PR #9573, opened 2024-10-02
[sd3] make sure height and size are divisible by `16`

yiyixuxu:
for `1920 X 1080` should get an input error and ask user to resize to `1920 x 1072` (on current main will get an error on scheduler step)

```
ValueError: `height` and `width` have to be divisible by 16 but are 1920 and 1080.You can use height 1920 and width 1072.
```

```python
import torch
import argparse

def main(height, width):
    device = torch.device("cuda:1")

    pipe = StableDiffusion3Pipeline.from_pretrained("stabilityai/stable-diffusion-3-medium-diffusers", torch_dtype=torch.float16)
    pipe = pipe.to(device)

    image = pipe(
        "A cat holding a sign that says hello world",
        negative_prompt="",
        num_inference_steps=28,
        height=height,
        width=width,
        guidance_scale=7.0,
        generator=generator,
    ).images[0]

    print(f"{height}x{width} -> image: {image.size}")
    image.save(f"yiyi_test_1_out_{height}_{width}.png")

if __name__ == "__main__":
    parser = argparse.ArgumentParser(description="Generate an image with Stable Diffusion 3")
    parser.add_argument("--height", type=int, default=1920, help="Height of the output image")
    parser.add_argument("--width", type=int, default=1080, help="Width of the output image")
    
    args = parser.parse_args()
    
    main(args.height, args.width)
```

### DIFF
--- a/src/diffusers/pipelines/controlnet_sd3/pipeline_stable_diffusion_3_controlnet_inpainting.py
+++ b/src/diffusers/pipelines/controlnet_sd3/pipeline_stable_diffusion_3_controlnet_inpainting.py
@@ -577,8 +577,14 @@ class StableDiffusion3ControlNetInpaintingPipeline(DiffusionPipeline, SD3LoraLoa
         callback_on_step_end_tensor_inputs=None,
         max_sequence_length=None,
     ):
-        if height % 8 != 0 or width % 8 != 0:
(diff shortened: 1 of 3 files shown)


asomoza (approved):
thanks, LGTM

PR #9580, opened 2024-10-04
FLUX: nonblocking and scheduler cpu indexing 

changjonathanc:
# What does this PR do?

- add nonblocking and avoids creating tensor on cpu then immediately move to device
- use cpu indexing in scheduler to avoid cuda sync/graph break 

## Who can review?
@a-r-r-o-w 

## Tests

main branch: there are 2 *visible* cuda sync
- first one happens here (it triggers cuda sync and has to wait for all preceding cuda ops to finish): https://github.com/huggingface/diffusers/blob/99f608218caa069a2f16dcf9efab46959b15aec0/src/diffusers/pipelines/flux/pipeline_flux.py#L363

- the second one happens inside of the first scheduler step:
https://github.com/huggingface/diffusers/blob/99f608218caa069a2f16dcf9efab46959b15aec0/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L288-L289

image

this branch: fixing both causes results in a nice profile view

image

profile is done by running the following code in a loop
```python
def bench_fn():
    image = pipe(prompt, num_inference_steps=4, width=1024, height=1024, output_type="pt")
    image = image.cpu()
```

### DIFF
--- a/src/diffusers/pipelines/flux/pipeline_flux.py
+++ b/src/diffusers/pipelines/flux/pipeline_flux.py
@@ -134,9 +134,7 @@ def retrieve_timesteps(
     else:
         scheduler.set_timesteps(num_inference_steps, device=device, **kwargs)
         timesteps = scheduler.timesteps
-    # a copy of timesteps that's always on cpu for indexing and code that requires cuda sync
-    timesteps_cpu = getattr(scheduler, "timesteps_cpu", None)
-    return timesteps, num_inference_steps, timesteps_cpu
+    return timesteps, num_inference_steps
 
 
 class FluxPipeline(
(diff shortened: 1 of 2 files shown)


yiyixuxu:
are you testing with flux pipeline? you can just add a set_begin_index from the pipeline to avoid that search step 

https://github.com/huggingface/diffusers/blob/acd6d2c42f0fa4fade262e8814279748a544b0ce/src/diffusers/pipelines/deepfloyd_if/pipeline_if.py#L677

cccntu:
Thanks for the pointer @yiyixuxu !, I updated the PR to use `set_begin_index`

a-r-r-o-w:
Hi, I apologies for the delay here. I have been profiling our pipelines and finding places for improvements continuosly to make Diffusers more production ready out-of-the-box. I will eventually club all PRs similar to this in the future and make sure that we don't have any synchronizations from start to end of pipeline execution

Issue #9534, opened 2024-09-26
[FluxMultiControlNetModel] object has no attribute 'config


darhsu:
### Describe the bug

This commit on main  https://github.com/huggingface/diffusers/commit/14a1b86fc7de53ff1dbf803f616cbb16ad530e45 seems to have broke FluxMultiControlNetModel. Reverting this commit fixes the issue on line `pipeline_flux_controlnet.py:844`

Referencing this issue:
https://huggingface.co/Shakker-Labs/FLUX.1-dev-ControlNet-Union-Pro/discussions/20

### Reproduction

```python
base_model = "/archive/me_flux_shallow/FLUX.1-dev"
controlnet_model_union = 'Shakker-Labs/FLUX.1-dev-ControlNet-Union-Pro'

controlnet_union = FluxControlNetModel.from_pretrained(controlnet_model_union, torch_dtype=torch.bfloat16)
controlnet = FluxMultiControlNetModel([controlnet_union])

pipe = FluxControlNetPipeline.from_pretrained(base_model, controlnet=controlnet, torch_dtype=torch.bfloat16)
pipe.to("cuda:0")

prompt = 'A bohemian-style female travel blogger with sun-kissed skin and messy beach waves.'
control_image_depth = load_image("/home/me/repos/flux_control_net/assets_depth.jpg")
control_mode_depth = 2

control_image_canny = load_image("/home/me/repos/flux_control_net/assets_canny.jpg")
control_mode_canny = 0

width, height = control_image_depth.size


image = pipe(
    prompt, 
    control_image=[control_image_depth, control_image_canny],
    control_mode=[control_mode_depth, control_mode_canny],
    width=width,
    height=height,
    controlnet_conditioning_scale=0.6,
    num_inference_steps=24, 
    guidance_scale=3.5,
    generator=torch.manual_seed(42),
).images[0]
```

### Logs

```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[8], line 1
----> 1 image = pipe(
      2     prompt, 
      3     control_image=[control_image_depth, control_image_canny],
      5     control_mode=[control_mode_depth, control_mode_canny],
      6     width=width,
      7     height=height,
      8     controlnet_conditioning_scale=0.6,
      9     num_inference_steps=24, 
     10     guidance_scale=3.5,
     11     generator=torch.manual_seed(42),
     12 ).images[0]

[...]

PR #9507, opened 2024-09-23
flux controlnet fix (control_modes batch & others)

yiyixuxu:
this PR:
1. fix `control_mode` shape when `num_images_per_prompt > 1`  (continue the PR from @christopher-beckham https://github.com/huggingface/diffusers/pull/9406)
2. fix a bug when multi-controlnet (follow up on from https://github.com/huggingface/diffusers/pull/9472) 
3. fix a bug when use multi-controlnet with regular controllers that may not have single blocks

slow tests here
```python

import torch
torch.cuda.set_device(2) 

# num_images_per_prompt = 1
num_images_per_prompt = 2
branch = "test" 
# branch = "main"

import gc
def flush():
    gc.collect()
    torch.cuda.empty_cache()

# test1: single controlnet (canny)
import torch
from diffusers.utils import load_image
from diffusers.pipelines.flux.pipeline_flux_controlnet import FluxControlNetPipeline
from diffusers.models.controlnet_flux import FluxControlNetModel

base_model = 'black-forest-labs/FLUX.1-dev'
controlnet_model = 'InstantX/FLUX.1-dev-Controlnet-Canny'
controlnet = FluxControlNetModel.from_pretrained(controlnet_model, torch_dtype=torch.bfloat16)
pipe = FluxControlNetPipeline.from_pretrained(base_model, controlnet=controlnet, torch_dtype=torch.bfloat16)
pipe.to("cuda")

control_image = load_image("https://huggingface.co/InstantX/FLUX.1-dev-Controlnet-Canny/resolve/main/canny.jpg")
prompt = "A girl in city, 25 years old, cool, futuristic"
generator = torch.Generator(device="cuda").manual_seed(42)
images_out = pipe(
    prompt, 
    control_image=control_image,
    controlnet_conditioning_scale=0.6,
    num_inference_steps=28, 
    num_images_per_prompt=num_images_per_prompt,
    guidance_scale=3.5,
    generator=generator,
).images
print(images_out)
for i, image in enumerate(images_out):
    image.save(f"yiyi_test_7_{branch}_num_images_per_prompt_{num_images_per_prompt}_test1_out_{i}.png")


del pipe
flush()

# test2: single controlnet (union)
import torch
from diffusers.utils import load_image
from diffusers import FluxControlNetPipeline, FluxControlNetModel

base_model = 'black-forest-labs/FLUX.1-dev'
controlnet_model = 'InstantX/FLUX.1-dev-Controlnet-Union'

[...]

PR #9586, opened 2024-10-04
config attribute not foud error for FluxImagetoImage Pipeline for multi controlnet solved

rshah240:
This  PR solves a bug in the file pipeline_flux_controlnet_image_to_image.py which was similar to this [issue](https://github.com/huggingface/diffusers/issues/9534)  I

# What does this PR do?

I made the similar changes to the pipeline_flux_controlnet_image_to_image.py by looking at the PR [#9507] .The changes are similar to those which are made by @yiyixuxu (https://github.com/huggingface/diffusers/pull/9507)


rshah240 (commented):
This is the same bug which was in the FluxControlNet pipeline. It was solved in the PR 9507. But the same bug was for FluxImg2ImgControlNet pipeline

yiyixuxu:
thanks! can we apply the same to inpaint controlnet too?

rshah240:
I have to check, I will test it and reply in this thread

yiyixuxu (approved):

PR #9614, opened 2024-10-09
Resolves [BUG] 'GatheredParameters' object is not callable

### LINKED ISSUE #9477: [BUG] 'GatheredParameters' object is not callable
### Describe the bug

'GatheredParameters' object is not callable  when we use zero3 in text-to-image.py

The `context_manager` has been initilized, why we need `with context_manager()`. I think it should be `with context_manager`. 
After replacing `with context_manager()` with `with context_manager`, the problem is solved.

https://github.com/huggingface/diffusers/commit/af2a237676ada656889de5e5b96ce609e37ed8c4#diff-5c07adf115dbb937f26dbfedff2aef0d31b4e8e3a58ba9cfb4a0c9f64672a8c5R213

### Reproduction

 when we use zero3 in text-to-image.py

### Logs

_No response_

### System Info

latest diffusers version

### Who can help?

@yiyixuxu @DN6 @sayakpaul

charchit7:
# What does this PR do?

Fixes #9477 

## Before submitting

## Who can review?
@sayakpaul @LinB203


sayakpaul (approved):
Thanks!

src/diffusers/training_utils.py
```diff
@@ -24,6 +24,9 @@
 if is_transformers_available():
     import transformers
 
+    if transformers.deepspeed.is_deepspeed_zero3_enabled():
```
a-r-r-o-w, line 27: `transformers.deepspeed` is deprecated and will be removed in a near future release. Could you look into the alternative `transformers.integrations`?

src/diffusers/training_utils.py
```diff
@@ -431,14 +434,12 @@ def step(self, parameters: Iterable[torch.nn.Parameter]):
...
         context_manager = contextlib.nullcontext
-        if is_transformers_available() and transformers.integrations.deepspeed.is_deepspeed_zero3_enabled():
-            import deepspeed
 
         if self.foreach:
             if is_transformers_available() and transformers.integrations.deepspeed.is_deepspeed_zero3_enabled():
                 context_manager = deepspeed.zero.GatheredParameters(parameters, modifier_rank=None)
 
-            with context_manager():
+            with context_manager:
```
a-r-r-o-w, line 442: This is correct for GatheredParameters but I think will fail in non-DeepSpeed scenarious where nullcontext is used. Could you create the context_manager object too?

```
context_manager = contextlib.nullcontext()
if deepspeed enabled:
    context_manager = GatheredParameters(...)
```

src/diffusers/training_utils.py
charchit7: Yes, give me sometime.

a-r-r-o-w (commented):
Thanks for fixing. But, now line 464 will fail which also needs to be taken into account, since it makes use of the same context_manager object.

Please change `with context_manager():` to `with context_manager:`.

These changes need to be made carefully and all easy to verify code paths should be tested, otherwise it will lead to many more bug reports down the line.

image

charchit7:
Got it, @a-r-r-o-w, thanks. 
I missed this one.

PR #9616, opened 2024-10-09
Slight performance improvement to `Euler`, `EDMEuler`, `FlowMatchHeun`, `KDPM2Ancestral`

hlky:
# What does this PR do?
This PR provides a slight performance improvement to `Euler`, `EDMEuler`, `FlowMatchHeun` by moving creation of `randn` tensor `noise` to inside the `gamma > 0` condition.

`gamma`, calculated as `min(s_churn / (len(self.sigmas) - 1), 2**0.5 - 1) if s_tmin <= sigma <= s_tmax else 0.0` is generally `0.0` as `s_churn`, `s_tmin` and `s_tmax` are typically not changed from their default values of `0.0`, `0.0` and `inf` respectively.

Similarly in `KDPM2Ancestral` we move creation of `noise` to inside the 2nd order path.

## Who can review?
@yiyixuxu


vladmandic:
@hlky love the attention you're giving schedulers.

btw, a bit off-topic, but though of mentioning it here:
this is executed on each step - and its definitely causing torch sync:
> sample = sample.to(torch.float32)

hlky:
Thanks! Looks like that cast and the mentioned precision issues comes from `rescale_betas_zero_snr` so perhaps the cast could be moved under `if self.config.rescale_betas_zero_snr`, or a [different value](https://github.com/huggingface/diffusers/blob/31058cdaef63ca660a1a045281d156239fba8192/src/diffusers/schedulers/scheduling_euler_discrete.py#L232-L235) like in [Comfy](https://github.com/comfyanonymous/ComfyUI/blob/203942c8b29dfbf59a7976dcee29e8ab44a1b32d/comfy_extras/nodes_model_advanced.py#L54-L71) and [Forge](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/7eb824a0daf2efe9e8e036a55e19de247f3f5d4d/modules/sd_models.py#L342-L358) could be used and the cast removed.

vladmandic:
> rescale_betas_zero_snr

that options is so rarely used that moving cast only if its true should be more than ok.

yiyixuxu (approved):
thanks!

yiyixuxu:
some of the tests failed as expected because the random seeds are used differently now
can we:
1. make a simple slow test that running a pipeline with each of these scheduler we changed and make sure the output is expected (even though it might be different from the main given same seed)
2. update the tests!

PR #9542, opened 2024-09-27
[LoRA] support Kohya Flux LoRAs that have text encoders as well

sayakpaul:
# What does this PR do?

https://huggingface.co/cocktailpeanut has a bunch of very nice Flux LoRAs that were trained using Kohya but has text encoder components too. This PR adds support for fully loading those LoRAs. 

Test code (has a slow test in this PR too):

```python
from diffusers import FluxPipeline
import torch 

pipeline = FluxPipeline.from_pretrained(
    "black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16
).to("cuda")

pipeline.load_lora_weights("cocktailpeanut/optimus", weight_name="optimus.safetensors")
  
prompts = [
    "optimus is cleaning the house with broomstick",
    "optimus is a DJ performing at a hip nightclub",
    "optimus is competing in a bboy break dancing competition",
    "optimus is playing tennis in a tennis court"
]
images = pipeline(
    prompts, 
    num_inference_steps=50,
    guidance_scale=3.5,
    max_sequence_length=512,
    generator=torch.manual_seed(0)
).images
for i, image in enumerate(images):
    image.save(f"{i}.png")
```

Favorite sample:
image

<sup>optimus is a DJ performing at a hip nightclub</sup>

PR #9618, opened 2024-10-09
[LoRA] make `set_adapters()` robust on silent failures.

sayakpaul:
# What does this PR do?

Currently, if we do 

```py
scales = {"text_encoder": 0.0, "text_encoder_2": 0.0, "unet": 0.0}
pipe.set_adapters("optimus", adapter_weights=scales)
```

where `pipe` is an instance of the `FluxPipeline` it doesn't error out whereas it should because Flux doesn't have any UNet and its `text_encoder_2` component isn't LoRA-loadable:

https://github.com/huggingface/diffusers/blob/31058cdaef63ca660a1a045281d156239fba8192/src/diffusers/loaders/lora_pipeline.py#L1650

Instead, we silently ignore things. This PR fixes this behavior. 

Thanks to @asomoza for the idea in https://github.com/huggingface/diffusers/pull/9542#issuecomment-2380411627!

### DIFF
--- a/src/diffusers/loaders/lora_base.py
+++ b/src/diffusers/loaders/lora_base.py
@@ -518,8 +518,18 @@ class LoraBaseMixin:
         adapter_names: Union[List[str], str],
         adapter_weights: Optional[Union[float, Dict, List[float], List[Dict]]] = None,
     ):
-        adapter_names = [adapter_names] if isinstance(adapter_names, str) else adapter_names
+        if isinstance(adapter_weights, dict):
+            components_passed = set(adapter_weights.keys())
+            lora_components = set(self._lora_loadable_modules)
 
+            invalid_components = components_passed - lora_components
+            if invalid_components:
+                raise ValueError(
+                    f"The following components in `adapter_weights` are not part of the pipeline: {invalid_components}. "
+                    f"Available components that are LoRA-compatible: {self._lora_loadable_modules}"
+                )
+
+        adapter_names = [adapter_names] if isinstance(adapter_names, str) else adapter_names
         adapter_weights = copy.deepcopy(adapter_weights)
 
         # Expand weights into a list, one entry per adapter
--- a/tests/lora/utils.py
+++ b/tests/lora/utils.py
@@ -1109,6 +1109,41 @@ class PeftLoraLoaderMixinTests:
         pipe.set_adapters("adapter-1")
         _ = pipe(**inputs, generator=torch.manual_seed(0))[0]
 
+    def test_multiple_wrong_adapter_name_raises_error(self):
+        scheduler_cls = self.scheduler_classes[0]
+        components, text_lora_config, denoiser_lora_config = self.get_dummy_components(scheduler_cls)
+        pipe = self.pipeline_class(**components)
+        pipe = pipe.to(torch_device)
+        pipe.set_progress_bar_config(disable=None)
+        _, _, inputs = self.get_dummy_inputs(with_generator=False)
+
+        if "text_encoder" in self.pipeline_class._lora_loadable_modules:
+            pipe.text_encoder.add_adapter(text_lora_config, "adapter-1")
+            self.assertTrue(check_if_lora_correctly_set(pipe.text_encoder), "Lora not correctly set in text encoder")
+
+        denoiser = pipe.transformer if self.unet_kwargs is None else pipe.unet
+        denoiser.add_adapter(denoiser_lora_config, "adapter-1")
+        self.assertTrue(check_if_lora_correctly_set(denoiser), "Lora not correctly set in denoiser.")
+
+        if self.has_two_text_encoders or self.has_three_text_encoders:
+            if "text_encoder_2" in self.pipeline_class._lora_loadable_modules:
+                pipe.text_encoder_2.add_adapter(text_lora_config, "adapter-1")
+                self.assertTrue(
+                    check_if_lora_correctly_set(pipe.text_encoder_2), "Lora not correctly set in text encoder 2"
+                )
+
+        scale_with_wrong_components = {"foo": 0.0, "bar": 0.0, "tik": 0.0}
+        with self.assertRaises(ValueError) as err_context:
+            pipe.set_adapters("adapter-1", adapter_weights=scale_with_wrong_components)
+
+        wrong_components = set(scale_with_wrong_components.keys())
+        msg = f"The following components in `adapter_weights` are not part of the pipeline: {wrong_components}"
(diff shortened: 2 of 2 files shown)


tests/lora/test_lora_layers_cogvideox.py
```diff
@@ -192,3 +192,7 @@ def test_simple_inference_with_text_lora_fused(self):
     @unittest.skip("Text encoder LoRA is not supported in CogVideoX.")
     def test_simple_inference_with_text_lora_save_load(self):
         pass
+
+    @unittest.skip("Not supported in CogVideoX.")
+    def test_simple_inference_with_text_denoiser_multi_adapter_block_lora(self):
+        pass
```
sayakpaul, line 198: Now that we're catching the error appropriately within the code, we should skip this test for unsupported models.

tests/lora/test_lora_layers_flux.py
```diff
@@ -162,6 +162,10 @@ def test_simple_inference_with_text_denoiser_block_scale_for_all_dict_options(se
     def test_modify_padding_mode(self):
         pass
 
+    @unittest.skip("Not supported in Flux.")
```
sayakpaul, line 165: Same as https://github.com/huggingface/diffusers/pull/9618/files#r1793336930.

tests/lora/test_lora_layers_sdxl.py
```diff
@@ -111,6 +112,10 @@ def tearDown(self):
         gc.collect()
         torch.cuda.empty_cache()
 
+    @is_flaky
```
sayakpaul, line 115: There's absolutely no reason for it to be flaky but I think okay for now.

tests/lora/test_lora_layers_sd3.py
```diff
@@ -108,3 +114,7 @@ def test_simple_inference_with_text_denoiser_block_scale_for_all_dict_options(se
     @unittest.skip("Not supported in SD3.")
     def test_modify_padding_mode(self):
         pass
+
+    @is_flaky
```
sayakpaul, line 118: Same as https://github.com/huggingface/diffusers/pull/9618#discussion_r1793367330.

tests/lora/utils.py
```diff
@@ -1109,6 +1109,41 @@ def test_wrong_adapter_name_raises_error(self):
...
+                self.assertTrue(
+                    check_if_lora_correctly_set(pipe.text_encoder_2), "Lora not correctly set in text encoder 2"
+                )
+
+        scale_with_wrong_components = {"foo": 0.0, "bar": 0.0, "tik": 0.0}
+        with self.assertRaises(ValueError) as err_context:
+            pipe.set_adapters("adapter-1", adapter_weights=scale_with_wrong_components)
+
+        wrong_components = set(scale_with_wrong_components.keys())
+        msg = f"The following components in `adapter_weights` are not part of the pipeline: {wrong_components}"
```
BenjaminBossan, line 1140: I wonder if this check is robust: The wrong components are coming from a set where order is not guaranteed, so the message could contain the keys in a different order. My approach is generally to sort the keys in the error message.

tests/lora/utils.py
```diff
@@ -1109,6 +1109,41 @@ def test_wrong_adapter_name_raises_error(self):
...
+        self.assertTrue(msg in str(err_context.exception))
+
```
BenjaminBossan, line 1142: Should this also include a test with some wrong and some correct components?

BenjaminBossan (approved):
Good idea to check this, LGTM. I have some suggestions for the test, please check if they make sense.

Apart from that, I'm wondering if we still need the checks starting here:

https://github.com/huggingface/diffusers/pull/9618/files#diff-005343920b6fa6c3f4c1f639ee23f180dfa424d6008e487fdc9962a82f498d15R568

Or are those now covered by the newly introduced check and can be removed?

tests/lora/utils.py
sayakpaul: Does 47ef80234 work for you?

sayakpaul:
@BenjaminBossan thanks!

> Apart from that, I'm wondering if we still need the checks starting here:

Good catch. Resolved in 271404336.

tests/lora/utils.py
BenjaminBossan: Yes, works. We could remove the set here since keys are already unique, but no big deal.

tests/lora/utils.py
BenjaminBossan: This would mainly be a test that the error does not accidentally contain valid components. This is really just a nit though.

sayakpaul:
@DN6 could you give this a look?

sayakpaul:
@DN6 a gentle ping.

sayakpaul:
@yiyixuxu @DN6 a gentle ping.

sayakpaul:
@asomoza @yiyixuxu @DN6 could you give this a look?

tests/lora/utils.py
```diff
@@ -1124,6 +1124,41 @@ def test_wrong_adapter_name_raises_error(self):
...
+        if self.has_two_text_encoders or self.has_three_text_encoders:
+            if "text_encoder_2" in self.pipeline_class._lora_loadable_modules:
+                pipe.text_encoder_2.add_adapter(text_lora_config, "adapter-1")
+                self.assertTrue(
+                    check_if_lora_correctly_set(pipe.text_encoder_2), "Lora not correctly set in text encoder 2"
+                )
+
+        scale_with_wrong_components = {"foo": 0.0, "bar": 0.0, "tik": 0.0}
+        with self.assertRaises(ValueError) as err_context:
```
DN6, line 1151: Should just checking warning is raised.

sayakpaul:
@DN6 I have changed from raising error to raising warnings. Additionally, we're removing the invalid components from the `adapter_weights`. LMK what you think.

sayakpaul:
@hlky could you give this a look?

src/diffusers/loaders/lora_base.py
```diff
@@ -562,12 +574,6 @@ def set_adapters(
             for adapter_name, weights in zip(adapter_names, adapter_weights):
                 if isinstance(weights, dict):
                     component_adapter_weights = weights.pop(component, None)
-
-                    if component_adapter_weights is not None and not hasattr(self, component):
-                        logger.warning(
-                            f"Lora weight dict contains {component} weights but will be ignored because pipeline does not have {component}."
```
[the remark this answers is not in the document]
sayakpaul: Yeah but `adapter_names` and `adapter_weights` formulation was wrong.

hlky (approved):

PR #9626, opened 2024-10-09
Adaptive Projected Guidance

### LINKED ISSUE #9585: APG: Eliminating Oversaturation and Artifacts of High Guidance Scales in Diffusion Models
New paper just released, a new way to do classifier-free guidance (CFG) without affecting image saturation, introducing artifacts, etc. Could be huge.

https://huggingface.co/papers/2410.02416

hlky:
# What does this PR do?

This PR implements APG (Adaptive Projected Guidance) from Algorithm 1 in [`Eliminating Oversaturation and Artifacts of High Guidance Scales in Diffusion Models`](https://arxiv.org/pdf/2410.02416).

Algorithm 1 is slightly modified to combine `project` into `normalized_guidance`, this simply reduces the number of methods to be copied between pipelines.

APG is added to `StableDiffusionPipeline` and `StableDiffusionXLPipeline`. The following parameters are introduced:
```
adaptive_projected_guidance (`bool`, *optional*):
    Use adaptive projected guidance from [Eliminating Oversaturation and Artifacts of High Guidance Scales
    in Diffusion Models](https://arxiv.org/pdf/2410.02416)
adaptive_projected_guidance_momentum (`float`, *optional*, defaults to `-0.5`):
    Momentum to use with adaptive projected guidance. Use `None` to disable momentum.
adaptive_projected_guidance_rescale_factor (`float`, *optional*, defaults to `15.0`):
    Rescale factor to use with adaptive projected guidance.
```

Default values are taken from Stable Diffusion XL in Table 10. The existing `eta` parameter is used, rather than adding a new parameter, as per the docstring `eta` is only used in `DDIMScheduler` so this should be ok.

Fixes #9585

Example usage:
```python
import torch

pipe = StableDiffusionXLPipeline.from_pretrained(
    variant="fp16",
    torch_dtype=torch.float16,
)
pipe.safety_checker = None
pipe.enable_vae_tiling()
prompt = "A 4k dslr photo of a raccoon wearing an astronaut helmet, photorealistic."

generator = torch.Generator().manual_seed(694208027600)
image = pipe(prompt, guidance_scale=15, generator=generator).images[0]
image

generator = torch.Generator().manual_seed(694208027600)
image = pipe(
    prompt,
    guidance_scale=15,
    adaptive_projected_guidance=True,
    adaptive_projected_guidance_momentum=-0.5,
    adaptive_projected_guidance_rescale_factor=15.0,
    generator=generator,
).images[0]
image
```

CFG:
original

APG:
apg

There's certainly some improvement, however further testing would be beneficial to confirm the findings in the paper.

## Who can review?

cc @yiyixuxu @asomoza


Msadat97:
Thanks for your interest in our work! Please note that we always convert the output of the model to the denoised predictions (pred_x0) and compute the guidance there. We found that APG performs better when applied to the denoised predictions. We also have a discussion on this step in Section 5.2 (Figure 12).

hlky:
@Msadat97 Thanks! I missed that section. Applying the guidance to the denoised predictions does indeed produce much better results:
apg

@yiyixuxu After fixing this locally for denoised predictions I'd have to agree, it needs a little more than present here, specifically we'd actually need some changes to schedulers to allow easy conversion between noised and denoised predictions. Making guidance a separate component sounds like a great idea, hope to see that in soon, I'd be happy to work with you on that and any changes to schedulers.

hlky:
Certainly, I've pushed those changes, however please note this will only work with some schedulers like Euler, and while it does work for 2nd order schedulers like Heun and DPM2 it's not 100% as the `sigma` used is incorrect for the 2nd order step, it won't work for schedulers like DDIM. The issue linked above aims to add methods to handle this for each scheduler.

yiyixuxu:
hi @hlky just so you know I made a Guider class in here and this is the use case I have in mind to try next https://github.com/huggingface/diffusers/pull/9672

PR #9627, opened 2024-10-09
[LoRA] fix dora test to catch the warning properly.

sayakpaul:


tests/lora/test_lora_layers_sdxl.py
```diff
@@ -619,12 +619,12 @@ def test_integration_logits_multi_adapter(self):
     @nightly
     def test_integration_logits_for_dora_lora(self):
         pipeline = StableDiffusionXLPipeline.from_pretrained("stabilityai/stable-diffusion-xl-base-1.0")
-        pipeline.load_lora_weights("hf-internal-testing/dora-trained-on-kohya")
-        pipeline.enable_model_cpu_offload()
 
         logger = logging.get_logger("diffusers.loaders.lora_pipeline")
         logger.setLevel(30)
         with CaptureLogger(logger) as cap_logger:
+            pipeline.load_lora_weights("hf-internal-testing/dora-trained-on-kohya")
```
sayakpaul, line 626: The warning gets raised when we load the DoRA weights. So, we need to put this line inside the context manager.

yiyixuxu (approved):

Issue #9629, opened 2024-10-09
[Schedulers] Add methods for conversion between noise and denoised predictions


hlky:
Guidance methods like APG (#9585 #9626) require conversion between noise and denoised predictions. The computation of these predictions varies across schedulers and depends on `prediction_type`. Currently, this is embedded in `step` functions of the schedulers which complicates implementation of APG.

Adding separate methods for conversion between noise and denoised predictions would simplify implementation and allow easier experimentation.

yiyixuxu:
Do you need a function like this we have for dpm? https://github.com/huggingface/diffusers/blob/07bd2fabb60dbd1c7da6fd176c62e592402b19e3/src/diffusers/schedulers/scheduling_dpmsolver_multistep.py#L594

hlky:
Yes like that, there's currently also something similar in UniPC, they'd need to be controllable with a parameter rather than config or internal variables, and handle different sigmas for 2nd order schedulers like Heun and KDPM2

yiyixuxu:
Make sense! 

Do you want to open up a PR for UniPC? We'd need to keep this method consistent across all schedulers that have them, so we need to use the same name and signature and return the same output etc 

For the rest of the schedulers, we can also open this task up to communities if you don't have much time: we would need to run slow tests for each scheduler we refactor here, so it might be a bit time-consuming for one person to work on!

hlky:
Sure, I'll start with UniPC. I don't mind doing the rest of the schedulers, I don't have much else going on, more community contributions in schedulers would be good though!

PR #9636, opened 2024-10-10
Add methods for conversion between noise and denoised predictions to UniPC

hlky:
# What does this PR do?

This PR adds `convert_noise_to_x0` and `convert_x0_to_noise` methods to UniPC scheduler for converting to original sample x0 from noise prediction and converting to noise prediction from original sample x0 respectively.

A test, `test_convert_model_output` is added, which performs a full loop, converting model output with `convert_noise_to_x0`, then back to noised prediction with `convert_x0_to_noise` and compares that result to model output. In the `v_prediction` case comparison is done at the end of the loop because the required tolerance seems to change per step.

`convert_noise_to_x0` and `convert_x0_to_noise` are added as separate methods than the existing `convert_model_output` for a few reasons:
1. `convert_model_output` has deprecated `timestep` in favor of internal `step_index`, however, conversion methods could be used before `step` is called when `step_index` isn't set. Conversion methods accept `timestep` and call `_init_step_index(timestep)` when `self.step_index is None`.
2. Using `convert_model_output` for the conversion resulted in numerical inaccuracies, I noticed a potential discrepancy in `convert_model_output` compared to the working `convert_x0_to_noise`, 
```python
# convert_model_output
if self.predict_x0:
    if self.config.prediction_type == "epsilon":
        x0_pred = (sample - sigma_t * model_output) / alpha_t
    elif self.config.prediction_type == "sample":
        x0_pred = model_output
...
else:
    if self.config.prediction_type == "epsilon":
        return model_output
    elif self.config.prediction_type == "sample":
        epsilon = (sample - alpha_t * model_output) / sigma_t
        return epsilon
```
```python
# convert_x0_to_noise
if self.config.prediction_type == "epsilon":
    x0_pred = (sample - alpha_t * pred_x0) / sigma_t
elif self.config.prediction_type == "sample":
    x0_pred = pred_x0
```
`convert_model_output` with `prediction_type == "epsilon"` returns `model_output` when it should possibly be returning `(sample - alpha_t * model_output) / sigma_t` like in `convert_x0_to_noise`. `predict_x0=False` path appears to be untested/unusued, so I'm not sure whether there's a mistake in `convert_model_output` or whether it's just specific to the scheduler's needs.

I'd like to get more eyes on this to decide whether `convert_model_output` needs fixing or whether to stick with separate methods. Input on the currently deprecated usage of timestep would also be useful.

Also as this is being implemented for guidance methods like APG, I have tested `convert_noise_to_x0` and `convert_x0_to_noise` in a pipeline with APG, the image result is as expected.

#9629

## Who can review?

@yiyixuxu

### DIFF
--- a/src/diffusers/schedulers/scheduling_unipc_multistep.py
+++ b/src/diffusers/schedulers/scheduling_unipc_multistep.py
@@ -527,6 +527,67 @@ class UniPCMultistepScheduler(SchedulerMixin, ConfigMixin):
         )
         return sigmas
 
+    def convert_noise_to_x0(self, model_output: torch.Tensor, sample: torch.Tensor, timestep: int) -> torch.Tensor:
+        """
+        Convert to original sample x0 from noise prediction.
+
+        Args:
+            model_output (`torch.Tensor`): The model output.
+            sample (`torch.Tensor`): A current instance of a sample created by the diffusion process.
+            timestep (`int`): The current discrete timestep in the diffusion chain.
+
+        Returns:
+            `torch.Tensor`: The predicted original sample (x0).
+        """
+        if self.step_index is None:
+            self._init_step_index(timestep)
+        sigma = self.sigmas[self.step_index]
+        alpha_t, sigma_t = self._sigma_to_alpha_sigma_t(sigma)
+
+        if self.config.prediction_type == "epsilon":
+            return (sample - sigma_t * model_output) / alpha_t
+        elif self.config.prediction_type == "sample":
+            return model_output
+        elif self.config.prediction_type == "v_prediction":
+            return alpha_t * sample - sigma_t * model_output
+        else:
+            raise ValueError(
+                f"prediction_type given as {self.config.prediction_type} must be one of `epsilon`, `sample`, or"
+                " `v_prediction`."
+            )
+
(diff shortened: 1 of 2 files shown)


yiyixuxu:
I think we can just add a step_index argument to the `convert_model_output` function

hlky:
I've pushed a version using `convert_model_output`, note that the tests will fail, I hope this will better demonstrate the issue with `convert_model_output`.

1. Convert model output to x0, this works
https://github.com/huggingface/diffusers/blob/951517873fd703b431d65d949fd9e9a747757fd4/tests/schedulers/test_scheduler_unipc.py#L300
2. Convert x0 to noise prediction, this doesn't work
https://github.com/huggingface/diffusers/blob/951517873fd703b431d65d949fd9e9a747757fd4/tests/schedulers/test_scheduler_unipc.py#L301
because it just returns `model_output`, which is still the x0
https://github.com/huggingface/diffusers/blob/951517873fd703b431d65d949fd9e9a747757fd4/src/diffusers/schedulers/scheduling_unipc_multistep.py#L589-L590

Perhaps another way to state the issue is that `convert_model_output` is a one-way conversion and we need two-way conversion.

PR #9646, opened 2024-10-11
[flux lora training] make LoRA target modules configurable

linoytsaban:
new feature for the Flux dreambooth lora training script:

-make LoRA target modules configurable through `--lora_blocks`

- change the current default target modules to not be attention layers only (?)

### DIFF
--- a/examples/dreambooth/train_dreambooth_lora_flux.py
+++ b/examples/dreambooth/train_dreambooth_lora_flux.py
@@ -555,6 +555,13 @@ def parse_args(input_args=None):
     )
 
     parser.add_argument(
+        "--lora_blocks",
+        type=str,
+        default=None,
+        help=('The transformer modules to apply LoRA training on. Please specify the layers in a comma seperated. E.g. - "q_proj,k_proj,v_proj,out_proj" will result in lora training of attention layers only'),
+    )
+
+    parser.add_argument(
         "--adam_epsilon",
         type=float,
         default=1e-08,
@@ -1188,12 +1195,18 @@ def main(args):
         if args.train_text_encoder:
             text_encoder_one.gradient_checkpointing_enable()
 
-    # now we will add new LoRA weights to the attention layers
+    if args.lora_blocks is not None:
+        target_modules = [block.strip() for block in args.lora_blocks.split(",")]
+    else:
+        target_modules = ["to_k", "to_q", "to_v", "to_out.0",
+                        "add_k_proj", "add_q_proj", "add_v_proj", "to_add_out", "ff.net.0.proj","ff.net.2", "ff_context.net.0.proj","ff_context.net.2"]
+
+    # now we will add new LoRA weights the transformer layers
     transformer_lora_config = LoraConfig(
         r=args.rank,
         lora_alpha=args.rank,
         init_lora_weights="gaussian",
-        target_modules=["to_k", "to_q", "to_v", "to_out.0"],
+        target_modules=target_modules,
     )
     transformer.add_adapter(transformer_lora_config)
     if args.train_text_encoder:


examples/dreambooth/train_dreambooth_flux.py
```diff
@@ -161,7 +161,7 @@ def log_validation(
         f"Running validation... \n Generating {args.num_validation_images} images with prompt:"
         f" {args.validation_prompt}."
     )
-    pipeline = pipeline.to(accelerator.device, dtype=torch_dtype)
```
sayakpaul, line 164: Let's provide the author courtesy here.

examples/dreambooth/train_dreambooth_lora_flux.py
```diff
@@ -554,6 +554,15 @@ def parse_args(input_args=None):
...
     )
 
+    parser.add_argument(
+        "--lora_layers",
+        type=str,
+        default=None,
+        help=(
+            'The transformer modules to apply LoRA training on. Please specify the layers in a comma seperated. E.g. - "to_k,to_q,to_v,to_out.0" will result in lora training of attention layers only'
+        ),
+    )
```
sayakpaul, line 564: Could do this with `nargs`. Better no?

examples/dreambooth/train_dreambooth_lora_flux.py
```diff
@@ -1186,12 +1195,30 @@ def main(args):
...
+            "attn.to_out.0",
+            "attn.add_k_proj",
+            "attn.add_q_proj",
+            "attn.add_v_proj",
+            "attn.to_add_out",
+            "ff.net.0.proj",
+            "ff.net.2",
+            "ff_context.net.0.proj",
+            "ff_context.net.2",
+        ]
```
sayakpaul, line 1214: Seems like a bit breaking no? Better to not do it and instead make a note from the README? 

WDYT?

sayakpaul (approved):
Thanks! Minor notes.

examples/dreambooth/train_dreambooth_lora_flux.py
apolinario: Breaking or just changing default behavior? I think it's geared more towards the latter, but I think it's in line with the other trainers & makes sense for Transformer based models, so maybe a  `Warning` note and a guide on how to train it the old way for e.g.?

sayakpaul: Yeah maybe a warning note at the beginning of the README should cut it. 

With this change, we're likely also increasing the total training wall-clock time so, that is worth noting.

examples/dreambooth/train_dreambooth_flux.py
sayakpaul: @linoytsaban did we resolve this?

examples/dreambooth/test_dreambooth_lora_flux.py
```diff
@@ -136,6 +136,42 @@ def test_dreambooth_lora_latent_caching(self):
...
+                --instance_data_dir {self.instance_data_dir}
+                --instance_prompt {self.instance_prompt}
+                --resolution 64
+                --train_batch_size 1
+                --gradient_accumulation_steps 1
+                --max_train_steps 2
+                --cache_latents
+                --learning_rate 5.0e-04
+                --scale_lr
+                --lora_layers single_transformer_blocks.0.attn.to_k
```
sayakpaul, line 153: (nit): could make "single_transformer_blocks.0.attn.to_k" a constant and then supply it here. This way we know what we're testing for immediately. WDYT?

sayakpaul (approved):
Thanks! Minor comments.

PR #9691, opened 2024-10-16
[CI] add a big GPU marker to run memory-intensive tests separately on CI

sayakpaul:
# What does this PR do?

I have only touched a handful of tests with the marker being introduced. I think we may need to change the slices based on the CI machine and infra. @a-r-r-o-w should consider marking the Cog tests similarly as well? 

@DN6 would love to get your thoughts on the design.

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -2,6 +2,7 @@ name: Nightly and release tests on main/release branch
 
 on:
   workflow_dispatch:
+  pull_request:
   schedule:
     - cron: "0 0 * * *" # every day at midnight
 
@@ -18,6 +19,7 @@ env:
 
 jobs:
   setup_torch_cuda_pipeline_matrix:
+    if: github.event_name == 'schedule'
     name: Setup Torch Pipelines CUDA Slow Tests Matrix
     runs-on:
       group: aws-general-8-plus
@@ -49,6 +51,7 @@ jobs:
           path: reports
 
   run_nightly_tests_for_torch_pipelines:
+    if: github.event_name == 'schedule'
     name: Nightly Torch Pipelines CUDA Tests
     needs: setup_torch_cuda_pipeline_matrix
     strategy:
@@ -106,6 +109,7 @@ jobs:
           python utils/log_reports.py >> $GITHUB_STEP_SUMMARY
 
   run_nightly_tests_for_other_torch_modules:
+    if: github.event_name == 'schedule'
     name: Nightly Torch CUDA Tests
     runs-on:
       group: aws-g4dn-2xlarge
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -180,6 +180,61 @@ jobs:
         pip install slack_sdk tabulate
         python utils/log_reports.py >> $GITHUB_STEP_SUMMARY
 
+  run_big_gpu_torch_tests:
+    name: Torch tests on big GPU (24GB)
+    strategy:
+      fail-fast: false
+      max-parallel: 8
+    runs-on:
+      group: aws-g6-4xlarge-plus
+    container:
+      image: diffusers/diffusers-pytorch-cuda
+      options: --shm-size "16gb" --ipc host --gpus 0
+    steps:
+      - name: Checkout diffusers
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - name: NVIDIA-SMI
+        run: nvidia-smi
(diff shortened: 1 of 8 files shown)


.github/workflows/nightly_tests.yml
```diff
@@ -2,6 +2,7 @@ name: Nightly and release tests on main/release branch
 
 on:
   workflow_dispatch:
+  pull_request:
```
sayakpaul, line 5: This is temporary.

.github/workflows/nightly_tests.yml
```diff
@@ -18,6 +19,7 @@ env:
 
 jobs:
   setup_torch_cuda_pipeline_matrix:
+    if: github.event_name == 'schedule'
```
sayakpaul, line 22: Temporary.

a-r-r-o-w:
> should consider marking the Cog tests similarly as well?

With model cpu offload and vae tiling, it should be < 16 GB, and I think we documented it [here](https://huggingface.co/docs/diffusers/main/en/api/pipelines/cogvideox#memory-optimization). Are we seeing Cog test failures due to memory? I see that they are passing [here](https://github.com/huggingface/diffusers/actions/runs/11361614795/job/31601801453)

sayakpaul:
Ah okay then. No issues.

.github/workflows/nightly_tests.yml
```diff
@@ -180,6 +184,61 @@ jobs:
         pip install slack_sdk tabulate
         python utils/log_reports.py >> $GITHUB_STEP_SUMMARY
 
+  run_big_gpu_torch_tests:
+    name: Torch tests on big GPU (24GB)
+    strategy:
+      fail-fast: false
+      max-parallel: 8
```
DN6, line 191: Would take this down to 2 so that we don't grab a bunch of big GPUs from the pool. Since we're running nightly and currently only a few models fall into this category it should be enough.

DN6 (approved):
Nice work! 👍🏽

sayakpaul:
@DN6 okay if I modified the failing tests to account for the machine change?

src/diffusers/utils/testing_utils.py
```diff
@@ -54,6 +54,7 @@
 ) > version.parse("4.33")
 
 USE_PEFT_BACKEND = _required_peft_version and _required_transformers_version
+BIG_GPU_MEMORY = 40
```
DN6, line 57: Would make configurable from the workflow file. 
```suggestion
BIG_GPU_MEMORY = os.getenv("BIG_GPU_MEMORY, 40)
```

.github/workflows/nightly_tests.yml
```diff
@@ -180,6 +184,61 @@ jobs:
         pip install slack_sdk tabulate
         python utils/log_reports.py >> $GITHUB_STEP_SUMMARY
 
+  run_big_gpu_torch_tests:
+    name: Torch tests on big GPU (24GB)
```
DN6, line 188: Will we use 24GB or 40? (The decorator has a hard coded value, which is why I'm asking?)

sayakpaul: 40GB. Will update.

sayakpaul:
@DN6 can you give this a look? I think the test failures should go away once the CI Bot has access to Flux. 

Once approved I will revert the changes which I have denoted as temporary (like [this](https://github.com/huggingface/diffusers/pull/9691/#discussion_r1802535985)).

tests/pipelines/controlnet_flux/test_controlnet_flux_img2img.py
```diff
@@ -222,70 +218,3 @@ def test_fused_qkv_projections(self):
         assert np.allclose(
             original_image_slice, image_slice_disabled, atol=1e-2, rtol=1e-2
         ), "Original outputs should match when fused QKV projections are disabled."
-
-
-@slow
-@require_torch_gpu
-class FluxControlNetImg2ImgPipelineSlowTests(unittest.TestCase):
```
sayakpaul, line 229: I don't think this test was correctly done as it doesn't pass the `controlnet` module to the pipeline and it also uses very dummy inputs which I think should be avoided for an integration test. LMK if you think otherwise.

sayakpaul:
@DN6 regarding https://github.com/huggingface/diffusers/actions/runs/11398910357/job/31716739483?pr=9691#step:7:67, my hunch is that there's some kind of leakage happening which is causing the worker to crash. When I SSH'd into the runner and manually ran the test, it passed.

DN6 (approved):

PR #9693, opened 2024-10-16
Added asserts and fixes for empty emb input in AdaLayerNorm

Godofnothing:
This PR raises asserts when conditioning is not provided to AdaLayerNorm. 

Specifically:
1) One has to provide either timestep or temb argument to `AdaLayerNorm`.
2) Modulation embedding has to be provided either via CombinedTimestepLabelEmbeddings or emb argument in `AdaLayerNormZero`.
3) `AdaLayerNormZeroSingle` handles the case when number of channels in `emb` is not equal to `x`.
4) `emb` argument in `AdaLayerNormZeroSingle` becomes mandatory (currently, code would crash if `emb` is None).


yiyixuxu:
hi:
is change this needed for a custom model you're making?

PR #9712, opened 2024-10-18
[feat] add `load_lora_adapter()` for compatible models

sayakpaul:
# What does this PR do?

Similar to `load_attn_procs()`, we want to have something similar for loading LoRAs into models, as the LoRA loading logic is generic. 

This way, we can reduce the LoC and have better maintainability. I am not too fixated on the `load_lora_adapter()` name. Could also do `load_adapter()`. 

@DN6 as discussed via Slack, could you give this a check? Could also add a `save_lora_adapter()` method to complement.


src/diffusers/loaders/lora_base.py
```diff
@@ -181,6 +181,123 @@ def _remove_text_encoder_monkey_patch(text_encoder):
         text_encoder._hf_peft_config_loaded = None
 
 
+def _fetch_state_dict(
```
sayakpaul, line 184: Just taking it out of the class to be able to better reuse.

DN6 (approved):
Nice clean up 👍🏽

src/diffusers/loaders/lora_base.py
```diff
@@ -233,126 +350,6 @@ def _optionally_disable_offloading(cls, _pipeline):
 
         return (is_model_cpu_offload, is_sequential_cpu_offload)
 
-    @classmethod
-    def _fetch_state_dict(
```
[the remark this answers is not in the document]
sayakpaul: Valid. I deprecated and added tests.

DN6 (approved):

DN6 (approved):

sayakpaul:
@BenjaminBossan could you give this a look too?

src/diffusers/loaders/lora_base.py
```diff
@@ -234,124 +350,16 @@ def _optionally_disable_offloading(cls, _pipeline):
...
-                )
-                state_dict = load_state_dict(model_file)
-        else:
-            state_dict = pretrained_model_name_or_path_or_dict
-
-        return state_dict
+    def _fetch_state_dict(cls, *args, **kwargs):
+        deprecation_message = f"Using the `_fetch_state_dict()` method from {cls} has been deprecated and will be removed in a future version. Please use `from diffusers.loaders.lora_base import _fetch_state_dict`."
+        deprecate("_fetch_state_dict", "0.35.0", deprecation_message)
+        _fetch_state_dict(*args, **kwargs)
```
BenjaminBossan, line 356:
```suggestion
        return _fetch_state_dict(*args, **kwargs)
```

or not?

src/diffusers/loaders/lora_base.py
```diff
@@ -234,124 +350,16 @@ def _optionally_disable_offloading(cls, _pipeline):
...
-        if len(targeted_files) > 1:
-            raise ValueError(
-                f"Provided path contains more than one weights file in the {file_extension} format. Either specify `weight_name` in `load_lora_weights` or make sure there's only one  `.safetensors` or `.bin` file in  {pretrained_model_name_or_path_or_dict}."
-            )
-        weight_name = targeted_files[0]
-        return weight_name
+    def _best_guess_weight_name(cls, *args, **kwargs):
+        deprecation_message = f"Using the `_best_guess_weight_name()` method from {cls} has been deprecated and will be removed in a future version. Please use `from diffusers.loaders.lora_base import _best_guess_weight_name`."
+        deprecate("_best_guess_weight_name", "0.35.0", deprecation_message)
+        _best_guess_weight_name(*args, **kwargs)
```
BenjaminBossan, line 362:
```suggestion
        return _best_guess_weight_name(*args, **kwargs)
```
same

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -1250,13 +1258,17 @@ def load_lora_weights(
...
             raise ValueError("Invalid LoRA checkpoint.")
 
-        self.load_lora_into_transformer(
-            state_dict,
-            transformer=getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer,
-            adapter_name=adapter_name,
-            _pipeline=self,
-            low_cpu_mem_usage=low_cpu_mem_usage,
-        )
+        transformer_state_dict = {k: v for k, v in state_dict.items() if "transformer." in k}
```
BenjaminBossan, line 1261: For my understanding: This is a fix independent of the main change of this PR, right? Would it be possible to move this check inside of `load_lora_into_transformer` or would that not be a good idea?

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -341,7 +339,9 @@ def load_lora_into_text_encoder(
             adapter_name (`str`, *optional*):
                 Adapter name to be used for referencing the loaded adapter model. If not specified, it will use
                 `default_{i}` where i is the total number of adapters being loaded.
-            Speed up model loading by only loading the pretrained LoRA weights and not initializing the random weights.:
+            low_cpu_mem_usage (`bool`, *optional*):
```
BenjaminBossan, line 342: I saw that too, thx for fixing.

BenjaminBossan (changes requested):
Thanks for this refactor, always happy to see more lines being removed than added. I didn't check the functions that were moved around, as I assume they were left identical. Regarding the rest, just some smaller comments.

src/diffusers/loaders/lora_pipeline.py
sayakpaul: I think it kind of depends on the entrypoint to the underlying method. 

We already have a similar check within `load_lora_adapter()`:
https://github.com/huggingface/diffusers/blob/e187b706a546a54c95abffb951f0ec0292400627/src/diffusers/loaders/peft.py#L153

So, I think it should be fine, without.

BenjaminBossan (approved):
Thanks for addressing my comments, LGTM.

sayakpaul:
Ran the Flux integration tests and they pass. Failing tests are unrelated.

PR #9709, opened 2024-10-18
[CI] pin max torch version to fix CI errors

a-r-r-o-w:
# What does this PR do?

Only pinned until we figure out what's going on with torch 2.5.0

Context: https://huggingface.slack.com/archives/C065E480NN9/p1729230272869569

#9691, #9704

- https://github.com/huggingface/diffusers/actions/runs/11398910363/job/31721831848?pr=9691
- https://github.com/huggingface/diffusers/actions/runs/11398910363/job/31721832317?pr=9691

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #9720, opened 2024-10-19
`make  deps_table_update` to fix CI tests

a-r-r-o-w:
# What does this PR do?

We missed this step (I did not know we had to do it) in #9709. Also contains a dummy change for now to trigger CI fast tests

## Who can review?

@sayakpaul @DN6

### DIFF
--- a/.github/workflows/pr_test_peft_backend.yml
+++ b/.github/workflows/pr_test_peft_backend.yml
@@ -94,11 +94,11 @@ jobs:
         python -m uv pip install -e [quality,test]
         if [ "${{ matrix.lib-versions }}" == "main" ]; then
             python -m pip install -U peft@git+https://github.com/huggingface/peft.git --no-deps
-            python -m uv pip install -U transformers@git+https://github.com/huggingface/transformers.git
-            pip uninstall accelerate -y && python -m uv pip install -U accelerate@git+https://github.com/huggingface/accelerate.git
+            python -m uv pip install -U transformers@git+https://github.com/huggingface/transformers.git --no-deps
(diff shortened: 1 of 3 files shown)


a-r-r-o-w:
@sayakpaul, seems like capping the torch version in diffusers to `<2.5.0` does not fix the failing tests, because 2.5 gets installed by peft (see [here](https://github.com/huggingface/diffusers/actions/runs/11420646495/job/31776602477?pr=9720#step:4:76)).

sayakpaul:
I am not sure if that is the case because from the logs of that action, I don't see any reinstallations of PyTorch, though. 

To confirm things, I would try print the PyTorch version just after the container is initialized. After this line:
https://github.com/huggingface/diffusers/blob/56d6d21baeb203a249b3049835916bfb0bfad061/.github/workflows/pr_test_peft_backend.yml#L94

If it's 2.5.0, then there's something wrong.

sayakpaul:
Hmm I am pretty sure that the updated Torch is coming from https://github.com/huggingface/diffusers/blob/56d6d21baeb203a249b3049835916bfb0bfad061/docker/diffusers-pytorch-cpu/Dockerfile#L32

So, we need to:

* Ping torch versions in the Dockerfiles 
* Verify Dockerfiles build 
* Trigger a Docker build and push workflow 

Just pinning in the setup.py file won't cut it and my bad it went unnoticed.

sayakpaul:
PR: https://github.com/huggingface/diffusers/pull/9709

Workflow: https://github.com/huggingface/diffusers/actions/runs/11423669772

DN6 (approved):

PR #9736, opened 2024-10-21
[core] Allegro T2V

a-r-r-o-w:
# What does this PR do?

Model: https://huggingface.co/rhymes-ai/Allegro
Github: https://github.com/rhymes-ai/Allegro


src/diffusers/models/autoencoders/autoencoder_kl_allegro.py
```diff
@@ -0,0 +1,1024 @@
...
+        else:
+            self.upsamplers = None
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        batch_size = hidden_states.shape[0]
+
+        for resnet, temp_conv in zip(self.resnets, self.temp_convs):
+            hidden_states = hidden_states.permute(0, 2, 1, 3, 4).flatten(0, 1)
+            hidden_states = resnet(hidden_states, temb=None)
+            hidden_states = hidden_states.unflatten(0, (batch_size, -1)).permute(0, 2, 1, 3, 4)
```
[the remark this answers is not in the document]
a-r-r-o-w: I think the current implementation is much cleaner, no? Maybe we could do a refactor of the existing `TemporalConvLayer` in a future PR? Either okay with me, so just need your confirmation before committing to this (as there are many instances of this and using 3 lines instead of 1 will look the code look a bit more bloatified)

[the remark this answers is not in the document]
a-r-r-o-w: Or did you mean to make this change just in the temporal layer, while leaving the other instances of unflatten/permute untouched?

```python
        hidden_states = (
            hidden_states[None, :].reshape((-1, num_frames) + hidden_states.shape[1:]).permute(0, 2, 1, 3, 4)
        )
```

src/diffusers/models/autoencoders/autoencoder_kl_allegro.py
```diff
@@ -0,0 +1,1058 @@
...
+        else:
+            self.downsamplers = None
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        batch_size = hidden_states.shape[0]
+
+        for resnet, temp_conv in zip(self.resnets, self.temp_convs):
+            hidden_states = hidden_states.permute(0, 2, 1, 3, 4).flatten(0, 1)
+            hidden_states = resnet(hidden_states, temb=None)
+            hidden_states = hidden_states.unflatten(0, (batch_size, -1)).permute(0, 2, 1, 3, 4)
```
yiyixuxu, line 203: like here, whenever it has a temp_conv layer, it has to apply this step, I think it is better we move this logic inside temp_conv so we only have to deal it there (once)

src/diffusers/models/autoencoders/autoencoder_kl_allegro.py
```diff
@@ -0,0 +1,1058 @@
...
+        self.resnets = nn.ModuleList(resnets)
+        self.temp_convs = nn.ModuleList(temp_convs)
+        self.attentions = nn.ModuleList(attentions)
+
+    def forward(self, hidden_states: torch.Tensor):
+        batch_size = hidden_states.shape[0]
+
+        hidden_states = hidden_states.permute(0, 2, 1, 3, 4).flatten(0, 1)
+        hidden_states = self.resnets[0](hidden_states, temb=None)
+        hidden_states = hidden_states.unflatten(0, (batch_size, -1)).permute(0, 2, 1, 3, 4)
```
yiyixuxu, line 403: here too

src/diffusers/models/autoencoders/autoencoder_kl_allegro.py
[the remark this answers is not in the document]
yiyixuxu: yeah I just glance through the code, I think this line  `hidden_states.unflatten(0, (batch_size, -1)).permute(0, 2, 1, 3, 4)` is everywhere, and basically everytime you call the temp conv layer, you need to do that - so it is easier to just process inside the temp conv I think  https://github.com/huggingface/diffusers/pull/9736/files#r1809793907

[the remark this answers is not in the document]
a-r-r-o-w: will do next, thanks for the review! this was just a 1:1 conversion of the original code using einops so i didn't try to get clever and break anything

a-r-r-o-w:
It looks like something broke when doing the VAE refactor - looking into it at the moment. Will fix the broken tests afterwards

a-r-r-o-w:
Update: not a bug in the VAE - made a mistake in the transformer. Doing a more careful run through it again. @yiyixuxu `repeat_interleave` and pad temporal changes are in and work as expected, thanks!

src/diffusers/pipelines/allegro/pipeline_allegro.py
```diff
@@ -0,0 +1,936 @@
...
+        latents = latents * self.scheduler.init_noise_sigma
+        return latents
+
+    def decode_latents(self, latents: torch.Tensor) -> torch.Tensor:
+        latents = 1 / self.vae.config.scaling_factor * latents
+        frames = self.vae.decode(latents).sample
+        frames = frames.permute(0, 2, 1, 3, 4)  # [batch_size, channels, num_frames, height, width]
+        return frames
+
+    def _prepare_rotary_positional_embeddings(
```
DN6, line 620: Not a blocker to merge. 

We currently have a mix of creating rotary embeddings like this in a few pipelines (Cog, Lumina, Hunyuan) 
Was there a specific reason I missed to go this route as opposed to creating a dedicated layer in the transformer (Flux)? Is it because we need access to height, width etc to create the embedding

tests/pipelines/allegro/test_allegro.py
```diff
@@ -0,0 +1,311 @@
...
+
+        inputs["callback_on_step_end"] = callback_inputs_change_tensor
+        inputs["callback_on_step_end_tensor_inputs"] = pipe._callback_tensor_inputs
+        output = pipe(**inputs)[0]
+        assert output.abs().sum() < 1e10
+
+    def test_inference_batch_single_identical(self):
+        self._test_inference_batch_single_identical(batch_size=3, expected_max_diff=1e-3)
+
+    # def test_attention_slicing_forward_pass(
```
DN6, line 209: Let's just mark this with a skip

tests/pipelines/allegro/test_allegro.py
```diff
@@ -0,0 +1,311 @@
...
+            height=720,
+            width=1280,
+            num_frames=88,
+            generator=generator,
+            num_inference_steps=2,
+            output_type="pt",
+        ).frames
+
+        video = videos[0]
+        expected_video = torch.randn(1, 88, 720, 1280, 3).numpy()
```
DN6, line 308: I assume this will be updated to a real video?

tests/pipelines/allegro/test_allegro.py
```diff
@@ -0,0 +1,311 @@
...
+        super().setUp()
+        gc.collect()
+        torch.cuda.empty_cache()
+
+    def tearDown(self):
+        super().tearDown()
+        gc.collect()
+        torch.cuda.empty_cache()
+
+    def test_cogvideox(self):
```
DN6, line 290:
```suggestion
    def test_allegro(self):
```

DN6 (approved):

DN6:
LGTM. There's a failing test that looks related to saving/loading the transformer.

src/diffusers/models/transformers/transformer_allegro.py
```diff
@@ -0,0 +1,422 @@
...
+            Whether or not to use bias in the attention projection layers.
+        sample_height (`int`, defaults to `90`):
+            The height of the input latents.
+        sample_width (`int`, defaults to `160`):
+            The width of the input latents.
+        sample_frames (`int`, defaults to `22`):
+            The number of frames in the input latents.
+        activation_fn (`str`, defaults to `"gelu-approximate"`):
+            Activation function to use in feed-forward.
+        norm_elementwise_affine (`bool`, defaults to `True`):
```
stevhliu, line 210:
```suggestion
        norm_elementwise_affine (`bool`, defaults to `False`):
```

src/diffusers/models/transformers/transformer_allegro.py
```diff
@@ -0,0 +1,422 @@
...
+            Whether or not to use elementwise affine in normalization layers.
+        norm_eps (`float`, defaults to `1e-5`):
```
stevhliu, line 212:
```suggestion
        norm_eps (`float`, defaults to `1e-6`):
```

src/diffusers/pipelines/allegro/pipeline_allegro.py
```diff
@@ -0,0 +1,936 @@
...
+                The prompt or prompts not to guide the image generation. If not defined, one has to pass
+                `negative_prompt_embeds` instead. Ignored when not using guidance (i.e., ignored if `guidance_scale` is
+                less than `1`).
+            num_inference_steps (`int`, *optional*, defaults to 100):
+                The number of denoising steps. More denoising steps usually lead to a higher quality image at the
+                expense of slower inference.
+            timesteps (`List[int]`, *optional*):
+                Custom timesteps to use for the denoising process. If not defined, equal spaced `num_inference_steps`
+                timesteps are used. Must be in descending order.
+            guidance_scale (`float`, *optional*, defaults to 7.0):
```
stevhliu, line 720:
```suggestion
            guidance_scale (`float`, *optional*, defaults to 7.5):
```

src/diffusers/pipelines/allegro/pipeline_allegro.py
```diff
@@ -0,0 +1,936 @@
...
+                Guidance scale as defined in [Classifier-Free Diffusion Guidance](https://arxiv.org/abs/2207.12598).
+                `guidance_scale` is defined as `w` of equation 2. of [Imagen
+                Paper](https://arxiv.org/pdf/2205.11487.pdf). Guidance scale is enabled by setting `guidance_scale >
+                1`. Higher guidance scale encourages to generate images that are closely linked to the text `prompt`,
+                usually at the expense of lower image quality.
+            num_images_per_prompt (`int`, *optional*, defaults to 1):
+                The number of images to generate per prompt.
```
stevhliu, line 727:
```suggestion
            num_videos_per_prompt (`int`, *optional*, defaults to 1):
                The number of videos to generate per prompt.
```

stevhliu (approved):
Very nice, just a few typos :)

src/diffusers/models/autoencoders/autoencoder_kl_allegro.py
```diff
@@ -0,0 +1,1155 @@
...
+            return (posterior,)
+        return AutoencoderKLOutput(latent_dist=posterior)
+
+    def _decode(self, z: torch.Tensor) -> torch.Tensor:
+        # TODO(aryan): refactor tiling implementation
+        # if self.use_tiling and (width > self.tile_latent_min_width or height > self.tile_latent_min_height):
+        if self.use_tiling:
+            return self.tiled_decode(z)
+
+        raise NotImplementedError("Decoding without tiling has not been implemented yet.")
```
a-r-r-o-w, line 888: @yiyixuxu Is this okay for now? There are some followups that we could look into later and maybe rewrite the tiling implementation similar to our other VAEs.

I don't think the model works well with lower number of frames (in which case not using tiling would be faster when decoding), so we should probably always just use tiling since we have 88 frames as the default (and recommended)

src/diffusers/models/normalization.py
```diff
@@ -266,6 +263,7 @@ def forward(
         hidden_dtype: Optional[torch.dtype] = None,
     ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
         # No modulation happening here.
+        added_cond_kwargs = added_cond_kwargs or {"resolution": None, "aspect_ratio": None}
```
yiyixuxu, line 266: oh what is this? is this fixing a current bug?

a-r-r-o-w: Yep. We accept `None` in `added_cond_kwargs` here, but we actually need to pass values for `resolution` and `aspect_ratio` in the following pixart embedding layer (which requires it as non-defaulted arg)

yiyixuxu (approved):
thanks!
we can merge once the tests are fixed 
I can look into refactoring the 3d rope in a follow up PR (or if you want to leave this a little bit longer it is ok too, up to you!)

Ednaordinary:
I have something really weird happening here. Feeding in a list of prompts (even if the list is only one prompt) results in a really bad video. It might be related to https://github.com/huggingface/diffusers/pull/9769#discussion_r1817614450, but I can't be sure. It seems cfg baked, so it would make sense if it was a cfg error.

prompt = ["Orbital shot of a squirrel nibbles on a nut while sitting in a tree"]:

https://github.com/user-attachments/assets/11eaab98-9be6-44d9-8904-76e762d91c60


prompt = "Orbital shot of a squirrel nibbles on a nut while sitting in a tree":


https://github.com/user-attachments/assets/13d15a6d-de81-4abe-b077-2e15beb7774c

DN6:
@Ednaordinary Can you share your code snippet? Including how you're loading the pipeline? I'm unable to reproduce the issue.

Ednaordinary:
I checked further, and it happens when the prompt and negative prompt are length one lists (even with the list being [None], I think), but not when the prompt is a list and negative is unspecified (my mistake). I have yet to test further

Ednaordinary:
@foreverpiano I don't believe so, as passing None in a list also seems to trigger it. It also looks suspiciously like cfg baking, which I can't be certain but I feel as if negative prompting with nothing wouldn't cause

a-r-r-o-w:
@Ednaordinary I think should be fixed with the latest commit. LMK if it still persists - if so, will fix in a follow-up PR

src/diffusers/pipelines/allegro/pipeline_allegro.py
a-r-r-o-w: Yes, this is on my mind. Will take up dedicated RoPE layer refactor for existing models that do it in the pipeline in a future PR

PR #9737, opened 2024-10-22
[Fix] Using sharded checkpoints with gated repositories

asomoza:
# What does this PR do?

When using sharded checkpoints with gated repositories we get an error because we don't pass the token to the `model_info` function. This PR fixes this.

## Who can review?

@sayakpaul @DN6


sayakpaul (approved):
Thank you!

PR #9746, opened 2024-10-23
fix bug in `require_accelerate_version_greater`

faaany:
## What does this PR do?
We should check `accelerate` rather than `peft`. I think this is a typo. Please have a review. 

@sayakpaul


sayakpaul (approved):
Thanks very much!

sayakpaul:
Failing test is unrelated.

PR #9749, opened 2024-10-23
[SD3-5 dreambooth lora] update model cards

linoytsaban:
small updates to model card saving


sayakpaul (approved):
Uff! The tags are so spot on!

PR #9758, opened 2024-10-23
Improve control net block index for sd3

linjiapro:
# What does this PR do?

The layer configuration for the Control Net in Stable Diffusion 3 models must adhere to the rule that the total number of layers in the SD3 model should be a multiple of the Control Net's layer count.

For SD3.5, which has 38 layers, the Control Net can only have three possible options: 2, 19, or 38. This leads to inefficiencies in the setup.

Also, qk_norm, context_pre_only_last_layer, dual_attention_layers are added to match the transformer architecture

## Who can review?

@sayakpaul @yiyixuxu @DN6


linjiapro:
cc @sayakpaul @yiyixuxu @DN6

linjiapro:
This is a very simple PR, but for some reason, the tests all failed for wired reasons such as:

Unable to find self-hosted runner group: 'aws-general-8-plus'.

sayakpaul:
Thanks for your contributions! Could you maybe also add a test for this?

linjiapro:
@sayakpaul 

There is an existing test for the pipeline of SD3 control net, I leveraged that. The layer number for the control net changed from 1 to 3, the number of layer of the transformer (4) is no longer the multiples of the number of layers of control net. This will test the code changes of this PR.

.gitignore
```diff
@@ -102,6 +102,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+myenv/
```
sayakpaul, line 105: Should be removed.

tests/pipelines/controlnet_sd3/test_controlnet_sd3.py
```diff
@@ -77,7 +77,7 @@ def get_dummy_components(self):
             sample_size=32,
             patch_size=1,
             in_channels=8,
-            num_layers=1,
+            num_layers=3,
```
sayakpaul, line 80: We should change this value here I think. Instead, we could make this method accept an argument like `num_controlnet_layers` and then leverage it as needed. WDYT?

sayakpaul (commented):
Thanks, I left some comments.

.gitignore
linjiapro: Done

tests/pipelines/controlnet_sd3/test_controlnet_sd3.py
linjiapro: Done

linjiapro:
@sayakpaul @yiyixuxu, shall we merge this?

bghira:
maybe @DN6 is more active.

src/diffusers/models/transformers/transformer_sd3.py
```diff
@@ -344,7 +345,8 @@ def custom_forward(*inputs):
 
             # controlnet residual
             if block_controlnet_hidden_states is not None and block.context_pre_only is False:
-                interval_control = len(self.transformer_blocks) // len(block_controlnet_hidden_states)
+                interval_control = len(self.transformer_blocks) / len(block_controlnet_hidden_states)
```
yiyixuxu, line 348: why are we making this change? it is not the same so a breaking change, no?

linjiapro: @yiyixuxu Good question. 

The revised code adapts the strategy used by ControlNet for Flux, introducing a significant improvement in flexibility. Here's why this change matters:

In the old code,  the number of transformer layers is divisible by the number of ControlNet layers. For example, with SD3.5 Large, which has 38 transformer layers, there were only two valid options for the number of ControlNet layers: 2 and 19. Setting the number of ControlNet layers to anything else, such as 5, would cause the old code to crash.

However, the Flux ControlNet approach removes this restriction, allowing greater flexibility in choosing the number of layers. The revised logic essentially mirrors the Flux implementation, enabling more versatile configurations.

Importantly, the new code maintains compatibility with existing setups. If the number of transformer layers is divisible by the number of ControlNet layers, the interval_control remains unchanged, ensuring all previous configurations continue to function seamlessly.

src/diffusers/models/controlnets/controlnet_sd3.py
```diff
@@ -248,7 +254,7 @@ def from_transformer(
         config = transformer.config
         config["num_layers"] = num_layers or config.num_layers
         config["extra_conditioning_channels"] = num_extra_conditioning_channels
-        controlnet = cls(**config)
+        controlnet = cls.from_config(**config)
```
yiyixuxu, line 257:
```suggestion
        controlnet = cls.from_config(config)
```

src/diffusers/models/transformers/transformer_sd3.py
yiyixuxu: thanks! I think it's indeed better, I'm just wondering if it would cause issue for controlnet is trained with the current logic
cc @haofanwang here

linjiapro: I don't think it will cause any issue with the trained controlnet using old code before this PR. 

The reason is that for the controlnet to be trained with the old code, the number of layers of the transformer has to be divisible by the number of layers of the controlnet, and the new logic after this PR does not change the behavior.

yiyixuxu (approved):

yiyixuxu:
can you run `make style` and `make fix-copies`?

PR #9784, opened 2024-10-28
[fix] sync ControlNet-SD3 attention parameters of dual_attention_layer and qk_norm with SD3

huchenz1:
When initializing ControlNet-SD3 with the from_transformer method, the dual_attention_layers and qk_norm parameters were not defined in the __init__ method, resulting in the following error:
```
[rank1]:   File "train_controlnet_sd3.py", line 989, in main
[rank1]:     controlnet = SD3ControlNetModel.from_transformer(transformer)
[rank1]:   File "/home/23124047r/anaconda3/envs/py38/lib/python3.8/site-packages/diffusers/models/controlnet_sd3.py", line 251, in from_transformer
[rank1]:     controlnet = cls(**config)
[rank1]:   File "/home/23124047r/anaconda3/envs/py38/lib/python3.8/site-packages/diffusers/configuration_utils.py", line 665, in inner_init
[rank1]:     init(self, *args, **init_kwargs)
[rank1]: TypeError: __init__() got an unexpected keyword argument 'dual_attention_layers'
```
To resolve this, the missing keyword arguments were added to ensure compatibility with the from_transformer method.

And pos_embed_input for conditioning images needs to be the same as it in original transformers, for the images shape are the same.

### DIFF
--- a/src/diffusers/models/controlnets/controlnet_sd3.py
+++ b/src/diffusers/models/controlnets/controlnet_sd3.py
@@ -56,6 +56,10 @@ class SD3ControlNetModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOriginal
         out_channels: int = 16,
         pos_embed_max_size: int = 96,
         extra_conditioning_channels: int = 0,
+        dual_attention_layers: Tuple[
+            int, ...
+        ] = (),  # () for sd3.0; (0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12) for sd3.5
+        qk_norm: Optional[str] = None,
     ):
         super().__init__()
         default_out_channels = in_channels
@@ -84,6 +88,8 @@ class SD3ControlNetModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOriginal
                     num_attention_heads=num_attention_heads,
                     attention_head_dim=self.config.attention_head_dim,
                     context_pre_only=False,
+                    qk_norm=qk_norm,
+                    use_dual_attention=True if i in dual_attention_layers else False,
                 )
                 for i in range(num_layers)
             ]
@@ -243,7 +249,7 @@ class SD3ControlNetModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOriginal
 
     @classmethod
     def from_transformer(
-        cls, transformer, num_layers=12, num_extra_conditioning_channels=1, load_weights_from_transformer=True
+        cls, transformer, num_layers=12, num_extra_conditioning_channels=0, load_weights_from_transformer=True
     ):
         config = transformer.config
         config["num_layers"] = num_layers or config.num_layers


src/diffusers/models/controlnet_sd3.py
```diff
@@ -99,7 +103,7 @@ def __init__(
             height=sample_size,
             width=sample_size,
             patch_size=patch_size,
-            in_channels=in_channels + extra_conditioning_channels,
+            in_channels=in_channels,
```
yiyixuxu, line 106: Is this change not related to SD 3.5?

huchenz1: It is releated to whether the control net is used for inpainting. 
The control image for inpainiting has additional mask channel, but the normal controlnet does not need it. [src/diffusers/pipelines/controlnet_sd3/pipeline_stable_diffusion_3_controlnet_inpainting.py](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/controlnet_sd3/pipeline_stable_diffusion_3_controlnet_inpainting.py#L746)

yiyixuxu: why are we removing this then? we can set it to 0 when it is not needed?

huchenz1: Sorry, you are right.  I cancelled the commit and I changed the  the default num_extra_conditioning_channels to 0 in the from_transformer method such that when using [script](https://github.com/huggingface/diffusers/blob/main/examples/controlnet/train_controlnet_sd3.py), erros will not occur.

src/diffusers/models/controlnet_sd3.py
```diff
@@ -241,9 +247,10 @@ def _set_gradient_checkpointing(self, module, value=False):
         if hasattr(module, "gradient_checkpointing"):
             module.gradient_checkpointing = value
 
+    # If controlnet is used for inpainting, the num_extra_conditioning_channels should be 1.
     @classmethod
     def from_transformer(
-        cls, transformer, num_layers=12, num_extra_conditioning_channels=1, load_weights_from_transformer=True
+        cls, transformer, num_layers=12, num_extra_conditioning_channels=0, load_weights_from_transformer=True
```
yiyixuxu, line 253: do we have to change the default? you can manually pass `num_extra_conditioning_channels=0` when you need it to be `0`, no?

otherwise I will tag the original author to see if it is ok to change here (to make sure it won't break anything)

huchenz1: Because the method in training [script ](https://github.com/huggingface/diffusers/blob/main/examples/controlnet/train_controlnet_sd3.py#L986) use the default parameters and the script is for add control with normal image instead of inpainting.  As the result, with num_extra_conditioning_channels=0 in this method, [script ](https://github.com/huggingface/diffusers/blob/main/examples/controlnet/train_controlnet_sd3.py#L986) will occur errors.
In addtion, the default value of num_extra_conditioning_channels in [__ini__](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/controlnets/controlnet_sd3.py#L58) method is also 0.

PR #9791, opened 2024-10-29
[fix] Correct progress bar initialization in train_instruct_pix2pix.py

GhostCai:
This PR ensures the progress bar starts from 0 when resuming from a checkpoint, fixing an issue where it previously started at global_step and caused wrong progress tracking.

### DIFF
--- a/examples/instruct_pix2pix/train_instruct_pix2pix.py
+++ b/examples/instruct_pix2pix/train_instruct_pix2pix.py
@@ -842,7 +842,7 @@ def main():
             resume_step = resume_global_step % (num_update_steps_per_epoch * args.gradient_accumulation_steps)
 
     # Only show the progress bar once on each machine.
-    progress_bar = tqdm(range(global_step, args.max_train_steps), disable=not accelerator.is_local_main_process)
+    progress_bar = tqdm(range(0, args.max_train_steps), disable=not accelerator.is_local_main_process)
     progress_bar.set_description("Steps")
 
     for epoch in range(first_epoch, args.num_train_epochs):


bghira:
this isn't correct. it's supposed to resume from global_step

GhostCai:
> this isn't correct. it's supposed to resume from global_step

Hi, thank you for the feedback! I believe there’s still an issue—the progress bar is updated within the resume-skipping section of the code (L855). To resolve this, we could either remove the progress bar update in that section or adjust its range accordingly. 

This issue was already addressed in `train_instruct_pix2pix_sdxl.py` by setting the full range, but it seems the fix wasn’t applied here somehow.

bghira:
the other scripts iirc are simply jumping straight to the resume range position, and then `continue` inside the loop is antiquated and should be removed instead

GhostCai:
Exactly. Using continue in this way is safer but can lead to unnecessary data loading. In this code, both a shortened range and continue are used together, resulting in an incorrect range. For example, if I want to train for 20k steps and resume from an 8k checkpoint, the range is set to range(8k, 20k). However, after all the continue statements, the current iteration counter updates to 16k, leaving only 4k steps instead of the expected 12k for training. So I think we should adjust the range.

bghira:
well, no - remove the continue statement, so that this script becomes consistent across the project's other script examples.

GhostCai:
OK that can solve the problem too. I’ve removed the continue statement and set the initial value to global_step instead.

examples/instruct_pix2pix/train_instruct_pix2pix.py
```diff
@@ -851,8 +851,6 @@ def collate_fn(examples):
         for step, batch in enumerate(train_dataloader):
             # Skip steps until we reach the resumed step
             if args.resume_from_checkpoint and epoch == first_epoch and step < resume_step:
-                if step % args.gradient_accumulation_steps == 0:
-                    progress_bar.update(1)
```
sayakpaul, line 855: Why are we removing this?

GhostCai: Because we have set `initial=global_step` in the progress_bar so it will be consistent with `train_instruct_pix2pix_sdxl.py`'s progress_bar setup? As is proposed by bghira in [here](https://github.com/huggingface/diffusers/pull/9791#issuecomment-2444592784). If we don't remove this statement, the initial should be set to zero.

sayakpaul:
Can we try to harmonize the changes by following what we do here?
https://github.com/huggingface/diffusers/blob/5588725e8e7be497839432e5328c596169385f16/examples/text_to_image/train_text_to_image_sdxl.py#L1058

I think it's better to have better consistency that way.

GhostCai:
> Can we try to harmonize the changes by following what we do here?
> 
> https://github.com/huggingface/diffusers/blob/5588725e8e7be497839432e5328c596169385f16/examples/text_to_image/train_text_to_image_sdxl.py#L1058
> 
> I think it's better to have better consistency that way.

Yes, we can work towards harmonizing the changes. The main difference here lies in whether or not we skip the loop N times. This file does while the SDXL file doesn't. However, based on our [previous discussion](https://github.com/huggingface/diffusers/pull/9791#issuecomment-2444565399), skipping the loop N times in this implementation is needed for maintaining reproducibility in training. So, to ensure consistency, it may actually be the SDXL version that needs to be aligned with this approach. Or, if we believe accelerate is robust enough, we could remove the continue in this file to match the SDXL version. Thoughts?

sayakpaul:
Thanks for explaining further! 

Could you please provide a visual example comparing the changes you would like to see if it's not too much?

sayakpaul:
Gentle ping @GhostCai

PR #9796, opened 2024-10-29
[training] use the lr when using 8bit adam.

sayakpaul:
# What does this PR do?

Oops.

@a-r-r-o-w FYI.


linoytsaban (approved):
Nice catch!

sayakpaul:
@linoytsaban sorry I think I went too soon with the PR. I realized that we ALWAYS pack "lr" in `params_to_optimize` irrespective of the optimizer class we're using. 

So, I think you were right not to pass the `lr`. However, when using "prodigy" we were still passing `lr`. So, to remove the inconsistency, I omitted those instances in this PR. Could you please give this another look? 

Cc: @a-r-r-o-w for the changed Cog scripts.

sayakpaul:
Correct. And this PR resolves that.

linoytsaban (approved):

PR #9812, opened 2024-10-30
fix `EMAModel.use_ema_warmup`

chenguolin:
# What does this PR do?

I do not understand for a long time why `use_ema_warmup` is forced to be `True` in `EMAModel.__init__()`. It is really inconvenient for really proferming fixed-decay EMA.

And it is also unreasonable to still use warmup by `cur_decay_value = (1 + step) / (10 + step)` when `use_ema_warmup=False`.

## Before submitting

## Who can review?

General functionalities: @sayakpaul @yiyixuxu @DN6

### DIFF
--- a/src/diffusers/training_utils.py
+++ b/src/diffusers/training_utils.py
@@ -341,9 +341,6 @@ class EMAModel:
             )
             parameters = parameters.parameters()
 
-            # set use_ema_warmup to True if a torch.nn.Module is passed for backwards compatibility
-            use_ema_warmup = True
-
         if kwargs.get("max_value", None) is not None:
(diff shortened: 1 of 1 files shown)


sayakpaul:
I think this becomes a breaking change for folks using this class. It's better to deprecate the bevahiour.

chenguolin:
Maybe adding a deprecating logging? The current decay behavior is unreasonable for me:
1. we can't set `self. use_ema_warmup` in the EMAModel initialization, which means input arg `use_ema_warmup` is useless;
2. EMA decay is still warmup when `self.use_ema_warmup=False`.

sayakpaul:
Deprecation is the way to go. We cannot introduce a breaking change like this.

PR #9817, opened 2024-10-31
[Flux] reduce explicit device transfers and typecasting in flux.

sayakpaul:
# What does this PR do?

Reduces explicit device transfers and type-casting in the Flux bunch of pipelines.


yiyixuxu (approved):
thanks!

PR #9820, opened 2024-10-31
Add Dynamic Device and DType to Initial DiffusionPipeline Example

DorsaRoh:
# What does this PR do?

In line with [PHILOSOPHY.md](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md) of prioritizing usability and accessibility, this update enables DiffusionPipeline to automatically select cuda with float16 precision if a GPU is available, otherwise defaults to cpu with float32. Since this is the first piece of code new users are introduced to in the entire repo, ensuring it works seamlessly across different hardware configurations is vital for an inclusive user experience. This enhancement maintains our commitment to making the library accessible to all without adding extra dependencies.

## Before submitting

## Who can review?

@stevhliu  @sayakpaul @yiyixuxu  and @asomoza

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,12 @@ Generating outputs is super easy with 🤗 Diffusers. To generate an image from 
 from diffusers import DiffusionPipeline
 import torch
 
-pipeline = DiffusionPipeline.from_pretrained("stable-diffusion-v1-5/stable-diffusion-v1-5", torch_dtype=torch.float16)
-pipeline.to("cuda")
+pipeline = DiffusionPipeline.from_pretrained(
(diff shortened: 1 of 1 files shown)


vladmandic:
what about rocm, directml, zluda, ipex, openvino, etc.?
imo - they all need special case and falling back to cpu simply because cuda is not available is worse than erroring out.

a-r-r-o-w:
I don't think the minimal front-page example for how to use diffusers should contain anything more than a simple snippet that would work for most users of the library out of the box. Closing PR for same reason. Feel free to suggest changes to docs in the optimization/backend section if something similar would be applicable there and not already implicitly conveyed, or require explicitly mentioning relevant info

PR #9449, opened 2024-09-17
[LoRA] fix: lora loading when using with a device_mapped model.

sayakpaul:
# What does this PR do?

Fixes LoRA loading behaviour when used with a model that is sharded into multiple devices. 

<details>
<summary>Minimal code</summary>

```python
"""
Minimal example to show how to load a LoRA into the Flux transformer
that is sharded in two GPUs. 

Limitation:
* Latency
* If the LoRA has text encoder layers then this needs to be revisited.
"""

from diffusers import FluxTransformer2DModel, FluxPipeline 
import torch 

ckpt_id = "black-forest-labs/FLUX.1-dev"
dtype = torch.bfloat16
transformer = FluxTransformer2DModel.from_pretrained(
    ckpt_id, 
    subfolder="transformer",
    device_map="auto",
    max_memory={0: "16GB", 1: "16GB"},
    torch_dtype=dtype
)
print(transformer.hf_device_map)
pipeline = FluxPipeline.from_pretrained(
    ckpt_id,
    text_encoder=None,
    text_encoder_2=None,
    tokenizer=None,
    tokenizer_2=None,
    vae=None,
    transformer=transformer,
    torch_dtype=dtype
)
pipeline.load_lora_weights("TheLastBen/Jon_Snow_Flux_LoRA", weight_name="jon_snow.safetensors")
# print(pipeline.transformer.hf_device_map)

# Essentially you'd pre-compute these embeddings beforehand.
# Reference: https://gist.github.com/sayakpaul/a9266fe2d0d510ec44a9cdc385b3dd74. 
example_inputs = {
    "prompt_embeds": torch.randn(1, 512, 4096, dtype=dtype, device="cuda"),
    "pooled_projections": torch.randn(1, 768, dtype=dtype, device="cuda"),
}

_ =  pipeline(
    prompt_embeds=example_inputs["prompt_embeds"],
    pooled_prompt_embeds=example_inputs["pooled_projections"],
    num_inference_steps=50,
    guidance_scale=3.5,
    height=1024,
    width=1024,
    output_type="latent",
)
```

</details>

Some internal discussions:

* https://huggingface.slack.com/archives/C03UQJENJTV/p1725527760353639
* https://huggingface.slack.com/archives/C04L3MWLE6B/p1726470631333599

Cc: @philschmid for awareness as you were interested in this feature. 

## TODOs

- [x] Tests
- [x] Docs

Once I get a sanity review from Marc and Benjamin, will request a review from Yiyi.

PR #9821, opened 2024-10-31
[device_map] fix device_map check behaviour.

sayakpaul:
# What does this PR do?

Follow-up of https://github.com/huggingface/diffusers/pull/9449. It introduced a bug when doing:
https://github.com/huggingface/diffusers/blob/41e4779d988ead99e7acd78dc8e752de88777d0f/src/diffusers/pipelines/pipeline_utils.py#L412

For models that have sharded checkpoints (such as `FluxTransformer2DModel`), we will always set `device_map`:
https://github.com/huggingface/diffusers/blob/41e4779d988ead99e7acd78dc8e752de88777d0f/src/diffusers/models/modeling_utils.py#L883  

So, this PR:

1. When a device map only has CPU we need to account for it. We also add a test for it in this PR. 
2. Reuse `model_has_device_map()` functionality.


sayakpaul:
Unrelated test failures.

sayakpaul:
Closing in light of https://github.com/huggingface/diffusers/pull/9827.

PR #9824, opened 2024-10-31
Add Diffusion Policy for Reinforcement Learning

DorsaRoh:
## What does this PR do?

Adds Diffusion Policy, a diffusion model to predict action sequences in reinforcement learning tasks, using the HuggingFace `diffusers` library. 

## Before submitting

## Who can review?

@sayakpaul @yiyixuxu @DN6 @a-r-r-o-w 

---


sayakpaul:
It seems like we're putting in the content of some repository within `diffusers`. I think it might be better off as a standalone repository than an example here as we're not a training focused library.

DorsaRoh:
> It seems like we're putting in the content of some repository within `diffusers`. I think it might be better off as a standalone repository than an example here as we're not a training focused library.

Thank you for the feedback! I have made the changes. Now, it includes only an inference example of using diffusers for diffusion policy

examples/reinforcement_learning/diffusion_policy.py
```diff
@@ -0,0 +1,69 @@
...
+            model_input = torch.cat([action, cond], dim=1)
+            
+            # predict and remove noise
+            noise_pred = self.model(model_input, t).sample
+            action = self.noise_scheduler.step(noise_pred, t, action).prev_sample
+        
+        return action.transpose(1, 2)  # [batch_size, sequence_length, action_dim]
+
+if __name__ == "__main__":
+    policy = DiffusionPolicy()
```
sayakpaul, line 64: Should we load any pre-trained model here?

DorsaRoh: Thanks for the valuable input!
Diffusion policies are frequently tailored to specific use cases, and incorporating pretrained weights into the inference example could highly limit its general applicability and confuse users working on different tasks. Although I have pretrained weights available for a specific task that I can add here, to maintain the example’s universality, I recommend initializing the model without loading them. This will allow users to train their own models or integrate relevant pretrained weights based on their own applications!

sayakpaul: I beg to differ. I think if we can document it sufficiently it would make more sense to showcase this with a pre-trained model.

DorsaRoh: Sounds good! I have made the changes. Now, the example loads from a pretrained model

examples/reinforcement_learning/README.md
```diff
@@ -1,4 +1,13 @@
-# Overview
+
+## Diffusion-based Policy Learning for RL
+
+`diffusion_policy` implements <a href="https://diffusion-policy.cs.columbia.edu/">Diffusion Policy</a>, a diffusion model that predicts robot action sequences in reinforcement learning tasks.
```
sayakpaul, line 4:
```suggestion
`diffusion_policy` implements [Diffusion Policy](https://diffusion-policy.cs.columbia.edu/), a diffusion model that predicts robot action sequences in reinforcement learning tasks.
```

examples/reinforcement_learning/diffusion_policy.py
```diff
@@ -0,0 +1,172 @@
...
+        ).to(device)
+
+        # noise scheduler that controls the denoising process
+        self.noise_scheduler = DDPMScheduler(
+            num_train_timesteps=100,                # number of denoising steps 
+            beta_schedule="squaredcos_cap_v2"       # type of noise schedule
+        )   
+        
+        # load pre-trained weights from HuggingFace
+        checkpoint = torch.load(hf_hub_download("dorsar/diffusion_policy", "push_tblock.pt"), map_location=device)
```
sayakpaul, line 95:
```suggestion
        checkpoint = torch.load(hf_hub_download("dorsar/diffusion_policy", "push_tblock.pt"), weights_only=True, map_location=device)
```

examples/reinforcement_learning/diffusion_policy.py
```diff
@@ -0,0 +1,172 @@
...
+class DiffusionPolicy:
+    """
+    Implements diffusion policy for generating robot arm trajectories.
+    Uses diffusion to generate sequences of positions for a robot arm, conditioned on 
+    the current state of the robot and the block it needs to push.
+
+    The model expects observations in pixel coordinates (0-512 range) and block angle in radians.
+    It generates trajectories as sequences of (x,y) coordinates also in the 0-512 range.
+    """
+    def __init__(self, state_dim=5, device="cuda" if torch.cuda.is_available() else "cpu"):
```
sayakpaul, line 66: Better to not conditionally assign a value for an argument, IMO. 

```suggestion
    def __init__(self, state_dim=5, device="cpu"):
```

sayakpaul (commented):
Thanks! Left some further comments.

examples/reinforcement_learning/diffusion_policy.py
```diff
@@ -0,0 +1,201 @@
...
+import numpy.core.multiarray as multiarray
+import torch
+import torch.nn as nn
+from huggingface_hub import hf_hub_download
+from torch.serialization import add_safe_globals
+
+from diffusers import DDPMScheduler, UNet1DModel
+
+
+add_safe_globals(
```
sayakpaul, line 11: Why do we need it?

sayakpaul (approved):

examples/reinforcement_learning/diffusion_policy.py
DorsaRoh: After setting weights_only=True (from False), an error occurs for any pretrained model unless we use add_safe_globals to make custom or third-party methods available (since weights_only=True skips the configuration loading). I believe it is a preventative measure by HuggingFace for security reasons, because it explicitly states that if we use weights_only=False, we must trust the authors of the model

DorsaRoh:
It appears the 2 failing tests are unrelated to the changes in this PR and may be due to external factors or environment issues. Do they need to be addressed?

examples/reinforcement_learning/diffusion_policy.py
sayakpaul: It is happening at `torch.load()` so, I don't think it has anything to do with Hugging Face. Which torch version are you using?

DorsaRoh: I see, thank you - I am using 2.5.1

sayakpaul:
Indeed. That is not merge-blocking. Thanks for the PR!

PR #9840, opened 2024-11-02
[bitsandbytes] allow directly CUDA placements of pipelines loaded with bnb components

sayakpaul:
# What does this PR do?

When a pipeline is loaded with models that have quantization config, we should still be able to call `to("cuda")` on the pipeline object. For GPUs that would allow the memory (such as a 4090), this has performance benefits (as demonstrated below).

| Model CPU Offload | Batch Size | Time (seconds) | Memory (GB)         |
|:-----------------:|:----------:|:--------------:|:-------------------:|
|       False       |     1      |     19.316     |       14.935        |
|       True        |     1      |     36.746     |       12.139        |
|       False       |     4      |     80.665     |       20.576        |
|       True        |     4      |     98.612     |       12.138        |  

*Flux.1 Dev, steps: 30*

Currently, calling `to("cuda")` is not possible because:

```py
from transformers import BitsAndBytesConfig as BnbConfig
import torch 

ckpt_id = "black-forest-labs/FLUX.1-dev"

text_encoder_2_config = BnbConfig(
    load_in_4bit=True,
    bnb_4bit_quant_type="nf4",
    bnb_4bit_compute_dtype=torch.bfloat16,
)
    ckpt_id,
    subfolder="text_encoder_2",
    quantization_config=text_encoder_2_config,
    torch_dtype=torch.bfloat16
)
print(text_encoder_2._hf_hook)
```

has:

```bash
AlignDevicesHook(execution_device=0, offload=False, io_same_device=True, offload_buffers=False, place_submodules=True, skip_keys=None)
```

This is why this line complains:
https://github.com/huggingface/diffusers/blob/c10f875ff042f3fd2bc14ed019db68d4ad9567b6/src/diffusers/pipelines/pipeline_utils.py#L413

This PR fixes that behavior. 

Benchmarking code:

<details>
<summary>Unroll</summary>

```python
from diffusers import DiffusionPipeline, FluxTransformer2DModel, BitsAndBytesConfig
from transformers import BitsAndBytesConfig as BnbConfig
import torch 
import fire

    t0 = benchmark.Timer(
        stmt="f(*args, **kwargs)",
    )

def load_pipeline(model_cpu_offload=False):
    ckpt_id = "black-forest-labs/FLUX.1-dev"

    transformer_config = BitsAndBytesConfig(
        load_in_4bit=True,
        bnb_4bit_quant_type="nf4",
        bnb_4bit_compute_dtype=torch.bfloat16,
    )
        ckpt_id, 
        subfolder="transformer",
        quantization_config=transformer_config,
        torch_dtype=torch.bfloat16
    )

    text_encoder_2_config = BnbConfig(
        load_in_4bit=True,
        bnb_4bit_quant_type="nf4",
        bnb_4bit_compute_dtype=torch.bfloat16,
    )
        ckpt_id,
        subfolder="text_encoder_2",
        quantization_config=text_encoder_2_config,
        torch_dtype=torch.bfloat16
    )

        transformer=transformer,
        torch_dtype=torch.bfloat16,
    )
    if model_cpu_offload:
        pipeline.enable_model_cpu_offload()
    else:
        pipeline = pipeline.to("cuda")

    pipeline.set_progress_bar_config(disable=True)
    return pipeline

def run_pipeline(pipeline, batch_size=1):
    _ = pipeline(
        prompt="a dog sitting besides a sea", 
        guidance_scale=3.5, 
        max_sequence_length=512, 
        num_inference_steps=30,
        num_images_per_prompt=batch_size
    )


def main(batch_size: int = 1, model_cpu_offload: bool = False):
    pipeline = load_pipeline(model_cpu_offload=model_cpu_offload)

    for _ in range(5):
        run_pipeline(pipeline)

    time = benchmark_fn(run_pipeline, pipeline, batch_size)
    memory = torch.cuda.max_memory_allocated() / 1024 / 1024 / 1024
    print(f"{model_cpu_offload=}, {batch_size=} {time=} seconds {memory=} GB.")

    image = pipeline(
        prompt="a dog sitting besides a sea", 
        guidance_scale=3.5, 
        max_sequence_length=512, 
        num_inference_steps=30,
        num_images_per_prompt=1
    ).images[0]
    img_name = f"mco@{model_cpu_offload}-bs@{batch_size}.png"
    image.save(img_name)


if __name__ == "__main__":
    fire.Fire(main)
```


src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -410,10 +413,14 @@ def module_is_offloaded(module):
...
+            (_check_bnb_status(module)[1] or _check_bnb_status(module)[-1]) for _, module in self.components.items()
+        )
         if pipeline_is_sequentially_offloaded and device and torch.device(device).type == "cuda":
-            raise ValueError(
-                "It seems like you have activated sequential model offloading by calling `enable_sequential_cpu_offload`, but are now attempting to move the pipeline to GPU. This is not compatible with offloading. Please, move your pipeline `.to('cpu')` or consider removing the move altogether if you use sequential offloading."
-            )
+            if not pipeline_has_bnb:
+                raise ValueError(
+                    "It seems like you have activated sequential model offloading by calling `enable_sequential_cpu_offload`, but are now attempting to move the pipeline to GPU. This is not compatible with offloading. Please, move your pipeline `.to('cpu')` or consider removing the move altogether if you use sequential offloading."
+                )
```
SunMarc, line 423: I think that here will still be a potential issue even with that. The issue is that hooks are added and might cause issues if we really move the model. To fix this, we should probably remove them. They were added [here](https://github.com/huggingface/accelerate/blob/bf4572b6ce0a534a9d73537485a0edf1d68144b8/src/accelerate/big_modeling.py#L355) in accelerate. We should just relax the case with 4-bit bnb by checking the version of bnb + removing the `getattr(model, "is_quantized", False)` condition. LMK if this is clear !

SunMarc (commented):
Thanks for the PR ! Left a suggestion

src/diffusers/pipelines/pipeline_utils.py
sayakpaul: Do you maybe want to push your changes directly to this PR?

SunMarc: To fix this at the accelerate level !

sayakpaul: Alright. I will wait for it then

SunMarc: please try with that ! https://github.com/huggingface/accelerate/pull/3223

sayakpaul:
@SunMarc WDYT now?

SunMarc (approved):
Thanks for adding this ! LGTM ! I'll marge the PR on accelerate also

sayakpaul:
Have run the integration tests and they are passing.

SunMarc:
> Have run the integration tests and they are passing.
On diffusers ?

sayakpaul:
@SunMarc yes, on `diffusers`. Anywhere else they need to be run?

SunMarc:
No, I read that as a question, my bad ;)

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -448,8 +460,19 @@ def module_is_offloaded(module):
...
+                    if is_transformers_version(">", "4.44.0"):
+                        module.to(device=device)
+                    else:
+                        logger.warning(
+                            f"{module.__class__.__name__} could not be placed on {device}. Module is still on {module.device}. Please update your `transformers` installation to the latest."
+                        )
+                # For `diffusers` it should not be a problem as we enforce the installation of a bnb version
+                # that already supports CPU placements.
+                else:
+                    module.to(device=device)
```
yiyixuxu, line 475: ok but this means for diffusers the transformer version would always met the requirement, no? i.e. the check `is_transformers_version(">", "4.44.0")` will aways pass

sayakpaul: Agree but for the `diffusers` codepath, probably don't care about the `transformer` version, no? 

Anything I am missing?

sayakpaul: For a `transformers` model, we need to guard it with a version-check (since device placement was introduced in a later version in `transformers` for bitsandbytes quantized models). But for a `diffusers` model, we don't need that. Hence, the different codepaths and it also helps to capture and log the warnings better. 

If we do just this (just like it is currently in `main`):

```py
if is_loaded_in_4bit_bnb and device is not None and is_transformers_version(">", "4.44.0"):
     module.to(device=device)
```

then we won't be breaking anything but IMO, it's better to be explicit here because `module` can be either a `diffusers` model or a `transformers` model. So, placing a `diffusers` model on the basis of `is_transformers_version` feels a bit confusing to me. Or would you not prefer that?

yiyixuxu: Because we did not differentiate between diffusers vs transformer models at all, all models need to meet this requirement `is_transfomers_version_`

sayakpaul: Updated.

sayakpaul:
@yiyixuxu I have resolved your comment. LMK if it's okay now. I have run the tests, too and they are passing.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -421,6 +424,15 @@ def module_is_offloaded(module):
                 "It seems like you have activated a device mapping strategy on the pipeline which doesn't allow explicit device placement using `to()`. You can call `reset_device_map()` first and then call `to()`."
             )
 
+        pipeline_has_bnb = any(
+            (_check_bnb_status(module)[1] or _check_bnb_status(module)[-1]) for _, module in self.components.items()
+        )
```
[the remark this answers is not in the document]
DN6: Running the test gives:
```
E           ValueError: It seems like you have activated sequential model offloading by calling `enable_sequential_cpu_offload`, but are now attempting to move the pipeline to GPU. This is not compatible with offloading. Please, move your pipeline `.to('cpu')` or consider removing the move altogether if you use sequential offloading.

src/diffusers/pipelines/pipeline_utils.py:417: ValueError
```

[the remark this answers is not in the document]
sayakpaul: Which test are you running?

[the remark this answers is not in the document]
sayakpaul: > If this check is placed after the sequential offloading check, placement would still fail right?

I have modified the placement of the logic. Could you check again? 

Re. tests, I just ran `pytest tests/quantization/bnb/test_4bit.py::SlowBnb4BitTests` and `pytest tests/quantization/bnb/test_mixed_int8.py::SlowBnb8bitTests` and everything passed.

You need this PR https://github.com/huggingface/accelerate/pull/3223/ for this to work.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -389,6 +392,13 @@ def to(self, *args, **kwargs):
 
         device = device or device_arg
 
+        pipeline_has_bnb = any(any((_check_bnb_status(module))) for _, module in self.components.items())
```
yiyixuxu, line 395: it seems to have some overlapping logics with the code just a little bit below this, no?https://github.com/huggingface/diffusers/blob/6db33337a49c2f20db1b8d2ad069cca10c552c68/src/diffusers/pipelines/pipeline_utils.py#L444

sayakpaul: Good point. 

However, the LoC you pointed out is relevant when we're transferring an 8bit quantized model from one device to the other. It's a log to let the users know that this model has already been placed on a GPU and will remain so. Requesting to put it on a CPU will be ineffective. 

We call `self.to("cpu")` when doing `enable_model_cpu_offload()`:
https://github.com/huggingface/diffusers/blob/963ffca43419a8dffa682d9e03c2299b76feeced/src/diffusers/pipelines/pipeline_utils.py#L1039

So, this kind of log becomes informative in the context of using `enable_model_cpu_offload()`, for example. 

So, maybe this apparent overlap is justified. LMK.

yiyixuxu: >This PR, however, allows users to move an entire pipeline to a GPU when the memory permits. Previously it wasn't possible.

did I miss something? 
this PR add a check which throw a value error under certain condition - not enable a new use case like you described here, no?

sayakpaul: Well, the enablement comes from the `accelerate` fix https://github.com/huggingface/accelerate/pull/3223/ and this PR adds a check for that as you described. Sorry for the wrong order of words 😅

If you have other comments on the PR happy to address them.

yiyixuxu: my previous comments stands, it has overlapping logic with other checks you have below and is very very confusing. 

you're not enable a new use case here, this PR correct a previous wrong error message and allow user to take correct action, I would simply update the warning message here, to add the other possible scenario that they are trying to call `to("cuda")` on a quantized model without offloading, and they need to upgrade accelerate in order to do that
https://github.com/huggingface/diffusers/blob/8421c1461bf4ab7801070d04d6ec1e6b28ee5b59/src/diffusers/pipelines/pipeline_utils.py#L426

sayakpaul: >  this PR correct a previous wrong error message 

What was the wrong error message? 

IIUC the [line](https://github.com/huggingface/diffusers/blob/8421c1461bf4ab7801070d04d6ec1e6b28ee5b59/src/diffusers/pipelines/pipeline_utils.py#L426) you're point to has nothing to do with the changes introduced in this PR and has been in the codebase for quite a while.

The problem line (fixed by the `accelerate` PR) was this:
https://github.com/huggingface/diffusers/blob/c10f875ff042f3fd2bc14ed019db68d4ad9567b6/src/diffusers/pipelines/pipeline_utils.py#L413

So, what I have done in https://github.com/huggingface/diffusers/pull/9840/commits/1779093d7f4507b45123885ca343acbde969a07f is as follows: 

Updated the condition of the error message:

```bash
"You are trying to call `.to('cuda')` on a pipeline that has models quantized with `bitsandbytes`. Your current `accelerate` installation does not support it. Please upgrade the installation."
```

to:

```py
if (
      not pipeline_is_offloaded
      and not pipeline_is_sequentially_offloaded
      and pipeline_has_bnb
      and torch.device(device).type == "cuda"
      and is_accelerate_version("<", "1.1.0.dev0")
):
```

This now also considers when the pipeline is not offloaded. Additionally, 
https://github.com/huggingface/diffusers/blob/8421c1461bf4ab7801070d04d6ec1e6b28ee5b59/src/diffusers/pipelines/pipeline_utils.py#L446

now also consider if the pipeline is not offloaded:

https://github.com/huggingface/diffusers/blob/1779093d7f4507b45123885ca343acbde969a07f/src/diffusers/pipelines/pipeline_utils.py#L460

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -428,6 +431,19 @@ def module_is_offloaded(module):
...
+        # PR: https://github.com/huggingface/accelerate/pull/3223/
+        if (
+            not pipeline_is_offloaded
+            and not pipeline_is_sequentially_offloaded
+            and pipeline_has_bnb
+            and torch.device(device).type == "cuda"
+            and is_accelerate_version("<", "1.1.0.dev0")
+        ):
+            raise ValueError(
```
yiyixuxu, line 443: the error message you want to throw against this scenario, no?
1. accelerator < 1.1.0.dev0
2. `you call pipeline.to("cuda")` on a pipeline that has bnb 

but if these 2 condition are met (older accelerator version + bnb):
1. not pipeline_is_sequentially_offloaded` will be `False` here and you will not reach the value error 
2. you will reach this check first and get an error message -this is the wrong error message I was talking about https://github.com/huggingface/diffusers/blob/8421c1461bf4ab7801070d04d6ec1e6b28ee5b59/src/diffusers/pipelines/pipeline_utils.py#L413

```python
 if (
            not pipeline_is_offloaded
            and not pipeline_is_sequentially_offloaded
            and pipeline_has_bnb
            and torch.device(device).type == "cuda"
            and is_accelerate_version("<", "1.1.0.dev0")
        ):
```

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -410,10 +410,16 @@ def module_is_offloaded(module):
         pipeline_is_sequentially_offloaded = any(
             module_is_sequentially_offloaded(module) for _, module in self.components.items()
         )
-        if pipeline_is_sequentially_offloaded and device and torch.device(device).type == "cuda":
-            raise ValueError(
-                "It seems like you have activated sequential model offloading by calling `enable_sequential_cpu_offload`, but are now attempting to move the pipeline to GPU. This is not compatible with offloading. Please, move your pipeline `.to('cpu')` or consider removing the move altogether if you use sequential offloading."
-            )
+        if device and torch.device(device).type == "cuda":
+            if pipeline_is_sequentially_offloaded and not pipeline_has_bnb:
```
yiyixuxu, line 414: my previous comments here apply almost exactly here so I will just repeat it
https://github.com/huggingface/diffusers/pull/9840

the error message you want to throw against this scenario:

* accelerator < 1.1.0.dev0
* you call `pipeline.to("cuda")` on a pipeline that has bnb

if these 2 condition are met (older accelerator version + bnb), it will not reach the error message you intended, it will be caught here at this firs check, and the error message is same as before this PR (about offloading)

can you do this? https://github.com/huggingface/diffusers/pull/9840#discussion_r1868969624

IF not, please remove the changes to `pipline_utils.py` and we can merge (I will work on it in a separate PR) I think the added tests are fine without the changes: if accecelrate version is new, it is not affected by the changes in this PR; if it is not, it throw a different error, that's all

yiyixuxu: ok I was wrong! will merge

sayakpaul: Sure that works but here's my last try. 

> if these 2 condition are met (older accelerator version + bnb), it will not reach the error message you intended, it will be caught here at this firs check, and the error message is same as before this PR (about offloading)

When you have:

```py
model_id = "hf-internal-testing/flux.1-dev-nf4-pkg"
t5_4bit = T5EncoderModel.from_pretrained(model_id, subfolder="text_encoder_2")
transformer_4bit = FluxTransformer2DModel.from_pretrained(model_id, subfolder="transformer")
pipeline_4bit = DiffusionPipeline.from_pretrained(
    text_encoder_2=t5_4bit,
    transformer=transformer_4bit,
    torch_dtype=torch.float16,
)
```

in `if pipeline_is_sequentially_offloaded and not pipeline_has_bnb`, `pipeline_is_sequentially_offloaded` will be True (older `accelerate` version), however, `not pipeline_has_bnb` will be False (as expected). 

To test, you can run the following with `accelerate` 1.0.1:

```py
from diffusers import DiffusionPipeline, FluxTransformer2DModel
import torch 

model_id = "hf-internal-testing/flux.1-dev-nf4-pkg"
t5_4bit = T5EncoderModel.from_pretrained(model_id, subfolder="text_encoder_2")
transformer_4bit = FluxTransformer2DModel.from_pretrained(model_id, subfolder="transformer")
pipeline_4bit = DiffusionPipeline.from_pretrained(
    text_encoder_2=t5_4bit,
    transformer=transformer_4bit,
    torch_dtype=torch.float16,
).to("cuda")
```

It throws:

```bash
ValueError: You are trying to call `.to('cuda')` on a pipeline that has models quantized with `bitsandbytes`. Your current `accelerate` installation does not support it. Please upgrade the installation.
```

Isn't this what we expect or am I missing something?

yiyixuxu: yeah I missed that `not pipeline_has_bnb` in the statement, it works

sayakpaul: Saw your comment. Thanks for beating it with me :)

PR #9845, opened 2024-11-02
[LoRA] allow big CUDA tests to run properly for LoRA (and others)

sayakpaul:
# What does this PR do?

Some comments in-line.


tests/pipelines/mochi/test_mochi.py
```diff
@@ -261,7 +264,10 @@ def test_vae_tiling(self, expected_diff_max: float = 0.2):
 
 
 @slow
+@nightly
```
a-r-r-o-w, line 267: @sayakpaul Is this right to add here?

sayakpaul: Yes, but also test this by SSHing into the runner and running the test actually: https://github.com/huggingface/diffusers/blob/99c0483b67427de467f11aa35d54678fd36a7ea2/.github/workflows/nightly_tests.yml#L189

tests/lora/test_lora_layers_sd3.py
```diff
@@ -130,9 +134,13 @@ def test_modify_padding_mode(self):
         pass
 
 
+@slow
```
DN6, line 137: Think this can just run on nightly

tests/pipelines/controlnet_flux/test_controlnet_flux.py
```diff
@@ -205,6 +206,7 @@ def test_flux_image_output_shape(self):
 
 
 @slow
```
DN6, line 208: Can also just run on nightly

tests/pipelines/mochi/test_mochi.py
DN6: Can remove the slow decorator

sayakpaul:
@DN6 comments addressed. @a-r-r-o-w could you also check if we should consider other pipeline tests in this PR (Allegro, etc.)? Okay if we do that in a separate PR. LMK.

sayakpaul:
Gently pinging @DN6 and @a-r-r-o-w. I think these tests are important to run nightly.

tests/pipelines/mochi/test_mochi.py
```diff
@@ -293,7 +297,7 @@ def test_mochi(self):
         ).frames
 
         video = videos[0]
-        expected_video = torch.randn(1, 16, 480, 848, 3).numpy()
+        expected_video = torch.randn(1, 19, 480, 848, 3).numpy()
```
a-r-r-o-w, line 300: Thanks for catching!

a-r-r-o-w (approved):
Thanks, LGTM! Off to @DN6

sayakpaul:
@DN6 a gentle ping here.

hlky (approved):

PR #9862, opened 2024-11-04
[LoRA] feat: `save_lora_adapter()`

sayakpaul:
# What does this PR do?

Complementing https://github.com/huggingface/diffusers/pull/9712, this PR adds a `save_lora_adapter()` for the models that support LoRA loading. It also adds tests to ensure things don't break. 

Additionally, this PR:

* Deprecates the `load_attn_procs()` method when it tries to load a LoRA state dict (and adds tests for it). 
* Replaces the `load_attn_procs()` method with `load_lora_adapter()` in `src/diffusers/loaders/lora_pipeline.py`. I have run the integration tests for the SD and SDXL LoRAs (as those are impacted by this change), and the tests passed. 
* Some minor code improvements on the loading side (comments in line). 

Additionally, just to be sure, I have run the integration tests under `tests/lora` to ensure they pass.


check_flux.py
```diff
@@ -0,0 +1,100 @@
+from diffusers import AutoencoderKL, FluxTransformer2DModel, FluxPipeline, FlowMatchEulerDiscreteScheduler
```
sayakpaul, line 1: Ignore.

tests/models/test_modeling_common.py
```diff
@@ -877,8 +892,6 @@ def _set_gradient_checkpointing_new(self, module, value=False):
         model = model_class_copy(**init_dict)
         model.enable_gradient_checkpointing()
 
-        print(f"{set(modules_with_gc_enabled.keys())=}, {expected_set=}")
```
sayakpaul, line 880: Unrelated but my hands were itching.

tests/models/test_modeling_common.py
```diff
@@ -902,6 +915,97 @@ def test_deprecated_kwargs(self):
...
+            state_dict_loaded = safetensors.torch.load_file(os.path.join(tmpdir, "pytorch_lora_weights.safetensors"))
+
+            model.unload_lora()
+            self.assertFalse(check_if_lora_correctly_set(model), "LoRA layers not set correctly")
+
+            model.load_lora_adapter(tmpdir, use_safetensors=True)
+            state_dict_retrieved = get_peft_model_state_dict(model, adapter_name="default_0")
+
+            print(f"{state_dict_loaded.keys()=}, {len(state_dict_loaded)}")
+            print(f"{state_dict_retrieved.keys()=} {len(state_dict_retrieved)}")
```
sayakpaul, line 965: @BenjaminBossan do you have an idea as why these two state dicts have different lengths? IMO, they should have the exact same length? Or am I being mistaken here?

The keys that are missing in the retrieved state dict have the `"single_transformer_blocks"` prefix. 

I am on the `peft` main (`0.13.3.dev0`).

BenjaminBossan: I debugged and the cause is this line:

https://github.com/huggingface/diffusers/pull/9862/files#diff-fdc2e14a8c1091d917cf31ca155be9baa8095d704bd2a2f58970ec31641c0caeR198

Since `prefix` is defined as `"transformer"` but the keys are starting with `"single_transformer_blocks"`, they don't match. In fact, even the other keys should not match: Even though they start with `"transformer"`, they don't start with `"transformer."` (note the dot), which _I think_ is the intended condition, since the next line replaces `f"{prefix}."`.

I think what happens here is that since this is a standalone module, not a transformer submodule of a larger model, it does not have a `"transformer."`  prefix. Therefore, the keys should not be filtered like that. Probably you know best what the correct solution would be here, but one suggestion is to make the `prefix` argument optional and only filter if it is not `None`.

Also, this problem probably rarely arise in real use cases, as users will typically not load the transformer as a standalone module.

sayakpaul: Thank you very much!

src/diffusers/loaders/peft.py
```diff
@@ -189,40 +193,47 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
+        if network_alphas is not None and prefix is None:
+            raise ValueError("`network_alphas` cannot be None when `prefix` is None.")
 
-        keys = list(state_dict.keys())
-        transformer_keys = [k for k in keys if k.startswith(prefix)]
-        if len(transformer_keys) > 0:
-            state_dict = {k.replace(f"{prefix}.", ""): v for k, v in state_dict.items() if k in transformer_keys}
+        if prefix is not None:
+            keys = list(state_dict.keys())
+            model_keys = [k for k in keys if k.startswith(f"{prefix}.")]
```
BenjaminBossan, line 201: Could be a set comprehension for a little bit extra performance, but probably not much of a difference overall.

src/diffusers/loaders/peft.py
```diff
@@ -189,40 +193,47 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
-
             rank = {}
             for key, val in state_dict.items():
                 if "lora_B" in key:
                     rank[key] = val.shape[1]
 
             if network_alphas is not None and len(network_alphas) >= 1:
-                alpha_keys = [k for k in network_alphas.keys() if k.startswith(prefix) and k.split(".")[0] == prefix]
+                alpha_keys = [
+                    k for k in network_alphas.keys() if k.startswith(f"{prefix}.") and k.split(".")[0] == prefix
```
BenjaminBossan, line 223: Hmm, AFAICT the two conditions `k.startswith(f"{prefix}.") and k.split(".")[0] == prefix` are redundant, right? In the previous code, we had `k.startswith(prefix)`, which is a bit different, but was also unnecessary.

src/diffusers/loaders/peft.py
```diff
@@ -189,40 +193,47 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
                 network_alphas = {k.replace(f"{prefix}.", ""): v for k, v in network_alphas.items() if k in alpha_keys}
 
             lora_config_kwargs = get_peft_kwargs(rank, network_alpha_dict=network_alphas, peft_state_dict=state_dict)
             if "use_dora" in lora_config_kwargs:
-                if lora_config_kwargs["use_dora"] and is_peft_version("<", "0.9.0"):
-                    raise ValueError(
-                        "You need `peft` 0.9.0 at least to use DoRA-enabled LoRAs. Please upgrade your installation of `peft`."
-                    )
+                if lora_config_kwargs["use_dora"]:
+                    if is_peft_version("<", "0.9.0"):
```
BenjaminBossan, line 230: This is the same code as before, just split into 2 lines, right?

src/diffusers/loaders/peft.py
```diff
@@ -276,6 +287,52 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
                 _pipeline.enable_sequential_cpu_offload()
             # Unsafe code />
 
+    def save_lora_adapter(
+        self,
+        save_directory,
+        adapter_name: str = "default",
+        upcast_before_saving: bool = False,
+        safe_serialization: bool = True,
+        weight_name: str = None,
```
BenjaminBossan, line 296:
```suggestion
        weight_name: Optional[str] = None,
```

src/diffusers/loaders/peft.py
```diff
@@ -276,6 +287,52 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
+        from .lora_base import LORA_WEIGHT_NAME, LORA_WEIGHT_NAME_SAFE
+
+        if adapter_name is None:
+            adapter_name = get_adapter_name(self)
+
+        if adapter_name not in getattr(self, "peft_config", {}):
+            raise ValueError(f"Adapter name {adapter_name} not found in the model.")
+
+        lora_layers_to_save = get_peft_model_state_dict(
+            self.to(dtype=torch.float32 if upcast_before_saving else None), adapter_name=adapter_name
```
BenjaminBossan, line 310: TIL about `.to(None)`.

src/diffusers/loaders/peft.py
```diff
@@ -276,6 +287,52 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
+        )
+        if os.path.isfile(save_directory):
+            logger.error(f"Provided path ({save_directory}) should be a directory, not a file")
+            return
```
BenjaminBossan, line 314: Does it make sense to return here instead of raising an error?

src/diffusers/loaders/peft.py
```diff
@@ -276,6 +287,52 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
+        os.makedirs(save_directory, exist_ok=True)
+
+        if weight_name is None:
+            if safe_serialization:
+                weight_name = LORA_WEIGHT_NAME_SAFE
+            else:
+                weight_name = LORA_WEIGHT_NAME
+
+        save_path = Path(save_directory, weight_name).as_posix()
+        save_function(lora_layers_to_save, save_path)
```
BenjaminBossan, line 333: I wonder if it would make sense to (optionally) save the peft config too or not.

tests/lora/utils.py
```diff
@@ -1784,18 +1784,15 @@ def test_missing_keys_warning(self):
...
-        )
+        logger = logging.get_logger("diffusers.loaders.peft")
         logger.setLevel(30)
         with CaptureLogger(logger) as cap_logger:
             pipe.load_lora_weights(state_dict)
 
         # Since the missing key won't contain the adapter name ("default_0").
         # Also strip out the component prefix (such as "unet." from `missing_key`).
         component = list({k.split(".")[0] for k in state_dict})[0]
+        print(f"{cap_logger.out=}")
```
BenjaminBossan, line 1795: Remove?

BenjaminBossan (commented):
Thanks for adding a method to save the LoRA adapter. Overall, this looks good, I have a few comments but no blockers.

src/diffusers/loaders/peft.py
sayakpaul: We could but we don't really support the loading. So, maybe something to consider for the future? I have made a note about it.

src/diffusers/loaders/peft.py
sayakpaul: Changed to a `ValueError`

src/diffusers/loaders/peft.py
sayakpaul: Yeah.

src/diffusers/loaders/peft.py
sayakpaul: Could just keep it to "k.startswith(f"{prefix}.")"?

src/diffusers/loaders/peft.py
sayakpaul: Yeah.

src/diffusers/loaders/peft.py
```diff
@@ -189,40 +193,45 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
+        if network_alphas is not None and prefix is None:
+            raise ValueError("`network_alphas` cannot be None when `prefix` is None.")
 
-        keys = list(state_dict.keys())
-        transformer_keys = [k for k in keys if k.startswith(prefix)]
-        if len(transformer_keys) > 0:
-            state_dict = {k.replace(f"{prefix}.", ""): v for k, v in state_dict.items() if k in transformer_keys}
+        if prefix is not None:
+            keys = list(state_dict.keys())
+            model_keys = [k for k in keys if k.startswith(f"{prefix}.")]
```
sayakpaul, line 201: Just a better and more robust way to filter out the state dict.

src/diffusers/loaders/peft.py
```diff
@@ -189,40 +193,45 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
+            if len(model_keys) > 0:
+                state_dict = {k.replace(f"{prefix}.", ""): v for k, v in state_dict.items() if k in model_keys}
+
+        if len(state_dict) > 0:
+            if adapter_name in getattr(self, "peft_config", {}):
+                raise ValueError(
+                    f"Adapter name {adapter_name} already in use in the model - please select a new adapter name."
+                )
```
sayakpaul, line 209: Catching this error earlier than previous.

src/diffusers/loaders/peft.py
```diff
@@ -189,40 +193,45 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
-                )
-
             rank = {}
             for key, val in state_dict.items():
                 if "lora_B" in key:
                     rank[key] = val.shape[1]
 
             if network_alphas is not None and len(network_alphas) >= 1:
-                alpha_keys = [k for k in network_alphas.keys() if k.startswith(prefix) and k.split(".")[0] == prefix]
+                alpha_keys = [k for k in network_alphas.keys() if k.startswith(f"{prefix}.")]
```
sayakpaul, line 222: Removing redundant conditions.

src/diffusers/loaders/peft.py
```diff
@@ -189,40 +193,45 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
-                    )
+                if lora_config_kwargs["use_dora"]:
+                    if is_peft_version("<", "0.9.0"):
+                        raise ValueError(
+                            "You need `peft` 0.9.0 at least to use DoRA-enabled LoRAs. Please upgrade your installation of `peft`."
+                        )
                 else:
-                    lora_config_kwargs.pop("use_dora")
+                    if is_peft_version("<", "0.9.0"):
+                        lora_config_kwargs.pop("use_dora")
```
sayakpaul, line 234: Breaking the conditionals to be more explicit.

tests/models/unets/test_models_unet_2d_condition.py
```diff
@@ -1078,30 +1078,7 @@ def test_load_sharded_checkpoint_device_map_from_hub_local_subfolder(self):
         assert new_output.sample.shape == (4, 4, 16, 16)
 
     @require_peft_backend
-    def test_lora(self):
```
sayakpaul, line 1081: Unneeded to test here now.

BenjaminBossan (approved):
Thanks for this PR, LGTM. I assume the docstring will be completed before merging.

DN6 (approved):
LGTM 👍🏽

sayakpaul:
@yiyixuxu could you give this a review?

bghira:
why isn't this merged yet?

sayakpaul:
Pending reviews from Yiyi. It will take a bit as she is off for some days.

yiyixuxu:
thanks!

PR #9864, opened 2024-11-04
Update ptxla training

entrpn:
- Updates TPU benchmark numbers.
- Updates the ptxla training example code.
- Adds flash attention to ptxla code running on TPUs.

@sayakpaul can you please review. This new PR supersedes the [other one](https://github.com/huggingface/diffusers/pull/9446) I had opened a while back, which I just closed. Thank you.

Fixes # (issue)

## Before submitting

## Who can review?


sayakpaul:
Cc: @yiyixuxu could you review the changes made to `attention_processor.py`?

yiyixuxu:
@entrpn can you use a custom attention instead? (without updating our default attention processor)

zpcore:
> @entrpn can you use a custom attention instead? (without updating our default attention processor)

Hi @yiyixuxu , we wrapped the flash attention kernel call under condition `if XLA_AVAILABLE`. This shouldn't touch the default attention processor behavior. Can you give more details about `use a custom attention`? Thanks

yiyixuxu:
I'm just wondering if it makes sense for Flash Attention to have its attention processor since this one is meant for SDPA

cc @DN6 here too

entrpn:
@yiyixuxu this makes sense.

@zpcore do you think you can implement it?

zpcore:
> @yiyixuxu this makes sense.
> 
> @zpcore do you think you can implement it?

Yes, I can follow up with the code change.

zpcore:
Hi @yiyixuxu , what about we create another AttnProcess with flash attention in parallel with `AttnProcessor2_0`? My concern is that majority of the code will be the same as `AttnProcessor2_0`.

yiyixuxu:
@zpcore 
that should not be a problem. a lot of our attention processors share majority of same code, e.g. https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention_processor.py#L732 and https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention_processor.py#L2443 

this way user can explicitly set to use flash attention if they want to

miladm:
@yiyixuxu - to better understand, can you please help me understand why wrapping the flash attention kernel call under condition `if XLA_AVAILABLE` causes a trouble? Do you want this functionality to be more generalized?

yiyixuxu:
is it not possible that XLA_AVAILABLE but the user does not want to use flash attention? 
our attention processors are designed to be very easy to switch & each one corresponding to a very specific method -> could be exformer, SDPA, or even like special method like fused has its own processor

sayakpaul:
@miladm @zpcore a gentle ping

zpcore:
Thanks for the review feedback. We split out the XLA flash attention process from AttnProcessor2_0 as requested in the review. PTAL

src/diffusers/models/attention_processor.py
```diff
@@ -2750,6 +2763,117 @@ def __call__(
...
+        hidden_states: torch.Tensor,
+        encoder_hidden_states: Optional[torch.Tensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
+        temb: Optional[torch.Tensor] = None,
+        *args,
+        **kwargs,
+    ) -> torch.Tensor:
+        if len(args) > 0 or kwargs.get("scale", None) is not None:
+            deprecation_message = "The `scale` argument is deprecated and will be ignored. Please remove it, as passing it will raise an error in the future. `scale` should directly be passed while calling the underlying pipeline component i.e., via `cross_attention_kwargs`."
+            deprecate("scale", "1.0.0", deprecation_message)
```
sayakpaul, line 2789: Since this is a new attention processor, I think we can safely remove this.

src/diffusers/models/attention_processor.py
```diff
@@ -2750,6 +2763,117 @@ def __call__(
         return hidden_states
 
 
+class XLAFlashAttnProcessor2_0:
```
sayakpaul, line 2766: So, this will be _automatically_ used when using the compatible models under an XLA environment, right?

src/diffusers/models/attention_processor.py
```diff
@@ -36,6 +36,13 @@
 else:
     xformers = None
 
+if is_torch_xla_available():
+    from torch_xla.experimental.custom_kernel import flash_attention
```
sayakpaul, line 40: Does this need to go through any version check guards too i.e., a minimum version known to have `flash_attention`?

src/diffusers/models/attention_processor.py
```diff
@@ -269,10 +276,16 @@ def __init__(
         # We use the AttnProcessor2_0 by default when torch 2.x is used which uses
         # torch.nn.functional.scaled_dot_product_attention for native Flash/memory_efficient_attention
         # but only if it has the default `scale` argument. TODO remove scale_qk check when we move to torch 2.1
+        # If torch_xla is available, we use pallas flash attention kernel to improve the performance.
         if processor is None:
-            processor = (
-                AttnProcessor2_0() if hasattr(F, "scaled_dot_product_attention") and self.scale_qk else AttnProcessor()
-            )
+            if hasattr(F, "scaled_dot_product_attention") and self.scale_qk:
+                if is_torch_xla_available:
```
sayakpaul, line 282: Same here too. Does this need to be guarded with a version check too?

sayakpaul (commented):
Thanks for working on this and for being patient with our feedback. 

I left a few minor comments. 

The other reviewer, @yiyixuxu will review this soon. Please allow for some time because of the thanksgiving week.

src/diffusers/models/attention_processor.py
zpcore: Yes, AttnProcessor2_0 will be replaced with XLAFlashAttnProcessor2_0 if XLA version condition satisfied.

src/diffusers/models/attention_processor.py
zpcore: Removed.

src/diffusers/models/attention_processor.py
zpcore: Introduced the version check function `is_torch_xla_version` in import_utils.py. Added the version check for torch_xla here.

src/diffusers/models/attention_processor.py
zpcore: Added the version check for torch_xla here too.

zpcore:
> I think @yiyixuxu's point here is valid:
> 
> > is it not possible that XLA_AVAILABLE but the user does not want to use flash attention?
> 
> IMO it's better to use a similar API to xformers to enable the XLA processor.
> 
> https://github.com/huggingface/diffusers/blob/784b351f32fad39ad8c8bb238faaa44090e00a08/src/diffusers/models/modeling_utils.py#L228

OK, now I get it! We have added functions like `enable_use_xla_flash_attention` similar to the `enable_xformers_memory_efficient_attention` to give user the option to enable xla flash attention or not. In the example ()[train_text_to_image_xla.py](https://github.com/huggingface/diffusers/pull/9864/files#diff-1bfda209ceb7442ed78c28c9931d5842cb3bc5b47349b542e5ab53eb0c2dc838) we give, we apply the kernel to the diffusion model unet. Thanks!

src/diffusers/models/attention_processor.py
```diff
@@ -2751,6 +2783,117 @@ def __call__(
...
+                    .masked_fill(attention_mask == 1, float(0.0))
+                )
+
+                # Apply attention mask to key
+                key = key + attention_mask
+            query /= math.sqrt(query.shape[3])
+            partition_spec = self.partition_spec if is_spmd() else None
+            hidden_states = flash_attention(query, key, value, causal=False, partition_spec=partition_spec)
+        else:
+            hidden_states = F.scaled_dot_product_attention(
```
yiyixuxu, line 2874: we don't need to support SDPA in this XLAFlash attention processor! - we can remove all the logics related to it!

src/diffusers/models/attention_processor.py
```diff
@@ -2751,6 +2783,117 @@ def __call__(
...
 
 
+class XLAFlashAttnProcessor2_0:
+    r"""
+    Processor for implementing scaled dot-product attention with pallas flash attention kernel if using `torch_xla`.
+    """
+
+    def __init__(self, partition_spec: Optional[Tuple[Optional[str], ...]] = None):
+        if not hasattr(F, "scaled_dot_product_attention"):
+            raise ImportError("XLAFlashAttnProcessor2_0 requires PyTorch 2.0, to use it, please upgrade PyTorch to 2.0.")
```
yiyixuxu, line 2793:
```suggestion
```
I think we don't need to support SDPA in XLA Flash attention processor! let's remove all the logics related to that to simplify things a bit!

src/diffusers/utils/import_utils.py
```diff
@@ -700,6 +700,19 @@ def is_torch_version(operation: str, version: str):
     return compare_versions(parse(_torch_version), operation, version)
 
 
+def is_torch_xla_version(operation: str, version: str):
```
yiyixuxu, line 703: can we make sure we can call `is_torch_xla_version()` when it is not installed? currently, I think you will have to run it together with `is_torch_xla_available()`, because the `_torch_xla_version` is not defined otherwise

we can do like this https://github.com/huggingface/diffusers/blob/243d9a49864ebb4562de6304a5fb9b9ebb496c6e/src/diffusers/utils/import_utils.py#L55

yiyixuxu (commented):
thanks! l think we can merge this soon!

src/diffusers/models/attention_processor.py
zpcore: There is a constraint when using the pallas kernel. We need this `all(tensor.shape[2] >= 4096 for tensor in [query, key, value])` or xla will error out.

However, we added a new error message when it fall back to `scaled_dot_product_attention` to avoid silently skip the kernel.

src/diffusers/models/attention_processor.py
zpcore: Please check my comment above for why we still keep it here. Thanks

src/diffusers/utils/import_utils.py
zpcore: Nice catch, updated.

src/diffusers/models/attention_processor.py
```diff
@@ -275,6 +284,29 @@ def __init__(
...
+            partition_spec (`Tuple[]`, *optional*):
+                Specify the partition specification if using SPMD. Otherwise None.
+        """
+        if (
+            use_xla_flash_attention
+            and is_torch_xla_available
+            and is_torch_xla_version('>', '2.2')
+            and (not is_spmd() or is_torch_xla_version('>', '2.3'))
+        ):
+            processor = XLAFlashAttnProcessor2_0(partition_spec)
```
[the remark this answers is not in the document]
zpcore: Sounds good, updated!

src/diffusers/models/attention_processor.py
yiyixuxu: ok thank you for explaining to me!

yiyixuxu (approved):
thank you!

entrpn:
Thank you all!

PR #9877, opened 2024-11-06
CogVideoX 1.5

zRzRzRzRzRzRzR:
# What does this PR do?

This PR is a draft about the new generation of CogVideoX, which has not yet been fully realized
Main work:
1. The new PatchEmbedding has increased t_patch.
2. The resolution needs to consider compatibility with previous versions; the new version has a higher resolution, but it is also a fixed value. The Rotray Embedding is estimated to need changes.
3. Image-to-video has more resolution features.
The SAT version is very slow, and the memory usage is quite high; this may need optimization in future PRs.

Currently achieved
1. The new PatchEmbedding has increased t_patch.

# Who Can help this
@a-r-r-o-w


scripts/convert_cogvideox_to_diffusers.py
```diff
@@ -187,6 +193,34 @@ def convert_vae(ckpt_path: str, scaling_factor: float, dtype: torch.dtype):
...
+
+    elif version == "1.5":
+        vae_scale_factor_spatial = 8
+        init_kwargs = {
+            "patch_size": 2,
+            "patch_size_t": 2,
+            "patch_bias": False,
+            "sample_height": 768 // vae_scale_factor_spatial,
+            "sample_width": 1360 // vae_scale_factor_spatial,
+            "sample_frames": 81, # TODO: Need Test with 161 for 10 seconds
```
a-r-r-o-w, line 216:
```suggestion
            "sample_frames": 81,
```

This is just to determine the default number of frames for sampling, so we do not need to make a modification here (which would affect the `config.json` of the converted transformer model). Users can still specify 161 frames (in the call to pipeline) for generation normally and we will still be compatible without needing any modifications here.

scripts/convert_cogvideox_to_diffusers.py
```diff
@@ -230,21 +281,27 @@ def get_args():
...
             args.num_attention_heads,
             args.use_rotary_positional_embeddings,
             args.i2v,
             dtype,
+            init_kwargs,
         )
     if args.vae_ckpt_path is not None:
-        vae = convert_vae(args.vae_ckpt_path, args.scaling_factor, dtype)
+        # Keep VAE in float32 for better quality
+        vae = convert_vae(args.vae_ckpt_path, args.scaling_factor, torch.float32)
```
a-r-r-o-w, line 296: @zRzRzRzRzRzRzR This is a bit of a breaking change. The SAT VAE is in fp32 but the diffusers format VAE is in bf16/fp16. This can lead to poorer quality, so it is best to just keep the VAE in fp32 and let users decide what configuration to use. I will open a PR to the other model weight CogVideoX repositories with the updated VAE weights soon.

cc @yiyixuxu @DN6 The VAE quality doesn't take too much of a hit, but best to have the default in FP32 and update all existing checkpoints. Apologies that this slipped through earlier but I definitely notice very minor differences in quality (atleast in training cc @sayakpaul). The transformer modeling weights don't use variants because there is no FP32 weights as training is done in BF16

scripts/convert_cogvideox_to_diffusers.py
```diff
@@ -276,11 +333,6 @@ def get_args():
         scheduler=scheduler,
     )
 
-    if args.fp16:
```
a-r-r-o-w, line 279: Due to the explanation above, we shouldn't typecast all weights in the pipeline. VAE is best in FP32, text encoder could be saved in FP32 but works well at lower precisions as well, and transformer is either in BF16, or FP16 for CogVideoX-2B text-to-video

src/diffusers/pipelines/cogvideo/pipeline_cogvideox.py
```diff
@@ -481,9 +486,9 @@ def __call__(
         self,
         prompt: Optional[Union[str, List[str]]] = None,
         negative_prompt: Optional[Union[str, List[str]]] = None,
-        height: int = 480,
-        width: int = 720,
-        num_frames: int = 49,
+        height: Optional[int] = None,
```
a-r-r-o-w, line 489: This is not a breaking change since we can determine these exact values using transformer config parameters

src/diffusers/pipelines/cogvideo/pipeline_cogvideox_fun_control.py
```diff
@@ -488,8 +488,13 @@ def _prepare_rotary_positional_embeddings(
...
         grid_width = width // (self.vae_scale_factor_spatial * self.transformer.config.patch_size)
-        base_size_width = 720 // (self.vae_scale_factor_spatial * self.transformer.config.patch_size)
-        base_size_height = 480 // (self.vae_scale_factor_spatial * self.transformer.config.patch_size)
+
+        p = self.transformer.config.patch_size
+        p_t = self.transformer.config.patch_size_t or 1
+
+        base_size_width = self.transformer.config.sample_width // p
+        base_size_height = self.transformer.config.sample_height // p
+        base_num_frames = (num_frames + p_t - 1) // p_t
```
a-r-r-o-w, line 497: This is not a breaking change either because they are mathematically equivalent for 1.0 models which do not use temporal patch embedding, but the ceil div is required for new 1.5 CogVideoX models.

src/diffusers/pipelines/cogvideo/pipeline_cogvideox_image2video.py
```diff
@@ -666,11 +671,6 @@ def __call__(
             `tuple`. When returning a tuple, the first element is a list with the generated images.
         """
 
-        if num_frames > 49:
```
a-r-r-o-w, line 669: We need to remove the frame restriction since the newer models can generate at higher frame resolutions. 49 frames will still be the default for 1.0 models because we determine that by the transformer config params

scripts/convert_cogvideox_to_diffusers.py
zRzRzRzRzRzRzR: Understood, this is the right thing to do.

src/diffusers/pipelines/cogvideo/pipeline_cogvideox_image2video.py
```diff
@@ -726,6 +725,22 @@ def __call__(
...
+                image_width = tgt_width
+
+            # Step 2: Ensure height and width are divisible by the divisor
+            image_height = (image_height // divisor) * divisor
+            image_width = (image_width // divisor) * divisor
+            return image_height, image_width
+
+        image_width, image_height = image.size[-2:]
+
+        height, width = adjust_resolution_to_divisible(image_height, image_width, height, width)
```
a-r-r-o-w, line 743:
```suggestion
```

I think this will not be required. We can mention it in the docs as well as add a comment saying that the image_height and image_width need to be divisible by 16. This check can also be added to the `check_inputs` method. The user needs to make sure to correctly pass the `height` and `width` parameters based on the input image dimension

kijai:
Hey,

I've been testing this draft within my ComfyUI wrapper nodes, and I'd like clarity on couple of things as I'm unsure how this should perform at this stage, mostly to know if I have made a mistake somewhere:

- With text to video model the input hidden_states are always padded, which seems to cause noise/corruption for the first latent/first 4 frames for the resulting video. Is this unavoidable?

- With both models many resolutions don't really work at all, especially smaller ones, for example the old default 720x480 is blurry mess with T2V and just colorful blocks with I2V. New default resolution is fine, also gotten really good results with square ones (640p, 768,p 1024p). I've made sure it's always divisibly by 16.

a-r-r-o-w:
Hi @kijai. Could you try experimenting with this branch: https://github.com/zRzRzRzRzRzRzR/diffusers/pull/1? 

It removes the padding of hidden_states which seems to cause corruption. Instead, you must specify frame values whose latent size is divisible by patch_size_t (such as 85 or 165). I believe there might still be some modifications remaining on how to handle frames based on my conversation with Yuxuan.

Regarding resolutions, the T2V model works best only at specific resolutions (the recommended is to always do 1360x768). The I2V model can generate at multiple resolutions I believe. Also, both models can generate best at 85 and 165 frames (as per my PR above, but this should actually be 81 and 161 to be consistent with original implementation. We'll try to figure out the best way to support this today)

kijai:
> Hi @kijai. Could you try experimenting with this branch: [zRzRzRzRzRzRzR#1](https://github.com/zRzRzRzRzRzRzR/diffusers/pull/1)?
> 
> It removes the padding of hidden_states which seems to cause corruption. Instead, you must specify frame values whose latent size is divisible by patch_size_t (such as 85 or 165). I believe there might still be some modifications remaining on how to handle frames based on my conversation with Yuxuan.
> 
> Regarding resolutions, the T2V model works best only at specific resolutions (the recommended is to always do 1360x768). The I2V model can generate at multiple resolutions. Also, both models can generate best at 85 and 165 frames (as per my PR above, but this should actually be 81 and 161 to be consistent with original implementation. We'll try to figure out the best way to support this today)

I did try without the padding already, but then you get the first 4 frames without movement, I'll check if that branch has something for that thanks.

To illustrate what I mean with the I2V and 720x480 (for example):

https://github.com/user-attachments/assets/010f9b16-9f97-4421-b256-b9e1f68ea22b

Also forgive me for not knowing anything about this, and fully admittedly without actually understanding what I was doing, I tried flipping the aspect for the rotary pos embed crop coords, (which probably results it not being applied at all or something), seemingly cleared it up for this resolution, but ruining it for the default one:

https://github.com/user-attachments/assets/6c75f56c-967b-4ba0-9d96-8c3e946c7f1e

zRzRzRzRzRzRzR:
> Hi @kijai. Could you try experimenting with this branch: [zRzRzRzRzRzRzR#1](https://github.com/zRzRzRzRzRzRzR/diffusers/pull/1)?嗨。你能试着在这个分支上实验一下吗： [zRzRzRzRzRzRzR#1](https://github.com/zRzRzRzRzRzRzR/diffusers/pull/1) ？
> 
> It removes the padding of hidden_states which seems to cause corruption. Instead, you must specify frame values whose latent size is divisible by patch_size_t (such as 85 or 165). I believe there might still be some modifications remaining on how to handle frames based on my conversation with Yuxuan.它去除了 hidden_states 的填充，这似乎导致了损坏。相反，您必须指定潜在大小可被 patch_size_t 整除的帧值（例如 85 或 165）。我相信根据我与 Yuxuan 的对话，可能仍然需要对如何处理帧进行一些修改。
> 
> Regarding resolutions, the T2V model works best only at specific resolutions (the recommended is to always do 1360x768). The I2V model can generate at multiple resolutions. Also, both models can generate best at 85 and 165 frames (as per my PR above, but this should actually be 81 and 161 to be consistent with original implementation. We'll try to figure out the best way to support this today)关于分辨率，T2V 模型仅在特定分辨率下效果最佳（推荐始终使用 1360x768）。I2V 模型可以在多个分辨率下生成。此外，这两个模型在 85 和 165 帧下生成效果最佳（根据我上面的 PR，但实际上应该是 81 和 161，以与原始实现保持一致。我们今天会尽力找出最佳支持方式）。

In this version, T2V works very well; I believe the main issue lies in I2V. For T2V, we indeed trained with only one resolution, 1360x768, so it’s quite normal that the original 720x480 resolution doesn’t work properly.

zRzRzRzRzRzRzR:
Hi @kijai. Could you try experimenting with this branch: [zRzRzRzRzRzRzR#1](https://github.com/zRzRzRzRzRzRzR/diffusers/pull/1)?
It removes the padding of hidden_states which seems to cause corruption. Instead, you must specify frame values whose latent size is divisible by patch_size_t (such as 85 or 165). I believe there might still be some modifications remaining on how to handle frames based on my conversation with Yuxuan.
Regarding resolutions, the T2V model works best only at specific resolutions (the recommended is to always do 1360x768). The I2V model can generate at multiple resolutions. Also, both models can generate best at 85 and 165 frames (as per my PR above, but this should actually be 81 and 161 to be consistent with original implementation. We'll try to figure out the best way to support this today)
>
I did try without the padding already, but then you get the first 4 frames without movement, I'll check if that branch has something for that thanks.
>
To illustrate what I mean with the I2V and 720x480 (for example):
CogVideoX-I2V_00001.10.mp4
>
Also forgive me for not knowing anything about this, and fully admittedly without actually understanding what I was doing, I tried flipping the aspect for the rotary pos embed crop coords, (which probably results it not being applied at all or something), seemingly cleared it up for this resolution, but ruining it for the default one:
cogvideo1_5_test.mp4

What I want to say is that it should actually be 21, not 22, because the first block of each latent is useless. Therefore, we need to try to remove the useless latents. This method should effectively output a 16fps video of 5 seconds, with the input num_frames being 81

zRzRzRzRzRzRzR:
Now, all changes are in the cogvideox1.1-5b branch.

a-r-r-o-w:
> Also forgive me for not knowing anything about this, and fully admittedly without actually understanding what I was doing, I tried flipping the aspect for the rotary pos embed crop coords, (which probably results it not being applied at all or something), seemingly cleared it up for this resolution, but ruining it for the default one:

@kijai It could be possible that there is something wrong with our implementation then. Could you provide the patch/changes you made to do the aspect ratio flip?

kijai:
> > Also forgive me for not knowing anything about this, and fully admittedly without actually understanding what I was doing, I tried flipping the aspect for the rotary pos embed crop coords, (which probably results it not being applied at all or something), seemingly cleared it up for this resolution, but ruining it for the default one:
> 
> @kijai It could be possible that there is something wrong with our implementation then. Could you provide the patch/changes you made to do the aspect ratio flip?

I really don't know if this means anything, I only simply swapped the base_size_width and base_size_height here in effort to understand what it even does, but it's not a real solution as it ruins the other resolutions then: 

https://github.com/huggingface/diffusers/blob/27441fc2daf79c7838b43ff280e4e61c7639a40d/src/diffusers/pipelines/cogvideo/pipeline_cogvideox.py#L458

Also I did just now replicate this to double check it indeed somehow fixes at least that video I showed above.

zRzRzRzRzRzRzR:
One point that needs clarification is that the video resolution in I2V implemented by CogVIdeoX1.5 needs to be the same as the image resolution, or 1360*768.

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,23 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
...
 
 There is one model available that can be used with the image-to-video CogVideoX pipeline:
-- [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V): The recommended dtype for running this model is `bf16`.
+- [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V): The recommended dtype for running this model is `torch.bfloat16`.
+- [`THUDM/CogVideoX-1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-1.5-5b-I2V): The recommended dtype for running this mdoel is `torch.bfloat16`.
+
+For the CogVideoX 1.5 series of models, note that:
+- Text-to-video works best at `1360 x 768` resolution because it is trained with that specific resolution
+- Image-to-video works for multiple resolutions. Width can vary from `256` to `1360`, and height can vary from `256` to `768`. Note that the height/width must be divisible by `16`.
+- Both T2V and I2V models support generation with `81` and `161` frames and work best at this value. It is recommended to export videos at 16 FPS.
```
a-r-r-o-w, line 44: @zRzRzRzRzRzRzR LMK if this is what you wanted to convey and correct

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,23 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
 
 This pipeline was contributed by [zRzRzRzRzRzRzR](https://github.com/zRzRzRzRzRzRzR). The original codebase can be found [here](https://huggingface.co/THUDM). The original weights can be found under [hf.co/THUDM](https://huggingface.co/THUDM).
 
-There are two models available that can be used with the text-to-video and video-to-video CogVideoX pipelines:
-- [`THUDM/CogVideoX-2b`](https://huggingface.co/THUDM/CogVideoX-2b): The recommended dtype for running this model is `fp16`.
-- [`THUDM/CogVideoX-5b`](https://huggingface.co/THUDM/CogVideoX-5b): The recommended dtype for running this model is `bf16`.
+There are three official models available that can be used with the text-to-video and video-to-video CogVideoX pipelines:
+- [`THUDM/CogVideoX-2b`](https://huggingface.co/THUDM/CogVideoX-2b): The recommended dtype for running this model is `torch.float16`.
+- [`THUDM/CogVideoX-5b`](https://huggingface.co/THUDM/CogVideoX-5b): The recommended dtype for running this model is `torch.bfloat16`.
+- [`THUDM/CogVideoX-1.5-5b`](https://huggingface.co/THUDM/CogVideoX-1.5-5b): The recommended dtype for running this mdoel is `torch.bfloat16`.
```
stevhliu, line 35:
```suggestion
There are three official CogVideoX checkpoints for text-to-video and video-to-video.
| checkpoints | recommended inference dtype |
|---|---|
| [`THUDM/CogVideoX-2b`](https://huggingface.co/THUDM/CogVideoX-2b) | torch.float16 |
| [`THUDM/CogVideoX-5b`](https://huggingface.co/THUDM/CogVideoX-5b) | torch.bfloat16 |
| [`THUDM/CogVideoX-1.5-5b`](https://huggingface.co/THUDM/CogVideoX-1.5-5b) | torch.bfloat16 |
```

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,23 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
...
 
 There is one model available that can be used with the image-to-video CogVideoX pipeline:
```
stevhliu, line 37:
```suggestion
There are two official CogVideoX checkpoints available for image-to-video.
| checkpoints | recommended inference dtype |
|---|---|
| [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V) | torch.bfloat16 |
| [`THUDM/CogVideoX-1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-1.5-5b-I2V) | torch.bfloat16 |
```

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,23 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
...
-- [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V): The recommended dtype for running this model is `bf16`.
+- [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V): The recommended dtype for running this model is `torch.bfloat16`.
+- [`THUDM/CogVideoX-1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-1.5-5b-I2V): The recommended dtype for running this mdoel is `torch.bfloat16`.
+
+For the CogVideoX 1.5 series of models, note that:
+- Text-to-video works best at `1360 x 768` resolution because it is trained with that specific resolution
+- Image-to-video works for multiple resolutions. Width can vary from `768` to `1360`, and height must be `768`. Note that the height/width must be divisible by `16`.
+- Both T2V and I2V models support generation with `81` and `161` frames and work best at this value. It is recommended to export videos at 16 FPS.
 
 There are two models that support pose controllable generation (by the [Alibaba-PAI](https://huggingface.co/alibaba-pai) team):
```
stevhliu, line 46:
```suggestion
There are two official CogVideoX checkpoints that support pose controllable generation (by the [Alibaba-PAI](https://huggingface.co/alibaba-pai) team).
| checkpoints | recommended inference dtype |
|---|---|
| [`alibaba-pai/CogVideoX-Fun-V1.1-2b-Pose`](https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.1-2b-Pose) | torch.bfloat16 |
| [`alibaba-pai/CogVideoX-Fun-V1.1-5b-Pose`](https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.1-5b-Pose) | torch.bfloat16 |
```

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,23 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
...
 
 There is one model available that can be used with the image-to-video CogVideoX pipeline:
-- [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V): The recommended dtype for running this model is `bf16`.
+- [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V): The recommended dtype for running this model is `torch.bfloat16`.
+- [`THUDM/CogVideoX-1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-1.5-5b-I2V): The recommended dtype for running this mdoel is `torch.bfloat16`.
+
+For the CogVideoX 1.5 series of models, note that:
+- Text-to-video works best at `1360 x 768` resolution because it is trained with that specific resolution
+- Image-to-video works for multiple resolutions. Width can vary from `768` to `1360`, and height must be `768`. Note that the height/width must be divisible by `16`.
+- Both T2V and I2V models support generation with `81` and `161` frames and work best at this value. It is recommended to export videos at 16 FPS.
```
stevhliu, line 44:
```suggestion
For the CogVideoX 1.5 series:
- Text-to-video (T2V) works best at a resolution of 1360x768 because it was trained with that specific resolution.
- Image-to-video (I2V) works for multiple resolutions. The width can vary from 768 to 1360, but the height must be 768. The height/width must be divisible by 16.
- Both T2V and I2V models support generation with 81 and 161 frames and work best at this value. Exporting videos at 16 FPS is recommended.
```

stevhliu (commented):
Cool, thanks!

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
```diff
@@ -1057,6 +1057,7 @@ def __init__(
         force_upcast: float = True,
         use_quant_conv: bool = False,
         use_post_quant_conv: bool = False,
+        invert_scale_latents: bool = False,
```
yiyixuxu, line 1060: I don't think this change is needed - we can just adjust when we scale it, ie.g instead of `* scale_factor`, we do * `1/scale_factor`

src/diffusers/models/transformers/cogvideox_transformer_3d.py
```diff
@@ -170,14 +170,16 @@ class CogVideoXTransformer3DModel(ModelMixin, ConfigMixin, PeftAdapterMixin):
             Whether to flip the sin to cos in the time embedding.
         time_embed_dim (`int`, defaults to `512`):
             Output dimension of timestep embeddings.
+        ofs_embed_dim (`int`, defaults to `512`):
+            scaling factor in the VAE process for the Image-to-Video (I2V) transformation in CogVideoX1.5-5B.
```
yiyixuxu, line 174: this cannot be right,no? what ofs?

yiyixuxu (approved):
thanks! left one comment
feel free to merge after addressing that!

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
a-r-r-o-w: For the original CogVideoX 1.0 models, we need to do `latents * scale_factor`, but for the new version we need to do `latents / scale_factor` in `prepare_latents` IIUC from my conversation with Yuxuan. So, I think we might need it :( I will check again

src/diffusers/models/transformers/cogvideox_transformer_3d.py
a-r-r-o-w: Oh yes, will update

docs/source/en/api/pipelines/cogvideox.md
zRzRzRzRzRzRzR: > @zRzRzRzRzRzRzR LMK if this is what you wanted to convey and correct
Here is the correction: to be precise, it is adapted to 768–1360, not 256–1360.

Actis92: ok, thank you but in the comment below i see this `Image-to-video (I2V) works for multiple resolutions. The width can vary from 768 to 1360, but the height must be 768. The height/width must be divisible by 16` so it means that should work for square and horizontal videos right? But trying with 768x768 is still quite bad 😅 

https://github.com/user-attachments/assets/e92ab46d-8330-4933-a081-487fbf66d309

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
zRzRzRzRzRzRzR: This part is necessary because there is indeed a difference in execution between 1.5 and 1.0 here. The training process for 1.5 forgot to use *scale_factor.

yiyixuxu: why cannot we just update the scale_factor here? for example if it is 5 , we just use 1/5 instead

yiyixuxu: we cannot update the `decode_latents` too? it is creating inconsistency, no?

src/diffusers/models/transformers/cogvideox_transformer_3d.py
zRzRzRzRzRzRzR: What are the specific modifications to this place? This ofs_embed_dim parameter exists in sat

kijai:
I'm still not convinced the rotary embed part is correct, the original SAT code seems to be able to do 768x1360 just fine, here's an example output, wasn't done by me so some settings and seed are different so it's not exact same motion, but it clearly works:

https://github.com/user-attachments/assets/9aafbcf2-7fa7-418a-96d3-7bf01b747c37

Here's same with this with my ComfyUI wrapper nodes using code from this current draft with diffusers:

https://github.com/user-attachments/assets/27961003-f8b1-4200-b05d-eba55145c182

Here's same with same seed etc. but doing that flipping of the base height/width for the `get_resize_crop_region_for_grid` I mentioned before:

https://github.com/user-attachments/assets/4ee21282-7210-4e03-ac8e-92e56015ed15

Still not ruling out I have something else wrong in my implementation, but this seems to go against the resolution docs.

a-r-r-o-w:
Thanks for the detailed response! I will try and take a look soon but will be traveling until tomorrow so it might be a while before I can respond. I'm going to make sure that there's nothing wrong going on in my testing script which numerically matched outputs to 0.03 absmax difference with the SAT codebase for I2V with the same initial settings/embeddings. It could be that some latest change broke the pipeline but last I checked, it was working great at 768x1360 and 768x1152 (HxW) atleast, so apologies if any mistake ended up getting reflected downstream

zRzRzRzRzRzRzR:
https://github.com/user-attachments/assets/fa845c33-6ba8-4cc0-94dd-4b5744fcaa09

We have also reproduced this issue in the latest version. There might indeed have been some changes in between; we will also check it out.

a-r-r-o-w:
@kijai Would you be able to try it out now? Please make sure to update the config with the correct sample height and width as done in the latest commit! It looks to me like the issue you mentioned is fixed (but I have only done limited testing with 768 x 1360, 1360 x 768, 768 x 1152, 1152 x 768).

Another possible bug might be the max_sequence_length. I'm unsure if this should be 226 or 224 for CogVideoX 1.5. @zRzRzRzRzRzRzR Could you clarify which is correct?

I have been on vacation for the past few days, so could not really invest time to carefully check out each change that was made, and missed a few commits made to the original repo. Apologies for the delay here but once @zRzRzRzRzRzRzR and you confirm that it is working as intended, we can proceed with merging here. Thanks for all the help testing!

src/diffusers/models/embeddings.py
```diff
@@ -497,7 +517,14 @@ def forward(self, hidden_states: torch.Tensor, encoder_hidden_states: torch.Tens
...
 
 def get_3d_rotary_pos_embed(
-    embed_dim, crops_coords, grid_size, temporal_size, theta: int = 10000, use_real: bool = True
+    embed_dim,
+    crops_coords,
+    grid_size,
+    temporal_size,
+    theta: int = 10000,
+    use_real: bool = True,
+    grid_type: str = "linspace",
```
a-r-r-o-w, line 526: cc @yiyixuxu

src/diffusers/models/transformers/cogvideox_transformer_3d.py
a-r-r-o-w, line 174:
```suggestion
            Output dimension of "ofs" embeddings used in CogVideoX-5b-I2B in version 1.5
```

src/diffusers/pipelines/cogvideo/pipeline_cogvideox_image2video.py
```diff
@@ -522,18 +538,38 @@ def _prepare_rotary_positional_embeddings(
...
+            base_num_frames = (num_frames + p_t - 1) // p_t
+
+            freqs_cos, freqs_sin = get_3d_rotary_pos_embed(
+                embed_dim=self.transformer.config.attention_head_dim,
+                crops_coords=None,
+                grid_size=(grid_height, grid_width),
+                temporal_size=base_num_frames,
+                grid_type="slice",
+                max_size=(base_size_height, base_size_width),
+            )
```
a-r-r-o-w, line 572: cc @yiyixuxu 

@zRzRzRzRzRzRzR I believe this change will also be required in the T2V script, yes? Because both 1.5 models use slices from the RoPE grid I think. The current implementation works because the base height/width as configured in transformer matches the grid height/width of latent, but for any other resolution, the outputs will numerically diverge.

a-r-r-o-w:
@yiyixuxu Re-requesting a review because of the changes made to 3D rope embeds

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
a-r-r-o-w: So, one way or the other, we will need to determine when to multiply by scale factor and when to divide. The encode method needs to support both (multiply for 1.0, divide for Cog 1.5). In the decode method, we just need to support divide (already exists).

Even if we invert the scale factor during conversion of VAE, we still need to be able to determine which version of the model is running in order for the encode method to work as expected. As of now, we can do this by a check like `if self.transformer.config.patch_embed_t is None: then version 1.0 else version is 1.5`, but I don't think this is a nice way to handle it.

kijai:
> @kijai Would you be able to try it out now? Please make sure to update the config with the correct sample height and width as done in the latest commit! It looks to me like the issue you mentioned is fixed (but I have only done limited testing with 768 x 1360, 1360 x 768, 768 x 1152, 1152 x 768).
> 
> Another possible bug might be the max_sequence_length. I'm unsure if this should be 226 or 224 for CogVideoX 1.5. @zRzRzRzRzRzRzR Could you clarify which is correct?
> 
> I have been on vacation for the past few days, so could not really invest time to carefully check out each change that was made, and missed a few commits made to the original repo. Apologies for the delay here but once @zRzRzRzRzRzRzR and you confirm that it is working as intended, we can proceed with merging here. Thanks for all the help testing!

I don't have time to test more currently as it's 3am here, but quickly testing both examples that failed before indeed works now:

https://github.com/user-attachments/assets/ab2aa814-5187-4e79-a6fc-069f6255f8bf

https://github.com/user-attachments/assets/9ac7e203-c8fb-4c88-a348-395624a258a8

Thank you for your hard work!

zRzRzRzRzRzRzR:
I have replied to some of the issues mentioned by @a-r-r-o-w on Slack. I am currently reproducing the issue mentioned in the latest issue report.

zRzRzRzRzRzRzR:
အားလုပ်ဆောင်မှု အကြောင်းအရာ အား ၇၆၈ * ၁၃၆၀ အတွက် မမှန်ကန်ပါဘူး၊ သို့ရာတွင် @kijai နဲ့ @a-r-r-o-w တို့နှစ်ဦးစလုံး အောင်မြင်ခဲ့ကြသလား?
ကျွန်ုပ်၏ အလုပ်လုပ်ခြင်းတွင် @kijai က ယခင်က ဖော်ပြခဲ့သော မှန်းမှန်းချက် အမှားတစ်ခု ကျယ်လောင်းခဲ့သည်။

kijai:
> The content of the activity for 768 x 1360 is incorrect, but did both @kijai and @a-r-r-o-w succeed? In my work, @kijai made a broad assumption about an error previously mentioned.

The latest fix from @a-r-r-o-w is working perfectly for me, every resolution I've tried has worked to some extend, even really small ones.

a-r-r-o-w:
@zRzRzRzRzRzRzR Let me know if you think this is good to merge now, and we can go ahead :)

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,33 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
...
-
-There are two models that support pose controllable generation (by the [Alibaba-PAI](https://huggingface.co/alibaba-pai) team):
-- [`alibaba-pai/CogVideoX-Fun-V1.1-2b-Pose`](https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.1-2b-Pose): The recommended dtype for running this model is `bf16`.
-- [`alibaba-pai/CogVideoX-Fun-V1.1-5b-Pose`](https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.1-5b-Pose): The recommended dtype for running this model is `bf16`.
+There are three official CogVideoX checkpoints for text-to-video and video-to-video.
+| checkpoints | recommended inference dtype |
+|---|---|
+| [`THUDM/CogVideoX-2b`](https://huggingface.co/THUDM/CogVideoX-2b) | torch.float16 |
+| [`THUDM/CogVideoX-5b`](https://huggingface.co/THUDM/CogVideoX-5b) | torch.bfloat16 |
+| [`THUDM/CogVideoX-1.5-5b`](https://huggingface.co/THUDM/CogVideoX-1.5-5b) | torch.bfloat16 |
```
a-r-r-o-w, line 37:
```suggestion
| [`THUDM/CogVideoX1.5-5b`](https://huggingface.co/THUDM/CogVideoX1.5-5b) | torch.bfloat16 |
```

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,33 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
...
+
+There are two official CogVideoX checkpoints available for image-to-video.
+| checkpoints | recommended inference dtype |
+|---|---|
+| [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V) | torch.bfloat16 |
+| [`THUDM/CogVideoX-1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-1.5-5b-I2V) | torch.bfloat16 |
+- [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V): The recommended dtype for running this model is `torch.bfloat16`.
+- [`THUDM/CogVideoX-1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-1.5-5b-I2V): The recommended dtype for running this mdoel is `torch.bfloat16`.
```
a-r-r-o-w, line 45:
```suggestion
- [`THUDM/CogVideoX1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX1.5-5b-I2V): The recommended dtype for running this mdoel is `torch.bfloat16`.
```

a-r-r-o-w (approved):

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,33 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
...
+- Image-to-video (I2V) works for multiple resolutions. The width can vary from 768 to 1360, but the height must be 768. The height/width must be divisible by 16.
+- Both T2V and I2V models support generation with 81 and 161 frames and work best at this value. Exporting videos at 16 FPS is recommended.
+
+There are two official CogVideoX checkpoints that support pose controllable generation (by the [Alibaba-PAI](https://huggingface.co/alibaba-pai) team).
+| checkpoints | recommended inference dtype |
+|---|---|
+| [`alibaba-pai/CogVideoX-Fun-V1.1-2b-Pose`](https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.1-2b-Pose) | torch.bfloat16 |
+| [`alibaba-pai/CogVideoX-Fun-V1.1-5b-Pose`](https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.1-5b-Pose) | torch.bfloat16 |
+- [`alibaba-pai/CogVideoX-Fun-V1.1-2b-Pose`](https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.1-2b-Pose): The recommended dtype for running this model is `torch.bfloat16`.
+- [`alibaba-pai/CogVideoX-Fun-V1.1-5b-Pose`](https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.1-5b-Pose): The recommended dtype for running this model is `torch.bfloat16`.
```
a-r-r-o-w, line 58:
```suggestion
```

docs/source/en/api/pipelines/cogvideox.md
```diff
@@ -29,16 +29,33 @@ Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers.m
...
+| [`THUDM/CogVideoX-5b`](https://huggingface.co/THUDM/CogVideoX-5b) | torch.bfloat16 |
+| [`THUDM/CogVideoX1.5-5b`](https://huggingface.co/THUDM/CogVideoX1.5-5b) | torch.bfloat16 |
+
+There are two official CogVideoX checkpoints available for image-to-video.
+| checkpoints | recommended inference dtype |
+|---|---|
+| [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V) | torch.bfloat16 |
+| [`THUDM/CogVideoX-1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-1.5-5b-I2V) | torch.bfloat16 |
+- [`THUDM/CogVideoX-5b-I2V`](https://huggingface.co/THUDM/CogVideoX-5b-I2V): The recommended dtype for running this model is `torch.bfloat16`.
+- [`THUDM/CogVideoX1.5-5b-I2V`](https://huggingface.co/THUDM/CogVideoX1.5-5b-I2V): The recommended dtype for running this mdoel is `torch.bfloat16`.
```
a-r-r-o-w, line 45:
```suggestion
```

Actis92:
I have tried now to compare the diffusers version and the SAT one, IMO the diffusers version is working better respect some days ago but still the quality is lower. For the experiment i have used this image that is 768x768 with the prompt "an old woman walking on the street"
a
in the first video you can see an example obtained using SAT version, the second video using the current diffusers version. Already the first frame is different seems over saturated, then SAT version is really smooth instead diffusers has artifacts. I still need to go deep in the code to find the reason but i hope that can be useful for you

https://github.com/user-attachments/assets/1200c931-bd59-4942-9dd2-d0fe97e07e4b

https://github.com/user-attachments/assets/f6f83f25-2acf-4a42-bf32-852edf39a5d5

YanzuoLu:
The current settings for T2V transformer model seem to problematic.
Any ideas on this [issue](https://github.com/THUDM/CogVideo/issues/520)?

a-r-r-o-w:
Just fixed in #9963. A config attribute was updated on the transformer checkpoint to make the T2V behaviour consistent with SAT codebase, but the specific change only existed in I2V for diffusers - but now should work same as before

Issue #9850, opened 2024-11-03
make gradient checkpointing with frozen model possible


MikeTkachuk:
### Describe the bug

https://github.com/huggingface/diffusers/blob/89e4d6219805975bd7d253a267e1951badc9f1c0/src/diffusers/models/unets/unet_2d_blocks.py#L862

hi, the clause i highlighted in the link above prevents a model from using gradient checkpointing in eval mode. this is particularly useful for e.g. LORAs.
perhaps you meant to check something like this instead?
`if torch.is_grad_enabled() and self.gradient_checkpointing:`

Same for any other module in the repo

### Reproduction
```python
import torch
from diffusers.models.unets.unet_2d_blocks import UNetMidBlock2DCrossAttn

block = UNetMidBlock2DCrossAttn(32, 32, 32, cross_attention_dim=32)
block.gradient_checkpointing = True
block.eval()

block(torch.randn((1, 32, 64, 64)), torch.randn((1, 32,)))

```

### System Info

- 🤗 Diffusers version: 0.30.3
- Platform: Windows-10-10.0.22631-SP0
- Running on Google Colab?: No
- Python version: 3.10.13
- PyTorch version (GPU?): 2.4.1+cu118 (True)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Huggingface_hub version: 0.25.1
- Transformers version: 4.45.1
- Accelerate version: 0.34.2
- PEFT version: 0.12.0
- Bitsandbytes version: 0.44.1
- Safetensors version: 0.4.5
- xFormers version: 0.0.28.post3
- Accelerator: NVIDIA GeForce RTX 4060 Laptop GPU, 8188 MiB
- Using GPU in script?: No
- Using distributed or parallel set-up in script?: No

### Who can help?

Other: @yiyixuxu @DN6

SahilCarterr:
The issue raised here highlights a specific use case: enabling gradient checkpointing during evaluation mode (i.e., eval() mode) when gradients are still needed. This scenario could apply in cases like LoRA (Low-Rank Adaptation) or other fine-tuning strategies where we might still compute gradients on a pre-trained model while in evaluation mode.

### Reproduction
```python
import torch
from diffusers.models.unets.unet_2d_blocks import UNetMidBlock2DCrossAttn

# Define the modified UNetMidBlock2DCrossAttn subclass
[...]

PR #9878, opened 2024-11-06
Enabling gradient checkpointing in eval() mode

MikeTkachuk:
Removed unnecessary `if self.training ...` check when using gradient checkpointing
https://github.com/huggingface/diffusers/issues/9850


MikeTkachuk:
Since all of the module implementations used the same `if self.training and self.gradient_checkpointing:` clause, I wonder if there is some reference template for `def forward` that others are using? Need to make sure future implementations are using the fixed clause

src/diffusers/models/transformers/cogvideox_transformer_3d.py
```diff
@@ -452,7 +452,7 @@ def forward(
 
         # 3. Transformer blocks
         for i, block in enumerate(self.transformer_blocks):
-            if self.training and self.gradient_checkpointing:
+            if self.gradient_checkpointing:
```
yiyixuxu, line 455: oh thanks! why do we also removed the `torch.is_grad_enabled()` check? gradient checkpointing isn't meaningful without gradient being computed, no?

MikeTkachuk: added it back, thanks for pointing it out. 
it does not break anything, but found that it throws an annoying warning when use_reentrant=True,

yiyixuxu: >but found that it throws an annoying warning when use_reentrant=True,

what do you mean by that?

MikeTkachuk: use_reentrant is an argument passed to torch.utils.checkpoint.checkpoint

if True one of the checks will print this to stderr
warnings.warn(
            "None of the inputs have requires_grad=True. Gradients will be None"
        )
 but diffusers are using use_reentrant=False anyway

yiyixuxu: oh got thanks, so the warning is specific to when we use gradient checkpointing when gradient is not enabled

yiyixuxu:
hi @MikeTkachuk unfortunately we have to update the branch and resolve conflicts now...
would you be able to do that? I'm cool with the change otherwise and can merge once it is synced with main

MikeTkachuk:
done, also fixed in a few other places I missed

docs/source/en/api/models/controlnet.md
```diff
@@ -39,12 +39,12 @@ pipe = StableDiffusionControlNetPipeline.from_single_file(url, controlnet=contro
 
 ## ControlNetOutput
 
-[[autodoc]] models.controlnet.ControlNetOutput
+[[autodoc]] models.controlnets.controlnet.ControlNetOutput
```
yiyixuxu, line 42: this change is not made from this PR 
it should not show up here, no?

yiyixuxu:
can you rebase again?
I saw a bunch of changes that are not made in this PR and the commit history includes commits  that's already been merged into main (from another PR)

MikeTkachuk:
fixed

yiyixuxu (approved):
thanks!

PR #9882, opened 2024-11-07
Use parameters + buffers when deciding upscale_dtype

universome:
Sometimes, the decoder might lack parameters and only buffers (e.g., this happens when we manually need to convert all the parameters to buffers — e.g. to avoid packing fp16 and fp32 parameters with FSDP)

# What does this PR do?

Fixes # (issue)

## Before submitting

## Who can review?


sayakpaul:
Could you maybe elaborate a bit more where this is useful?

sayakpaul:
@universome a gentle ping

DN6 (approved):

DN6:
@universome Would it be possible to handle the merge conflict? We can merge this after.

universome:
Hi @DN6 , sorry for the late reply. Basically, my use-case is the following.
I'm training a video latent diffusion model with the CogVideoX autoencoder with FSDP wrapping.
I instantiate the autoencoder inside the base model class, wrap the base model in FSDP, and now I need to make sure that it does not get FSDP-wrapped (to avoid extra syncs for it since it's quite small in terms of the parameter counts anyway).
One way to do this is to simply convert AE parameters to buffers (which would cause the error above). The other is to add changes to FSDP wrap policy to make sure it's not wrapped. I felt that the conversion to buffers is simpler since in this case I also do not need to bother about setting requires_grad to False for them.

PR #9903, opened 2024-11-11
[refactor] CogVideoX VAE

a-r-r-o-w:
# What does this PR do?

Refactors the CogVideoX VAE to make the implementation similar to Mochi VAE for consistency. Will follow-up on Allegro in a separate PR

## Who can review?

@yiyixuxu @sayakpaul


src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
```diff
@@ -1108,18 +1122,11 @@ def __init__(
...
 
         # We make the minimum height and width of sample for tiling half that of the generally supported
-        self.tile_sample_min_height = sample_height // 2
-        self.tile_sample_min_width = sample_width // 2
-        self.tile_latent_min_height = int(
-            self.tile_sample_min_height / (2 ** (len(self.config.block_out_channels) - 1))
-        )
-        self.tile_latent_min_width = int(self.tile_sample_min_width / (2 ** (len(self.config.block_out_channels) - 1)))
+        self.tile_sample_min_height = 256
+        self.tile_sample_min_width = 256
```
sayakpaul, line 1126: Can we make sure this is 100% not backwards-breaking?

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
```diff
@@ -1108,18 +1122,11 @@ def __init__(
...
 
-        # These are experimental overlap factors that were chosen based on experimentation and seem to work best for
-        # 720x480 (WxH) resolution. The above resolution is the strongly recommended generation resolution in CogVideoX
-        # and so the tiling implementation has only been tested on those specific resolutions.
-        self.tile_overlap_factor_height = 1 / 6
-        self.tile_overlap_factor_width = 1 / 5
+        self.tile_sample_stride_height = 192
+        self.tile_sample_stride_width = 192
```
sayakpaul, line 1129: Why are we completely getting rid of the `overlap_factor`s? It is still used in `enable_tiling()` no? Oh because we're deprecating them. Okay.

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
```diff
@@ -1142,24 +1151,28 @@ def enable_tiling(
...
+            tile_sample_stride_width (`int`, *optional*):
+                The stride between two consecutive horizontal tiles. This is to ensure that there are no tiling
+                artifacts produced across the width dimension.
         """
+
+        if tile_overlap_factor_height is not None or tile_overlap_factor_width is not None:
+            deprecate(
+                "tile_overlap_factor",
+                "1.0.0",
+                "The parameters `tile_overlap_factor_height` and `tile_overlap_factor_width` are deprecated. Please use `tile_sample_stride_height` and `tile_sample_stride_width` instead.",
```
sayakpaul, line 1166:
```suggestion
                "The parameters `tile_overlap_factor_height` and `tile_overlap_factor_width` are deprecated and will be ignored. Please use `tile_sample_stride_height` and `tile_sample_stride_width` instead. For now, we will use these flags automatically without breaking the existing behaviour.",
```

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
```diff
@@ -1142,24 +1151,28 @@ def enable_tiling(
...
+            )
+            tile_sample_stride_height = int((1 - tile_overlap_factor_height) * self.tile_sample_min_height) // 8 * 8
+            tile_sample_stride_width = int((1 - tile_overlap_factor_width) * self.tile_sample_min_width) // 8 * 8
```
sayakpaul, line 1169: Should `8` be kept into a sensible variable for better readability?

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
```diff
@@ -1085,9 +1085,23 @@ def __init__(
...
+
+        # When decoding spatially large video latents, the memory requirement is very high. By breaking the video latent
+        # frames spatially into smaller tiles and performing multiple forward passes for decoding, and then blending the
+        # intermediate tiles together, the memory requirement can be lowered.
         self.use_tiling = False
 
+        # When decoding temporally long video latents, the memory requirement is very high. By decoding latent frames
+        # at a fixed frame batch size (based on `self.num_latent_frames_batch_sizes`), the memory requirement can be lowered.
+        self.use_framewise_encoding = True
+        self.use_framewise_decoding = True
```
sayakpaul, line 1103: If we're fixing them to `True`, does it even make sense to have these flags then? If we're mutating later, then maybe add a note about it here to make it clear?

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
```diff
@@ -1323,44 +1337,48 @@ def tiled_encode(self, x: torch.Tensor) -> torch.Tensor:
...
-        row_limit_width = self.tile_latent_min_width - blend_extent_width
-        frame_batch_size = self.num_sample_frames_batch_size
+        blend_height = tile_latent_min_height - tile_latent_stride_height
+        blend_width = tile_latent_min_width - tile_latent_stride_width
 
         # Split x into overlapping tiles and encode them separately.
         # The tiles have an overlap to avoid seams between tiles.
         rows = []
-        for i in range(0, height, overlap_height):
+        for i in range(0, height, self.tile_sample_stride_height):
```
sayakpaul, line 1354: Feel free to disregard this comment entirely but I think it could make sense to add some comments before the for loops as each one of them includes a lot of indexing operations and it can be difficult for readability.

sayakpaul (commented):
This is promising! I left some comments, many of which are nits. LMK if they make sense.

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
a-r-r-o-w: The overlap factors are more complex to understand and work with IMO, and their usage to calculate strides are prone to rounding errors (for example, if you wanted the stride to be 192, you would have to try to figure out the correct overlap factor fraction, and even then you could end up with something like 191 or 193). I feel like this makes things easier to understand and simple, but it is mostly a personal preference. I know that the CogVideoX ComfyUI wrapper allows users to modify these settings, and some overlap values can lead to artifacts when decoding.

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
a-r-r-o-w: Ah yeah. This should be self.spatial_compression_ratio

src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
a-r-r-o-w: Remember we discussed about allowing one shot decoding and applying quantization on the activations? Allowing framewise stuff to default, but configurably opt-out would be helpful for that because until now we did not support oneshot decoding

a-r-r-o-w:
Gentle ping @DN6

a-r-r-o-w:
Converting to draft for a bit. I had something in mind that I wanted to try. Will open for review again soon

PR #9711, opened 2024-10-18
[refactor] enhance readability of flux related pipelines

Luciennnnnnn:
# What does this PR do?

This PR refactors the FLUX-related pipelines to improve clarity and consistency, particularly in regards to height and width determination. The changes align the implementation more closely with other pipelines in the library, enhancing overall code coherence and maintainability.

The original code obscured the details of how latents packing affects resolution, but this change makes the process more transparent and easier to understand.

@yiyixuxu  @asomoza

PR #9929, opened 2024-11-14
Flux latents fix

DN6:
# What does this PR do?
This PR: #9711 changed the latent preparation step so that the `vae_scale_factor` and default sample size are more clear. 

However, we overlooked the fact that this operation (with scale_factor == 16)

```python
        height = 2 * (int(height) // self.vae_scale_factor)
        width = 2 * (int(width) // self.vae_scale_factor)
```

Is not equivalent to this operation with scale_factor == 8
```python
height = int(height) // self.vae_scale_factor
```
For resolutions that are divisible by 8 but not by 16 the current pipeline on main will error out, because the implementation does not account for the fact that the latent width and height must be divisible by 2 for the packing step to work.

The old implementation scaled the height and width by `//16` and then multiplied by 2. This results in slightly resized images if the height or width aren't multiples of 16.

This PR:
1. Adds steps to account for latent packing. Which will result in resized output images (like we had before)
2. Adds a warning that the image will be resized if an incompatible image is provided
3. Adds fast tests to check Flux with expected and unexpected image shapes to verify that the scaling is properly applied.      

Fixes # (issue)

## Before submitting

## Who can review?


bghira:
this seems really important but considering how slowly things move around here, not so sure it will be merged in a timely manner.

src/diffusers/pipelines/flux/pipeline_flux.py
```diff
@@ -501,8 +505,10 @@ def prepare_latents(
         generator,
         latents=None,
     ):
-        height = int(height) // self.vae_scale_factor
-        width = int(width) // self.vae_scale_factor
+        # VAE applies 8x compression on images but we must also account for packing which requires
+        # latent height and width to be divisible by 2.
+        height = int(height) // self.vae_scale_factor - ((int(height) // self.vae_scale_factor) % 2)
```
yiyixuxu, line 510: is this same as this? (basically revert back to the original code but divisible by `self.vae_scale_factor*2`
a little bit easier to understand, no?
```python
height = 2 * (int(height) // (self.vae_scale_factor*2))
```

yiyixuxu (approved):
oh thanks for the fix!
I made a comment

DN6:
Failing LoRA tests are unrelated. Merging.

PR #9943, opened 2024-11-17
[LoRA] enable LoRA for Mochi-1

sayakpaul:
# What does this PR do?

Needed for fine-tuning.


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2362,7 +2362,7 @@ def save_lora_weights(
 
 class CogVideoXLoraLoaderMixin(LoraBaseMixin):
     r"""
-    Load LoRA layers into [`CogVideoXTransformer3DModel`]. Specific to [`CogVideoX`].
+    Load LoRA layers into [`CogVideoXTransformer3DModel`]. Specific to [`CogVideoXPipeline`].
```
sayakpaul, line 2365: Unrelated change but doesn't hurt I guess.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2667,6 +2667,314 @@ def unfuse_lora(self, components: List[str] = ["transformer", "text_encoder"], *
         super().unfuse_lora(components=components)
 
 
+class Mochi1LoraLoaderMixin(LoraBaseMixin):
```
sayakpaul, line 2670: A copy-paste of the Cog LoRA loader classes, indicated by the "Copied from ..." comments.

src/diffusers/loaders/lora_pipeline.py
a-r-r-o-w: Looks okay to fix here

tests/lora/test_lora_layers_mochi.py
```diff
@@ -0,0 +1,173 @@
...
+
+            self.assertTrue(np.isnan(out).all())
+
+    def test_simple_inference_with_text_lora_denoiser_fused_multi(self):
+        super().test_simple_inference_with_text_lora_denoiser_fused_multi(expected_atol=9e-3)
+
+    def test_simple_inference_with_text_denoiser_lora_unfused(self):
+        super().test_simple_inference_with_text_denoiser_lora_unfused(expected_atol=9e-3)
+
+    @unittest.skip("Not supported in Mochi.")
```
a-r-r-o-w, line 143: Looks good since not supporting T5 finetuning!

a-r-r-o-w (approved):
Looks great! Don't see any major differences in the lora mixin, so I think everything should be good to merge once the finetuning script is working - going to take a look at it now, thanks!

sayakpaul:
> so I think everything should be good to merge once the finetuning script is working

@a-r-r-o-w thanks! Do you think it could make sense to merge regardless with the tests, etc.? Because this will also allow others to experiment from `main`. I am okay otherwise too.

yiyixuxu (approved):

sayakpaul:
The tests failing are related to fact that our CI is currently in PyTorch 2.4. https://github.com/huggingface/diffusers/pull/9961 should fix this. Additional info:
https://huggingface.slack.com/archives/C014N4749J9/p1732002990541779

sayakpaul:
Seems like the `NaN` tests are failing with PyTorch 2.5 and when using "CPU" as the test device. Will investigate deeper. It doesn't happen on PyTorch 2.4. Note that all the tests pass when using a GPU even with PyTorch 2.5. 

@yiyixuxu can I merge this PR or should we first investigate the `NaN` issue? It's failing for other PRs too. 

Cc: @BenjaminBossan as well. It's the `safe_merge` argument in the `merge()` method of `peft`. Does this sound familiar?

yiyixuxu:
> @yiyixuxu can I merge this PR or should we first investigate the NaN issue? It's failing for other PRs too.

 @a-r-r-o-w wants to test it with fine-tune script first, no?

sayakpaul:
> @a-r-r-o-w wants to test it with fine-tune script first, no?

Well, that is currently done by me here https://github.com/a-r-r-o-w/cogvideox-factory/pull/90/ with Aryan's reviews. However, I don't think the LoRA implementation is dependent on the fine-tuning experiments as it's less likely gonna change (and it's similar to CogVideoX). But okay to wait and will defer to @a-r-r-o-w.

a-r-r-o-w:
I think should be okay to merge without waiting for finetuning script to work or having an available checkpoint since it will unblock others trying to work on finetuning scripts based on Diffusers (I don't know if there is anyone apart from us yet though). Thanks!

PR #9945, opened 2024-11-17
Correct pipeline_output.py to the type Mochi

twobob:
# What does this PR do?

The autodocs are incorrect as a result of this oversight in the header

Mochi is the correct pipeline type

image

## Before submitting
- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

## Who can review?
originally 
Co-authored-by: Dhruv Nair 
Co-authored-by: yiyixuxu 

@yiyixuxu 
@stevhliu @sayakpaul


yiyixuxu (approved):
thanks!

PR #9963, opened 2024-11-19
Make CogVideoX RoPE implementation consistent

a-r-r-o-w:
# What does this PR do?

CogVideoX 1.5 models use a slightly different RoPE implementation compared to 1.0. When we merged the CogVideoX 1.5 PR, this was the status of RoPE implementations:
- 1.0 T2V and I2V used 1.0 RoPE implementation
- 1.5 I2V used 1.5 RoPE implementation
- 1.5 T2V used 1.0 RoPE implementation (inconsistent)

All models still worked as expected and there weren't any bugs. This PR just makes sure that 1.5 T2V uses 1.5 RoPE implementation (the reason why it still worked before is because linspace and slice implementation are equivalent at the 1360x768 resolution, which is the only resolution 1.5 T2V supports, but differ otherwise). The 1.5 T2V needs to use sliced implementation though to keep things consistent with the original code base (even though the outputs match with 1.0 implementation).

The config for 1.5 T2V on the official checkpoints has been correctly updated (https://huggingface.co/THUDM/CogVideoX1.5-5B/commit/6b80d8556f820bf58e4e9719ec71450aff55c246), but unless we also make these changes, it will become incompatible. So requesting a speedy merge @DN6

### DIFF
--- a/src/diffusers/pipelines/cogvideo/pipeline_cogvideox.py
+++ b/src/diffusers/pipelines/cogvideo/pipeline_cogvideox.py
@@ -446,19 +446,34 @@ class CogVideoXPipeline(DiffusionPipeline, CogVideoXLoraLoaderMixin):
         p = self.transformer.config.patch_size
         p_t = self.transformer.config.patch_size_t or 1
 
-        base_size_width = self.transformer.config.sample_width // p
-        base_size_height = self.transformer.config.sample_height // p
-        base_num_frames = (num_frames + p_t - 1) // p_t
+        if p_t is None:
+            # CogVideoX 1.0
+            base_size_width = self.transformer.config.sample_width // p
+            base_size_height = self.transformer.config.sample_height // p
 
-        grid_crops_coords = get_resize_crop_region_for_grid(
-            (grid_height, grid_width), base_size_width, base_size_height
-        )
-        freqs_cos, freqs_sin = get_3d_rotary_pos_embed(
-            embed_dim=self.transformer.config.attention_head_dim,
-            crops_coords=grid_crops_coords,
-            grid_size=(grid_height, grid_width),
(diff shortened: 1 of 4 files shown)


src/diffusers/pipelines/cogvideo/pipeline_cogvideox.py
```diff
@@ -446,19 +446,34 @@ def _prepare_rotary_positional_embeddings(
         p = self.transformer.config.patch_size
         p_t = self.transformer.config.patch_size_t or 1
 
-        base_size_width = self.transformer.config.sample_width // p
-        base_size_height = self.transformer.config.sample_height // p
-        base_num_frames = (num_frames + p_t - 1) // p_t
+        if p_t is None:
+            # CogVideoX 1.0
+            base_size_width = self.transformer.config.sample_width // p
+            base_size_height = self.transformer.config.sample_height // p
```
DN6, line 452: This part is the same if `p_t == None` or not right? Can't we just move it outside the if/else so the lines aren't duplicated?

DN6 (approved):

a-r-r-o-w:
Gentle ping to @zRzRzRzRzRzRzR. Maybe @Kijai as well to reflect downstream but I'm assuming you've done that already

src/diffusers/pipelines/cogvideo/pipeline_cogvideox.py
a-r-r-o-w: Oh yep, moving it out now

a-r-r-o-w:
Running slow tests now to confirm all checkpoints still work as expected

a-r-r-o-w:
All outputs are matching, so looks good to merge. Failing test is unrelated

Issue #9487, opened 2024-09-20
Add GGUF loader for FluxTransformer2DModel


vladmandic:
[GGUF](https://huggingface.co/docs/hub/en/gguf) is becoming a preferred means of distribution of FLUX fine-tunes.

Transformers recently added general support for GGUF and are slowly adding support for [additional model types](https://github.com/huggingface/transformers/issues/33260).
(implementation is by adding `gguf_file` param to `from_pretrained` method)

[This PR](https://github.com/huggingface/transformers/pull/33389) adds support for loading GGUF files to `T5EncoderModel`.
I've tested the code with quants available at <https://huggingface.co/city96/t5-v1_1-xxl-encoder-gguf/tree/main> and its working with current Flux implementation in diffusers.

However, as `FluxTransformer2DModel` is defined in diffusers library, support has to be added here to be able to load actual transformer model which is most (if not all) of Flux finetunes.

Examples that can be used:
- <https://civitai.com/models/657607/gguf-fastflux-flux1-schnell-merged-with-flux1-dev>  
  with weights quantized as q4_0, q4_1, q5_0, q5_1
- <https://civitai.com/models/662958/flux1-dev-gguf-f16>  
  with weights simply converted from f16

cc: @yiyixuxu @sayakpaul @DN6

sayakpaul:
Perhaps after #9213. 

Note that exotic FPX schemes are already supported (FP6, FP5, FP4) with torchao. Check out this repo for that: https://github.com/sayakpaul/diffusers-torchao

vladmandic:
yes, i'm following that pr closely :)
also, torchao work makes all this easier. request here is not to reimplement any of the quantization work done so far, but to add diffusers equivalent of `transformers.modeling_gguf_pytorch_utils.load_gguf_checkpoint` returns `state_dict` (with key re-mapping as needed) and then the rest of the load can be as-is.

sayakpaul:
Yeah for sure. Thanks for following along!

sayakpaul:
Right up our alley. Cc: @DN6

vladmandic:
@sayakpaul @DN6 if you want to take a look...

simple implementation of generic gguf loader that loads `state_dict`:
<https://github.com/vladmandic/automatic/blob/dev/modules/ggml/__init__.py>

[...]

PR #9964, opened 2024-11-19
[Single File] Add GGUF support

DN6:
# What does this PR do?
Adds support for loading GGUF checkpoints via `from_single_file`. 

```python
import torch

from diffusers import FluxPipeline, FluxTransformer2DModel, GGUFQuantizationConfig

ckpt_path = (
    "https://huggingface.co/city96/FLUX.1-dev-gguf/blob/main/flux1-dev-Q2_K.gguf"
)
transformer = FluxTransformer2DModel.from_single_file(
    ckpt_path,
    torch_dtype=torch.bfloat16,
)
    transformer=transformer,
    torch_dtype=torch.bfloat16,
)
image = pipe(
    prompt_embeds=prompt_embeds, pooled_prompt_embeds=pooled_prompt_embeds
).images[0]
image.save("flux-gguf.png")
```

Notes:
1. API for loading GGUF is a bit overkill, but it's consistent with quantized loading in `from_pretrained`. GGUF files have enough metadata that we can automatically infer everything we need from the file itself. We don't really need a quantization config, but it becomes necessary as we expand to support to other quant loading methods (BnB, TorchAO etc)

TODOS:
- Benchmark loading and inference speed.
- Verify output quality
- Add tests
  

Fixes # (issue)
 #9487

## Before submitting

## Who can review?


src/diffusers/quantizers/bitsandbytes/bnb_quantizer.py
```diff
@@ -204,7 +204,10 @@ def create_quantized_param(
 
         module._parameters[tensor_name] = new_value
 
-    def check_quantized_param_shape(self, param_name, current_param_shape, loaded_param_shape):
+    def check_quantized_param_shape(self, param_name, current_param, loaded_param):
```
DN6, line 207: GGUF needs to access the tensor quant type to run a shape check. So this needs to change from passing in shapes to passing in params directly.

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,369 @@
...
+# # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# # See the License for the specific language governing permissions and
+# # limitations under the License.
+
+import gguf
+import torch
+import torch.nn as nn
+
+
+def _replace_with_gguf_linear(model, compute_dtype):
```
DN6, line 20: GGUF files contain a mix of quantized linear and unquantized linear layers. It's not trivial to selectively replace layers. We can replace all of them and then check the parameter type when running forward instead.

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,369 @@
...
+    for name, module in model.named_children():
+        if isinstance(module, nn.Linear):
+            model._modules[name] = GGUFLinear(
+                module.in_features,
+                module.out_features,
+                module.bias is not None,
+                compute_dtype=compute_dtype,
+            )
+            model._modules[name].source_cls = type(module)
+            # Force requires grad to False to avoid unexpected errors
```
bghira, line 30: what is Force?

src/diffusers/quantizers/gguf/utils.py
DN6: nvm

src/diffusers/quantizers/bitsandbytes/bnb_quantizer.py
sayakpaul: Why not add this method to the `gguf_quantizer.py` file instead of modifying this? This would be a breaking change no?

sayakpaul: I see you're already adding this to the GGUF quantizer class. So, maybe okay to not modify this?

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,392 @@
...
+
+        if isinstance(module, nn.Linear) and _should_convert_to_gguf(module, state_dict, module_prefix):
+            model._modules[name] = GGUFLinear(
+                module.in_features,
+                module.out_features,
+                module.bias is not None,
+                compute_dtype=compute_dtype,
+            )
+            model._modules[name].source_cls = type(module)
+            # Force requires grad to False to avoid unexpected errors
```
sayakpaul, line 48:
```suggestion
            # Force requires_grad to False to avoid unexpected errors
```

yiyixuxu (commented):
oh thanks!
I left some comments but overall looks good to me! (I didn't look into all these methods on utils.py, but I can go over them in details if you need me to 😛)

yiyixuxu:
@SunMarc can you do a review too?

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,109 @@
...
+
+        self.compute_dtype = quantization_config.compute_dtype
+        self.pre_quantized = True
+
+    def validate_environment(self, *args, **kwargs):
+        if not is_accelerate_available() or is_accelerate_version("<", "0.26.0"):
+            raise ImportError(
+                "Loading GGUF Parameters requires `accelerate` installed in your enviroment: `pip install 'accelerate>=0.26.0'`"
+            )
+        if not is_gguf_available():
```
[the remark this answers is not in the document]
SunMarc: I think that just checking it here should be enough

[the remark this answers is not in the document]
SunMarc: I think that just checking it here should be enough + check version

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,384 @@
...
+    n_blocks = tensor.numel() // type_size
+    blocks = tensor.reshape((n_blocks, type_size))
+
+    dequant = dequant_fn(blocks, block_size, type_size)
+    dequant = dequant.reshape(shape)
+
+    return dequant.as_tensor()
+
+
+class GGUFParameter(torch.Tensor):
```
sayakpaul, line 326: Would it make sense to also include this and `GGUFLinear` classes in the `__init__.py` of `gguf` module?

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,384 @@
...
+
+import gguf
+import torch
+import torch.nn as nn
+
+
+def _replace_with_gguf_linear(model, compute_dtype, state_dict, prefix=""):
+    def _should_convert_to_gguf(module, state_dict, prefix):
+        weight_key = prefix + "weight"
+        return weight_key in state_dict and isinstance(state_dict[weight_key], GGUFParameter)
```
sayakpaul, line 24: This seems to be an independent function and can stay outside of `_replace_with_gguf_linear()`?

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,384 @@
...
+
+    has_children = list(model.children())
+    if not has_children:
+        return
+
+    for name, module in model.named_children():
+        module_prefix = prefix + name + "."
+        _replace_with_gguf_linear(module, compute_dtype, state_dict, module_prefix)
+
+        if isinstance(module, nn.Linear) and _should_convert_to_gguf(module, state_dict, module_prefix):
```
sayakpaul, line 34: Should we keep a placeholder to have `modules_to_not_convert`? Could be very important for video models, no?

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def as_tensor(self):
+        return torch.Tensor._make_subclass(torch.Tensor, self, self.requires_grad)
+
+    @classmethod
+    def __torch_function__(cls, func, types, args=(), kwargs=None):
+        if kwargs is None:
+            kwargs = {}
+
+        result = super().__torch_function__(func, types, args, kwargs)
```
sayakpaul, line 342: Do we always have to compute this?

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,384 @@
...
+        compute_dtype=None,
+        device=None,
+    ) -> None:
+        super().__init__(in_features, out_features, bias, device)
+        self.compute_dtype = compute_dtype
+
+    def forward(self, inputs):
+        weight = dequantize_gguf_tensor(self.weight)
+        weight = weight.to(self.compute_dtype)
+        bias = self.bias.to(self.compute_dtype)
```
sayakpaul, line 381: The `bias` doesn't need any dequantization?

sayakpaul (commented):
Looking very nice! I wonder if it would be possible to support saving models with quantization from GGUF. That would be amazing!

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,109 @@
...
+        model: "ModelMixin",
+        param_value: "torch.Tensor",
+        param_name: str,
+        target_device: "torch.device",
+        state_dict: Dict[str, Any],
+        unexpected_keys: Optional[List[str]] = None,
+    ):
+        module, tensor_name = get_module_from_name(model, param_name)
+        if tensor_name not in module._parameters:
+            raise ValueError(f"{module} does not have a parameter or a buffer named {tensor_name}.")
```
SunMarc, line 86: You should check for buffers also in `module._buffers` as it is not included in `module._parameters`

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,384 @@
...
+    scales = scales.reshape((n_blocks, 3, 4))
+
+    d, m, m_d = torch.split(scales, scales.shape[-2] // 3, dim=-2)
+
+    sc = torch.cat([d & 0x3F, (m_d & 0x0F) | ((d >> 2) & 0x30)], dim=-1)
+    min = torch.cat([m & 0x3F, (m_d >> 4) | ((m >> 2) & 0x30)], dim=-1)
+
+    return (sc.reshape((n_blocks, 8)), min.reshape((n_blocks, 8)))
+
+
```
SunMarc, line 80: gguf library have a dequantize function, can't be make use of it ? 

`def dequantize(data: np.ndarray, qtype: GGMLQuantizationType) -> np.ndarray:`

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,384 @@
...
+    return dequant.as_tensor()
+
+
+class GGUFParameter(torch.Tensor):
+    def __new__(cls, data, requires_grad=False, quant_type=None):
+        data = data if data is not None else torch.empty(0)
+        self = torch.Tensor._make_subclass(cls, data, requires_grad)
+        self.quant_type = quant_type
+
+        return self
```
SunMarc, line 332: Nice ! Great use of tensor subclasses

src/diffusers/loaders/single_file_model.py
```diff
@@ -215,6 +218,12 @@ def from_single_file(cls, pretrained_model_link_or_path_or_dict: Optional[str] =
                 local_files_only=local_files_only,
                 revision=revision,
             )
+        if quantization_config is not None:
+            hf_quantizer = DiffusersAutoQuantizer.from_config(quantization_config)
+            hf_quantizer.validate_environment()
+
+        else:
+            hf_quantizer = None
```
SunMarc, line 226: For GGUF files, I'm thinking if it would be nice to allow the user to load the model without having necessarily to specify `quantization_config=GGUFQuantizationConfig(compute_dtype=xxx)`. If we detect that this is a gguf, we can set by default `quantization_config = GGUFQuantizationConfig(compute_dtype=torch.float32)`. 
I'm suggesting this because usually, when you pass a `quantization_config`, it means either that the model is not quantized (bnb) or that the model is quantized (there is a quantization_config in the config.json) but we want to change a few arguments. 

Also, what happens when the user pass a gguf without specifying the `quantization_config` ?

SunMarc (commented):
Thanks for this nice implementation. I think transformers integration could benefit from what you did here to allow users to run the compressed model cc @Isotr0py. Left a few comments.

src/diffusers/quantizers/gguf/utils.py
sayakpaul: Great to see you again on `diffusers`, @Isotr0py!

> BTW, if we want a runtime optimized dequantize function, we might need a torch compatible GGUF kernel, just like what we do in vLLM...

That is a great point. Perhaps we could ship the first iteration without and make use of those kernels (iff possible) in a follow-up PR? WDYT?

src/diffusers/loaders/single_file_model.py
sayakpaul: Yeah this is a good point! I think for most users, the entrypoint for GGUF files is going to be through `from_single_file()` and I agree with the logic you mentioned.

src/diffusers/quantizers/gguf/utils.py
Isotr0py: Agreed, we can port the GGUF kernel in a follow-up PR later, because I'm going to update the GGUF kernel implementation in vLLM since it's pretty out of date. (Although I haven't had bandwidth to start it. 😅)

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -0,0 +1,384 @@
...
+def _quant_shape_from_byte_shape(shape, type_size, block_size):
+    return (*shape[:-1], shape[-1] // type_size * block_size)
+
+
+def dequantize_gguf_tensor(tensor):
+    if not hasattr(tensor, "quant_type"):
+        return tensor
+
+    quant_type = tensor.quant_type
+    dequant_fn = dequantize_functions[quant_type]
```
sayakpaul, line 310: Beautiful!

src/diffusers/quantizers/gguf/utils.py
Isotr0py: Yes, in GGUF file, only weights are quantized while all bias is still saved as FP32.

src/diffusers/quantizers/gguf/utils.py
Isotr0py, line 332: Just curious, can we inherit from `nn.Parameter`? Because here is "Parameter" and `nn.Parameter` is also a subclass of `torch.Tensor`.

asomoza:
I did some tests with this PR and I found that it was really slow at loading, so I tested just loading the transformer model like this:

```python
import torch

from diffusers import FluxTransformer2DModel, GGUFQuantizationConfig


ckpt_path = "https://huggingface.co/city96/FLUX.1-dev-gguf/blob/main/flux1-dev-Q2_K.gguf"
transformer = FluxTransformer2DModel.from_single_file(
    ckpt_path,
    torch_dtype=torch.bfloat16,
).to("cuda")
```

I have 32GB of RAM and 16 of VRAM for this test, and it's really slow to load the model and uses all the RAM without using the VRAM and after a while becomes irresponsive so it's also swapping to disk which is a lot of RAM usage for a 4 GB model.

Tested the same model with ComfyUI and it loads faster and doesn't use any RAM at all. 

Inference speed was faster with diffusers though, 1:05 mins versus 1:35

sayakpaul:
@asomoza thanks! What happens when we use FP16? Does the ComfyUI node also use BF16? Asking to rule out if there's any incompatible dtype issues.

Additionally, if you could maybe help to localize the segment in the code that is taking time during loading, that would be very helpful!

src/diffusers/models/model_loading_utils.py
```diff
@@ -396,3 +401,77 @@ def _fetch_index_file_legacy(
...
+    for tensor in tqdm(reader.tensors, desc="Loading GGUF Parameters: "):
+        name = tensor.name
+        quant_type = tensor.tensor_type
+
+        # if the tensor is a torch supported dtype do not use GGUFParameter
+        is_gguf_quant = quant_type not in [gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16]
+        weights = torch.from_numpy(tensor.data.copy())
+        parsed_parameters[name] = GGUFParameter(weights, quant_type=quant_type) if is_gguf_quant else weights
+
+    if len(reader_keys) > 0:
```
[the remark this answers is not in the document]
DN6: Copied this from transformers. But these aren't tensor keys. They're metadata keys
```
['GGUF.version', 'GGUF.tensor_count', 'GGUF.kv_count', 'general.architecture', 'general.quantization_version', 'general.file_type']
```
This can probably just be removed since the info isn't too relevant.

src/diffusers/loaders/single_file_model.py
DN6: I agree that this is a nice convenience. GGUF does have all the information we need to auto fetch the config (honestly it's possible to skip the config all together), but it would mean that loading semantics would be different for GGUF vs other quant types. e.g.

GGUF 
```python
model = FluxTransformer2DModel.from_single_file("<>.gguf")
```

BnB and TorchAO (assuming these can be supported):  

```python
model = FluxTransformer2DModel.from_single_file("<path>", quantization_config=BnBConfig)
model = FluxTransformer2DModel.from_single_file("<path>", quantization_config=TorchAOConfig)
```

When loading with `from_pretrained` we would have to pass a quant config. I understand it's not ideal, but I feel it's better to preserve consistency across the different quant loading methods. 

@SunMarc if the config isn't passed you get shape mismatch errors when you hit `load_model_dict_into_meta` since the quant shapes are different from the expected shapes.

DN6:
@asomoza I profiled the RAM usage for the Q2K checkpoint and it seems okay. 
```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    14    650.0 MiB    650.0 MiB           1   @profile
    15                                         def load_model():
    16    650.0 MiB      0.0 MiB           1       ckpt_path = (
    17    650.0 MiB      0.0 MiB           1           "https://huggingface.co/city96/FLUX.1-dev-gguf/blob/main/flux1-dev-Q2_K.gguf"
    18                                             )
    19   4519.5 MiB   3869.5 MiB           2       transformer = FluxTransformer2DModel.from_single_file(
    20    650.0 MiB      0.0 MiB           1           ckpt_path,
    21    650.0 MiB      0.0 MiB           1           quantization_config=GGUFQuantizationConfig(compute_dtype=torch.bfloat16),
    22    650.0 MiB      0.0 MiB           1           torch_dtype=torch.bfloat16,
    23                                             )

```

There is a slight lag in loading compared to `from_pretrained`. I'm looking into it. Additionally, single file models load to CPU first before moving to GPU (device map isn't supported)

asomoza:
@DN6 I'm using windows (because most webui users do) and I also test with just looking at the RAM consumption, it goes all the way up to 80GB but if I profile the code I get this:

```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
     7    537.4 MiB    537.4 MiB           1   @profile
     8                                         def load_model():
     9    537.4 MiB      0.0 MiB           1       ckpt_path = "https://huggingface.co/city96/FLUX.1-dev-gguf/blob/main/flux1-dev-Q2_K.gguf"
    10  23090.4 MiB  22552.9 MiB           3       transformer = FluxTransformer2DModel.from_single_file(
    11    537.4 MiB      0.0 MiB           1           ckpt_path,
    12    537.4 MiB      0.0 MiB           1           quantization_config=GGUFQuantizationConfig(compute_dtype=torch.bfloat16),
    13    537.4 MiB      0.0 MiB           1           torch_dtype=torch.bfloat16,
    14    173.6 MiB -22916.8 MiB           1       ).to("cuda")
```

This is why I also use normal tools to monitor and not just profile the code. I must add that this doesn't happen with ComfyUI.

I still need to test the proposed tests from @sayakpaul and also just in case test in linux.

This is my env:

```
- 🤗 Diffusers version: 0.32.0.dev0
- Platform: Windows-11-10.0.22631-SP0
- Python version: 3.12.4
- PyTorch version (GPU?): 2.4.1+cu124 (True)
- Jax version: not installed
- Huggingface_hub version: 0.26.1
- Transformers version: 4.45.2
- Accelerate version: 1.0.1
- PEFT version: 0.13.2
- Bitsandbytes version: 0.44.1
- Safetensors version: 0.4.5
- Accelerator: NVIDIA GeForce RTX 4090 Laptop GPU, 16376 MiB
```

I'll update pytorch and the libraries also to see if the problem lies there.

src/diffusers/loaders/single_file_model.py
yiyixuxu: >I'm suggesting this because usually, when you pass a quantization_config, it means either that the model is not quantized (bnb) or that the model is quantized (there is a quantization_config in the config.json) but we want to change a few arguments.

yeah I thought about that too, but I think the API for `from_single_file` and `from_pretrained` might just have to be different. It is a bit confusing but I'm not sure if there is a way to make it consistent between `from_single_file` and `from_pertrained`, if we also want to make sure the same API is consistent across different quant types

GGUF is a special case here because it has built-in config. Normally, for single-file it is just a checkpoint without config, so you will always have to pass a config (at least I think so, is it? @DN6 ).  So for loading a regular quantized model (e.g. BNB) we can load it with `from_pretrained` without passing a config, but for `from_single_file`, we will have to manually pass a config

so agree with @DN6 here I think it more important to make the same API (`from_pretrained API` or `from_single_file`) consistent for different quant types; if we have to choose one

but if there a way to make it consistent between from_pretrained and from_single_file and across all quant types it will be great!

yiyixuxu: also, want to know this: do we plan to support quantizing a model in`from_single_file`? @DN6

sayakpaul: > GGUF is a special case here because it has built-in config. Normally, for single-file it is just a checkpoint without config, so you will always have to pass a config (at least I think so, is it? @DN6 ). So for loading a regular quantized model (e.g. BNB) we can load it with from_pretrained without passing a config, but for from_single_file, we will have to manually pass a config

Would it make sense to at least make the user aware when the passed config and the determined config mismatch and if that could lead to unintentional consequences? 

> also, want to know this: do we plan to support quantizing a model infrom_single_file? @DN6

Supporting quantizing in the GGUF format (regardless of `from_pretrained()` or `from_single_file()`) would be reallllly nice.

DN6: @yiyixuxu Yeah we can definitely support quantizing a model via single file. For GGUF I can look into in a follow up because we would have to port the quantize functions to torch (the gguf library uses numpy).  

I think with torch AO I'm pretty sure it would work just out of the box. 

You would have to save it with `save_pretrained` though since we don't support serializing single file checkpoints.

DN6:
@asomoza Could you try now? Think I figured out the issue. I wasn't creating meta tensors when replacing the Linear layers.

src/diffusers/loaders/single_file_model.py
sayakpaul: So, what I am hearing is saving a GGUF quantized model would be added in a follow-up PR? That is also okay but it could be quite an enabling factor for the community. 

> For GGUF I can look into in a follow up because we would have to port the quantize functions to torch (the gguf library uses numpy). We could use the gguf library interally to quantize but it's quite slow since we would have to move tensors off GPU, convert to numpy and then quantize.

I think the porting option is more preferrable. 

> I think with torch AO I'm pretty sure it would work just out of the box.

You mean serializing with `torchao` but with quantization configs similar to the ones provided in GGUF?

src/diffusers/models/model_loading_utils.py
```diff
@@ -176,11 +182,9 @@ def load_model_dict_into_meta(
     hf_quantizer=None,
     keep_in_fp32_modules=None,
 ) -> List[str]:
-    if hf_quantizer is None:
-        device = device or torch.device("cpu")
+    device = device or torch.device("cpu")
```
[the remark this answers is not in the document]
sayakpaul: @a-r-r-o-w has an open PR for this https://github.com/huggingface/diffusers/pull/10069

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,109 @@
...
+
+logger = logging.get_logger(__name__)
+
+
+class GGUFQuantizer(DiffusersQuantizer):
+    def __init__(self, quantization_config, **kwargs):
+        super().__init__(quantization_config, **kwargs)
+
+        self.compute_dtype = quantization_config.compute_dtype
+        self.pre_quantized = True
```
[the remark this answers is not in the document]
sayakpaul: Hmm seems like it's always going to be prequantized for this PR.

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,109 @@
...
+    ) -> bool:
+        if isinstance(param_value, GGUFParameter):
+            return True
+
+        return False
+
+    def create_quantized_param(
+        self,
+        model: "ModelMixin",
+        param_value: "torch.Tensor",
```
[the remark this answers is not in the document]
sayakpaul: My bad. Not needed here. I think we could just set them to `Optional` then no?

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,109 @@
...
+        param_name: str,
+        target_device: "torch.device",
+        state_dict: Dict[str, Any],
+        unexpected_keys: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
sayakpaul: My bad. Not needed here. I think we could just set them to `Optional` then no?

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,132 @@
...
+        param_name: str,
+        state_dict: Dict[str, Any],
+        **kwargs,
+    ) -> bool:
+        if isinstance(param_value, GGUFParameter):
+            return True
+
+        return False
+
+    def create_quantized_param(
```
sayakpaul, line 95: For bitsandbytes, the bias is often left untouched and is kept as the original `dtype` (`torch.bfloat16`) for example. Does the same apply to GGUF or we can quantize everything?

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,132 @@
...
+        if tensor_name in module._parameters:
+            module._parameters[tensor_name] = param_value.to(target_device)
+        if tensor_name in module._buffers:
+            module._buffers[tensor_name] = param_value.to(target_device)
+
+    def _process_model_before_weight_loading(
+        self,
+        model: "ModelMixin",
+        device_map,
+        keep_in_fp32_modules: List[str] = [],
```
sayakpaul, line 117: But thisn't used any where no? Is that expected?

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,132 @@
...
+        if not is_accelerate_available() or is_accelerate_version("<", "0.26.0"):
+            raise ImportError(
+                "Loading GGUF Parameters requires `accelerate` installed in your enviroment: `pip install 'accelerate>=0.26.0'`"
+            )
+        if not is_gguf_available() or is_gguf_version("<", "0.10.0"):
+            raise ImportError(
+                "To load GGUF format files you must have `gguf` installed in your environment: `pip install gguf>=0.10.0`"
+            )
+
+    def adjust_max_memory(self, max_memory: Dict[str, Union[int, str]]) -> Dict[str, Union[int, str]]:
```
sayakpaul, line 52: Could use a `Copied from ...` comment:
https://github.com/huggingface/diffusers/blob/8421c1461bf4ab7801070d04d6ec1e6b28ee5b59/src/diffusers/quantizers/bitsandbytes/bnb_quantizer.py#L217

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,109 @@
...
+    is_torch_available,
+    logging,
+)
+
+
+if is_torch_available() and is_gguf_available():
+    import gguf
+    import torch
+
+    from .utils import GGUFParameter, _quant_shape_from_byte_shape, _replace_with_gguf_linear
```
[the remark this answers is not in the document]
DN6: When we import `quantizers.quantization_config` in the diffusers init, it also imports the DiffusersAutoQuantizers object which is in the quantizers module init file.  DiffusersAutoQuantizer imports BnB and GGUF quantizers so our 
dependency test complains when running without torch or gguf installed.

src/diffusers/quantizers/gguf/gguf_quantizer.py
city96: The default logic in llama.cpp is [here](https://github.com/ggerganov/llama.cpp/blob/59f4db10883a4f3e855cffbf2c3ab68430e95272/src/llama.cpp#L18922-L18926), the weights quantized with the comfy repo follow this logic, so biases and 1d tensors are left unquantized.

They're FP32 internally in llama.cpp since they're fairly small (and the operation for numpy/gguf FP32 -> torch FP32 doesn't require any dequantization, so doing that might be ever so slightly faster. Same goes for FP16, but *not* BF16 due to not being a native numpy datatype).

sayakpaul: Checking with @DN6 if we're following the same.

tests/quantization/gguf/test_gguf.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+        model.to("cuda")
+        inputs = self.get_dummy_inputs()
+
+        torch.cuda.reset_peak_memory_stats()
+        torch.cuda.empty_cache()
+        with torch.no_grad():
+            model(**inputs)
+        max_memory = torch.cuda.max_memory_allocated()
+        assert (max_memory / 1024**3) < 5
```
sayakpaul, line 92: @DN6 can't we leverage the `get_memory_footprint()` method here?

https://github.com/huggingface/diffusers/blob/98d0cd5778afef0f8361908ed613ebcc285c1581/src/diffusers/models/modeling_utils.py#L1277

tests/quantization/gguf/test_gguf.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_gguf_linear_layers(self):
+        quantization_config = GGUFQuantizationConfig(compute_dtype=self.torch_dtype)
+        model = FluxTransformer2DModel.from_single_file(self.ckpt_path, quantization_config=quantization_config)
+
+        for name, module in model.named_modules():
+            if isinstance(module, torch.nn.Linear) and hasattr(module.weight, "quant_type"):
+                assert module.weight.dtype == torch.uint8
+
+    def test_gguf_memory(self):
```
sayakpaul, line 78: This is checking the memory usage of the model. So, I would do:

```suggestion
    def test_gguf_memory_usage(self):
```

tests/quantization/gguf/test_gguf.py
sayakpaul: This could be similar to:
https://github.com/huggingface/transformers/blob/1ed1de2fecdb5fe82f67e4fe742417f737370324/tests/quantization/bnb/test_4bit.py#L162

tests/quantization/gguf/test_gguf.py
```diff
@@ -0,0 +1,130 @@
...
+            "img_ids": torch.randn((4096, 3), generator=torch.Generator("cpu").manual_seed(0)).to(
+                torch_device, self.torch_dtype
+            ),
+            "txt_ids": torch.randn((512, 3), generator=torch.Generator("cpu").manual_seed(0)).to(
+                torch_device, self.torch_dtype
+            ),
+            "guidance": torch.tensor([3.5]).to(torch_device, self.torch_dtype),
+        }
+
+    def test_gguf_parameters(self):
```
sayakpaul, line 60: Should we check for all the supported quant types here?

tests/quantization/gguf/test_gguf.py
```diff
@@ -0,0 +1,130 @@
...
+
+if is_gguf_available():
+    from diffusers.quantizers.gguf.utils import GGUFParameter
+
+
+@nightly
+@require_big_gpu_with_torch_cuda
+@require_accelerate
+@require_gguf_version_greater_or_equal("0.10.0")
+class GGUFSingleFileTests(unittest.TestCase):
```
sayakpaul, line 25: Maybe also a big model marker?

tests/quantization/gguf/test_gguf.py
```diff
@@ -0,0 +1,130 @@
...
+        with self.assertRaises(ValueError):
+            # Tries with a cast
+            model.float()
+
+        with self.assertRaises(ValueError):
+            # Tries with a cast
+            model.half()
+
+        # This should work
+        model.to("cuda")
```
sayakpaul, line 130: Should there also be a full integration test for models like Flux?

src/diffusers/quantizers/gguf/gguf_quantizer.py
DN6: Biases and 1D tensors are unquantized. @city96 created the GGUF quants for Flux and SD3.* so they're the expert here :)

sayakpaul: Thanks! I don't remember any specific code / tests for checking that.

sayakpaul: Ah seems like this:
https://github.com/huggingface/diffusers/pull/9964/files#diff-1aab1b9f58b9a73465bb88dbd921eeee13cca2913f9cb5808d14c3ca88f5b1b8R389

Maybe having a test for this would be nice!

src/diffusers/quantizers/gguf/utils.py
DN6: It's only used in the recursive replacement. Fine to keep in here for now. We can revisit.

src/diffusers/quantizers/gguf/gguf_quantizer.py
DN6: Updated. It's used now.

src/diffusers/quantizers/gguf/gguf_quantizer.py
```diff
@@ -0,0 +1,109 @@
...
+        self,
+        model: "ModelMixin",
+        device_map,
+        keep_in_fp32_modules: List[str] = [],
+        **kwargs,
+    ):
+        state_dict = kwargs.get("state_dict", None)
+        _replace_with_gguf_linear(model, self.compute_dtype, state_dict)
+
+    def _process_model_after_weight_loading(self, model: "ModelMixin", **kwargs):
```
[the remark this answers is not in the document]
DN6: These model attributes don't seem to be used anywhere? Why register?

src/diffusers/loaders/single_file_utils.py
```diff
@@ -81,8 +81,14 @@
     "open_clip_sd3": "text_encoders.clip_g.transformer.text_model.embeddings.position_embedding.weight",
     "stable_cascade_stage_b": "down_blocks.1.0.channelwise.0.weight",
     "stable_cascade_stage_c": "clip_txt_mapper.weight",
-    "sd3": "model.diffusion_model.joint_blocks.0.context_block.adaLN_modulation.1.bias",
-    "sd35_large": "model.diffusion_model.joint_blocks.37.x_block.mlp.fc1.weight",
+    "sd3": [
```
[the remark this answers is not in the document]
sayakpaul: Anything special for Flux?

src/diffusers/models/model_loading_utils.py
```diff
@@ -396,3 +401,77 @@ def _fetch_index_file_legacy(
...
+            "https://pytorch.org/ and https://github.com/ggerganov/llama.cpp/tree/master/gguf-py for installation instructions."
+        )
+        raise ImportError("Please install torch and gguf>=0.10.0 to load a GGUF checkpoint in PyTorch.")
+
+    reader = GGUFReader(gguf_checkpoint_path)
+    fields = reader.fields
+    reader_keys = list(fields.keys())
+
+    parsed_parameters = {}
+    for tensor in tqdm(reader.tensors, desc="Loading GGUF Parameters: "):
```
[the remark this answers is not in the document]
sayakpaul: @DN6 don't think it's typical of us to do q TQDM here no?

tests/quantization/gguf/test_gguf.py
```diff
@@ -0,0 +1,359 @@
...
+        quantization_config = GGUFQuantizationConfig(compute_dtype=self.torch_dtype)
+        model = self.model_cls.from_single_file(self.ckpt_path, quantization_config=quantization_config)
+
+        for name, module in model.named_modules():
+            if isinstance(module, torch.nn.Linear):
+                if name in model._keep_in_fp32_modules:
+                    assert module.weight.dtype == torch.float32
+        self.model_cls._keep_in_fp32_modules = _keep_in_fp32_modules
+
+    def test_dtype_assignment(self):
```
sayakpaul, line 91:
```suggestion
    def test_device_dtype_assignment(self):
```

src/diffusers/quantizers/gguf/gguf_quantizer.py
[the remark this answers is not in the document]
sayakpaul: For consistency. No big deal.

tests/quantization/gguf/test_gguf.py
```diff
@@ -0,0 +1,359 @@
...
+                assert hasattr(param, "quant_type")
+                assert param.dtype == quant_storage_type
+
+    def test_gguf_linear_layers(self):
+        quantization_config = GGUFQuantizationConfig(compute_dtype=self.torch_dtype)
+        model = self.model_cls.from_single_file(self.ckpt_path, quantization_config=quantization_config)
+
+        for name, module in model.named_modules():
+            if isinstance(module, torch.nn.Linear) and hasattr(module.weight, "quant_type"):
+                assert module.weight.dtype == torch.uint8
```
sayakpaul, line 55: should this test also check if the bias and 1D params are in the respective dtypes?

DN6: 1D Params don't need to be tested. If they're normal FP32 parameters they're loaded as is. They won't be GGUFLinear layers of GGUFParameters.

docs/source/en/quantization/gguf.md
```diff
@@ -13,10 +13,19 @@ specific language governing permissions and limitations under the License.
...
 
+Before starting please install gguf in your environment
+
+```shell
+pip install -U gguf
+```
+
+Since GGUF is a single file format, we will be using `from_single_file` to load the model and pass in the `GGUFQuantizationConfig` when loading the model.
+
+When using GGUF checkpoints, the quantized weights remain in a low memory `dtype`, typically `torch.unint8` and are dynamically dequantized and cast to the configured `compute_dtype` when running a forward pass through each module in the model. The `GGUFQuantizationConfig` allows you to set the `compute_dtype` for the forward pass of each module. The functions used for dynamic dequantizatation are based on the great work done by [city96](https://github.com/city96/ComfyUI-GGUF)
```
city96, line 28: A lot of the pytorch dequantization code is based on the [numpy code from llama.cpp](https://github.com/ggerganov/llama.cpp/blob/master/gguf-py/gguf/quants.py) written by @compilade - I believe he should be credited here as well :)

sayakpaul:
@stevhliu could you review the doc related changes please?

docs/source/en/quantization/gguf.md
```diff
@@ -0,0 +1,68 @@
...
+
+The following example will load the [FLUX.1 DEV](https://huggingface.co/black-forest-labs/FLUX.1-dev) transformer model using the GGUF Q2_K quantization variant.
+
+Before starting please install gguf in your environment
+
+```shell
+pip install -U gguf
+```
+
+Since GGUF is a single file format, we will be using `from_single_file` to load the model and pass in the `GGUFQuantizationConfig` when loading the model.
```
stevhliu, line 26:
```suggestion
Since GGUF is a single file format, use [`~FromSingleFileMixin.from_single_file`] to load the model and pass in the [`GGUFQuantizationConfig`] when loading the model.
```

stevhliu (approved):
LGTM, thanks!

sayakpaul:
@Isotr0py, now that this PR is merged, would you like to help us integrate the support for launching appropriate GGUF kernels as done in vLLM?

sayakpaul:
Wonderful. I think we could start minimal, gauge the performance gains and then iteratively ship. Happy to work with you on this!

PR #8768, opened 2024-07-02
[Core] introduce `controlnet` module

sayakpaul:
# What does this PR do?

We have several ControlNet models now and I think it makes sense to now have a separate `controlnets` module under `src/diffisers/models`.

PR #9968, opened 2024-11-20
fix controlnet module refactor 

yiyixuxu:
https://github.com/huggingface/diffusers/pull/8768 breaks the legacy API for all controlnets it moved: flux, sd3, sdxl, sd, sparse; 
e.g. docstring example for flux controlnet won't run on `main` https://huggingface.co/InstantX/FLUX.1-dev-Controlnet-Canny

this PR fixes that


src/diffusers/models/controlnet.py
```diff
@@ -24,19 +25,91 @@
...
     def __init__(self, *args, **kwargs):
         deprecation_message = "Importing `ControlNetOutput` from `diffusers.models.controlnet` is deprecated and this will be removed in a future version. Please use `from diffusers.models.controlnets.controlnet import ControlNetOutput`, instead."
-        deprecate("ControlNetOutput", "0.34", deprecation_message)
+        deprecate("diffusers.models.controlnet.ControlNetOutput", "0.34", deprecation_message)
         super().__init__(*args, **kwargs)
 
 
 class ControlNetModel(ControlNetModel):
-    def __init__(self, *args, **kwargs):
+    def __init__(
```
yiyixuxu, line 33: for models that inherits from `ConfigMixin`, it relies on its `__init__` method signature to identify expected keys and only pass the expected keys to the model https://github.com/huggingface/diffusers/blob/99c0483b67427de467f11aa35d54678fd36a7ea2/src/diffusers/configuration_utils.py#L465

Currently with its signature being `*args` and `**kwargs`, None of the configs are being passed to the `ControlNetModel` when importing from this legacy location.

src/diffusers/models/controlnet.py
```diff
@@ -24,19 +25,91 @@
 class ControlNetOutput(ControlNetOutput):
     def __init__(self, *args, **kwargs):
         deprecation_message = "Importing `ControlNetOutput` from `diffusers.models.controlnet` is deprecated and this will be removed in a future version. Please use `from diffusers.models.controlnets.controlnet import ControlNetOutput`, instead."
-        deprecate("ControlNetOutput", "0.34", deprecation_message)
```
yiyixuxu, line 27: currently, you're getting a deprecation message like this - it is confusing because we are not deprecating `ControlNetModel`; we only deprecate importing from the previous file

```
FutureWarning: `ControlNetModel` is deprecated and will be removed in version 0.34. Importing `ControlNetModel` from `diffusers.models.controlnet` is deprecated and this will be removed in a future version. Please use `from diffusers.models.controlnets.controlnet import ControlNetModel`, instead.
  deprecate("ControlNetModel", "0.34", deprecation_message)
```

src/diffusers/__init__.py
```diff
@@ -107,6 +107,7 @@
             "ModelMixin",
             "MotionAdapter",
             "MultiAdapter",
+            "MultiControlNetModel",
```
yiyixuxu, line 110: add `MultiControlNetModel` to top-level import because every other `MultiControlNetModels` are, e.g, `SD3MultiControlNetModel`, `FluxMultiControlNetModel`

a-r-r-o-w (approved):
Not sure if I've missed something, but the following changes look correctly done:
- Docstring fixes to correctly use `~models` as opposed to `~diffusers` or anything else.
- Init arguments are handled correctly as opposed to args and kwargs because otherwise our ConfigMixin won't work as expected.
- I've matched the default init parameters for all models and they seem to be the same as before #8768, so this looks correct.
- The deprecation messages are no longer ambiguous.
- MultiControlNetModel is now a top-level import accessible directly from `diffusers.models` but was not previously (or after #8768)
- Using `cls.from_config` as opposed to creating the class object directly from config is the preferred way to do things for `from_*` methods (?)

If these were all the intended changes, I think this is good to merge. With this PR, the linked Flux controlnet model works correctly as well, thanks!

PR #9975, opened 2024-11-20
[Tests] skip nan lora tests on PyTorch 2.5.1 CPU.

sayakpaul:
# What does this PR do?

See: https://github.com/huggingface/diffusers/actions/runs/11925826067/job/33238594250?pr=9943#step:7:261. This PR skips the NaN `fuse_lora()` tests when detected PyTorch version is 2.5 or above and the device is CPU. 

It's likely a PyTorch bug as the following passes on PyTorch 2.4.1 CPU but not on PyTorch 2.5.1 CPU. Thanks to Ben for helping to verifying. 

```py
import torch

device = torch.device("cuda:0") if torch.cuda.is_available() else torch.device("cpu")
print(f"Device: {device}, torch version: {torch.__version__}")

with torch.device(device):
    query = torch.randn(2, 4, 16, 8)
    key = torch.randn(2, 4, 16, 8)
    value = torch.randn(2, 4, 16, 8)

        query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False
    )
    assert torch.isfinite(hidden_states).all()

    query += torch.nan
        query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False
    )
    assert torch.isnan(hidden_states).all()
```

See relevant PyTorch thread: https://github.com/pytorch/pytorch/issues/141128.


tests/lora/test_lora_layers_cogvideox.py
```diff
@@ -126,6 +127,10 @@ def get_dummy_inputs(self, with_generator=True):
         return noise, input_ids, pipeline_inputs
 
     @skip_mps
+    @unittest.skipIf(
+        torch.device(torch_device).type == "cpu" and is_torch_version(">=", "2.5"),
+        "Test not supported on PyTorch 2.5 and CPU.",
+    )
```
sayakpaul, line 133: @a-r-r-o-w not important for this PR but any reason this had to be rewritten for Cog?

BenjaminBossan (approved):
Thanks, LGTM. My only concern is that if it gets fixed in a future PyTorch version, it can easily go unnoticed and tests continue to be skipped. That's why I usually prefer `pytest.mark.xfail(..., strict=True)` for such cases.

sayakpaul:
@BenjaminBossan @DN6 how about now?

tests/lora/test_lora_layers_cogvideox.py
```diff
@@ -126,6 +127,7 @@ def get_dummy_inputs(self, with_generator=True):
         return noise, input_ids, pipeline_inputs
 
     @skip_mps
+    @pytest.mark.xfail("Test currently fails on CPU and PyTorch 2.5.1 but not on PyTorch 2.4.1.", strict=True)
```
BenjaminBossan, line 130: I think you need to add the [`condition`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-xfail-ref) that the device is CPU, otherwise the test will fail on GPU, since it's `strict`.

BenjaminBossan (approved):
Thanks, LGTM.

sayakpaul:
@DN6 okay to merge, pinging because of the CI failures across the board.

PR #9981, opened 2024-11-21
Fix prepare latent image ids and vae sample generators for flux

a-r-r-o-w:


yiyixuxu (approved):
thanks!

a-r-r-o-w:
Wow, changing a randomly sampled latent to an actually properly seeded sampled latent breaks the test 😭

yiyixuxu:
@a-r-r-o-w that's expected, no?

src/diffusers/pipelines/flux/pipeline_flux_controlnet.py
```diff
@@ -512,7 +526,7 @@ def prepare_latents(
         shape = (batch_size, num_channels_latents, height, width)
 
         if latents is not None:
-            latent_image_ids = self._prepare_latent_image_ids(batch_size, height, width, device, dtype)
+            latent_image_ids = self._prepare_latent_image_ids(batch_size, height // 2, width // 2, device, dtype)
```
sayakpaul, line 529: I can recollect two PRs here:

* https://github.com/huggingface/diffusers/pull/9711
* https://github.com/huggingface/diffusers/pull/9929

Gently pinging @DN6

a-r-r-o-w: Oh, do you mean I am reverting a previous change that was necessary for something? I have been trying to pass `latents` for testing purposes but it always fails at this point unless the height and width are correctly adjusted -- it makes sense to me when comparing with other pipelines and doing rough calculation with tensor shapes, but LMK if this is wrong

DN6: Not this is fine. We missed scaling the height and width when passing in latents. We do it here
https://github.com/huggingface/diffusers/blob/e564abe292750b7d2eef07f2b49ea2056df391ab/src/diffusers/pipelines/flux/pipeline_flux.py#L528

a-r-r-o-w: Oh okay. Then I'll update the slices and merge once CI is green

a-r-r-o-w:
I think only need to update slice for `test_controlnet_flux.py` despite there being similar changes for other flux pipelines with generator, because `pipeline_flux_controlnet_image_to_image` and `pipeline_flux_controlnet_inpainting` don't seem to have individual tests.

Issue #9966, opened 2024-11-19
Add support for SD 3.5 IP-Adapters


vladmandic:
First IP-Adapter for SD 3.5 just released at <https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter>
with code for the modified pipeline available in the same location.

Ask is to integrate support for SD 3.5 IP-Adapter into standard t2i/i2i/inpaint pipelines

@yiyixuxu @sayakpaul @DN6 @asomoza

sayakpaul:
CC @fabiorigano as well. Since the code is already out I think it might be nice to open this to the community and we help along the way. Maybe we give it a week and if it's not picked up we integrate ourselves. @asomoza meanwhile, maybe you could run some tests as well?

yiyixuxu:
cc @haofanwang too

guiyrt:
I started working on it and have a first draft without modifying the original code, but takes a while to test because SD3.5 is too big for my GPU.

For this PR, is the goal to have the pipeline with no changes as a community pipeline or as an optimized pipeline in src/pipelines? I noticed that it's not inheriting `IPAdapterMixin` loader, for example. This is my first contribution, still learning about repo structure :)

guiyrt:
Is was expanding the integration into SD3 ControlNet Inpainting pipeline, but I don't think we can test it. I can only find the inpainting controlnet for [SD3-Medium](https://huggingface.co/alimama-creative/SD3-Controlnet-Inpainting), and we only have IP-Adapter for [SD3.5-Large](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter). Do you know of any IP-Adapter for SD3-Medium or Inpainting controlnet for SD3.5-Large?

I also noticed that the HF PR with updated keys and safetensors is not merged yet https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/discussions/10, which means that for the diffusers integration to work, we need to provide the commit-id of the proposed change, otherwise we get mismatched keys when loading the IP-Adapter (as someone noticed [here](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/discussions/12)). So, we either kindly ask for someone of the InstantX team to approve the PR, or we need to add runtime conversion for the expected keys.

hlky:
[...]

PR #9987, opened 2024-11-21
[WIP] SD3.5 IP-Adapter Pipeline Integration

guiyrt:
# What does this PR do?

Integrates [IP-Adapter for SD3.5](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter) pipeline, as discussed in #9966.

## Before submitting
- [X] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/diffusers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/diffusers/tree/main/docs#writing-source-documentation).

## Who can review?
@yiyixuxu
@sayakpaul
@DN6 
@asomoza 
@fabiorigano 
@haofanwang


guiyrt:
I have now a more robust integration of [InstantX/SD3.5-Large-IP-Adapter](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter), but there are still some things to clean-up and improve :) A few points that I noted:

- After looking into `IPAdapterMixin`, I thought maybe it would be easier to create a new IP Adapter loader for SD3.5? It handle behavior related different types of `encoder_hid_proj`, which I think don't apply to SD3.5. Also, the old IP Adapter attention processors are not useful for SD3.5. If you agree, I can create a new loader class for SD3.5, which would have the `load_ip_adapter` method that is currently in the pipeline class (as placeholder), `set_ip_adapter_scale` and other relevant methods.
- I used `StableDiffusionXLPipeline` as reference and noticed a difference in image encoding compared with the code from the InstantX team. In the [SDXL pipeline](https://github.com/huggingface/diffusers/blob/6394d905da45236670570ae87803afd5c4cddb07/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L522), when `output_hidden_states` is `True`, torch.zeros_like for `uncond_image_enc` is passed through the image encoder, but in [InstantX/SD3.5-Large-IP-Adapter.pipeline_stable_diffusion_3_ipa.py](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/blob/main/pipeline_stable_diffusion_3_ipa.py#L924) it is not. Which one should it be?
- I changed the default argument of `joint_attn_kwargs` in the SD3.5 pipeline `forward()` to empty dict so that we don't need to check if is None before unpacking. I think it makes it simpler, but let me know if it should be reverted.
- When IP Adapter scale=0, can we just ignore the IP Adapter? Adding zeros to skip image processing means doing image-related compute for nothing, so skipping that would be ideal. I'm not an expert yet on this, so let me know if this makes sense 😄 
- Is it valuable using `.eval()` and `@torch.inference_mode()`? I removed it because I didn't see it often in other pipelines.

A couple things needed to change before merging (I think):
- num_images_per_prompt and batch_size are not considered yet.
- The image_proj resampler is very similar to the existing `IPAdapterPlusImageProjection`, plus timestep embeddings. We could simplify `TimePerceiverResampler` to make use of existing implementations.
- Extending IP Adapters to other SD3.5 pipelines when things are refined.
- Add/update tests.

Also this (I had a few more points, but can share them later):
- Not super into the `feature_extractor`as property name for IP Adapter image processor , but image_processor is already taken by VaeImageProcessor. Maybe ip_image_processor?
- At the moment, only one IP Adapter can be loaded. Perhaps out of scope, but might be useful to use `MultiIPAdapterImageProjection` in the future (Happy to work on that if you find it valuable).

As I mentioned in the issue, as a proud member of the GPU poor community, I can't fit the pipeline in my puny 16GB GPU 😅 
I tested the pipeline for 1 inference step (only took 5hr), for functionality check, but can't verify image generation quality.
If someone could try it or let me know of a way for me to run the pipeline, I would greatly appreciate it! Here is the inference code I used (is there a better way to share this?):

```python
import torch
from PIL import Image

from diffusers.models.transformers import SD3Transformer2DModel
from diffusers.pipelines.stable_diffusion_3.pipeline_stable_diffusion_3 import StableDiffusion3Pipeline
from transformers import SiglipVisionModel, SiglipImageProcessor

model_path = 'stabilityai/stable-diffusion-3.5-large'
image_encoder_path = "google/siglip-so400m-patch14-384"
ip_adapter_path = "InstantX/SD3.5-Large-IP-Adapter"

device = "cuda"
image_name = "image.png"

transformer = SD3Transformer2DModel.from_pretrained(
    model_path, subfolder="transformer", torch_dtype=torch.bfloat16
)

feature_extractor = SiglipImageProcessor.from_pretrained(
    image_encoder_path, torch_dtype=torch.bfloat16
)

image_encoder = SiglipVisionModel.from_pretrained(
    image_encoder_path, torch_dtype=torch.bfloat16
)

pipe = StableDiffusion3Pipeline.from_pretrained(
    model_path,
    transformer=transformer,
    torch_dtype=torch.bfloat16,
    image_encoder=image_encoder
).to(device)

pipe.load_ip_adapter(ip_adapter_path, "ip-adapter.bin")
pipe.set_ip_adapter_scale(0.6)

ref_img = Image.open(image_name).convert('RGB')

# please note that SD3.5 Large is sensitive to highres generation like 1536x1536
image = pipe(
    width=1024,
    height=1024,
    prompt='a cat',
    num_inference_steps=24,
    guidance_scale=5.0,
    generator=torch.Generator(device).manual_seed(42),
    ip_adapter_image=ref_img,
).images[0]

image.save('result.jpg')
```

src/diffusers/models/attention_processor.py
```diff
@@ -18,6 +18,7 @@
 import torch
 import torch.nn.functional as F
 from torch import nn
+from einops import rearrange
```
hlky, line 21: We like to refactor usage of `einops` `rearrange` to avoid the dependency. Let me know if you need assistance with that!

src/diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py
```diff
@@ -691,17 +734,19 @@ def __call__(
         negative_prompt_embeds: Optional[torch.FloatTensor] = None,
         pooled_prompt_embeds: Optional[torch.FloatTensor] = None,
         negative_pooled_prompt_embeds: Optional[torch.FloatTensor] = None,
+        ip_adapter_image: Optional[PipelineImageInput] = None,
+        ip_adapter_image_embeds: Optional[List[torch.Tensor]] = None,
         output_type: Optional[str] = "pil",
         return_dict: bool = True,
-        joint_attention_kwargs: Optional[Dict[str, Any]] = None,
+        joint_attention_kwargs: Dict[str, Any] = {},
```
hlky, line 741: We may change the default for `joint_attention_kwargs` across all pipelines etc but for now we should keep it consistent

src/diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py
```diff
@@ -911,13 +972,30 @@ def __call__(
...
+                        timestep.to(dtype=latents.dtype),
+                        need_temb=True,
+                    )
+
+                    image_prompt_embeds = dict(
+                        ip_hidden_states = ip_hidden_states,
+                        temb = temb
+                    )
+                else:
+                    image_prompt_embeds = {}
```
hlky, line 987: If `self.joint_attention_kwargs` is not None we can `update` it, otherwise we can set the value to `image_prompt_embeds`, then we don't need the `else` path.

src/diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py
```diff
@@ -911,13 +972,30 @@ def __call__(
...
                     hidden_states=latent_model_input,
                     timestep=timestep,
                     encoder_hidden_states=prompt_embeds,
                     pooled_projections=pooled_prompt_embeds,
-                    joint_attention_kwargs=self.joint_attention_kwargs,
                     return_dict=False,
+                    joint_attention_kwargs={
+                        **image_prompt_embeds,
+                        **self.joint_attention_kwargs,                        
+                    }
```
hlky, line 998: and then we can still use `self.joint_attention_kwargs` here and below

src/diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py
hlky: Note we will remove the `None` checks in the future if we end up changing the default

hlky (commented):
Thanks for your contribution @guiyrt

hlky:
@guiyrt Could you share some example outputs here?

guiyrt:
Thanks @hlky for fixing the tests 😃 
I managed to test the pipeline with a big enough GPU, and found there was a bug in loading the IP adapter state dict, but it's fixed now. I ran the pipeline with some images, and took the chance to test with `num_images_per_prompt=4`. Here are some examples, with the same text prompt as in the official examples, 'a cat'. The first two images are from the `assets` folder in [InstantX/SD3.5-Large-IP-Adapter](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/tree/main/assets).
grid

guiyrt:
In SD3.5 architecture, is it still relevant to have layer-specific ip_adapter scale? Right now, each attention processor has its own scale property. If this is still the case, we can have faster execution by checking also in each attn_proc for scale = 0, and cat zeros directly to `hidden_states`. If not relevant anymore, we can have one ip_adapter_scale in the transformer instead.

The only thing left (I think) before adding `SD3IPAdapterMixin` to the other pipelines is checking if we can refactor a bit `TimePerceiverAttention` and`TimePerceiverResampler`. It is very similar to `IPAdapterFaceIDPlusImageProjection`, but I might need some guidance in what might be the best option here. I'm still getting familiar with the internals, and might be overcomplicating things 😅 For example, we might be able to use the existing Attention class, but kv are fused, it's using `scale = 1 / math.sqrt(math.sqrt(self.dim_head))` in the dot product, and I'm not sure how to leverage existing attention processors for that. Also, we might need a new AdaLayerNorm as we need to get `shift_msa, scale_msa, shift_mlp, scale_mlp`, which I didn't find in existing implementations.

I will dive into the `attention.py` and `attention_processors.py` to learn more in the meantime, but let me know if you have an idea of how to improve it 😄

hlky:
Layer-specific `scale` should still have effect but we will need transformer version of these functions to support that in `set_ip_adapter_scale`
https://github.com/huggingface/diffusers/blob/c9e4fab42ca481fe8e0d2456b54ec900fb57730d/src/diffusers/loaders/unet_loader_utils.py#L41

[a87895e](https://github.com/huggingface/diffusers/pull/9987/commits/a87895e94ba97545b09eae35107e9a3103d73fa1) this check should be sufficient for skipping computation when `scale == 0`

As for refactoring TimePerceiverAttention and TimePerceiverResampler, it looks like we can easily base `TimePerceiverResampler` off `IPAdapterPlusImageProjection` and add `time_proj`/`time_embedding`, calling it something like `IPAdapterTimeImageProjection`

https://github.com/huggingface/diffusers/blob/c9e4fab42ca481fe8e0d2456b54ec900fb57730d/src/diffusers/models/embeddings.py#L1968-L2024

However refactoring the layers (`IPAdapterPlusImageProjectionBlock` in `IPAdapterPlusImageProjection`) is a bit more involved and as there's limited checkpoints (1?) available we need to consider whether its worth changing then needing convert to Diffusers script/uploading converted checkpoints, so I'll let @yiyixuxu comment on that.
https://github.com/huggingface/diffusers/blob/a87895e94ba97545b09eae35107e9a3103d73fa1/src/diffusers/models/embeddings.py#L2203-L2215

yiyixuxu:
>However refactoring the layers (IPAdapterPlusImageProjectionBlock in IPAdapterPlusImageProjection) is a bit more involved and as there's limited checkpoints (1?) available we need to consider whether its worth changing then needing convert to Diffusers script/uploading converted checkpoints, so I'll let @yiyixuxu comment on that.

agree here - I think it is ok to keep these layers as they are! (and it will be nice to not having to convert)

src/diffusers/models/embeddings.py
```diff
@@ -2104,6 +2104,98 @@ def forward(self, id_embeds: torch.Tensor) -> torch.Tensor:
...
+    def forward(self, x, latents, timestep_emb):
+        shift_msa, scale_msa, shift_mlp, scale_mlp = self.adaln_proj(self.adaln_silu(timestep_emb))
+
+        x = self.ln0(x)
+        latents = self.ln1(latents) * (1 + scale_msa[:, None]) + shift_msa[:, None]
+        latents = self.attn(x, latents) + latents
+
+        residual = latents
+        latents = self.adaln_norm(latents) * (1 + scale_mlp[:, None]) + shift_mlp[:, None]
+        return self.ff(latents) + residual
```
hlky, line 2151: Have you verified the numerical accuracy of this vs the original?

guiyrt: I'm testing it right now, the former `TimePerceiverAttention` class can be replaced by `Attention` and `FusedAttnProcessor2_0` processor, but that would require doing some conversions in the state_dict. As I have it right now, we need to change these keys:

```python
image_proj_state_dict = {}
for key, value in state_dict["image_proj"].items():
    for idx in range(4):
        key = key.replace(f"layers.{idx}.0.norm1", f"layers.{idx}.ln0")
        key = key.replace(f"layers.{idx}.0.norm2", f"layers.{idx}.ln1")
        key = key.replace(f"layers.{idx}.0.to_q", f"layers.{idx}.attn.to_q")
        key = key.replace(f"layers.{idx}.0.to_kv", f"layers.{idx}.attn.to_kv")
        key = key.replace(f"layers.{idx}.0.to_out", f"layers.{idx}.attn.to_out.0")
        key = key.replace(f"layers.{idx}.1.0", f"layers.{idx}.adaln_norm")
        key = key.replace(f"layers.{idx}.1.1", f"layers.{idx}.ff.net.0.proj")
        key = key.replace(f"layers.{idx}.1.3", f"layers.{idx}.ff.net.2")
        key = key.replace(f"layers.{idx}.2.1", f"layers.{idx}.adaln_proj")
    image_proj_state_dict[key] = value
```

This is to account for the `IPAdapterTimeImageProjectionBlock` class that replaces ModuleList that previously was as:

```python
nn.ModuleList(
  [
      # msa
      TimePerceiverAttention(dim=hidden_dim, dim_head=dim_head, heads=heads),
      # ff
      nn.Sequential(
          nn.LayerNorm(hidden_dim),
          nn.Linear(hidden_dim, ff_inner_dim, bias=False),
          nn.GELU(),
          nn.Linear(ff_inner_dim, hidden_dim, bias=False),
      ),
      # adaLN
      nn.Sequential(
          nn.SiLU(),
          nn.Linear(hidden_dim, ff_inner_dim, bias=True)
      )
  ]
)
```

I think using existing `diffusers` components makes it easier to understand and minimizes the introduction of somewhat duplicated and single-use code, but does come with the downside of adjusting the state_dict. But as you said, there is only one checkpoint (afaik) available right now, so it's also the best time to make this decision.

I saw @yiyixuxu's message after making most of these changes, but it was a good exercise to learn more about diffusers nonetheless 😄 We can either roll back to the original implementation using `TimePerceiverAttention` or test the current version with `Attention` and `FusedAttnProcessor2_0`, as you prefer.

guiyrt: I found that in fact these are not numerically equivalent, and one clear diference is how scale is applied in the dot product. In the PyTorch implementation, using [`F.scaled_dot_product_attention`](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html#torch-nn-functional-scaled-dot-product-attention), we have:

```python
attn_weight = query @ key.transpose(-2, -1) * scale_factor
```

But in the `PerceiverAttention` module, we have:
```python
attn_weight = (query * scale_factor) @ (key * scale_factor).transpose(-2, -1)
```

Not sure why this difference exists, which one should we follow? If it's PyTorch's, we can use `FusedAttnProcessor2_0`. If we go with the `PerceiverAttention` implementation, we can just ignore the attention processor and compute it directly in `forward()` of `IPAdapterTimeImageProjectionBlock`.

I created a script to check output equality from both implementations and played around changing things, and in fact the only difference is the dot product op. I tried debugging with the PyTorch official equivalent implementation, but even that produces a different output from calling `F.scaled_dot_product_attention` (I found that somewhat weird, you can check the final part of the script).

<details>
<summary>Analysis script</summary>

```python
import math
import torch
import torch.nn as nn
from diffusers.models.embeddings import IPAdapterTimeImageProjection, IPAdapterTimeImageProjectionBlock, TimestepEmbedding, Timesteps
from examples.community.pipeline_stable_diffusion_3_ipa import TimeResampler, PerceiverAttention, reshape_tensor


DEVICE = "cuda"
DTYPE = torch.bfloat16

# ip_adapter from https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/resolve/main/ip-adapter.bin?download=true
state_dict = torch.load("ip-adapter.bin", "cpu")

image_proj_state_dict = {}
for key, value in state_dict["image_proj"].items():
    if key.startswith("layers."):
        idx = key.split(".")[1]
        key = key.replace(f"layers.{idx}.0.norm1", f"layers.{idx}.ln0")
        key = key.replace(f"layers.{idx}.0.norm2", f"layers.{idx}.ln1")
        key = key.replace(f"layers.{idx}.0.to_q", f"layers.{idx}.attn.to_q")
        key = key.replace(f"layers.{idx}.0.to_kv", f"layers.{idx}.attn.to_kv")
        key = key.replace(f"layers.{idx}.0.to_out", f"layers.{idx}.attn.to_out.0")
        key = key.replace(f"layers.{idx}.1.0", f"layers.{idx}.adaln_norm")
        key = key.replace(f"layers.{idx}.1.1", f"layers.{idx}.ff.net.0.proj")
        key = key.replace(f"layers.{idx}.1.3", f"layers.{idx}.ff.net.2")
        key = key.replace(f"layers.{idx}.2.1", f"layers.{idx}.adaln_proj")
    image_proj_state_dict[key] = value

# Image projetion parameters
embed_dim = image_proj_state_dict["proj_in.weight"].shape[1]
output_dim = image_proj_state_dict["proj_out.weight"].shape[0]
hidden_dim = image_proj_state_dict["proj_in.weight"].shape[0]
heads = image_proj_state_dict["layers.0.attn.to_q.weight"].shape[0] // 64
num_queries = image_proj_state_dict["latents"].shape[1]
timestep_in_dim = image_proj_state_dict["time_embedding.linear_1.weight"].shape[1]


############################ Full image proj test #############################
print("\n>> Image proj test")
# Image projection
ti_proj = IPAdapterTimeImageProjection(
    embed_dim=embed_dim,
    output_dim=output_dim,
    hidden_dim=hidden_dim,
    dim_head=64,
    heads=heads,
    num_queries=num_queries
).to(DEVICE, DTYPE)

tr = TimeResampler(
    dim=hidden_dim,
    depth=4,
    dim_head=64,
    heads=heads,
    num_queries=num_queries,
    embedding_dim=embed_dim,
    output_dim=output_dim
).to(DEVICE, DTYPE)

# Load weights
ti_proj.load_state_dict(image_proj_state_dict)
tr.load_state_dict(state_dict["image_proj"])

# Random variables
x = torch.rand(2, 279, 1152, device=DEVICE, dtype=DTYPE)
t = torch.rand(2, device=DEVICE, dtype=DTYPE)

latents_ti_proj, timestep_emb_ti_proj = ti_proj(x, t)
latents_tr, timestep_emb_tr = tr(x, t, need_temb=True)

print("image_proj out:", latents_tr.equal(latents_ti_proj))
print("timestep embed:", timestep_emb_tr.equal(timestep_emb_ti_proj))

###############################################################################


################################# Block test ##################################
print("\n>> Block attention test")

# Common projection params
time_proj = Timesteps(timestep_in_dim, True, 0).to(DEVICE, DTYPE)
time_embedding = TimestepEmbedding(timestep_in_dim, hidden_dim, act_fn="silu").to(DEVICE, DTYPE)
latents_params = nn.Parameter(image_proj_state_dict["latents"]).to(DEVICE, DTYPE)
proj_in = nn.Linear(embed_dim, hidden_dim).to(DEVICE, DTYPE)
proj_out = nn.Linear(hidden_dim, output_dim).to(DEVICE, DTYPE)
norm_out = nn.LayerNorm(output_dim).to(DEVICE, DTYPE)

# New implementation
block = IPAdapterTimeImageProjectionBlock(
    hidden_dim=hidden_dim,
    dim_head=64,
    heads=heads
).to(DEVICE, DTYPE)

# Original implementation
adaln_silu = nn.SiLU()
adaln_proj = nn.Linear(hidden_dim, 4 * hidden_dim, bias=True).to(DEVICE, DTYPE)
pa = PerceiverAttention(
    dim=hidden_dim,
    dim_head=64,
    heads=heads
).to(DEVICE, DTYPE)

# Load common params from state_dict
time_embedding.load_state_dict({
    "linear_1.weight": image_proj_state_dict["time_embedding.linear_1.weight"],
    "linear_1.bias": image_proj_state_dict["time_embedding.linear_1.bias"],
    "linear_2.weight": image_proj_state_dict["time_embedding.linear_2.weight"],
    "linear_2.bias": image_proj_state_dict["time_embedding.linear_2.bias"],
})
proj_in.load_state_dict({"weight": image_proj_state_dict["proj_in.weight"], "bias": image_proj_state_dict["proj_in.bias"]})
proj_out.load_state_dict({"weight": image_proj_state_dict["proj_out.weight"], "bias": image_proj_state_dict["proj_out.bias"]})
norm_out.load_state_dict({"weight": image_proj_state_dict["norm_out.weight"], "bias": image_proj_state_dict["norm_out.bias"]})

# Load first proj layer (original)
layer_0_pa_state_dict = {k.replace("layers.0.0.", ""): v for k,v in state_dict["image_proj"].items() if k.startswith("layers.0.0")}
pa.load_state_dict(layer_0_pa_state_dict)
adaln_proj.load_state_dict({"weight": state_dict["image_proj"]["layers.0.2.1.weight"], "bias": state_dict["image_proj"]["layers.0.2.1.bias"]})

# Load first proj layer (new)
layer_0_block_state_dict = {k.replace("layers.0.", ""): v for k,v in image_proj_state_dict.items() if k.startswith("layers.0.")}
block.load_state_dict(layer_0_block_state_dict)

# Random variables
x = torch.rand(2, 279, 1152, device=DEVICE, dtype=DTYPE)
t = torch.rand(2, device=DEVICE, dtype=DTYPE)

# Processing prior to attention
timestep_emb = time_proj(t).to(DEVICE, DTYPE)
timestep_emb = time_embedding(timestep_emb)

latents = latents_params.repeat(x.size(0), 1, 1)

x = proj_in(x)
x = x + timestep_emb[:, None]

x_input = x
latents_input = latents

############## Projection block - Original ##############
shift_msa1, scale_msa1, shift_mlp1, scale_mlp1 = adaln_proj(adaln_silu(timestep_emb)).chunk(4, dim=1)
        
x1_1 = pa.norm1(x_input)
latents1_1 = pa.norm2(latents_input)
latents1_2 = latents1_1 * (1 + scale_msa1.unsqueeze(1)) + shift_msa1.unsqueeze(1)

b1, l1, _ = latents1_2.shape

q1_1 = pa.to_q(latents1_2)
kv_input1 = torch.cat((x1_1, latents1_2), dim=-2)
k1_1, v1_1 = pa.to_kv(kv_input1).chunk(2, dim=-1)

q1_2 = reshape_tensor(q1_1, pa.heads)
k1_2 = reshape_tensor(k1_1, pa.heads)
v1_2 = reshape_tensor(v1_1, pa.heads)

scale1 = 1 / math.sqrt(math.sqrt(pa.dim_head))
weight1_1 = (q1_2 * scale1) @ (k1_2 * scale1).transpose(-2, -1) * scale1 # More stable with f16 than dividing afterwards
weight1_2 = torch.softmax(weight1_1.float(), dim=-1).type(weight1_1.dtype)
latents1_3 = weight1_2 @ v1_2

latents1_4 = latents1_3.permute(0, 2, 1, 3).reshape(b1, l1, -1)
latents1_5 = pa.to_out(latents1_4)

############## Projection block - New ##############
emb = block.adaln_proj(block.adaln_silu(timestep_emb))
shift_msa2, scale_msa2, shift_mlp2, scale_mlp2 = emb.chunk(4, dim=1)

x2_1 = block.ln0(x_input)
latents2_1 = block.ln1(latents_input) * (1 + scale_msa2[:, None]) + shift_msa2[:, None]
kv_input2 = torch.cat((x2_1, latents2_1), dim=-2)

b2, l2, _ = kv_input2.shape

q2_1 = block.attn.to_q(latents2_1)
kv2 = block.attn.to_kv(kv_input2)
split_size = kv2.shape[-1] // 2
k2_1, v2_1 = torch.split(kv2, split_size, dim=-1)

inner_dim = k2_1.shape[-1]
head_dim = inner_dim // block.attn.heads

q2_2 = q2_1.view(b2, -1, block.attn.heads, head_dim).transpose(1, 2)
k2_2 = k2_1.view(b2, -1, block.attn.heads, head_dim).transpose(1, 2)
v2_2 = v2_1.view(b2, -1, block.attn.heads, head_dim).transpose(1, 2)

# the output of sdp = (batch, num_heads, seq_len, head_dim)
latents2_2 = F.scaled_dot_product_attention(
    q2_2, k2_2, v2_2, attn_mask=None, dropout_p=0.0, is_causal=False, scale=block.attn.scale
)

latents2_3 = latents2_2.transpose(1, 2).reshape(b2, -1, block.attn.heads * head_dim)
latents2_4 = latents2_3.to(q2_2.dtype)

# linear proj
latents2_5 = block.attn.to_out[0](latents2_4)
# dropout
latents2_6 = block.attn.to_out[1](latents2_5)

latents2_7 = latents2_6 / block.attn.rescale_output_factor


############## Checks ##############
print("query:", q2_2.equal(q1_2))
print("key:", k2_2.equal(k1_2))
print("value:", v2_2.equal(v1_2))
print("sdpa out:", latents2_2.equal(latents1_3))
print("final output:", latents2_7.equal(latents1_5))

###############################################################################


################################# SDPA tests ##################################
print("\n>> SDPA tests")

# Efficient implementation equivalent to the following
# https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html
def eq_scaled_dot_product_attention(query, key, value, attn_mask=None, dropout_p=0.0,
        is_causal=False, scale=None, enable_gqa=False) -> torch.Tensor:
    L, S = query.size(-2), key.size(-2)
    scale_factor = 1 / math.sqrt(query.size(-1)) if scale is None else scale
    attn_bias = torch.zeros(L, S, dtype=query.dtype, device=query.device)
    if is_causal:
        assert attn_mask is None
        temp_mask = torch.ones(L, S, dtype=torch.bool).tril(diagonal=0)
        attn_bias.masked_fill_(temp_mask.logical_not(), float("-inf"))
        attn_bias.to(query.dtype)

    if attn_mask is not None:
        if attn_mask.dtype == torch.bool:
            attn_bias.masked_fill_(attn_mask.logical_not(), float("-inf"))
        else:
            attn_bias += attn_mask

    if enable_gqa:
        key = key.repeat_interleave(query.size(-3)//key.size(-3), -3)
        value = value.repeat_interleave(query.size(-3)//value.size(-3), -3)

    attn_weight = query @ key.transpose(-2, -1) * scale_factor
    attn_weight += attn_bias
    attn_weight = torch.softmax(attn_weight, dim=-1)
    attn_weight = torch.dropout(attn_weight, dropout_p, train=True)
    return attn_weight @ value

for backend in (
    nn.attention.SDPBackend.FLASH_ATTENTION,
    nn.attention.SDPBackend.EFFICIENT_ATTENTION,
    nn.attention.SDPBackend.MATH
):
    with nn.attention.sdpa_kernel(backend):
        # This should be equal?
        print(
            f"Equality check with backed {backend}:",
            F.scaled_dot_product_attention(
                q2_2, k2_2, v2_2, attn_mask=None, dropout_p=0.0, is_causal=False, scale=1.0
            ).equal(
                eq_scaled_dot_product_attention(
                    q2_2, k2_2, v2_2, attn_mask=None, dropout_p=0.0, is_causal=False, scale=1.0
                )
            )
        )
``` 
</details>

The script outputs the following:
```
>> Image proj test
image_proj out: False
timestep embed: True

>> Block attention test
query: True
key: True
value: True
sdpa out: False
final output: False

>> SDPA tests
Equality check with backed SDPBackend.FLASH_ATTENTION: False
Equality check with backed SDPBackend.EFFICIENT_ATTENTION: False
Equality check with backed SDPBackend.MATH: False
```

hlky: It's ok to just use TimePerceiverAttention

guiyrt: Done, now the outputs are numerically equivalent

```python
import torch
from diffusers.models.embeddings import IPAdapterTimeImageProjection
from examples.community.pipeline_stable_diffusion_3_ipa import TimeResampler


DEVICE = "cuda"
DTYPE = torch.bfloat16

# ip_adapter from https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/resolve/main/ip-adapter.bin?download=true
state_dict = torch.load("ip-adapter.bin", "cpu", weights_only=True)

# Convert image_proj state dict to diffusers
image_proj_state_dict = {}
for key, value in state_dict["image_proj"].items():
    if key.startswith("layers."):
        idx = key.split(".")[1]
        key = key.replace(f"layers.{idx}.0.norm1", f"layers.{idx}.ln0")
        key = key.replace(f"layers.{idx}.0.norm2", f"layers.{idx}.ln1")
        key = key.replace(f"layers.{idx}.0.to_q", f"layers.{idx}.attn.to_q")
        key = key.replace(f"layers.{idx}.0.to_kv", f"layers.{idx}.attn.to_kv")
        key = key.replace(f"layers.{idx}.0.to_out", f"layers.{idx}.attn.to_out.0")
        key = key.replace(f"layers.{idx}.1.0", f"layers.{idx}.adaln_norm")
        key = key.replace(f"layers.{idx}.1.1", f"layers.{idx}.ff.net.0.proj")
        key = key.replace(f"layers.{idx}.1.3", f"layers.{idx}.ff.net.2")
        key = key.replace(f"layers.{idx}.2.1", f"layers.{idx}.adaln_proj")
    image_proj_state_dict[key] = value

# Image projetion parameters
embed_dim = image_proj_state_dict["proj_in.weight"].shape[1]
output_dim = image_proj_state_dict["proj_out.weight"].shape[0]
hidden_dim = image_proj_state_dict["proj_in.weight"].shape[0]
heads = image_proj_state_dict["layers.0.attn.to_q.weight"].shape[0] // 64
num_queries = image_proj_state_dict["latents"].shape[1]
timestep_in_dim = image_proj_state_dict["time_embedding.linear_1.weight"].shape[1]

# Image projection
ti_proj = IPAdapterTimeImageProjection(
    embed_dim=embed_dim,
    output_dim=output_dim,
    hidden_dim=hidden_dim,
    dim_head=64,
    heads=heads,
    num_queries=num_queries
).to(DEVICE, DTYPE)

tr = TimeResampler(
    dim=hidden_dim,
    depth=4,
    dim_head=64,
    heads=heads,
    num_queries=num_queries,
    embedding_dim=embed_dim,
    output_dim=output_dim
).to(DEVICE, DTYPE)

# Load weights
ti_proj.load_state_dict(image_proj_state_dict)
tr.load_state_dict(state_dict["image_proj"])

# Random variables
x = torch.rand(2, 279, 1152, device=DEVICE, dtype=DTYPE)
t = torch.rand(2, device=DEVICE, dtype=DTYPE)

latents_ti_proj, timestep_emb_ti_proj = ti_proj(x, t)
latents_tr, timestep_emb_tr = tr(x, t, need_temb=True)

print("image_proj out:", latents_tr.equal(latents_ti_proj))
print("timestep embed:", timestep_emb_tr.equal(timestep_emb_ti_proj))
```

examples/community/pipeline_stable_diffusion_3_ipa.py
```diff
@@ -0,0 +1,1403 @@
+# Copyright 2024 Stability AI, The HuggingFace Team and The InstantX Team. All rights reserved.
```
hlky, line 1: Do we need this?

src/diffusers/models/attention_processor.py
```diff
@@ -3927,9 +3927,8 @@ def __call__(
             key = attn.norm_k(key)
 
         # the output of sdp = (batch, num_heads, seq_len, head_dim)
-        # TODO: add support for attn.scale when we move to Torch 2.1
         hidden_states = F.scaled_dot_product_attention(
-            query, key, value, attn_mask=attention_mask, dropout_p=0.0, is_causal=False
+            query, key, value, attn_mask=attention_mask, dropout_p=0.0, is_causal=False, scale=attn.scale
```
hlky, line 3931: Not needed now? also not sure we can support that yet anyway

src/diffusers/models/embeddings.py
```diff
@@ -21,7 +21,7 @@
 
 from ..utils import deprecate
 from .activations import FP32SiLU, get_activation
-from .attention_processor import Attention
+from .attention_processor import Attention, FusedAttnProcessor2_0
```
hlky, line 24: Unused?

examples/community/pipeline_stable_diffusion_3_ipa.py
guiyrt: That's the original implementation, we can remove

guiyrt:
I'm doing new tests with the same images to make sure the output is the same as before, will post them here in a bit

guiyrt:
Back from my witch hunt, TLDR;

The output is different compared with the images I posted before at [c78c4fd](https://github.com/huggingface/diffusers/pull/9987/commits/c78c4fd12a474ecdd2295f20f9ac0e8d8fda0f73), but the current version is correct one, at least the one that follows the original implementation. If you recall:

> > In the [SDXL pipeline](https://github.com/huggingface/diffusers/blob/6394d905da45236670570ae87803afd5c4cddb07/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L522), when output_hidden_states is True, torch.zeros_like for uncond_image_enc is passed through the image encoder, but in [InstantX/SD3.5-Large-IP-Adapter.pipeline_stable_diffusion_3_ipa.py](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/blob/main/pipeline_stable_diffusion_3_ipa.py#L924) it is not. Which one should it be?
> 
> Let's follow the original code for `InstantX/SD3.5-Large-IP-Adapter`.

The images I posted before had `single_negative_image_embeds` set as `self.image_encoder(torch.zeros_like(image), output_hidden_states=True).hidden_states[-2]`, but the original implementation (and as done now) sets it as `torch.zeros_like(image_enc_hidden_states)` (changed in [0f6c607](https://github.com/huggingface/diffusers/pull/9987/commits/0f6c6079df30e9469cd8b606c6b6c833a2eba986)).

I'm a bit short on time rn, so I'll make a more detailed post later with images. But seems to check out 😄

guiyrt:
As it turns out, original vs current output was sometimes ever so slightly off. In most examples, just a few details were different, in others no difference, and in anothers the changes were more noticeable. After leaving no stone unturned (and I do mean it), the culprit was a rescaling that squares off the image before passing it to `feature_extractor` (`SiglipImageProcessor`):

```python
clip_image = clip_image.resize((max(clip_image.size), max(clip_image.size)))
```

As we already have the image preparation in `feature_extrator`, I don't think we need it. This would have no effect in square images, but would affect images proportionally to aspect ratio. Below is an example showing the difference in outputs.

<table style="border: none;" cellspacing="0" cellpadding="0" border="0">
  <tr style="border: none;">
    <td width="33%" style="border: none;">
      <div align="center">
        <h3>With resize</h3>
        
      </div>
    </td>
    <td width="33%" style="border: none;">
      <div align="center">
        <h3>No resize</h3>
        
      </div>
    </td>
    <td width="33%" style="border: none;">
      <div align="center">
        <h3>Diff</h3>
        
      </div>
    </td>
  </tr>
</table>

This was the only change that affected the output, so now we have a version that works as intended :tada: We can now focus more on review, documentation and then changing the other pipelines

hlky (approved):
Nice work @guiyrt! Indeed resizing the image should happen in `feature_extractor`. I'm going to approve this on my end and we'll ask @yiyixuxu for another review, and @asomoza can help with further testing.

Here's a couple example PRs for documentation https://github.com/huggingface/diffusers/pull/10021/files  https://github.com/huggingface/diffusers/pull/10131/files files under `docs/`. Ping @stevhliu if you need help with the docs or when they're ready for review.

src/diffusers/models/transformers/transformer_sd3.py
```diff
@@ -330,6 +331,77 @@ def _set_gradient_checkpointing(self, module, value=False):
...
+                key = key.replace(f"layers.{idx}.0.norm1", f"layers.{idx}.ln0")
+                key = key.replace(f"layers.{idx}.0.norm2", f"layers.{idx}.ln1")
+                key = key.replace(f"layers.{idx}.0.to_q", f"layers.{idx}.attn.to_q")
+                key = key.replace(f"layers.{idx}.0.to_kv", f"layers.{idx}.attn.to_kv")
+                key = key.replace(f"layers.{idx}.0.to_out", f"layers.{idx}.attn.to_out.0")
+                key = key.replace(f"layers.{idx}.1.0", f"layers.{idx}.adaln_norm")
+                key = key.replace(f"layers.{idx}.1.1", f"layers.{idx}.ff.net.0.proj")
+                key = key.replace(f"layers.{idx}.1.3", f"layers.{idx}.ff.net.2")
+                key = key.replace(f"layers.{idx}.2.1", f"layers.{idx}.adaln_proj")
+            image_proj_state_dict[key] = value
```
guiyrt, line 380: Should we update the checkpoint [InstantX/SD3.5-Large-IP-Adapter](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter) (convert to safetensors while we're at it?). Otherwise we might need some extra checks here

stevhliu:
> I noticed none of the IP-Adapter related attention processors are part of the API documentation

Feel free to add `IPAdapterJointAttnProcessor2_0`! The other IP-Adapter related attention processors should also be included, and I'll open a separate PR for them

src/diffusers/models/transformers/transformer_sd3.py
yiyixuxu: that's a good idea!

yiyixuxu (approved):
thanks!

src/diffusers/models/transformers/transformer_sd3.py
```diff
@@ -330,6 +331,89 @@ def _set_gradient_checkpointing(self, module, value=False):
         if hasattr(module, "gradient_checkpointing"):
             module.gradient_checkpointing = value
 
+    def _load_ip_adapter_weights(self, state_dict: Dict, low_cpu_mem_usage: bool) -> None:
```
yiyixuxu, line 334: we should probably make a loaderMixin for transformers at some point but can be a future PR (if we need to implement more loading related methods)

src/diffusers/models/transformers/transformer_sd3.py
guiyrt: This should do it, I will upload the checkpoint and propose the change.

```python
from diffusers.utils import _get_model_file
from diffusers.models.modeling_utils import load_state_dict

# Get weights from the hub
model_file = _get_model_file(
    "InstantX/SD3.5-Large-IP-Adapter",
    weights_name="ip-adapter.bin",
)
state_dict = load_state_dict(model_file)

# ip_adapter stays the same
safetensors_dict = {f"ip_adapter.{key}": value for key, value in state_dict["ip_adapter"].items()}

# image_proj needs some adjustments
for key, value in state_dict["image_proj"].items():
    if key.startswith("layers."):
        idx = key.split(".")[1]
        key = key.replace(f"layers.{idx}.0.norm1", f"layers.{idx}.ln0")
        key = key.replace(f"layers.{idx}.0.norm2", f"layers.{idx}.ln1")
        key = key.replace(f"layers.{idx}.0.to_q", f"layers.{idx}.attn.to_q")
        key = key.replace(f"layers.{idx}.0.to_kv", f"layers.{idx}.attn.to_kv")
        key = key.replace(f"layers.{idx}.0.to_out", f"layers.{idx}.attn.to_out.0")
        key = key.replace(f"layers.{idx}.1.0", f"layers.{idx}.adaln_norm")
        key = key.replace(f"layers.{idx}.1.1", f"layers.{idx}.ff.net.0.proj")
        key = key.replace(f"layers.{idx}.1.3", f"layers.{idx}.ff.net.2")
        key = key.replace(f"layers.{idx}.2.1", f"layers.{idx}.adaln_proj")
    safetensors_dict[f"image_proj.{key}"] = value

# export!
save_file(safetensors_dict, "ip_adapter.safetensors")
```

src/diffusers/models/transformers/transformer_sd3.py
guiyrt: Oh I see, similarly to `UNet2DConditionLoadersMixin`? For now we just have this one method, but I can do it now, less things to move around in the future. Should it be in a new file like `diffusers/loaders/transformers.py`?

yiyixuxu: maybe `diffusers/loaders/transformers_sd3.py` because I think flux adapter would needs its own method, no?

guiyrt: Right, and it's also more consistent with the existing code structure as well, thanks for pointing it out :)

src/diffusers/loaders/ip_adapter.py
```diff
@@ -348,3 +351,228 @@ def unload_ip_adapter(self):
...
+                proxies=proxies,
+                local_files_only=local_files_only,
+                token=token,
+                revision=revision,
+                subfolder=subfolder,
+                user_agent=user_agent,
+            )
+            if weight_name.endswith(".safetensors"):
+                state_dict = {"image_proj": {}, "ip_adapter": {}}
+                with safe_open(model_file, framework="pt", device="cpu") as f:
```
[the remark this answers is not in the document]
guiyrt: It is, it processes _safetensors_ checkpoints to mimic the structure of torch pickle checkpoints, as we can't have nested values in safetensors. This was the way it was implemented previously in `IPAdapterMixin`, should we be doing something differently?

src/diffusers/models/transformers/transformer_sd3.py
hlky: Indeed Flux IP Adapter has its own method, we can also rename `IPAdapterJointAttnProcessor2_0` to `SD3IPAdapterJointAttnProcessor2_0`, I've used `FluxIPAdapterJointAttnProcessor2_0` for Flux.

guiyrt: Just updated `IPAdapterJointAttnProcessor2_0` to `SD3IPAdapterJointAttnProcessor2_0`, and created `SD3Transformer2DLoadersMixin`.  Maybe docs could be improved a bit, but it only has `_load_ip_adapter_weights` so far :sweat_smile:

src/diffusers/models/transformers/transformer_sd3.py
guiyrt: PR with updated checkpoints open [here](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/discussions/10), adding safetensors as well.

guiyrt:
In terms of documentation, I think we're only missing updating the SD3 pipeline, I'll spend a bit more time on this one, as it is the one more important to the user. Should/can we add an example with images using IP-Adapter?

hlky:
An example would be great @guiyrt

src/diffusers/loaders/transformers_sd3.py
```diff
@@ -0,0 +1,89 @@
+# Copyright 2024 The HuggingFace Team. All rights reserved.
```
hlky, line 1: Can we rename the file to `transformer_sd3`? Same for the file in docs.

hlky:
Thanks again for the iterations on this @guiyrt. I'll run some examples on this today, should be good to go though.

guiyrt:
Just updated the SD3 pipeline documentation. I added some context, a code example and some generated images. I'm not sure what's the best way to upload images for the docs, right now its pointing to the image I uploded here as a placeholder https://github.com/user-attachments/assets/bc93dd29-ff50-48de-971f-c306653a3a10.

sd3 ip_adapter

docs/source/en/api/pipelines/stable_diffusion/stable_diffusion_3.md
```diff
@@ -59,9 +59,76 @@ image.save("sd3_hello_world.png")
 - [`stabilityai/stable-diffusion-3.5-large`](https://huggingface.co/stabilityai/stable-diffusion-3-5-large)
 - [`stabilityai/stable-diffusion-3.5-large-turbo`](https://huggingface.co/stabilityai/stable-diffusion-3-5-large-turbo)
 
+## Image Prompting with IP-Adapters
+
+An IP-Adapter lets you prompt SD3 with images, in addition to the text prompt. This is especially useful when describing complex concepts that are difficult to articulate through text alone and you have reference images. To load and use an IP-Adapter, you need:
+
+- `image_encoder`: Pre-trained vision model used to obtain image features, usually a CLIP image encoder.
+- `feature_extractor`: Image processor that prepares the input image for the choosen `image_encoder`.
```
stevhliu, line 67:
```suggestion
- `feature_extractor`: Image processor that prepares the input image for the chosen `image_encoder`.
```

docs/source/en/api/pipelines/stable_diffusion/stable_diffusion_3.md
```diff
@@ -59,9 +59,76 @@ image.save("sd3_hello_world.png")
...
+- `ip_adapter_id`: Checkpoint containing parameters of image cross attention layers and image projection. 
+
+IP-Adapters are trained for a specific model architecture, so they also work in finetuned variations of the base model. You can use the `set_ip_adapter_scale()` function to adjust how strongly the output aligns with the image prompt. The higher the value, the more closely the model follows the image prompt. A default value of 0.5 is typically a good balance, ensuring the model considers both the text and image prompts equally.
```
stevhliu, line 70:
```suggestion
IP-Adapters are trained for a specific model architecture, so they also work in finetuned variations of the base model. You can use the [`~IPAdapterMixin.set_ip_adapter_scale`] function to adjust how strongly the output aligns with the image prompt. The higher the value, the more closely the model follows the image prompt. A default value of 0.5 is typically a good balance, ensuring the model considers both the text and image prompts equally.
```

stevhliu:
Thanks for adding more docs! For images, you can open a PR [here](https://huggingface.co/datasets/huggingface/documentation-images/tree/main/diffusers) and I can merge it for you :)

guiyrt:
>  I'll run some examples on this today, should be good to go though.

You can try with this script, should work

```python
import os
import torch
from PIL import Image
from pathlib import Path

from transformers import SiglipVisionModel, SiglipImageProcessor

model_path = "stabilityai/stable-diffusion-3.5-large"
image_encoder_path = "google/siglip-so400m-patch14-384"
ip_adapter_path = "InstantX/SD3.5-Large-IP-Adapter"

# Change to the dir with prompt images
images_dir = Path("base_images")
image_names = os.listdir(images_dir)

results_dir = Path("results")
os.makedirs(results_dir, exist_ok=True)

feature_extractor = SiglipImageProcessor.from_pretrained(
    image_encoder_path, torch_dtype=torch.bfloat16
)

image_encoder = SiglipVisionModel.from_pretrained(
    image_encoder_path, torch_dtype=torch.bfloat16
).to("cuda")

pipe = StableDiffusion3Pipeline.from_pretrained(
    model_path,
    torch_dtype=torch.bfloat16,
    image_encoder=image_encoder,
)

# Load ip-adapter.safetensors checkpoint
# Revision is from PR for checkpoints with updated keys
# https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/discussions/10
pipe.load_ip_adapter(ip_adapter_path, revision="f1f54ca369ae759f9278ae9c87d46def9f133c78")
pipe.set_ip_adapter_scale(0.6)

# Works for 16GB VRAM, but call pipe.to("cuda") if you have big GPU
# enable_model_cpu_offload not working atm, taking a look

for image_name in image_names:
    ref_img = Image.open(images_dir/image_name).convert("RGB")

    # please note that SD3.5 Large is sensitive to highres generation like 1536x1536
    images = pipe(
        width=1024,
        height=1024,
        prompt="a cat",
        num_inference_steps=24,
        num_images_per_prompt=4,
        guidance_scale=5.0,
        ip_adapter_image=ref_img
    ).images

    base_name = os.path.splitext(image_name)[0]
        image.save(results_dir/f"{base_name}_{i}.jpg")
```

guiyrt:
> Thanks for adding more docs! For images, you can open a PR [here](https://huggingface.co/datasets/huggingface/documentation-images/tree/main/diffusers) and I can merge it for you :)

Thanks! Just opened it https://huggingface.co/datasets/huggingface/documentation-images/discussions/404

guiyrt:
I was trying to figure out why `enable_sequential_cpu_offload()` or `enable_model_cpu_offload()` would only work with `_exclude_from_cpu_offload=["image_encoder"]`, and I think the root cause is the PyTorch implementation of `MultiHeadAttention`.  As you can verify in [torch.nn.modules.activation.py](https://github.com/pytorch/pytorch/blob/f9da639950865dee33fc9b2a6e18def7cf2ea8b2/torch/nn/modules/activation.py#L1368C13-L1388C14), the arguments `self.out_proj.weight` and `self.out_proj.bias` are accessed directly, and not in their classes' `forward()`, which means `AlignDevicesHook` doesn't move that data to the GPU (and boom, device mismatch). This happens because unlike `in_proj_weight` and `in_proj_bias`, which are `nn.Parameter`, `out_proj` is of type `NonDynamicallyQuantizableLinear`. Apparently this is to avoid improper quantization [[note in class implementation](https://github.com/pytorch/pytorch/blob/f9da639950865dee33fc9b2a6e18def7cf2ea8b2/torch/nn/modules/linear.py#L131C1-L148C1), pytorch/pytorch#58969] and is supposed to be removed at some point. 

Maybe this was already known before I spiraled into this rabbit hole, but the easy fix for now is to always keep `image_encoder` on the GPU. Maybe there's an easy way to just keep all `NonDynamicallyQuantizableLinear` on GPU, that could be a compromise for a patch, but I'm not sure if CPU offloading is that crucial to go that far.

guiyrt:
Oh, I should mention that this started because I was also breaking the confinement of `SD3Transformer2DModel.forward()` by accessing `self.transformer.image_proj` from `StableDiffusion3Pipeline.__call__`. Now the image embeddings from `image_encoder` are passed to `transformer`, and `image_proj` is called from there.

Let me know if I should change something, as this is changing the signature of `SD3Transformer2DModel.forward()`. I'm a bit unsure if I should change the argument name `ip_adapter_image_embeds` to `ip_adapter_image_hidden_states` for consistency, as all the other embeddings variables have the suffix hidden_states.

hlky:
Would you mind posting the traceback? Maybe there's something we can do but if the issue is in [Siglip](https://github.com/huggingface/transformers/blob/9613933b022ddbf085e2c593ed4ceea4c734179a/src/transformers/models/siglip/modeling_siglip.py#L1113-L1134) we may need to raise it with Transformers team. We probably don't want to keep Siglip on GPU, it's relatively heavy like CLIP Vision right?

Has passing `image_embeds` fixed the issue or no? I think we can still pass it through kwargs instead of changing the signature.

hlky:
Awesome, thanks for checking into this. I think the `enable_sequential_cpu_offload` case is not a blocker, I'll discuss it with the team, I think we'll need to raise it with Transformers team or Accelerate team, let's add a note in the docstring to set `_exclude_from_cpu_offload = ["image_encoder"]` when using `enable_sequential_cpu_offload`, the attribute is available by default on all pipelines so `pipe._exclude_from_cpu_offload = ["image_encoder"]` or `pipe._exclude_from_cpu_offload.append("image_encoder")` should work.

As for `joint_attention_kwargs`, let's add `ip_adapter_image_embeds` from the pipeline pop in `SD3Transformer2DModel` then add `ip_hidden_states`.

hlky:
Sure, let's add that warning, nice idea. Thank you once again for all the iterations on this.

guiyrt:
> Sure, let's add that warning, nice idea. Thank you once again for all the iterations on this.

I enjoyed and learned a lot over the course of this PR, thanks a lot for the guidance @hlky @yiyixuxu @stevhliu :) Not bad for a first PR ahah

Unless we need to update any pipeline tests or you have more change suggestions, I'd say we're golden! We can add IP-Adapters to the rest of the SD3 pipelines, but that could maybe be a different PR, I've seen interest to use it especially with controlnet pipelines (#10129). If that's up for grabs, I'm happy to go for it :)

We should try to merge [Update checkpoints according to diffusers integration](https://huggingface.co/InstantX/SD3.5-Large-IP-Adapter/discussions/10) also once this is merged, so the checkpoints can be used.

yiyixuxu:
I think we can allow the user pass the `_exclude_from_cpu_offload` and `model_cpu_offload_seq` as variables to these offloading methods ( we can do in a separate PR)

src/diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py
```diff
@@ -974,6 +1083,7 @@ def __call__(
                         noise_pred_skip_layers = self.transformer(
                             hidden_states=latent_model_input,
                             timestep=timestep,
+                            ip_adapter_image_embeds=ip_adapter_image_embeds,
```
yiyixuxu, line 1086:
```suggestion
```

PR #9985, opened 2024-11-21
Flux Fill, Canny, Depth, Redux

a-r-r-o-w:
we've opened PR to add diffusers checkpoint to hub, before they are merged, here is how you can test these 4 models

```python
# test all model card examples for flux redux/fill/canny/depth

# flux redux
import torch
from diffusers import FluxPriorReduxPipeline, FluxPipeline
from diffusers.utils import load_image

pipe_prior_redux = FluxPriorReduxPipeline.from_pretrained("black-forest-labs/FLUX.1-Redux-dev", revision="refs/pr/8", torch_dtype=torch.bfloat16).to("cuda")
pipe = FluxPipeline.from_pretrained(
    "black-forest-labs/FLUX.1-dev" , 
    text_encoder=None,
    text_encoder_2=None,
    torch_dtype=torch.bfloat16
).to("cuda")

image = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/robot.png")
pipe_prior_output = pipe_prior_redux(image)
images = pipe(
    guidance_scale=2.5,
    num_inference_steps=50,
    generator=torch.Generator("cpu").manual_seed(0),
    **pipe_prior_output,
).images
images[0].save("flux-dev-redux.png")

del pipe_prior_redux
del pipe
import gc
gc.collect()
torch.cuda.empty_cache()

# flux fill
import torch
from diffusers import FluxFillPipeline
from diffusers.utils import load_image

image = load_image("https://huggingface.co/datasets/diffusers/diffusers-images-docs/resolve/main/cup.png")
mask = load_image("https://huggingface.co/datasets/diffusers/diffusers-images-docs/resolve/main/cup_mask.png")

pipe = FluxFillPipeline.from_pretrained("black-forest-labs/FLUX.1-Fill-dev", torch_dtype=torch.bfloat16, revision="refs/pr/4").to("cuda")
image = pipe(
    prompt="a white paper cup",
    image=image,
    mask_image=mask,
    height=1632,
    width=1232,
    guidance_scale=30,
    num_inference_steps=50,
    max_sequence_length=512,
    generator=torch.Generator("cpu").manual_seed(0)
).images[0]
image.save(f"flux-fill-dev.png")

del pipe
import gc
gc.collect()
torch.cuda.empty_cache()

# flux canny

# please install controlnet-aux if you haven't 
# # !pip install -U controlnet-aux

import torch
from controlnet_aux import CannyDetector
from diffusers import FluxControlPipeline
[...]

PR #9988, opened 2024-11-22
Flux Redux

yiyixuxu:
part of https://github.com/huggingface/diffusers/pull/9985

This PR adds Flux Redux 

# TO-DO

- [x] test if we use all zero prompts will make a difference in output, if not we do not need t5 when redux is used 
- [ ] doc and tests
- [ ] test flux with all other pipelines, including the control + fill model that just released (this will be in a different PR)

I tested empty prompt vs. zero prompt embeds, I think the results are similar; in that case, we can recommend running redux without text_encoders, Here are the results: left is with text_encoders, and the right is without

cc @asomoza here, can you test a little bit and let me know what you think 

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

### to use with t5 (same as in original impl, `prompt=""`)
```python
# test 1
import torch
from PIL import Image

device = "cuda"
dtype = torch.bfloat16

from diffusers import FluxPriorReduxPipeline, FluxPipeline

repo_redux = "YiYiXu/yiyi-redux"
repo_base = "black-forest-labs/FLUX.1-dev"

pipe = FluxPipeline.from_pretrained(repo_base, torch_dtype=torch.bfloat16)
pipe_prior_redux = FluxPriorReduxPipeline.from_pretrained(
    repo_redux, 
    text_encoder=pipe.text_encoder,
    tokenizer=pipe.tokenizer,
    text_encoder_2=pipe.text_encoder_2,
    tokenizer_2=pipe.tokenizer_2,
    torch_dtype=dtype
)
pipe_prior_redux.to(device)

img_path = "/raid/yiyi/flux-new/assets/robot.webp"
image = Image.open(img_path).convert("RGB")

pipe_prior_output = pipe_prior_redux(image)


pipe.enable_model_cpu_offload() #save some VRAM by offloading the model to CPU. Remove this if you have enough GPU power
latents = torch.load("/raid/yiyi/flux-new/redux_latents.pt")
print(latents.shape)
image = pipe(
    guidance_scale=2.5,
    height=768,
    width=1360,
    num_inference_steps=50,
    max_sequence_length=512,
    generator=torch.Generator("cpu").manual_seed(0),
    latents=latents,
    **pipe_prior_output,
).images[0]
image.save("yiyi_test_5_out.png")

# Clean up memory
del pipe
del pipe_prior_redux
import gc
gc.collect()
torch.cuda.empty_cache()
```
### run without t5 (use zero prompt embeds) 
```python
# test 2 (zero prompt embeds)

import torch
from PIL import Image
from diffusers import FluxPriorReduxPipeline, FluxPipeline

device = "cuda"
dtype = torch.bfloat16


repo_redux = "YiYiXu/yiyi-redux"
repo_base = "black-forest-labs/FLUX.1-dev"

pipe_prior_redux = FluxPriorReduxPipeline.from_pretrained(repo_redux, torch_dtype=dtype)
pipe_prior_redux.to(device)

img_path = "/raid/yiyi/flux-new/assets/robot.webp"
image = Image.open(img_path).convert("RGB")
pipe_prior_output = pipe_prior_redux(image)

    repo_base, 
    text_encoder=None,
    tokenizer=None,
    text_encoder_2=None,
    tokenizer_2=None,
    torch_dtype=torch.bfloat16
)
pipe.to(device) #save some VRAM by offloading the model to CPU. Remove this if you have enough GPU power
latents = torch.load("/raid/yiyi/flux-new/redux_latents.pt")
print(latents.shape)
image = pipe(
    guidance_scale=2.5,
    height=768,
    width=1360,
    num_inference_steps=50,
    max_sequence_length=512,
    generator=torch.Generator("cpu").manual_seed(0),
    latents=latents,
    **pipe_prior_output,
).images[0]
image.save("yiyi_test_5_out_2.png")```

### CODE REVIEW

**Comment by wsxwd** [comment]
black-forest-labs/FLUX.1-Redux-dev only support image prompt, right？

**Comment by yiyixuxu** [comment]
@wsxwd yes

**Comment by Adenialzz** [comment]
why only image prompt ? It seems to work with both image prompt and text prompt according to the official example.
image

File: src/diffusers/pipelines/flux/pipeline_flux_prior_redux.py [hunk 1]
```diff
@@ -0,0 +1,370 @@
...
+        image_embeds = image_embeds.to(device=device)
+
+        # 3. Prepare (dummy) text embeddings
+        if hasattr(self, "text_encoder") and self.text_encoder is not None:
+            (
+                prompt_embeds,
+                pooled_prompt_embeds,
+                _,
+            ) = self.encode_prompt(
+                prompt=[""] * batch_size,
```
yiyixuxu, line 348: In the original implementation, even though flux-dev-redux is an image variation pipeline and it does not take a text input, in the code, it still puts an empty string through the text_encoders to create the prompt embeds and then concatenates with the image embeds to create model inputs

However, the model itself does not seem to be controlled by text prompts at all, i.e. if you change the empty prompt to a different prompt, the generation does not seem to be affected by the prompt itself, even though the output might be slightly different.

so here for now, if the text_encoders are explicitly added to the redux pipeline, we use ""; otherwise use a zero prompt embeds. @asomoza , could you test a little bit and let me know if it affects generation? maybe we could remove this logic and just use the zero embeds 

If, in any case,  the prompt embeds generated from empty string is better, we can add prompt/prompt_embeds argument to redux pipeline, this way they can use the prompt_embeds and use it as inputs (so that they don't have to load the text encoders)

I added the scripts for both use cases in PR description

src/diffusers/pipelines/flux/pipeline_flux_prior_redux.py
```diff
@@ -0,0 +1,370 @@
...
+
+if is_torch_xla_available():
+    XLA_AVAILABLE = True
+else:
+    XLA_AVAILABLE = False
+
+
+logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
+
+EXAMPLE_DOC_STRING = """
```
a-r-r-o-w, line 52: Example needs update with correct pipeline usage

a-r-r-o-w (approved):
The changes look great and seem to be working (I haven't tried matching numerically to original but I believe you've covered it)! We can handle tests, docs and example usage doc string in the combined PR. Thanks!

DN6 (approved):
Nice 👍🏽

asomoza:
With more complex images the difference it's also minimal and the image quality it's the same, so with the added benefit of using less VRAM, without the T5 is the way to go if people want to use this.

|source|with TE|without TE|
|---|---|---|
|cat_square|||
|capyrabbit|||
|ip_image|||

the different variations and image sizes works flawlessly too.

asomoza (approved):

src/diffusers/pipelines/flux/pipeline_flux_prior_redux.py
bghira: repeat the prompt a lot and it starts to impact things. the token count of the image space is just very large

yiyixuxu: can you share an example?

lhjlhj11:
> part of #9985
> 
> This PR adds Flux Redux
> # TO-DO
> 
>     * [x]  test if we use all zero prompts will make a difference in output, if not we do not need t5 when redux is used
> 
>     * [ ]  doc and tests
> 
>     * [ ]  test flux with all other pipelines, including the control + fill model that just released (this will be in a different PR)
> 
> 
> I tested empty prompt vs. zero prompt embeds, I think the results are similar; in that case, we can recommend running redux without text_encoders, Here are the results: left is with text_encoders, and the right is without
> 
> cc @asomoza here, can you test a little bit and let me know what you think
>  	
> ### to use with t5 (same as in original impl, `prompt=""`)
> 
> ```python
> # test 1
> import torch
> from PIL import Image
> 
> device = "cuda"
> dtype = torch.bfloat16
> 
> from diffusers import FluxPriorReduxPipeline, FluxPipeline
> from diffusers.utils import load_image
> 
> repo_redux = "YiYiXu/yiyi-redux"
> repo_base = "black-forest-labs/FLUX.1-dev"
> 
> pipe = FluxPipeline.from_pretrained(repo_base, torch_dtype=torch.bfloat16)
> pipe_prior_redux = FluxPriorReduxPipeline.from_pretrained(
>     repo_redux, 
>     text_encoder=pipe.text_encoder,
>     tokenizer=pipe.tokenizer,
>     text_encoder_2=pipe.text_encoder_2,
>     tokenizer_2=pipe.tokenizer_2,
>     torch_dtype=dtype
> )
> pipe_prior_redux.to(device)
> 
> img_path = "/raid/yiyi/flux-new/assets/robot.webp"
> image = Image.open(img_path).convert("RGB")
> 
> pipe_prior_output = pipe_prior_redux(image)
> 
> 
> pipe.enable_model_cpu_offload() #save some VRAM by offloading the model to CPU. Remove this if you have enough GPU power
> latents = torch.load("/raid/yiyi/flux-new/redux_latents.pt")
> print(latents.shape)
> image = pipe(
>     guidance_scale=2.5,
>     height=768,
>     width=1360,
>     num_inference_steps=50,
>     max_sequence_length=512,
>     generator=torch.Generator("cpu").manual_seed(0),
>     latents=latents,
>     **pipe_prior_output,
> ).images[0]
> image.save("yiyi_test_5_out.png")
> 
> # Clean up memory
> del pipe
> del pipe_prior_redux
> import gc
> gc.collect()
> torch.cuda.empty_cache()
> ```
> 
> ### run without t5 (use zero prompt embeds)
> 
> ```python
> # test 2 (zero prompt embeds)
> 
> import torch
> from PIL import Image
> from diffusers import FluxPriorReduxPipeline, FluxPipeline
> 
> device = "cuda"
> dtype = torch.bfloat16
> 
> 
> repo_redux = "YiYiXu/yiyi-redux"
> repo_base = "black-forest-labs/FLUX.1-dev"
> 
> pipe_prior_redux = FluxPriorReduxPipeline.from_pretrained(repo_redux, torch_dtype=dtype)
> pipe_prior_redux.to(device)
> 
> img_path = "/raid/yiyi/flux-new/assets/robot.webp"
> image = Image.open(img_path).convert("RGB")
> pipe_prior_output = pipe_prior_redux(image)
> 
> pipe = FluxPipeline.from_pretrained(
>     repo_base, 
>     text_encoder=None,
>     tokenizer=None,
>     text_encoder_2=None,
>     tokenizer_2=None,
>     torch_dtype=torch.bfloat16
> )
> pipe.to(device) #save some VRAM by offloading the model to CPU. Remove this if you have enough GPU power
> latents = torch.load("/raid/yiyi/flux-new/redux_latents.pt")
> print(latents.shape)
> image = pipe(
>     guidance_scale=2.5,
>     height=768,
>     width=1360,
>     num_inference_steps=50,
>     max_sequence_length=512,
>     generator=torch.Generator("cpu").manual_seed(0),
>     latents=latents,
>     **pipe_prior_output,
> ).images[0]
> image.save("yiyi_test_5_out_2.png")```
> ```

So does the redux support text prompt input now?

bghira:
yes it works with prompts. you need attention bias to upweight the importance of the prompt vs img tokens.

Issue #8955, opened 2024-07-24
RuntimeError: Error(s) in loading state_dict for CLIPTextModelWithProjection


dann2333:
### Describe the bug

Hi, I'm trying to fine-tuning stabilityai/stable-diffusion-3-medium-diffusers and using [the official diffuser scripts](https://github.com/huggingface/diffusers/blob/main/examples/dreambooth/train_dreambooth_lora_sd3.py). The train process was normal expect the loss cannot reduce. I hope to add a validation prompt to see if every thing works ok, so I used Ctrl-C stopped the training process and then added the --validation_prompt and --validation_epochs params. However, when I tried to re-start the train I only found the error below. I tried to use other checkpoints and delete that two params, however no one works.

### Reproduction

Here are the checkpoints link: https://drive.google.com/drive/folders/16RbJa_W4H7aQiGf7QhTXVEJV53LPuS8n?usp=sharing , https://drive.google.com/drive/folders/1zT3LmB7SNtavHP3tgbodTgE13VT0cYvb?usp=sharing
The train command is: ` accelerate launch train_dreambooth_lora_sd3.py --pretrained_model_name_or_path="stabilityai/stable-diffusion-3-medium-diffusers" --output_dir=sd3-lora --instance_data_dir="sample-imgs" --instance_prompt="xxx" --resolution=900 --train_batch_size=1 --train_text_encoder --gradient_accumulation_steps=16 --optimizer="adamw" --learning_rate=1e-6  --text_encoder_lr=1e-6 --lr_scheduler="cosine" --lr_warmup_steps=500 --max_train_steps=4000 --rank=32 --seed="42" --gradient_checkpointing --resume_from_checkpoint latest --center_crop  --report_to="wandb" --checkpointing_steps 20 --checkpoints_total_limit 3 --validation_prompt="xxx" --validation_epochs=1`

### Logs

```shell
07/24/2024 08:16:14 - INFO - __main__ - ***** Running training *****
07/24/2024 08:16:14 - INFO - __main__ -   Num examples = 93
07/24/2024 08:16:14 - INFO - __main__ -   Num batches each epoch = 93
07/24/2024 08:16:14 - INFO - __main__ -   Num Epochs = 667
07/24/2024 08:16:14 - INFO - __main__ -   Instantaneous batch size per device = 1
07/24/2024 08:16:14 - INFO - __main__ -   Total train batch size (w. parallel, distributed & accumulation) = 16
[...]

PR #9993, opened 2024-11-22
[SD3 dreambooth lora] smol fix to checkpoint saving

linoytsaban:
minor change to fix saving of text-encoder layers in lora training script, as we have in place in the dreambooth training script.

related to https://github.com/huggingface/diffusers/issues/8955


examples/dreambooth/train_dreambooth_lora_sd3.py
```diff
@@ -1294,10 +1294,13 @@ def save_model_hook(models, weights, output_dir):
...
-                    text_encoder_one_lora_layers_to_save = get_peft_model_state_dict(model)
-                elif isinstance(model, type(unwrap_model(text_encoder_two))):
-                    text_encoder_two_lora_layers_to_save = get_peft_model_state_dict(model)
+                elif isinstance(model, type(unwrap_model(text_encoder_one))):  # or text_encoder_two
+                    # check hidden size to distinguish between text_encoder_one and two
+                    hidden_size = unwrap_model(model).config.hidden_size
+                    if hidden_size == 768:
+                        text_encoder_one_lora_layers_to_save = get_peft_model_state_dict(model)
+                    elif hidden_size == 1280:
+                        text_encoder_two_lora_layers_to_save = get_peft_model_state_dict(model)
```
sayakpaul, line 1303: Works for me. We cannot distinguish with the classes here because both have the same class. Maybe this reasoning as a comment?

sayakpaul (approved):
Thanks for fixing!

examples/dreambooth/train_dreambooth_lora_sd3.py
linoytsaban: Done :)

PR #10001, opened 2024-11-23
Add `beta`, `exponential` and `karras` sigmas to `FlowMatchEulerDiscreteScheduler`

hlky:
# What does this PR do?

Add `beta`, `exponential` and `karras` sigmas to `FlowMatchEulerDiscreteScheduler`.

```python
from diffusers.schedulers import FlowMatchEulerDiscreteScheduler
import numpy as np


def calculate_shift(
    image_seq_len,
    base_seq_len: int = 256,
    max_seq_len: int = 4096,
    base_shift: float = 0.5,
    max_shift: float = 1.16,
):
    m = (max_shift - base_shift) / (max_seq_len - base_seq_len)
    b = base_shift - m * base_seq_len
    mu = image_seq_len * m + b
    return mu

image_seq_len = 4096
mu = calculate_shift(image_seq_len)
num_inference_steps = 8

flow_match = FlowMatchEulerDiscreteScheduler.from_pretrained("black-forest-labs/FLUX.1-dev", subfolder="scheduler", use_beta_sigmas=True)
flow_match.set_timesteps(sigmas=sigmas, mu=mu)
print(f"flow_match beta (sigmas) {flow_match.sigmas} {flow_match.sigmas.shape[0]}")
print(f"flow_match beta (sigmas) {flow_match.timesteps} {flow_match.timesteps.shape[0]}")

flow_match.set_timesteps(num_inference_steps, mu=mu)
print(f"flow_match beta ({num_inference_steps}) {flow_match.sigmas} {flow_match.sigmas.shape[0]}")
print(f"flow_match beta ({num_inference_steps}) {flow_match.timesteps} {flow_match.timesteps.shape[0]}")

flow_match = FlowMatchEulerDiscreteScheduler.from_pretrained("black-forest-labs/FLUX.1-dev", subfolder="scheduler", use_exponential_sigmas=True)
flow_match.set_timesteps(sigmas=sigmas, mu=mu)
print(f"flow_match exponential (sigmas) {flow_match.sigmas} {flow_match.sigmas.shape[0]}")
print(f"flow_match exponential (sigmas) {flow_match.timesteps} {flow_match.timesteps.shape[0]}")

flow_match.set_timesteps(num_inference_steps, mu=mu)
print(f"flow_match exponential ({num_inference_steps}) {flow_match.sigmas} {flow_match.sigmas.shape[0]}")
print(f"flow_match exponential ({num_inference_steps}) {flow_match.timesteps} {flow_match.timesteps.shape[0]}")

flow_match = FlowMatchEulerDiscreteScheduler.from_pretrained("black-forest-labs/FLUX.1-dev", subfolder="scheduler", use_beta_sigmas=True)
flow_match.set_timesteps(sigmas=sigmas, mu=mu)
print(f"flow_match karras (sigmas) {flow_match.sigmas} {flow_match.sigmas.shape[0]}")
print(f"flow_match karras (sigmas) {flow_match.timesteps} {flow_match.timesteps.shape[0]}")

flow_match.set_timesteps(num_inference_steps, mu=mu)
print(f"flow_match karras ({num_inference_steps}) {flow_match.sigmas} {flow_match.sigmas.shape[0]}")
print(f"flow_match karras ({num_inference_steps}) {flow_match.timesteps} {flow_match.timesteps.shape[0]}")
```
```python
flow_match beta (sigmas) tensor([1.0000, 0.9511, 0.8510, 0.7242, 0.5888, 0.4620, 0.3619, 0.3130, 0.0000]) 9
flow_match beta (sigmas) tensor([1000.0000,  951.1288,  851.0334,  724.2367,  588.8108,  462.0141,
         361.9187,  313.0475]) 8
flow_match beta (8) tensor([1.0000, 0.9291, 0.7838, 0.5998, 0.4033, 0.2193, 0.0741, 0.0032, 0.0000]) 9
flow_match beta (8) tensor([1000.0000,  929.0844,  783.8388,  599.8478,  403.3352,  219.3442,
          74.0985,    3.1830]) 8
flow_match exponential (sigmas) tensor([1.0000, 0.8471, 0.7176, 0.6079, 0.5150, 0.4362, 0.3695, 0.3130, 0.0000]) 9
flow_match exponential (sigmas) tensor([1000.0000,  847.1188,  717.6103,  607.9012,  514.9645,  436.2361,
         369.5438,  313.0475]) 8
flow_match exponential (8) tensor([1.0000, 0.4398, 0.1934, 0.0851, 0.0374, 0.0165, 0.0072, 0.0032, 0.0000]) 9
flow_match exponential (8) tensor([1000.0000,  439.8065,  193.4298,   85.0717,   37.4151,   16.4554,
           7.2372,    3.1830]) 8
flow_match karras (sigmas) tensor([1.0000, 0.9511, 0.8510, 0.7242, 0.5888, 0.4620, 0.3619, 0.3130, 0.0000]) 9
flow_match karras (sigmas) tensor([1000.0000,  951.1288,  851.0334,  724.2367,  588.8108,  462.0141,
         361.9187,  313.0475]) 8
flow_match karras (8) tensor([1.0000, 0.9291, 0.7838, 0.5998, 0.4033, 0.2193, 0.0741, 0.0032, 0.0000]) 9
flow_match karras (8) tensor([1000.0000,  929.0844,  783.8388,  599.8478,  403.3352,  219.3442,
          74.0985,    3.1830]) 8
```

## Who can review?

cc @yiyixuxu @asomoza

### DIFF
--- a/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
+++ b/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
@@ -20,9 +20,12 @@ import numpy as np
 import torch
 
 from ..configuration_utils import ConfigMixin, register_to_config
-from ..utils import BaseOutput, logging
+from ..utils import BaseOutput, is_scipy_available, logging
 from .scheduling_utils import SchedulerMixin
 
+
+if is_scipy_available():
+    import scipy.stats
 
 logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
 
@@ -72,7 +75,16 @@ class FlowMatchEulerDiscreteScheduler(SchedulerMixin, ConfigMixin):
         base_image_seq_len: Optional[int] = 256,
         max_image_seq_len: Optional[int] = 4096,
         invert_sigmas: bool = False,
+        use_karras_sigmas: Optional[bool] = False,
+        use_exponential_sigmas: Optional[bool] = False,
+        use_beta_sigmas: Optional[bool] = False,
     ):
+        if self.config.use_beta_sigmas and not is_scipy_available():
+            raise ImportError("Make sure to install scipy if you want to use beta sigmas.")
+        if sum([self.config.use_beta_sigmas, self.config.use_exponential_sigmas, self.config.use_karras_sigmas]) > 1:
+            raise ValueError(
+                "Only one of `config.use_beta_sigmas`, `config.use_exponential_sigmas`, `config.use_karras_sigmas` can be used."
+            )
         timesteps = np.linspace(1, num_train_timesteps, num_train_timesteps, dtype=np.float32)[::-1].copy()
         timesteps = torch.from_numpy(timesteps).to(dtype=torch.float32)
 
@@ -185,22 +197,32 @@ class FlowMatchEulerDiscreteScheduler(SchedulerMixin, ConfigMixin):
             device (`str` or `torch.device`, *optional*):
                 The device to which the timesteps should be moved to. If `None`, the timesteps are not moved.
         """
-
         if self.config.use_dynamic_shifting and mu is None:
             raise ValueError(" you have a pass a value for `mu` when `use_dynamic_shifting` is set to be `True`")
 
         if sigmas is None:
-            self.num_inference_steps = num_inference_steps
             timesteps = np.linspace(
                 self._sigma_to_t(self.sigma_max), self._sigma_to_t(self.sigma_min), num_inference_steps
             )
 
             sigmas = timesteps / self.config.num_train_timesteps
+        else:
+            num_inference_steps = len(sigmas)
+        self.num_inference_steps = num_inference_steps
 
         if self.config.use_dynamic_shifting:
             sigmas = self.time_shift(mu, 1.0, sigmas)
         else:
             sigmas = self.config.shift * sigmas / (1 + (self.config.shift - 1) * sigmas)
+
+        if self.config.use_karras_sigmas:
+            sigmas = self._convert_to_karras(in_sigmas=sigmas, num_inference_steps=num_inference_steps)
+
+        elif self.config.use_exponential_sigmas:
+            sigmas = self._convert_to_exponential(in_sigmas=sigmas, num_inference_steps=num_inference_steps)
+
+        elif self.config.use_beta_sigmas:
+            sigmas = self._convert_to_beta(in_sigmas=sigmas, num_inference_steps=num_inference_steps)
 
         sigmas = torch.from_numpy(sigmas).to(dtype=torch.float32, device=device)
         timesteps = sigmas * self.config.num_train_timesteps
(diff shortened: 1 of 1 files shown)


src/diffusers/schedulers/scheduling_euler_discrete.py
```diff
@@ -196,13 +196,21 @@ def __init__(
...
         final_sigmas_type: str = "zero",  # can be "zero" or "sigma_min"
+        shift: float = 1.0,
+        use_dynamic_shifting=False,
+        base_shift: Optional[float] = 0.5,
+        max_shift: Optional[float] = 1.15,
+        base_image_seq_len: Optional[int] = 256,
+        max_image_seq_len: Optional[int] = 4096,
+        invert_sigmas: bool = False,
     ):
         if self.config.use_beta_sigmas and not is_scipy_available():
```
yiyixuxu, line 215: this whole section to calculation ` betas` -> `alphas_cumprod` is not relevant to flow matching, no? since it's only used to calculate sigmas when  `not use_flow_match`; but it isn't clear from the code because is it outside of the `if use_flow_match ...  else ...` block

src/diffusers/schedulers/scheduling_euler_discrete.py
```diff
@@ -234,20 +242,39 @@ def __init__(
...
+                # when use_dynamic_shifting is True, we apply the timestep shifting on the fly based on the image resolution
+                sigmas = shift * sigmas / (1 + (shift - 1) * sigmas)
+        else:
+            sigmas = (((1 - self.alphas_cumprod) / self.alphas_cumprod) ** 0.5).flip(0)
+
         # setable values
         self.num_inference_steps = None
 
         # TODO: Support the full EDM scalings for all prediction types and timestep types
         if timestep_type == "continuous" and prediction_type == "v_prediction":
```
yiyixuxu, line 267: is `v_prediction` relevant to flow matching? can people configure `use_flow_match` + `v_prediction`? based on the code, it is possible. But does this make sense?

src/diffusers/schedulers/scheduling_euler_discrete.py
```diff
@@ -362,57 +392,113 @@ def set_timesteps(
...
+        elif sigmas is not None and self.config.use_flow_match:
+            sigmas = np.array(sigmas).astype(np.float32)
+            if self.config.use_flow_match:
+                if self.config.use_dynamic_shifting:
+                    sigmas = self.time_shift(mu, 1.0, sigmas)
+                else:
+                    sigmas = self.config.shift * sigmas / (1 + (self.config.shift - 1) * sigmas)
 
-        else:
+            if self.config.use_karras_sigmas:
```
yiyixuxu, line 429: we have this argument `sigmas` is to accept a custom `sigmas` schedule from user, i.e. they should either pass a custom `sigmas` or choose to use one of the pre-set sigma schedules (e.g. karras, beta, exponential) ; 

so we should not have this logic here
```python
if sigma is not None and self.config.use_flow_match:
    ...
    if self.config.use_karras_sigmas:
          ...
```

yiyixuxu (commented):
thanks for the initiative!! @hlky 

I left some comment as I was going through the the PR; however, now after going through all the code changes, I actually think it becomes pretty clear to me these two things do not naturally combine: they share very little common logic together, and most of the code should be moved inside a big `if use_flow_match: ... else: ...` block if they have not already in there

I think we should:
1. add the karras/beta/exponential to flow matching scheduler
2. in a future PR, we should think after separating the sigmas schedule as its own abstraction, like you suggested to me:)

let me know what you think!

hlky:
Thanks for the review. This was mainly to see whether it could be combined, there are only a few key differences, it should be possible to refactor Euler in a way that FlowMatchEuler only overrides a few things instead. For now I'll change this PR to add karras/beta/exponential to flow match euler and think about that future PR with an abstraction of noise schedulers.

src/diffusers/schedulers/scheduling_euler_discrete.py
hlky: Yes this was due to how pipelines are currently set up and I wanted to test the noise schedules without modifying the pipelines, some of the logic is the same as existing FlowMatchEuler which applies shifting etc to the supplied sigmas, karras etc might actually work better when applied before the scaling so the future refactor will be really useful here.

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
```diff
@@ -185,23 +197,33 @@ def set_timesteps(
...
 
         if sigmas is None:
-            self.num_inference_steps = num_inference_steps
             timesteps = np.linspace(
                 self._sigma_to_t(self.sigma_max), self._sigma_to_t(self.sigma_min), num_inference_steps
             )
 
             sigmas = timesteps / self.config.num_train_timesteps
+        else:
+            num_inference_steps = len(sigmas)
```
hlky, line 210: Just to note passing sigmas to `Euler`  vs `FlowMatchEuler` is different
https://github.com/huggingface/diffusers/blob/074e12358bc17e7dbe111ea4f62f05dbae8a49d5/docs/source/en/using-diffusers/scheduler_features.md?plain=1#L133
https://github.com/huggingface/diffusers/blob/074e12358bc17e7dbe111ea4f62f05dbae8a49d5/src/diffusers/schedulers/scheduling_euler_discrete.py#L367
https://github.com/huggingface/diffusers/blob/074e12358bc17e7dbe111ea4f62f05dbae8a49d5/src/diffusers/pipelines/flux/pipeline_flux.py#L702

yiyixuxu: yes, it is a little bit different (even though I think it should be the same); that's why we created it in the pipeline for convenience (and had planned to move the logic to inside scheduler too, but just haven't done that)

The current logic of `set_timesteps` (for sigma is None path) in FlowMatchEuler is implemented based on SD3( it was the first flow-matching model we have), and it is somehow a little bit different from all the flow-matching models following it; I think the main difference is it applied the shift twice, first time here https://github.com/huggingface/diffusers/blob/074e12358bc17e7dbe111ea4f62f05dbae8a49d5/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L82

and then

https://github.com/huggingface/diffusers/blob/074e12358bc17e7dbe111ea4f62f05dbae8a49d5/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L195

yiyixuxu: BTW, it is very clear the flux method is the "standard" way by now

We had planned to move the flux sigmas to the scheduler, too, but since we want to separate the noise scheduler logic way from the scheduler; maybe we don't have to spend time on that for now because it might be the opposite direction :)

yiyixuxu: oh sorry this here is just explain your question here
> why Flux pipelines pass sigmas and why those are calculated differently to the sigmas is None path 

I think i might put it in the wrong thread here, sorry!

ukaprch:
@hlky @yiyixuxu @asomoza
Before you folks implement this PR I'd like to offer another approach which I used in my (hopefully) soon to be implemented: 
scheduling_flow_match_dpmsolver_multistep.py
which I think is more straightforward, makes more sense and makes use of betas to calculate sigmas as I feel they do make a difference. I will come up with this scheduler hopefully today / tomorrow which you can test for yourselves. The user will only have to make one decision whether to use the betas or not. The betas will be governed by the variables beta_start and beta_end as is currently the case for SDXL.

yiyixuxu:
@hlky is this PR ready for review? can we see some images? :)

@ukaprch thanks! Happy to take a look!

hlky:
Yes it's ready for review

```python
import torch
from diffusers import FluxPipeline, FlowMatchEulerDiscreteScheduler
pipe.enable_vae_tiling()
pipe = pipe.to("cuda")
config = pipe.scheduler.config
euler_flow_beta = FlowMatchEulerDiscreteScheduler.from_config(config, use_beta_sigmas=True)

euler_flow_exponential = FlowMatchEulerDiscreteScheduler.from_config(config, use_exponential_sigmas=True)

euler_flow_karras = FlowMatchEulerDiscreteScheduler.from_config(config, use_karras_sigmas=True)

pipe.scheduler = euler_flow_beta
generator = torch.Generator("cuda").manual_seed(0)
image = pipe(prompt, num_inference_steps=30, guidance_scale=3.5, generator=generator).images[0]
image.save("flow_beta.png")

pipe.scheduler = euler_flow_exponential
generator = torch.Generator("cuda").manual_seed(0)
image = pipe(prompt, num_inference_steps=30, guidance_scale=3.5, generator=generator).images[0]
image.save("flow_exponential.png")

pipe.scheduler = euler_flow_karras
generator = torch.Generator("cuda").manual_seed(0)
image = pipe(prompt, num_inference_steps=30, guidance_scale=3.5, generator=generator).images[0]
image.save("flow_karras.png")
```

beta
flow_beta

exponential
flow_exponential

karras
flow_karras

ukaprch:
The proposed scheduler can handle SD 3.5 Medium/Large & Flux equally well. Working out final details.
Here's a taste of what the proposed FlowMatchEulerDiscrete scheduler can do using **SD 3.5 Medium**:
sigma scheduler = karras
use beta sigmas = True
30 steps
seed = 114747598
prompt = a cat holding a sign that says "hello world"
imgname_0

yiyixuxu (approved):
thanks!

yiyixuxu:
@ukaprch thanks!

I'm going to merge this PR now since it is consistent with the current design. We realize that our current scheduler has many limitations and are very eager to refactor and improve.

I look forward to seeing your PR!! We can discuss it from there. And if we need to change the current design, we will apply it across all schedulers :)

ukaprch:
One of the limitations of FlowMatch for sigma schedules like 'karras', 'exponential' and 'lambdas' is that the current scheme of creating sigmas for them is not biased (weighted) enough in the early steps delaying convergence as opposed to what was used in SDXL. The use of betas is instrumental in building the necessary timesteps / sigmas to take this into account. I felt this feature should not be ignored in our new FlowMatch schedulers. The problem was how to implement for FlowMatch.
The image outputs generated during preliminary testing justify using beta/sigmas as a variable option for Flux and especially for SD 3.5.
Below are (2) images using the same generational data with the exception that 1 uses std flux/karras sigma generation and the other uses beta/sigma generation. 
FLUXKARRASNB
FLUXKARRASUB

PR #10020, opened 2024-11-25
Sd35 controlnet

yiyixuxu:


src/diffusers/models/controlnets/controlnet_sd3.py
```diff
@@ -318,9 +339,27 @@ def forward(
...
+            raise ValueError("hidden_states must be 4D when pos_embed is used")
+
+        # SD3.5 8b controlnet does not have a `pos_embed`,
+        # it use the `pos_embed` from the transformer to process input before passing to controlnet
+        elif self.pos_embed is None and hidden_states.ndim != 3:
+            raise ValueError("hidden_states must be 3D when pos_embed is not used")
+
+        if self.context_embedder is not None and encoder_hidden_states is None:
+            raise ValueError("encoder_hidden_states must be provided when context_embedder is used")
+        # SD3.5 8b controlnet does not have a `context_embedder`, it does not use `encoder_hidden_states
```
yiyixuxu, line 352:
```suggestion
        # SD3.5 8b controlnet does not have a `context_embedder`, it does not use `encoder_hidden_states`
```

src/diffusers/models/controlnets/controlnet_sd3.py
```diff
@@ -58,40 +59,60 @@ def __init__(
...
-                    num_attention_heads=num_attention_heads,
-                    attention_head_dim=self.config.attention_head_dim,
-                    context_pre_only=False,
-                    qk_norm=qk_norm,
-                    use_dual_attention=True if i in dual_attention_layers else False,
-                )
-                for i in range(num_layers)
-            ]
-        )
+        if joint_attention_dim is not None:
```
sayakpaul, line 86: I think this is a good enough condition for now. Because based `joint_attention_dim` we initialize both the `context_embedded` and `transformer_blocks` (that have the `JointTransformerBlock` type). I am okay with it.

src/diffusers/models/controlnets/controlnet_sd3.py
```diff
@@ -318,9 +339,27 @@ def forward(
...
+        # SD3.5 8b controlnet does not have a `pos_embed`,
+        # it use the `pos_embed` from the transformer to process input before passing to controlnet
+        elif self.pos_embed is None and hidden_states.ndim != 3:
+            raise ValueError("hidden_states must be 3D when pos_embed is not used")
+
+        if self.context_embedder is not None and encoder_hidden_states is None:
+            raise ValueError("encoder_hidden_states must be provided when context_embedder is used")
+        # SD3.5 8b controlnet does not have a `context_embedder`, it does not use `encoder_hidden_states`
+        elif self.context_embedder is None and encoder_hidden_states is not None:
+            raise ValueError("encoder_hidden_states should not be provided when context_embedder is not used")
```
sayakpaul, line 354: Very useful!

src/diffusers/models/transformers/transformer_sd3.py
```diff
@@ -18,21 +18,94 @@
...
 from ..modeling_outputs import Transformer2DModelOutput
 
 
 logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
 
 
+@maybe_allow_in_graph
+class SD3SingleTransformerBlock(nn.Module):
+    r"""
+    A Single Transformer block as part of the MMDiT architecture, used in Stable Diffusion 3 ControlNet.
```
sayakpaul, line 46: Perhaps we could make this more explicit by:

* Specifying it's needed for the SD3.5 ControlNet.
* Providing a reference https://stability.ai/news/sd3-5-large-controlnets

src/diffusers/pipelines/controlnet_sd3/pipeline_stable_diffusion_3_controlnet.py
```diff
@@ -858,6 +858,12 @@ def __call__(
         height = height or self.default_sample_size * self.vae_scale_factor
         width = width or self.default_sample_size * self.vae_scale_factor
 
+        controlnet_config = (
+            self.controlnet.config
+            if isinstance(self.controlnet, SD3ControlNetModel)
+            else self.controlnet.nets[0].config
```
sayakpaul, line 864: I guess this is okay for now but could there be a case where we may have incompatibility configs when `len(self.controlnet.nets) > 1`? I guess we will know when we will know.

sayakpaul (approved):
My comments are quite minor and not merge-blockers. 

I guess we could add docs and tests merge merging.

a-r-r-o-w (approved):
I'm not completely familiar with SD3 so all changes seem correct to me so far and inference works well 😅 

LGTM!

yiyixuxu:
I will add tests once the weights PR are merged

vladmandic:
this pr adds support for `from_pretrained` method, but actual stabiliyai models were released as single-file-safetensors and that does not work:

> ValueError: FromOriginalModelMixin is currently only compatible with StableCascadeUNet, UNet2DConditionModel, AutoencoderKL, ControlNetModel, SD3Transformer2DModel, MotionAdapter, SparseControlNetModel, FluxTransformer2DModel

when adding new support, it should be in ORIGINAL format published by authors first, then internally-converted stuff.
right now, this PR adds support for something that does not exist - loading from stabilityai controlnet from diffusers repo  
(or should we really point users to use `diffusers-internal-dev` INSTEAD of originally published models from StabilityAI).

yiyixuxu:
@vladmandic 

we try to support single-file format as best as we can but will not prioritize single-file format over diffusers format. 
The checkpoints PR will be merged soon, for now the script here used https://github.com/huggingface/diffusers/pull/10020#issue-2692654706


>when adding new support, it should be in ORIGINAL format published by authors first, then internally-converted stuff.

vladmandic:
it also fails in current form for any type of offloading - which is kind of a necessity given the model size.

> /home/vlado/dev/sdnext/venv/lib/python3.12/site-packages/diffusers/pipelines/controlnet_sd3/pipeline_stable_diffusion_3_controlnet.py:1047 in __call__

```log
│   1046 │   │   │   │   │   # sd35 (offical) 8b controlnet
│ ❱ 1047 │   │   │   │   │   controlnet_model_input = self.transformer.pos_embed(latent_model_input)
```

RuntimeError: Input type (CUDABFloat16Type) and weight type (CPUBFloat16Type) should be the same

yiyixuxu:
@vladmandic ahh thanks! indeed

yiyixuxu:
@vladmandic ohh, actually, but controlnet was never able to be offloaded because it was used at same time with transformers https://github.com/huggingface/diffusers/blob/fdec8bd6754f8ae5428fb542f08707e0a5aba24e/src/diffusers/pipelines/controlnet/pipeline_controlnet.py#L199

vladmandic:
somehow, i dont have such problems with `InstantX/SD3-Controlnet-Canny` or other InstantX controlnets for SD3.
but it could be due to fact that they are 6x smaller than SAI models.
anyhow, any controlnet for sd35 or flux if pointless if it doesn't support offloading due to sheer model size.

yiyixuxu:
oh make sense, we will look into this

PR #10033, opened 2024-11-27
Fix Mochi Quality Issues

DN6:
# What does this PR do?
We're seeing some quality issues with Mochi due to missing upcasts and differences between how attention is handled in the original repo. 

This PR:

1. Matches the transformer implementation 1:1 so that norms are upcast and run in the same precision as the original repo. 2. Changes the MochiAttnProcessor to match the original approach of dropping padding tokens. 
3. Runs the CFG and Sampling step in FP32     

I'll update the docs PR: #9934 with a guide on how to reproduce the original repo results exactly once this PR is merged. 

Fixes # (issue)

## Before submitting

## Who can review?


src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+from ..normalization import (
+    AdaLayerNormContinuous,
+)
 
 
-logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
+logger = logging.get_logger(__name__)  # pylint: disable=invalid-n
+
+
+class MochiModulatedRMSNorm(nn.Module):
```
a-r-r-o-w, line 37: Love seeing the single file format for modeling!

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
 from ..modeling_outputs import Transformer2DModelOutput
 from ..modeling_utils import ModelMixin
-from ..normalization import AdaLayerNormContinuous, LuminaLayerNormContinuous, MochiRMSNormZero, RMSNorm
+from ..normalization import (
+    AdaLayerNormContinuous,
+)
 
 
-logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
+logger = logging.get_logger(__name__)  # pylint: disable=invalid-n
```
a-r-r-o-w, line 34:
```suggestion
logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
```

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+    def __init__(self, eps: float):
+        super().__init__()
+
+        self.eps = eps
+
+    def forward(self, hidden_states, scale=None):
+        hidden_states_dtype = hidden_states.dtype
+
+        variance = hidden_states.to(torch.float32).pow(2).mean(-1, keepdim=True)
+        hidden_states = hidden_states.to(torch.float32) * torch.rsqrt(variance + self.eps)
```
a-r-r-o-w, line 47: I think we could do one cast for the hidden states instead of two here?

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+        if elementwise_affine:
+            self.weight = nn.Parameter(torch.ones(dim))
+        else:
+            self.weight = None
+
+    def forward(self, hidden_states):
+        hidden_states_dtype = hidden_states.dtype
+
+        variance = hidden_states.to(torch.float32).pow(2).mean(-1, keepdim=True)
+        hidden_states = hidden_states.to(torch.float32) * torch.rsqrt(variance + self.eps)
```
a-r-r-o-w, line 71: Same here, a bit neater to cast hdiden_states before the following statements (not really problem though)

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+        x: torch.Tensor,
+        conditioning_embedding: torch.Tensor,
+    ) -> torch.Tensor:
+        input_dtype = x.dtype
+
+        # convert back to the original dtype in case `conditioning_embedding`` is upcasted to float32 (needed for hunyuanDiT)
+        scale = self.linear_1(self.silu(conditioning_embedding).to(x.dtype))
+        x = self.norm(x, (1 + scale.unsqueeze(1).to(torch.float32)))
+
+        return x.to(input_dtype)
```
a-r-r-o-w, line 110: I think this pattern is very common for some kinds of layers. Do you think we could work on a refactor in the future where we decorate the forward methods with something like `@upcast_to_fp32` so that the type conversions occur outside of the forward and they look like clean mathematical input-to-output mapping? This way, it could also be disabled with a global context manager if upcasting is not required for certain models, but be enabled by default.

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+        emb = self.linear(self.silu(emb))
+        scale_msa, gate_msa, scale_mlp, gate_mlp = emb.chunk(4, dim=1)
+
+        hidden_states = self.norm(hidden_states, (1 + scale_msa[:, None].to(torch.float32)))
+        hidden_states = hidden_states.to(hidden_states_dtype)
+
+        return hidden_states, gate_msa, scale_mlp, gate_mlp
+
+
+class MochiAttention(nn.Module):
```
a-r-r-o-w, line 144: This is interesting to see! I did not know we wanted to move from the central Attention class as well -- only thought we would be breaking up BasicTransformerBlock. Super cool and clean!

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -142,22 +415,25 @@ def forward(
             hidden_states=norm_hidden_states,
             encoder_hidden_states=norm_encoder_hidden_states,
             image_rotary_emb=image_rotary_emb,
+            attention_mask=encoder_attention_mask,
         )
 
-        hidden_states = hidden_states + self.norm2(attn_hidden_states) * torch.tanh(gate_msa).unsqueeze(1)
-        norm_hidden_states = self.norm3(hidden_states) * (1 + scale_mlp.unsqueeze(1))
+        hidden_states = hidden_states + self.norm2(attn_hidden_states, torch.tanh(gate_msa).unsqueeze(1))
+        norm_hidden_states = self.norm3(hidden_states, (1 + scale_mlp.unsqueeze(1).to(torch.float32)))
```
a-r-r-o-w, line 422: I think I'm probably missing something here, but can you point me to where do we handle the downcast for this upcast? Because otherwise all successive computations would be in float32, no?

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+            valid_query = torch.cat([query[idx][None, :], valid_encoder_query], dim=2)
+            valid_key = torch.cat([key[idx][None, :], valid_encoder_key], dim=2)
+            valid_value = torch.cat([value[idx][None, :], valid_encoder_value], dim=2)
+
+            attn_output = F.scaled_dot_product_attention(
+                valid_query, valid_key, valid_value, dropout_p=0.0, is_causal=False
+            )
+            valid_sequence_length = attn_output.size(2)
+            attn_output = F.pad(attn_output, (0, 0, 0, total_length - valid_sequence_length))
+            attn_outputs.append(attn_output)
```
a-r-r-o-w, line 293: Awesome catch and fix!

I think this  might not play well with data parallel implementations due to the loop. i will profile this in the future and we can try a different implementation in the future.

a-r-r-o-w (approved):
Thank you for all the fixes Dhruv!

Very grateful to @YanzuoLu and @Ednaordinary for their continuous help with testing different things out - thank you!

jzhang38:
@DN6 Hey can you give an example code to run your fixed branch? am really curious about the quality changes with those precision fixes.

yiyixuxu:
@DN6 can we get this merged now? :) the tests need to be fixed

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+    def forward(
+        self,
+        x: torch.Tensor,
+        conditioning_embedding: torch.Tensor,
+    ) -> torch.Tensor:
+        input_dtype = x.dtype
+
+        # convert back to the original dtype in case `conditioning_embedding`` is upcasted to float32 (needed for hunyuanDiT)
+        scale = self.linear_1(self.silu(conditioning_embedding).to(x.dtype))
+        x = self.norm(x, (1 + scale.unsqueeze(1).to(torch.float32)))
```
yiyixuxu, line 108: I commented on `MochiModulatedRMSNorm` - we should apply `scale` here, instead of passing it to `MochiModulatedRMSNorm`

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+
+    def forward(
+        self, hidden_states: torch.Tensor, emb: torch.Tensor
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+        hidden_states_dtype = hidden_states.dtype
+
+        emb = self.linear(self.silu(emb))
+        scale_msa, gate_msa, scale_mlp, gate_mlp = emb.chunk(4, dim=1)
+
+        hidden_states = self.norm(hidden_states, (1 + scale_msa[:, None].to(torch.float32)))
```
yiyixuxu, line 138: same comments here, `scale` should be applied here

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -202,7 +478,10 @@ def _get_positions(
         return positions
 
     def _create_rope(self, freqs: torch.Tensor, pos: torch.Tensor) -> torch.Tensor:
-        freqs = torch.einsum("nd,dhf->nhf", pos, freqs.float())
+        with torch.autocast(freqs.device.type, enabled=False):
```
yiyixuxu, line 481: what does this do here? are we disable autocast here in case it is enabled 
ok here if this is what causing the quality issue - but we should come up with something better :)

yiyixuxu (commented):
ok i went over the PR so turns out there are a lot of refactorings (in addition to the fix)
i left some comments, I think it is not ready to merge yet and there are some clean up to be done :)

love that we start to take an action on having model-specific attention/attention class now!  and +1 on @a-r-r-o-w's comments here to think about a better way to manage precision https://github.com/huggingface/diffusers/pull/10033/files#r1860070716

also, do we know what actually cause the issue? all of them together?

src/diffusers/models/transformers/transformer_mochi.py
sayakpaul: For data parallelism, we're replicating the model on each worker, though.

a-r-r-o-w: maybe used wrong terminology. what I meant was parallelizing across batch dimension

src/diffusers/models/transformers/transformer_mochi.py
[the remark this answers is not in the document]
DN6: Agreed. Should I add in this PR or follow up? I don't think Mochi needs those methods since it just has a single processor.

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+        heads: int = 8,
+        dim_head: int = 64,
+        dropout: float = 0.0,
+        bias: bool = False,
+        added_kv_proj_dim: Optional[int] = None,
+        added_proj_bias: Optional[bool] = True,
+        out_dim: int = None,
+        out_context_dim: int = None,
+        out_bias: bool = True,
+        context_pre_only: bool = False,
```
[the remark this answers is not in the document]
DN6: I think these are arguments are all used in the Mochi Attention layer.

src/diffusers/models/transformers/transformer_mochi.py
DN6: Hmm actually I think with all the manual casts we have in place, just setting `torch_dtype=torch.bfloat16` should allow us to effectively run Mochi as if autocast is enabled. 

We could just remove this autocast context manager. Will test.

src/diffusers/models/transformers/transformer_mochi.py
DN6: Downcast here
https://github.com/huggingface/diffusers/blob/b9043256273abcbdf098791a5810639c5622b26b/src/diffusers/models/transformers/transformer_mochi.py#L53

src/diffusers/models/transformers/transformer_mochi.py
DN6: So the reason to pass this in is because of how upcasts are handled in Mochi. The input hidden state is always upcast, but the scaling parameter isn't always upcast. And the scaling/modulation happens between the upcast hidden state and scale parameter.

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+
+        if scale is not None:
+            hidden_states = hidden_states * scale
+
+        hidden_states = hidden_states.to(hidden_states_dtype)
+
+        return hidden_states
+
+
+class MochiRMSNorm(nn.Module):
```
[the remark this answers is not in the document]
yiyixuxu: why cannot we upcast the input instead?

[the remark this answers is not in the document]
yiyixuxu: in fact this operation 
```
hidden_states = hidden_states * torch.rsqrt(variance + self.eps)
```

if the variance is in float32 and `hidden_states` in a lower precision, pytorch will automatically upcast float32 anyway

a little demo script 

```python
import torch

# Create a float16 tensor (simulating hidden_states)
hidden_states = torch.ones(2, 3, dtype=torch.float16)
print("Initial hidden_states dtype:", hidden_states.dtype)  # float16

# Calculate variance in float32 (like the code)
variance = hidden_states.to(torch.float32).pow(2).mean(-1, keepdim=True)
print("Variance dtype:", variance.dtype)  # float32

# Perform the multiplication (this is the line we're testing)
eps = 1e-5
result = hidden_states * torch.rsqrt(variance + eps)
print("Result dtype:", result.dtype)  # float32

# Print all values to verify
print("\nValues:")
print("hidden_states:", hidden_states)
print("variance:", variance)
print("result:", result)
```

```
Initial hidden_states dtype: torch.float16
Variance dtype: torch.float32
Result dtype: torch.float32

Values:
hidden_states: tensor([[1., 1., 1.],
        [1., 1., 1.]], dtype=torch.float16)
variance: tensor([[1.],
        [1.]])
result: tensor([[1.0000, 1.0000, 1.0000],
        [1.0000, 1.0000, 1.0000]])
```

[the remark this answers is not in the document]
yiyixuxu: so I really don't think we need new class here

src/diffusers/models/transformers/transformer_mochi.py
```diff
@@ -17,19 +17,297 @@
...
+
+        self.eps = eps
+
+    def forward(self, hidden_states, scale=None):
+        hidden_states_dtype = hidden_states.dtype
+
+        variance = hidden_states.to(torch.float32).pow(2).mean(-1, keepdim=True)
+        hidden_states = hidden_states.to(torch.float32) * torch.rsqrt(variance + self.eps)
+
+        if scale is not None:
```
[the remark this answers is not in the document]
yiyixuxu: for this comment
>>The way it's used is a bit different in Mochi. The outputs from the linear layer in MochiRMSNormZero (gate_msa, gate_mlp etc) are passed to the modulation after attention.


I don't think it's different, I think this is the case for all our DIT models, maybe with slight difference in implementation from model to mode,l but we always apply the "modulate" part explicitly inside the transformer blocks, we should do the same for Mochi too. see some examples here
flux: https://github.com/huggingface/diffusers/blob/c96bfa5c80eca798d555a79a491043c311d0f608/src/diffusers/models/transformers/transformer_flux.py#L192
sd3:
https://github.com/huggingface/diffusers/blob/c96bfa5c80eca798d555a79a491043c311d0f608/src/diffusers/models/transformers/transformer_sd3.py#L99
cogvideo
https://github.com/huggingface/diffusers/blob/c96bfa5c80eca798d555a79a491043c311d0f608/src/diffusers/models/transformers/transformer_cogview3plus.py#L117


`MochiModulatedRMSNorm` you defined from scratch but it is actually just `MochiRMSNorm`with a scale, and can be write as 
```python
class MochiModulatedRMSNorm(nn.Module):
    def __init__(self, eps: float):
        super().__init__()
        self.norm = MochiRMSNorm(dim=None, eps=eps, elementwise_affine=False)

    def forward(self, hidden_states, scale=None):
        hidden_states = self.norm(hidden_states)
        
        if scale is not None:
            hidden_states = hidden_states * scale

        return hidden_states
```
so for layers that's are currently using `MochiModulatedRMSNorm`, we should easily rewrite with `RMSNorm` too

src/diffusers/models/transformers/transformer_mochi.py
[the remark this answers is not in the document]
yiyixuxu: follow PR is ok! but let's have a rough plan for the follow-up PR before merge
we only have one attention processor for mochi, but these methods are for people to use custom attention processors, also 
methods like `attn_processor` depends on this, and they are useful 
https://github.com/huggingface/diffusers/blob/c96bfa5c80eca798d555a79a491043c311d0f608/src/diffusers/models/transformers/transformer_flux.py#L296

also, let's make sure we don't have something we need relhy with this kind of logic
https://github.com/huggingface/diffusers/blob/c96bfa5c80eca798d555a79a491043c311d0f608/src/diffusers/models/unets/unet_3d_condition.py#L533

src/diffusers/models/transformers/transformer_mochi.py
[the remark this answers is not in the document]
DN6: So in this example the output of the `self.norm` would be `torch.bfloat16`. 

When applying modulation the hidden state has to be in FP32. Just upcasting the output of the norm unfortunately isn't enough to reproduce the original result.  

Downcasting to BF16 only happens after modulation
https://github.com/genmoai/mochi/blob/f3a800aea5862b4af13e66ff77eea1967c8c3a7f/src/genmo/mochi_preview/dit/joint_model/mod_rmsnorm.py#L15

And in the case of tanh modulation, right before adding to the residual
https://github.com/genmoai/mochi/blob/f3a800aea5862b4af13e66ff77eea1967c8c3a7f/src/genmo/mochi_preview/dit/joint_model/residual_tanh_gated_rmsnorm.py#L6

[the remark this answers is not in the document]
DN6: And in the original repo the RMSNorm used in the QK Norm (which we define here as MochiRMSNorm) also upcasts to FP32 before multiplying by the weight and then downcasting back to BF16
https://github.com/genmoai/mochi/blob/f3a800aea5862b4af13e66ff77eea1967c8c3a7f/src/genmo/mochi_preview/dit/joint_model/layers.py#L179 

Our RMSNorm only upcasts when computing variance.`hidden_states` remains in BF16
https://github.com/huggingface/diffusers/blob/c96bfa5c80eca798d555a79a491043c311d0f608/src/diffusers/models/normalization.py#L531-L533

src/diffusers/models/transformers/transformer_mochi.py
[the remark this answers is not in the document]
DN6: Oh yes, the RMSNorm in the Attentions can be replaced.

src/diffusers/models/transformers/transformer_mochi.py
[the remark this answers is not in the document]
DN6: Oh. I see what you mean. Upcast hidden states in MochiModulatedNorm and the output is equivalent. Got it.

src/diffusers/pipelines/ltx/pipeline_ltx.py
```diff
@@ -198,7 +198,6 @@ def __init__(
             self.tokenizer.model_max_length if hasattr(self, "tokenizer") and self.tokenizer is not None else 128
         )
 
-    # Copied from diffusers.pipelines.mochi.pipeline_mochi.MochiPipeline._get_t5_prompt_embeds with 256->128
```
DN6, line 201: Mochi zeros out negative prompt embeds if they are an empty string.

src/diffusers/models/transformers/transformer_mochi.py
yiyixuxu: @DN6 we can commit this suggestion!

src/diffusers/models/transformers/transformer_mochi.py
yiyixuxu: I think it will automtically upcast, no? here in `hidden_states = hidden_states * scale` if the hidden_states is in float32 an scale is in bfloat16, the operation will be in float32 and we do not need to explicitly upcast scale, no?

this is non-merge blocking! just curious if I misunderstood something 

```python
    def forward(self, hidden_states, scale=None):
        ....
        if scale is not None:
            hidden_states = hidden_states * scale

        hidden_states = hidden_states.to(hidden_states_dtype)

        return hidden_states
```

yiyixuxu (approved):
thank you for the fix!!!!
so much work went into this PR:)

src/diffusers/models/attention_processor.py
```diff
@@ -906,6 +906,177 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+        return self.processor(
+            self,
+            hidden_states,
+            encoder_hidden_states=encoder_hidden_states,
+            attention_mask=attention_mask,
+            **kwargs,
+        )
+
+
+class MochiAttnProcessor2_0:
```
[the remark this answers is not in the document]
a-r-r-o-w: I think we should do the following going forward as design choice (just personal opinion so let's try to forumalate a plan for consistency):
- Apart from transformer model code, the transformer files will also contain all the relevant attention processor implementation.
- If an alternate `Attention` class is required, let's keep it in the transformer file as well. These custom classes require some common methods that will probably not change between implementantions. For this, let's create a `AttentionMixin` class - for changing/getting attention processors, fusing, etc.
- If an attention processor is required in both transformer and VAE (and possibly a different file) because of some common parts shared, let's keep the implementation in transformer file, and import it in the vae. If there's no common attention processor, let's keep the implementation respectively in transformer or vae.
- If some specific layers are shared between transformer and vae (for example, GLUMB convolution in Sana), let's keep the implementation in transformer file too.
- Let's create dedicated RoPE classes for each implementation. Any concerns about speed due to recreating the embeddings every inference step can be addressed by caching. Something as simple as `functools.cache` works here if we make the rope calculation forward dependant on just the num_frames, height, width. But a save hook would work as well. This - we can look into a bit later

WDYT?

src/diffusers/models/transformers/transformer_mochi.py
DN6: It's the 1 + scale. The addition has to happen FP32 otherwise we get numerical differences.

DN6:
Failing LoRA tests are unrelated. Seem to be timing out on SDXL tests. Merging.

Issue #10013, opened 2024-11-25
[Core] refactor model loading


sayakpaul:
Currently, we have got two codepaths:

1. For non-sharded checkpoints we do: https://github.com/huggingface/diffusers/blob/047bf492914ddc9393070b8f73bba5ad5823eb29/src/diffusers/models/modeling_utils.py#L855
2. For sharded checkpoints we do: https://github.com/huggingface/diffusers/blob/047bf492914ddc9393070b8f73bba5ad5823eb29/src/diffusers/models/modeling_utils.py#L886

And then for the (bnb) quantized checkpoints, we merge a sharded checkpoint:
https://github.com/huggingface/diffusers/blob/047bf492914ddc9393070b8f73bba5ad5823eb29/src/diffusers/models/modeling_utils.py#L775

Essentially, we shouldn't have to merge sharded checkpoints even if it's quantized. 

This will also allow us to more generally use `keep_module_in_fp32` for sharded checkpoints. Currently, we have this logic for casting a model (which is [tested](https://github.com/huggingface/diffusers/blob/43534a8d1fd405fd0d1e74f991ab97f743bd3e59/tests/quantization/bnb/test_4bit.py#L190) thoroughly):
https://github.com/huggingface/diffusers/blob/43534a8d1fd405fd0d1e74f991ab97f743bd3e59/src/diffusers/models/modeling_utils.py#L997

When using `load_model_dict_into_meta()`, we do consider `keep_module_in_fp32`:
https://github.com/huggingface/diffusers/blob/43534a8d1fd405fd0d1e74f991ab97f743bd3e59/src/diffusers/models/model_loading_utils.py#L177

But since for sharded checkpoints, we use `load_checkpoint_and_dispatch()`, there is no way to pass `keep_module_in_fp32`:
https://huggingface.co/docs/accelerate/main/en/package_reference/big_modeling#accelerate.load_checkpoint_and_dispatch

As discussed with @SunMarc, it's better to uniformize this so that we don't have to maintain two different codepaths and rely completely on `load_model_dict_into_meta()`. Marc has kindly agreed to open a PR to attempt this (this could be done in a series of PRs if needed). But I will join if any help is needed.

sayakpaul:
@huggingface/diffusers Marc has started working on this 🥳

PR #10037, opened 2024-11-27
[FEAT] DDUF format

SunMarc:
# What does this PR do?

This PR adds support to DDUF format. More docs about this format [here](https://github.com/huggingface/huggingface_hub/pull/2692).

Requires transformers PR : https://github.com/huggingface/transformers/pull/35217/files (merged)
Requires huggingface_hub PR: https://github.com/huggingface/huggingface_hub/pull/2692 (released)

### Load from the hub 
```python
pipe = StableDiffusion3Pipeline.from_pretrained("DDUF/stable-diffusion-3-medium-diffusers-DDUF",
                                                dduf_file="stable-diffusion-3-medium-diffusers.dduf")
pipe = pipe.to("cuda")

image = pipe(
    "A cat holding a sign that says hello world",
    negative_prompt="",
    num_inference_steps=28,
    guidance_scale=7.0,
).images[0]
image.save("cat.png")

```
### Save to DDUF format
```python
from huggingface_hub import export_folder_as_dduf

pipe = StableDiffusion3Pipeline.from_pretrained("stabilityai/stable-diffusion-3-medium-diffusers", 
                                                torch_dtype=torch.float16)

pipe.save_pretrained(save_directory="stable-diffusion-3-medium-diffusers")

export_folder_as_dduf("stable-diffusion-3-medium-diffusers.dduf", folder_path="stable-diffusion-3-medium-diffusers")

```

TODO list:

- [x] stream directly from the .dduf 
- [x] replace loader with huggingface_hub parser https://github.com/huggingface/huggingface_hub/pull/2692#event-15529630331
- [x] SD pipeline
- [x] Flux pipeline


src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,6 +306,34 @@ def is_saveable_module(name, value):
...
+                    mode = "a"
+                else:
+                    # Open in write mode to create it if it doesn't exist
+                    mode = "w:"
+                with tarfile.open(dduf_file_path, mode) as tar:
+                    dir_to_archive = os.path.join(save_directory, pipeline_component_name)
+                    if os.path.isdir(dir_to_archive):
+                        tar.add(dir_to_archive, arcname=os.path.basename(dir_to_archive))
+                        # remove from save_directory after we added it to the archive
+                        shutil.rmtree(dir_to_archive)
```
sayakpaul, line 335: Might wanna consider using a ZIP archive instead, as also mentioned over Slack.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,6 +306,34 @@ def is_saveable_module(name, value):
...
+                        f"Removing the existing folder {dduf_file_path} so that we can save the DDUF archive."
+                    )
+                    shutil.rmtree(dduf_file_path)
+                if (
+                    os.path.exists(dduf_file_path)
+                    and os.path.isfile(dduf_file_path)
+                    and tarfile.is_tarfile(dduf_file_path)
+                ):
+                    # Open in append mode if the file exists
+                    mode = "a"
```
sayakpaul, line 326: Hmm, could this be problematic if we had the same file but it exists in a corrupted state?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,6 +306,34 @@ def is_saveable_module(name, value):
...
+                else:
+                    # Open in write mode to create it if it doesn't exist
+                    mode = "w:"
+                with tarfile.open(dduf_file_path, mode) as tar:
+                    dir_to_archive = os.path.join(save_directory, pipeline_component_name)
+                    if os.path.isdir(dir_to_archive):
```
sayakpaul, line 332: We also rely on `model_index.json` to help with loading and a few verifications. We should package that in the tar archive, IMO. Any reason not to?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -617,6 +652,8 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
             variant (`str`, *optional*):
                 Load weights from a specified variant filename such as `"fp16"` or `"ema"`. This is ignored when
                 loading `from_flax`.
+            dduf(`str`, *optional*):
+                Load weights from the specified dduf archive or folder.
```
sayakpaul, line 656: nit

```suggestion
            dduf (`str`, *optional*):
                Load weights from the specified dduf archive or folder.
```

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1346,6 +1407,10 @@ def download(cls, pretrained_model_name, **kwargs) -> Union[str, os.PathLike]:
             allow_patterns += [f"{custom_pipeline}.py"] if f"{custom_pipeline}.py" in filenames else []
             # also allow downloading config.json files with the model
             allow_patterns += [os.path.join(k, "config.json") for k in model_folder_names]
+            # also allow downloading the dduf
+            # TODO: check that the file actually exist
+            if dduf is not None:
+                allow_patterns += [dduf]
```
sayakpaul, line 1413: Should `dduf` argument values go through some validations?

sayakpaul (commented):
Thanks for starting this! Apart from the initial comments I left, I have a few higher level questions:

We're assuming that the archive will contain all the pipeline-related files in the `diffusers` format. I am fine with that. This means we'll be shipping DDUF only at the pipeline-level for now, correct?

sayakpaul:
I just tried out `bitsanbdytes` and it seems to work as well. Checkpoint: https://huggingface.co/DDUF/Flux.1-Dev-Tnf4-TE2NF4 (private for now)

<details>
<summary>Serialization</summary>

```py
from diffusers import FluxTransformer2DModel, DiffusionPipeline 
from huggingface_hub import create_repo, upload_folder
import torch 

repo_id = "DDUF/Flux.1-Dev-Tnf4-TE2NF4"
repo_id = create_repo(repo_id=repo_id, private=True, exist_ok=True).repo_id

    "hf-internal-testing/flux.1-dev-nf4-pkg", subfolder="transformer"
)
    "hf-internal-testing/flux.1-dev-nf4-pkg", subfolder="text_encoder_2"
)
    transformer=transformer, 
    text_encoder=text_encoder_2,
    torch_dtype=torch.bfloat16
)

folder = "/raid/.cache/huggingface/sayak/flux.1-dduf"
pipeline.save_pretrained(folder, dduf_format=True, dduf_filename="Flux.1-Dev-Tnf4-TE2NF4.DDUF")
upload_folder(repo_id=repo_id, folder_path=folder, create_pr=True)
```

</details>

<details>
<summary>Loading</summary>

```py
import torch

repo_id = "DDUF/Flux.1-Dev-Tnf4-TE2NF4"
    repo_id, dduf="Flux.1-Dev-Tnf4-TE2NF4.DDUF", torch_dtype=torch.bfloat16
)
pipeline.enable_model_cpu_offload()

image = pipeline(
    "a cat standing by the sea waiting for its companion to come and console it.",
    guidance_scale=4.5,
    num_inference_steps=50,
    max_sequence_length=512,
    generator=torch.manual_seed(0)
).images[0]
image.save("dduf_out.png")
```

</details>

src/diffusers/pipelines/pipeline_utils.py
SunMarc: For this case, then we need to check before anything else if we do have an achieve and delete then. The logic for the append mode is to create the archive progressively as the models are getting saved. This way we use less disk memory

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -193,6 +193,8 @@ def save_pretrained(
         variant: Optional[str] = None,
         max_shard_size: Optional[Union[int, str]] = None,
         push_to_hub: bool = False,
+        dduf_format: bool = False,
+        dduf_filename: Optional[Union[str, os.PathLike]] = None,
```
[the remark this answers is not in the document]
SunMarc: I will just use that then !

src/diffusers/pipelines/pipeline_utils.py
SunMarc: Yeah i left it as a todo above but we would have to check that the file do actually exist. Any other suggestion is welcomed.

sayakpaul: Then perhaps a comment of TODO would be nice.

src/diffusers/pipelines/pipeline_utils.py
sayakpaul: Okay make sense but just wanted to flag this.

src/diffusers/pipelines/pipeline_utils.py
pcuenca: Also, expanding the archive and removing it could be problematic.

- If I understand correctly, we would be removing the symlink to the blob, but not the blob itself. We won't be saving disk space.
- If we did remove the blob, I think it'd trigger another download next time we call `.from_pretrained`.
- Removing the reference could maybe have consequences when checking for newer versions in the Hub (not sure).

When we change to using zips, I wonder if we could extract files progressively in memory during the load process, and keeping the big .dduf file as is. Having a disk snapshot is convenient, but I can see no easy solution to avoid duplicating space.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -762,6 +801,25 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
+            extract_to = os.path.join(cached_folder, f"{dduf}_extracted")
+            # if tar file, we need to extract the tarfile and remove it
+            if os.path.isfile(tar_file_path):
+                if tarfile.is_tarfile(tar_file_path):
+                    with tarfile.open(tar_file_path, "r") as tar:
+                        tar.extractall(extract_to)
+                    # remove tar archive to free memory
+                    os.remove(tar_file_path)
+                    # rename folder to match the name of the dduf archive
+                    os.rename(extract_to, tar_file_path)
```
pcuenca, line 817: (Moving my comment to the right place, sorry)

I think that expanding the archive and removing it could be problematic.

- If I understand correctly, we would be removing the symlink to the blob, but not the blob itself. We won't be saving disk space.
- If we did remove the blob, I think it'd trigger another download next time we call `.from_pretrained`.
- Removing the reference could maybe have consequences when checking for newer versions in the Hub (not sure).

When we change to using zips, I wonder if we could extract files progressively in memory during the load process, and keeping the big .dduf file as is. Having a disk snapshot is convenient, but I can see no easy solution to avoid duplicating space.

src/diffusers/pipelines/pipeline_utils.py
pcuenca: I agree with Sayak, `model_index.json` can serve as a first approach to containing pipeline metadata (Even if it's duplicated in the repo as a separate file anyway).

src/diffusers/pipelines/pipeline_utils.py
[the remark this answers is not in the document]
pcuenca: I think that most, if not all, the components of the file name should be inferred. One of the reasons to use them would be to provide confidence about the contents based on filename inspection, so we should try to make it work with something like an optional prefix, but build the rest automatically. So something like `dduf_name="my-lora"` resulting in the library saving to `"my-lora-stable-diffusion-3-medium-diffusers-fp16.dduf"`, or something like that.

src/diffusers/pipelines/pipeline_utils.py
SunMarc: Sounds good, I'll package it !

src/diffusers/pipelines/pipeline_utils.py
SunMarc: I agree. I will check if it is possible to stream directly from the zip or at least extract progressively in memory as you said

src/diffusers/pipelines/pipeline_utils.py
SunMarc: Switched to zip !

src/diffusers/pipelines/pipeline_utils.py
SunMarc: Done in the latest commit

src/diffusers/pipelines/pipeline_utils.py
SunMarc: fixed !

src/diffusers/pipelines/pipeline_utils.py
SunMarc: Thanks for the link ! I'll have a look !

julien-c: BTW on the Hub side we want to constraint the types of Zip a lot:
- no compression only (to make sure it's deduplicated well by XetHub)
- single disk (internal simplifications of Zip)
- no encryption etc

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,9 +302,54 @@ def is_saveable_module(name, value):
...
+                    # Open in append mode if the file exists
+                    mode = "a"
+                else:
+                    # Open in write mode to create it if it doesn't exist
+                    mode = "w"
+                with zipfile.ZipFile(dduf_file_path, mode=mode, compression=zipfile.ZIP_STORED) as zipf:
+                    dir_to_archive = os.path.join(save_directory, pipeline_component_name)
+                    if os.path.isdir(dir_to_archive):
+                        zipdir(dir_to_archive, zipf)
+                        shutil.rmtree(dir_to_archive)
```
julien-c, line 340: important: make sure the resulting zip file is uncompressed and its layout is "clean" (we will provide a tool to check this ASAP on the Hub side)

julien-c (commented):
BTW a bit orthogonal but do you have examples of quantized diffusers repos right now?

sayakpaul:
> BTW a bit orthogonal but do you have examples of quantized diffusers repos right now?

This is one example: https://huggingface.co/sayakpaul/FLUX.1-Fill-dev-nf4. Note that it doesn't include all the models needed to run Flux as it's not typical to quantize the VAE and the other CLIP-based text encoder.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,9 +302,54 @@ def is_saveable_module(name, value):
 
             save_method(os.path.join(save_directory, pipeline_component_name), **save_kwargs)
 
+            if dduf_filename:
+                import shutil
+                import zipfile
+
+                def zipdir(dir_to_archive, zipf):
+                    "zip a directory"
```
sayakpaul, line 310:
```suggestion
                    """Archive a directory"""
```

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,9 +302,54 @@ def is_saveable_module(name, value):
 
             save_method(os.path.join(save_directory, pipeline_component_name), **save_kwargs)
 
+            if dduf_filename:
```
sayakpaul, line 305: Maybe a "TODO" on validating `dduf_filename` and that validation should probably go more at the beginning so that we can error out earlier?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,9 +302,54 @@ def is_saveable_module(name, value):
...
+                                os.path.basename(dir_to_archive), os.path.relpath(file_path, start=dir_to_archive)
+                            )
+                            zipf.write(file_path, arcname=arcname)
+
+                dduf_file_path = os.path.join(save_directory, dduf_filename)
+
+                if os.path.isdir(dduf_file_path):
+                    logger.warning(
+                        f"Removing the existing folder {dduf_file_path} so that we can save the DDUF archive."
+                    )
```
sayakpaul, line 324: Would `dduf_file_path` be a directory with the ".dduf" extension? If so, should this be checked?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,9 +302,54 @@ def is_saveable_module(name, value):
...
         # finally save the config
         self.save_config(save_directory)
 
+        if dduf_filename:
+            import zipfile
+
+            with zipfile.ZipFile(dduf_file_path, mode="a", compression=zipfile.ZIP_STORED) as zipf:
+                config_path = os.path.join(save_directory, self.config_name)
+                zipf.write(config_path, arcname=os.path.basename(config_path))
+                os.remove(config_path)
```
sayakpaul, line 351:
```suggestion
        # Takes care of including the "model_index.json" inside the ZIP.
        # TODO: Include a DDUF a metadata file. 
        if dduf_filename:
            import zipfile

            with zipfile.ZipFile(dduf_file_path, mode="a", compression=zipfile.ZIP_STORED) as zipf:
                config_path = os.path.join(save_directory, self.config_name)
                zipf.write(config_path, arcname=os.path.basename(config_path))
                os.remove(config_path)
```

sayakpaul:
@SunMarc I don't think loading is working:

```py

repo_id = "DDUF/stable-diffusion-3-medium-diffusers-DDUF"
filename = "stable-diffusion-3-medium-diffusers.dduf"
pipeline = DiffusionPipeline.from_pretrained(repo_id, dduf=filename)
```

Leads to:

```bash
The above exception was the direct cause of the following exception:

  File "/fsx/sayak/diffusers/load_dduf.py", line 23, in <module>
    pipeline = DiffusionPipeline.from_pretrained(repo_id, dduf=filename)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
  File "/fsx/sayak/diffusers/src/diffusers/pipelines/pipeline_utils.py", line 776, in from_pretrained
    cached_folder = cls.download(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
  File "/fsx/sayak/diffusers/src/diffusers/pipelines/pipeline_utils.py", line 1379, in download
    config_file = hf_hub_download(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/file_download.py", line 862, in hf_hub_download
    return _hf_hub_download_to_cache_dir(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/file_download.py", line 925, in _hf_hub_download_to_cache_dir
    (url_to_download, etag, commit_hash, expected_size, head_call_error) = _get_metadata_or_catch_error(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/file_download.py", line 1376, in _get_metadata_or_catch_error
    metadata = get_hf_file_metadata(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/file_download.py", line 1296, in get_hf_file_metadata
    r = _request_wrapper(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/file_download.py", line 277, in _request_wrapper
    response = _request_wrapper(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/file_download.py", line 301, in _request_wrapper
    hf_raise_for_status(response)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/huggingface_hub/utils/_http.py", line 417, in hf_raise_for_status
    raise _format(EntryNotFoundError, message, response) from e
huggingface_hub.errors.EntryNotFoundError: 404 Client Error. (Request ID: Root=1-674931ef-1fdc6f274e6f862c31db623e;2056ffe1-9bd0-4667-806f-273827e95101)

Entry Not Found for url: https://huggingface.co/DDUF/stable-diffusion-3-medium-diffusers-DDUF/resolve/main/model_index.json.
```

which is expected.

src/diffusers/pipelines/pipeline_utils.py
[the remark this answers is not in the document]
julien-c: ok to use some kind of convention, but i don't think it should be an absolute requirement, ie. people can rename their files if they want to (but the metadata should be inside the file anyways)

src/diffusers/pipelines/pipeline_utils.py
Wauplin: To ensure restrictions are met, we'll provide tooling in Python to save/load DDUF files in a consistent way: https://github.com/huggingface/huggingface_hub/pull/2692. This will be part of `huggingface_hub` to that it can be easily reused by any library without having necessarily `diffusers` as a dependency.

src/diffusers/pipelines/pipeline_utils.py
SunMarc: validation will be done from huggingface_hub side

SunMarc:
> Update: I have pushed my updates in sayak-dduf which fixes it (it's incomplete but gives a heads-up). Feel free to cherry-pick commits from it. It has progressive extraction with a progressbar as well.

Merged !

src/diffusers/pipelines/pipeline_utils.py
julien-c: marking this as resolved

src/diffusers/configuration_utils.py
```diff
@@ -552,9 +565,12 @@ def extract_init_dict(cls, config_dict, **kwargs):
         return init_dict, unused_kwargs, hidden_config_dict
 
     @classmethod
-    def _dict_from_json_file(cls, json_file: Union[str, os.PathLike]):
-        with open(json_file, "r", encoding="utf-8") as reader:
-            text = reader.read()
+    def _dict_from_json_file(cls, json_file: Union[str, os.PathLike], dduf_reader=None):
```
sayakpaul, line 568: Do we wanna annotate the type hint of `dduf_reader`? No biggie, though.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -50,6 +50,7 @@
     CONFIG_NAME,
     DEPRECATED_REVISION_ARGS,
     BaseOutput,
+    DDUFReader,
```
sayakpaul, line 53: TODO: take this from https://github.com/huggingface/huggingface_hub/pull/2692/

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -301,9 +303,56 @@ def is_saveable_module(name, value):
...
+                        f"Removing the existing folder {dduf_file_path} so that we can save the DDUF archive."
+                    )
+                    shutil.rmtree(dduf_file_path)
+                if (
+                    os.path.exists(dduf_file_path)
+                    and os.path.isfile(dduf_file_path)
+                    and zipfile.is_zipfile(dduf_file_path)
+                ):
+                    # Open in append mode if the file exists
+                    mode = "a"
```
sayakpaul, line 333: Think this needs to be replaced with the export methods from https://github.com/huggingface/huggingface_hub/pull/2692/?

src/diffusers/configuration_utils.py
```diff
@@ -347,6 +347,7 @@ def load_config(
         _ = kwargs.pop("mirror", None)
         subfolder = kwargs.pop("subfolder", None)
         user_agent = kwargs.pop("user_agent", {})
+        dduf_reader = kwargs.pop("dduf_reader", None)
```
sayakpaul, line 350: `load_config()` is a public method, so I have a few considerations here:

* What happens both `cached_folder` and `dduf_reader` are supplied?
* Should the combination of `cached_folder` and `dduf_reader` values be validated?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -757,7 +812,14 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
             )
             logger.warning(warn_msg)
 
-        config_dict = cls.load_config(cached_folder)
+        dduf_reader = None
+        if dduf:
+            zip_file_path = os.path.join(cached_folder, dduf)
+            dduf_reader = DDUFReader(zip_file_path)
+            # The reader contains already all the files needed, no need to check it again
+            cached_folder = ""
```
sayakpaul, line 820: See: https://github.com/huggingface/diffusers/pull/10037/files#r1870594275.

src/diffusers/pipelines/stable_audio/pipeline_stable_audio.py
```diff
@@ -34,6 +34,7 @@
 from ..pipeline_utils import AudioPipelineOutput, DiffusionPipeline
```
sayakpaul, line 34: Changes here can be removed.

src/diffusers/utils/dduf.py
```diff
@@ -0,0 +1,41 @@
+import zipfile
+
+
+class DDUFReader:
```
sayakpaul, line 4: TODO:

Take this from https://github.com/huggingface/huggingface_hub/pull/2692/

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -193,6 +194,7 @@ def save_pretrained(
         variant: Optional[str] = None,
         max_shard_size: Optional[Union[int, str]] = None,
         push_to_hub: bool = False,
+        dduf_filename: Optional[Union[str, os.PathLike]] = None,
```
Wauplin, line 197:
```suggestion
        dduf_file: Optional[Union[str, os.PathLike]] = None,
```

I would rename it `dduf_file` for consistency within `diffusers` and with `transformers`'s `.from_pretrained(model_id, gguf_file=filename)`

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -617,6 +668,8 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
             variant (`str`, *optional*):
                 Load weights from a specified variant filename such as `"fp16"` or `"ema"`. This is ignored when
                 loading `from_flax`.
+            dduf (`str`, *optional*):
+                Load weights from the specified dduf archive or folder.
```
Wauplin, line 672:
```suggestion
            dduf_file (`str`, *optional*):
                Load weights from the specified dduf file.
```

Same here, I would also rename it `dduf_file` for consistency within `diffusers` and with `transformers`'s `.from_pretrained(model_id, gguf_file=filename)`

Wauplin (commented):
(I haven't thoroughly reviewed the PR)

src/diffusers/pipelines/pipeline_utils.py
SunMarc: Sounds good to me. I didn't put the same name before as I extracted the archive and it ended up being a folder.

src/diffusers/pipelines/pipeline_utils.py
SunMarc: switched !

src/diffusers/configuration_utils.py
SunMarc: yeah I will add the annotation !

src/diffusers/utils/dduf.py
SunMarc: switched to huggingface hub api

src/diffusers/pipelines/pipeline_utils.py
SunMarc: WDYM ?

src/diffusers/configuration_utils.py
SunMarc: passing `dduf_reader` or `dduf_entites` means that we already downloaded the files and prepared everything to read the file. So cached_folder is not really useful at this stage I think.

src/diffusers/configuration_utils.py
```diff
@@ -358,8 +359,22 @@ def load_config(
...
+        if dduf_entries:
+            if subfolder is not None:
+                if os.path.join(pretrained_model_name_or_path, subfolder, cls.config_name) in dduf_entries:
+                    config_file = os.path.join(subfolder, cls.config_name)
+                else:
+                    raise ValueError(
+                        f"We did not manage to find the file {os.path.join(pretrained_model_name_or_path, subfolder, cls.config_name)} in the dduf file. We only have the following files {dduf_entries.keys()}"
+                    )
+            elif os.path.join(pretrained_model_name_or_path, cls.config_name) in dduf_entries:
+                config_file = os.path.join(pretrained_model_name_or_path, cls.config_name)
```
sayakpaul, line 372: Bit cleaner no?

```suggestion
            if subfolder is not None:
                config_file = if os.path.join(pretrained_model_name_or_path, subfolder, cls.config_name) in dduf_entries
            else:
                config_file = if os.path.join(pretrained_model_name_or_path, cls.config_name)
            if config_filepath not in dduf_entries:
                raise ValueError(
                    f"We did not manage to find the file {os.path.join(pretrained_model_name_or_path, subfolder, cls.config_name)} in the dduf file. We only have the following files {dduf_entries.keys()}"
                )
```

src/diffusers/configuration_utils.py
```diff
@@ -358,8 +359,22 @@ def load_config(
...
+            else:
+                raise ValueError(
+                    f"We did not manage to find the file {os.path.join(pretrained_model_name_or_path, cls.config_name)} in the dduf file. We only have the following files {dduf_entries.keys()}"
+                )
```
sayakpaul, line 376: Bit cleaner no?

```suggestion
            if subfolder is not None:
                config_file = if os.path.join(pretrained_model_name_or_path, subfolder, cls.config_name) in dduf_entries
            else:
                config_file = if os.path.join(pretrained_model_name_or_path, cls.config_name)
            if config_filepath not in dduf_entries:
                raise ValueError(
                    f"We did not manage to find the file {os.path.join(pretrained_model_name_or_path, subfolder, cls.config_name)} in the dduf file. We only have the following files {dduf_entries.keys()}"
                )
```

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -763,7 +820,23 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
             )
             logger.warning(warn_msg)
 
-        config_dict = cls.load_config(cached_folder)
+        dduf_entries = None
+        if dduf_file:
+            if not is_huggingface_hub_version(">", "0.26.3"):
+                (">=", "0.17.0.dev0")
```
sayakpaul, line 826: I think this needs to go away

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -763,7 +820,23 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
+                raise RuntimeError(
+                    "In order to load a dduf file, you need to install huggingface_hub>0.26.3. "
+                    "You can install it with the following: `pip install --upgrade huggingface_hub"
+                )
```
sayakpaul, line 830:
```suggestion
                raise RuntimeError(
                    "To load a dduf file, you need to install huggingface_hub>0.26.3. "
                    "You can install it with the following: `pip install --upgrade huggingface_hub`."
                )
```

tests/pipelines/test_pipelines_common.py
```diff
@@ -1902,6 +1902,30 @@ def test_StableDiffusionMixin_component(self):
             )
         )
 
+    # @pytest.mark.xfail(condition=not os.getenv("RUN_DDUF_TEST", False), strict=True)
+    # Should consider guarding the test with proper transformers and huggingface_hub versions.
+    def test_save_load_dduf(self):
```
sayakpaul, line 1907: @SunMarc I added this simple test. However, it's failing with:

```bash
FAILED tests/pipelines/flux/test_pipeline_flux.py::FluxPipelineFastTests::test_save_load_dduf - AttributeError: 'ZipInfo' object has no attribute '_end_offset'
```

I am on Python 3.10.12. 

@Wauplin is this expected? 

To run:

```bash
pytest tests/pipelines/flux/test_pipeline_flux.py::FluxPipelineFastTests::test_save_load_dduf
```

Wauplin: Not expected no :smile: Seems like I'll have to address the "TODO:" earlier than what I thought :disappointed:

src/diffusers/pipelines/pipeline_utils.py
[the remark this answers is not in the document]
sayakpaul: Yeah that is the plan.

src/diffusers/pipelines/pipeline_utils.py
sayakpaul: I think it's resolved. So, disregard.

tests/pipelines/test_pipelines_common.py
Wauplin: @sayakpaul fixed it in https://github.com/huggingface/huggingface_hub/pull/2692/commits/ba1e6a4e05070d7a902b4d66db29fc9955150b2c. I don't use the private arg anymore :)

sayakpaul: @SunMarc I think we would need to merge the commit Lucain is mentioning in your branch `dduf-parser-more-feature`. Otherwise, I still have the same [error](https://github.com/huggingface/diffusers/pull/10037/files#r1874795428) when running the test.

SunMarc: done !

src/diffusers/pipelines/pipeline_loading_utils.py
```diff
@@ -721,7 +722,10 @@ def load_sub_model(
             loading_kwargs["low_cpu_mem_usage"] = False
 
     # check if the module is in a subdirectory
-    if os.path.isdir(os.path.join(cached_folder, name)):
+    if dduf_entries:
+        loading_kwargs["dduf_entries"] = dduf_entries
+        loaded_sub_model = load_method(name, **loading_kwargs)
+    elif os.path.isdir(os.path.join(cached_folder, name)):
```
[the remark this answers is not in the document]
Wauplin: see suggested workaround to avoid `transformers` update: https://github.com/huggingface/diffusers/pull/10171

SunMarc:
Thanks for merging your PR @Wauplin ! I fixed the tokenizer and the transformers model loading part. It should work now with the following PR in transformers: https://github.com/huggingface/transformers/pull/35217/files

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -300,9 +305,29 @@ def is_saveable_module(name, value):
 
             save_method(os.path.join(save_directory, pipeline_component_name), **save_kwargs)
 
+            if dduf_file:
+                import shutil
+
+                from huggingface_hub import export_folder_as_dduf
```
sayakpaul, line 311: No need to version-check this as we're modifying the `setup.py` already.

tests/pipelines/test_pipelines_common.py
sayakpaul: @SunMarc @Wauplin currently, it's failing with the following:

```bash
FAILED tests/pipelines/flux/test_pipeline_flux.py::FluxPipelineFastTests::test_save_load_dduf - huggingface_hub.errors.DDUFExportError: Missing required 'model_index.json' entry in DDUF file.
```

SunMarc: can you try the latest commit from my branch on huggingface hub ?

sayakpaul: Same.

Wauplin: As mentioned [on slack](https://huggingface.slack.com/archives/C0821RXC9T5/p1733832282906659) (private) and in https://github.com/huggingface/huggingface_hub/pull/2692#issuecomment-2531395376 I really think we should post-pone the `pipe.save_pretrained(tmpdir, dduf_file=dduf_filename)` feature and therefore remove this test for now.

sayakpaul: Yeah I am going to modify the test.

sayakpaul: @Wauplin this is the current test:

```py
def test_save_load_dduf(self):
    from huggingface_hub import export_folder_as_dduf

    pipe = pipe.to(torch_device)

    inputs = self.get_dummy_inputs(device="cpu")
    inputs.pop("generator")
    inputs["generator"] = torch.manual_seed(0)

    pipeline_out = pipe(**inputs).images

        dduf_filename = f"{pipe.__class__.__name__.lower()}.dduf"
        pipe.save_pretrained(tmpdir, safe_serialization=True)
        export_folder_as_dduf(dduf_filename, folder_path=tmpdir)
        loaded_pipe = self.pipeline_class.from_pretrained(tmpdir, dduf_file=dduf_filename).to(torch_device)

    inputs["generator"] = torch.manual_seed(0)
    loaded_pipeline_out = loaded_pipe(**inputs).images

    assert np.allclose(pipeline_out, loaded_pipeline_out)
```

As you can see, I am now leveraging the `export_folder_as_dduf()` functionality instead of directly serializing DDUF with `save_pretrained()`. With this, it's currently failing:

```bash
FAILED tests/pipelines/flux/test_pipeline_flux.py::FluxPipelineFastTests::test_save_load_dduf - huggingface_hub.errors.DDUFExportError: Invalid DDUF file structure.
```

Is this expected?

sayakpaul: Sorry about my oversight. The scheduler's config should be `scheduler_config.json` and not `config.json`. 

```py
DDUF_FOLDER_REQUIRED_ENTRIES = {
    # Each folder must contain at least one of these entries
    "config.json",
    "tokenizer_config.json",
    "image_processor.json",
    "scheduler_config.json"
}
```

sayakpaul: Concretely, this is `git diff`:

```diff
diff --git a/src/huggingface_hub/serialization/_dduf.py b/src/huggingface_hub/serialization/_dduf.py
index 23d40505..20cdcea3 100644
--- a/src/huggingface_hub/serialization/_dduf.py
+++ b/src/huggingface_hub/serialization/_dduf.py
@@ -27,6 +27,7 @@ DDUF_FOLDER_REQUIRED_ENTRIES = {
     "config.json",
     "tokenizer_config.json",
     "image_processor.json",
+    "scheduler_config.json"
 }
```

sayakpaul: With this change, the test is successfully passing but I am seeing:

```bash
src/diffusers/pipelines/transformers_loading_utils.py:96:55: F821 Undefined name `tmp_generation_config_file`
```

while running `make style && make quality`, which makes sense to me.

sayakpaul: Cc: @SunMarc for the error ^

src/diffusers/configuration_utils.py
```diff
@@ -358,8 +359,17 @@ def load_config(
...
                 "`ConfigMixin`. Please make sure to define `config_name` in a class inheriting from `ConfigMixin`"
             )
-
-        if os.path.isfile(pretrained_model_name_or_path):
+        # Custom path for now
+        if dduf_entries:
+            if subfolder is not None:
+                config_file = os.path.join(pretrained_model_name_or_path, subfolder, cls.config_name)
+            else:
+                config_file = os.path.join(pretrained_model_name_or_path, cls.config_name)
```
Wauplin, line 367: It feels weird to me to use `pretrained_model_name_or_path` and `subfolder` to generate a DDUF entry. These two parameters refers to paths on a system. But when dduf entries are passed, the entry names are not based on paths on the system.

Maybe `subfolder` can be useful to load from a subfolder in the DDUF archive (but I haven't spotted where this is done). For `pretrained_model_name_or_path` it seems to be explicitly set to `""` below to make it work here by ignoring the value in `os.path.join`. That feels like a weird behavior to me (I'd just not used these values here).

Wauplin: Also, paths inside a DDUF file must always be `"/"`-separated so this code could produce an error with Windows paths `"\\"`

src/diffusers/configuration_utils.py
```diff
@@ -552,9 +560,12 @@ def extract_init_dict(cls, config_dict, **kwargs):
         return init_dict, unused_kwargs, hidden_config_dict
 
     @classmethod
-    def _dict_from_json_file(cls, json_file: Union[str, os.PathLike]):
-        with open(json_file, "r", encoding="utf-8") as reader:
-            text = reader.read()
+    def _dict_from_json_file(cls, json_file: Union[str, os.PathLike], dduf_entries=None):
```
Wauplin, line 563:
```suggestion
    def _dict_from_json_file(cls, json_file: Union[str, os.PathLike], dduf_entries: Optional[Dict[str, DDUFEntry]] = None):
```

src/diffusers/configuration_utils.py
```diff
@@ -347,6 +347,7 @@ def load_config(
         _ = kwargs.pop("mirror", None)
         subfolder = kwargs.pop("subfolder", None)
         user_agent = kwargs.pop("user_agent", {})
+        dduf_entries = kwargs.pop("dduf_entries", None)
```
Wauplin, line 350:
```suggestion
        dduf_entries: Optional[Dict[str, DDUFEntry]] = kwargs.pop("dduf_entries", None)
```

src/diffusers/models/model_loading_utils.py
```diff
@@ -128,7 +128,7 @@ def _fetch_remapped_cls_from_config(config, old_class):
         return old_class
 
 
-def load_state_dict(checkpoint_file: Union[str, os.PathLike], variant: Optional[str] = None):
+def load_state_dict(checkpoint_file: Union[str, os.PathLike], variant: Optional[str] = None, dduf_entries=None):
```
Wauplin, line 131:
```suggestion
def load_state_dict(checkpoint_file: Union[str, os.PathLike], variant: Optional[str] = None, dduf_entries: Optional[Dict[str, DDUFEntry]] = None):
```

src/diffusers/models/model_loading_utils.py
```diff
@@ -274,6 +280,7 @@ def _fetch_index_file(
     revision,
     user_agent,
     commit_hash,
+    dduf_entries=None,
```
Wauplin, line 283:
```suggestion
    dduf_entries: Optional[Dict[str, DDUFEntry]] = None,
```

src/diffusers/models/model_loading_utils.py
```diff
@@ -350,6 +358,7 @@ def _fetch_index_file_legacy(
...
```
Wauplin, line 361:
```suggestion
    dduf_entries: Optional[Dict[str, DDUFEntry]] = None,
```

src/diffusers/models/modeling_utils.py
```diff
@@ -586,6 +586,7 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
         variant = kwargs.pop("variant", None)
         use_safetensors = kwargs.pop("use_safetensors", None)
         quantization_config = kwargs.pop("quantization_config", None)
+        dduf_entries = kwargs.pop("dduf_entries", None)
```
Wauplin, line 589:
```suggestion
        dduf_entries: Optional[Dict[str, DDUFEntry]] = kwargs.pop("dduf_entries", None)
```

src/diffusers/models/model_loading_utils.py
Wauplin: (nit) (unrelated to this PR) from what I can see, it doesn't seem that the `variant` parameter is ever used in the method

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -300,9 +305,29 @@ def is_saveable_module(name, value):
...
+
+                dduf_file_path = os.path.join(save_directory, dduf_file)
+                dir_to_archive = os.path.join(save_directory, pipeline_component_name)
+                if os.path.isdir(dir_to_archive):
+                    export_folder_as_dduf(dduf_file_path, dir_to_archive)
+                    shutil.rmtree(dir_to_archive)
```
Wauplin: to remove?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -300,9 +305,29 @@ def is_saveable_module(name, value):
...
         self.save_config(save_directory)
 
+        # Takes care of including the "model_index.json" inside the ZIP.
+        if dduf_file:
+            from huggingface_hub import add_entry_to_dduf
+
+            config_path = os.path.join(save_directory, self.config_name)
+            # add config.json to the root of the dduf_file_path
+            add_entry_to_dduf(dduf_file_path, self.config_name, content=config_path)
+            os.remove(config_path)
```
Wauplin: to remove?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1273,6 +1314,7 @@ def download(cls, pretrained_model_name, **kwargs) -> Union[str, os.PathLike]:
         use_onnx = kwargs.pop("use_onnx", None)
         load_connected_pipeline = kwargs.pop("load_connected_pipeline", False)
         trust_remote_code = kwargs.pop("trust_remote_code", False)
+        dduf_file = kwargs.pop("dduf_file", None)
```
Wauplin, line 1294:
```suggestion
        dduf_file: Optional[Dict[str, DDUFEntry]] = kwargs.pop("dduf_file", None)
```

src/diffusers/utils/__init__.py
```diff
@@ -35,6 +35,7 @@
     WEIGHTS_INDEX_NAME,
     WEIGHTS_NAME,
 )
+from .dduf import DDUFReader
```
Wauplin, line 38:
```suggestion
```

src/diffusers/utils/hub_utils.py
```diff
@@ -291,9 +291,21 @@ def _get_model_file(
...
-    if os.path.isfile(pretrained_model_name_or_path):
+
+    if dduf_entries:
+        if os.path.join(pretrained_model_name_or_path, weights_name) in dduf_entries:
+            return os.path.join(pretrained_model_name_or_path, weights_name)
+        elif (
+            subfolder is not None
+            and os.path.join(pretrained_model_name_or_path, subfolder, weights_name) in dduf_entries
+        ):
+            return os.path.join(pretrained_model_name_or_path, weights_name)
```
Wauplin, line 305: same as above, mixing paths on disk and DDUF entries feels weird + can lead to errors on Windows machines

src/diffusers/utils/import_utils.py
```diff
@@ -789,6 +798,21 @@ def is_k_diffusion_version(operation: str, version: str):
...
+        operation (`str`):
+            A string representation of an operator, such as `">"` or `"<="`
+        version (`str`):
+            A version string
+    """
+    if not _huggingface_hub_available:
+        return False
+    return compare_versions(parse(_huggingface_hub_version), operation, version)
+
+
```
Wauplin, line 815:
```suggestion
```

(same)

src/diffusers/configuration_utils.py
SunMarc: Actually, `pretrained_model_name_or_path` will take for value "tokenizer" or "text_encoder" for example. We discard the paths on the system but only keep the structure of the folder.

src/diffusers/configuration_utils.py
SunMarc: Done !

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -772,6 +809,7 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
         # We retrieve the information by matching whether variant model checkpoints exist in the subfolders.
         # Example: `diffusion_pytorch_model.safetensors` -> `diffusion_pytorch_model.fp16.safetensors`
         # with variant being `"fp16"`.
+        # TODO: adapt logic for DDUF files (at the moment, scans the local directory which doesn't make sense in DDUF context)
```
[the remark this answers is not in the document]
SunMarc: After discussions with Sayak, it doesn't make sense to do this for dduf file.

src/diffusers/utils/dduf.py
```diff
@@ -0,0 +1,41 @@
+import zipfile
```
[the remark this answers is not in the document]
SunMarc: indeed ! done !

src/diffusers/configuration_utils.py
Wauplin: > Actually, pretrained_model_name_or_path will take for value "tokenizer" or "text_encoder" for example. We discard the paths on the system but only keep the structure of the folder.

Oh ok, good! But then `subfolder` is not needed at this stage? Since DDUF files can only have 1 level of directory.

src/diffusers/utils/import_utils.py
SunMarc: merged

src/diffusers/configuration_utils.py
SunMarc: Make sense, we only allow for 1 level of directory, then it shouldn't be necessary.

SunMarc: fixed it

src/diffusers/pipelines/pipeline_utils.py
SunMarc: yes

src/diffusers/utils/hub_utils.py
SunMarc: fixed !

tests/pipelines/test_pipelines_common.py
SunMarc: Fixed quality issue !

SunMarc:
CI is failing as `huggingface_hub` 0.27.0 is still not released cc @sayakpaul

SunMarc:
Where should we put the docs reegarding DDUF format in the docs @sayakpaul @stevhliu ? I was thinking of just a basic summary + link to huggingface_hub docs + small script to show how to load and save.

stevhliu:
I think you can add the DDUF docs in our [Model files and layout](https://huggingface.co/docs/diffusers/main/en/using-diffusers/other-formats#files) guide. Basic usage as it pertains to using diffusion models is perfect! 👌

Wauplin:
@stevhliu  I also tagged you on the docs PR: https://github.com/huggingface/hub-docs/pull/1526 which is not diffusers-specific but would still benefit from an example. It describes the format with motivations and specifications.

tests/pipelines/test_pipelines.py
```diff
@@ -1802,6 +1804,25 @@ def test_pipe_same_device_id_offload(self):
         sd.maybe_free_model_hooks()
         assert sd._offload_gpu_id == 5
 
+    @require_hf_hub_version_greater("0.26.5")
```
sayakpaul, line 1807: @SunMarc I have added a loading from Hub test as well.

src/diffusers/pipelines/transformers_loading_utils.py
```diff
@@ -0,0 +1,121 @@
...
+                tmp_entry_path = os.path.join(tmp_dir, *entry_name.split("/"))
+                # need to create intermediary directory if they don't exist
+                os.makedirs(os.path.dirname(tmp_entry_path), exist_ok=True)
+                with open(tmp_entry_path, "wb") as f:
+                    with entry.as_mmap() as mm:
+                        f.write(mm)
+        return cls.from_pretrained(os.path.dirname(tmp_entry_path), **kwargs)
+
+
+def load_transformers_model_from_dduf(
```
sayakpaul, line 59: Should we make these methods private?

src/diffusers/configuration_utils.py
```diff
@@ -358,8 +359,21 @@ def load_config(
...
-        if os.path.isfile(pretrained_model_name_or_path):
+        # Custom path for now
+        if dduf_entries:
+            if subfolder is not None:
+                raise ValueError("DDUF file only allow for 1 level of directory. Please check the DDUF structure")
+            # paths inside a DDUF file must always be "/"
+            config_file = (
+                cls.config_name
+                if pretrained_model_name_or_path == ""
+                else "/".join([pretrained_model_name_or_path, cls.config_name])
```
sayakpaul, line 370:
```suggestion
                else os.path.join(pretrained_model_name_or_path, cls.config_name)
```

src/diffusers/configuration_utils.py
```diff
@@ -358,8 +359,21 @@ def load_config(
                 "`self.config_name` is not defined. Note that one should not load a config from "
                 "`ConfigMixin`. Please make sure to define `config_name` in a class inheriting from `ConfigMixin`"
             )
-
-        if os.path.isfile(pretrained_model_name_or_path):
+        # Custom path for now
+        if dduf_entries:
+            if subfolder is not None:
+                raise ValueError("DDUF file only allow for 1 level of directory. Please check the DDUF structure")
```
sayakpaul, line 365: Can we please supplement this with an example in the comments?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -672,6 +678,7 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
         offload_state_dict = kwargs.pop("offload_state_dict", False)
         low_cpu_mem_usage = kwargs.pop("low_cpu_mem_usage", _LOW_CPU_MEM_USAGE_DEFAULT)
         variant = kwargs.pop("variant", None)
+        dduf_file = kwargs.pop("dduf_file", None)
```
sayakpaul, line 681: This needs to be also version-guarded with a `transformers` version, right?

src/diffusers/utils/hub_utils.py
```diff
@@ -291,9 +292,23 @@ def _get_model_file(
...
     pretrained_model_name_or_path = str(pretrained_model_name_or_path)
-    if os.path.isfile(pretrained_model_name_or_path):
+
+    if dduf_entries:
+        if subfolder is not None:
+            raise ValueError("DDUF file only allow for 1 level of directory. Please check the DDUF structure")
+        model_file = (
+            weights_name
+            if pretrained_model_name_or_path == ""
+            else "/".join([pretrained_model_name_or_path, weights_name])
```
sayakpaul, line 305:
```suggestion
            else os.path.join(pretrained_model_name_or_path, weights_name)
```

sayakpaul (commented):
I think we're close to merge for this one. Thank you!

sayakpaul:
@stevhliu @SunMarc I have added basic docs. LMK.

src/diffusers/configuration_utils.py
Wauplin: in a DDUF file, one should always use `/` separator (same as on the Hub)

src/diffusers/utils/hub_utils.py
Wauplin: same as above https://github.com/huggingface/diffusers/pull/10037#discussion_r1883660992

src/diffusers/pipelines/pipeline_utils.py
Wauplin: or update the lower-bound in `setup.py` ?

docs/source/en/using-diffusers/other-formats.md
```diff
@@ -240,6 +240,50 @@ Benefits of using a single-file layout include:
...
+pipe = DiffusionPipeline.from_pretrained(
+    "DDUF/FLUX.1-dev-DDUF", dduf_file="FLUX.1-dev.dduf", torch_dtype=torch.bfloat16
+).to("cuda")
+image = pipe(
+    "photo a cat holding a sign that says Diffusers", num_inference_steps=50, guidance_scale=3.5
+).images[0]
+image.save("cat.png")
+```
+
+To obtain `.dduf` checkpoint, we rely on `huggingface_hub`'s `export_folder_as_dduf()` utility, which takes care of all the necessary file-level validations:
```
Wauplin, line 268:
```suggestion
To save a pipeline as a `.dduf` checkpoint, we rely on `huggingface_hub`'s `export_folder_as_dduf()` utility, which takes care of all the necessary file-level validations:
```

(since we can "obtain" a checkpoint by downloading it)

src/diffusers/pipelines/pipeline_utils.py
sayakpaul: That works too.

src/diffusers/configuration_utils.py
SunMarc: You mean add an example of path that doesn't work ?

src/diffusers/pipelines/pipeline_utils.py
SunMarc: I added a version guard a bit later in the pipeline. Will this be enough or you prefer to raise an error earlier ?

SunMarc: I prefer not updating the lower-bound of transformers if possible.

tests/pipelines/test_pipelines.py
SunMarc: Thanks !

src/diffusers/configuration_utils.py
sayakpaul: Exactly.

src/diffusers/pipelines/pipeline_utils.py
sayakpaul: I think we should raise errors as early as possible.

src/diffusers/pipelines/transformers_loading_utils.py
SunMarc: Yeah it will be better, I fixed it

docs/source/en/using-diffusers/other-formats.md
```diff
@@ -240,6 +240,50 @@ Benefits of using a single-file layout include:
 1. Easy compatibility with diffusion interfaces such as [ComfyUI](https://github.com/comfyanonymous/ComfyUI) or [Automatic1111](https://github.com/AUTOMATIC1111/stable-diffusion-webui) which commonly use a single-file layout.
 2. Easier to manage (download and share) a single file.
 
+### DDUF
+
+<Tip warning={true}>
+
+DDUF is an experimental file format and APIs related to it can change in the future.
+
+</Tip>
```
stevhliu, line 249: Prefer using this syntax style now since its shorter/easier :)

```suggestion
> [!WARNING]
>DDUF is an experimental file format and APIs related to it can change in the future.
```

docs/source/en/using-diffusers/other-formats.md
```diff
@@ -240,6 +240,50 @@ Benefits of using a single-file layout include:
...
+
+DDUF, aka (**D**DUF’s **D**iffusion **U**nified **F**ormat) is a file format designed to make storing, distributing, and using diffusion models much easier. Built on the ZIP file format, DDUF offers a standardized, efficient, and flexible way to package all parts of a diffusion model into a single, easy-to-manage file. It tries to provide a sweet spot between our multi-folder format and widely popular single-file format. To learn more about it, please check out the documentation [here](https://huggingface.co/docs/hub/dduf).
```
stevhliu, line 251:
```suggestion
DDUF (**D**DUF **D**iffusion **U**nified **F**ormat) is a file format designed to make storing, distributing, and using diffusion models much easier. Built on the ZIP file format, DDUF offers a standardized, efficient, and flexible way to package all parts of a diffusion model into a single, easy-to-manage file. It provides a balance between Diffusers multi-folder format and the widely popular single-file format.

Learn more details about DDUF on the Hugging Face Hub [documentation](https://huggingface.co/docs/hub/dduf).
```

docs/source/en/using-diffusers/other-formats.md
```diff
@@ -240,6 +240,50 @@ Benefits of using a single-file layout include:
...
+
+Below we show, how to load a DDUF checkpoint in a [`DiffusionPipeline`]:
```
stevhliu, line 253:
```suggestion
Pass a checkpoint to the `dduf_file` parameter to load it in [`DiffusionPipeline`].
```

docs/source/en/using-diffusers/other-formats.md
```diff
@@ -240,6 +240,50 @@ Benefits of using a single-file layout include:
...
+pipe = DiffusionPipeline.from_pretrained(
+    "DDUF/FLUX.1-dev-DDUF", dduf_file="FLUX.1-dev.dduf", torch_dtype=torch.bfloat16
+).to("cuda")
+image = pipe(
+    "photo a cat holding a sign that says Diffusers", num_inference_steps=50, guidance_scale=3.5
+).images[0]
+image.save("cat.png")
+```
+
+To save a pipeline as a `.dduf` checkpoint, we rely on `huggingface_hub`'s `export_folder_as_dduf()` utility, which takes care of all the necessary file-level validations:
```
stevhliu, line 268:
```suggestion
To save a pipeline as a `.dduf` checkpoint, use the [`~huggingface_hub.export_folder_as_dduf`] utility, which takes care of all the necessary file-level validations.
```

docs/source/en/using-diffusers/other-formats.md
```diff
@@ -240,6 +240,50 @@ Benefits of using a single-file layout include:
...
+
+save_folder = "flux-dev"
+pipe.save_pretrained("flux-dev")
+export_folder_as_dduf("flux-dev.dduf", folder_path=save_folder)
+
+<Tip>
+
+We support packaging and loading quantized checkpoints in the DDUF format as long as they respect the multi-folder structure.
+
+</Tip>
```
stevhliu, line 285:
```suggestion
> [!TIP]
> Packaging and loading quantized checkpoints in the DDUF format is supported as long as they respect the multi-folder structure.
```

stevhliu (approved):
Nice, thanks for adding!

sayakpaul:
@DN6 @yiyixuxu I think PR is ready to be reviewed. @SunMarc is away for this week. So, expect delay in addressing comments that are more involved. I think I can address / comment on some.

src/diffusers/models/modeling_utils.py
```diff
@@ -808,7 +811,8 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
 
             model = load_flax_checkpoint_in_pytorch_model(model, model_file)
         else:
-            if is_sharded:
+            # in the case it is sharded, we have already the index
+            if is_sharded and not dduf_entries:
```
yiyixuxu, line 815: do we support dduf_entries for sharded checkpoints? 
the`is_sharded` logic (on line 786) depends on whether the `index_file` exists in the filesystem, which is just never going to be `True` for dduf because all the files are contained in the `dduf_entires`  - it is unclear to me whether this is intended or not

yiyixuxu (commented):
thanks for working on this! @SunMarc! 
I left a question about shared dduf checkpoint, looking great otherwise and hope to see this merged soon:)

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1453,10 +1488,14 @@ def download(cls, pretrained_model_name, **kwargs) -> Union[str, os.PathLike]:
                 return snapshot_folder
 
         user_agent = {"pipeline_class": cls.__name__}
-        if custom_pipeline is not None and not custom_pipeline.endswith(".py"):
+        if not dduf_file and custom_pipeline is not None and not custom_pipeline.endswith(".py"):
```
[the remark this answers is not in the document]
SunMarc: cc @sayakpaul as you added this part

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1313,7 +1342,14 @@ def download(cls, pretrained_model_name, **kwargs) -> Union[str, os.PathLike]:
                 local_files_only = True
                 model_info_call_error = e  # save error to reraise it if model is not cached locally
 
-        if not local_files_only:
+        if (
```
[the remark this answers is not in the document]
SunMarc: cc @sayakpaul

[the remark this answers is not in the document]
sayakpaul: Yeah decoupling these into different methods like suggested is a nice idea.

src/diffusers/pipelines/pipeline_utils.py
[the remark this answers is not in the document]
sayakpaul: Yeah we should throw an error when `custom_pipeline` and `dduf_file` both are passed.

src/diffusers/models/modeling_utils.py
sayakpaul: Hmm, the loading refactoring (https://github.com/huggingface/diffusers/issues/10013) is turning out to be more and more urgent which @DN6 wants to tackle. If we can make `dduf_entries` distinct from `is_sharded` in this PR, I think that might be a better option to explore for the minimal DDUF PoC?

src/diffusers/pipelines/pipeline_utils.py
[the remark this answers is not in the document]
sayakpaul: Done in 6a163c7d3.

vladmandic:
one slightly offtopic note - if we're coming up with new format, lets make sure it covers the painpoints of existing ones, otherwise its-just-another-standard.
as is `dduf` does have benefits, but it would be great if it could cover one of the most common use-cases: partial models.
e.g.
- typical sdxl model in single-file format includes unet and vae, but te1 and te2 are up to user to load.
- same sdxl model in diffusers folder-style format includes all components, but that creates significant duplication of storage
why not have model config that can point each model component not just to subfolder, but to other repo as well.
so for example, sdxl model could have unet-only with links to vae, te1, te2 in reference repo if they are unchanged.

for example:
```json
{
  "_class_name": "StableDiffusionXLPipeline",
  "scheduler": ["diffusers", "EulerDiscreteScheduler"],
  "unet": ["diffusers", "UNet2DConditionModel"],
  "vae": ["diffusers", "AutoencoderKL"],
  "text_encoder": ["transformers", "CLIPTextModel", "location": "stabilityai/stable-diffusion-xl-base-1.0/text_encoder"],
  "text_encoder_2": ["transformers", "CLIPTextModelWithProjection", "stabilityai/stable-diffusion-xl-base-1.0/text_encoder_2"],
  "tokenizer": ["transformers", "CLIPTokenizer", "location": "stabilityai/stable-diffusion-xl-base-1.0/tokenizer"],
  "tokenizer_2": ["transformers", "CLIPTokenizer", "location": "stabilityai/stable-diffusion-xl-base-1.0/tokenizer_2"]
}
```

src/diffusers/pipelines/pipeline_utils.py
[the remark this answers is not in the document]
sayakpaul: @yiyixuxu we have taken care of it. Let us know if the current changes make sense.

sayakpaul:
@SunMarc in the recent commits, I have:

* Taken care of Yiyi's comment except for [this one](https://github.com/huggingface/diffusers/pull/10037#discussion_r1898162330). Leaving that to you. 
* Added more tests and properly skipped a couple where DDUF isn't supported. 
* Instead of `require_*` markers, I have used `xfail` and conditioned it with the right versions of HF Hub and `transformers`, instead.

PTAL at the changes and let me know your thoughts. Even though I have run the tests myself but I would appreciate if you could run them as well just to be sure.

src/diffusers/models/modeling_utils.py
SunMarc: For now, I used the `_merge_sharded_checkpoints` path just like how we do for quantized model cc @sayakpaul . But after the refactoring in a follow-up PR, we will be able to make it a lot simpler.

src/diffusers/models/modeling_utils.py
```diff
@@ -818,10 +822,13 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
                     user_agent=user_agent,
                     revision=revision,
                     subfolder=subfolder or "",
+                    dduf_entries=dduf_entries,
                 )
                 # TODO: https://github.com/huggingface/diffusers/issues/10013
                 if hf_quantizer is not None:
```
sayakpaul, line 828: However, hf_quantizer can be None and we may have to merge the checkpoints in case `dduf_entries` is None, no?

sayakpaul (commented):
Thanks! Left a single comment.

src/diffusers/models/modeling_utils.py
SunMarc: this is fixed in the latest commit, i must have deleted the change when I fixed some conflicts !

tests/pipelines/test_pipelines.py
```diff
@@ -1812,6 +1875,18 @@ def test_wrong_model(self):
...
+    @require_transformers_version_greater("4.47.1")
+    def test_dduf_load_sharded_checkpoint_diffusion_model(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            pipe = DiffusionPipeline.from_pretrained(
+                "hf-internal-testing/tiny-flux-dev-pipe-sharded-checkpoint-DDUF",
+                dduf_file="tiny-flux-dev-pipe-sharded-checkpoint.dduf",
+                cache_dir=tmpdir,
+            ).to(torch_device)
+
+            pipe(prompt="dog", num_inference_steps=5, generator=torch.manual_seed(0), output_type="np").images
```
sayakpaul, line 1888: Here, I would:

1. save the `pipe` and load it.
2. run inference with the same params
3. compare the outputs to what we got from `pipe(prompt="dog", num_inference_steps=5, generator=torch.manual_seed(0), output_type="np").images`.

sayakpaul (approved):
Thanks for handling the final changes.

yiyixuxu:
@DN6 can you do a review here? it is a single file

src/diffusers/models/model_loading_utils.py
```diff
@@ -327,14 +340,19 @@ def _merge_sharded_checkpoints(sharded_ckpt_cached_folder, sharded_metadata):
     # Load tensors from each unique file
     for file_name in files_to_load:
         part_file_path = os.path.join(sharded_ckpt_cached_folder, file_name)
-        if not os.path.exists(part_file_path):
+        if not os.path.exists(part_file_path) and (dduf_entries and part_file_path not in dduf_entries):
```
yiyixuxu, line 343:
```suggestion
        if not os.path.exists(part_file_path) or (dduf_entries and part_file_path not in dduf_entries):
```
this seem to be `or` to me (because not it only throw an error for dduf) but let me know if not

src/diffusers/configuration_utils.py
```diff
@@ -616,6 +622,27 @@ def to_json_file(self, json_file_path: Union[str, os.PathLike]):
         with open(json_file_path, "w", encoding="utf-8") as writer:
             writer.write(self.to_json_string())
 
+    @classmethod
+    def _get_config_file_from_dduf(
+        cls, pretrained_model_name_or_path: str, subfolder: str, dduf_entries: Dict[str, DDUFEntry]
+    ):
+        if subfolder is not None:
```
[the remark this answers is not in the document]
SunMarc: removed !

tests/pipelines/test_pipelines.py
```diff
@@ -981,6 +983,18 @@ def test_download_ignore_files(self):
             assert not any(f in ["vae/diffusion_pytorch_model.bin", "text_encoder/config.json"] for f in files)
             assert len(files) == 14
 
+    def test_download_dduf_with_custom_pipeline_raises_error(self):
+        with self.assertRaises(NotImplementedError):
```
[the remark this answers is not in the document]
SunMarc: Maybe it is easier to set to True by default then set to to False for pipeline that doesn't work ? This helps finding which pipeline doesn't work with dduf and it makes sure that when adding a model, we try to make it compatible with dduf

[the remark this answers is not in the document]
SunMarc: LMK what you think about the last commit

yiyixuxu (approved):
thanks for working on this! @SunMarc

tests/pipelines/test_pipelines.py
```diff
@@ -1802,6 +1816,55 @@ def test_pipe_same_device_id_offload(self):
...
+            _ = DiffusionPipeline.from_pretrained(
+                "DDUF/tiny-flux-dev-pipe-dduf", dduf_file="fluxpipeline.dduf", custom_pipeline="my_pipeline"
+            )
+
+    def test_dduf_raises_error_with_connected_pipeline(self):
+        with self.assertRaises(NotImplementedError):
+            _ = DiffusionPipeline.from_pretrained(
+                "DDUF/tiny-flux-dev-pipe-dduf", dduf_file="fluxpipeline.dduf", load_connected_pipeline=True
+            )
```
sayakpaul, line 1866: Seems like these two are duplicates?

sayakpaul (approved):
Thanks, Marc!

tests/pipelines/test_pipelines.py
SunMarc: one is custom pipeline and the user is for connected pipeline I think !

DN6 (approved):
LGTM 👍🏽 Good to merge after resolving conflicts.

sayakpaul:
@SunMarc I proceeded with resolving the conflicts. LMK.

sayakpaul:
@SunMarc not sure why the test is failing. Could you give it a look?

src/diffusers/models/model_loading_utils.py
```diff
@@ -342,8 +342,12 @@ def _merge_sharded_checkpoints(
...
     for file_name in files_to_load:
         part_file_path = os.path.join(sharded_ckpt_cached_folder, file_name)
-        if not os.path.exists(part_file_path) or (dduf_entries and part_file_path not in dduf_entries):
-            raise FileNotFoundError(f"Part file {file_name} not found.")
+        if dduf_entries:
+            if part_file_path not in dduf_entries:
+                raise FileNotFoundError(f"Part file {file_name} not found.")
+        else:
+            if not os.path.exists(part_file_path):
+                raise FileNotFoundError(f"Part file {file_name} not found.")
```
SunMarc, line 350: This could be more concise with :
```python 
if part_file_path not in dduf_entries if dduf_entries else not os.path.exists(part_file_path):
    raise FileNotFoundError(f"Part file {file_name} not found.")
```
LMK what you prefer @sayakpaul

PR #10042, opened 2024-11-28
[CI] merge peft pr workflow into the main pr workflow.

sayakpaul:
# What does this PR do?

As discussed.


.github/workflows/pr_tests.yml
```diff
@@ -234,3 +234,75 @@ jobs:
...
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+
+    - name: Install dependencies
+      run: |
+        python -m venv /opt/venv && export PATH="/opt/venv/bin:$PATH"
+        python -m uv pip install -e [quality,test]
+        # TODO (sayakpaul, DN6): revisit `--no-deps`
+        if [ "${{ matrix.lib-versions }}" == "main" ]; then
```
DN6, line 269: Hmm wondering if we actually just run the test using the PEFT version on main since it allows us to catch issues from recent changes. If something breaks we can just manually check against latest. Would give us a bit of speed up on the PR CI.

sayakpaul:
@DN6 works for you now?

DN6 (approved):

PR #10051, opened 2024-11-29
Attention mask for transformer flux

### LINKED ISSUE #10025: attention mask for transformer Flux
### Describe the bug

Is it possible to get back the `attention_mask` argument in the flux attention processor 

```
hidden_states = F.scaled_dot_product_attention(query, key, value, dropout_p=0.0, is_causal=False,attn_mask=attention_mask)
```

https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention_processor.py#L1910

in order to tweak things a bit ? otherwise the argument  `attention_mask`  is unused.

Thanks a lot 

### Reproduction

pip install diffusers

### Logs

_No response_

### System Info

Ubuntu

### Who can help?

@yiyixuxu @sayakpaul @DN6 @asomoza

christopher5106:
Unused params `attention_mask` in transformer flux attention processors is useful to reweight prompt or image prompt.

Fixes #10025 

@yiyixuxu @sayakpaul

### DIFF
--- a/src/diffusers/models/attention_processor.py
+++ b/src/diffusers/models/attention_processor.py
@@ -1762,7 +1762,9 @@ class FluxAttnProcessor2_0:
             query = apply_rotary_emb(query, image_rotary_emb)
             key = apply_rotary_emb(key, image_rotary_emb)
 
-        hidden_states = F.scaled_dot_product_attention(query, key, value, dropout_p=0.0, is_causal=False)
+        hidden_states = F.scaled_dot_product_attention(
+            query, key, value, dropout_p=0.0, is_causal=False, attn_mask=attention_mask,
+        )
         hidden_states = hidden_states.transpose(1, 2).reshape(batch_size, -1, attn.heads * head_dim)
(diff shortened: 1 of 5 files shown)


src/diffusers/models/attention_processor.py
```diff
@@ -1762,7 +1762,9 @@ def __call__(
             query = apply_rotary_emb(query, image_rotary_emb)
             key = apply_rotary_emb(key, image_rotary_emb)
 
-        hidden_states = F.scaled_dot_product_attention(query, key, value, dropout_p=0.0, is_causal=False)
+        hidden_states = F.scaled_dot_product_attention(
```
yiyixuxu, line 1765: i see you added this to F.scaled_dot_product_attention
but just wonder how would this help? it is currently not passing down from the flux transformer to attenton_processor https://github.com/huggingface/diffusers/blob/c96bfa5c80eca798d555a79a491043c311d0f608/src/diffusers/models/attention.py#L190

PR #10053, opened 2024-11-29
use attention mask parameter in flux attention

### LINKED ISSUE #10025: attention mask for transformer Flux

christopher5106:
Unused params `attention_mask` in transformer flux attention processors is useful to reweight prompt or image prompt.

Fixes #10025 

@yiyixuxu @sayakpaul

### DIFF
--- a/scripts/convert_flux_to_diffusers.py
+++ b/scripts/convert_flux_to_diffusers.py
@@ -31,7 +31,7 @@ python scripts/convert_flux_to_diffusers.py  \
 --vae
 """
 
-CTX = init_empty_weights if is_accelerate_available else nullcontext
+CTX = init_empty_weights if is_accelerate_available() else nullcontext
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--original_state_dict_repo_id", default=None, type=str)
--- a/src/diffusers/models/attention_processor.py
+++ b/src/diffusers/models/attention_processor.py
@@ -1907,7 +1907,9 @@ class FluxAttnProcessor2_0:
             query = apply_rotary_emb(query, image_rotary_emb)
             key = apply_rotary_emb(key, image_rotary_emb)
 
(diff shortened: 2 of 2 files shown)


bghira:
this is incomplete. more changes are required to make use of the text encoder attn masking since it's being combined with the image tokens.

bonlime:
Thumbs up to this request. Diffusers is not only a set of pipelines, but also a library of use-full components, even if currently there is no diffusers pipeline that uses attention masking for flux, it doesn't mean it shouldn't be passed. I personally have a flux pipeline that requires passing attention mask and currently had to copy-paste existing attn processor only to start using this argument

bghira:
i have code to do attn masking in flux and would like it to have feature parity with what Diffusers will do. hence this is incomplete.there is more than just passing args in using generic parameter names. you can use image_attention_mask and encoder_attention_mask and fill with ones if encoder mask is apllied butnimage mask isnt.

however i think more importantly there is a fundamental misapplication of attention masking happening here, because it is not for re-weighting inputs.

bonlime:
This has been closed by #10122

bghira:
i'll just keep mentioning that it's an incomplete fix that has been merged and that this is not yet resolved

PR #10065, opened 2024-12-01
Fix a bug for SD35 control net training and improve control net block index

linjiapro:
# What does this PR do?

In SD3.5, ControlNet uses a single-layer transformer and does not rely on encoder_hidden_states. The code has been updated accordingly to prevent crashes.

For SD3.5 Large, which consists of 38 transformer layers, setting the ControlNet layers to 16 layers caused inefficiencies during inference. This was due to the old index logic calculating interval_control as an integer 3, which resulted in the 38-layer SD3.5 Large transformer using only up to layer 13 of the 16-layer ControlNet. To address this, the indexing logic has been revised to calculate interval_control as a floating-point number, ensuring all 16 layers of ControlNet are utilized. This update is backward-compatible, meaning no changes occur for a 19-layer ControlNet.

## Before submitting

## Who can review?


linjiapro:
cc @yiyixuxu @sayakpaul

src/diffusers/models/transformers/transformer_sd3.py
```diff
@@ -424,8 +424,7 @@ def custom_forward(*inputs):
             # controlnet residual
             if block_controlnet_hidden_states is not None and block.context_pre_only is False:
                 interval_control = len(self.transformer_blocks) / len(block_controlnet_hidden_states)
-                interval_control = int(np.ceil(interval_control))
-                hidden_states = hidden_states + block_controlnet_hidden_states[index_block // interval_control]
+                hidden_states = hidden_states + block_controlnet_hidden_states[int(index_block / interval_control)]
```
linjiapro, line 427: Please note that this is a change that is backward compatible. 

When len(self.transformer_blocks) is divisible by len(block_controlnet_hidden_states), which is 100% cases for all shipped controlnets, the logic does not change anything.

sayakpaul:
Thanks for the PR. Could you also add some tests to the PR?

src/diffusers/models/transformers/transformer_sd3.py
yiyixuxu: ok! but why do we have to change then?

yiyixuxu:
thanks!
can you `make style`?

linjiapro:
@yiyixuxu make style is done!

yiyixuxu (approved):

yiyixuxu:
thank you @linjiapro !

Issue #984, opened 2022-10-25
`F.interpolate(hidden_states, scale_factor=2.0, mode="nearest")` breaks for large bsz


NouamaneTazi:
### Describe the bug

Thanks to the amazing work done in the [memory efficient PR](https://github.com/huggingface/diffusers/pull/532), I can now run Stable Diffusion in fp16, on TITAN RTX (24Go VRAM) until a batch size of 31 with no issue.
```python
pipe = StableDiffusionPipeline.from_pretrained(
    "CompVis/stable-diffusion-v1-4", 
    use_auth_token=True,
    revision="fp16",
    torch_dtype=torch.float16,
).to("cuda")

batch_size = 32

with torch.inference_mode():
    image = pipe([prompt] * batch_size, num_inference_steps=5).images[0]
```
When I try a batch size of 32, I get the following error:
```
Traceback (most recent call last):
  File "/home/nouamane/projects/diffusers/a.py", line 45, in <module>
    image = pipe([prompt] * batch_size, num_inference_steps=5).images[0]
  File "/home/nouamane/miniconda3/envs/dev/lib/python3.9/site-packages/torch/autograd/grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "/home/nouamane/projects/diffusers/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py", line 353, in __call__
    image = self.vae.decode(latents).sample
  File "/home/nouamane/projects/diffusers/src/diffusers/models/vae.py", line 577, in decode
    dec = self.decoder(z)
  File "/home/nouamane/miniconda3/envs/dev/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/nouamane/projects/diffusers/src/diffusers/models/vae.py", line 217, in forward
    sample = up_block(sample)
  File "/home/nouamane/miniconda3/envs/dev/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/nouamane/projects/diffusers/src/diffusers/models/unet_blocks.py", line 1281, in forward
    hidden_states = upsampler(hidden_states)
  File "/home/nouamane/miniconda3/envs/dev/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
[...]

PR #10067, opened 2024-12-01
Interpolate fix on cuda for large output tensors

### LINKED ISSUE #10040: [BUG - STABLE DIFFUSION 3] Grey images generated
### Describe the bug

I'm running the SD3 model: [stabilityai/stable-diffusion-3-medium](https://huggingface.co/stabilityai/stable-diffusion-3-medium) with the following settings:
- Height: 1024
- Width: 1024
- Inference steps: 50
- Guidance scale: 7
- Batch size: 32 (using the same input prompt: "A men jumps from a high building")

The model generates 32 images, but half of them are grey.

### Reproduction

Reproduction code:

```
import os

os.environ['CUDA_VISIBLE_DEVICES'] = '7'

import torch

base_dir = os.path.join(os.path.dirname(__file__), '..', '..')

def run():
    input_prompts = ["A men jumps from a high building"] * 32
    img_save_dir = f'{base_dir}/data/test_generated_img'
    os.makedirs(f'{img_save_dir}', exist_ok=True)

    pipe = StableDiffusion3Pipeline.from_pretrained("stabilityai/stable-diffusion-3-medium-diffusers", torch_dtype=torch.float16, cache_dir="/data0/tien/cache")
    pipe.to("cuda")

    torch.set_grad_enabled(False)
    images = pipe(
        prompt=input_prompts,
        negative_prompt="",
        height=1024,
        width=1024,
        num_inference_steps=50,
        guidance_scale=7.0
    ).images

    for j in range(len(input_prompts)):
        images[j].save(os.path.join(f'{img_save_dir}', f'{j}.jpg'))

    torch.cuda.empty_cache()
    
run()
```

### Logs

_No response_

### System Info

- OS: Ubuntu 22.04.3 LTS
- Python: 3.10.15
- diffusers: 0.31.0

### Who can help?

_No response_

pcuenca:
Fixes #10040.

As described in https://github.com/pytorch/pytorch/issues/141831, there is a silent bug in PyTorch (CUDA) that causes the result of the upsampling operation to be wrong when the output size is beyond a threshold.

## Before submitting

cc @sayakpaul @asomoza as they are involved in #10040. In addition to that issue, I think the silent failure mode could come up in other cases as well and would be difficult to track.

### DIFF
--- a/src/diffusers/models/upsampling.py
+++ b/src/diffusers/models/upsampling.py
@@ -167,7 +167,9 @@ class Upsample2D(nn.Module):
         if self.interpolate:
             # upsample_nearest_nhwc also fails when the number of output elements is large
             # https://github.com/pytorch/pytorch/issues/141831
-            scale_factor = 2 if output_size is None else max([f/s for f,s in zip(output_size, hidden_states.shape[-2:])])
+            scale_factor = (
+                2 if output_size is None else max([f / s for f, s in zip(output_size, hidden_states.shape[-2:])])
+            )
             if hidden_states.numel() * scale_factor > pow(2, 31):
                 hidden_states = hidden_states.contiguous()
 
--- a/src/diffusers/models/upsampling.py
+++ b/src/diffusers/models/upsampling.py
@@ -167,7 +167,7 @@ class Upsample2D(nn.Module):
         if self.interpolate:
             # upsample_nearest_nhwc also fails when the number of output elements is large
             # https://github.com/pytorch/pytorch/issues/141831
-            scale_factor = 2 if output_size is None else max(output_size)
+            scale_factor = 2 if output_size is None else max([f/s for f,s in zip(output_size, hidden_states.shape[-2:])])
             if hidden_states.numel() * scale_factor > pow(2, 31):
                 hidden_states = hidden_states.contiguous()
 
--- a/src/diffusers/models/upsampling.py
+++ b/src/diffusers/models/upsampling.py
@@ -165,6 +165,12 @@ class Upsample2D(nn.Module):
         # if `output_size` is passed we force the interpolation output
         # size and do not make use of `scale_factor=2`
         if self.interpolate:
+            # upsample_nearest_nhwc also fails when the number of output elements is large
+            # https://github.com/pytorch/pytorch/issues/141831
+            scale_factor = 2 if output_size is None else max(output_size)
+            if hidden_states.numel() * scale_factor > pow(2, 31):
+                hidden_states = hidden_states.contiguous()
+
             if output_size is None:
                 hidden_states = F.interpolate(hidden_states, scale_factor=2.0, mode="nearest")
             else:


src/diffusers/models/upsampling.py
```diff
@@ -165,6 +165,14 @@ def forward(self, hidden_states: torch.Tensor, output_size: Optional[int] = None
         # if `output_size` is passed we force the interpolation output
         # size and do not make use of `scale_factor=2`
         if self.interpolate:
+            # upsample_nearest_nhwc also fails when the number of output elements is large
+            # https://github.com/pytorch/pytorch/issues/141831
+            scale_factor = (
+                2 if output_size is None else max([f / s for f, s in zip(output_size, hidden_states.shape[-2:])])
```
sayakpaul, line 171: `[f / s for f, s in zip(output_size, hidden_states.shape[-2:])]`

-> this could be unpacked into a variable with a sensible name and reused?

src/diffusers/models/upsampling.py
```diff
@@ -165,6 +165,14 @@ def forward(self, hidden_states: torch.Tensor, output_size: Optional[int] = None
...
+            )
+            if hidden_states.numel() * scale_factor > pow(2, 31):
```
sayakpaul, line 173: Consider keeping `pow(2, 31)` in a variable and then using it.

sayakpaul (commented):
Thanks, Pedro. The fix indeed looks like 🧠

I think this should be tested as well. Could we add a test here?

src/diffusers/models/upsampling.py
```diff
@@ -165,6 +165,14 @@ def forward(self, hidden_states: torch.Tensor, output_size: Optional[int] = None
         # if `output_size` is passed we force the interpolation output
         # size and do not make use of `scale_factor=2`
         if self.interpolate:
+            # upsample_nearest_nhwc also fails when the number of output elements is large
```
sayakpaul, line 168: Do we need to check for `channels_first` or `channels_last` memory layout?

pcuenca:
> I think this should be tested as well. Could we add a test here?

Sure, I can add some tests. They will require GPU and a good amount of GPU RAM, but I think they should fit in the CI environment.

src/diffusers/models/upsampling.py
pcuenca: I'm not sure, adding a variable that is only used when `output_size` is provided (which is not the usual case) can be more confusing, in my opinion, and I don't think we should complicate things further with an `if` branch. Also the value will never be reused. Perhaps I can add another line to the comment instead.

src/diffusers/models/upsampling.py
pcuenca: I feel that `pow(2, 31)` is easy to understand and provides explicit documentation about what's happening. Using another level of indirection wouldn't help, unless we are concerned about performance, in which case I can use a constant.

src/diffusers/models/upsampling.py
pcuenca: My assumption is it will behave the same way for any non-contiguous layouts, and that calling `.contiguous()` on a contiguous tensor will be a no-op. I'm also mimicking the same structure as in [lines 161-163 above](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/upsampling.py#L161-L163).

src/diffusers/models/upsampling.py
sayakpaul: I don't see how assigning `max([f / s for f, s in zip(output_size, hidden_states.shape[-2:])])` to a variable is not a better thing to do here as far as readability is concerned. But I am not going sweat over it, either. So, up to you.

src/diffusers/models/upsampling.py
sayakpaul: Along the same lines as https://github.com/huggingface/diffusers/pull/10067#discussion_r1865770022.

sayakpaul:
> Sure, I can add some tests. They will require GPU and a good amount of GPU RAM, but I think they should fit in the CI environment.

Do you have an estimate on how much? 16GB shouldn't be enough?

src/diffusers/models/upsampling.py
pcuenca: Readability is important to me, I'm happy to adapt to any recommendations! To be more specific, these are the options I see:

This PR:
```
        if self.interpolate:
            scale_factor = (
                2 if output_size is None else max([f / s for f, s in zip(output_size, hidden_states.shape[-2:])])
            )
```

With a var:
```
        if self.interpolate:
             if output_size is not None:
                 max_scale_factor = max([f / s for f, s in zip(output_size, hidden_states.shape[-2:])]
            scale_factor = 2 if output_size is None else max_scale_factor
```

Or:
```
        if self.interpolate:
             if output_size is None:
                  scale_factor = 2
              else:
                 scale_factor = max([f / s for f, s in zip(output_size, hidden_states.shape[-2:])]
```

I don't like the long expression in this PR, but I find it more difficult to understand the flow at a glance in the second alternative with the branches. The third one looks less Pythonic to me.

Happy to use what the team prefers :)

sayakpaul: Thanks for the consideration. I would prefer the third option. We don't go strictly Pythonic some times to have more explicit nature. But will let others comment,

yiyixuxu (approved):
thanks for the fix!
is this same bug as https://github.com/huggingface/diffusers/issues/984 (that's been there for 2 years?)

yiyixuxu:
also I'm cool with the code as it is! 
for the test @sayakpaul do you want to help add one?

PR #10081, opened 2024-12-02
Add `sigmas` to Flux pipelines

hlky:
# What does this PR do?

This PR adds `sigmas` argument to Flux pipelines and removes `timesteps` argument as passing `timesteps` to `FlowMatch` schedulers is currently unsupported.

Fixes #10080

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@yiyixuxu @asomoza

PR #10088, opened 2024-12-02
Apply `np.array` to provided sigmas in Flux pipelines

hlky:
# What does this PR do?

https://github.com/huggingface/diffusers/pull/10081#issuecomment-2512995300

This PR applies `np.array` to provided sigmas as this was tested with `np.ndarray` but Pipeline input for `sigmas` is supposed to be `list` according to type hints. 

## Who can review?

@yiyixuxu @vladmandic


yiyixuxu:
that's really fast!! :rolling_on_the_floor_laughing:
can we also do this on `scheduler.set_timesteps` too?  (we only have a few that support custom sigmas)

yiyixuxu (approved):

hlky:
Thanks! I've changed it to apply in `set_timesteps` like in non-FlowMatch
https://github.com/huggingface/diffusers/blob/6db33337a49c2f20db1b8d2ad069cca10c552c68/src/diffusers/schedulers/scheduling_euler_discrete.py#L372

Issue #10080, opened 2024-12-02
Proposal to add sigmas option to FluxPipeline


John6666cat:
**Is your feature request related to a problem? Please describe.**
Flux's image generation is great, but it seems to have a tendency to remove too much detail.

**Describe the solution you'd like.**
Add the `sigmas` option to `FluxPipeline` to enable adjustment of the degree of noise removal.

**Additional context.**
I propose adding the `sigmas` option to `FluxPipeline`.
The details are as follows.

[Yntec](https://huggingface.co/Yntec) picked up a FLUX modification proposal from [Reddit](https://www.reddit.com/r/comfyui/comments/1g9wfbq/simple_way_to_increase_detail_in_flux_and_remove/), and [r3gm](https://huggingface.co/r3gm), the author of [stablepy](https://github.com/R3gm/stablepy), wrote the code for the logic part of the FLUX pipeline modification, and I made [a demo](https://huggingface.co/spaces/John6666/flux-sigmas-test) by making [a test commit on github](https://github.com/huggingface/diffusers/commit/ad3344e2be033887d854d2731757db8b80dcfb06), and it turned out that it was working as expected.

This time, we only modified the pipeline for T2I for testing. During discussions with r3gm, it was discovered that [the `sigmas` option, which exists in StableDiffusionXLPipeline](https://github.com/huggingface/diffusers/blob/c96bfa5c80eca798d555a79a491043c311d0f608/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L841), does not exist in `FluxPipeline`, so the actual implementation was switched to porting the `sigmas` option.
Also, in the current `FluxPipeline`, I also found a bug where specifying `timesteps` would probably result in an error because `sigmas` are hard-coded, even though the SDXL pipeline code is reused, so I fixed it while I was at it.

If you want to use Reddit's suggested **0.95**, specify it as follows.
```py
factor = 0.95
sigmas = np.linspace(1.0, 1 / num_inference_steps, num_inference_steps)
sigmas = sigmas * factor

image_sigmas = pipe(
        prompt=prompt,
        guidance_scale=guidance_scale,
        num_inference_steps=num_inference_steps,
        width=width,
        height=height,
        generator=generator,
        output_type="pil",
        sigmas=sigmas
    ).images[0]
```

I will post some samples that were actually generated in the demo.

Prompt: anthropomorphic pig Programmer with laptop, colorfull, funny / Seed: 9119
[...]

PR #10113, opened 2024-12-04
Test `timesteps` in `PipelineTesterMixin`

hlky:
# What does this PR do?

Companion to #10101. Adds `test_timesteps` to `PipelineTesterMixin`, some pipelines already test `timesteps`, useful to test all of them, for example this could have caught #10080 by raising that `timesteps` isn't supported by the scheduler then it might have been noticed that `sigmas` was unsupported in the pipeline.

As expected this test reveals pipelines that are not working with `timesteps` argument. This PR is created so everyone can refer to test results for the list of affected pipelines, as such this PR should not be merged in its current state, we will investigate each pipeline to determine the appropriate course of action (removing the unsupported argument, fixing an error, etc.)

## Who can review?


DN6:
Looks good. But I think we would have to adjust based on the scheduler (think you mentioned this already in #10101)

Issue #10112, opened 2024-12-04
Detail-Daemon diffusers


NicholasCao:
**Describe the solution you'd like.**
Detail-Daemon:  https://github.com/Jonseed/ComfyUI-Detail-Daemon

How to implement Detail-Daemon in diffusers, as seen in https://github.com/Jonseed/ComfyUI-Detail-Daemon. Will there be a better official component in the future?

hlky:
This should currently work by passing the custom sigmas that are created to pipelines with `sigmas` argument. We can consider this as an example for upcoming Modular Diffusers. We are also planning on separating "sampler"/Scheduler and "noise schedule"/sigmas, so this kind of functionality (modifying sigmas) could fit in there.

NicholasCao:
```
sigmas = np.linspace(1.0, 1 / num_inference_steps, num_inference_steps)
sigmas[10:20] = sigmas[10:20] * 0.95
pipeline(sigmas=sigmas, xxx)
```
So, it can be implemented this way in Diffusers?

NicholasCao:
@hlky got！thx！

blepping:
> Yes, see #10080 for a similar use case. Let me know if you encounter any pipelines that don't support passing sigmas.

Unfortunately this won't be enough (unless I'm misunderstanding, which is possible since I'm not a Diffusers person). Detail Daemon accomplishes its effect with a mismatch between the schedule used for sampling and the sigmas the model gets called with. For example, given `x` is current latent, `sigma` is the current sigma and `sigma_next` is the next sigma, a normal Euler step would look like:

```python
denoised = model(x, sigma)
dt = sigma_next - sigma
d = (x - denoised) / sigma
x_next = x + d * dt
```

and a "Detail Daemon" step would look like:

```python
dd_adjustment = -0.05
denoised = model(x, sigma * dd_adjustment)
dt = sigma_next - sigma
d = (x - denoised) / sigma
x_next = x + d * dt
```

If the user is able to replace the model with a wrapper function that performs the adjustment then that would probably be the easiest way to do it.

hlky:
Thanks @blepping! We'll be able to do that with modular Diffusers when its ready #9672 

cc @yiyixuxu for interest in another use case of modular

blepping:
> Thanks @blepping! We'll be able to do that with modular Diffusers when its ready #9672

[...]

PR #10116, opened 2024-12-04
Add `sigmas` to pipelines using FlowMatch

hlky:
# What does this PR do?

Part of the fixes for #10113.

Pipelines in this PR all use FlowMatch scheduler which does not currently support passing timesteps so the timesteps argument is non-functional, we replace timesteps argument with `sigmas` which is currently supported by the scheduler. See #10080 #10112 for example use case of passing `sigmas`.

## Who can review?

### DIFF
--- a/src/diffusers/pipelines/aura_flow/pipeline_aura_flow.py
+++ b/src/diffusers/pipelines/aura_flow/pipeline_aura_flow.py
@@ -387,7 +387,6 @@ class AuraFlowPipeline(DiffusionPipeline):
         prompt: Union[str, List[str]] = None,
         negative_prompt: Union[str, List[str]] = None,
         num_inference_steps: int = 50,
-        timesteps: List[int] = None,
         sigmas: List[float] = None,
         guidance_scale: float = 3.5,
         num_images_per_prompt: Optional[int] = 1,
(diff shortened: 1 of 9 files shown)


src/diffusers/pipelines/aura_flow/pipeline_aura_flow.py
```diff
@@ -387,7 +387,6 @@ def __call__(
         prompt: Union[str, List[str]] = None,
         negative_prompt: Union[str, List[str]] = None,
         num_inference_steps: int = 50,
-        timesteps: List[int] = None,
```
sayakpaul, line 390: Even though these are non-functional but isn't this a breaking change? If so, maybe we should deprecate (with a short cycle, for sure)?

sayakpaul (commented):
Thanks! My only comment is that this could be a breaking change and we might want to consider a deprecation. Or maybe that's a stretch.

src/diffusers/pipelines/aura_flow/pipeline_aura_flow.py
hlky: In #10116 for Flux pipelines we replaced `timesteps` with `sigmas` but it's a good point. Let's bring in @yiyixuxu for a second opinion.

yiyixuxu: it doesn't because the flow match scheduler does not accept `timesteps`

thanks for the fix @hlky!

yiyixuxu (approved):
thanks!

src/diffusers/pipelines/aura_flow/pipeline_aura_flow.py
sayakpaul: That makes sense.

PR #10122, opened 2024-12-04
pass attn mask arg for flux

yiyixuxu:
fix https://github.com/huggingface/diffusers/issues/10025#issuecomment-2518335274


dxqbYD:
This didn't fix the problem at all. See here: https://github.com/huggingface/diffusers/issues/10194

PR #10126, opened 2024-12-05
[LoRA] depcrecate save_attn_procs().

sayakpaul:
# What does this PR do?

Should have been done in https://github.com/huggingface/diffusers/pull/9712 but somehow got missed.

### DIFF
--- a/tests/models/unets/test_models_unet_2d_condition.py
+++ b/tests/models/unets/test_models_unet_2d_condition.py
@@ -1119,6 +1119,24 @@ class UNet2DConditionModelTests(ModelTesterMixin, UNetTesterMixin, unittest.Test
             lora_sample_1, lora_sample_2, atol=1e-4, rtol=1e-4
         ), "Loading from a saved checkpoint should produce identical results."
 
+    @require_peft_backend
+    def test_save_attn_procs_raise_warning(self):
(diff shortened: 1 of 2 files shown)


tests/models/unets/test_models_unet_2d_condition.py
```diff
@@ -1119,6 +1119,24 @@ def test_load_attn_procs_raise_warning(self):
...
+        model = self.model_class(**init_dict)
+        model.to(torch_device)
+
+        unet_lora_config = get_unet_lora_config()
+        model.add_adapter(unet_lora_config)
+
+        assert check_if_lora_correctly_set(model), "Lora not correctly set in UNet."
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            with self.assertWarns(FutureWarning) as warning:
```
BenjaminBossan, line 1134: You could use `assertWarnsRegex` to check the message.

BenjaminBossan (approved):
Thanks for adding the warning.

tests/models/unets/test_models_unet_2d_condition.py
sayakpaul: We use `assertWarns` throughout `diffusers`. So, just kept that for consistency.

PR #10134, opened 2024-12-05
Remove duplicate checks for len(generator) != batch_size when generator is a list

a-r-r-o-w:
Removes unnecessary duplicate checks on length of generator, as it is already checked in a previous condition


src/diffusers/pipelines/animatediff/pipeline_animatediff_video2video.py
```diff
@@ -662,12 +662,6 @@ def prepare_latents(
                 self.vae.to(dtype=torch.float32)
 
             if isinstance(generator, list):
-                if len(generator) != batch_size:
```
yiyixuxu, line 665: oh where is it checked?

a-r-r-o-w: Just a few lines above

image

yiyixuxu (approved):
thanks!

PR #10143, opened 2024-12-06
[Tests] update always test pipelines list.

sayakpaul:
# What does this PR do?

As discussed with Dhruv, it makes sense to update 
https://github.com/huggingface/diffusers/blob/18f9b990884883533491fc87f303e7305dc27d75/utils/fetch_torch_cuda_pipeline_test_matrix.py#L13

I don't think pipelines like DeepFloyd IF have usage high enough to be always tested.


DN6 (approved):

hlky (approved):
Model downloads on Hub for reference

Kandinsky 2.1	58k
Kandinsky 2.2	23k
wuerstchen	10k
DeepFloyd IF	20k-55k

sayakpaul:
Failing test is unrelated.

Issue #10158, opened 2024-12-09
Post processing performance can be improved


soof-golan:
## Problem

Images generated in batches pay a performance penalty in the post-processing step of the diffusion pipeline.

A lot of calls to image_processor.denormalize are made instead of batching the computation.

#### Using multiplication instead of division

This is a freebie, numerically the same but much cheaper on all compute platforms

```python
def denormalize(images: Union[np.ndarray, torch.Tensor]) -> Union[np.ndarray, torch.Tensor]:
    return (images * 0.5 + 0.5).clamp(0, 1)
```

#### Adding Fast path in an all-or-nothing denormalization scenario

Instead of calling `torch.stack` on a batch where all (or none) of the images require denormalization, apply the operation on the full tensor directly

```python
def post_process(...):
    # ... 
    if do_denormalize is None:
        return self.denormalize(images) if self.config.do_normalize else images
    # ...
```

#### Denormalize first, stack later

Invoking the `denormalize` call multiple times incurs a lot of overhead on the compute pipeline, more so than "wasting" compute of denormalizing everything and calling `torch.stack` on the results.

```python
def post_process(...):
    # ...
    denormalized = self.denormalize(image)
    image = torch.stack([
        denormalized[i] if do_denormalize[i] else image[i] for i in range(image.shape[0])
    ])
    # ...
```

## Benchmarks

https://colab.research.google.com/drive/1H1SKUlyEZduUeU50V8SVEZgmFrdl6Ego?usp=sharing

These were ran in a Colab T4 instance on CUDA. The batch is of the shape `[1024,3,512,512]` and dtype `fp16`. Further combinations can be made if requested.

### Baseline

<img width="745" alt="image" src="https://github.com/user-attachments/assets/3bd79c84-bce0-47f4-aae9-5ba86b77a2de">

### Suggested Improvements

<img width="729" alt="image" src="https://github.com/user-attachments/assets/3c739ea0-71e6-467d-b36a-e07e42593003">

yiyixuxu:
can you look into this? @hlky

PR #10159, opened 2024-12-09
Improve post-processing performance

soof-golan:
# What does this PR do?

Fixes: #10158 

* Use multiplication instead of division in `VaeImageProcessor.denormalize`
* Avoid splitting and re-stacking tensors to reduce memory bandwidth and CPU-GPU syncs

### DIFF
--- a/src/diffusers/image_processor.py
+++ b/src/diffusers/image_processor.py
@@ -236,7 +236,7 @@ class VaeImageProcessor(ConfigMixin):
             `np.ndarray` or `torch.Tensor`:
                 The denormalized image array.
         """
-        return (images / 2 + 0.5).clamp(0, 1)
+        return (images * 0.5 + 0.5).clamp(0, 1)
 
     @staticmethod
     def convert_to_rgb(image: PIL.Image.Image) -> PIL.Image.Image:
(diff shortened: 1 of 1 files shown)


soof-golan:
> Thanks for your contribution @soof-golan! Can you run make style?

@hlky Done!

hlky (approved):

src/diffusers/image_processor.py
```diff
@@ -537,6 +537,27 @@ def binarize(self, image: PIL.Image.Image) -> PIL.Image.Image:
...
+            do_denormalize (`Optional[List[bool]`, *optional*, defaults to `None`):
+                A list of booleans indicating whether to denormalize each image in the batch. If `None`, will use the
+                value of `do_normalize` in the `VaeImageProcessor` config.
+        """
+        if do_denormalize is None:
+            return self.denormalize(images) if self.config.do_normalize else images
+
+        # De-normalizing a batch and selectively torch.stack'ing the results turns out to be
+        # significantly faster than performing a lot of smaller denormalizations
+        denormalized = self.denormalize(images)
```
yiyixuxu, line 558: I think there is some context to that, each image in the batch may have a different value for `do_normalize` for sd1.5, see the code here https://github.com/huggingface/diffusers/blob/6131a93b969f87d171148bd367fd9990d5a49b6b/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py#L1075

Since most of the new pipelines (sdxl, sd3, flux), we do not pass the `do_normalize` from the pipeline, i.e. `do_normalize is None` here , see SDXL https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L1304 , 

and we already did this so it will be batch processed for all the new pipelines already, I think this line is sufficient here
```
        if do_denormalize is None:
            return self.denormalize(images) if self.config.do_normalize else images
```

src/diffusers/image_processor.py
```diff
@@ -752,12 +773,7 @@ def postprocess(
         if output_type == "latent":
             return image
 
-        if do_denormalize is None:
-            do_denormalize = [self.config.do_normalize] * image.shape[0]
-
-        image = torch.stack(
-            [self.denormalize(image[i]) if do_denormalize[i] else image[i] for i in range(image.shape[0])]
-        )
+        image = self._denormalize_conditionally(image, do_denormalize)
```
yiyixuxu, line 776:
```python

if do_denormalize is None:
            image = self.denormalize(images) if self.config.do_normalize 
else: 
        image = torch.stack(
            [self.denormalize(image[i]) if do_denormalize[i] else image[i] for i in range(image.shape[0])]
        )

soof-golan:
@yiyixuxu I've opened up #10170 with only the simpler optimization, leaving the homogeneous case untouched.

Choose whichever you see fit for the scope of the project :)

hlky:
Closed by https://github.com/huggingface/diffusers/pull/10170

PR #10168, opened 2024-12-10
[Single file] Support `revision` argument when loading single file config

a-r-r-o-w:
Revealed in #10135 that we can't load config from a provided revision. This is handy for testing single file changes on our weights PRs and has helped for both DCAE and LTX integration


src/diffusers/loaders/single_file_model.py
```diff
@@ -243,7 +243,7 @@ def from_single_file(cls, pretrained_model_link_or_path_or_dict: Optional[str] =
                 original_config=original_config, checkpoint=checkpoint, **config_mapping_kwargs
             )
         else:
-            if config:
+            if config is not None:
```
a-r-r-o-w, line 246: This branch should only be entered when no config is provided. If, for example, I were to pass `{}`, I want the model to be initialized with default parameters, but currently this does not seem possible.

DN6 (approved):

PR #10170, opened 2024-12-10
Improve post-processing performance

### LINKED ISSUE #10158: Post processing performance can be improved

soof-golan:
# What does this PR do?

Closes #10158 

* Use multiplication instead of division
* Add fast path when denormalizing all or none of the images

## Context

As suggested by @yiyixuxu, this is a smaller PR than #10159 and only adds the fast path in the all-or-nothing scenario. leaving the selective denormalization as-is in the heterogeneous case.

### DIFF
--- a/src/diffusers/image_processor.py
+++ b/src/diffusers/image_processor.py
@@ -236,7 +236,7 @@ class VaeImageProcessor(ConfigMixin):
             `np.ndarray` or `torch.Tensor`:
                 The denormalized image array.
         """
-        return (images / 2 + 0.5).clamp(0, 1)
+        return (images * 0.5 + 0.5).clamp(0, 1)
 
     @staticmethod
     def convert_to_rgb(image: PIL.Image.Image) -> PIL.Image.Image:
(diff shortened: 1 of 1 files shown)


yiyixuxu (approved):
thanks!

PR #10179, opened 2024-12-10
[community pipeline rf-inversion] - fix example in doc

linoytsaban:
fix inference example


hlky (approved):
Thanks @linoytsaban! We could use `load_image` from `diffusers.utils` here, the `download_image` is used in some examples though so either is ok.

linoytsaban:
thanks for the speedy response @hlky!

Issue #9950, opened 2024-11-18
Improve SD35 LoRA support to cover most popular LoRA formats


vladmandic:
SD3.x pipeline does implement `SD3LoraLoaderMixin` and as such `load_lora_weights` on SD3.x does "work".  

However, attempting to load any of the most popular LoRAs results in silent failure:  
load is successful without any warnings, but loads ZERO keys.

Looking at implementation at: <https://github.com/huggingface/diffusers/blob/345907f32de71c8ca67f3d9d00e37127192da543/src/diffusers/loaders/lora_pipeline.py#L1214>

Shows that it there are no matching keys in `text_encoder_state_dict`, `text_encoder_state_dict`, `text_encoder_2_state_dict`, there is no warning raised, the method simply does NOTHING.

When looking at `state_dict` of loaded LoRA, it shows that keys are NOT in the expected format - they need remapping.
For example:
```dict_keys(['lora_unet_joint_blocks_0_context_block_adaLN_modulation_1.alpha', 'lora_unet_joint_blocks_0_context_block_adaLN_modulation_1.lora_down.weight', 'lora_unet_joint_blocks_0_context_block_adaLN_modulation_1.lora_up.weight', 'lora_unet_joint_blocks_0_context_block_attn_proj.alpha' ...])```

Below are download links for some of the popular LoRAs for SD35, all experience same behavior:
- <https://civitai.com/models/943882/tokino-sd35-ambient-scene?modelVersionId=1056722>
- <https://civitai.com/models/539031/lineart-sd35fluxsdxlpony?modelVersionId=1052194>
- <https://civitai.com/models/948259/brian-sum-mech-lora-sd35-large?modelVersionId=1061663>

CC @yiyixuxu @sayakpaul @DN6 @asomoza

sayakpaul:
Ah I know why this is happening. 

It's because when we supported SD3 LoRA, there weren't popular non-diffusers LoRAs. But clearly, things have changed now. I will look into it.

bghira:
could you please ensure that the missing keys error out in a more generalised sense @sayakpaul ? that way this might not impact newer models when implementing their LoRA loaders

sayakpaul:
> could you please ensure that the missing keys error out in a more generalised sense 

I don't understand what that means.

sayakpaul:
[...]

PR #10187, opened 2024-12-11
[LoRA] Improve warning messages when LoRA loading becomes a no-op

sayakpaul:
# What does this PR do?

As discussed with Dhruv, this PR improves our logging for LoRAs when the specified state dict isn't an appropriate one, leaving the `load_lora_weights()`, `load_lora_into_text_encoder()`, and `load_lora_adapter()` methods essentially a no-op. 

https://github.com/huggingface/diffusers/issues/9950 mentions this problem, and I think it's better to at least let the user know that the specified LoRA state dict is either not a correct one (which we already do) or is an ineffective one (this PR).

This PR also standardizes a bit how we log these kinds of info across our LoRA loading utilities. 

Some comments are in-line.  

- [x] Run the integration tests for SD LoRA
- [x] Run the integration tests for SDXL LoRA
- [x] Run the integration tests for Flux LoRA (FluxControlLoRA integration tests will pass after https://github.com/huggingface/diffusers/pull/10397)
- [x] Run the integration tests for SD3 LoRA


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -294,22 +294,15 @@ def load_lora_into_unet(
...
-            # Load the layers corresponding to UNet.
-            logger.info(f"Loading {cls.unet_name}.")
-            unet.load_lora_adapter(
-                state_dict,
-                prefix=cls.unet_name,
-                network_alphas=network_alphas,
-                adapter_name=adapter_name,
-                _pipeline=_pipeline,
-                low_cpu_mem_usage=low_cpu_mem_usage,
-            )
```
sayakpaul, line 312: We're handling all of this within the `load_lora_adapter()` method now which I think is more appropriate as:

1. It takes care of logging when users try to load LoRA into a model via `pipe.load_lora_weights()`. 
2. Users try to load LoRAs directly into a model with the `load_lora_adapter()` method (with something like `unet.load_lora_adapter()`. 

Helps to avoid duplication. I have run the integration tests, too and nothing is breaking due to this.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -660,18 +658,16 @@ def load_lora_weights(
...
-            self.load_lora_into_text_encoder(
-                text_encoder_state_dict,
-                network_alphas=network_alphas,
-                text_encoder=self.text_encoder,
-                prefix="text_encoder",
-                lora_scale=self.lora_scale,
-                adapter_name=adapter_name,
-                _pipeline=self,
-                low_cpu_mem_usage=low_cpu_mem_usage,
-            )
```
sayakpaul, line 674: Similar philosophy as explained above.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -297,19 +297,15 @@ def load_lora_into_unet(
...
+        logger.info(f"Loading {cls.unet_name}.")
+        unet.load_lora_adapter(
+            state_dict,
+            prefix=cls.unet_name,
+            network_alphas=network_alphas,
+            adapter_name=adapter_name,
+            _pipeline=_pipeline,
+            low_cpu_mem_usage=low_cpu_mem_usage,
+        )
```
sayakpaul, line 308: In case `prefix` is not None and no `prefix` matched state dict keys are found, we log from the `load_lora_adapter()` method. 

This way, we cover for both `load_lora_weights()` which is pipeline-level and `load_lora_adapter()` which model-level.

yiyixuxu:
@hlky could you do a review here? let's merge if it looks good 
@sayakpaul sorry we missed this PR

hlky (approved):
Thanks @sayakpaul

sayakpaul:
Ran the integration tests, too. They are passing. Will merge once the CI is green.

BenjaminBossan:
@sayakpaul This PR broke LoRA loading. It seems like after calling `load_lora_adapter`, there is no LoRA being loaded. Reproducer:

```python
import os
import tempfile

import torch

from peft import LoraConfig
from peft.tuners.lora import LoraLayer
from diffusers import UNet2DConditionModel
from diffusers.utils.testing_utils import floats_tensor

torch_device = 0

class TestIt:
    def get_small_unet(self):
        # from diffusers UNet2DConditionModelTests
        init_dict = {
            "block_out_channels": (4, 8),
            "norm_num_groups": 4,
            "down_block_types": ("CrossAttnDownBlock2D", "DownBlock2D"),
            "up_block_types": ("UpBlock2D", "CrossAttnUpBlock2D"),
            "cross_attention_dim": 8,
            "attention_head_dim": 2,
            "out_channels": 4,
            "in_channels": 4,
            "layers_per_block": 1,
            "sample_size": 16,
        }
        model = UNet2DConditionModel(**init_dict)
        return model.to(torch_device)

    def test_it(self):
        lora_config = LoraConfig(
            target_modules=["to_q"],
            init_lora_weights=False,
        )
        unet = self.get_small_unet()
        unet.add_adapter(lora_config, adapter_name="adapter0")
        num_lora_before = sum(isinstance(m, LoraLayer) for m in unet.modules())
        assert num_lora_before > 0

        with tempfile.TemporaryDirectory() as tmp_dirname:
            unet.save_lora_adapter(tmp_dirname, safe_serialization=True, adapter_name="adapter0")
            del unet

            torch.manual_seed(0)
            unet = self.get_small_unet()
            file_name = os.path.join(tmp_dirname, "pytorch_lora_weights.safetensors")
            unet.load_lora_adapter(file_name, safe_serialization=True, adapter_name="adapter0")
            num_lora_after = sum(isinstance(m, LoraLayer) for m in unet.modules())

            # this fails, no LoRA layer is loaded
            assert num_lora_after > 0
```

The previous commit, `0703ce88008b2765ef6636c6e5cb013d227c42ca`, still works. I think this is quite urgent.

sayakpaul:
Use:

```
unet.load_lora_adapter(file_name, safe_serialization=True, prefix=None, adapter_name="adapter0")
```

This is reflected in the tests:
https://github.com/huggingface/diffusers/blob/733b44ac82193afc601421f0ca563132c627cb2a/tests/models/test_modeling_common.py#L1094

BenjaminBossan:
Thanks for the quick reply. Isn't this a backwards incompatible change on a public API? At least for me this was surprising, especially since there is no error message or warning.

sayakpaul:
Can you check if you're hitting?
https://github.com/huggingface/diffusers/blob/733b44ac82193afc601421f0ca563132c627cb2a/src/diffusers/loaders/peft.py#L357

If not, then there's something wrong that I can look into fixing.

BenjaminBossan:
> there's something wrong that I can look into

When I load without `prefix=None`, I hit this code path:

> "No LoRA keys associated to UNet2DConditionModel found with the prefix='transformer'. This is safe to ignore if LoRA state dict didn't originally have any UNet2DConditionModel related params. Open an issue if you think it's unexpected: https://github.com/huggingface/diffusers/issues/new"

sayakpaul:
Earlier you said

> At least for me this was surprising, especially since there is no error message or warning.

But you are indeed hitting the right message here. Would you suggest turning this into a warning?

BenjaminBossan:
Since `logger.info` is not shown by default, I had no clue about this message. At least to me, when I call

`unet.load_lora_adapter(file_name, safe_serialization=True, adapter_name="adapter0")`

but nothing happens, it would be useful to get a warning. Maybe there is a good reason not to have a warning because there are legitimate use cases, but it feels dangerous to me. As mentioned, everything appears to work on the surface, but the outputs are not what I expect, so this could lead to hard to debug silent errors in code that used to work.

Also, I don't find the message to be very helpful, since the solution you provided is not mentioned. But perhaps in the bigger context of things, this is obvious?

sayakpaul:
https://github.com/huggingface/diffusers/pull/11044

PR #10188, opened 2024-12-11
Test error raised when loading normal and expanding loras together in Flux

a-r-r-o-w:
Context: https://github.com/huggingface/diffusers/pull/10182#discussion_r1879533319


tests/lora/test_lora_layers_flux.py
```diff
@@ -430,6 +430,97 @@ def test_correct_lora_configs_with_different_ranks(self):
...
+        # The first lora expanded the input features of x_embedder. Here, we are trying to load a lora with the correct
+        # input features before expansion. This should raise an error about the weight shapes being incompatible.
+        self.assertRaisesRegex(
+            RuntimeError,
+            "size mismatch for x_embedder.lora_A.adapter-2.weight",
+            pipe.load_lora_weights,
+            lora_state_dict,
+            "adapter-2",
+        )
+
```
sayakpaul, line 478: Here I would run another inference round and make sure the outputs match with the LoRA that was correctly loaded. This will help us check if this loading error didn't leave the pipeline in a broken state, which is important.

tests/lora/test_lora_layers_flux.py
```diff
@@ -430,6 +430,97 @@ def test_correct_lora_configs_with_different_ranks(self):
...
+        # Test the opposite case where the first lora has the correct input features and the second lora has expanded input features.
+        # This should raise a runtime error on input shapes being incompatible. But it doesn't. This is because PEFT renames the
+        # original layers as `base_layer` and the lora layers with the adapter names. This makes our logic to check if a lora
+        # weight is compatible with the current model incorrect. This should be addressed when attempting support for
+        # https://github.com/huggingface/diffusers/issues/10180 (TODO)
```
sayakpaul, line 483: Could you provide some concrete LoCs as references for what you mean by:

> This makes our logic to check if a lora weight is compatible with the current model incorrect.

Would also love to understand how this relates to how `peft` names things.

a-r-r-o-w: The lines in question are: https://github.com/huggingface/diffusers/blob/d041dd504058ac6b0fde3eb767eb6844d8d577b8/src/diffusers/loaders/lora_pipeline.py#L2319-L2322

When the first lora layer is loaded, assuming it is named `adapter-1` and the layer in question is `x_embedder`, the nn.Linear layer names are something like [`x_embedder`]. [This](https://github.com/huggingface/diffusers/blob/d041dd504058ac6b0fde3eb767eb6844d8d577b8/src/diffusers/loaders/lora_pipeline.py#L2321) check passes because `x_embedder.lora_A.weight` is indeed a key in the lora state dict.

After the first lora is loaded, peft updates the layer names to something like: [`x_embedder.base_layer`, `x_embedder.adapter-1.lora_A`, `x_embedder.adapter-1.lora_B`]. 

So, when the second lora is loaded, it tries to find `x_embedder.base_layer.lora_A.weight` in the lora state dict, which does not exist. It needs to instead search for `x_embedder.lora_A.weight`. But this won't happen because the model state dict contains x_embedder original layers in a renamed key.

Note that I don't recall the exact layer names, so it may differ when you test and I'm just giving an example. The rough idea is that the current logic only works for the loading the first lora correctly IF it is a normal lora, and the second lora being loaded is a shape expander lora.

But as discussed in DM, this was not an anticipated use case - we only wanted to make control lora work as expected so the shape mismatch error on weights instead of inputs is OK for now.

sayakpaul: Hmm got it. Thanks Aryan. Just noting this is enough for now.

sayakpaul:
@a-r-r-o-w could you expand on what is the issue?

a-r-r-o-w:
> could you expand on what is the issue?

Yes. So you had asked me to add a test before and after the second lora was loaded (which would error out because of weight shape mismatch).

> > Here I would run another inference round and make sure the outputs match with the LoRA that was correctly loaded. This will help us check if this loading error didn't leave the pipeline in a broken state, which is important.

It seems like the pipeline is in a permanent broken state. After loading the first lora (shape expander), the inference runs correctly. After trying to load the second lora, which will error out, we run inference again to verify outputs don't change. But here they do, and are all NaNs. I did not find time to investigate why yet

sayakpaul:
Thanks! Let me look into this :)

sayakpaul:
@a-r-r-o-w WDYT about the changes in [7b5037f](https://github.com/huggingface/diffusers/pull/10188/commits/7b5037f980fe92ead1980d2ce5aa01d8708fb4e0)? 

@BenjaminBossan WDYT about the changes introduced in the PEFT front to handle the following scenario. We first inject the adapter config in the requested model and then we set PEFT model state dict:
https://github.com/huggingface/diffusers/blob/main/src/diffusers/loaders/peft.py#L319-L320

Now, if the second step is unsuccessful, I think we should remove the corresponding layers we had injected and make a good error message. LMK what you think.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2337,12 +2337,19 @@ def _maybe_expand_transformer_param_shape_or_error_(
...
+                    f"expanded from {module_in_features} to {in_features}"
                 )
+                if module_out_features != out_features:
+                    debug_message += (
+                        ", and the number of output features will be "
+                        f"expanded from {module_out_features} to {out_features}."
+                    )
+                else:
+                    debug_message += "."
+                logger.debug(debug_message)
```
sayakpaul, line 2352: Better crafting of the debug message I guess?

BenjaminBossan:
> WDYT about the changes introduced in the PEFT front to handle the following scenario. We first inject the adapter config in the requested model and then we set PEFT model state dict: https://github.com/huggingface/diffusers/blob/main/src/diffusers/loaders/peft.py#L319-L320
> 
> Now, if the second step is unsuccessful, I think we should remove the corresponding layers we had injected and make a good error message. LMK what you think.

To be sure I understand your suggestion correctly, do you want `inject_adapter_in_model` to catch potential errors during injection, and on error, try to roll back the changes and re-raise?

BenjaminBossan:
Yes, I think it's reasonable, I'll put it on the back log. How high priority is it?

sayakpaul:
Oh sorry I think I conveyed it wrong. I am asking what do you think of the changes I have already introduced here? I think okay for me if the changes remain in diffusers.

BenjaminBossan:
> Oh sorry I think I conveyed it wrong. I am asking what do you think of the changes I have already introduced here? I think okay for me if the changes remain in diffusers.

Ah okay, I got you now. So this is more about this part of the code, right?

https://github.com/huggingface/diffusers/pull/10188/files#diff-fdc2e14a8c1091d917cf31ca155be9baa8095d704bd2a2f58970ec31641c0caeR325-R335

I think it serves the purpose well enough. It is not a 100% rollback, as PEFT could do other things (mutate the PEFT config and toggle `requires_grad` on the base model) but I think those won't matter on this context.

Still, I'll leave the task in the backlog, as it would be useful to have in PEFT too (code would be similar to the code here).

sayakpaul:
Alright then. Could you provide an approval then?

sayakpaul:
@a-r-r-o-w WDYT about the changes? Will let you merge

BenjaminBossan (approved):
Thanks, LGTM.

> Alright then. Could you provide an approval then?

I was not asked for review ;)

src/diffusers/loaders/peft.py
```diff
@@ -316,8 +317,22 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
             if is_peft_version(">=", "0.13.1"):
                 peft_kwargs["low_cpu_mem_usage"] = low_cpu_mem_usage
 
-            inject_adapter_in_model(lora_config, self, adapter_name=adapter_name, **peft_kwargs)
-            incompatible_keys = set_peft_model_state_dict(self, state_dict, adapter_name, **peft_kwargs)
+            # To handle scnearios where we cannot successfully set state dict. If it's unsucessful,
```
a-r-r-o-w, line 320:
```suggestion
            # To handle scenarios where we cannot successfully set state dict. If it's unsucessful,
```

a-r-r-o-w (commented):
lgtm!

PR #10189, opened 2024-12-11
Check correct model type is passed to `from_pretrained`

### LINKED ISSUE #10093: [pipelines] add better checking when a wrong model is passed when initializing a pipeline
With this code:

```py
import torch 

repo_id = "black-forest-labs/FLUX.1-dev"
text_encoder_2 = T5EncoderModel.from_pretrained(repo_id, subfolder="text_encoder_2")
    repo_id, text_encoder=text_encoder_2, torch_dtype=torch.bfloat16
).to("cuda")

image = pipe("A dog").images[0]
```

where we're doing `text_encoder=text_encoder_2` (should have been `text_encoder_2=text_encoder_2`), we get an extremely uninformative error:

<details>
<summary>Error</summary>

```bash
...
... [51,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [52,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [53,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [54,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [55,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [56,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [57,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [58,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [59,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [60,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [61,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [62,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
../aten/src/ATen/native/cuda/Indexing.cu:1308: indexSelectLargeIndex: block: [95,0,0], thread: [63,0,0] Assertion `srcIndex < srcSelectDimSize` failed.
  File "/home/sayak/diffusers/check_incompatible_flux.py", line 11, in <module>
    image = pipe("A dog").images[0]
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/home/sayak/diffusers/src/diffusers/pipelines/flux/pipeline_flux.py", line 677, in __call__
    ) = self.encode_prompt(
  File "/home/sayak/diffusers/src/diffusers/pipelines/flux/pipeline_flux.py", line 353, in encode_prompt
    pooled_prompt_embeds = self._get_clip_prompt_embeds(
  File "/home/sayak/diffusers/src/diffusers/pipelines/flux/pipeline_flux.py", line 289, in _get_clip_prompt_embeds
    prompt_embeds = self.text_encoder(text_input_ids.to(device), output_hidden_states=False)
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1747, in _call_impl
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/transformers/models/t5/modeling_t5.py", line 1972, in forward
    encoder_outputs = self.encoder(
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1747, in _call_impl
  File "/home/sayak/.pyenv/versions/3.10.12/envs/diffusers/lib/python3.10/site-packages/transformers/models/t5/modeling_t5.py", line 1029, in forward
    attention_mask = torch.ones(batch_size, mask_seq_length, device=inputs_embeds.device)
RuntimeError: CUDA error: device-side assert triggered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

</details>

I know it's on the users to ensure that they're initializing a pipeline properly with the right components. However, the thrown error makes it worse to do any further debugging on the user side. IMO. 

It would be greatly helpful from the perspective of a user if we could throw a better warning/error instead. As soon as we detect wrong types (I think `model_index.json` info can be used in this case), we throw a warning/error. Is this something doable?

Cc: @DN6 @yiyixuxu 
Ccing @hlky as well in case you're interested.

hlky:
# What does this PR do?

Example
```python
import torch 

repo_id = "black-forest-labs/FLUX.1-dev"
text_encoder_2 = T5EncoderModel.from_pretrained(repo_id, subfolder="text_encoder_2")
    repo_id, text_encoder=text_encoder_2, torch_dtype=torch.bfloat16
)
```
```python
ValueError: Expected CLIPTextModel for text_encoder, got T5EncoderModel.
```

Fixes #10093

## Who can review?

cc @sayakpaul


src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -833,6 +833,13 @@ def load_module(name, value):
 
         init_dict = {k: v for k, v in init_dict.items() if load_module(k, v)}
 
+        for key, (_, expected_class_name) in zip(init_dict.keys(), init_dict.values()):
+            if key not in passed_class_obj:
+                continue
+            class_name = passed_class_obj[key].__class__.__name__
+            if class_name != expected_class_name:
+                raise ValueError(f"Expected {expected_class_name} for {key}, got {class_name}.")
```
sayakpaul, line 841: Let's add a test for this too?

sayakpaul:
Thanks @hlky! This looks leaner than I had thought. thank you!

hlky:
Thanks @sayakpaul. I've added a test, trimmed `Flax` from `class_name` and we skip checking `scheduler`, this should fix the failed tests.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -833,6 +833,14 @@ def load_module(name, value):
 
         init_dict = {k: v for k, v in init_dict.items() if load_module(k, v)}
 
+        for key, (_, expected_class_name) in zip(init_dict.keys(), init_dict.values()):
+            if key not in passed_class_obj or key == "scheduler":
+                continue
```
sayakpaul, line 838: If we pass `scheduler=text_encoder` that should be errored out as well, right?

tests/pipelines/test_pipelines.py
```diff
@@ -1802,6 +1802,17 @@ def test_pipe_same_device_id_offload(self):
...
+    def test_wrong_model(self):
+        tokenizer = CLIPTokenizer.from_pretrained("hf-internal-testing/tiny-random-clip")
+        with self.assertRaises(ValueError) as error_context:
+            _ = StableDiffusionPipeline.from_pretrained(
+                "hf-internal-testing/diffusers-stable-diffusion-tiny-all", text_encoder=tokenizer
+            )
+
+        assert "Expected" in str(error_context.exception)
+        assert "text_encoder" in str(error_context.exception)
+        assert "CLIPTokenizer" in str(error_context.exception)
```
sayakpaul, line 1814:
```suggestion
        assert f"{tokenizer.__class__.__name}" in str(error_context.exception)
```

tests/pipelines/test_pipelines.py
```diff
@@ -1802,6 +1802,17 @@ def test_pipe_same_device_id_offload(self):
...
+    def test_wrong_model(self):
+        tokenizer = CLIPTokenizer.from_pretrained("hf-internal-testing/tiny-random-clip")
+        with self.assertRaises(ValueError) as error_context:
+            _ = StableDiffusionPipeline.from_pretrained(
+                "hf-internal-testing/diffusers-stable-diffusion-tiny-all", text_encoder=tokenizer
+            )
+
+        assert "Expected" in str(error_context.exception)
+        assert "text_encoder" in str(error_context.exception)
+        assert f"{tokenizer.__class__.__name}" in str(error_context.exception)
```
sayakpaul, line 1814: Maybe also a check for the `scheduler` as that is handled slightly differently?

sayakpaul (approved):
Go to go from my side once the scheduler related tests are added. Thanks!

tests/pipelines/test_pipelines.py
hlky: Will add it. For context this is what we're handling:

https://github.com/huggingface/diffusers/blob/d041dd504058ac6b0fde3eb767eb6844d8d577b8/src/diffusers/schedulers/scheduling_utils.py#L33-L48

sayakpaul: That's cool. But I don't see the flow matching schedulers here. So, if I do assign a text encoder to `scheduler` in an RF pipeline (`FluxPipeline`, for example), would it still work as expected?

hlky: Yes that also works, for pipelines like Flux we're getting the `type`
```python
<class 'diffusers.schedulers.scheduling_flow_match_euler_discrete.FlowMatchEulerDiscreteScheduler'>
```
For SD etc we get the `enum`
```python
<enum 'KarrasDiffusionSchedulers'>
```
```python
[<KarrasDiffusionSchedulers.DDIMScheduler: 1>, <KarrasDiffusionSchedulers.DDPMScheduler: 2>, <KarrasDiffusionSchedulers.PNDMScheduler: 3>, <KarrasDiffusionSchedulers.LMSDiscreteScheduler: 4>, <KarrasDiffusionSchedulers.EulerDiscreteScheduler: 5>, <KarrasDiffusionSchedulers.HeunDiscreteScheduler: 6>, <KarrasDiffusionSchedulers.EulerAncestralDiscreteScheduler: 7>, <KarrasDiffusionSchedulers.DPMSolverMultistepScheduler: 8>, <KarrasDiffusionSchedulers.DPMSolverSinglestepScheduler: 9>, <KarrasDiffusionSchedulers.KDPM2DiscreteScheduler: 10>, <KarrasDiffusionSchedulers.KDPM2AncestralDiscreteScheduler: 11>, <KarrasDiffusionSchedulers.DEISMultistepScheduler: 12>, <KarrasDiffusionSchedulers.UniPCMultistepScheduler: 13>, <KarrasDiffusionSchedulers.DPMSolverSDEScheduler: 14>, <KarrasDiffusionSchedulers.EDMEulerScheduler: 15>]
```
So we apply the same processing (`str`, `split`, `strip` applies for `type` case) to get a list of scheduler
```python
['FlowMatchEulerDiscreteScheduler']
```
```python
['DDIMScheduler', 'DDPMScheduler', 'PNDMScheduler', 'LMSDiscreteScheduler', 'EulerDiscreteScheduler', 'HeunDiscreteScheduler', 'EulerAncestralDiscreteScheduler', 'DPMSolverMultistepScheduler', 'DPMSolverSinglestepScheduler', 'KDPM2DiscreteScheduler', 'KDPM2AncestralDiscreteScheduler', 'DEISMultistepScheduler', 'UniPCMultistepScheduler', 'DPMSolverSDEScheduler', 'EDMEulerScheduler']
```
If it's not a scheduler it will raise or if it's the wrong type of scheduler.

hlky: We now also support `Union`

hlky:
Note that we add `LCMScheduler` to `KarrasDiffusionSchedulers`, it's more of a "compatibles" because some don't actually support `Karras` and `LCMScheduler` is generally compatible with pipelines using `KarrasDiffusionSchedulers`, otherwise we'd need to change most (all?) instances of `KarrasDiffusionSchedulers` to `Union[LCMScheduler, KarrasDiffusionSchedulers]`

tests/pipelines/test_pipelines.py
hlky, line 1814: Using `f"{tokenizer.__class__.__name}"` failed with `type object 'CLIPTokenizer' has no attribute '_PipelineFastTests__name'`

tests/pipelines/test_pipelines.py
hlky: Tests for wrong scheduler are added.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -832,6 +834,28 @@ def load_module(name, value):
...
+        if "scheduler" in expected_types:
+            scheduler_types = []
+            for scheduler_type in expected_types["scheduler"]:
+                if isinstance(scheduler_type, enum.EnumMeta):
+                    scheduler_types.extend(list(scheduler_type))
+                else:
+                    scheduler_types.extend([str(scheduler_type)])
+            scheduler_types = [str(scheduler).split(".")[-1].strip("'>") for scheduler in scheduler_types]
+
+        for key, (_, expected_class_name) in zip(init_dict.keys(), init_dict.values()):
```
DN6, line 847: Since we already have the types extracted in `expected_types` can't we fetch them using the key and then check if the passed object is an instance of the type? If the expected type is an enum then we can just check if the passed obj class name exists in the keys?

tests/pipelines/test_pipelines.py
sayakpaul: Sorry for that.

src/diffusers/pipelines/pipeline_utils.py
hlky: https://github.com/huggingface/diffusers/pull/10189#discussion_r1880030678

DN6: I think it might be better to make this check more agnostic to the component names. 

We have a few pipelines with Union types on non-scheduler components (mostly AnimateDiff). So this snippet would fail even though it's valid, because init_dict is based on the model_index.json which doesn't support multiple types.  

```python
from diffusers import (
	AnimateDiffPipeline,
	UNetMotionModel,
)

unet = UNetMotionModel()
pipe = AnimateDiffPipeline.from_pretrained(
	"hf-internal-testing/tiny-sd-pipe", unet=unet
)
```

Enforcing scheduler types might be a breaking change cc: @yiyixuxu . e.g. Using DDIM with Kandinsky is currently valid, but with this change any downstream code doing this it would break. It would be good to enforce on the Flow based schedulers though (perhaps via a new Enum)

I would try 

```python
        for key, (_, expected_class_name) in zip(init_dict.keys(), init_dict.values()):
            if key not in passed_class_obj:
                continue

            class_obj = passed_class_obj[key]
            _expected_class_types = []
            for expected_type in expected_types[key]:
                if isinstance(expected_type, enum.EnumMeta):
                    _expected_class_types.extend(expected_type.__members__.keys())
                else:
                    _expected_class_types.append(expected_type.__name__)

            _is_valid_type = class_obj.__class__.__name__ in _expected_class_types
            if isinstance(class_obj, SchedulerMixin) and not _is_valid_type:
                # Handle case where scheduler is still valid 
                # raise if scheduler is meant to be a Flow based scheduler?
            elif not _is_valid_type:
                raise ValueError(f"Expected types for {key}: {_expected_class_types}, got {class_obj.__class__.__name__}.")


```

hlky: Added this for scheduler

```

Comment (yiyixuxu) [reply src/diffusers/pipelines/pipeline_utils.py]: I think we don't need a value error here, a warning is enough, no?

Comment (hlky) [reply src/diffusers/pipelines/pipeline_utils.py]: A warning should be sufficient, it's mainly for the situation here https://github.com/huggingface/diffusers/issues/10093#issue-2713961860 where the wrong text encoder is given because the resulting error is uninformative.

Comment (yiyixuxu) [reply src/diffusers/pipelines/pipeline_utils.py]: let's do a warning then:)

File: tests/pipelines/test_pipelines.py [hunk 6]
```diff
@@ -1802,6 +1804,42 @@ def test_pipe_same_device_id_offload(self):
...
 
+    def test_wrong_model(self):
+        tokenizer = CLIPTokenizer.from_pretrained("hf-internal-testing/tiny-random-clip")
+        with self.assertRaises(ValueError) as error_context:
+            _ = StableDiffusionPipeline.from_pretrained(
+                "hf-internal-testing/diffusers-stable-diffusion-tiny-all", text_encoder=tokenizer
+            )
+
+        assert "is of type" in str(error_context.exception)
+        assert "but should be" in str(error_context.exception)
```
hlky, line 1815: We're now using warning, but for this case, `CLIPTokenizer` in `text_encoder` we still get a `ValueError` later on from here
https://github.com/huggingface/diffusers/blob/63243406ba5510c10d5cac931882918ceba926f9/src/diffusers/pipelines/pipeline_utils.py#L893-L897

So it's a little inconsistent and needs further testing to determine which other cases this already applies to.

src/diffusers/pipelines/pipeline_utils.py
sayakpaul: Just chiming here a bit to share a perspective as a user (not a strong opinion). Related to https://github.com/huggingface/diffusers/pull/10189#discussion_r1884592432. 

Here

https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/pipeline_loading_utils.py#L287-L290

if there's an unexpected module passed we raise a value error. I think the check is almost along similar lines -- users are passing assigning components that are unexpected / incompatible. We probably cannot predict the consequences of allowing the loading without raising any errors but if we raise an error, users would know what to do to fix the in correct behaviour.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -833,6 +835,31 @@ def load_module(name, value):
...
+            class_obj = passed_class_obj[key]
+            _expected_class_types = []
+            for expected_type in expected_types[key]:
+                if isinstance(expected_type, enum.EnumMeta):
+                    _expected_class_types.extend(expected_type.__members__.keys())
+                else:
+                    _expected_class_types.append(expected_type.__name__)
+
+            _is_valid_type = class_obj.__class__.__name__ in _expected_class_types
+            if isinstance(class_obj, SchedulerMixin) and not _is_valid_type:
```
DN6, line 851:
```suggestion
            if (isinstance(class_obj, SchedulerMixin) or isinstance(class_obj, FlaxSchedulerMixin)) and not _is_valid_type:
```

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -833,6 +835,31 @@ def load_module(name, value):
...
+                _requires_flow_match = any("FlowMatch" in class_type for class_type in _expected_class_types)
```
DN6, line 852: I think checking against a `FlowMatchSchedulers` enum would be better in case we end up not using "FlowMatch" in the class name. 

```python
_requires_flow_match = any(class_type in FlowMatchSchedulers.__members__ for class_type in _expected_class_types)
_is_flow_match =  class_obj.__class__.__name__ in FlowMatchSchedulers
```  

cc: @yiyixuxu

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -833,6 +835,31 @@ def load_module(name, value):
...
+                _is_flow_match = "FlowMatch" in class_obj.__class__.__name__
+                if _requires_flow_match and not _is_flow_match:
+                    logger.warning(f"Expected FlowMatch scheduler, got {class_obj.__class__.__name__}.")
```
DN6, line 855: Probably okay to raise an error here because `scheduler.scale_noise` would raise an error in the flow matching pipelines if a non-FlowMatch scheduler is used.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -833,6 +835,31 @@ def load_module(name, value):
...
+                elif not _requires_flow_match and _is_flow_match:
+                    logger.warning(f"Expected non-FlowMatch scheduler, got {class_obj.__class__.__name__}.")
+            elif not _is_valid_type:
+                logger.warning(
```
DN6, line 859: I think if it's not a scheduler and the types don't match it's okay to raise an error. I think it would break in the model loading step anyway in this case. wdyt @yiyixuxu?

yiyixuxu: I prefer a warning because:
1. I think there is very little /no benefits in raising an error vs a warning here
2. in case we make a mistake in type hint, we will throw an error by mistake

src/diffusers/pipelines/pipeline_utils.py
yiyixuxu: I really don't want to raise any error here because type hint was not something enforced in this library and it is hard even for us to tell which schedulers can be used/cannot.

yiyixuxu: e.g kandinsky if memory serves I think ddim may also works with some of the pipelines, and the compatibility may change

src/diffusers/pipelines/pipeline_utils.py
yiyixuxu: we just added `use_flow_sigma` to a few non-flow match schedulers with the SANA pr, and also we plan to refactor them but don't have a design finalized yet
given that, I think maybe we can skip checking for scheduler altogether for now, and revisit later. let me know what you guys think!

hlky: I've removed scheduler related changes for now, I think we can revisit that later, as @yiyixuxu mentioned above type hints haven't been strictly enforced there are probably some missing/wrong, especially for schedulers. Warning is better because of that too, if there is some wrong type hint that makes its way into a release we'd have to issue a hotfix release to fix it, that just creates headaches and issue reports.

yiyixuxu (approved):
thanks @hlky !

DN6 (approved):

sayakpaul:
Thanks for taking it up, @hlky! To me it's a really nice QoL improvement from a DX perspective.

PR #10200, opened 2024-12-12
refactor  StableDiffusionXLControlNetUnion

hlky:
# What does this PR do?
https://github.com/huggingface/diffusers/pull/10131#issuecomment-2537167847

## Who can review?

@yiyixuxu @vladmandic

### DIFF
--- a/src/diffusers/models/controlnets/__init__.py
+++ b/src/diffusers/models/controlnets/__init__.py
@@ -15,7 +15,7 @@ if is_torch_available():
         SparseControlNetModel,
         SparseControlNetOutput,
     )
-    from .controlnet_union import ControlNetUnionInput, ControlNetUnionInputProMax, ControlNetUnionModel
+    from .controlnet_union import ControlNetUnionModel
     from .controlnet_xs import ControlNetXSAdapter, ControlNetXSOutput, UNetControlNetXSModel
     from .multicontrolnet import MultiControlNetModel
 
--- a/src/diffusers/models/controlnets/controlnet_union.py
+++ b/src/diffusers/models/controlnets/controlnet_union.py
@@ -11,14 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import torch
 from torch import nn
 
 from ...configuration_utils import ConfigMixin, register_to_config
-from ...image_processor import PipelineImageInput
 from ...loaders.single_file_model import FromOriginalModelMixin
 from ...utils import logging
 from ..attention_processor import (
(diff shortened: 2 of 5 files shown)


vladmandic:
looks great, thanks!

yiyixuxu (approved):
ok but I actually like the `ControlNetUnionInput`, it is pretty difficult to use otherwise having to figure out which mode is which number for given checkpoint

hlky:
Yeah I like it too 🤷‍♂️ seems like integration in some ui is the same whether it's mapping the names in the class or the indices, and we know that the class should be used because it's the specific pipeline.

vladmandic:
tiny, but breaking issue in:
```py
    _callback_tensor_inputs = [
        "latents",
        "prompt_embeds",
        "negative_prompt_embeds",
        "add_text_embeds",
        "add_time_ids",
        "negative_pooled_prompt_embeds",
        "negative_add_time_ids",
        "image",
    ]
```
but `image` doesnt exist anymore - remove `image` from callback list or there is an instant runtime error.

hlky:
I tested this with the same examples as before, I've just ran it again and it works, is there something specific you're doing to get "an instant runtime error"? would you mind sharing all the relevant information?

vladmandic:
if you have a callback defined, then you get a runtime since there is no `image` param.
your tests do not run with callback. only thing that is needed is to remove `image` from callback params listed above. you can replace it with `control_image` if you want, but its pointless since its not processed anymore at that time.

hlky:
#10218

vladmandic:
yup, thats exactly it - thanks!

will you also add mappings in `pipelines/auto_pipeline.py` to link new txt/img/inpaint pipelines?

hlky:
> and what are the actual values for control_mode? i'm guessing its

Yes, this is why we had the dataclass btw.

vladmandic:
in general, i like the concept of dataclass, just overriding `image` when its not-an-image was messing things badly.
anyhow, i think we're really close, just need the autopipeline mappings.

hlky:
> just overriding image when its not-an-image was messing things badly.
So we can understand the effect downstream could you link to the code that didn't work when it's not an image?

#10219 for autopipeline, only for `from_pretrained` because it's very late, let me know how much of a priority `from_pipe` is.

vladmandic:
i've tested with #10218 + #10219 and seems ok.

fyi, typical sdnext workflow is to:
- load a model
- switch a pipeline in runtime depending on user inputs. 
  e.g., if user provided input image, we'll switch to img2img. if mask is provided, we'll switch to inpaint. otherwise we'll run in txt2img.
- inspect pipeline for available args and set them as appropriate.
  using `inspect.signature(cls.__call__, follow_wrapped=True, eval_str=True)`
  e.g. if `prompt_embeds` are allowed as param, we'll try to pre-calcululate them, otherwise we'll pass prompt as-is.
  thus overloading any known param such as `image` that changes its meaning would be a nightmare.  
  if `image` param is available, its going to be handled. but it cannot be a `PIL.Image` in one pipeline and a `DataClass` in another.

vladmandic:
one more issue - `num_control_type` is hard coded as 6 in
(https://github.com/huggingface/diffusers/blob/a5f35ee4731b731d6bd8977525873b0bc480cb42/src/diffusers/models/controlnets/controlnet_union.py#L193)

but its not the same number of Union and ProMax - even in your example, you used tile mode, but tile has index 6 which is larger than allowed num_control_type.

hlky:
How are you loading ProMax? Try using `brad-twinkl/controlnet-union-sdxl-1.0-promax` or `OzzyGT/controlnet-union-promax-sdxl-1.0`, these are uploaded correctly with the right config. If you're loading [diffusion_pytorch_model_promax.safetensors](https://huggingface.co/xinsir/controlnet-union-sdxl-1.0/blob/main/diffusion_pytorch_model_promax.safetensors) from the original repo the config will be wrong, should work if you pass `num_control_type=8` to `from_pretrained`.

vladmandic:
right - yes, that has correct config, ignore the last msg.

PR #10203, opened 2024-12-12
Remove `negative_*` from SDXL callback

### LINKED ISSUE #9906: Stable Diffusion and SDXL Callbacks are fundamentally broken for prompt_embeds
### Describe the bug

I'm working on implementing scheduled prompting in SDNext and realized that the callback in SDXL pipelines is fundamentally not functional. This also applies to Stable Diffusion Pipelines and likely others. 

https://github.com/huggingface/diffusers/blob/dac623b59f52c58383a39207d5147aa34e0047cd/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L1245-L1252

Of the preceding lines the following variables are completely useless as they are not referenced inside the denoising loop and thus adjusting them in the callback can do nothing:

- `negative_prompt_embeds`
- `negative_pooled_prompt_embeds` 
- `negative_add_time_ids` 

If `self.do_classifier_free_guidance` is True, the other variable are concatenations of themselves and their negative counterparts, and not what their variables state.

https://github.com/huggingface/diffusers/blob/dac623b59f52c58383a39207d5147aa34e0047cd/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L1126

For the callback to operate intuitively, the code following `# 7` should be split out into a separate function, and following the variables NOT overwritten:

- `prompt_embeds`
- `add_text_embeds` (naming convention should be fixed)
- `add_time_ids`

That function should then be called post-callback to update all of the associated concat variables. 

### Reproduction

N/A

This is a code only issue.

### Logs

_No response_

### System Info

N/A

### Who can help?

@yiyixuxu @sayakpaul

hlky:
# What does this PR do?

Fixes #9906

## Who can review?

@yiyixuxu @asomoza

### DIFF
--- a/examples/community/README_community_scripts.md
+++ b/examples/community/README_community_scripts.md
@@ -241,27 +241,15 @@ from diffusers import StableDiffusionPipeline
 from diffusers.callbacks import PipelineCallback, MultiPipelineCallbacks
 from diffusers.configuration_utils import register_to_config
 import torch
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Tuple, Union
 
 
-pipeline: StableDiffusionPipeline = StableDiffusionPipeline.from_pretrained(
-    "stable-diffusion-v1-5/stable-diffusion-v1-5",
-    torch_dtype=torch.float16,
-    variant="fp16",
-    use_safetensors=True,
-).to("cuda")
-pipeline.safety_checker = None
-pipeline.requires_safety_checker = False
-
-
-class SDPromptScheduleCallback(PipelineCallback):
+class SDPromptSchedulingCallback(PipelineCallback):
     @register_to_config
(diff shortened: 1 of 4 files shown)


asomoza:
thanks @hlky, can you please change the example [here](https://github.com/huggingface/diffusers/blob/main/examples/community/README_community_scripts.md#prompt-scheduling-callback) for the one where the encoding is done outside of the loop? Probably best to have the more correct example  in the showcase.

hlky:
@asomoza I've changed the example for version with encoding outside of the loop and added the XL version.

yiyixuxu (approved):
thanks!

asomoza (approved):
thanks!

Issue #10202, opened 2024-12-12
Flux Control Lora not unloaded correctly


christopher5106:
### Describe the bug

Hi,

There is a bug while switching pipeline from Flux dev when lora control has been loaded:

### Reproduction

```python

import torch
from controlnet_aux import CannyDetector
from diffusers import FluxControlPipeline
from diffusers.utils import load_image
from diffusers import FluxImg2ImgPipeline


model = "black-forest-labs/FLUX.1-dev"
pipe = FluxControlPipeline.from_pretrained(
    model, 
    torch_dtype=torch.bfloat16
).to("cuda")

pipe.load_lora_weights(
    "black-forest-labs/FLUX.1-Canny-dev-lora"
)

prompt = "A robot made of exotic candies and chocolates of different kinds. The background is filled with confetti and celebratory gifts."
control_image = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/robot.png")

processor = CannyDetector()
control_image = processor(control_image, low_threshold=50, high_threshold=200, detect_resolution=1024, image_resolution=1024)

image = pipe(
    prompt=prompt,
    control_image=control_image,
    num_inference_steps=50,
    guidance_scale=30.0,
).images[0]

# pipe = FluxImg2ImgPipeline.from_pretrained(
#     "black-forest-labs/FLUX.1-dev", 
#     torch_dtype=torch.bfloat16
# )

pipe = FluxImg2ImgPipeline.from_pipe(
    pipe,
    torch_dtype=torch.bfloat16
)

pipe = pipe.to("cuda")

init_image = load_image("https://raw.githubusercontent.com/CompVis/stable-diffusion/main/assets/stable-samples/img2img/sketch-mountains-input.jpg").resize((1024, 1024))
prompt = "cat wizard, gandalf, lord of the rings, detailed, fantasy, cute, adorable, Pixar, Disney, 8k"
image = pipe(
    prompt=prompt, 
    image=init_image, 
    num_inference_steps=28, 
    strength=0.5, 
    guidance_scale=2.5
).images[0]

```

Remplacing the `from_pipe` loading by standard loading shows the previous code should work

```python
pipe = FluxImg2ImgPipeline.from_pretrained(
     "black-forest-labs/FLUX.1-dev", 
     torch_dtype=torch.bfloat16
)
```

### System Info

Ubuntu

### Who can help?

@yiyixuxu @sayakpaul @DN6 @asomoza

[...]

PR #10206, opened 2024-12-13
[LoRA] feat: support `unload_lora_weights()` for Flux Control.

sayakpaul:
# What does this PR do?

Fixes: https://github.com/huggingface/diffusers/issues/10202. 

Will request for reviews from others later.


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2283,6 +2283,50 @@ def unload_lora_weights(self):
...
+                    module_weight = module.weight.data
+                    module_bias = module.bias.data if module.bias is not None else None
+                    bias = module_bias is not None
+
+                    parent_module_name, _, current_module_name = name.rpartition(".")
+                    parent_module = transformer.get_submodule(parent_module_name)
+
+                    current_param_weight = overwritten_params[f"{name}.weight"]
+                    in_features, out_features = current_param_weight.shape[1], current_param_weight.shape[0]
+                    with torch.device("meta"):
```
sayakpaul, line 2306: Since we already pin `torch` version this is safe enough.

BenjaminBossan (approved):
Looks pretty good, well tested, no issues from my side.

sayakpaul:
@yiyixuxu @a-r-r-o-w could you give this a look?

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Also cc: @a-r-r-o-w. Something we should consider doing in:
https://github.com/huggingface/diffusers/blob/main/src/diffusers/loaders/lora_pipeline.py#L2351-L2354

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2283,6 +2283,49 @@ def unload_lora_weights(self):
...
+                        original_module = torch.nn.Linear(
+                            in_features,
+                            out_features,
+                            bias=bias,
+                            device=module_weight.device,
```
a-r-r-o-w, line 2310: We're in the meta device initialization context but also pass device here. I think the context takes precedence, but in that case we should not pass a device here.

```suggestion
```

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2378,6 +2422,14 @@ def _maybe_expand_transformer_param_shape_or_error_(
                     setattr(transformer.config, attribute_name, new_value)
                     logger.info(f"Set the {attribute_name} attribute of the model to {new_value} from {old_value}.")
 
+                    # For `unload_lora_weights()`.
+                    overwritten_params[f"{current_module_name}.weight"] = module_weight
```
a-r-r-o-w, line 2426: I think this would have a small but significant memory overhead. For inference purposes only with loras, maybe this could be made opt-out if we know we never want call `unload_lora_weights`. Not a blocker though and can be tackled in a different PR but lmk your thoughts

a-r-r-o-w (approved):
Thanks for supporting this! Changes look good to me

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Yeah this could be tackled with `discard_original_layers`.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2283,6 +2283,50 @@ def unload_lora_weights(self):
...
+                            out_features,
+                            bias=bias,
+                            dtype=module_weight.dtype,
+                        )
+
+                    tmp_state_dict = {"weight": current_param_weight}
+                    if module_bias is not None:
+                        tmp_state_dict.update({"bias": overwritten_params[f"{name}.bias"]})
+                    original_module.load_state_dict(tmp_state_dict, assign=True, strict=True)
+                    setattr(parent_module, current_module_name, original_module)
```
sayakpaul, line 2317: @a-r-r-o-w thanks for flagging the `device` assignment while initializing `original_module`. `device` takes priority so `original_module` was not getting initialized on "meta", rending the previous `copy_()` ops ineffective. 

LMK what you think about the current changes (have run the corresponding tests on a GPU and they pass). 

@DN6 LMK your comments here too.

a-r-r-o-w (approved):

PR #10156, opened 2024-12-09
Use `torch` in `get_2d_sincos_pos_embed` and `get_3d_sincos_pos_embed`

hlky:
# What does this PR do?

Refactors get_2d_sincos_pos_embed and get_3d_sincos_pos_embed to use torch instead of numpy, and adds device argument so that tensors can be created on e.g. cuda.

Usage of get_2d_sincos_pos_embed and get_3d_sincos_pos_embed is updated to pass device where applicable (we don't specify device during initialization so we don't pass device to the function when used from `__init__`, the device from weights would just be cpu)

torch and numpy versions match numerically.

<details>
  <summary>Reproduction `get_2d_sincos_pos_embed`</summary>

  ```python
from diffusers.models.embeddings import get_2d_sincos_pos_embed
import torch
from typing import Optional


def get_2d_sincos_pos_embed_torch(
    embed_dim,
    grid_size,
    cls_token=False,
    extra_tokens=0,
    interpolation_scale=1.0,
    base_size=16,
    device: Optional[torch.device] = None,
):
    """
    Creates 2D sinusoidal positional embeddings.

    Args:
        embed_dim (`int`):
            The embedding dimension.
        grid_size (`int`):
            The size of the grid height and width.
        cls_token (`bool`, defaults to `False`):
            Whether or not to add a classification token.
        extra_tokens (`int`, defaults to `0`):
            The number of extra tokens to add.
        interpolation_scale (`float`, defaults to `1.0`):
            The scale of the interpolation.

    Returns:
        pos_embed (`torch.Tensor`):
            Shape is either `[grid_size * grid_size, embed_dim]` if not using cls_token, or `[1 + grid_size*grid_size,
            embed_dim]` if using cls_token
    """
    if isinstance(grid_size, int):
        grid_size = (grid_size, grid_size)

    grid_h = (
        torch.arange(grid_size[0], device=device, dtype=torch.float32)
        / (grid_size[0] / base_size)
        / interpolation_scale
    )
    grid_w = (
        torch.arange(grid_size[1], device=device, dtype=torch.float32)
        / (grid_size[1] / base_size)
        / interpolation_scale
    )
    grid = torch.meshgrid(grid_w, grid_h, indexing="xy")  # here w goes first
    grid = torch.stack(grid, dim=0)

    grid = grid.reshape([2, 1, grid_size[1], grid_size[0]])
    pos_embed = get_2d_sincos_pos_embed_from_grid_torch(embed_dim, grid)
[...]

PR #10212, opened 2024-12-13
Add enable_vae_tiling to AllegroPipeline, fix example

hlky:
# What does this PR do?

Allegro doesn't support VAE without tiling and is missing `enable_vae_tiling` etc on the pipeline.

https://github.com/huggingface/diffusers/blob/cef0e3677e9a32fcf27cdfccfdf809f689a6f908/src/diffusers/models/autoencoders/autoencoder_kl_allegro.py#L851

Note that the default `num_inference_steps=100` is very slow (~1h on A40, only slightly faster on A6000 Ada) so the example could have the value changed as well.

## Who can review?

@sayakpaul @yiyixuxu @DN6

### DIFF
--- a/src/diffusers/pipelines/allegro/pipeline_allegro.py
+++ b/src/diffusers/pipelines/allegro/pipeline_allegro.py
@@ -59,6 +59,7 @@ EXAMPLE_DOC_STRING = """
 
         >>> vae = AutoencoderKLAllegro.from_pretrained("rhymes-ai/Allegro", subfolder="vae", torch_dtype=torch.float32)
         >>> pipe = AllegroPipeline.from_pretrained("rhymes-ai/Allegro", vae=vae, torch_dtype=torch.bfloat16).to("cuda")
+        >>> pipe.enable_vae_tiling()
 
         >>> prompt = (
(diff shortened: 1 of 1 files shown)


a-r-r-o-w (approved):
Thanks!

Regarding the number of inference steps being high, inference with 50 steps produces noisy videos. The default steps in the official repo is 100 as well. For faster inference, we could explore the different cache techniques soon.

hlky:
@a-r-r-o-w Is lowering `num_frames` also known to produce noisy videos? I'm getting noisy video with `num_inference_steps=100` and `num_frames=22`. I'm trying to test if outputs are changed with #10156 so I wanted to to be faster.

a-r-r-o-w:
I believe that is expected. My memory is a bit hazy since it's been a while since we integrated it, but IIRC Allegro only works at 88 frames and specific resolutions (?).

I think it is okay to not run the full test if it's time consuming. What you could do instead is - save outputs of all intermediate transformers blocks and after decoding for say 2 inference steps, and compare that to your sincos/rope changes by looking at absmax of the intermediates from both runs. This should be a sufficient check imo

PR #10213, opened 2024-12-13
Rename LTX blocks and docs title

a-r-r-o-w:
Slipped through in the original PR where I only renamed the transformer and vae but not the internal blocks. This is because LTX is the name of company and LTXVideo is the name of model


a-r-r-o-w:
Gentle ping @yiyixuxu before we go into release

DN6 (approved):

PR #10215, opened 2024-12-13
Kolors UNet2DConditionModel `from_single_file`

### LINKED ISSUE #10207: KolorsPipeline does not support from_single_file
from diffusers import KolorsPipeline
KolorsPipeline.from_single_file("models/kolrs-8steps.safetensors")
 

How does KolorsPipeline load a single file model?

hlky:
# What does this PR do?

Context https://github.com/huggingface/diffusers/issues/10207#issuecomment-2541507223

Fixes #10207

This partially enables `from_single_file` for other UNet2DConditionModel already in Diffusers format, others will need Diffusers keys adding to `CHECKPOINT_KEY_NAMES` and `infer_diffusers_model_type` updating.

https://github.com/huggingface/diffusers/blob/cef0e3677e9a32fcf27cdfccfdf809f689a6f908/src/diffusers/loaders/single_file_utils.py#L60 

## Who can review?

cc @a-r-r-o-w @DN6

### DIFF
--- a/src/diffusers/loaders/single_file_utils.py
+++ b/src/diffusers/loaders/single_file_utils.py
@@ -100,7 +100,7 @@ CHECKPOINT_KEY_NAMES = {
     ],
     "autoencoder-dc": "decoder.stages.1.op_list.0.main.conv.conv.bias",
     "autoencoder-dc-sana": "encoder.project_in.conv.bias",
-    "kolors": "down_blocks.1.attentions.0.transformer_blocks.0.attn2.to_k.weight"
+    "kolors": "down_blocks.1.attentions.0.transformer_blocks.0.attn2.to_k.weight",
 }
 
 DIFFUSERS_DEFAULT_PIPELINE_PATHS = {
--- a/src/diffusers/loaders/single_file_utils.py
+++ b/src/diffusers/loaders/single_file_utils.py
@@ -100,6 +100,7 @@ CHECKPOINT_KEY_NAMES = {
     ],
     "autoencoder-dc": "decoder.stages.1.op_list.0.main.conv.conv.bias",
     "autoencoder-dc-sana": "encoder.project_in.conv.bias",
+    "kolors": "down_blocks.1.attentions.0.transformer_blocks.0.attn2.to_k.weight"
 }
 
 DIFFUSERS_DEFAULT_PIPELINE_PATHS = {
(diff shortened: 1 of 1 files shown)


DN6:
Hmm seems to be a recurring issue. I don't think it's sustainable to maintain a large mapping of both diffusers keys and original keys in single_file_utils. 

In the case of Flux the keys in CHECKPOINT_KEY_NAMES are original checkpoint keys with different prefixes (comfy uses the prefixed version). ControlNet is a bit of an exception since we published the initial ControlNets for SDXL when it was released, so the diffusers format checkpoint keys sort of stuck. 

Single File loading wasn't really designed to support diffusers format checkpoints. I think it might be better to add logic that raises a helpful error message if we can't identify the model type.    

Or we add a flag in single file, such as `convert=False` to skip the model inferring and conversion step and it works when you pass in a model repo id to the `config` argument. e.g

```python
import torch
from diffusers import KolorsPipeline, EulerAncestralDiscreteScheduler

pipe = KolorsPipeline.from_single_file(
    "kolors8AcceleratedVersionOf8_accelerated.safetensors", config="Kwai-Kolors/Kolors-diffusers",  convert=False
)
```

hlky:
Thanks @DN6, `convert`/`config` will work, I don't think it's very common to load Diffusers models with `from_single_file` anyway, the linked issue with Kolors is a special case because a `8 step`/"turbo" version of the UNet has been released on civit, it's already in Diffusers format but unless I'm missing something we just don't have a nice way to load it to the pipeline other than something like this.

```python
import safetensors.torch

weights = safetensors.torch.load_file("kolors8AcceleratedVersionOf8_accelerated.safetensors")
pipe = KolorsPipeline.from_pretrained(
    "Kwai-Kolors/Kolors-diffusers", variant="fp16", torch_dtype=torch.float16
)
pipe.unet.load_state_dict(weights)
```

Ideally the user could do something like

```python
unet = UNet2DConditionModel.from_single_file("kolors8AcceleratedVersionOf8_accelerated.safetensors", config="Kwai-Kolors/Kolors-diffusers",  convert=False)
pipe = KolorsPipeline.from_pretrained(
    "Kwai-Kolors/Kolors-diffusers", unet=unet, variant="fp16", torch_dtype=torch.float16
)
```

DN6:
Nice. I'll open a PR to add convert/config then 👍🏽

PR #10225, opened 2024-12-15
Add offload option in flux-control training

Adenialzz:
Set model offloading to be optional in args, which will reduce the training memory usage but also reduce the training speed.

@sayakpaul

### DIFF
--- a/examples/flux-control/train_control_flux.py
+++ b/examples/flux-control/train_control_flux.py
@@ -541,7 +541,7 @@ def parse_args(input_args=None):
         default=1.29,
         help="Scale of mode weighting scheme. Only effective when using the `'mode'` as the `weighting_scheme`.",
     )
-    parser.add_argument("--offload", action="store_true", help="Whether to do model offloading")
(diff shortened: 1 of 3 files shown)


examples/flux-control/train_control_flux.py
```diff
@@ -541,6 +541,7 @@ def parse_args(input_args=None):
         default=1.29,
         help="Scale of mode weighting scheme. Only effective when using the `'mode'` as the `weighting_scheme`.",
     )
+    parser.add_argument("--offload", action="store_true", help="Whether to do model offloading")
```
sayakpaul, line 544:
```suggestion
    parser.add_argument("--offload", action="store_true", help="Whether to offload the VAE and the text encoders to CPU when they are not used.")
```

sayakpaul (approved):
Nice, thank you!

sayakpaul:
Could you run `make style && make quality`?

Adenialzz:
fixed now. ruff version issue

sayakpaul:
Thanks for your PR!

PR #7763, opened 2024-04-24
[Utils] add utilities for checking if certain utilities are properly documented

sayakpaul:
# What does this PR do?

We have a bunch of attention processors, normalization layers, and activation layers in the codebase and they have thin documentation. This is fine because users can always see their usage in the context of a model or a pipeline to get the fuller picture. 

But since we document them in separate pages too, (such as [this](https://huggingface.co/docs/diffusers/main/en/api/attnprocessor), [this](https://huggingface.co/docs/diffusers/main/en/api/activations), and [this](https://huggingface.co/docs/diffusers/main/en/api/normalization)), I think it makes sense to ensure these docs properly reflect what all we support. 

This PR adds utilities to do that. @stevhliu I think you will like it :-)

PR #10230, opened 2024-12-16
[Docs] add rest of the lora loader mixins to the docs.

sayakpaul:
# What does this PR do?

We can easily catch these during the PRs with https://github.com/huggingface/diffusers/pull/7763.


stevhliu (approved):
LGTM, thanks for adding!

PR #10242, opened 2024-12-16
Fix `use_flow_sigmas`

hlky:
# What does this PR do?

Error when using `use_flow_sigmas` due to missing `.copy()`

```
ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().)
```

## Who can review?

@yiyixuxu @vladmandic


vladmandic:
that's fast! thanks @hlky

PR #10243, opened 2024-12-16
Use `t` instead of `timestep` in `_apply_perturbed_attention_guidance`

hlky:
# What does this PR do?

To match usages of `_apply_perturbed_attention_guidance` we pass `t` instead of `timestep` (which is expanded).

```
RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

## Who can review?

@yiyixuxu @vladmandic


yiyixuxu (approved):
thanks!

Issue #10180, opened 2024-12-10
Can't load multiple loras when using Flux Control LoRA 


jonathanyin12:
### Describe the bug

I was trying out the FluxControlPipeline with the Control LoRA introduced in #9999 , but had issues loading in multiple loras. 

For example, if I load the depth lora first and then the 8-step lora, it errors on the 8-step lora, and if I load the 8-step lora first and then the depth lora, it errors when loading the depth lora. 

### Reproduction

```
from diffusers import FluxControlPipeline
from huggingface_hub import hf_hub_download
import torch

control_pipe = FluxControlPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16).to("cuda")
control_pipe.load_lora_weights("black-forest-labs/FLUX.1-Depth-dev-lora")
control_pipe.load_lora_weights(hf_hub_download("ByteDance/Hyper-SD", "Hyper-FLUX.1-dev-8steps-lora.safetensors"))

```

### Logs

```shell
AttributeError                            Traceback (most recent call last)
Cell In[6], line 8
      5 control_pipe = FluxControlPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16).to("cuda")
      7 control_pipe.load_lora_weights("black-forest-labs/FLUX.1-Depth-dev-lora")
----> 8 control_pipe.load_lora_weights(
      9         hf_hub_download(
     10             "ByteDance/Hyper-SD", "Hyper-FLUX.1-dev-8steps-lora.safetensors"
     11         ),
     12         adapter_name="HyperFlux",
     13     )

File ~/.venv/lib/python3.10/site-packages/diffusers/loaders/lora_pipeline.py:1856, in FluxLoraLoaderMixin.load_lora_weights(self, pretrained_model_name_or_path_or_dict, adapter_name, **kwargs)
   1849 transformer_norm_state_dict = {
   1850     k: state_dict.pop(k)
   1851     for k in list(state_dict.keys())
   1852     if "transformer." in k and any(norm_key in k for norm_key in self._control_lora_supported_norm_keys)
   1853 }
   1855 transformer = getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer
-> 1856 has_param_with_expanded_shape = self._maybe_expand_transformer_param_shape_or_error_(
   1857     transformer, transformer_lora_state_dict, transformer_norm_state_dict
   1858 )
[...]

Issue #10227, opened 2024-12-15
NotImplementedError: Only LoRAs with input/output features higher than the current module's input/output features are currently supported


JakobLS:
### Describe the bug

Unable to merge LoRAs with input/output features higher than the current module's input/output features. See the example and logs below.

### Reproduction

```python
from diffusers import FluxFillPipeline, FluxTransformer2DModel

transformer = FluxTransformer2DModel.from_pretrained(
    "xiaozaa/catvton-flux-alpha", 
    torch_dtype=torch.bfloat16
)
pipe = FluxFillPipeline.from_pretrained(
    "black-forest-labs/FLUX.1-dev",
    transformer=transformer,
    torch_dtype=torch.bfloat16
).to("cuda")

adapter_id = "alimama-creative/FLUX.1-Turbo-Alpha"
pipe.load_lora_weights(adapter_id)
pipe.fuse_lora()
```

### Logs

```shell
NotImplementedError                       Traceback (most recent call last)
Cell In[7], line 48
     46 # Load and fuse Lora
     47 adapter_id = "alimama-creative/FLUX.1-Turbo-Alpha"
---> 48 pipe.load_lora_weights(adapter_id)
     49 pipe.fuse_lora()

File ~/miniconda3/envs/env-p11/lib/python3.11/site-packages/diffusers/loaders/lora_pipeline.py:1856, in FluxLoraLoaderMixin.load_lora_weights(self, pretrained_model_name_or_path_or_dict, adapter_name, **kwargs)
   1849 transformer_norm_state_dict = {
   1850     k: state_dict.pop(k)
   1851     for k in list(state_dict.keys())
   1852     if "transformer." in k and any(norm_key in k for norm_key in self._control_lora_supported_norm_keys)
   1853 }
   1855 transformer = getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer
-> 1856 has_param_with_expanded_shape = self._maybe_expand_transformer_param_shape_or_error_(
   1857     transformer, transformer_lora_state_dict, transformer_norm_state_dict
   1858 )
   1860 if has_param_with_expanded_shape:
   1861     logger.info(
   1862         "The LoRA weights contain parameters that have different shapes that expected by the transformer. "
   1863         "As a result, the state_dict of the transformer has been expanded to match the LoRA parameter shapes. "
   1864         "To get a comprehensive list of parameter names that were modified, enable debug logging."
   1865     )

[...]

Issue #10184, opened 2024-12-11
flux fill cannot use lora(flux turbo lora)


Suprhimp:
### Describe the bug

I want to use flux fill pipeline with turbo lora, but when I load pipeline and load lora model, than gives error 

### Reproduction

```
from diffusers import FluxFillPipeline

def model_fn(model_dir: str) -> FluxFillPipeline:

    pipe = FluxFillPipeline.from_pretrained(
       "black-forest-labs/FLUX.1-Fill-dev", torch_dtype=torch.bfloat16
    ).to("cuda")
    
    pipe.load_lora_weights(f"alimama-creative/FLUX.1-Turbo-Alpha")
    pipe.fuse_lora()
    
    
    return pipe

```

### Logs

```shell
NotImplementedError: Only LoRAs with input/output features higher than the current module's input/output features are currently supported. The provided LoRA contains in_features=64 and out_features=3072, which are lower than module_in_features=384 and module_out_features=3072. If you require support for this please open an issue at https://github.com/huggingface/diffusers/issues.
```

### System Info

latest(github version diffusers), python3.10, ubuntu with nvidia gpu

### Who can help?

@sayakpaul

sayakpaul:
I am not sure you can. You're trying to load a LoRA that was obtained on Flux.1-Dev into something different.

Suprhimp:
I thought that I can use it  becuase I tried with comfyui first. it worked well with comfyui

sayakpaul:
Cc: @a-r-r-o-w could be nice to try the LoRA expansion idea here as well.

a-r-r-o-w:
Will have to take a look at what Comfy is doing to be sure they're not simply dropping the input lora layer (which requires expansion/shrinking) and using all the other layers - in this case it will work easily. Or perhaps they are not passing the channelwise-concatenated latents through the lora layer and just passing the denoising latents. I currently don't have the bandwidth to try this out, but will be sure to look into it soon!

sayakpaul:
> Or perhaps they are not passing the channelwise-concatenated latents through the lora layer and just passing the denoising latents.

Valid.

Suprhimp:
I hope I can use turbo for flux fill soon :)

kadirnar:
@sayakpaul 
Does this work?

https://github.com/nftblackmagic/catvton-flux/blob/main/tryon_inference_lora.py#L26

Code:
```python
        transformer = FluxTransformer2DModel.from_pretrained(
[...]

PR #10259, opened 2024-12-17
[LoRA] feat: support loading regular Flux LoRAs into Flux Control, and Fill

sayakpaul:
# What does this PR do?
Fixes https://github.com/huggingface/diffusers/issues/10180, https://github.com/huggingface/diffusers/issues/10227, #10184 

In short, this PR enables few-steps inference for Flux Control, Fill, Redux, etc. 

<details>
<summary>Fill + Turbo LoRA</summary>

```py
from diffusers import FluxFillPipeline
import torch

pipe = FluxFillPipeline.from_pretrained(
    "black-forest-labs/FLUX.1-Fill-dev", torch_dtype=torch.bfloat16
).to("cuda")

adapter_id = "alimama-creative/FLUX.1-Turbo-Alpha"
pipe.load_lora_weights(adapter_id)

image = load_image("https://huggingface.co/datasets/diffusers/diffusers-images-docs/resolve/main/cup.png")
mask = load_image("https://huggingface.co/datasets/diffusers/diffusers-images-docs/resolve/main/cup_mask.png")

image = pipe(
    prompt="a white paper cup",
    image=image,
    mask_image=mask,
    height=1632,
    width=1232,
    guidance_scale=30,
    num_inference_steps=8,
    max_sequence_length=512,
).images[0]
image.save("flux-fill-dev.png")
```

</details>

<details>
<summary>Flux Control LoRA + Turbo LoRA (different from the previous one)</summary>

```py
from diffusers import FluxControlPipeline
from image_gen_aux import DepthPreprocessor
import torch

control_pipe = FluxControlPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16)
control_pipe.load_lora_weights("black-forest-labs/FLUX.1-Depth-dev-lora", adapter_name="depth")
control_pipe.load_lora_weights(
    hf_hub_download("ByteDance/Hyper-SD", "Hyper-FLUX.1-dev-8steps-lora.safetensors"), adapter_name="hyper-sd"
)
control_pipe.set_adapters(["depth", "hyper-sd"], adapter_weights=[0.85, 0.125])
control_pipe.enable_model_cpu_offload()

control_image = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/robot.png")

processor = DepthPreprocessor.from_pretrained("LiheYoung/depth-anything-large-hf")
control_image = processor(control_image)[0].convert("RGB")

image = control_pipe(
    prompt=prompt,
    control_image=control_image,
    height=1024,
    width=1024,
    num_inference_steps=8,
    guidance_scale=10.0,
).images[0]
image.save("output.png")
```

</details>

## Todods

- [ ] Integration tests
- [x] Docs


sayakpaul:
Requesting for a review from @BenjaminBossan for initial stuff. Then will request reviews from others.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2329,56 +2331,90 @@ def _maybe_expand_transformer_param_shape_or_error_(
...
 
-                # TODO: consider initializing this under meta device for optims.
-                expanded_module = torch.nn.Linear(
-                    in_features, out_features, bias=bias, device=module_weight.device, dtype=module_weight.dtype
-                )
-                # Only weights are expanded and biases are not.
-                new_weight = torch.zeros_like(
-                    expanded_module.weight.data, device=module_weight.device, dtype=module_weight.dtype
+    @classmethod
+    def _maybe_expand_lora_state_dict(cls, transformer, lora_state_dict):
```
BenjaminBossan, line 2387: Just for my understanding, only LoRA is expanded because only the input shape is larger, but output shape stays the same?

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2329,56 +2331,90 @@ def _maybe_expand_transformer_param_shape_or_error_(
...
-                    attribute_name = _MODULE_NAME_TO_ATTRIBUTE_MAP_FLUX[current_module_name]
-                    new_value = int(expanded_module.weight.data.shape[1])
-                    old_value = getattr(transformer.config, attribute_name)
-                    setattr(transformer.config, attribute_name, new_value)
-                    logger.info(f"Set the {attribute_name} attribute of the model to {new_value} from {old_value}.")
-
-        return has_param_with_shape_update
+        if expanded_module_names:
+            logger.info(
+                f"Found some LoRA modules for which the weights were zero-padded: {expanded_module_names}. Please open an issue if you think this was unexpected - https://github.com/huggingface/diffusers/issues/new."
```
BenjaminBossan, line 2415: I think the message could be a bit misleading, since it's diffusers that did the padding, but it reads like diffusers just discovered this. How about a more active phrasing:

"The following LoRA modules were zero padded to [reason]: {expanded_module_names}.".

Also, could be changed to debug log IMHO, unless there is good reason to be believe it could fire inadvertently.

BenjaminBossan (approved):
Thanks for extending the functionality of loading LoRA adapters when shapes need to be expanded. The PR LGTM, I only have a nit.

One question that came up (maybe it was already discussed and I just missed it or forgot): Right now, this type of expansion is permanent, right? I.e. even after unloading the LoRA that made the expansion necessary in the first place, the expansion is not undone. Probably that would be quite hard to add and not worth the effort, I'm just curious.

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Yeah. Only the inputs.

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Our users prefer explicit logging I believe. So, I logged it as an info. But no strong opinions.

sayakpaul: Done fixing the error message.

sayakpaul:
@BenjaminBossan 

> Right now, this type of expansion is permanent, right? I.e. even after unloading the LoRA that made the expansion necessary in the first place, the expansion is not undone. Probably that would be quite hard to add and not worth the effort, I'm just curious.

The LoRA state dict expansion is permanent. But model-level state dict expansion can be undone is being added in https://github.com/huggingface/diffusers/pull/10206

BenjaminBossan:
> The LoRA state dict expansion is permanent. But model-level state dict expansion can be undone is being added in #10206

Yes, something similar for LoRA would be nice, but it's not as important, as the overhead for LoRA should be relatively small.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2329,56 +2331,102 @@ def _maybe_expand_transformer_param_shape_or_error_(
...
+                if out_features > module_out_features or in_features > module_in_features:
+                    has_param_with_shape_update = True
+                    parent_module_name, _, current_module_name = name.rpartition(".")
+                    parent_module = transformer.get_submodule(parent_module_name)
+
+                    with torch.device("meta"):
+                        expanded_module = torch.nn.Linear(
+                            in_features, out_features, bias=bias, dtype=module_weight.dtype
+                        )
+                    # Only weights are expanded and biases are not.
```
a-r-r-o-w, line 2360:
```suggestion
                    # Only weights are expanded and biases are not. This is because only the input dimensions
                    # are changed while the output dimensions remain the same. The shape of the weight tensor
                    # is (out_features, in_features), while the shape of bias tensor is (out_features,), which
                    # explains the reason why only weights are expanded.
```

tests/lora/test_lora_layers_flux.py
```diff
@@ -478,21 +463,16 @@ def test_lora_expanding_shape_with_normal_lora_raises_error(self):
             "transformer.x_embedder.lora_B.weight": normal_lora_B.weight,
         }
 
-        # The first lora expanded the input features of x_embedder. Here, we are trying to load a lora with the correct
```
a-r-r-o-w, line 481: So this error is now removed because expanding lora state dicts is now supported

tests/lora/test_lora_layers_flux.py
```diff
@@ -535,17 +515,107 @@ def test_lora_expanding_shape_with_normal_lora_raises_error(self):
...
         }
 
-        # We should check for input shapes being incompatible here. But because above mentioned issue is
-        # not a supported use case, and because of the PEFT renaming, we will currently have a shape
-        # mismatch error.
+        # We should check for input shapes being incompatible here.
         self.assertRaisesRegex(
             RuntimeError,
-            "size mismatch for x_embedder.lora_A.adapter-2.weight",
+            "x_embedder.lora_A.weight",
```
a-r-r-o-w, line 521: Why is that part of the comment being removed, and why is `size mismatch` part being removed?

Just to confirm, we still don't support:
- Load N different "normal" loras
- Try to load "shape expander" lora
- Will the N different normal loras be expanded?

tests/lora/test_lora_layers_flux.py
```diff
@@ -535,17 +515,107 @@ def test_lora_expanding_shape_with_normal_lora_raises_error(self):
...
+        self.assertFalse(np.allclose(lora_output, lora_output_2, atol=1e-3, rtol=1e-3))
+        self.assertFalse(np.allclose(lora_output, lora_output_3, atol=1e-3, rtol=1e-3))
+        self.assertFalse(np.allclose(lora_output_2, lora_output_3, atol=1e-3, rtol=1e-3))
+
+        pipe.fuse_lora(lora_scale=1.0, adapter_names=["adapter-1", "adapter-2"])
+        lora_output_4 = pipe(**inputs, generator=torch.manual_seed(0))[0]
+        self.assertTrue(np.allclose(lora_output_3, lora_output_4, atol=1e-3, rtol=1e-3))
+
+    def test_load_regular_lora(self):
+        # This test checks if a regular lora (think of one trained Flux.1 Dev for example) can be loaded
```
a-r-r-o-w, line 586:
```suggestion
        # This test checks if a regular lora (think of one trained on Flux.1 Dev for example) can be loaded
```

a-r-r-o-w (approved):
Thanks for making this work @sayakpaul! Just some tiny nits and a question

tests/lora/test_lora_layers_flux.py
sayakpaul: Exactly!

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2309,16 +2312,17 @@ def _maybe_expand_transformer_param_shape_or_error_(
...
         for name, module in transformer.named_modules():
             if isinstance(module, torch.nn.Linear):
                 module_weight = module.weight.data
                 module_bias = module.bias.data if module.bias is not None else None
                 bias = module_bias is not None
 
-                lora_A_weight_name = f"{name}.lora_A.weight"
-                lora_B_weight_name = f"{name}.lora_B.weight"
-                if lora_A_weight_name not in state_dict.keys():
+                lora_base_name = name.replace(".base_layer", "") if is_peft_loaded else name
```
a-r-r-o-w, line 2322: Awesome!

tests/lora/test_lora_layers_flux.py
```diff
@@ -546,6 +519,98 @@ def test_lora_expanding_shape_with_normal_lora_raises_error(self):
...
+        lora_output = pipe(**inputs, generator=torch.manual_seed(0))[0]
+
+        normal_lora_A = torch.nn.Linear(in_features, rank, bias=False)
+        normal_lora_B = torch.nn.Linear(rank, out_features, bias=False)
+        lora_state_dict = {
+            "transformer.x_embedder.lora_A.weight": normal_lora_A.weight,
+            "transformer.x_embedder.lora_B.weight": normal_lora_B.weight,
+        }
+
+        pipe.load_lora_weights(lora_state_dict, "adapter-2")
```
a-r-r-o-w, line 564: Very cool that this would now be possible! :fire:

a-r-r-o-w (approved):
Thanks Sayak! LGTM

The case for the following can be looked into in a separate PR because it is not too important.
- Step 1: Load N "normal" loras
- Step 2: Load shape expander lora
- Step 3; Run inference

This should currently raise an input size mismatch error on the normal loras when the inputs pass through the normal lora computation path - which is okay. This is because we don't handle the case of rejigging and expanding all the previous N loras if a shape expander lora is loaded after them.

Maybe if someone reports an issue, it would be worth looking into. But otherwise, the recommendation would be to load shape expander lora first, and then all the others loras. Thanks again for making this possible!

sayakpaul:
Just completed running ALL the integration tests for Flux LoRA (`test_lora_layers_flux.py`). All green 🟢

PR #10263, opened 2024-12-17
[tests] Remove/rename unsupported quantization torchao type

a-r-r-o-w:
We decided to not support some quantization shorthands (such as `{dtype}_aXbY`). I forgot to remove/rename those in the tests, which will cause the tests to fail unless updated correctly


a-r-r-o-w:
Failing tests are unrelated

PR #10273, opened 2024-12-18
[tests] remove nullop import checks from lora tests

a-r-r-o-w:


sayakpaul (approved):
Thanks. Reducing LoC is always a treat no matter how small the amount.

PR #10286, opened 2024-12-18
change the default Sana pre-trained model for LoRA to BF16 version

lawrence-cj:
# What does this PR do?
change the default pre-train model for LoRA training of Sana to the BF16 version for better performance

Cc: @sayakpaul 

-->


lawrence-cj:
Close as already exist.

PR #10293, opened 2024-12-18
Bump minimum TorchAO version to 0.7.0

a-r-r-o-w:
Context: https://huggingface.slack.com/archives/C065E480NN9/p1734425021147699

cc @yiyixuxu


tests/quantization/torchao/test_torchao.py
```diff
@@ -593,7 +589,7 @@ def get_dummy_inputs(self, device: torch.device, seed: int = 0):
 
     def _test_quant_type(self, quantization_config, expected_slice):
         components = self.get_dummy_components(quantization_config)
-        pipe = FluxPipeline(**components).to(dtype=torch.bfloat16)
```
a-r-r-o-w, line 596: I think this was incorrect thing to do here and it slipped past us in previous PR. We should not be calling `.to(dtype)` on the pipeline directly if there has been a model that has been quantized.

The GGUF PR introduced a check in `modeling_utils.py` [here]() that catches this behaviour.

a-r-r-o-w:
Gentle ping @DN6

DN6 (approved):

PR #10295, opened 2024-12-19
Hunyuan VAE tiling fixes and transformer docs

a-r-r-o-w:
While training loras, found some things that slipped by when we merged HunyuanVideo

- Enables gradient checkpointing in transformer
- Fixes the `encode` implementation in VAE
- Lowers the default frame tile size. A tile size of 64 requires ~24 GB memory, while a tile size of 16 requires only ~8 GB at 768x512 resolution
- Adds docstring for transformer


yiyixuxu (approved):
thanks!

a-r-r-o-w:
The test now also matches the VAE in model test with VAE in transformer test for consistency

yiyixuxu:
feel free to merge!

PR #10304, opened 2024-12-19
[Tests] QoL improvements to the LoRA test suite

sayakpaul:
# What does this PR do?

* Removes `test_lora_fuse_nan()` from the LTX LoRA test suite as that is handled in `tests/lora/utils.py`. 
* Moves `test_lora_B_bias` and `test_correct_lora_configs_with_different_ranks` to `utils.py` from the Control LoRA tests for Flux. 
* Other misc things.

### DIFF
--- a/tests/lora/test_lora_layers_ltx_video.py
+++ b/tests/lora/test_lora_layers_ltx_video.py
@@ -15,8 +15,6 @@
 import sys
 import unittest
 
-import numpy as np
-import pytest
 import torch
 from transformers import AutoTokenizer, T5EncoderModel
 
@@ -26,18 +24,12 @@ from diffusers import (
     LTXPipeline,
     LTXVideoTransformer3DModel,
 )
-from diffusers.utils.testing_utils import (
-    floats_tensor,
-    is_torch_version,
-    require_peft_backend,
-    skip_mps,
-    torch_device,
-)
+from diffusers.utils.testing_utils import floats_tensor, require_peft_backend
 
 
 sys.path.append(".")
 
-from utils import PeftLoraLoaderMixinTests, check_if_lora_correctly_set  # noqa: E402
+from utils import PeftLoraLoaderMixinTests  # noqa: E402
 
 
 @require_peft_backend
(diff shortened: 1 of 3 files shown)


tests/lora/utils.py
```diff
@@ -89,12 +89,12 @@ class PeftLoraLoaderMixinTests:
...
-    text_encoder_3_cls, text_encoder_3_id, text_encoder_3_subfolder = None, None, None
-    tokenizer_cls, tokenizer_id, tokenizer_subfolder = None, None, None
-    tokenizer_2_cls, tokenizer_2_id, tokenizer_2_subfolder = None, None, None
-    tokenizer_3_cls, tokenizer_3_id, tokenizer_3_subfolder = None, None, None
+    text_encoder_cls, text_encoder_id, text_encoder_subfolder = None, None, ""
+    text_encoder_2_cls, text_encoder_2_id, text_encoder_2_subfolder = None, None, ""
+    text_encoder_3_cls, text_encoder_3_id, text_encoder_3_subfolder = None, None, ""
+    tokenizer_cls, tokenizer_id, tokenizer_subfolder = None, None, ""
+    tokenizer_2_cls, tokenizer_2_id, tokenizer_2_subfolder = None, None, ""
+    tokenizer_3_cls, tokenizer_3_id, tokenizer_3_subfolder = None, None, ""
```
sayakpaul, line 97: Otherwise, we get 

```bash
E       TypeError: expected str, bytes or os.PathLike object, not NoneType
```

tests/lora/test_lora_layers_ltx_video.py
```diff
@@ -107,41 +99,6 @@ def get_dummy_inputs(self, with_generator=True):
 
         return noise, input_ids, pipeline_inputs
 
-    @skip_mps
-    @pytest.mark.xfail(
-        condition=torch.device(torch_device).type == "cpu" and is_torch_version(">=", "2.5"),
-        reason="Test currently fails on CPU and PyTorch 2.5.1 but not on PyTorch 2.4.1.",
-        strict=True,
-    )
-    def test_lora_fuse_nan(self):
```
a-r-r-o-w, line 116: Looks good now that the attn1 case is supported in the lora tests directly!

a-r-r-o-w (approved):
LGTM! The test for lora_B bias is moved under `utils.py` which means it will be run with every model now, no? This is okay, but maybe good to make note that it only matters for Flux currently as there are no other lora checkpoints with lora B bias yet I think

PR #10254, opened 2024-12-17
[LoRA] Support HunyuanVideo

SHYuanBest:
# What does this PR do?

Finetuning script on-the-way.

- [ ] The current code requires high video memory, and even an 80G graphics card may not be able to run it.
- [ ] The current code encounter erro in `enable gradient_checkpointing`.
- [ ] Need someone to continuous optimize and verify this code

To run training, use:

```
#!/bin/bash
# CUDA_VISIBLE_DEVICES=0
export WANDB_MODE="offline"
export MODEL_PATH="tencent/HunyuanVideo"
export DATASET_PATH="Disney-VideoGeneration-Dataset"
export OUTPUT_PATH="hunyuanvideo-lora-single-node"
export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True

# if you are not using wth 8 gus, change `accelerate_config_machine_single.yaml` num_processes as your gpu number
accelerate launch --config_file accelerate_config_machine_single.yaml \
  train.py \
  --gradient_checkpointing \
  --pretrained_model_name_or_path $MODEL_PATH \
  --enable_tiling \
  --enable_slicing \
  --instance_data_root $DATASET_PATH \
  --caption_column prompt_1.txt \
  --video_column videos_1.txt \
  --validation_prompt "DISNEY A black and white animated scene unfolds with an anthropomorphic goat surrounded by musical notes and symbols, suggesting a playful environment. Mickey Mouse appears, leaning forward in curiosity as the goat remains still. The goat then engages with Mickey, who bends down to converse or react. The dynamics shift as Mickey grabs the goat, potentially in surprise or playfulness, amidst a minimalistic background. The scene captures the evolving relationship between the two characters in a whimsical, animated setting, emphasizing their interactions and emotions:::A panda, dressed in a small, red jacket and a tiny hat, sits on a wooden stool in a serene bamboo forest. The panda's fluffy paws strum a miniature acoustic guitar, producing soft, melodic tunes. Nearby, a few other pandas gather, watching curiously and some clapping in rhythm. Sunlight filters through the tall bamboo, casting a gentle glow on the scene. The panda's face is expressive, showing concentration and joy as it plays. The background includes a small, flowing stream and vibrant green foliage, enhancing the peaceful and magical atmosphere of this unique musical performance" \
  --validation_prompt_separator ::: \
  --num_validation_videos 1 \
  --validation_epochs 100 \
  --seed 42 \
  --rank 128 \
  --lora_alpha 64 \
[...]

PR #10307, opened 2024-12-19
Fix failing lora tests after HunyuanVideo lora

a-r-r-o-w:
Sorry :/ I misread the failing tests. There were two failing tests in https://github.com/huggingface/diffusers/pull/10254 - one from wrong lora test changes and one from timeout. I only saw the timeout ones and missed the lora related ones, so CI was in a broken state.

The correct fix is to default the `subfolder` parameter to an empty string


yiyixuxu (approved):
thanks!

PR #10340, opened 2024-12-22
[Sana bug] bug fix for 2K model config

lawrence-cj:
# What does this PR do?
This PR fix the Positional Embedding missing bug. We re-introduce PE into > 2K resolution models. So we need to add PE back in `SanaTransformer2DModel`

Cc: @yiyixuxu @sayakpaul @a-r-r-o-w , thanks in advance for merging this PR to fix the bug.

Before the PR:
image

After the PR:
sana2

### DIFF
--- a/tests/models/test_modeling_common.py
+++ b/tests/models/test_modeling_common.py
@@ -1013,6 +1013,8 @@ class ModelTesterMixin:
         base_output = model(**inputs_dict)
 
         model_size = compute_module_sizes(model)[""]
+        buffer_size = compute_module_sizes(model, buffers_only=True)[""]
+        model_size = model_size - buffer_size
         # We test several splits of sizes to make sure it works.
         max_gpu_sizes = [int(p * model_size) for p in self.model_split_percents[1:]]
         with tempfile.TemporaryDirectory() as tmp_dir:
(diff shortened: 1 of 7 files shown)


nitinmukesh:
Adding comment to get notification when it's merged.

yiyixuxu (approved):

tests/models/test_modeling_common.py
```diff
@@ -1042,7 +1110,7 @@ def test_disk_offload_without_safetensors(self):
         torch.manual_seed(0)
         base_output = model(**inputs_dict)
 
-        model_size = compute_module_sizes(model)[""]
+        model_size = compute_module_persistent_sizes(model)[""]
```
yiyixuxu, line 1113: we estimate the checkpoint size here, so should not include non-persistent buffers; technically we could pass `remove_non_persistent` to the `named_module_tensors` here but it will require a PR to accelerate so adding a function here for now
https://github.com/huggingface/accelerate/blob/200c9eb7833cfa505907f6f224ebf5a275aa6d92/src/accelerate/utils/modeling.py#L724

DN6 (approved):

tests/models/test_modeling_common.py
lawrence-cj: Is there anything I need to do here? I'm not familiar with what you said above. 🤔

PR #10366, opened 2024-12-24
fix offload gpu tests etc

yiyixuxu:
this PR:
1. fix the gpu offload tests
2. refator sana tranformer so it can work with device map "auto"

### DIFF
--- a/src/diffusers/models/transformers/sana_transformer.py
+++ b/src/diffusers/models/transformers/sana_transformer.py
@@ -86,12 +86,12 @@ class SanaModulatedNorm(nn.Module):
     def __init__(self, dim: int, elementwise_affine: bool = False, eps: float = 1e-6):
         super().__init__()
         self.norm = nn.LayerNorm(dim, elementwise_affine=elementwise_affine, eps=eps)
-    
-    def forward(self, hidden_states: torch.Tensor, temb: torch.Tensor, scale_shift_table: torch.Tensor) -> torch.Tensor:
+
+    def forward(
+        self, hidden_states: torch.Tensor, temb: torch.Tensor, scale_shift_table: torch.Tensor
+    ) -> torch.Tensor:
         hidden_states = self.norm(hidden_states)
-        shift, scale = (
-            scale_shift_table[None] + temb[:, None].to(scale_shift_table.device)
-        ).chunk(2, dim=1)
+        shift, scale = (scale_shift_table[None] + temb[:, None].to(scale_shift_table.device)).chunk(2, dim=1)
         hidden_states = hidden_states * (1 + scale) + shift
         return hidden_states
 
@@ -235,7 +235,7 @@ class SanaTransformer2DModel(ModelMixin, ConfigMixin, PeftAdapterMixin):
     """
 
     _supports_gradient_checkpointing = True
-    _no_split_modules = ["SanaTransformerBlock", "PatchEmbed"]
+    _no_split_modules = ["SanaTransformerBlock", "PatchEmbed", "SanaModulatedNorm"]
 
     @register_to_config
     def __init__(
--- a/src/diffusers/models/transformers/sana_transformer.py
+++ b/src/diffusers/models/transformers/sana_transformer.py
@@ -82,6 +82,20 @@ class GLUMBConv(nn.Module):
         return hidden_states
 
 
+class SanaModulatedNorm(nn.Module):
+    def __init__(self, dim: int, elementwise_affine: bool = False, eps: float = 1e-6):
+        super().__init__()
+        self.norm = nn.LayerNorm(dim, elementwise_affine=elementwise_affine, eps=eps)
+    
+    def forward(self, hidden_states: torch.Tensor, temb: torch.Tensor, scale_shift_table: torch.Tensor) -> torch.Tensor:
+        hidden_states = self.norm(hidden_states)
+        shift, scale = (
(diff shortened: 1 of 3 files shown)


src/diffusers/models/transformers/sana_transformer.py
```diff
@@ -462,13 +475,8 @@ def custom_forward(*inputs):
                 )
 
         # 3. Normalization
-        shift, scale = (
```
yiyixuxu, line 465: cc @a-r-r-o-w 

any modular with this code pattern means that  it cannot be split (with the `device_map = "auto"`); so we better pack it into its won module 

this fails in `main`
```import torch
from diffusers import SanaPipeline, SanaTransformer2DModel

branch = "main"
repo = "Efficient-Large-Model"
ckpt = "Sana_1600M_1024px_BF16_diffusers"

transformer = SanaTransformer2DModel.from_pretrained(
    f"{repo}/{ckpt}", torch_dtype=torch.bfloat16, subfolder="transformer", device_map="auto", max_memory={0: "1GiB", 1: "10GiB"}
)
print(f" transformer.hf_device_map: {transformer.hf_device_map}")
pipe = SanaPipeline.from_pretrained(
    f"{repo}/{ckpt}", torch_dtype=torch.float32, transformer=transformer, device_map="balanced",
)
# pipe.to("cuda")
print(f" pipe.hf_device_map: {pipe.hf_device_map}")
generator = torch.Generator(device="cuda").manual_seed(42)
image = pipe(
    prompt="a cyberpunk cat with a neon sign that says 'Sana'",
    generator=generator,
)[0]
image[0].save(f"yiyi_test_2_output_{ckpt}_{branch}.png")
```

tests/models/test_modeling_common.py
```diff
@@ -1080,7 +1080,7 @@ def test_cpu_offload(self):
         torch.manual_seed(0)
         base_output = model(**inputs_dict)
 
-        model_size = compute_module_persistent_sizes(model)[""]
+        model_size = compute_module_sizes(model)[""]
```
yiyixuxu, line 1083: follow up fix for this https://github.com/huggingface/diffusers/pull/10340/files#r1895134336

src/diffusers/models/transformers/sana_transformer.py
```diff
@@ -288,8 +302,7 @@ def __init__(
 
         # 4. Output blocks
         self.scale_shift_table = nn.Parameter(torch.randn(2, inner_dim) / inner_dim**0.5)
-
-        self.norm_out = nn.LayerNorm(inner_dim, elementwise_affine=False, eps=1e-6)
+        self.norm_out = SanaModulatedNorm(inner_dim, elementwise_affine=False, eps=1e-6)
```
a-r-r-o-w, line 305: This change will affect the state_dict, no? Instead of the key being `transformer_blocks.{x}.norm_out`, it will now be `transformer_blocks.{x}.norm_out.norm`

src/diffusers/models/transformers/sana_transformer.py
a-r-r-o-w: Thanks, will keep this in mind in future integrations

src/diffusers/models/transformers/sana_transformer.py
```diff
@@ -82,6 +82,20 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+class SanaModulatedNorm(nn.Module):
+    def __init__(self, dim: int, elementwise_affine: bool = False, eps: float = 1e-6):
+        super().__init__()
+        self.norm = nn.LayerNorm(dim, elementwise_affine=elementwise_affine, eps=eps)
+
+    def forward(
+        self, hidden_states: torch.Tensor, temb: torch.Tensor, scale_shift_table: torch.Tensor
+    ) -> torch.Tensor:
+        hidden_states = self.norm(hidden_states)
+        shift, scale = (scale_shift_table[None] + temb[:, None].to(scale_shift_table.device)).chunk(2, dim=1)
```
a-r-r-o-w, line 94: Not really a fan of this kind of device casting in `forward` but okay to keep it since we don't have better solution yet. These usually end up creating problems for anything that modifies device/dtype with hooks and we then have to use some workarounds.

Going forward, I think `nn.Parameter`'s can be put in their own dummy `nn.Module` so that device map, or other things we're introducing (like group offloading or fp8 layerwise upcasting), works out of the box (as they will handle the weight/type-casting of inputs in overwritten pre-hook methods). If this sounds good, will do future model integrations with this design

a-r-r-o-w (approved):
LGTM

src/diffusers/models/transformers/sana_transformer.py
a-r-r-o-w: Oh sorry, just saw that elementwise_affine=False, so there will be no state dict changes. Looks good 👍

src/diffusers/models/transformers/sana_transformer.py
yiyixuxu: ohh I actually did not think about this at all (I just copied from the original code) - could you explain why do we need this device casting here?

a-r-r-o-w: Ah okay, I see. I think I missed it when reviewing the PR that added Sana, otherwise would have probably removed it then. I'm not really sure why it is needed here, and think it might be okay to remove

DN6 (approved):

PR #10384, opened 2024-12-25
flux: make scheduler config params optional

vladmandic:
now that we have flow support for most of the schedulers, we really should not rely on hardcoded optional scheduler config values inside the pipeline - they should be optional.

without this change, flux pipeline will fail for most non-default schedulers (with flow mode enabled) due to missing config value.

note that this change needs to be mirrored for all flux pipelines, but due to how diffusers ci works, i'm really not sure what needs to be manually updated vs not.

@hlky @yiyixuxu


src/diffusers/pipelines/flux/pipeline_flux.py
```diff
@@ -73,11 +74,24 @@
...
+        deprecate(
+            "calculate_shift scheduler",
+            "1.0.0",
+            deprecation_message,
+            standard_warn=False,
+        )
+    base_seq_len = base_seq_len or scheduler.config.get("base_image_seq_len", 256)
+    max_seq_len = max_seq_len or scheduler.config.get("max_image_seq_len", 4096)
+    base_shift = base_shift or scheduler.config.get("base_shift", 0.5)
+    max_shift = max_shift or scheduler.config.get("max_shift", 1.16)
```
hlky, line 94: Thanks for the contribution @vladmandic, it will be fun to try to those scheduler changes with Flux.

As you mentioned CI, I've ran `make fix-copies` to demonstrate. Every occurrence of `# Copied from diffusers.pipelines.flux.pipeline_flux.calculate_shift` is updated, this highlights the effects in our own codebase, we'll need to update the call of `calculate_shift` in every pipeline and there can be effects downstream, so I've pushed this version which adds a deprecation message. We'll run `make fix-copies` again when the changes here are finalised.

vladmandic: thanks!

examples/community/pipeline_flux_rf_inversion.py
```diff
@@ -992,10 +1003,7 @@ def invert(
         image_seq_len = (int(height) // self.vae_scale_factor // 2) * (int(width) // self.vae_scale_factor // 2)
         mu = calculate_shift(
             image_seq_len,
-            self.scheduler.config.base_image_seq_len,
-            self.scheduler.config.max_image_seq_len,
-            self.scheduler.config.base_shift,
-            self.scheduler.config.max_shift,
+            scheduler=self.scheduler,
```
yiyixuxu, line 1006:
```suggestion
            scheduler.config.get("base_image_seq_len", 256)
            scheduler.config.get("max_image_seq_len", 4096)
            scheduler.config.get("base_shift", 0.5)
            scheduler.config.get("max_shift", 1.16)
```

examples/community/pipeline_flux_rf_inversion.py
```diff
@@ -91,11 +92,24 @@
 # Copied from diffusers.pipelines.flux.pipeline_flux.calculate_shift
 def calculate_shift(
     image_seq_len,
-    base_seq_len: int = 256,
-    max_seq_len: int = 4096,
-    base_shift: float = 0.5,
-    max_shift: float = 1.16,
+    base_seq_len: Optional[int] = 256,
```
yiyixuxu, line 95: I think we don't need to change this function, we can just change how the parameters are passed from the pipeline

yiyixuxu (commented):
thanks! I left a comment

hlky (approved):

yiyixuxu:
thanks @vladmandic @hlky !

PR #10388, opened 2024-12-26
Fix Flux multiple Lora loading bug

maxs-kan:
# What does this PR do?

The current approach of checking for a key with a `base_layer` suffix may lead to a bug when multiple Lora models are loaded. If the first loaded Lora model does not have weights for layer `n`, and the second one does, loading the second model will lead to an error since the transformer state dict currently does not have key `n.base_layer.weight`. So I explicitly check for the presence of a key with the `base_layer` suffix.

@yiyixuxu

### DIFF
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -2460,13 +2460,17 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
         if unexpected_modules:
             logger.debug(f"Found unexpected modules: {unexpected_modules}. These will be ignored.")
 
-        is_peft_loaded = getattr(transformer, "peft_config", None) is not None
+        transformer_base_layer_keys = {
+            k[: -len(".base_layer.weight")] for k in transformer_state_dict.keys() if ".base_layer.weight" in k
+        }
         for k in lora_module_names:
             if k in unexpected_modules:
                 continue
 
             base_param_name = (
-                f"{k.replace(prefix, '')}.base_layer.weight" if is_peft_loaded else f"{k.replace(prefix, '')}.weight"
+                f"{k.replace(prefix, '')}.base_layer.weight"
+                if k in transformer_base_layer_keys
+                else f"{k.replace(prefix, '')}.weight"
             )
             base_weight_param = transformer_state_dict[base_param_name]
             lora_A_param = lora_state_dict[f"{prefix}{k}.lora_A.weight"]
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -2460,18 +2460,15 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
         if unexpected_modules:
             logger.debug(f"Found unexpected modules: {unexpected_modules}. These will be ignored.")
 
-        transformer_base_layer_keys = {
-            k[: -len(".base_layer.weight")] for k in transformer_state_dict.keys() if ".base_layer.weight" in k
-        }
+        is_peft_loaded = getattr(transformer, "peft_config", None) is not None
         for k in lora_module_names:
             if k in unexpected_modules:
                 continue
 
-            base_param_name = (
-                f"{k.replace(prefix, '')}.base_layer.weight"
-                if k in transformer_base_layer_keys
-                else f"{k.replace(prefix, '')}.weight"
-            )
+            base_param_name = f"{k.replace(prefix, '')}.weight"
+            base_layer_name = f"{k.replace(prefix, '')}.base_layer.weight"
+            if is_peft_loaded and base_layer_name in transformer_state_dict:
+                base_param_name = base_layer_name
             base_weight_param = transformer_state_dict[base_param_name]
             lora_A_param = lora_state_dict[f"{prefix}{k}.lora_A.weight"]
 
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -2465,10 +2465,11 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
             if k in unexpected_modules:
                 continue
 
-            base_param_name = f"{k.replace(prefix, '')}.weight"
-            base_layer_name = f"{k.replace(prefix, '')}.base_layer.weight"
-            if is_peft_loaded and base_layer_name in transformer_state_dict:
-                base_param_name = base_layer_name
+            base_param_name = (
+                f"{k.replace(prefix, '')}.base_layer.weight"
+                if is_peft_loaded and f"{k.replace(prefix, '')}.base_layer.weight" in transformer_state_dict
+                else f"{k.replace(prefix, '')}.weight"
+            )
             base_weight_param = transformer_state_dict[base_param_name]
             lora_A_param = lora_state_dict[f"{prefix}{k}.lora_A.weight"]
 
--- a/tests/lora/test_lora_layers_flux.py
+++ b/tests/lora/test_lora_layers_flux.py
@@ -251,14 +251,14 @@ class FluxLoRATests(unittest.TestCase, PeftLoraLoaderMixinTests):
 
         pipe.set_adapters(["one", "two"])
         self.assertTrue(check_if_lora_correctly_set(pipe.transformer), "Lora not correctly set in transformer")
-        images_lora_with_absent_keys = pipe(**inputs, generator=torch.manual_seed(0)).images
+        images_lora_with_extra_keys = pipe(**inputs, generator=torch.manual_seed(0)).images
 
         self.assertFalse(
-            np.allclose(images_lora, images_lora_with_absent_keys, atol=1e-3, rtol=1e-3),
+            np.allclose(images_lora, images_lora_with_extra_keys, atol=1e-3, rtol=1e-3),
             "Different LoRAs should lead to different results.",
         )
         self.assertFalse(
-            np.allclose(output_no_lora, images_lora_with_absent_keys, atol=1e-3, rtol=1e-3),
+            np.allclose(output_no_lora, images_lora_with_extra_keys, atol=1e-3, rtol=1e-3),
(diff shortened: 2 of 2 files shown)


hlky:
Hi @maxs-kan, thanks for your contribution, can you share some example lora checkpoints that may lead to a bug?

maxs-kan:
Sure, try in same order:
`pipe.load_lora_weights(hf_hub_download("TTPlanet/Migration_Lora_flux","Migration_Lora_cloth.safetensors"), adapter_name="cloth")`
`pipe.load_lora_weights("alimama-creative/FLUX.1-Turbo-Alpha", adapter_name="turbo")`

hlky (approved):
<details>
  <summary>Code</summary>

  ```python
import torch

)
pipe.load_lora_weights(
    hf_hub_download("TTPlanet/Migration_Lora_flux", "Migration_Lora_cloth.safetensors"),
    adapter_name="cloth",
)
pipe.load_lora_weights("alimama-creative/FLUX.1-Turbo-Alpha", adapter_name="turbo")
```
</details>

File: src/diffusers/loaders/lora_pipeline.py [hunk 1]
```diff
@@ -2460,13 +2460,17 @@ def _maybe_expand_lora_state_dict(cls, transformer, lora_state_dict):
         if unexpected_modules:
             logger.debug(f"Found unexpected modules: {unexpected_modules}. These will be ignored.")
 
-        is_peft_loaded = getattr(transformer, "peft_config", None) is not None
+        transformer_base_layer_keys = {
+            k[: -len(".base_layer.weight")] for k in transformer_state_dict.keys() if ".base_layer.weight" in k
+        }
```
sayakpaul, line 2465: Note `base_layer` substring can only be present when the underlying pipeline has at least one LoRA loaded that affects the layer under consideration. So, perhaps it's better to have an `is_peft_loaded` check?

sayakpaul: In your PR description you mention:

> If the first loaded Lora model does not have weights for layer n, and the second one does, loading the second model will lead to an error since the transformer state dict currently does not have key n.base_layer.weight.

Note that we may also have an opposite situation i.e., the first LoRA ckpt may have the params while the second LoRA may not. This is what I considered in https://github.com/huggingface/diffusers/pull/10388.

sayakpaul:
Also, I gave @hlky's [code snippet here](https://github.com/huggingface/diffusers/pull/10388#pullrequestreview-2523869439) a try in https://github.com/huggingface/diffusers/pull/10396/ branch and it seems to work.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2460,13 +2460,17 @@ def _maybe_expand_lora_state_dict(cls, transformer, lora_state_dict):
...
         for k in lora_module_names:
             if k in unexpected_modules:
                 continue
 
             base_param_name = (
-                f"{k.replace(prefix, '')}.base_layer.weight" if is_peft_loaded else f"{k.replace(prefix, '')}.weight"
+                f"{k.replace(prefix, '')}.base_layer.weight"
+                if k in transformer_base_layer_keys
+                else f"{k.replace(prefix, '')}.weight"
             )
```
hlky, line 2474:
```python
base_param_name = f"{k.replace(prefix, '')}.weight"
base_layer_name = f"{k.replace(prefix, '')}.base_layer.weight"
if is_peft_loaded and base_layer_name in transformer_state_dict:
    base_param_name = base_layer_name
```
Something like this might be better.

src/diffusers/loaders/lora_pipeline.py
sayakpaul: In this case, we are considering that LoRA params for certain modules exist in the first checkpoint while they don't exist in the second checkpoint (or any other subsequent checkpoint). 

In this case, we don't want to expand no? Or am I missing something? Perhaps better expressed through a short test case like the one I added [here](https://github.com/huggingface/diffusers/pull/10396)?

hlky: The [test case](https://github.com/huggingface/diffusers/blob/e99735c0557ef999f84ed6c319a9d23a2f0b496e/tests/lora/test_lora_layers_flux.py#L166) passes on `main`, the test case should be in the reverse order:

```python
        with tempfile.TemporaryDirectory() as tmpdirname:
            denoiser_state_dict = get_peft_model_state_dict(pipe.transformer)
            self.pipeline_class.save_lora_weights(tmpdirname, transformer_lora_layers=denoiser_state_dict)

            self.assertTrue(os.path.isfile(os.path.join(tmpdirname, "pytorch_lora_weights.safetensors")))
            pipe.unload_lora_weights()
            # Modify the state dict to exclude "x_embedder" related LoRA params.
            lora_state_dict = safetensors.torch.load_file(os.path.join(tmpdirname, "pytorch_lora_weights.safetensors"))
            lora_state_dict_without_xembedder = {k: v for k, v in lora_state_dict.items() if "x_embedder" not in k}
            pipe.load_lora_weights(lora_state_dict_without_xembedder, adapter_name="two")

            # Load state dict with `x_embedder`.
            pipe.load_lora_weights(os.path.join(tmpdirname, "pytorch_lora_weights.safetensors"), adapter_name="one")
```

```python
            base_param_name = (
                f"{k.replace(prefix, '')}.base_layer.weight" if is_peft_loaded else f"{k.replace(prefix, '')}.weight"
            )
>           base_weight_param = transformer_state_dict[base_param_name]
E           KeyError: 'x_embedder.base_layer.weight'

src\diffusers\loaders\lora_pipeline.py:2471: KeyError
```

I think we still want to check whether the param needs to be expanded

sayakpaul: Cool, I understand it better now. Thanks!

Might be better to ship this PR with proper testing then. Okay with me.

sayakpaul (commented):
@hlky thanks!

Do you wanna propagate your suggestions, too?

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2465,9 +2465,10 @@ def _maybe_expand_lora_state_dict(cls, transformer, lora_state_dict):
             if k in unexpected_modules:
                 continue
 
-            base_param_name = (
-                f"{k.replace(prefix, '')}.base_layer.weight" if is_peft_loaded else f"{k.replace(prefix, '')}.weight"
-            )
+            base_param_name = f"{k.replace(prefix, '')}.weight"
+            base_layer_name = f"{k.replace(prefix, '')}.base_layer.weight"
+            if is_peft_loaded and base_layer_name in transformer_state_dict:
+                base_param_name = base_layer_name
```
hlky, line 2471: Could be
```python
            base_param_name = (
                f"{k.replace(prefix, '')}.base_layer.weight"
                if is_peft_loaded and f"{k.replace(prefix, '')}.base_layer.weight" in transformer_state_dict
                else f"{k.replace(prefix, '')}.weight"
            )
```

tests/lora/test_lora_layers_flux.py
```diff
@@ -162,6 +163,56 @@ def test_with_alpha_in_state_dict(self):
         )
         self.assertFalse(np.allclose(images_lora_with_alpha, images_lora, atol=1e-3, rtol=1e-3))
 
+    def test_lora_expansion_works_for_absent_keys(self):
```
hlky, line 166: `test_lora_expansion_works_for_extra_keys`? we could add the original case `test_lora_expansion_works_for_absent_keys`

hlky (approved):
Thanks @sayakpaul. I've left some other comments but should be good to go

Thanks for the PR @maxs-kan

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Yes this is better. Let's go with this.

tests/lora/test_lora_layers_flux.py
sayakpaul: That works for me.

tests/lora/test_lora_layers_flux.py
```diff
@@ -162,6 +163,56 @@ def test_with_alpha_in_state_dict(self):
...
+
+            self.assertTrue(os.path.isfile(os.path.join(tmpdirname, "pytorch_lora_weights.safetensors")))
+            pipe.unload_lora_weights()
+            # Modify the state dict to exclude "x_embedder" related LoRA params.
+            lora_state_dict = safetensors.torch.load_file(os.path.join(tmpdirname, "pytorch_lora_weights.safetensors"))
+            lora_state_dict_without_xembedder = {k: v for k, v in lora_state_dict.items() if "x_embedder" not in k}
+            pipe.load_lora_weights(lora_state_dict_without_xembedder, adapter_name="two")
+
+            # Load state dict with `x_embedder`.
+            pipe.load_lora_weights(os.path.join(tmpdirname, "pytorch_lora_weights.safetensors"), adapter_name="one")
```
sayakpaul, line 201:
```suggestion
            lora_state_dict_without_xembedder = {k: v for k, v in lora_state_dict.items() if "x_embedder" not in k}
            pipe.load_lora_weights(lora_state_dict_without_xembedder, adapter_name="one")

            # Load state dict with `x_embedder`.
            pipe.load_lora_weights(os.path.join(tmpdirname, "pytorch_lora_weights.safetensors"), adapter_name="two")
```

sayakpaul (approved):
Thanks. 

I think the test mimics the code [here](https://github.com/huggingface/diffusers/pull/10388#pullrequestreview-2523869439) that was producing the error. So, I think we should be good to go.

yiyixuxu:
thank you all! @maxs-kan @hlky @sayakpaul

bghira:
this is a pretty big regression that caused some consumers to need to pull a custom build of diffusers with this patch included. can there perhaps be a hotfix pushed for v0.32.2?

sayakpaul:
Sorry this should go into a patch release. @yiyixuxu I am happy to do the patch release if you're okay.

PR #10393, opened 2024-12-27
Default values in SD3 pipelines when submodules are not loaded

hlky:
# What does this PR do?

Values like `vae_scale_factor` in pipelines are derived from configurations of component modules e.g. `self.vae_scale_factor = 2 ** (len(self.vae.config.block_out_channels) - 1)`, some pipelines are missing default values when the module is not loaded

```python
self.vae_scale_factor = (
    2 ** (len(self.vae.config.block_out_channels) - 1) if hasattr(self, "vae") and self.vae is not None else 8
)
```

This PR adds default values in SD3 pipelines.

Reproduction
```python
from diffusers import StableDiffusion3Img2ImgPipeline
import torch

pipe = StableDiffusion3Img2ImgPipeline.from_pretrained(
    "stabilityai/stable-diffusion-3.5-large",
    vae=None, # or `tokenizer` or `transformer`
    torch_dtype=torch.bfloat16,
)
```

## Who can review?


yiyixuxu (approved):
thank you!

PR #10397, opened 2024-12-27
[LoRA] fix: lora unloading when using expanded Flux LoRAs.

sayakpaul:
# What does this PR do?

https://github.com/huggingface/diffusers/pull/10206 added support for properly unloading Control LoRA weights from Flux and allowing the pipeline to be used with `from_pipe()`. 

However, this led to the following problem. Say, if we do:

```py
...
pipeline.load_lora_weights("black-forest-labs/FLUX.1-Canny-dev-lora")
pipeline.fuse_lora()
pipeline.unload_lora_weights()
...
```

[This](https://github.com/huggingface/diffusers/blob/83da817f73776aa61086683f55432bf2915d0748/src/diffusers/loaders/lora_pipeline.py#L2324C21-L2331C26) would revert the `transformer` config back to the original (`in_channels=64`), rendering [this line](https://github.com/huggingface/diffusers/blob/83da817f73776aa61086683f55432bf2915d0748/src/diffusers/pipelines/flux/pipeline_flux_control.py#L764) to output 8 (whereas it should be 16). This is fine for use cases https://github.com/huggingface/diffusers/pull/10206 served but it's not effective for the above use case. 

This went uncaught because we didn't have a proper test for it and I didn't run the integration tests for Control LoRA. Sorry about that. 

This PR fixes this. To do that, I am adding a flag `reset_to_overwritten_params` (not fixated on the name) to the `unload_lora_weights()` of `FluxLoraLoaderMixin` and suggesting to add a note about it in the docs.

### DIFF
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -2277,8 +2277,24 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
 
         super().unfuse_lora(components=components)
 
-    # We override this here account for `_transformer_norm_layers`.
+    # We override this here account for `_transformer_norm_layers` and `_overwritten_params`.
     def unload_lora_weights(self, reset_to_overwritten_params=False):
+        """
+        Unloads the LoRA parameters.
+
+        Args:
+            reset_to_overwritten_params (`bool`, defaults to `False`): Whether to reset the LoRA-loaded modules
+                to their original params. Refer to the [Flux
+                documentation](https://huggingface.co/docs/diffusers/main/en/api/pipelines/flux) to learn more.
+
+        Examples:
+
+        ```python
+        >>> # Assuming `pipeline` is already loaded with the LoRA parameters.
+        >>> pipeline.unload_lora_weights()
+        >>> ...
+        ```
+        """
         super().unload_lora_weights()
 
         transformer = getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -2278,7 +2278,7 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
         super().unfuse_lora(components=components)
 
     # We override this here account for `_transformer_norm_layers`.
-    def unload_lora_weights(self, reset_to_overwrriten_params=False):
+    def unload_lora_weights(self, reset_to_overwritten_params=False):
         super().unload_lora_weights()
 
         transformer = getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer
@@ -2286,7 +2286,7 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
             transformer.load_state_dict(transformer._transformer_norm_layers, strict=False)
             transformer._transformer_norm_layers = None
 
-        if reset_to_overwrriten_params and getattr(transformer, "_overwritten_params", None) is not None:
+        if reset_to_overwritten_params and getattr(transformer, "_overwritten_params", None) is not None:
             overwritten_params = transformer._overwritten_params
             module_names = set()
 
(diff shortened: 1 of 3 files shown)


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2278,15 +2278,15 @@ def unfuse_lora(self, components: List[str] = ["transformer", "text_encoder"], *
         super().unfuse_lora(components=components)
 
     # We override this here account for `_transformer_norm_layers`.
-    def unload_lora_weights(self):
+    def unload_lora_weights(self, reset_to_overwrriten_params=False):
```
sayakpaul, line 2281: Not too fixated on the arg name. 

Open to also tackling the problem in other better ways.

a-r-r-o-w, line 2281:
```suggestion
    def unload_lora_weights(self, reset_to_overwritten_params=False):
```

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2278,15 +2278,15 @@ def unfuse_lora(self, components: List[str] = ["transformer", "text_encoder"], *
...
         super().unload_lora_weights()
 
         transformer = getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer
         if hasattr(transformer, "_transformer_norm_layers") and transformer._transformer_norm_layers:
             transformer.load_state_dict(transformer._transformer_norm_layers, strict=False)
             transformer._transformer_norm_layers = None
 
-        if getattr(transformer, "_overwritten_params", None) is not None:
+        if reset_to_overwrriten_params and getattr(transformer, "_overwritten_params", None) is not None:
```
a-r-r-o-w, line 2289:
```suggestion
        if reset_to_overwritten_params and getattr(transformer, "_overwritten_params", None) is not None:
```

tests/lora/test_lora_layers_flux.py
```diff
@@ -606,7 +606,7 @@ def test_lora_unload_with_parameter_expanded_shapes(self):
         self.assertTrue(pipe.transformer.config.in_channels == 2 * in_features)
         self.assertTrue(cap_logger.out.startswith("Expanding the nn.Linear input/output features for module"))
 
-        control_pipe.unload_lora_weights()
+        control_pipe.unload_lora_weights(reset_to_overwrriten_params=True)
```
a-r-r-o-w, line 609:
```suggestion
        control_pipe.unload_lora_weights(reset_to_overwritten_params=True)
```

tests/lora/test_lora_layers_flux.py
```diff
@@ -624,6 +624,65 @@ def test_lora_unload_with_parameter_expanded_shapes(self):
...
+
+        inputs["control_image"] = control_image
+        lora_out = control_pipe(**inputs, generator=torch.manual_seed(0))[0]
+
+        self.assertFalse(np.allclose(original_out, lora_out, rtol=1e-4, atol=1e-4))
+        self.assertTrue(pipe.transformer.x_embedder.weight.data.shape[1] == 2 * in_features)
+        self.assertTrue(pipe.transformer.config.in_channels == 2 * in_features)
+        self.assertTrue(cap_logger.out.startswith("Expanding the nn.Linear input/output features for module"))
+
+        control_pipe.unload_lora_weights(reset_to_overwrriten_params=False)
```
a-r-r-o-w, line 675:
```suggestion
        control_pipe.unload_lora_weights(reset_to_overwritten_params=False)
```

a-r-r-o-w (approved):
Maybe a docstring for what `reset_to_overwritten_params` does and why we added it would be helpful for someone looking at it in future, but no problem if it doesn't make sense to. 

The test looks good. Would maybe check if an error is raised when trying to run inference with `reset_to_overwritten_params=True`. Off to YiYi

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2278,15 +2278,15 @@ def unfuse_lora(self, components: List[str] = ["transformer", "text_encoder"], *
         super().unfuse_lora(components=components)
 
     # We override this here account for `_transformer_norm_layers`.
-    def unload_lora_weights(self):
+    def unload_lora_weights(self, reset_to_overwritten_params=False):
```
sayakpaul, line 2281: Not too fixated on the name. Open to suggestions. 

Also open to tackling the problem in better ways.

sayakpaul:
@a-r-r-o-w docs and docstrings done.

> Would maybe check if an error is raised when trying to run inference with reset_to_overwritten_params=True

I am not sure what is meant here.

PR #10420, opened 2025-01-01
fix ltx checkpoint id.

sayakpaul:


a-r-r-o-w:
But this example is specific to LTX 0.9.1. This points to the original repo, which don't have the weights for 0.9.1, so this would be incorrect.

We do support loading from original repo with from_single_file for both 0.9.0 and 0.9.1.

We didn't settle on how to host the weights yet (conversation on Slack).

PR #9169, opened 2024-08-13
refactor SD3, speedup tests

a-r-r-o-w:
# What does this PR do?

- Refactor some of the SD3 modeling implementation. Accessing configuration parameters in `__init__` via the config attribute is not necessary or clean. Can use the parameters directly
- Improve docs
- Speedup tests

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #10423, opened 2025-01-01
Update Flux docstrings

a-r-r-o-w:
- Updates some docs and type hints
- Removes the access of attributes via config in transformer init
- torch.FloatTensor -> torch.Tensor because FloatTensor only exists for backward-compatibility reasons

### DIFF
--- a/src/diffusers/models/transformers/transformer_flux.py
+++ b/src/diffusers/models/transformers/transformer_flux.py
@@ -85,11 +85,11 @@ class FluxSingleTransformerBlock(nn.Module):
 
     def forward(
         self,
-        hidden_states: torch.FloatTensor,
-        temb: torch.FloatTensor,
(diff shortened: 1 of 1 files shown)


src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -259,39 +280,39 @@ def __init__(
     ):
         super().__init__()
         self.out_channels = out_channels or in_channels
-        self.inner_dim = self.config.num_attention_heads * self.config.attention_head_dim
```
a-r-r-o-w, line 262: don't really prefer accessing readily available init attributes via the config. I think we were okay with doing this [here](https://github.com/huggingface/diffusers/pull/9169#discussion_r1730686452)

hlky (approved):

PR #10431, opened 2025-01-02
[Sana] 1k PE bug fixed

lawrence-cj:
# What does this PR do?
This PR fix the Positional Embedding bug for Sana 1K and 2K models

Cc: @yiyixuxu @a-r-r-o-w @sayakpaul


yiyixuxu (approved):
thanks!

PR #10439, opened 2025-01-03
Use Pipelines without scheduler

hlky:
# What does this PR do?

#10438 tests `encode_prompt` independently. This revealed some pipelines that cannot be used without passing `scheduler` due to checks like `if hasattr(scheduler.config, ...`

This PR fixes the checks.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@sayakpaul

PR #10438, opened 2025-01-03
[tests] test `encode_prompt()` in isolation

sayakpaul:
# What does this PR do?

`encode_prompt()` is often used in isolation to lift off memory requirements and to also compute prompt embeddings in many trainers. It's important that we test `encode_prompt()` properly. 

This PR adds a test suite to do that. Since we cannot reliably map the outputs of `encode_prompt()` into keyword arguments of a pipeline call, I propose to use the `ast` module to determine the names of the variables returned from a particular `encode_prompt()` method. This PR is a PoC of what the changes might look like at the test level. 

The closest test we have currently is this:
https://github.com/huggingface/diffusers/blob/15d4569ec3fdd154508aef78c178d024dff7605f/tests/pipelines/test_pipelines_common.py#L2165

But, IMO, this should be extended to the other pipelines on a more general level.

## TODOs

- [x] Propagate to the rest of the pipelines once we agree on the structure.
- [x] Fix `encode_prompt()` to work in isolation in the pipelines if needed.
- [ ] Remove `SDXLOptionalComponentsTesterMixin` if needed.

PR #10440, opened 2025-01-03
Use Pipelines without unet

hlky:
# What does this PR do?

As in #10439 this PR guards some checks involving `unet` in a pipeline's `__init__` that prevent it from being used without `unet` i.e.  for text encoder only.

I'll cover default values for `vae_scale_factor` like in #10393 for all pipelines in another PR.

## Who can review?

@sayakpaul @yiyixuxu


sayakpaul (approved):
Sleek!

hlky:
Some of these checks are before `register_modules` and use `unet`, `vae` etc. directly, others are after `register_modules` and like `if hasattr(self, "unet") and self.unet is not None and ...`, it's ok to do either there shouldn't be any side effects from `register_modules` but we could consider uniformizing this.

hlky:
Oh there's actually a reason some of these `unet` and `scheduler` checks are before `register_modules`, they are changing values that in turn are registered to the module's config. Others like `unet.config.in_channels` and `default_sample_size` could be after `register_modules`.

sayakpaul:
> Oh there's actually a reason some of these `unet` and `scheduler` checks are before `register_modules`, they are changing values that in turn are registered to the module's config. Others like `unet.config.in_channels` and `default_sample_size` could be after `register_modules`.

But I guess that doesn't block this PR no?

hlky:
> But I guess that doesn't block this PR no?

We can leave reorganising as a follow-up.

sayakpaul:
You mean the uniformization? If so, yes, I think that is okay. That also doesn't block https://github.com/huggingface/diffusers/pull/10438 I think

src/diffusers/pipelines/deprecated/alt_diffusion/pipeline_alt_diffusion.py
```diff
@@ -253,10 +253,14 @@ def __init__(
                 " checker. If you do not want to use the safety checker, you can pass `'safety_checker=None'` instead."
             )
 
-        is_unet_version_less_0_9_0 = hasattr(unet.config, "_diffusers_version") and version.parse(
-            version.parse(unet.config._diffusers_version).base_version
-        ) < version.parse("0.9.0.dev0")
-        is_unet_sample_size_less_64 = hasattr(unet.config, "sample_size") and unet.config.sample_size < 64
+        is_unet_version_less_0_9_0 = is_unet_version_less_0_9_0 = (
```
yiyixuxu, line 256:
```suggestion
        is_unet_version_less_0_9_0 = (
```
it this intended here?

yiyixuxu (approved):
thanks!

PR #10441, opened 2025-01-03
Use pipelines without vae

hlky:
# What does this PR do?

As in https://github.com/huggingface/diffusers/pull/10439 and #10440 this PR guards some checks involving `vae` in a pipeline's `__init__` that prevent it from being used without `vae` i.e. for text encoder only.

This is mainly `vae_scale_factor`, the default is set to `8` as this is the most common value. `vae_scale_factor` effects several parts of the pipeline, default height/width calculation, `prepare_latents`, `check_inputs` etc. Pipelines are generally designed to allow use without loading `vae` with `output_type="latents"` for example, so it will be important to identify the pipelines where `vae_scale_factor` is not `8`.

In some Flux pipelines we add a guard for `self.vae.config.latent_channels` used in `mask_processor` with default `16`, this value is taken from Flux configs.

Checking issues with pipeline's `__call__` and `output_type="latents"` will be covered in another PR. Example in `FluxFillPipeline` and the fix already present in `FluxPipeline`:

https://github.com/huggingface/diffusers/blob/c28db0aa5bad353b8d812981d76f6b0b414aa195/src/diffusers/pipelines/flux/pipeline_flux_fill.py#L843
https://github.com/huggingface/diffusers/blob/c28db0aa5bad353b8d812981d76f6b0b414aa195/src/diffusers/pipelines/flux/pipeline_flux.py#L810

## Who can review?

@sayakpaul @yiyixuxu


examples/community/adaptive_mask_inpainting.py
```diff
@@ -450,7 +450,9 @@ def __init__(
             safety_checker=safety_checker,
             feature_extractor=feature_extractor,
         )
-        self.vae_scale_factor = 2 ** (len(self.vae.config.block_out_channels) - 1)
+        self.vae_scale_factor = (
+            2 ** (len(self.vae.config.block_out_channels) - 1) if hasattr(self, "vae") and self.vae is not None else 8
+        )
```
sayakpaul, line 455:
```suggestion
        self.vae_scale_factor = (
            2 ** (len(self.vae.config.block_out_channels) - 1) if getattr(self, "vae", None) is not None else 8
        )
```

leaner.

hlky: Thanks, I've applied this everywhere.

yiyixuxu (approved):
thanks! very nice initiative

PR #10442, opened 2025-01-03
Update and disable `push_tests_mps` workflow

hlky:
# What does this PR do?

Updates `push_tests_mps` to separate tests by category and fix some environment issues. For now a lot of failures are due to a problem with GitHub's runners so we also disable the workflow.

## Who can review?

@sayakpaul

### DIFF
--- a/.github/workflows/push_tests_mps.yml
+++ b/.github/workflows/push_tests_mps.yml
@@ -16,6 +16,8 @@ env:
   HF_HUB_ENABLE_HF_TRANSFER: 1
   PYTEST_TIMEOUT: 600
   RUN_SLOW: no
+  HF_TOKEN: ${{ secrets.HF_TOKEN }}
+  PYTORCH_MPS_HIGH_WATERMARK_RATIO: 0.0
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


sayakpaul:
Does removing them help launch the tests? Could we maybe trigger it manually?

hlky:
I don't know if we can trigger it manually. Currently the tests launch on push to `main` and eventually segfault after some scheduler tests

https://github.com/huggingface/diffusers/actions/runs/12588771861/job/35087350062#step:7:8211

Most pipeline tests seem to be failing and there are some warnings

```
There was a problem when trying to write in your cache folder (/System/Volumes/Data/mnt/cache/hub). You should set the environment variable TRANSFORMERS_CACHE to a writable directory.
There was a problem when trying to write in your cache folder (/System/Volumes/Data/mnt/cache/hub). Please, ensure the directory exists and can be written to.
```

It suggests some pipeline tests are failing because of it, but there are no tracebacks for individual test failures, probably due to the segfault.

sayakpaul:
Sorry should have been clearer.

What we could do to ensure the MPS stuff runs successfully: 

* Temporarily have a workflow in this PR with the same contents as the one in `.github/workflows/push_tests_mps.yml`.
* Have the temporary workflow triggered on PRs so that we can verify it. 
* Once we confirm things are working as expected, we delete the temporary workflow. 

Does this sound good?

sayakpaul:
Nice seems like the tests are getting picked up? We can let them run and fix anything in a follow-up PR? WDYT?

hlky:
There are still some failures so far, I'll compare to one of the last runs to check if anything is fixed. I think priority is fixing the segfault but I see no indication of anything that could be causing it. We could separate the tests like models, pipelines, schedulers, might help break it down as there are a lot of `FAILED` and a lot of `SKIPPED`.

sayakpaul:
Makes sense and okay for me to tackle that in this PR. Thanks! Let me know if I can help.

src/diffusers/schedulers/scheduling_consistency_models.py
```diff
@@ -220,7 +220,7 @@ def set_timesteps(
             self.num_inference_steps = num_inference_steps
 
             step_ratio = self.config.num_train_timesteps // self.num_inference_steps
-            timesteps = (np.arange(0, num_inference_steps) * step_ratio).round()[::-1].copy().astype(np.int64)
+            timesteps = (np.arange(0, num_inference_steps) * step_ratio).round()[::-1].copy().astype(np.int32)
```
sayakpaul, line 223: Won't it cause any numerical problems or qualitative problems?

hlky: Should be ok but I'll cherry-pick this + the other fix (https://github.com/huggingface/diffusers/pull/10442/commits/64553b36ef2fdd65d9eec9c25a8a22ddf2d7eb76) and review with @yiyixuxu

hlky: PR is now only workflow changes, we may want to just disable it until we get the new runner though.

sayakpaul: That is okay with me.

PR #9517, opened 2024-09-24
Fix the bug that `joint_attention_kwargs` is not passed to the FLUX's transformer attention processors

HorizonWind2004:
Issue link is below:

#9516

PR #10456, opened 2025-01-05
Support pass kwargs to cogvideox custom attention processor

huanngzh:
# What does this PR do?

Pass kwargs to custom attention processor is preferable when implementing layout control, attention replacing or introducing side effect to attention computation. Now this feature is supported by sd, sdxl, FLUX, sd3 series, but not CogVideoX. This PR fix the problem, and may support future training-free experiments in CogVideoX series models.

Example code:
```Python
class CustomCogVideoXAttnProcessor2_0:
    def __init__(self, **custom_args):
        super().__init__()
        # some initialization
    
    def __call__(self, ..., added_argu=None) -> torch.FloatTensor:
        # custom attention forward pass

custom_processor = CustomCogVideoXAttnProcessor2_0(**custom_args)
pipe.transformer.set_attn_processor(custom_processor) # cogvideox pipeline
attention_kwargs = {"added_argu": 100} # prepare passing added_argu to attention processor
pipe.transformer(..., attention_kwargs=attention_kwargs) # custom transformer forward pass
```

## Before submitting

## Who can review?

Hello @yiyixuxu @sayakpaul, I just want to do the same to CogVideoX as to FLUX in https://github.com/huggingface/diffusers/pull/9517 and to SD3 in https://github.com/huggingface/diffusers/pull/9818

### DIFF
--- a/src/diffusers/models/attention_processor.py
+++ b/src/diffusers/models/attention_processor.py
@@ -2813,7 +2813,6 @@ class CogVideoXAttnProcessor2_0:
         encoder_hidden_states: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
         image_rotary_emb: Optional[torch.Tensor] = None,
-        *args,
         **kwargs,
     ) -> torch.Tensor:
         text_seq_length = encoder_hidden_states.size(1)
@@ -2886,7 +2885,6 @@ class FusedCogVideoXAttnProcessor2_0:
         encoder_hidden_states: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
         image_rotary_emb: Optional[torch.Tensor] = None,
-        *args,
         **kwargs,
     ) -> torch.Tensor:
         text_seq_length = encoder_hidden_states.size(1)
--- a/src/diffusers/models/attention_processor.py
+++ b/src/diffusers/models/attention_processor.py
@@ -2813,7 +2813,6 @@ class CogVideoXAttnProcessor2_0:
         encoder_hidden_states: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
         image_rotary_emb: Optional[torch.Tensor] = None,
-        **kwargs,
     ) -> torch.Tensor:
         text_seq_length = encoder_hidden_states.size(1)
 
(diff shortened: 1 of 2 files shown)


src/diffusers/models/attention_processor.py
```diff
@@ -2813,6 +2813,8 @@ def __call__(
         encoder_hidden_states: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
         image_rotary_emb: Optional[torch.Tensor] = None,
+        *args,
```
yiyixuxu, line 2816: we don't need to accept more arguments here, no? you are going to use the custom attention processor anyway

a-r-r-o-w: Maybe should add attention_kwargs by default whenever adding a new model for use cases like this. Regional prompting, conditional attention masking, etc. seem to be significantly used by power-users of UIs, so if that's okay with you, will do newer model additions with this in mind

huanngzh: Perhaps accepting more arguments is not necessary here, because it will be detected and intercepted by the Attention forward. If agreed, I will delete these two lines.

a-r-r-o-w: I think what we mean here is to keep `**kwargs` but remove `*args`. Your changes look good to merge except for the two occurences of `*args` in the attention processors

huanngzh: Have removed it.

a-r-r-o-w (approved):
Thanks!

src/diffusers/models/attention_processor.py
```diff
@@ -2813,6 +2813,7 @@ def __call__(
         encoder_hidden_states: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
         image_rotary_emb: Optional[torch.Tensor] = None,
+        **kwargs,
```
yiyixuxu, line 2816: oh I just saw the example code here you provided 

https://github.com/huggingface/diffusers/pull/10456#issue-2769170591

your custom attention processor should not inherit from the official one, so I don't think the **kwargs here is needed, no

huanngzh: Hi, I have updated my example code of custom attention processor.

yiyixuxu (approved):
thanks!

PR #10461, opened 2025-01-06
[Training] QoL improvements in the Flux Control training scripts

sayakpaul:
# What does this PR do?

* Introduce a flag `only_target_transformer_blocks` in the `train_control_flux.py` script. This only trains the transformer blocks along with the `x_embedder` layer. [Results show](https://wandb.ai/sayakpaul/flux_train_control/runs/j13wv2jr) that it works.
* Introduce a flag `log_dataset_samples` which lets users log samples to `wandb` (properly guarded). One can check [this wandb run page](https://wandb.ai/sayakpaul/flux_train_control/reports/dataset_samples-25-01-06-07-43-11---VmlldzoxMDgyNTMyMw) to see that it works.
* Modifies the dataloader such that when a dataset caption column has a list of different captions we only take the largest one. This lets us train on datasets like this: https://huggingface.co/datasets/TIGER-Lab/OmniEdit-Filtered-1.2M. We could randomly sample, too but for now, this is okay IMO. 
* Some additional changes to the README, fixing some argument names.


examples/flux-control/train_control_flux.py
```diff
@@ -797,6 +807,12 @@ def main(args):
     assert torch.all(flux_transformer.x_embedder.weight[:, initial_input_channels:].data == 0)
     flux_transformer.register_to_config(in_channels=initial_input_channels * 2, out_channels=initial_input_channels)
 
+    if args.only_target_transformer_blocks:
```
sayakpaul, line 810: Doesn't apply to the LoRA script.

a-r-r-o-w (approved):
Thanks!

PR #10466, opened 2025-01-06
Update transformer to accelerator.device with it's weight_dtype

Viditnegi:
By default, the transformer always gets transferred to the device with float32 whether we specify the dtype as bfloat16 or float16. This fixes it.

# What does this PR do?

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/examples/dreambooth/train_dreambooth_flux.py
+++ b/examples/dreambooth/train_dreambooth_flux.py
@@ -1157,6 +1157,7 @@ def main(args):
             "Mixed precision training with bfloat16 is not supported on MPS. Please use fp16 (recommended) or fp32 instead."
         )
 
+    transformer.to(accelerator.device, dtype=weight_dtype)
     vae.to(accelerator.device, dtype=weight_dtype)
     if not args.train_text_encoder:
         text_encoder_one.to(accelerator.device, dtype=weight_dtype)


examples/dreambooth/train_dreambooth_flux.py
```diff
@@ -1157,6 +1157,7 @@ def main(args):
             "Mixed precision training with bfloat16 is not supported on MPS. Please use fp16 (recommended) or fp32 instead."
         )
 
+    transformer.to(accelerator.device, dtype=weight_dtype)
```
sayakpaul, line 1160: We should not be doing it as we're fine-tuning the `transformer`. `accelerator.prepare()` handles this for us (plus any additional hook placements that might be required).

Viditnegi: The transformer gets transferred to the gpu in float32 by default. Takes 40 gb memory at once!

sayakpaul: Well it will trained in mixed-precision through autocast. This is handled by `accelerate`. If the user wants to cast it to a lower precision, it should be done through an CLI argument, not by default IMO.

Viditnegi: Sure, but unlike this, flux control net training script explicitly transfers the model to accelerator.device in the mixed precision dtype.

linoytsaban: for context, we do have this line - `transformer.to(accelerator.device, dtype=weight_dtype)` in the LoRA training scripts, and the flag `--upcast_before_saving` (controls wether to upcast to fp32 before saving) that is False by default indeed due to the memory reqs

Viditnegi: Okay thanks!

DN6:
Can we close this since it seems we can resolve the issue with a CLI arg?

Viditnegi:
> Can we close this since it seems we can resolve the issue with a CLI arg? 

Yes, just closed the pr!

PR #10469, opened 2025-01-06
UNet2DModel mid_block_type

### LINKED ISSUE #10460: `UNet2DModel` is missing the documented `mid_block_type` argument
### Describe the bug

The [documentation](https://huggingface.co/docs/diffusers/api/models/unet2d) for the `UNet2DModel` class claims that the constructor accepts a `mid_block_type` argument.  However, this argument does not seem to actually be present in the code, and you get an error if you try to provide it.

### Reproduction

```pycon
>>> import diffusers
>>> diffusers.UNet2DModel(mid_block_type=None)
  File "<stdin>", line 1, in <module>
  File "/home/kale/.pyenv/versions/3.12.4/lib/python3.12/site-packages/diffusers/configuration_utils.py", line 665, in inner_init
    init(self, *args, **init_kwargs)
TypeError: UNet2DModel.__init__() got an unexpected keyword argument 'mid_block_type'
```

### Logs

_No response_

### System Info

- 🤗 Diffusers version: 0.32.1
- Platform: Linux-6.12.8-arch1-1-x86_64-with-glibc2.40
- Python version: 3.12.4
- PyTorch version (GPU?): 2.5.1+cu124 (True)
- Jax version: not installed
- Huggingface_hub version: 0.24.6
- Transformers version: not installed
- Accelerate version: not installed
- PEFT version: not installed
- Safetensors version: 0.4.4
- Accelerator: NVIDIA GeForce RTX 4060 Ti, 16380 MiB

### Who can help?

@sayakpaul

hlky:
# What does this PR do?

`mid_block_type` is unsupported in `UNet2DModel` despite appearing in documentation.

https://github.com/huggingface/diffusers/blob/b5726358cf125f2fa1a596dce321e91a225a57e4/src/diffusers/models/unets/unet_2d_blocks.py#L274-L322

Of these `UNet2DModel` only works with `UNetMidBlock2D`. `None` can also be supported and is added in this PR.

Fixes #10460

## Who can review?

@sayakpaul

### DIFF
--- a/src/diffusers/models/unets/unet_2d.py
+++ b/src/diffusers/models/unets/unet_2d.py
@@ -58,7 +58,7 @@ class UNet2DModel(ModelMixin, ConfigMixin):
         down_block_types (`Tuple[str]`, *optional*, defaults to `("DownBlock2D", "AttnDownBlock2D", "AttnDownBlock2D", "AttnDownBlock2D")`):
             Tuple of downsample block types.
         mid_block_type (`str`, *optional*, defaults to `"UNetMidBlock2D"`):
-            Block type for middle of UNet, it can be either `UNetMidBlock2D` or `UnCLIPUNetMidBlock2D`.
+            Block type for middle of UNet, it can be either `UNetMidBlock2D` or `None`.
         up_block_types (`Tuple[str]`, *optional*, defaults to `("AttnUpBlock2D", "AttnUpBlock2D", "AttnUpBlock2D", "UpBlock2D")`):
             Tuple of upsample block types.
         block_out_channels (`Tuple[int]`, *optional*, defaults to `(224, 448, 672, 896)`):
@@ -103,6 +103,7 @@ class UNet2DModel(ModelMixin, ConfigMixin):
         freq_shift: int = 0,
         flip_sin_to_cos: bool = True,
         down_block_types: Tuple[str, ...] = ("DownBlock2D", "AttnDownBlock2D", "AttnDownBlock2D", "AttnDownBlock2D"),
+        mid_block_type: Optional[str] = "UNetMidBlock2D",
         up_block_types: Tuple[str, ...] = ("AttnUpBlock2D", "AttnUpBlock2D", "AttnUpBlock2D", "UpBlock2D"),
         block_out_channels: Tuple[int, ...] = (224, 448, 672, 896),
         layers_per_block: int = 2,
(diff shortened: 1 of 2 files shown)


src/diffusers/models/unets/unet_2d.py
```diff
@@ -194,19 +195,22 @@ def __init__(
...
-            output_scale_factor=mid_block_scale_factor,
-            resnet_time_scale_shift=resnet_time_scale_shift,
-            attention_head_dim=attention_head_dim if attention_head_dim is not None else block_out_channels[-1],
-            resnet_groups=norm_num_groups,
-            attn_groups=attn_norm_num_groups,
-            add_attention=add_attention,
-        )
+        if mid_block_type is None:
+            self.mid_block = None
+        else:
```
sayakpaul, line 200: Hmm, if it's unsupported then should we not be disallowing its usage? I understand that's a breaking change and in that case, a deprecation would be preferred. WDYT?

hlky: If we don't want to support `mid_block_type=None` then we can just remove the docstring, the argument itself currently doesn't exist.

https://github.com/huggingface/diffusers/blob/b5726358cf125f2fa1a596dce321e91a225a57e4/src/diffusers/models/unets/unet_2d_condition.py#L1244

We support `mid_block_type` `None` in `UNet2DConditionModel` so it seemed appropriate to support in `UNet2DModel`.

We could use `get_mid_block` here but simpler like this as we'd need to guard against the other 2 `CrossAttn` mid blocks.

sayakpaul: > We support mid_block_type None in UNet2DConditionModel so it seemed appropriate to support in UNet2DModel.

I think since it's unsupported anyway in `UNet2DModel`, I am not seeing the reason to pass a `None` but if it's simpler to do so maintenance-wise, I am okay with that,

hlky: See the linked issue https://github.com/huggingface/diffusers/issues/10460
```python
>>> import diffusers
>>> diffusers.UNet2DModel(mid_block_type=None)
```

sayakpaul: Okay works for me then but I will note that supporting just `None` values does feel weird to me.

DN6 (approved):

PR #10481, opened 2025-01-07
[Sana][bug fix]change clean_caption from True to False.

lawrence-cj:
# What does this PR do?

Due to SanaPipeline's support of multiple languages, changing the default clean_caption from True to False is better for better performance.

Cc: @yiyixuxu


yiyixuxu (approved):
thanks!

PR #10493, opened 2025-01-08
[Sana 4K]

lawrence-cj:
# What does this PR do?

This PR adds support for Sana 4096x4096px image generation.
Cc: @yiyixuxu 

Generated result(compressed due to file size limitation):

8441736261830_ pic_compress


yiyixuxu (approved):
thank you!

elismasilva:
ok i think you forgot to do the same for this pipeline
https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/pag/pipeline_pag_sana.py

PR #10503, opened 2025-01-09
Module Group Offloading

a-r-r-o-w:
- `enable_model_cpu_offload` onloads the entire transformer model at once. The minimal memory requirements for this is, therefore, determined by the size of the transformer. For large models, it is sometimes impossible to even load the memory on GPU
- `enable_sequential_cpu_offload` has very minimal memory requirements, but is too slow because of lots of synchronous device transfers. We can speed this up with async cuda streams to "hide" the HtoD and DtoH transfer latency by overlapping with computation. The implementation with cuda sterams would be required to come from `accelerate` since we rely on it for memory management in this case.
- UIs usually rely on such a model management system. They require finegrained control on what layers to offload/onload to device, and the requirement may change dynamically during runtime based on chosen user settings. This PR enables that in limited capacity at the moment.

|   model_id    |  offloading_type  |  use_stream  |  num_blocks  |    time |   model_memory |   inference_memory |
|:-------------:|:-----------------:|:------------:|:------------:|--------:|---------------:|-------------------:|
| cogvideox-1.0 |       none        |    False     |              | 242.485 |         19.697 |             24.396 |
| cogvideox-1.0 |       model       |    False     |              | 328.429 |          0.059 |             15.314 |
| cogvideox-1.0* |    block_level    |    False     |            8 | 289.809 |           9.35 |             16.023 |
| cogvideox-1.0* |    block_level    |     True     |            8 | 247.966 |          9.332 |             17.996 |
| cogvideox-1.0* |    block_level    |    False     |            1 | 282.885 |          9.332 |             14.457 |
| cogvideox-1.0* |    block_level    |     True     |            1 | 244.795 |           9.35 |             14.549 |
| cogvideox-1.0 |    leaf_level     |     True     |              | 248.396 |          0.479 |              5.344 |
| hunyuan_video |       none        |    False     |              | 69.436 |         24.436 |             26.451 |
| hunyuan_video |       model       |    False     |              | 141.01 |          0.041 |             25.949 |
| hunyuan_video |    block_level    |    False     |            8 |139.316 |          0.525 |              8.273 |
| hunyuan_video |    block_level    |     True     |            8 | 83.975 |          0.701 |             13.508 |
| hunyuan_video |    block_level    |    False     |            1 |137.182 |          0.525 |              3.844 |
| hunyuan_video |    block_level    |     True     |            1 | 77.245 |          0.721 |              4.666 |
| hunyuan_video |    leaf_level     |     True     |              | 80.375 |          0.533 |              2.908 |
| ltx_video  |       none        |    False     |              | 36.702 |           4.85 |              6.475 |
| ltx_video  |       model       |    False     |              | 51.525 |          0.025 |              5.678 |
| ltx_video  |    block_level    |    False     |            8 | 56.477 |          0.848 |              3.564 |
| ltx_video  |    block_level    |     True     |            8 | 41.053 |          1.121 |              5.025 |
| ltx_video  |    block_level    |    False     |            1 | 55.057 |          0.848 |              2.658 |
| ltx_video  |    block_level    |     True     |            1 | 38.443 |          1.121 |              3.057 |
| ltx_video  |    leaf_level     |     True     |              | 38.456 |          1.121 |              2.826 |
|    flux    |       none        |    False     |              | 16.811 |         31.467 |             32.088 |
|    flux    |       model       |    False     |              |178.134 |          0.041 |              22.84 |
|    flux*    |    block_level    |    False     |            8 |116.088 |          9.307 |             14.891 |
|    flux*    |    block_level    |     True     |            8 | 50.228 |          9.309 |             20.168 |
|    flux*    |    block_level    |    False     |            1 |119.193 |          9.307 |             10.461 |
|    flux*    |    block_level    |     True     |            1 | 50.097 |          9.309 |             11.307 |
|    flux    |    leaf_level     |     True     |              | 52.519 |          0.227 |               1.09   |

<sub><sup>*</sup>The benchmarks were run with a mistake in the offloading code. This caused text encoder to be on the GPU instead of being offloaded, making the comparison unfair to those runs marked without a `*`</sub>

<details>

```python
import argparse
import gc
import pathlib
import traceback

import git
import pandas as pd
import torch
from diffusers import (
    AllegroPipeline,
    CogVideoXPipeline,
    FluxPipeline,
    HunyuanVideoPipeline,
    LattePipeline,
    LTXPipeline,
    MochiPipeline,
)
from diffusers.models import HunyuanVideoTransformer3DModel
from diffusers.utils.logging import set_verbosity_info, set_verbosity_debug
from tabulate import tabulate


repo = git.Repo(path="/home/aryan/work/diffusers")
branch = repo.active_branch


def pretty_print_results(results, precision: int = 3):
    def format_value(value):
        if isinstance(value, float):
            return f"{value:.{precision}f}"
        return value

    filtered_table = {k: format_value(v) for k, v in results.items()}
    print(tabulate([filtered_table], headers="keys", tablefmt="pipe", stralign="center"))


    torch.cuda.synchronize()

    start.record()
    output = f(*args, **kwargs)
    end.record()
    torch.cuda.synchronize()
    elapsed_time = round(start.elapsed_time(end) / 1000, 3)

    return elapsed_time, output


def prepare_allegro(dtype: torch.dtype, compile: bool = False, **kwargs):
    model_id = "rhymes-ai/Allegro"
    cache_dir = None

    pipe = AllegroPipeline.from_pretrained(model_id, torch_dtype=dtype, cache_dir=cache_dir)
    pipe.to("cuda")
    pipe.vae.enable_tiling()

    if compile:
        pipe.transformer = torch.compile(
            pipe.transformer, mode="max-autotune-no-cudagraphs", fullgraph=True, dynamic=False
        )

    for key, value in list(kwargs.items()):
        if torch.is_tensor(value):
            kwargs[key] = value.to(device="cuda", dtype=dtype)

    generation_kwargs = {
        "prompt": "A seaside harbor with bright sunlight and sparkling seawater, with many boats in the water. From an aerial view, the boats vary in size and color, some moving and some stationary. Fishing boats in the water suggest that this location might be a popular spot for docking fishing boats.",
        "height": 720,
        "width": 1280,
        "num_inference_steps": 50,
        "guidance_scale": 5.0,
        **kwargs,
    }

    return pipe, generation_kwargs


def prepare_cogvideox_1_0(dtype: torch.dtype, compile: bool = False, **kwargs):
    model_id = "THUDM/CogVideoX-5b"
    cache_dir = None

    pipe = CogVideoXPipeline.from_pretrained(model_id, torch_dtype=dtype, cache_dir=cache_dir)
    pipe.to("cuda")

    prompt_embeds, negative_prompt_embeds = pipe.encode_prompt(
        prompt=(
            "A panda, dressed in a small, red jacket and a tiny hat, sits on a wooden stool in a serene bamboo forest. "
            "The panda's fluffy paws strum a miniature acoustic guitar, producing soft, melodic tunes. Nearby, a few other "
            "pandas gather, watching curiously and some clapping in rhythm. Sunlight filters through the tall bamboo, "
            "casting a gentle glow on the scene. The panda's face is expressive, showing concentration and joy as it plays. "
            "The background includes a small, flowing stream and vibrant green foliage, enhancing the peaceful and magical "
            "atmosphere of this unique musical performance."
        ),
        device="cuda",
        dtype=dtype,
    )

    pipe.text_encoder.to("cpu")
    del pipe.text_encoder

    for key, value in list(kwargs.items()):
        if torch.is_tensor(value):
            kwargs[key] = value.to(device="cuda", dtype=dtype)

    generation_kwargs = {
        "prompt_embeds": prompt_embeds,
        "negative_prompt_embeds": negative_prompt_embeds,
        "height": 480,
        "width": 720,
        "num_frames": 49,
        "num_inference_steps": 50,
        "guidance_scale": 5.0,
        **kwargs,
    }

    return pipe, generation_kwargs


def prepare_flux(dtype: torch.dtype, compile: bool = False, **kwargs) -> None:
    model_id = "black-forest-labs/Flux.1-Dev"
    cache_dir = "/raid/.cache/huggingface"

    pipe = FluxPipeline.from_pretrained(model_id, torch_dtype=dtype, cache_dir=cache_dir)
    pipe.to("cuda")

    prompt_embeds, pooled_prompt_embeds, _ = pipe.encode_prompt(
        prompt="A cat holding a sign that says hello world", prompt_2=None, device="cuda"
    )

    pipe.text_encoder.to("cpu")
    pipe.text_encoder_2.to("cpu")

    for key, value in list(kwargs.items()):
        if torch.is_tensor(value):
            kwargs[key] = value.to(device="cuda", dtype=dtype)

    generation_kwargs = {
        "prompt_embeds": prompt_embeds,
        "pooled_prompt_embeds": pooled_prompt_embeds,
        "height": 768,
        "width": 768,
        "num_inference_steps": 50,
        "guidance_scale": 5.0,
        **kwargs,
    }

    return pipe, generation_kwargs


def prepare_hunyuan_video(dtype: torch.dtype, compile: bool = False, **kwargs):
    cache_dir = None

    )
    pipe = HunyuanVideoPipeline.from_pretrained(
        model_id, transformer=transformer, torch_dtype=torch.float16, cache_dir=cache_dir
    )
    pipe.to("cuda")

    prompt_embeds, pooled_prompt_embeds, prompt_attention_mask = pipe.encode_prompt(
        prompt="A cat wearing sunglasses and working as a lifeguard at pool.", device="cuda", dtype=torch.float16
    )
    pipe.text_encoder.to("cpu")
    pipe.text_encoder_2.to("cpu")
    del pipe.text_encoder, pipe.text_encoder_2

    for key, value in list(kwargs.items()):
        if torch.is_tensor(value):
            kwargs[key] = value.to(device="cuda", dtype=dtype)

    generation_kwargs = {
        "prompt_embeds": prompt_embeds,
        "pooled_prompt_embeds": pooled_prompt_embeds,
        "prompt_attention_mask": prompt_attention_mask,
        "height": 320,
        "width": 512,
        "num_frames": 61,
        "num_inference_steps": 30,
    }

    return pipe, generation_kwargs


def prepare_latte(dtype: torch.dtype, compile: bool = False, **kwargs):
    model_id = "maxin-cn/Latte-1"
    cache_dir = None

    pipe = LattePipeline.from_pretrained(model_id, torch_dtype=dtype, cache_dir=cache_dir)
    pipe.to("cuda")

    prompt_embeds, negative_prompt_embeds = pipe.encode_prompt(
        prompt="A cat wearing sunglasses and working as a lifeguard at pool.",
        num_videos_per_prompt=1,
        device="cuda",
    )
    pipe.text_encoder.to("cpu")
    del pipe.text_encoder

    for key, value in list(kwargs.items()):
        if torch.is_tensor(value):
            kwargs[key] = value.to(device="cuda", dtype=dtype)

    generation_kwargs = {
        "prompt_embeds": prompt_embeds,
        "negative_prompt_embeds": negative_prompt_embeds,
        "height": 512,
        "width": 512,
        "video_length": 16,
        "num_inference_steps": 50,
    }

    return pipe, generation_kwargs


def prepare_ltx_video(dtype: torch.dtype, compile: bool = False, **kwargs):
    model_id = "a-r-r-o-w/LTX-Video-diffusers"
    cache_dir = None

    pipe = LTXPipeline.from_pretrained(model_id, torch_dtype=dtype, cache_dir=cache_dir)
    pipe.to("cuda")

    (
        prompt_embeds,
        prompt_attention_mask,
        negative_prompt_embeds,
        negative_prompt_attention_mask,
    ) = pipe.encode_prompt(
        prompt="A woman with long brown hair and light skin smiles at another woman with long blonde hair. The woman with brown hair wears a black jacket and has a small, barely noticeable mole on her right cheek. The camera angle is a close-up, focused on the woman with brown hair's face. The lighting is warm and natural, likely from the setting sun, casting a soft glow on the scene. The scene appears to be real-life footage",
        negative_prompt="worst quality, inconsistent motion, blurry, jittery, distorted",
        num_videos_per_prompt=1,
        device="cuda",
    )
    pipe.text_encoder.to("cpu")
    del pipe.text_encoder

    for key, value in list(kwargs.items()):
        if torch.is_tensor(value):
            kwargs[key] = value.to(device="cuda", dtype=dtype)
    
    generation_kwargs = {
        "prompt_embeds": prompt_embeds,
        "prompt_attention_mask": prompt_attention_mask,
        "negative_prompt_embeds": negative_prompt_embeds,
        "negative_prompt_attention_mask": negative_prompt_attention_mask,
        "width": 768,
        "height": 512,
        "num_frames": 161,
        "num_inference_steps": 50,
    }

    return pipe, generation_kwargs


def prepare_mochi(dtype: torch.dtype, compile: bool = False, **kwargs):
    model_id = "genmo/mochi-1-preview"
    cache_dir = None

    pipe = MochiPipeline.from_pretrained(model_id, torch_dtype=dtype, cache_dir=cache_dir)
    pipe.to("cuda")
    pipe.vae.enable_tiling()

    for key, value in list(kwargs.items()):
        if torch.is_tensor(value):
            kwargs[key] = value.to(device="cuda", dtype=dtype)

    generation_kwargs = {
        "prompt": "Close-up of a chameleon's eye, with its scaly skin changing color. Ultra high resolution 4k.",
        "height": 480,
        "width": 848,
        "num_frames": 85,
        "num_inference_steps": 50,
    }

    return pipe, generation_kwargs


def decode_allegro(pipe: AllegroPipeline, latents: torch.Tensor, filename: pathlib.Path, **kwargs):
    filename = f"{filename.as_posix()}.mp4"
    video = pipe.decode_latents(latents)
    video = pipe.video_processor.postprocess_video(video=video, output_type="pil")[0]
    export_to_video(video, filename, fps=8)
    return filename


def decode_cogvideox_1_0(pipe: CogVideoXPipeline, latents: torch.Tensor, filename: pathlib.Path, **kwargs):
    filename = f"{filename.as_posix()}.mp4"
    video = pipe.decode_latents(latents)
    video = pipe.video_processor.postprocess_video(video=video, output_type="pil")[0]
    export_to_video(video, filename, fps=8)
    return filename


def decode_flux(pipe: FluxPipeline, latents: torch.Tensor, filename: pathlib.Path, **kwargs):
    height = kwargs["height"]
    width = kwargs["width"]
    filename = f"{filename.as_posix()}.png"
    latents = pipe._unpack_latents(latents, height, width, pipe.vae_scale_factor)
    latents = (latents / pipe.vae.config.scaling_factor) + pipe.vae.config.shift_factor
    image = pipe.vae.decode(latents, return_dict=False)[0]
    image = pipe.image_processor.postprocess(image, output_type="pil")[0]
    image.save(filename)
    return filename


def decode_hunyuan_video(pipe: HunyuanVideoPipeline, latents: torch.Tensor, filename: pathlib.Path, **kwargs):
    filename = f"{filename.as_posix()}.mp4"
    latents = latents.to(pipe.vae.dtype) / pipe.vae.config.scaling_factor
    video = pipe.vae.decode(latents, return_dict=False)[0]
    video = pipe.video_processor.postprocess_video(video, output_type="pil")[0]
    export_to_video(video, filename, fps=8)
    return filename


def decode_latte(pipe: LattePipeline, latents: torch.Tensor, filename: pathlib.Path, **kwargs):
    filename = f"{filename.as_posix()}.mp4"
    video = pipe.decode_latents(latents, video_length=kwargs["video_length"])
    video = pipe.video_processor.postprocess_video(video=video, output_type="pil")[0]
    export_to_video(video, filename, fps=8)
    return filename


def decode_ltx_video(pipe: LTXPipeline, latents: torch.Tensor, filename: pathlib.Path, **kwargs):
    filename = f"{filename.as_posix()}.mp4"
    latent_num_frames = (kwargs["num_frames"] - 1) // pipe.vae_temporal_compression_ratio + 1
    latent_height = kwargs["height"] // pipe.vae_spatial_compression_ratio
    latent_width = kwargs["width"] // pipe.vae_spatial_compression_ratio

    latents = pipe._unpack_latents(
        latents,
        latent_num_frames,
        latent_height,
        latent_width,
        pipe.transformer_spatial_patch_size,
        pipe.transformer_temporal_patch_size,
    )
    latents = pipe._denormalize_latents(
        latents, pipe.vae.latents_mean, pipe.vae.latents_std, pipe.vae.config.scaling_factor
    )
    latents = latents.to(pipe.vae.dtype)

    timestep = None
    video = pipe.vae.decode(latents, timestep, return_dict=False)[0]
    video = pipe.video_processor.postprocess_video(video, output_type="pil")[0]
    export_to_video(video, filename, fps=24)
    return filename


def decode_mochi(pipe: MochiPipeline, latents: torch.Tensor, filename: pathlib.Path, **kwargs):
    filename = f"{filename.as_posix()}.mp4"
    latents_mean = torch.tensor(pipe.vae.config.latents_mean).view(1, 12, 1, 1, 1).to(latents.device, latents.dtype)
    latents_std = torch.tensor(pipe.vae.config.latents_std).view(1, 12, 1, 1, 1).to(latents.device, latents.dtype)
    latents = latents * latents_std / pipe.vae.config.scaling_factor + latents_mean
    video = pipe.vae.decode(latents, return_dict=False)[0]
    video = pipe.video_processor.postprocess_video(video=video, output_type="pil")[0]
    export_to_video(video, filename, fps=8)
    return filename


def reset_memory():
    gc.collect()
    torch.cuda.empty_cache()
    torch.cuda.ipc_collect()
    torch.cuda.synchronize()
    torch.cuda.reset_accumulated_memory_stats()


MODEL_MAPPING = {
    "allegro": {
        "prepare": prepare_allegro,
        "decode": decode_allegro,
    },
    "cogvideox-1.0": {
        "prepare": prepare_cogvideox_1_0,
        "decode": decode_cogvideox_1_0,
    },
    "flux": {
        "prepare": prepare_flux,
        "decode": decode_flux,
    },
    "hunyuan_video": {
        "prepare": prepare_hunyuan_video,
        "decode": decode_hunyuan_video,
    },
    "latte": {
        "prepare": prepare_latte,
        "decode": decode_latte,
    },
    "ltx_video": {
        "prepare": prepare_ltx_video,
        "decode": decode_ltx_video,
    },
    "mochi": {
        "prepare": prepare_mochi,
        "decode": decode_mochi,
    },
}

STR_TO_COMPUTE_DTYPE = {
    "bf16": torch.bfloat16,
    "fp16": torch.float16,
    "fp32": torch.float32,
}


def run_inference(pipe, generation_kwargs):
    generator = torch.Generator().manual_seed(181201)
    output = pipe(generator=generator, output_type="latent", **generation_kwargs)[0]
    torch.cuda.synchronize()
    return output


@torch.no_grad()
def main(model_id: str, output_dir: str, dtype: str, offloading_type: str, num_blocks_per_group: int, use_stream: bool, compile: bool):
    if model_id not in MODEL_MAPPING.keys():
        raise ValueError("Unsupported `model_id` specified.")

    output_dir = pathlib.Path(output_dir)
    output_dir.mkdir(parents=True, exist_ok=True)
    csv_filename = output_dir / f"{model_id}.csv"

    compute_dtype = STR_TO_COMPUTE_DTYPE[dtype]
    model = MODEL_MAPPING[model_id]
    reset_memory()

    try:
        # 1. Prepare inputs and generation kwargs
        pipe, generation_kwargs = model["prepare"](dtype=compute_dtype)

        # 2. Apply group offloading
        if offloading_type == "model":
        elif offloading_type == "sequential":
        elif offloading_type in ["block_level", "leaf_level"]:
            apply_group_offloading(
                pipe.transformer,
                offload_type=offloading_type,
                num_blocks_per_group=num_blocks_per_group,
                offload_device=torch.device("cpu"),
                onload_device=torch.device("cuda"),
                # force_offload=True for a more fair comparison against model offloading
                # If we set to True -> lower memory
                # If we set to False -> lower time required
                force_offload=True,
                non_blocking=True,
                use_stream=use_stream,
            )

        reset_memory()
        model_max_memory_reserved = round(torch.cuda.max_memory_reserved() / 1024**3, 3)

        if compile:
            pipe.transformer = torch.compile(
                pipe.transformer, mode="max-autotune-no-cudagraphs", fullgraph=True, dynamic=False
            )

        # 3. Warmup
        num_warmups = 1
        original_num_inference_steps = generation_kwargs["num_inference_steps"]
        generation_kwargs["num_inference_steps"] = 2
        for _ in range(num_warmups):
            run_inference(pipe, generation_kwargs)
        generation_kwargs["num_inference_steps"] = original_num_inference_steps

        # 4. Benchmark
        time, latents = benchmark_fn(run_inference, pipe, generation_kwargs)
        inference_max_memory_reserved = round(torch.cuda.max_memory_reserved() / 1024**3, 3)

        # 5. Decode latents
        filename = output_dir / f"{model_id}---dtype-{dtype}---offloading_type-{offloading_type}---num_blocks_per_group-{num_blocks_per_group}---use_stream-{use_stream}---compile-{compile}"
        filename = model["decode"](
            pipe,
            latents,
            filename,
            height=generation_kwargs["height"],
            width=generation_kwargs["width"],
            num_frames=generation_kwargs.get("num_frames", None),
            video_length=generation_kwargs.get("video_length", None),
        )

        # 6. Save artifacts
        info = {
            "model_id": model_id,
            "offloading_type": offloading_type,
            "use_stream": use_stream,
            "num_blocks": num_blocks_per_group,
            "time": time,
            "model_memory": model_max_memory_reserved,
            "inference_memory": inference_max_memory_reserved,
            "compile": compile,
            "compute_dtype": dtype,
            "branch": branch,
            "filename": filename,
            "exception": None,
        }

    except Exception as e:
        print(f"An error occurred: {e}")
        traceback.print_exc()

        # 6. Save artifacts
        info = {
            "model_id": model_id,
            "offloading_type": offloading_type,
            "use_stream": use_stream,
            "num_blocks": num_blocks_per_group,
            "time": None,
            "model_memory": None,
            "inference_memory": None,
            "compile": compile,
            "compute_dtype": dtype,
            "branch": branch,
            "filename": None,
            "exception": str(e),
        }

    pretty_print_results(info, precision=3)

    df = pd.DataFrame([info])
    df.to_csv(csv_filename.as_posix(), mode="a", index=False, header=not csv_filename.is_file())


if __name__ == "__main__":
    parser.add_argument(
        "--model_id",
        type=str,
        default="flux",
        choices=["flux", "cogvideox-1.0", "latte", "allegro", "hunyuan_video", "mochi", "ltx_video"],
        help="Model to run benchmark for.",
    )
    parser.add_argument(
        "--output_dir", required=True, type=str, help="Path where the benchmark artifacts and outputs are the be saved."
    )
    parser.add_argument("--dtype", type=str, help="torch.dtype to use for inference")
    parser.add_argument("--offloading_type", type=str, default="none", choices=["none", "model", "block_level", "leaf_level"], help="Type of offloading to use.")
    parser.add_argument("--num_blocks_per_group", type=int, default=None, help="Number of layers per group for group offloading.")
    parser.add_argument("--use_stream", action="store_true", default=False, help="Whether to use CUDA streams for offloading.")
    parser.add_argument(
        "--compile",
        action="store_true",
        default=False,
        help="Whether to torch.compile the denoiser.",
    )
    parser.add_argument("-v", "--verbose", action="store_true", help="Enable verbose logging.")
    args = parser.parse_args()

    if args.verbose:
        set_verbosity_debug()
    else:
        set_verbosity_info()

    main(
        args.model_id,
        args.output_dir,
        args.dtype,
        args.offloading_type,
        args.num_blocks_per_group,
        args.use_stream,
        args.compile,
    )
```
</details>

Some goals of this PR:
- Opt-in choice to completely eliminate/hide any device transfer latency where possible by overlapping computation/transfer. This usually comes at a slightly higher memory requirement than choosing to not hide latency
- ~Fully compatible with torch.compile~ There are a few recompiles triggered. Not really sure how to get away with it

In a way, these changes can enable both enable_model_cpu_offload and enable_sequential_cpu_offload because of the way `offload_group_patterns` can be leveraged, but that is not the goal.


yiyixuxu:
I think this fits well in the offloading I'm working on in modular diffusers

yiyixuxu:
Maybe we should consolidate a bit - I will separate the offloading part into its own PR

src/diffusers/hooks/group_offloading.py
```diff
@@ -0,0 +1,218 @@
...
+
+import re
+from typing import List, Optional, Union
+
+import torch
+
+from .hooks import HookRegistry, ModelHook
+
+
+_COMMON_STACK_IDENTIFIERS = {
```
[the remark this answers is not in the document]
a-r-r-o-w: Actually will remove this completely. This should be applicable on any model containing ModuleList or Sequential because we know for sure, atleast in Diffusers, that the call order of these layers are sequential and not in some weird access pattern.

So, will make the check to just look for the above two classes with isinstance

a-r-r-o-w:
Some more numbers after latest changes:

```
| model_id   | offloading_type | num_blocks | non_blocking | time   | model_memory | inference_memory | cuda_stream |
|------------|-----------------|------------|--------------|--------|--------------|------------------|-------------|
| ltx_video  | none            |            | False        | 36.852 | 4.85         | 6.475            | False       |
| ltx_video  | group           | 8          | True         | 53.5   | 1.205        | 3.531            | False       |
| ltx_video  | group           | 8          | True         | 36.715 | 0.848        | 4.787            | True        |
| ltx_video  | group           | 1          | True         | 52.181 | 1.205        | 2.811            | False       |
| ltx_video  | group           | 1          | True         | 36.611 | 0.848        | 2.818            | True        |
```

Continuing from our internal thread, we have positive signal that sequential CPU offloading can be done without any hit to time required for inference when using cuda streams for transfer.

a-r-r-o-w:
After some discussing and re-iterating, the two main offloading strategies in this PR are now:
- Taking a module with a stack of ModuleList/Sequential layers and being able to offload/onload "groups" of them when required
- Taking a module and being able to offload/onload the leaf-level modules when required (similar to our current `enable_sequential_cpu_offload`)

The latter has minimal memory requirements, but can be very slow. If layer prefetching is utilized, there is some time overhead but not much if there is sufficient computation to overlap with (video models are great use case, or image models with bigger batch size).

The former is more beneficial for non-CUDA devices since it allows offloading at the inner module levels. This helps reduce the contribution to memory usage by requiring entire model on GPU (normal CPU offloading aka `enable_model_cpu_offload`). Initially, I wanted to get to a stage where any group of layers could be grouped together, but there's no real benefit to it, so a simple implementation is now used (only module lists and sequentials).

From this, it might be apparent that the main target users are people with CUDA devices supporting streams - low memory usage without much overhead to generation time.

Also, if you have tested the PR before, you might find the latest version slightly faster and using a few hundred megabytes lesser :)

Some code examples:

<details>
<summary> Offloading LTX Transformer and Text encoder </summary>

```python
import torch

set_verbosity_debug()

pipe.vae.to("cuda")
pipe.vae.enable_tiling()

apply_group_offloading(
    pipe.text_encoder,
    offload_type="leaf_level",
    offload_device=torch.device("cpu"),
    onload_device=torch.device("cuda"),
    use_stream=True,
)
apply_group_offloading(
    pipe.transformer,
    offload_type="leaf_level",
    offload_device=torch.device("cpu"),
    onload_device=torch.device("cuda"),
    use_stream=True,
)

video = pipe(
    prompt=prompt,
    width=704,
    height=480,
    num_frames=161,
    num_inference_steps=50,
).frames[0]
print(f"Max memory reserved: {torch.cuda.max_memory_reserved() / 1024 ** 3:.2f} GB")
export_to_video(video, "output.mp4", fps=24)
```
</details>

<details>
<summary> HunyuanVideo LoRA inference </summary>

```python
import argparse
import os
from pathlib import Path

import torch
from diffusers.hooks.group_offloading import apply_group_offloading

set_verbosity_debug()


def get_args():
    parser.add_argument("--lora_path", type=str, default="none")
    parser.add_argument("--id_token", type=str, default="")
    parser.add_argument("--prompts_path", type=str, required=True)
    parser.add_argument("--output_dir", type=str, required=True)
    parser.add_argument("--lora_strength", type=float, default=1.0)
    parser.add_argument("--height", type=int, default=320)
    parser.add_argument("--width", type=int, default=512)
    parser.add_argument("--num_frames", type=int, default=61)
    parser.add_argument("--num_inference_steps", type=int, default=30)
    return parser.parse_args()


def string_to_filename(x):
    return x.replace(" ", "_").replace(",", "").replace(".", "").replace(":", "").replace(";", "").replace("!", "").replace("?", "").replace("'", "").replace('"', "")


args = get_args()
output_dir = Path(args.output_dir)

output_dir.mkdir(parents=True, exist_ok=True)

)
if args.lora_path != "none":
    pipe.load_lora_weights(args.lora_path, adapter_name="hunyuan-lora")
    pipe.set_adapters("hunyuan-lora", args.lora_strength)
pipe.vae.enable_tiling()

pipe.text_encoder.to("cuda")
pipe.text_encoder_2.to("cuda")
pipe.vae.to("cuda")
apply_group_offloading(
    pipe.transformer,
    offload_type="leaf_level",
    offload_device=torch.device("cpu"),
    onload_device=torch.device("cuda"),
    force_offload=True,
    non_blocking=True,
    use_stream=True,
)

with open(args.prompts_path) as file:
    prompts = [line.strip() for line in file if len(line.strip()) > 0]

for prompt in prompts:
    if args.id_token:
        prompt = f"{args.id_token} {prompt}"
    print(prompt)
    output = pipe(
        prompt=prompt,
        height=args.height,
        width=args.width,
        num_frames=args.num_frames,
        num_inference_steps=args.num_inference_steps,
    ).frames[0]
    filename = string_to_filename(prompt)[:25]
    filename = f"{filename}---lora_strength-{args.lora_strength}---height-{args.height}---width-{args.width}---num_frames-{args.num_frames}---num_inference_steps-{args.num_inference_steps}"
    filepath = output_dir / f"{filename}.mp4"
    export_to_video(output, filepath.as_posix(), fps=15)
```
</details>

@DN6 @yiyixuxu Could you give this a review?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1020,6 +1020,26 @@ def _execution_device(self):
...
+                hook = registry.get_hook("group_offloading")
+                if hook is not None:
+                    diffusers_hook_device = hook.group.onload_device
+                    break
+
+            if diffusers_hook_device is not None:
+                break
+
+        if diffusers_hook_device is not None:
+            return diffusers_hook_device
```
a-r-r-o-w, line 1041:
```suggestion
                    return hook.group.onload_device
```

src/diffusers/hooks/hooks.py
```diff
@@ -177,6 +177,7 @@ def get_hook(self, name: str) -> Optional[ModelHook]:
         return self.hooks.get(name, None)
 
     def remove_hook(self, name: str, recurse: bool = True) -> None:
+        num_hooks = len(self._hook_order)
```
a-r-r-o-w, line 180:
```suggestion
```

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1020,6 +1020,26 @@ def _execution_device(self):
         [`~DiffusionPipeline.enable_sequential_cpu_offload`] the execution device can only be inferred from
         Accelerate's module hooks.
         """
+        diffusers_hook_device = None
```
a-r-r-o-w, line 1023:
```suggestion
```

tests/models/test_modeling_common.py
```diff
@@ -1433,6 +1434,52 @@ def get_memory_usage(storage_dtype, compute_dtype):
             or abs(fp8_e4m3_fp32_max_memory - fp32_max_memory) < MB_TOLERANCE
         )
 
+    @require_torch_gpu
+    def test_group_offloading(self):
+        init_dict, inputs_dict = self.prepare_init_args_and_inputs_for_common()
+        torch.manual_seed(0)
+
+        def run_forward(model):
```
DN6, line 1442: We should add a memory check here as well against a forward pass that's fully on GPU.

tests/models/autoencoders/test_models_autoencoder_oobleck.py
```diff
@@ -132,6 +132,15 @@ def test_layerwise_casting_inference(self):
     def test_layerwise_casting_memory(self):
         pass
 
+    @unittest.skip(
```
DN6, line 135: For these exceptions, I would add a `_supports_group_offloading=True` to ModelMixin and set to False on unsupported models.

tests/models/transformers/test_models_transformer_hunyuan_dit.py
```diff
@@ -111,3 +111,11 @@ def test_set_xformers_attn_processor_for_determinism(self):
     @unittest.skip("HunyuanDIT use a custom processor HunyuanAttnProcessor2_0")
     def test_set_attn_processor_for_determinism(self):
         pass
+
+    @unittest.skip(
```
DN6, line 115: This means HunyuanDiT can't use group offload?

src/diffusers/hooks/group_offloading.py
```diff
@@ -0,0 +1,604 @@
...
+        onload_self=True,
+    )
+
+    # When using streams, we need to know the layer execution order for applying prefetching (to overlap data transfer
+    # and computation). Since we don't know the order beforehand, we apply a lazy prefetching hook that will find the
+    # execution order and apply prefetching in the correct order.
+    if stream is None:
+        _apply_group_offloading_hook(module, unmatched_group, force_offload, None)
+    else:
+        _apply_lazy_group_offloading_hook(module, unmatched_group, force_offload, None)
```
DN6, line 521: Little confused here. Don't we only need to run this dummy when using streams? What happens when it runs without streams? What am I missing?

DN6 (commented):
Excellent work. Very clean. Just left some minor comments/questions. And we should allow enabling via a model method like layerwise.

src/diffusers/pipelines/pipeline_utils.py
a-r-r-o-w: Yes, not really needed since we immediate return if a match is found. Will refactor soon

src/diffusers/hooks/group_offloading.py
```diff
@@ -0,0 +1,604 @@
...
+    for name, submodule in module.named_children():
+        if not isinstance(submodule, (torch.nn.ModuleList, torch.nn.Sequential)):
+            unmatched_modules.append((name, submodule))
+            modules_with_group_offloading.add(name)
+            continue
+
+        for i in range(0, len(submodule), num_blocks_per_group):
+            current_modules = submodule[i : i + num_blocks_per_group]
+            group = ModuleGroup(
+                modules=submodule[i : i + num_blocks_per_group],
```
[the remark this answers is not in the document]
a-r-r-o-w: Ah my bad, overlooked it

tests/models/transformers/test_models_transformer_hunyuan_dit.py
a-r-r-o-w: Yes, HunyuanDiT cannot be used with group offload at the moment. This is due to some code patterns I did not really account for and haven't thought about. See the implementation for example:

https://github.com/huggingface/diffusers/blob/3e35f56b00d73bc3c2d3bb69615176d0909fab8a/src/diffusers/models/embeddings.py#L1684-L1692

As the module weights are used directly, the forward pass is never invoked leading to no onloading. This causes device mismatch error

src/diffusers/hooks/group_offloading.py
a-r-r-o-w: When not using streams:
- we load a group of modules, perform computation, unload them, sequentially
- every group has `onload_self` set to True, meaning the leader module onloads all modules in the group when its forward pass is invoked
- the `unmatched_group` does not do anything in this case, except onload the top-level parameters and buffers

When not using streams:
- for the first forward pass, all groups have `onload_self=True`. This pass is used to determine the layer execution order and is a bit slower than all future forward passes
  (btw we can use FX tracing but it fails for many models I tried to test. I think we would need similar implementation to `transformers`, to generate mock tensors, to be able to trace layer execution order fast and without problems)
- once we know the layer execution order, we can apply prefetching with streams since each layer knows what layer is going to be executed next
- in the post_forward of LazyPrefetchGroupOffloadingHook, we can see that the `onload_self` attribute of all but the first module is set to `False`. This is because the groups no longer have to onload themselves (as they're going to be fetched by the module that is invoked before them)
- However, as `onload_self` is set to False for all groups, there needs to be atleast one group that triggers the onloading of the first group. This is done by the `unmatched_group` (top-level module), which does not really do anything but serve as a dummy, as we know for sure its forward will always be invoked first.

a-r-r-o-w:
@DN6 I've tried addressing all review comments. LMK what you think about the changes added. 

There's now a `ModelMixin::enable_group_offloading()` method. I don't think we should add anything at pipeline-level, because, as you mentioned, it will be tricky handling the different parameters.

I've added GPU memory tests with a dummy model. It is separate from the per-model tests because there are no memory savings for certain models and the test would fail -- the reason being that they're typically a single block and not bound by intermediate activation tensor size.

@stevhliu Could you give this a look too for the docs please? I'm not quite sure if this is the best place to mention, but seemed ideal. LMK if you think it should be added elsewhere

docs/source/en/optimization/memory.md
```diff
@@ -158,6 +158,46 @@ In order to properly offload models after they're called, it is required to run
 
 </Tip>
 
+## Group offloading
+
+Group offloading is a middle ground between the two above methods. It works by offloading groups of internal layers (either `torch.nn.ModuleList` or `torch.nn.Sequential`). This method uses lower memory than model-level offloading. It is also faster than sequential-level offloading, as the number of device synchronizations is reduced.
```
stevhliu, line 163:
```suggestion
Group offloading is the middle ground between CPU and model offloading. It works by offloading groups of internal layers (either `torch.nn.ModuleList` or `torch.nn.Sequential`), which uses less memory than model-level offloading. It is also faster than sequential-level offloading because the number of device synchronizations is reduced.
```

docs/source/en/optimization/memory.md
```diff
@@ -158,6 +158,46 @@ In order to properly offload models after they're called, it is required to run
...
+
+Another supported feature (for CUDA devices with support for asynchronous data transfer streams) is the ability to overlap data transfer and computation to reduce the overall execution time compared to sequential offloading. This is enabled using layer prefetching with CUDA streams, i.e., the layer that is to be executed next starts onloading to the accelerator device while the current layer is being executed - this increases the memory requirements slightly. Note that this implementation also supports leaf-level offloading but can be made much faster when using streams.
```
stevhliu, line 165: Maybe also add a link to what leaf-level offloading is or briefly explain it.

```suggestion
Group offloading (for CUDA devices with support for asynchronous data transfer streams) overlaps data transfer and computation to reduce the overall execution time compared to sequential offloading. This is enabled using layer prefetching with CUDA streams. The next layer to be executed is loaded onto the accelerator device while the current layer is being executed - this increases the memory requirements slightly. Group offloading also supports leaf-level offloading but can be made much faster when using streams.
```

docs/source/en/optimization/memory.md
```diff
@@ -158,6 +158,46 @@ In order to properly offload models after they're called, it is required to run
...
+
+To enable group offloading, either call the [`~ModelMixin.enable_group_offloading`] method on the model or pass use [`~hooks.group_offloading.apply_group_offloading`]:
```
stevhliu, line 167: Very nice! I would move this code snippet up after the first paragraph so users can get started quicker with the code :)

```suggestion
To enable group offloading, call the [`~ModelMixin.enable_group_offloading`] method on the model if it is a Diffusers model implementation. For any other model implementation, use [`~hooks.group_offloading.apply_group_offloading`]:
```

stevhliu (approved):
Thanks, perfect place for these docs! 🔥

src/diffusers/hooks/group_offloading.py
a-r-r-o-w: Ah right, yes should work. Updated

DN6 (approved):
Very nicely done @a-r-r-o-w 🚀

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1910,6 +1914,24 @@ def from_pipe(cls, pipeline, **kwargs):
...
+        from ..hooks import HookRegistry
+        from ..hooks.group_offloading import _GROUP_OFFLOADING
+
+        for name, component in self.components.items():
+            if not isinstance(component, torch.nn.Module):
+                continue
+            for module in component.modules():
+                if not hasattr(module, "_diffusers_hook"):
+                    continue
+                registry: HookRegistry = module._diffusers_hook
```
DN6, line 1927: Don't need the type here I think.

a-r-r-o-w: Has been rewritten a bit, so this no longer exists

a-r-r-o-w:
@DN6 Based on our chat about handling `.to()`, I've made the relevant changes to DiffusionPipeline and ModelMixin. I think we should do a follow-up refactoring `.to()` a bit for the pipeline because it is getting harder to understand with all the code branches for model/sequential/group offloading + quantization handling separately

DN6 (approved):

a-r-r-o-w:
@zhangvia If you use group offloading with `use_stream=False`, it will be slower than blockswap IFF blockswap is keeping some layers on GPU while offloading others, and if blockswap is only onloading 1-2 layers at a time, then it will be same speed because there is a cuda stream synchronization after each onload.

If you use group offloading with `use_stream=True`, there will almost never be any noticeable difference as the weight transfer between devices is overlapped with computation (one can thank the amazing `@gau-nerst` for the implementation). I've benchmarked this to be significantly faster than blockswap implementations (and you can see the numbers in the description comparing non-offload time vs group-offload time)

PR #10505, opened 2025-01-09
[chore] remove prints from tests.

sayakpaul:


hlky (approved):
Thanks. Can we add a message to the asserts like `f"output is different from expected, {image_slice.flatten()}"` in a follow-up?

sayakpaul:
Sure.

Issue #10540, opened 2025-01-12
Loading Flux Dev transformers from_single_file fails since support for Flux Fill was added


dxqbYD:
### Describe the bug

Flux models are sometimes distributed as transformer only, because all the other pipeline components are usually not changed.

Examples include:
 - finetunes on CivitAI, to offer a smaller download size

 - https://huggingface.co/lllyasviel/flux1-dev-bnb-nf4/tree/main

Since support for Flux Fill was added here https://github.com/huggingface/diffusers/commit/862a7d5038c1c53641ffcab146a7eeb5ab683656, `diffusers.FluxTransformer2DModel.from_single_file` fails to load these files:

```
File "/.../src/diffusers/src/diffusers/loaders/single_file_utils.py", line 598, in infer_diffusers_model_type
    if checkpoint["img_in.weight"].shape[1] == 384:
       ~~~~~~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'img_in.weight'
```

### Reproduction

Download one of the example mentioned above, and run

```
from diffusers import FluxTransformer2DModel

transformer = FluxTransformer2DModel.from_single_file(your_file)
```

### System Info

HEAD

### Who can help?

@DN6

dxqbYD:
the bnb-nf4 checkpoint linked above is apparently not only the transformer - but it has the same problem anyway. Other affected transformer-only files can be found on CivitAI, or generated using this Comfy workflow: https://civitai.com/models/982277/flux-lora-merge

DN6:
@dxqbYD I'm not sure the bnb checkpoint will load. We don't support bnb single checkpoints at the moment. But for the others, can you try loading with this branch?
https://github.com/huggingface/diffusers/pull/10545

DN6:
@dxqbYD Do you have a link to a specific checkpoint that is failing to load that isn't an bnb NF4 checkpoint?

dxqbYD:
> [@dxqbYD](https://github.com/dxqbYD) I'm not sure the bnb checkpoint will load. We don't support bnb single checkpoints at the moment. But for the others, can you try loading with this branch? [#10545](https://github.com/huggingface/diffusers/pull/10545)

Can confirm that this PR works on all my test cases, which are:
 - a civitai checkpoint that offers downloads for 20 GB (full pipeline) and another file with 11 GB (transformer only I suppose). The 11 GB file can be loaded now
 - a file that I've created myself using https://civitai.com/models/982277/flux-lora-merge
[...]

PR #10541, opened 2025-01-12
Fix loading Flux Dev transformers from_single_file

dxqbYD:
Description and reproducer in:
https://github.com/huggingface/diffusers/issues/10540

Please check the added condition before you merge.
This code works, but I am not sure what the key in question does. It seems to be absent in transformer-only Flux.Dev files. But it could also be absent in transformer-only Flux Fill files? I don't know. In that case, a different condition to tell the two apart is needed.

Fixes https://github.com/huggingface/diffusers/issues/10540 (issue)

Who can review?

@DN6

### DIFF
--- a/src/diffusers/loaders/single_file_utils.py
+++ b/src/diffusers/loaders/single_file_utils.py
@@ -605,7 +605,10 @@ def infer_diffusers_model_type(checkpoint):
         if any(
             g in checkpoint for g in ["guidance_in.in_layer.bias", "model.diffusion_model.guidance_in.in_layer.bias"]
         ):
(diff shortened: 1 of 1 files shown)


bghira:
@sayakpaul for point release consideration

DN6:
@dxqbYD Flux Transformers all contain an `img_in.weight` key. Issue https://github.com/huggingface/diffusers/issues/10540 is related to the `model.diffusion_model` prefix that is added to Comfy checkpoints. The change here would cause issues with single file checkpoints that do not have this prefix. I've opened a fix here:
https://github.com/huggingface/diffusers/pull/10545

PR #10549, opened 2025-01-13
Scheduling fixes on MPS

hlky:
# What does this PR do?

These fixes for issues with `int64` and `randn_like` are cherry-picked from #10442 where we investigated CI failures on MPS, that PR is now only updating MPS workflow so we can separately review the potential effects of changing `int64`->`int32`.

## Who can review?

### DIFF
--- a/src/diffusers/schedulers/scheduling_consistency_models.py
+++ b/src/diffusers/schedulers/scheduling_consistency_models.py
@@ -207,7 +207,7 @@ class CMStochasticIterativeScheduler(SchedulerMixin, ConfigMixin):
                     f" {self.config.num_train_timesteps}."
                 )
 
-            timesteps = np.array(timesteps, dtype=np.int64)
+            timesteps = np.array(timesteps, dtype=np.int32)
             self.custom_timesteps = True
         else:
             if num_inference_steps > self.config.num_train_timesteps:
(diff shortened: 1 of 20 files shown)


yiyixuxu:
maybe it is easier to only use np32 for MPs? Some of the models we recently integrated are very sensitive to precision (e.g. MOCHI, LTX) 

cc @bghira here for this opinions too

bghira:
Sana is especially sensitive but it could be like the RoPE for Flux where we went from fp64 to fp32 and saw no real degradation. if it won't work on MPS maybe some CPU fallback code can work for those systems, but that sounds like an upstream pytorch limitation.

i guess i'd give it a whirl and see if the known sensitive models have an issue, and document potential instabilities with Pytorch on MPS (which is in general a good idea to hamper expectations)

hlky:
Given the timesteps range casting `int64`->`int32` should be lossless, no? and when the timestep is cast to float type before model `int32`->`float16` etc. should also be lossless, no? Anyway it seems like the main fix for CI failure is 

https://github.com/huggingface/diffusers/blob/28c1f3b97f1690e1890175d58674082ee0f0156e/tests/schedulers/test_schedulers.py#L725

There are some issues on PyTorch regarding `*_like` failures on MPS.

`self.timesteps` is cast to `torch.int64` in some and `add_noise` (which is what the failing test was for) already handles casting for MPS, so we can revert these `np.int64`->`np.int32` changes.

hlky:
`np.int64`->`np.int32` changes were not needed. With this PR scheduler tests on MPS are all passing.

```
917 passed, 15 skipped, 46 deselected, 15 warnings in 15.48s
```

yiyixuxu (approved):
thanks!

PR #10551, opened 2025-01-13
[LoRA] improve failure handling for peft.

sayakpaul:
# What does this PR do?

In case, `inject_adapter_in_model()` fails midway for whatever reason before injecting `peft_config` into the base model, we should handle have a proper guard. This PR adds hat guard.

### DIFF
--- a/src/diffusers/loaders/peft.py
+++ b/src/diffusers/loaders/peft.py
@@ -302,7 +302,7 @@ class PeftAdapterMixin:
                 incompatible_keys = set_peft_model_state_dict(self, state_dict, adapter_name, **peft_kwargs)
             except Exception as e:
                 # In case `inject_adapter_in_model()` was unsuccessful even before injecting the `peft_config`.
-                if getattr(self, "peft_config", None) is not None:
+                if hasattr(self, "peft_config"):
                     for module in self.modules():
                         if isinstance(module, BaseTunerLayer):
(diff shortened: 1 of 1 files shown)


src/diffusers/loaders/peft.py
```diff
@@ -300,15 +300,17 @@ def load_lora_adapter(self, pretrained_model_name_or_path_or_dict, prefix="trans
...
-                    if isinstance(module, BaseTunerLayer):
-                        active_adapters = module.active_adapters
-                        for active_adapter in active_adapters:
-                            if adapter_name in active_adapter:
-                                module.delete_adapter(adapter_name)
-
-                self.peft_config.pop(adapter_name)
+            except Exception as e:
+                # In case `inject_adapter_in_model()` was unsuccessful even before injecting the `peft_config`.
+                if getattr(self, "peft_config", None) is not None:
```
BenjaminBossan, line 305: Nit: Could be simplified to:
```suggestion
                if hasattr(self, "peft_config"):
```

BenjaminBossan (approved):
Thanks for this fix. It assumes that the PEFT config is always added first. This is probably a safe assumption, I just want to be sure.

sayakpaul:
@DN6 okay for you to merge?

sayakpaul:
It doesn't replace. It's orthogonal.

PR #10556, opened 2025-01-13
Fix Nightly AudioLDM2PipelineFastTests

hlky:
# What does this PR do?

https://github.com/huggingface/diffusers/actions/runs/12738450237/job/35517796258

`test_cpu_offload_forward_pass_twice`/`test_model_cpu_offload_forward_pass` by updating signature of overridden `enable_model_cpu_offload`

`test_dict_tuple_outputs_equivalent` fails ~1e-5 over tolerance, `AssertionError: np.float32(0.00021260581) not less than 0.0002`, it's already noted that tolerance is increased due to large composite model so should be ok to increase it a bit more.

`test_audioldm2_tts` should pass with `phonemizer` installed.

Leaving `test_sequential_offload_forward_pass_twice - NotImplementedError: Cannot copy out of meta tensor; no data!` as a TODO.

## Who can review?

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -74,7 +74,6 @@ jobs:
           python -m uv pip install -e [quality,test]
           pip uninstall accelerate -y && python -m uv pip install -U accelerate@git+https://github.com/huggingface/accelerate.git
           python -m uv pip install pytest-reportlog
-          python -m uv pip install phonemizer
       - name: Environment
         run: |
           python utils/print_env.py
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly_tests.yml
```diff
@@ -74,6 +74,7 @@ jobs:
           python -m uv pip install -e [quality,test]
           pip uninstall accelerate -y && python -m uv pip install -U accelerate@git+https://github.com/huggingface/accelerate.git
           python -m uv pip install pytest-reportlog
+          python -m uv pip install phonemizer
```
DN6, line 77: We can update the test dep here
https://github.com/huggingface/diffusers/blob/edb8c1bce67e81f0de90a7e4c16b2f6537d39f2d/setup.py#L212

And it will get picked up when running `python -m uv pip install -e [quality,test]`

hlky: Nice, thanks!

PR #10563, opened 2025-01-13
Fix documentation for FluxPipeline

ohm314:
Found a tiny mistake in the documentation of FluxPipeline, more specifically the part showing how the Flux model can be loaded in a quantized way. In the code example the text encoder model is passed to the wrong argument in the FluxPipeline.from_pretrained function.

# What does this PR do?

## Before submitting

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. @stevhliu will be interested?


stevhliu (approved):
Good catch, thanks for the fix 🤗

Issue #10550, opened 2025-01-13
[LoRA] loading LoRA into a quantized base model


sayakpaul:
Similar issues: 

1. https://github.com/huggingface/diffusers/issues/10512
2. https://github.com/huggingface/diffusers/issues/10496

<details>
<summary>Reproduction</summary>

```py
import torch
from diffusers import BitsAndBytesConfig as DiffusersBitsAndBytesConfig, FluxTransformer2DModel, FluxPipeline
from huggingface_hub import hf_hub_download


transformer_8bit = FluxTransformer2DModel.from_pretrained(
    "black-forest-labs/FLUX.1-dev",
    subfolder="transformer",
    quantization_config=DiffusersBitsAndBytesConfig(load_in_8bit=True),
    torch_dtype=torch.bfloat16,
)
pipe = FluxPipeline.from_pretrained(
    "black-forest-labs/FLUX.1-dev",
    transformer=transformer_8bit,
    torch_dtype=torch.bfloat16,
).to("cuda")

pipe.load_lora_weights(
    hf_hub_download("ByteDance/Hyper-SD", "Hyper-FLUX.1-dev-8steps-lora.safetensors"), 
    adapter_name="hyper-sd"
)
pipe.set_adapters("hyper-sd", adapter_weights=0.125)

prompt = "A robot made of exotic candies and chocolates of different kinds. The background is filled with confetti and celebratory gifts."

image = pipe(
    prompt=prompt,
    height=1024,
    width=1024,
    max_sequence_length=512,
    num_inference_steps=8,
    guidance_scale=50,
    generator=torch.Generator().manual_seed(42),
).images[0]
image[0].save("out.jpg")
```

</details>

Happens on `main` as well as `v0.31.0-release` branch as well. 

<details>
<summary>Error</summary>

```bash
Traceback (most recent call last):
  File "/home/sayak/diffusers/load_loras_flux.py", line 18, in <module>
    pipe.load_lora_weights(
  File "/home/sayak/diffusers/src/diffusers/loaders/lora_pipeline.py", line 1846, in load_lora_weights
    self.load_lora_into_transformer(
  File "/home/sayak/diffusers/src/diffusers/loaders/lora_pipeline.py", line 1948, in load_lora_into_transformer
    inject_adapter_in_model(lora_config, transformer, adapter_name=adapter_name, **peft_kwargs)
[...]

PR #10576, opened 2025-01-14
[Tests] add: test to check 8bit bnb quantized models work with lora loading.

sayakpaul:
# What does this PR do?

To prevent from https://github.com/huggingface/diffusers/issues/10550. Only adds this test to 8bit. 4bit has issues and I will open a follow-up in a bit.

Slightly related: https://github.com/huggingface/diffusers/issues/10550.

### DIFF
--- a/tests/quantization/bnb/test_mixed_int8.py
+++ b/tests/quantization/bnb/test_mixed_int8.py
@@ -512,7 +512,7 @@ class SlowBnb8bitFluxTests(Base8bitTests):
         self.assertTrue(max_diff < 1e-3)
 
     @require_peft_version_greater("0.14.0")
-    def test_lora_loading_works(self):
+    def test_lora_loading(self):
         self.pipeline_8bit.load_lora_weights(
             hf_hub_download("ByteDance/Hyper-SD", "Hyper-FLUX.1-dev-8steps-lora.safetensors"), adapter_name="hyper-sd"
         )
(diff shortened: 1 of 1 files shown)


tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -509,6 +511,29 @@ def test_quality(self):
         max_diff = numpy_cosine_similarity_distance(expected_slice, out_slice)
         self.assertTrue(max_diff < 1e-3)
 
+    @require_peft_version_greater("0.14.0")
+    def test_lora_loading_works(self):
```
DN6, line 515:
```suggestion
    def test_lora_loading(self):
```

DN6: Can we add 4bit testing too?

sayakpaul: Contingent on https://github.com/huggingface/diffusers/pull/10578/ and is being added in the same PR.

DN6:
Could consolidate by adding this to: #10578

sayakpaul:
> Could consolidate by adding this to: #10578

I would prefer to do it differently because:

1. Both are independent of one another. We never supported 8bit bnb models + LoRA loading (PEFT-side bug which recently got fixed). 
2. This is why the test being added in this PR has a requirement on the PEFT version.
3. 4bit bnb + LoRA loading for Flux was supported in 0.31.0 but got broken. I think the scope of that is different from what we're doing in this PR.

DN6 (approved):

PR #10578, opened 2025-01-14
[WIP] [LoRA] feat: support loading loras into 4bit quantized models.

sayakpaul:
# What does this PR do?

We broke LoRA loading for 4bit BnB Flux in 0.32.0, when supporting Flux Control LoRAs (yeah only applies to Flux).

To reproduce:

<details>
<summary>Code</summary>

```py
import torch
from diffusers import BitsAndBytesConfig as DiffusersBitsAndBytesConfig, FluxTransformer2DModel, FluxPipeline


transformer_4bit = FluxTransformer2DModel.from_pretrained(
    subfolder="transformer",
    quantization_config=DiffusersBitsAndBytesConfig(
        load_in_4bit=True, bnb_4bit_quant_type="nf4", bnb_4bit_compute_dtype=torch.bfloat16
    ),
    torch_dtype=torch.bfloat16,
)
    transformer=transformer_4bit,
    torch_dtype=torch.bfloat16,
).to("cuda")

pipe.load_lora_weights(
    hf_hub_download("ByteDance/Hyper-SD", "Hyper-FLUX.1-dev-8steps-lora.safetensors"), 
    adapter_name="hyper-sd"
)
pipe.set_adapters("hyper-sd", adapter_weights=0.125)

image = pipe(
    prompt=prompt,
    height=1024,
    width=1024,
    max_sequence_length=512,
    num_inference_steps=8,
    guidance_scale=50,
).images[0]
image.save("out.jpg")
```

</details>

This above code will run with `v0.31.0-release` branch but will fail with `v0.32.0-release` along with `main`. This went uncaught because we don't test for it.

This PR attempts to partially fix the problem so that we can at least resort to a behavior similar to what was happening in the `v0.31.0-release` branch. Want to use this PR to refine how we're doing that. I want to ship this PR first and tackle the TODOs in a follow-up. Once this PR is done, we might have to do a patch release.

Related issue: https://github.com/huggingface/diffusers/issues/10550

### DIFF
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -1982,9 +1982,19 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
                 out_features = state_dict[lora_B_weight_name].shape[0]
 
                 # This means there's no need for an expansion in the params, so we simply skip.
-                if tuple(module_weight.shape) == (out_features, in_features):
+                module_weight_shape = module_weight.shape
+                expansion_shape = (out_features, in_features)
+                quantization_config = getattr(transformer, "quantization_config", None)
+                if quantization_config and quantization_config.quant_method == "bitsandbytes":
+                    if quantization_config.load_in_4bit:
+                        expansion_shape = torch.Size(expansion_shape).numel()
+                        expansion_shape = ((expansion_shape + 1) // 2, 1)
+
+                if tuple(module_weight_shape) == expansion_shape:
                     continue
 
+                # TODO (sayakpaul): We still need to consider if the module we're expanding is
+                # quantized and handle it accordingly if that is the case.
                 module_out_features, module_in_features = module_weight.shape
                 debug_message = ""
                 if in_features > module_in_features:
@@ -2080,13 +2090,22 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
             base_weight_param = transformer_state_dict[base_param_name]
             lora_A_param = lora_state_dict[f"{prefix}{k}.lora_A.weight"]
 
-            if base_weight_param.shape[1] > lora_A_param.shape[1]:
+            # TODO (sayakpaul): Handle the cases when we actually need to expand.
+            base_out_feature_shape = base_weight_param.shape[1]
+            lora_A_out_feature_shape = lora_A_param.shape[1]
+            quantization_config = getattr(transformer, "quantization_config", None)
+            if quantization_config and quantization_config.quant_method == "bitsandbytes":
+                if quantization_config.load_in_4bit:
+                    lora_A_out_feature_shape = lora_A_param.shape.numel()
+                    lora_A_out_feature_shape = ((lora_A_out_feature_shape + 1) // 2, 1)[1]
+
+            if base_out_feature_shape > lora_A_out_feature_shape:
                 shape = (lora_A_param.shape[0], base_weight_param.shape[1])
                 expanded_state_dict_weight = torch.zeros(shape, device=base_weight_param.device)
                 expanded_state_dict_weight[:, : lora_A_param.shape[1]].copy_(lora_A_param)
                 lora_state_dict[f"{prefix}{k}.lora_A.weight"] = expanded_state_dict_weight
                 expanded_module_names.add(k)
-            elif base_weight_param.shape[1] < lora_A_param.shape[1]:
+            elif lora_A_out_feature_shape < lora_A_out_feature_shape:
                 raise NotImplementedError(
                     f"This LoRA param ({k}.lora_A.weight) has an incompatible shape {lora_A_param.shape}. Please open an issue to file for a feature request - https://github.com/huggingface/diffusers/issues/new."
                 )
--- a/tests/quantization/bnb/test_4bit.py
+++ b/tests/quantization/bnb/test_4bit.py
@@ -20,6 +20,7 @@ import unittest
 import numpy as np
 import pytest
 import safetensors.torch
+from huggingface_hub import hf_hub_download
 
 from diffusers import BitsAndBytesConfig, DiffusionPipeline, FluxTransformer2DModel, SD3Transformer2DModel
 from diffusers.utils import is_accelerate_version, logging
@@ -32,6 +33,7 @@ from diffusers.utils.testing_utils import (
     numpy_cosine_similarity_distance,
     require_accelerate,
     require_bitsandbytes_version_greater,
+    require_peft_version_greater,
     require_torch,
     require_torch_gpu,
     require_transformers_version_greater,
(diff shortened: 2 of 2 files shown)


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -1982,9 +1982,19 @@ def _maybe_expand_transformer_param_shape_or_error_(
...
 
                 # This means there's no need for an expansion in the params, so we simply skip.
-                if tuple(module_weight.shape) == (out_features, in_features):
+                module_weight_shape = module_weight.shape
+                expansion_shape = (out_features, in_features)
+                quantization_config = getattr(transformer, "quantization_config", None)
+                if quantization_config and quantization_config.quant_method == "bitsandbytes":
+                    if quantization_config.load_in_4bit:
+                        expansion_shape = torch.Size(expansion_shape).numel()
+                        expansion_shape = ((expansion_shape + 1) // 2, 1)
```
sayakpaul, line 1991: Only 4bit bnb models flatten.

BenjaminBossan: How about adding a comment along the lines of: "Handle 4bit bnb weights, which are flattened and compress 2 params into 1".

I'm not quite sure why we need (shape+1) // 2, maybe this could be added to the comment too.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -1982,9 +1982,19 @@ def _maybe_expand_transformer_param_shape_or_error_(
                 out_features = state_dict[lora_B_weight_name].shape[0]
 
                 # This means there's no need for an expansion in the params, so we simply skip.
-                if tuple(module_weight.shape) == (out_features, in_features):
+                module_weight_shape = module_weight.shape
+                expansion_shape = (out_features, in_features)
+                quantization_config = getattr(transformer, "quantization_config", None)
+                if quantization_config and quantization_config.quant_method == "bitsandbytes":
```
BenjaminBossan, line 1988: Would it make sense to have a utility function to get the shape to avoid code duplication?

BenjaminBossan (commented):
Thanks for fixing the issue with loading LoRA into Flux models that are quantized with 4 bit bnb.

The issue of the actual parameter shape is a common trap, it would be great if the _original_ shape could be retrieved from bnb, not sure if there is a method for that.

Sayak, could you quickly sketch why this used to work and no longer does? IIRC, there was no special handling for 4 bit bnb previously, was there?

sayakpaul:
> Sayak, could you quickly sketch why this used to work and no longer does? IIRC, there was no special handling for 4 bit bnb previously, was there?

It used to because we didn't have any support for loading Flux Control LoRA, some relevant pieces of vital code:

https://github.com/huggingface/diffusers/blob/be62c85cd973f2001ab8c5d8919a9a6811fc7e43/src/diffusers/loaders/lora_pipeline.py#L1941

https://github.com/huggingface/diffusers/blob/be62c85cd973f2001ab8c5d8919a9a6811fc7e43/src/diffusers/loaders/lora_pipeline.py#L2055

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Yes, this needs to happen. As I mentioned this PR is very much a PoC to gather feedback and I will refine it. But I wanted to first explore if this a good way to approach the problem.

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Yeah, this comes from `bitsandbytes`. Cc: @matthewdouglas

BenjaminBossan:
> It used to because we didn't have any support for loading Flux Control LoRA

Okay, so Flux control LoRA + 4bit bnb was never possible. From the initial description, I got the wrong impression that this is a full regression.

sayakpaul:
> Okay, so Flux control LoRA + 4bit bnb was never possible. From the initial description, I got the wrong impression that this is a full regression.

1. Flux control LoRA + 4bit bnb was never possible.
2. However, Flux LoRA + 4bit bnb was possible. So, I think Flux Control LoRA support introduced a regression at this point as we cannot do Flux LoRA + 4bit bnb anymore? What am I missing?

BenjaminBossan:
> 1. Flux control LoRA + 4bit bnb was never possible.
> 2. However, Flux LoRA + 4bit bnb was possible. So, I think Flux Control LoRA support introduced a regression at this point as we cannot do Flux LoRA + 4bit bnb anymore? What am I missing?

We fully agree there. What I meant is that I misunderstood your original message to mean that Flux control LoRA + 4 bit bnb used to work and was trying to understand why it breaks now. This is what I meant by "full regression".

src/diffusers/loaders/lora_pipeline.py
matthewdouglas: This is for rounding, i.e. if `expansion_shape` is odd it will have an additional 8bit tensor with just one value packed into it.

matthewdouglas:
> The issue of the actual parameter shape is a common trap, it would be great if the _original_ shape could be retrieved from bnb, not sure if there is a method for that.

@BenjaminBossan That's a good point. The `Params4bit` and `Linear4bit` should have a `quant_state` attribute on them, which has a `shape` to indicate the original shape. I haven't used this much myself but it could be an option here.

sayakpaul:
> We fully agree there. What I meant is that I misunderstood your original message to mean that Flux control LoRA + 4 bit bnb used to work and was trying to understand why it breaks now. This is what I meant by "full regression".

Thanks, @BenjaminBossan. I will work on the suggestions to make it ready for another review pass.

> That's a good point. The Params4bit and Linear4bit should have a quant_state attribute on them, which has a shape to indicate the original shape. I haven't used this much myself but it could be an option here.

Thanks @matthewdouglas! Will try to see if we can use it here.

sayakpaul:
Just confirmed that this works:

```py
from diffusers import FluxTransformer2DModel

model = FluxTransformer2DModel.from_pretrained(
    "hf-internal-testing/flux.1-dev-nf4-pkg", subfolder="transformer"
)
print(model.context_embedder.weight.quant_state.shape)
# torch.Size([3072, 4096])
```

sayakpaul:
@BenjaminBossan I have addressed the rest of the feedback. PTAL.

@DN6 ready for your review, too. Just as FYI, I have run all the LoRA related integration tests for Flux and they all pass.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2098,6 +2108,28 @@ def _maybe_expand_lora_state_dict(cls, transformer, lora_state_dict):
...
+        def _get_weight_shape(weight: torch.Tensor):
+            return weight.quant_state.shape if weight.__class__.__name__ == "Params4bit" else weight.shape
+
+        if base_module is not None:
+            return _get_weight_shape(base_module.weight)
+        elif base_weight_param_name is not None:
+            module_path = (
+                base_weight_param_name.rsplit(".weight", 1)[0]
+                if base_weight_param_name.endswith(".weight")
+                else base_weight_param_name
```
DN6, line 2126: In what case would this param name not end with `weight`? Since the subsequent call to `get_weight_shape` assumes that there is a weight attribute?

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -2098,6 +2108,28 @@ def _maybe_expand_lora_state_dict(cls, transformer, lora_state_dict):
...
         return lora_state_dict
 
+    @staticmethod
+    def _calculate_module_shape(
+        model: "torch.nn.Module",
+        base_module: "torch.nn.Linear" = None,
+        base_weight_param_name: str = None,
+    ) -> "torch.Size":
+        def _get_weight_shape(weight: torch.Tensor):
+            return weight.quant_state.shape if weight.__class__.__name__ == "Params4bit" else weight.shape
```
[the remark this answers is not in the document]
sayakpaul: They do.

src/diffusers/loaders/lora_pipeline.py
sayakpaul: True. Let me modify that. Thanks for catching.

sayakpaul:
@DN6 applied changes and re-run all the required tests and they passed.

DN6 (approved):

BenjaminBossan:
Thanks for fixing this, Sayak.

PR #10592, opened 2025-01-16
[chore] update notes generation spaces

sayakpaul:
# What does this PR do?

We rely on https://huggingface.co/spaces/lysandre/github-release to fetch release notes but sometimes, it can be confusing to determine the appropriate tag and branch to obtain the release notes. 

This changed Space (https://huggingface.co/spaces/sayakpaul/auto-release-notes-diffusers) bypasses this need by automatically fetching the appropriate tag and branch so that we don't have to struggle. It also has a token secret, so, we don't have sweat that part.

Screenshot 2025-01-16 at 9 00 09 AM


DN6 (approved):
Thanks!

PR #10602, opened 2025-01-17
Enable dreambooth lora finetune example on other devices

jiqing-feng:
This PR mainly changed 2 points to enable the example on other devices:
1. Replace `torch.cuda.amp.autocast()` by `torch.amp.autocast(device)` so other devices can also use it
2. Empty cache for xpu.


kaixuanliu:
LGTM

jiqing-feng:

examples/dreambooth/train_dreambooth_lora.py
```diff
@@ -151,14 +151,14 @@ def log_validation(
     if args.validation_images is None:
         images = []
         for _ in range(args.num_validation_images):
-            with torch.cuda.amp.autocast():
+            with torch.amp.autocast(pipeline.device.type):
```
sayakpaul, line 154:
```suggestion
            with torch.amp.autocast(accelerator.device.type):
```

examples/dreambooth/train_dreambooth_lora.py
```diff
@@ -151,14 +151,14 @@ def log_validation(
...
                 image = pipeline(**pipeline_args, generator=generator).images[0]
                 images.append(image)
     else:
         images = []
         for image in args.validation_images:
             image = Image.open(image)
-            with torch.cuda.amp.autocast():
+            with torch.amp.autocast(pipeline.device.type):
```
sayakpaul, line 161: Same as abpve.

examples/dreambooth/train_dreambooth_lora.py
```diff
@@ -177,7 +177,10 @@ def log_validation(
             )
 
     del pipeline
-    torch.cuda.empty_cache()
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
+    if hasattr(torch, "xpu") and torch.xpu.is_available():
+        torch.xpu.empty_cache()
```
sayakpaul, line 183: Do we want to introduce these changes to the `free_memory()` here?
https://github.com/huggingface/diffusers/blob/328e0d20a7b996f9bdb04180457eb08c1b42a76e/src/diffusers/training_utils.py#L290

And then call the `free_memory()` method here instead?

examples/dreambooth/train_dreambooth_lora.py
```diff
@@ -831,6 +834,8 @@ def main(args):
             del pipeline
             if torch.cuda.is_available():
                 torch.cuda.empty_cache()
+            if hasattr(torch, "xpu") and torch.xpu.is_available():
+                torch.xpu.empty_cache()
```
sayakpaul, line 838: Same as above.

examples/dreambooth/train_dreambooth_lora.py
```diff
@@ -1085,7 +1090,10 @@ def compute_text_embeddings(prompt):
         tokenizer = None
 
         gc.collect()
-        torch.cuda.empty_cache()
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
+        if hasattr(torch, "xpu") and torch.xpu.is_available():
+            torch.xpu.empty_cache()
```
sayakpaul, line 1096: Same as above.

examples/dreambooth/train_dreambooth_lora.py
```diff
@@ -793,7 +796,7 @@ def main(args):
         cur_class_images = len(list(class_images_dir.iterdir()))
 
         if cur_class_images < args.num_class_images:
-            torch_dtype = torch.float16 if accelerator.device.type == "cuda" else torch.float32
+            torch_dtype = torch.float16 if accelerator.device.type in ("cuda", "xpu") else torch.float32
```
sayakpaul, line 799: We cannot train in mixed-precision?

sayakpaul (commented):
Thanks for contribution! I just left some comments, LMK if they make sense.

examples/dreambooth/train_dreambooth_lora.py
jiqing-feng: I wasn't aware of it, just want to keep the XPU device has the same precision as CUDA.

jiqing-feng:
Hi @sayakpaul . I have fixed your comments. For mixed-precision, I am not sure because we have `torch_dtype` and [weight_dtype](https://github.com/huggingface/diffusers/blob/main/examples/dreambooth/train_dreambooth_lora.py#L885-L889) here. In my view, the mix-precision should work on lora module but not SD based model because the weight_dtype has already been converted to the mix-precision.

examples/dreambooth/train_dreambooth_lora.py
```diff
@@ -75,6 +75,13 @@
 logger = get_logger(__name__)
 
 
+def free_memory():
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
+    if hasattr(torch, "xpu") and torch.xpu.is_available():
+        torch.xpu.empty_cache()
```
sayakpaul, line 82: Let's directly update this function no?
https://github.com/huggingface/diffusers/blob/4ace7d0483d9b5016575a2d51119c33172c005ed/src/diffusers/training_utils.py#L290

sayakpaul (approved):
Just one comment and we should be good to go.

Have you verified if it works effectively? If so, could you share some results?

jiqing-feng:
> Just one comment and we should be good to go.
> 
> Have you verified if it works effectively? If so, could you share some results?

The result of Intel Xeon CPU is:
image


Sorry, I cannot release the latency performance data cause it is not allowed by Intel.

sayakpaul:
> For cuda, it just enabled the DDP finetune and had no impact on the result as previously.

What do you mean by this? I don't see the dog from the training set appear here. Am I missing something?

jiqing-feng:
> > For cuda, it just enabled the DDP finetune and had no impact on the result as previously.
> 
> What do you mean by this? I don't see the dog from the training set appear here. Am I missing something?

Sorry for I didn't upload the cuda result image successfully, I will run again to get the result and give you feedback ASAP.

jiqing-feng:
Hi @sayakpaul , here is the cuda result runs on 2*A100 cards
image

sayakpaul:
And how about the XPU result?

jiqing-feng:
> And how about the XPU result?

Just got the XPU result with 2 DDP across 2 cards:
image

The XPU needs to tune the args to get a high-quality result.

sayakpaul:
Nice, this is good.

jiqing-feng:
The failed test seems unrelated to my changes. Please let me know what needs to be changed or request any other reviewers before merging. Thanks!

sayakpaul:
Thanks much!

PR #10603, opened 2025-01-17
[feat]Add strength in flux_fill pipeline (denoising strength for fluxfill)

Suprhimp:
# What does this PR do?

allows the fluxfill pipeline to have a denoising strength (denoising strength).

## Before submitting

## Who can review?
@yiyixuxu and @asomoza

I use fluxfill pipeline to edit image or outpatin. Yes it works very well but I don't agree `Flux Fill pipeline does not require `strength` as an input like regular inpainting pipelines` this line in the docs.

When remove object in the image, We need denoising strength I guess. without denoising strength I can't get clean image that remove object that I want.
let me gives you an example I tested.

mask
original_image

this is mask and original image that I want to edit.

And this is the result with denoising strength (0.6 with revised pipeline) and none(default pipeline)

cbimage (51)
7fabf5e8-9d0d-47d8-af7b-0371b4f29f21

I did lama inpaint to remove more clearly in both case but output is different.

As you can see with denoising strength we can have more control with the image quality. So I think there is no reason to not use denoising strength in fluxfill pipeline 

I changed the fluxfill pipeline code with reference sdxl inpaint pipeline.
So, I think there would be things that need to be fixed.

Thanks for reading :)


src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -225,10 +225,9 @@ def __init__(
         # Flux latents are turned into 2x2 patches and packed. This means the latent width and height has to be divisible
         # by the patch size. So the vae scale factor is multiplied by the patch size to account for this
         self.image_processor = VaeImageProcessor(vae_scale_factor=self.vae_scale_factor * 2)
-        latent_channels = self.vae.config.latent_channels if getattr(self, "vae", None) else 16
         self.mask_processor = VaeImageProcessor(
             vae_scale_factor=self.vae_scale_factor * 2,
-            vae_latent_channels=latent_channels,
+            vae_latent_channels=self.vae.config.latent_channels,
```
hlky, line 230: We need to use `latent_channels` here not the config directly. This allows pipelines to be used without the component e.g. `FluxFillPipeline(vae=None, ...)`.

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -627,6 +659,8 @@ def disable_vae_tiling(self):
     # Copied from diffusers.pipelines.flux.pipeline_flux.FluxPipeline.prepare_latents
```
hlky, line 659: Can we copy from `FluxImg2ImgPipeline.prepare_latents` or `FluxInpaintPipeline.prepare_latents`?

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -809,6 +866,10 @@ def __call__(
         self._joint_attention_kwargs = joint_attention_kwargs
         self._interrupt = False
 
+        original_image = image
```
hlky, line 869: Unused?

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -855,13 +952,13 @@ def __call__(
         if masked_image_latents is not None:
             masked_image_latents = masked_image_latents.to(latents.device)
         else:
-            image = self.image_processor.preprocess(image, height=height, width=width)
+            # image = self.image_processor.preprocess(image, height=height, width=width)
```
hlky, line 955:
```suggestion
```

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -838,9 +901,43 @@ def __call__(
             lora_scale=lora_scale,
         )
 
+        # 6. Prepare timesteps
```
hlky, line 904:
```suggestion
        # 4. Prepare timesteps
```

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -838,9 +901,43 @@ def __call__(
...
+        timesteps, num_inference_steps = self.get_timesteps(num_inference_steps, strength, device)
+
+        if num_inference_steps < 1:
+            raise ValueError(
+                f"After adjusting the num_inference_steps by strength parameter: {strength}, the number of pipeline"
+                f"steps is {num_inference_steps} which is < 1 and not appropriate for this pipeline."
+            )
+        latent_timestep = timesteps[:1].repeat(batch_size * num_images_per_prompt)
+
         # 4. Prepare latent variables
```
hlky, line 936:
```suggestion
        # 5. Prepare latent variables
```

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -855,13 +952,13 @@ def __call__(
         if masked_image_latents is not None:
```
hlky, line 952: Above `# 6. Prepare mask and masked image latents`.

hlky (commented):
I've left some comments. However, we should note that Flux Fill is intended to fill areas based on a text description and not for object removal.

Suprhimp:
Thankyou for the fast and kind review, I'll revise it and test :)

src/diffusers/pipelines/flux/pipeline_flux_fill.py
Suprhimp: Sure , That would be more clean. Thanks for the review :)

Suprhimp:
> I've left some comments. However, we should note that Flux Fill is intended to fill areas based on a text description and not for object removal.

Hmm, Maby I missed the intention than ;)
Although I think we can use fluxfill with more controllable, with denosing strength:)

@hlky Thanks for the review again, I changed the code that you've reviewed.

Should I change docs also?

asomoza:
Hi, you're saying you used lama to remove the object before doing this? If that's what you did, you're just essentially using this pipeline as an img2img one and the model as just a refiner.

The original fill model was trained to do this without the need of changing the strength and this is from the original implementation, I'm not saying that what you're doing is wrong though, people used redux in a more different and creative way and it works alright but I want to understand your use case since I haven't encountered the need to lower the strength to make it work.

Did you try what you're doing here with just the regular Flux model and the img2img pipeline?

Suprhimp:
>Did you try what you're doing here with just the regular Flux model and the img2img pipeline?

Thanks for your response :)

I tried with sdxl model with img2img model. But not flux image2image.
And yep I understand what you said, Maby I'm just using the fluxfill pipeline as an refiner.

As I use this pipeline an image edit pipeline, I just want to use the single pipeline as an object adder or outpainter but also object remover.

If I have to use my case with flux img2img pipeline I can do that, but as you know flux model is quite large.
So I'd like to use it as  refiner but also object adder or outpainter etc.

If I can use fluxfill pipeline with `strength` parameter, I don't need any other pipeline :)

hlky:
@Suprhimp To confirm, you're using lama cleaner to remove the object first?

Suprhimp:
> To confirm, you're using lama cleaner to remove the object first?

Yep I used lamacleaner before both case (my pipeline, and default pipeline)

ukaprch:
lama with the refiner is very good at removing objects. I thought the whole point of Flux fill was to replace or insert an object. Big difference there.

Suprhimp:
Hmm I think even we use fluxfill without lamainpainter to replace or insert an object, There will be certian point to need stength. Maby, keep the original color masked but change object, or change object but referenced based image in masked area.

You guys think it is unecessary to use strength referer to the orignal masked area in replace or insert an object?

asomoza:
> There will be certian point to need stength. Maby, keep the original color masked but change object, or change object but referenced based image in masked area.

This are valid use cases but that it's again just a nasked img2img, in fact, if you want to preserve as much of the original image intact, you should always do it with img2img instead of this pipeline, and just replace the new part in the original image.

Still I think they're valid points and also the one that you don't want to switch models because they're really big, but this will make the pipeline more complex, @yiyixuxu WDYT?

jonluca:
This would be a welcome addition, would be good to be able to use strength with flux fill

asomoza:
> This would be a welcome addition, would be good to be able to use strength with flux fill

@jonluca we're still waiting for @yiyixuxu feedback here, but in the meantime, do you have some demo images of using the strength with the vanilla flux fill pipeline? I mean without using it as a refiner only or with another inpainting model like lama cleaner.

yiyixuxu:
@asomoza I think it's ok to support if you think the use case is meaningful :)

lvwilson:
I +1 this use case.

asomoza:
sure, I think it's worth it because of the VRAM usage of Flux, I'll do some quick tests as soon as possible. In the meantime @Suprhimp would you mind resolving the conflicts or do you prefer if I take it over?

Suprhimp:
I'll try right away:) @asomoza

Suprhimp:
@asomoza  I fixed conflict and I did test with my code, it seems work well to me

asomoza:
@bot /style

asomoza (approved):

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -636,28 +669,38 @@ def prepare_latents(
...
+        image_latents = self._encode_vae_image(image=image, generator=generator)
+        if batch_size > image_latents.shape[0] and batch_size % image_latents.shape[0] == 0:
+            # expand init_latents for batch_size
+            additional_image_per_prompt = batch_size // image_latents.shape[0]
+            image_latents = torch.cat([image_latents] * additional_image_per_prompt, dim=0)
+        elif batch_size > image_latents.shape[0] and batch_size % image_latents.shape[0] != 0:
             raise ValueError(
-                f"You have passed a list of generators of length {len(generator)}, but requested an effective batch"
-                f" size of {batch_size}. Make sure the batch size matches the length of the generators."
+                f"Cannot duplicate `image` of batch size {image_latents.shape[0]} to {batch_size} text prompts."
```
asomoza, line 696: this part it's not the same as the `prepare_latents`  in FluxImg2ImgPipeline

asomoza (changes requested):
I took the liberty to fix the bad name in the `prepare_latents` copy, but now we have an issue with it.

src/diffusers/pipelines/flux/pipeline_flux_fill.py
Suprhimp: Thanks for the detail review, and yes, since FluxImg2ImgPipeline was updated 2 weeks ago, It has to be changed.
So I updated `prepare_latents` function as same with in FluxImg2ImgPipeline.

And also I tested with my code, it works well. (honestly, it works better than before lol)

asomoza (approved):

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -838,9 +897,37 @@ def __call__(
...
+        # 4. Prepare timesteps
+        sigmas = np.linspace(1.0, 1 / num_inference_steps, num_inference_steps) if sigmas is None else sigmas
+        image_seq_len = (int(height) // self.vae_scale_factor // 2) * (int(width) // self.vae_scale_factor // 2)
+        mu = calculate_shift(
+            image_seq_len,
+            self.scheduler.config.base_image_seq_len,
+            self.scheduler.config.max_image_seq_len,
+            self.scheduler.config.base_shift,
+            self.scheduler.config.max_shift,
+        )
```
hlky, line 909:
```suggestion
        mu = calculate_shift(
            image_seq_len,
            self.scheduler.config.get("base_image_seq_len", 256),
            self.scheduler.config.get("max_image_seq_len", 4096),
            self.scheduler.config.get("base_shift", 0.5),
            self.scheduler.config.get("max_shift", 1.15),
        )
```

We need to keep this for scheduler compatibility, see https://github.com/huggingface/diffusers/pull/10384#issue-2758882751

hlky (approved):

src/diffusers/pipelines/flux/pipeline_flux_fill.py
asomoza: @Suprhimp want to update the code or should I do it and merge?

Suprhimp: Than can I ask for help for this part please? 😂

hlky: We only need to hit "Commit suggestion" button

asomoza: yeah I know but I always like to give the opportunity first to the contributor (to learn) and also the commit has a different author if we do it.

Suprhimp:
Thank you! I wanted to revise the last part as well, but I couldn't respond because I couldn't use the computer at that time. But I wanted to merge this PR quickly, so I asked for help.

PR #10347, opened 2024-12-23
[core] Layerwise Upcasting

a-r-r-o-w:
[...continuation of #9177]

Pytorch has had support for `float8_e4m3fn` and `float8_e5m2` as storage dtypes for a while now. This allows one to store model weights in a lower precision dtype and upcast them on-the-fly when a layer is required for proceeding with computation.

<details>
<summary> Code </summary>

```python
import argparse
import gc
import pathlib
import traceback

import git
import pandas as pd
import torch
from diffusers import AllegroPipeline, CogVideoXPipeline, LattePipeline, FluxPipeline, HunyuanVideoPipeline, HunyuanVideoTransformer3DModel, MochiPipeline, LTXPipeline
from diffusers.utils import export_to_video, load_image
from diffusers.utils.logging import set_verbosity_info, set_verbosity_debug
from tabulate import tabulate


repo = git.Repo(path="/home/aryan/work/diffusers")
branch = repo.active_branch


def pretty_print_results(results, precision: int = 3):
    def format_value(value):
        if isinstance(value, float):
            return f"{value:.{precision}f}"
        return value

    filtered_table = {k: format_value(v) for k, v in results.items()}
    print(tabulate([filtered_table], headers="keys", tablefmt="pipe", stralign="center"))


def benchmark_fn(f, *args, **kwargs):
    torch.cuda.synchronize()
    start = torch.cuda.Event(enable_timing=True)
    end = torch.cuda.Event(enable_timing=True)

    start.record()
    output = f(*args, **kwargs)
    end.record()
    torch.cuda.synchronize()
    elapsed_time = round(start.elapsed_time(end) / 1000, 3)

    return elapsed_time, output


def prepare_flux(dtype: torch.dtype, compile: bool = False, **kwargs) -> None:
    model_id = "black-forest-labs/Flux.1-Dev"
    cache_dir = "/raid/.cache/huggingface"

    pipe = FluxPipeline.from_pretrained(model_id, torch_dtype=dtype, cache_dir=cache_dir)
    pipe.to("cuda")

    if compile:
        pipe.transformer = torch.compile(
            pipe.transformer, mode="max-autotune-no-cudagraphs", fullgraph=True, dynamic=False
        )

    for key, value in list(kwargs.items()):
        if torch.is_tensor(value):
            kwargs[key] = value.to(device="cuda", dtype=dtype)

    generation_kwargs = {
        "prompt": "A cat holding a sign that says hello world",
        "height": 768,
        "width": 768,
[...]

Issue #9343, opened 2024-09-02
FLUX error when loading with low_cpu_mem_usage=False and ignore_mismatched_sizes=True


primecai:
### Describe the bug

I'd like to change the input layers of FLUX for training some img2img stuff, but got:
`TypeError: expected str, bytes or os.PathLike object, not NoneType`
when loading `FluxTransformer2DModel` with `low_cpu_mem_usage=False, ignore_mismatched_sizes=True`.

### Reproduction

```
from diffusers.models import FluxTransformer2DModel

transformer = FluxTransformer2DModel.from_pretrained(
                "black-forest-labs/FLUX.1-dev",
                subfolder="transformer",
                torch_dtype=weight_dtype,
                low_cpu_mem_usage=False, ignore_mismatched_sizes=True,
                revision=None, variant=None)
```

### Logs

```shell
expected str, bytes or os.PathLike object, not NoneType
Traceback (most recent call last):
  File "/home/xxxx/repos/xxxx/.venv/lib/python3.11/site-packages/diffusers/models/model_loading_utils.py", line 104, in load_state_dict
    file_extension = os.path.basename(checkpoint_file).split(".")[-1]
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen posixpath>", line 142, in basename
TypeError: expected str, bytes or os.PathLike object, not NoneType

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/xxxx/repos/xxxx/train.py", line xxxx, in <module>
    main()
  File "/home/xxxx/repos/xxxx/train.py", line xxx, in main
    transformer = load_flux(args, weight_dtype)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxxx/repos/xxxx/xxxx.py", line xx, in load_flux
    transformer = FluxTransformer2DModel.from_pretrained(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxxx/repos/xxxx/.venv/lib/python3.11/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/xxxx/repos/xxxx/.venv/lib/python3.11/site-packages/diffusers/models/modeling_utils.py", line 828, in from_pretrained
[...]

PR #10604, opened 2025-01-17
[FEAT] Model loading refactor

### LINKED ISSUE #10013: [Core] refactor model loading

SunMarc:
# What does this PR do?

Fixes #10013 . This PR refactors model loading in diffusers. Here's a list of major changes in this PR.  

- only two loading paths (`low_cpu_mem_usage=True` and `low_cpu_mem_usage = False`). We don't rely on `load_checkpoint_and_dispatch` anymore and we don't merge sharded checkpoint also.  
- support for sharded checkpoints for both loading paths
- `keep_module_in_fp32` support for sharded checkpoints
- better support for displaying warning due to error/unexpected/missing/mismatched keys

For `low_cpu_mem_usage = False`:
- Faster initialization (thanks to skipping the init + `assign_to_params_buffers`). I didn't benchmarked it but it should be as fast as `low_cpu_mem_usage=True` or maybe even faster. We did a similar PR in transformers thanks to @muellerzr. 
- Better `torch_dtype support` We don't initialize anymore the model in fp32 then cast the model to a specific dtype after finishing to load the weights. 

For `low_cpu_mem_usage = True` or `device_map!=None`:
- one path, we don't rely anymore on `load_checkpoint_and_dispatch`
- `device_map` support for quantization
- non persistance buffer support through `dispatch_model` ( the test you added is passing cc @hlky  ) 

Single format file: 
- Simplified the single file format loading through `from_pretrained`. This way we have the same features as this function (device_map, quantization ...). Feel free to share your opinion @DN6, I didn't expect to touch this but I felt that we could simplify a bit 

TODO (some items can be done in follow-up PRs):
- [x] Check if we have any regression / tests issues
- [x] Add more tests
- [ ] Deal with missing keys in the model for both paths (before, it only worked when `low_cpu_mem_usage=False` since we are  initializing the whole model)
- [x] Fix typing 
- [ ] Better support for offload with safetensors (like in transformers)

Please let me know your thoughts on the PR ! 

cc @sayakpaul, @DN6 , @yiyixuxu , @hlky , @a-r-r-o-w


SunMarc:
FLAX CPU failing test is unrelated, failing in other PRs too

src/diffusers/loaders/single_file_model.py
```diff
@@ -13,15 +13,11 @@
...
-import re
-from contextlib import nullcontext
 from typing import Optional
 
-import torch
 from huggingface_hub.utils import validate_hf_hub_args
 
-from ..quantizers import DiffusersAutoQuantizer
-from ..utils import deprecate, is_accelerate_available, logging
+from ..utils import deprecate, logging
```
sayakpaul, line 20: Will let @DN6 comment on the single-file related changes.

src/diffusers/models/model_loading_utils.py
```diff
@@ -223,72 +241,93 @@ def load_model_dict_into_meta(
...
         # bnb params are flattened.
         # gguf quants have a different shape based on the type of quantization applied
         if empty_state_dict[param_name].shape != param.shape:
             if (
                 is_quantized
                 and hf_quantizer.pre_quantized
-                and hf_quantizer.check_if_quantized_param(model, param, param_name, state_dict, param_device=device)
+                and hf_quantizer.check_if_quantized_param(
+                    model, param, param_name, state_dict, param_device=param_device
+                )
```
sayakpaul, line 288: (nit): not sure if this indentation is needed.

src/diffusers/models/model_loading_utils.py
```diff
@@ -188,23 +204,25 @@ def load_state_dict(
...
-    if hf_quantizer is None:
-        device = device or torch.device("cpu")
-    dtype = dtype or torch.float32
-    is_quantized = hf_quantizer is not None
+    """
+    This is somewhat similar to `_load_state_dict_into_model`, but deals with a model that has some or all of its
+    params on a `meta` device. It replaces the model params with the data from the `state_dict`
+    """
 
-    accepts_dtype = "dtype" in set(inspect.signature(set_module_tensor_to_device).parameters.keys())
```
sayakpaul, line 205: Perhaps an explanation on why this is going away would help the reviewer.

src/diffusers/models/model_loading_utils.py
```diff
@@ -188,23 +204,25 @@ def load_state_dict(
...
+    error_msgs = []
+    is_quantized = hf_quantizer is not None
     empty_state_dict = model.state_dict()
-    unexpected_keys = [param_name for param_name in state_dict if param_name not in empty_state_dict]
```
sayakpaul, line 207: Same as above.

src/diffusers/models/model_loading_utils.py
```diff
@@ -223,72 +241,93 @@ def load_model_dict_into_meta(
...
         else:
-            if accepts_dtype:
-                set_module_tensor_to_device(model, param_name, device, value=param, **set_module_kwargs)
-            else:
-                set_module_tensor_to_device(model, param_name, device, value=param)
+            set_module_tensor_to_device(model, param_name, param_device, value=param, **set_module_kwargs)
 
-    if named_buffers is None:
-        return unexpected_keys
+    return error_msgs, offload_index, state_dict_index
```
sayakpaul, line 307: I don't think `error_msgs` isn't populated in this method. Is that expected?

src/diffusers/models/model_loading_utils.py
```diff
@@ -223,72 +241,93 @@ def load_model_dict_into_meta(
...
+            hf_quantizer.create_quantized_param(model, param, param_name, param_device, state_dict, unexpected_keys)
         else:
-            if accepts_dtype:
-                set_module_tensor_to_device(model, param_name, device, value=param, **set_module_kwargs)
-            else:
-                set_module_tensor_to_device(model, param_name, device, value=param)
+            set_module_tensor_to_device(model, param_name, param_device, value=param, **set_module_kwargs)
 
-    if named_buffers is None:
-        return unexpected_keys
```
sayakpaul, line 259: Why is this going away?

src/diffusers/models/model_loading_utils.py
```diff
@@ -343,46 +382,6 @@ def _fetch_index_file(
     return index_file
 
 
-# Adapted from
```
sayakpaul, line 346: 👋

src/diffusers/models/modeling_utils.py
```diff
@@ -134,6 +172,57 @@ def find_tensor_attributes(module: nn.Module) -> List[Tuple[str, Tensor]]:
...
+        )
+        return False
+
+    # If the model does, the incoming `state_dict` and the `model_to_load` must be the same dtype
+    first_key = next(iter(model_to_load.state_dict().keys()))
+    if start_prefix + first_key in state_dict:
+        return state_dict[start_prefix + first_key].dtype == model_to_load.state_dict()[first_key].dtype
+
+    # For cases when the `state_dict` doesn't contain real weights to the model (`test_model_weights_reload_no_missing_tied_weights`)
+    return False
```
sayakpaul, line 202: Nice to be forward thinking but just as an FYI, diffusion models don't really use weight tying. So, we can remove code to handle weight tying from this PR to keep things relevant and add support as needed if that's not too much.

src/diffusers/models/modeling_utils.py
```diff
@@ -134,6 +172,57 @@ def find_tensor_attributes(module: nn.Module) -> List[Tuple[str, Tensor]]:
...
+
+
+@contextmanager
+def no_init_weights():
```
sayakpaul, line 206: Do we still need this when we're under `torch.device("meta")` or `init_empty_weights()`?

src/diffusers/models/modeling_utils.py
```diff
@@ -609,6 +698,8 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
         quantization_config = kwargs.pop("quantization_config", None)
         dduf_entries: Optional[Dict[str, DDUFEntry]] = kwargs.pop("dduf_entries", None)
         disable_mmap = kwargs.pop("disable_mmap", False)
+        state_dict = kwargs.pop("state_dict", None)
+        config = kwargs.pop("config", None)
```
sayakpaul, line 702: Are these two necessary for this PR? If not, let's maybe support these in a follow-up?

src/diffusers/models/modeling_utils.py
```diff
@@ -758,89 +851,106 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
-
-            # Convert the weights
-            from .modeling_pytorch_flax_utils import load_flax_checkpoint_in_pytorch_model
-
-            model = load_flax_checkpoint_in_pytorch_model(model, model_file)
-        else:
-            # in the case it is sharded, we have already the index
-            if is_sharded:
-                sharded_ckpt_cached_folder, sharded_metadata = _get_checkpoint_shard_files(
+        resolved_archive_file = None
```
sayakpaul, line 856: `resolved_archive_file` is a confusing name, IMO. From what I understand it's associated to:

* file containing the model state dict
* shard filenames

So, I propose `resolve_state_dict_file` and adding comments on what all it can be.

src/diffusers/models/modeling_utils.py
```diff
@@ -1257,6 +1357,33 @@ def _get_no_split_modules(self, device_map: str):
...
+        `torch.int64` is passed. So if a non-float `dtype` is passed this functions will throw an exception.
+        """
+        if not dtype.is_floating_point:
+            raise ValueError(
+                f"Can't instantiate {cls.__name__} model under dtype={dtype} since it is not a floating point dtype"
+            )
+
+        logger.info(f"Instantiating {cls.__name__} model under default dtype {dtype}.")
+        dtype_orig = torch.get_default_dtype()
+        torch.set_default_dtype(dtype)
```
sayakpaul, line 1384: Have we fully considered the consequences of this especially under things like "layerwise upcasting"? (see #10347)

src/diffusers/models/modeling_utils.py
```diff
@@ -853,183 +963,98 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
-                        named_buffers=named_buffers,
-                    )
+        with ContextManagers(init_contexts):
+            model = cls.from_config(config, **unused_kwargs)
 
-                    if cls._keys_to_ignore_on_load_unexpected is not None:
-                        for pat in cls._keys_to_ignore_on_load_unexpected:
-                            unexpected_keys = [k for k in unexpected_keys if re.search(pat, k) is None]
+        if dtype_orig is not None:
+            torch.set_default_dtype(dtype_orig)
```
sayakpaul, line 988: `_set_default_torch_dtype()` already calls `set_default_dtype()`, is that still needed here?

src/diffusers/models/modeling_utils.py
```diff
@@ -853,183 +963,98 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
-                    "error_msgs": [],
-                }
-            else:
-                model = cls.from_config(config, **unused_kwargs)
+        if is_sharded:
+            loaded_keys = sharded_metadata["all_checkpoint_keys"]
+        else:
+            loaded_keys = list(state_dict.keys())
+        # TODO: hacky solution
+        loaded_keys = list(model._fix_state_dict_keys_on_load({key: "" for key in loaded_keys}))
```
sayakpaul, line 1001: Let's try not to rely on private methods from PyTorch.

src/diffusers/models/modeling_utils.py
```diff
@@ -1397,56 +1529,7 @@ def recursive_find_attn_block(name, module):
                 state_dict[f"{path}.to_out.0.weight"] = state_dict.pop(f"{path}.proj_attn.weight")
             if f"{path}.proj_attn.bias" in state_dict:
                 state_dict[f"{path}.to_out.0.bias"] = state_dict.pop(f"{path}.proj_attn.bias")
-
-    def _temp_convert_self_to_deprecated_attention_blocks(self) -> None:
```
sayakpaul, line 1401: Why is this going away?

src/diffusers/quantizers/bitsandbytes/bnb_quantizer.py
```diff
@@ -235,18 +235,15 @@ def update_torch_dtype(self, torch_dtype: "torch.dtype") -> "torch.dtype":
...
-    #         device_map = {"": torch.cuda.current_device()}
-    #         logger.info(
-    #             "The device_map was not initialized. "
-    #             "Setting device_map to {'':torch.cuda.current_device()}. "
-    #             "If you want to use the model for inference, please set device_map ='auto' "
-    #         )
-    #     return device_map
+    def update_device_map(self, device_map):
+        if device_map is None:
+            device_map = {"": torch.cuda.current_device()}
```
sayakpaul, line 240: `torch.cuda.current_device()` usually returns the index, which is an integer (example: 0). If we use this condition without care, it can have bad consequences. Like directly, using the `device_map` value which could be just 0. `if 0`, for example.

We could maybe always prefer this format `cuda:{device_index}`?

sayakpaul (commented):
Thanks for starting this! Left some comments from a first pass. 

I think we will need to also add tests for seeing if `device_map` works as expected for quantization. Okay to not test that a bit later once there is consensus about the design changes. Maybe we could add that as a TODO.

Other tests could include checking if we can do `low_cpu_mem_usage=True` along with some changed config values. This will ensure we're well tested for cases like https://github.com/huggingface/diffusers/issues/9343.

sayakpaul:
@SunMarc, 

Additionally, I ran some tests on `audace` (two RTX 4090s). Some tests that are failing (they fail on `main` too):

<details>
<summary>Failures</summary>

```bash
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_load_sharded_checkpoint_from_hub_0_hf_internal_testing_unet2d_sharded_dummy - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argumen...
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_load_sharded_checkpoint_from_hub_1_hf_internal_testing_tiny_sd_unet_sharded_latest_format - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argumen...
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_load_sharded_checkpoint_from_hub_local - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argumen...
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_load_sharded_checkpoint_from_hub_local_subfolder - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argumen...
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_load_sharded_checkpoint_from_hub_subfolder_0_hf_internal_testing_unet2d_sharded_dummy_subfolder - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argumen...
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_load_sharded_checkpoint_from_hub_subfolder_1_hf_internal_testing_tiny_sd_unet_sharded_latest_format_subfolder - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argumen...
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_sharded_checkpoints - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argumen...
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_sharded_checkpoints_device_map - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argumen...
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_sharded_checkpoints_with_variant - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1! (when checking argument...
```

</details>

^^ passes when using with `CUDA_VISIBLE_DEVICES=0` (same with `main`). Expected? 

Same for following:

```bash
FAILED tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelTests::test_sharded_checkpoints_device_map - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1!
```

And then I also ran:

```bash
RUN_SLOW=1 pytest tests/pipelines/stable_diffusion/test_stable_diffusion.py::StableDiffusionPipelineDeviceMapTests
```

Everything passes.

src/diffusers/models/modeling_utils.py
hlky:
```python
import torch
torch.set_default_dtype(torch.float8_e4m3fn)
# TypeError: couldn't find storage object Float8_e4m3fnStorage
```

src/diffusers/models/model_loading_utils.py
```diff
@@ -134,15 +135,14 @@ def _fetch_remapped_cls_from_config(config, old_class):
...
+    map_location: Optional[Union[str, torch.device]] = None,
 ):
     """
     Reads a checkpoint file, returning properly formatted errors if they arise.
     """
-    # TODO: We merge the sharded checkpoints in case we're doing quantization. We can revisit this change
-    # when refactoring the _merge_sharded_checkpoints() method later.
+    # TODO: maybe refactor a bit this part where we pass a dict here
     if isinstance(checkpoint_file, dict):
         return checkpoint_file
```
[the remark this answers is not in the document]
SunMarc: I left this because we might pass a state_dict in this method. This was done in order to make the code simpler.

src/diffusers/models/model_loading_utils.py
```diff
@@ -152,19 +152,35 @@ def load_state_dict(
...
+                metadata = f.metadata()
+                if metadata is not None and metadata.get("format") not in ["pt", "flax"]:
+                    raise OSError(
+                        f"The safetensors archive passed at {checkpoint_file} does not contain the valid metadata. Make sure "
+                        "you save your model with the `save_pretrained` method."
+                    )
             if disable_mmap:
                 return safetensors.torch.load(open(checkpoint_file, "rb").read())
             else:
                 return safetensors.torch.load_file(checkpoint_file, device="cpu")
```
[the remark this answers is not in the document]
SunMarc: Indeed !

src/diffusers/models/model_loading_utils.py
```diff
@@ -188,23 +204,25 @@ def load_state_dict(
 def load_model_dict_into_meta(
     model,
     state_dict: OrderedDict,
-    device: Optional[Union[str, torch.device]] = None,
     dtype: Optional[Union[str, torch.dtype]] = None,
     model_name_or_path: Optional[str] = None,
     hf_quantizer=None,
     keep_in_fp32_modules=None,
-    named_buffers: Optional[Iterator[Tuple[str, torch.Tensor]]] = None,
+    device_map=None,
```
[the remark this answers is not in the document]
SunMarc: Yeah I will add the type annotations, didn't have time for that yet

src/diffusers/models/model_loading_utils.py
SunMarc: This was needed for older version of accelerate but since diffusers requires accelerate v0.31.0, this is not needed anymore

src/diffusers/models/model_loading_utils.py
SunMarc: Removed as we define it a bit earlier now, along with missing keys.

src/diffusers/models/model_loading_utils.py
SunMarc: yeah, it is to replicate the behavior of `load_state_dict` path (just like in transformers) but we can just skip this if you prefer !

src/diffusers/models/model_loading_utils.py
SunMarc: By adding `dispatch_model` at the end, we won't have the issue with buffers anymore. The test that @hlky added is passing this is PR.

src/diffusers/models/modeling_utils.py
SunMarc: Oh yeah this needs to be removed. I just copy pasted this method.

src/diffusers/models/model_loading_utils.py
SunMarc: just tried to remove the indent but it gets reverted

src/diffusers/models/modeling_utils.py
SunMarc: No as in this case, we don't even create the weights with `init_empty_weights`.

src/diffusers/models/modeling_utils.py
SunMarc: These are needed for the single file model changes I made but I can remove these if the changes in the single file model don't make sense or leave it for a follow-up PR !

src/diffusers/models/modeling_utils.py
SunMarc: At the end, `resolved_archive_file` or `resolve_state_dict_file` will be a list of 
- shared filenames (sharded case)
- one filename (not sharded case)
- one state_dict (case where we have already loaded the state_dict) -> I did that in order to make things simpler

Let me know in this case what is the best name !

src/diffusers/models/modeling_utils.py
SunMarc: Thanks for the feedback ! For `torch.float8_e4m3fn` dtype, we can just make an exception for this dtype and skip that.

src/diffusers/models/modeling_utils.py
SunMarc: It is to set back the default dtype to the original dtype `dtype_orig`. This way, if the user continue to create tensors, it will be back to the default dtype they are expected e.g. FP32

src/diffusers/models/modeling_utils.py
SunMarc: This is not a private methods from Pytorch, I defined it ! It is to replace `_convert_deprecated_attention_blocks` method with something a bit more general

src/diffusers/models/modeling_utils.py
SunMarc: This was needed when loading with accelerate load_checkpoint_and_dispatch. Since we don't rely on that, this is not needed !

src/diffusers/models/model_loading_utils.py
```diff
@@ -223,72 +241,93 @@ def load_model_dict_into_meta(
...
-                set_module_tensor_to_device(model, param_name, device, value=param, **set_module_kwargs)
-            else:
-                set_module_tensor_to_device(model, param_name, device, value=param)
+            set_module_tensor_to_device(model, param_name, param_device, value=param, **set_module_kwargs)
 
-    if named_buffers is None:
-        return unexpected_keys
+    return error_msgs, offload_index, state_dict_index
 
-    for param_name, param in named_buffers:
```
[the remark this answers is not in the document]
SunMarc: The changes I did should also cover this use case. The test you added should pass with my PR. The is mainly due to adding the dispatch_model function at the end.

src/diffusers/quantizers/bitsandbytes/bnb_quantizer.py
SunMarc: Yeah I can switch to that !

src/diffusers/models/model_loading_utils.py
```diff
@@ -152,19 +152,35 @@ def load_state_dict(
...
                 with dduf_entries[checkpoint_file].as_mmap() as mm:
                     return safetensors.torch.load(mm)
+            # Check format of the archive
+            with safetensors.safe_open(checkpoint_file, framework="pt") as f:
+                metadata = f.metadata()
+                if metadata is not None and metadata.get("format") not in ["pt", "flax"]:
+                    raise OSError(
+                        f"The safetensors archive passed at {checkpoint_file} does not contain the valid metadata. Make sure "
+                        "you save your model with the `save_pretrained` method."
+                    )
```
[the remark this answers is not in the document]
SunMarc: done !

src/diffusers/models/modeling_utils.py
```diff
@@ -758,89 +850,106 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
-                    model_file = _get_model_file(
+                    except IOError as e:
+                        logger.error(f"An error occurred while trying to fetch {pretrained_model_name_or_path}: {e}")
+                        if not allow_pickle:
+                            raise
+                        logger.warning(
+                            "Defaulting to unsafe serialization. Pass `allow_pickle=False` to raise an error instead."
+                        )
+
+                if resolved_archive_file is None:
```
yiyixuxu, line 949:
```suggestion
                if resolved_archive_file is None and not is_sharded:
```
not sure, but it was like this before and shared checkpoints use `_get_checkpoint_shard_files`, no?

src/diffusers/models/modeling_utils.py
```diff
@@ -853,183 +962,98 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
-                        keep_in_fp32_modules=keep_in_fp32_modules,
-                        named_buffers=named_buffers,
-                    )
+        with ContextManagers(init_contexts):
+            model = cls.from_config(config, **unused_kwargs)
 
-                    if cls._keys_to_ignore_on_load_unexpected is not None:
-                        for pat in cls._keys_to_ignore_on_load_unexpected:
-                            unexpected_keys = [k for k in unexpected_keys if re.search(pat, k) is None]
+        if dtype_orig is not None:
```
yiyixuxu, line 986: based on the function, I think `dtype_orig` won't be `None` - but I guess it won't affect results

```
        dtype_orig = torch.get_default_dtype()
        torch.set_default_dtype(dtype)
        return dtype_orig
```

src/diffusers/models/modeling_utils.py
```diff
@@ -853,183 +962,98 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
+            torch.set_default_dtype(dtype_orig)
 
-                    if len(unexpected_keys) > 0:
-                        logger.warning(
-                            f"Some weights of the model checkpoint were not used when initializing {cls.__name__}: \n {[', '.join(unexpected_keys)]}"
-                        )
+        if not is_sharded and state_dict is None:
```
yiyixuxu, line 989: can we do something like this? 
current code is correct so not a huge deal, just bit easier to follow the logic
```python
if state_dict:
  ...
elif is_sharded:
 ...
else:
  ...
```

src/diffusers/models/modeling_utils.py
yiyixuxu: is there any particular reason we do not run `model._fix_state_dict_keys_on_load` now? is it because we want to extend that to shared checkpoints also?
the shared checkpoint support was added way after these attention_blocks were deprecated, so it is impossible for any shared checkpoint to have this deprecated attention blocks
so maybe we don't need this hacky solution here?

src/diffusers/models/modeling_utils.py
```diff
@@ -1119,54 +1144,135 @@ def _load_pretrained_model(
...
+        error_msgs = []
+
+        # Deal with offload
+        is_safetensors = False
+        if device_map is not None and "disk" in device_map.values():
+            archive_file = (
+                resolved_archive_file[0] if isinstance(resolved_archive_file, (list, tuple)) else resolved_archive_file
+            )
+            is_safetensors = archive_file.endswith(".safetensors")
+            if offload_folder is None and not is_safetensors:
```
yiyixuxu, line 1185: what does this mean?
ok to have `offload_folder = None` for safetensors?

src/diffusers/models/model_loading_utils.py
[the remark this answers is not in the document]
yiyixuxu: ok with this if it makes code simpler :)

src/diffusers/models/modeling_utils.py
SunMarc: Yeah, we can be extra cautious with that but `resolved_archive_file` should be different from `None` after  `_get_checkpoint_shard_files`. Otherwise, it will trigger an error in that function.

src/diffusers/models/modeling_utils.py
SunMarc: It will be `None` only if the user don't specify `torch_dtype` !

src/diffusers/models/modeling_utils.py
SunMarc: This is something I forgot to remove ! Thanks !

SunMarc: With safetensors, you don't have to pass an `offload_folder` but in this PR, I didn't implement this yet. See point in the TODO list

src/diffusers/models/modeling_utils.py
SunMarc: In `_load_pretrained_model`, we do run `model._fix_state_dict_keys_on_load(state_dict)` ! Here, this is just to deal with the case where we have to load the state dict a bit earlier.

SunMarc: > the sharded checkpoint support was added way after these attention_blocks were deprecated, so it is impossible for any sharded checkpoint to have these deprecated attention blocks
> maybe we don't need this hacky solution here?

Then we can just do the fix here in modeling_utils.py. LMK what you think ! 

```python 
if not is_sharded and state_dict is None:
            # Time to load the checkpoint
            state_dict = load_state_dict(
                resolved_archive_file[0], disable_mmap=disable_mmap, dduf_entries=dduf_entries
            )
           -> ###### Put it here ######

if is_sharded:
            loaded_keys = sharded_metadata["all_checkpoint_keys"]
        else:
            loaded_keys = list(state_dict.keys())
        # TODO: hacky solution
        loaded_keys = list(model._fix_state_dict_keys_on_load({key: "" for key in loaded_keys}))
```

src/diffusers/models/modeling_utils.py
```diff
@@ -853,183 +962,98 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
-                    local_files_only=local_files_only,
-                    token=token,
-                    revision=revision,
-                    subfolder=subfolder,
-                    user_agent=user_agent,
-                    commit_hash=commit_hash,
-                    dduf_entries=dduf_entries,
+        # set dtype to instantiate the model under:
+        # 1. If torch_dtype is not None, we use that dtype
+        dtype_orig = None
```
DN6, line 970: There is a use case where we might want to support loading checkpoints that are in mixed precision. e.g The Mochi Video model needs to preserve norms in FP32 (we can't load in FP16/BF16 and then cast back to FP32 with _keep_in_fp32_modules)    
https://huggingface.co/Kijai/Mochi_preview_comfy/blob/main/mochi_preview_dit_fp8_e4m3fn.safetensors

We were thinking of introducing an `auto` dtype for such cases. 

Additionally, torch FP8 is a valid and popular storage type in the Diffusion community that is dynamically upcast during inference time (a feature we will add soon). 
https://github.com/huggingface/diffusers/pull/10347/files

I think this might break if a user tries something like `.from_pretrained(.., torch_dtype=torch.float8_e4m3fn), which would be a breaking change for us.  

Think we need to update the casting here to account for these cases.

SunMarc: > Additionally, torch FP8 is a valid and popular storage type in the Diffusion community that is dynamically upcast during inference time (a feature we will add soon).
> https://github.com/huggingface/diffusers/pull/10347/files
> 
> I think this might break if a user tries something like `.from_pretrained(.., torch_dtype=torch.float8_e4m3fn), which would be a breaking change for us.
> 
> Think we need to update the casting here to account for these cases.

Yes, I will update the code to reflect this.

SunMarc: > There is a use case where we might want to support loading checkpoints that are in mixed precision. e.g The Mochi Video model needs to preserve norms in FP32 (we can't load in FP16/BF16 and then cast back to FP32 with _keep_in_fp32_modules)
> https://huggingface.co/Kijai/Mochi_preview_comfy/blob/main/mochi_preview_dit_fp8_e4m3fn.safetensors

Under `low_cpu_mem_usage = True`, it won't load the the model in FP16/BF16 then cast it back to FP32. With 
_keep_in_fp32_modules, we should be able to make sure that the param stays in FP32. 

Of course, if we have more complicated use case where the params are a mix of many stype then it would make sense to introduce the dtype `auto`, so that we use the dtype of the state dict.

src/diffusers/models/modeling_utils.py
```diff
@@ -1119,54 +1144,130 @@ def _load_pretrained_model(
...
+        if device_map is not None and "disk" in device_map.values():
+            if offload_folder is None:
+                raise ValueError(
+                    "The current `device_map` had weights offloaded to the disk. Please provide an `offload_folder`"
+                    " for them. Alternatively, make sure you have `safetensors` installed if the model you are using"
+                    " offers the weights in this format."
+                )
+            if offload_folder is not None:
+                os.makedirs(offload_folder, exist_ok=True)
+            if offload_state_dict is None:
```
yiyixuxu, line 1188: we defaulted offload_state_dict to be False though https://github.com/huggingface/diffusers/blob/21f94a185602d1e3326d56e173addb3e95c0229f/src/diffusers/models/modeling_utils.py#L693

I think maybe it makes sense to default to `None`, no?

src/diffusers/models/modeling_utils.py
yiyixuxu: but this is loading from the model level though, and should not affect the unify of from_pretrained and `from_single_file` from pipeline level 

Just thought this refactor was done really really elegantly, and would probably have been a lot easier if we didn't have to consider `state_dict` as input.

src/diffusers/models/modeling_utils.py
DN6: We can introduce `dtype="auto"` after this PR is merged. Just wanted to flag

src/diffusers/models/modeling_utils.py
SunMarc: it doesn't look like the layerwise upcasting PR passes torch_dtype = torch.float8_e4m3fn in from_pretrained. cc @a-r-r-o-w . LMK if I should still take care of this case or we can deal with that in a follow-up PR when we need that.

src/diffusers/models/modeling_utils.py
SunMarc: As said [here](https://github.com/huggingface/diffusers/pull/10604#discussion_r1927159867), I don't think we need to change anything yet cc @a-r-r-o-w

src/diffusers/models/modeling_utils.py
SunMarc: I will let you take care of that in a follow-up PR then ! I reverted the changes.

src/diffusers/loaders/single_file_model.py
```diff
@@ -362,17 +362,18 @@ def from_single_file(cls, pretrained_model_link_or_path_or_dict: Optional[str] =
 
         if is_accelerate_available():
             param_device = torch.device(device) if device else torch.device("cpu")
-            named_buffers = model.named_buffers()
-            unexpected_keys = load_model_dict_into_meta(
+            unexpected_keys = [
```
sayakpaul, line 365: Are the single-file related changes to uniformize the use of `load_model_dict_into_meta()` (with the new signature)?

src/diffusers/models/model_loading_utils.py
sayakpaul: I think it might be easier to just remove the vars which are not used at all.

src/diffusers/models/model_loading_utils.py
sayakpaul: Maybe I am missing something but I couldn't spot `dispatch_model()` in https://github.com/huggingface/diffusers/blob/model-loading-refactor/src/diffusers/models/model_loading_utils.py.

sayakpaul: Nevermind, found it: https://github.com/huggingface/diffusers/blob/18d61bb86cb8c8ccca1c26d82cb96c55e8f48f10/src/diffusers/models/modeling_utils.py#L1144

It's a tad bit easier for reviewers if we could just provide these links going forward.

src/diffusers/models/modeling_utils.py
```diff
@@ -61,16 +64,49 @@
...
+
+    def __exit__(self, *args, **kwargs):
+        self.stack.__exit__(*args, **kwargs)
+
+
 logger = logging.get_logger(__name__)
 
 _REGEX_SHARD = re.compile(r"(.*?)-\d{5}-of-\d{5}")
 
+TORCH_INIT_FUNCTIONS = {
```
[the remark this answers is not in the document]
sayakpaul: Something like this could work:

```py
import torch.nn.init as init

init_functions = {
    name: getattr(init, name) for name in dir(init) if callable(getattr(init, name)) 
    and name.endswith("_")
    and not name.startswith("_")
}

print("Available initialization functions:")
for name in init_functions:
    print(name)
```

Prints:

```bash
Available initialization functions:
constant_
dirac_
eye_
kaiming_normal_
kaiming_uniform_
normal_
ones_
orthogonal_
sparse_
trunc_normal_
uniform_
xavier_normal_
xavier_uniform_
zeros_
```

src/diffusers/models/modeling_utils.py
sayakpaul: Could you briefly then elaborate what happens in this codepath?

src/diffusers/models/modeling_utils.py
sayakpaul: If you perhaps run the `test_layerwise_casting_inference` tests to confirm, that would be great.

tests/models/test_modeling_common.py
```diff
@@ -199,12 +199,12 @@ class ModelUtilsTest(unittest.TestCase):
...
 
-    def test_accelerate_loading_error_message(self):
-        with self.assertRaises(ValueError) as error_context:
+    def test_missing_key_loading_warning_message(self):
+        with self.assertLogs("diffusers.models.modeling_utils", level="WARNING") as logs:
             UNet2DConditionModel.from_pretrained("hf-internal-testing/stable-diffusion-broken", subfolder="unet")
 
         # make sure that error message states what keys are missing
-        assert "conv_out.bias" in str(error_context.exception)
+        assert "conv_out.bias" in " ".join(logs.output)
```
sayakpaul, line 207: Explain the changes?

sayakpaul (commented):
Some more comments. 

I am running the 4bit quantization tests currently. And so far things are looking nice! Some tests that might be worth including/consdering:

* Device map with quantization 
* Effectiveness of `keep_modules_in_fp32` when not using quantization.

WDYT?

Edit: 4bit tests are passing. 8bit running.

src/diffusers/loaders/single_file_model.py
SunMarc: yeah that's right !

src/diffusers/models/model_loading_utils.py
SunMarc: i will do that then !

src/diffusers/models/modeling_utils.py
SunMarc: We modify torch initialization function to skip them when loading the model. This makes the loading a lot faster when loading without accelerate. Related to this [PR](https://github.com/huggingface/transformers/pull/27709) in transformers. The difference with transformers is that in transformers, we also have the case where we call an `_init_weights` function to initialize each layer. In diffusers, we rely on the default initialization. 

```python
@contextmanager
def no_init_weights():
    """
    Context manager to globally disable weight initialization to speed up loading large models. To do that, all the
    torch.nn.init function are all replaced with skip.
    """

    def _skip_init(*args, **kwargs):
        pass

    # # Save the original initialization functions
    for name, init_func in TORCH_INIT_FUNCTIONS.items():
        setattr(torch.nn.init, name, _skip_init)
    try:
        yield
    finally:
        # # Restore the original initialization functions
        for name, init_func in TORCH_INIT_FUNCTIONS.items():
            setattr(torch.nn.init, name, init_func)
```

tests/models/test_modeling_common.py
SunMarc: I switched from raising an error to just a warning. See [here](https://github.com/huggingface/diffusers/pull/10604/files#:~:text=%3E%200%3A-,raise%20ValueError(,f%22Cannot%20load%20%7Bcls%7D%20from,-%7Bpretrained_model_name_or_path%7D).

src/diffusers/models/modeling_utils.py
SunMarc: all 37 tests are passing !

src/diffusers/models/modeling_utils.py
SunMarc: Any strong opinion on what to choose @sayakpaul ?

src/diffusers/models/model_loading_utils.py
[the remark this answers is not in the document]
SunMarc: done !

src/diffusers/models/modeling_utils.py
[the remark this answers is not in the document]
SunMarc: WDYT @DN6 ? I'm fine with either. Also there are some missing function since you only choose the one finishing with "_"
```python
    "uniform": nn.init.uniform,
    "normal": nn.init.normal,
    "xavier_uniform": nn.init.xavier_uniform,
    "xavier_normal": nn.init.xavier_normal,
    "kaiming_uniform": nn.init.kaiming_uniform,
    "kaiming_normal": nn.init.kaiming_normal,
```

src/diffusers/models/modeling_utils.py
a-r-r-o-w: @SunMarc Sorry for taking so long to review - had a huge backlog of emails to check :/

Wanted to provide more context for [this](https://github.com/huggingface/diffusers/pull/10604#discussion_r1927159867) comment.

We did not use `torch.float8_e4m3fn` in `from_pretrained()` for layerwise casting because it:
- either errored out somewhere (I don't exactly recall what the problem was), or
- did not handle the case of keeping layers like normalization and modulation (specified in `_skip_layerwise_casting_patterns`) in torch.bfloat16 or torch.float32 precision -- as loading those weights in fp8 leads to poorer generation quality.

Currently, the process is to load transformer on CPU in bf16 and then apply fp8 hooks. This uses more CPU Memory. But, if we could load in fp8 (while keeping certain layers like norms in bf16) and then apply fp8 to bf16 upcasting hooks, it would use lower CPU memory as well.

So, we need to definitely try supporting loading in `torch.float8*` if not already (not a blocker though and can be done in a future PR)

src/diffusers/models/modeling_utils.py
```diff
@@ -1217,54 +1237,129 @@ def _load_pretrained_model(
...
-
         expected_keys = list(model_state_dict.keys())
-
-        original_loaded_keys = loaded_keys
-
         missing_keys = list(set(expected_keys) - set(loaded_keys))
+        if hf_quantizer is not None:
+            missing_keys = hf_quantizer.update_missing_keys(model, missing_keys, prefix="")
         unexpected_keys = list(set(loaded_keys) - set(expected_keys))
+        # # Some models may have keys that are not in the state by design, removing them before needlessly warning
```
a-r-r-o-w, line 1260:
```suggestion
        # Some models may have keys that are not in the state by design, removing them before needlessly warning
```

src/diffusers/models/modeling_utils.py
SunMarc: > I think this might break if a user tries something like `.from_pretrained(.., torch_dtype=torch.float8_e4m3fn), which would be a breaking change for us.

For now, I've reverted the changes for float8. But as @a-r-r-o-w said, it is not really supported to use `torch.float8_e4m3fn` in `from_pretrained`

SunMarc:
> I am running the 4bit quantization tests currently. And so far things are looking nice! Some tests that might be worth including/consdering:
> 
> Device map with quantization
> Effectiveness of keep_modules_in_fp32 when not using quantization.

Done ! Please check

src/diffusers/models/modeling_utils.py
SunMarc: changed to `resolved_model_file`

src/diffusers/models/modeling_utils.py
SunMarc: > Thanks for the feedback ! For torch.float8_e4m3fn dtype, we can just make an exception for this dtype and skip that.

skipped it in the latest commit

src/diffusers/models/modeling_utils.py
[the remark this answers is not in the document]
DN6: Hmm perhaps just clearer to leave this as is then 👍🏽

tests/models/test_modeling_common.py
```diff
@@ -334,6 +334,28 @@ def test_weight_overwrite(self):
...
+        A simple tests to check if the modules under `_keep_in_fp32_modules` are kept in fp32 when we load the model in fp16/bf16
+        Also ensures if inference works.
+        """
+        fp32_modules = SD3Transformer2DModel._keep_in_fp32_modules
+
+        for torch_dtype in [torch.bfloat16, torch.float16]:
+            SD3Transformer2DModel._keep_in_fp32_modules = ["proj_out"]
+
+            model = SD3Transformer2DModel.from_pretrained(
+                "stabilityai/stable-diffusion-3-medium-diffusers", subfolder="transformer", torch_dtype=torch_dtype
```
DN6, line 348: Would it be possible to use the transformer in the tiny version of SD3 here. 
`https://huggingface.co/hf-internal-testing/tiny-sd3-pipe/tree/main/transformer`

DN6 (approved):
LGTM 👍🏽 Thanks @SunMarc

tests/models/test_modeling_common.py
sayakpaul: Yeah let's definitely update this to the tiny checkpoint.

tests/models/test_modeling_common.py
```diff
@@ -334,6 +334,28 @@ def test_weight_overwrite(self):
...
+            )
+
+            for name, module in model.named_modules():
+                if isinstance(module, torch.nn.Linear):
+                    if name in model._keep_in_fp32_modules:
+                        self.assertTrue(module.weight.dtype == torch.float32)
+                    else:
+                        self.assertTrue(module.weight.dtype == torch_dtype)
+        SD3Transformer2DModel._keep_in_fp32_modules = fp32_modules
```
sayakpaul, line 357: Two things:

* Would it make sense to make this test a part of `ModelTesterMixin`? Not strongly opinionated about it. 
* Let's make sure we also perform inference -- this helps us validate the effectiveness of `_keep_modules_in_fp32` even better. WDYT?

tests/quantization/bnb/test_4bit.py
```diff
@@ -136,7 +136,7 @@ def setUp(self):
             bnb_4bit_compute_dtype=torch.float16,
         )
         self.model_4bit = SD3Transformer2DModel.from_pretrained(
-            self.model_name, subfolder="transformer", quantization_config=nf4_config
+            self.model_name, subfolder="transformer", quantization_config=nf4_config, device_map=torch_device
```
sayakpaul, line 139: Would it make sense to test with "auto" device_map instead?

sayakpaul (commented):
Thanks a bunch, @SunMarc! I left some mixed comments across the board. But I don't think any of them are major.

Really great work!

src/diffusers/models/model_loading_utils.py
```diff
@@ -132,17 +134,29 @@ def _fetch_remapped_cls_from_config(config, old_class):
...
+                "you save your model with the `save_pretrained` method."
+            )
+
+
 def load_state_dict(
     checkpoint_file: Union[str, os.PathLike],
-    variant: Optional[str] = None,
     dduf_entries: Optional[Dict[str, DDUFEntry]] = None,
     disable_mmap: bool = False,
+    map_location: Optional[Union[str, torch.device]] = None,
```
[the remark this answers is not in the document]
SunMarc: Indeed, I see that it is also used for safetensors. I will change that

src/diffusers/models/model_loading_utils.py
```diff
@@ -214,81 +238,98 @@ def load_model_dict_into_meta(
...
             ):
                 param = param.to(torch.float32)
-                if accepts_dtype:
-                    set_module_kwargs["dtype"] = torch.float32
+                set_module_kwargs["dtype"] = torch.float32
             else:
                 param = param.to(dtype)
-                if accepts_dtype:
-                    set_module_kwargs["dtype"] = dtype
+                set_module_kwargs["dtype"] = dtype
```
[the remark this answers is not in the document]
SunMarc: not sure what you mean

src/diffusers/models/model_loading_utils.py
```diff
@@ -214,81 +238,98 @@ def load_model_dict_into_meta(
...
+        if device_map is None:
+            param_device = "cpu"
+        else:
+            # find next higher level module that is defined in device_map:
+            # bert.lm_head.weight -> bert.lm_head -> bert -> ''
+            while len(module_name) > 0 and module_name not in device_map:
+                module_name = ".".join(module_name.split(".")[:-1])
+            if module_name == "" and "" not in device_map:
+                raise ValueError(f"{param_name} doesn't have any device set.")
+            param_device = device_map[module_name]
```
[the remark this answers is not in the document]
SunMarc: I can do that !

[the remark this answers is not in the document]
SunMarc: done

src/diffusers/models/model_loading_utils.py
```diff
@@ -214,81 +238,98 @@ def load_model_dict_into_meta(
...
     state_dict = state_dict.copy()
     error_msgs = []
 
     # PyTorch's `_load_from_state_dict` does not copy parameters in a module's descendants
     # so we need to apply the function recursively.
-    def load(module: torch.nn.Module, prefix: str = ""):
-        args = (state_dict, prefix, {}, True, [], [], error_msgs)
+    def load(module: torch.nn.Module, prefix: str = "", assign_to_params_buffers: bool = False):
+        local_metadata = {}
+        local_metadata["assign_to_params_buffers"] = assign_to_params_buffers
```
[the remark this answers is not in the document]
SunMarc: Just checked and indeed I think that we need to guard this as it requires torch >= 2.1

src/diffusers/models/modeling_utils.py
```diff
@@ -159,6 +197,56 @@ def find_tensor_attributes(module: nn.Module) -> List[Tuple[str, Tensor]]:
...
+    Context manager to globally disable weight initialization to speed up loading large models. To do that, all the
+    torch.nn.init function are all replaced with skip.
+    """
+
+    def _skip_init(*args, **kwargs):
+        pass
+
+    # # Save the original initialization functions
+    for name, init_func in TORCH_INIT_FUNCTIONS.items():
+        setattr(torch.nn.init, name, _skip_init)
```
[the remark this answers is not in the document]
SunMarc: indeed, i'll remove that for clarify. The code should be explicit enough

tests/quantization/bnb/test_4bit.py
SunMarc: it would work also but I kept `torch_device` for simplicity as we do in transformers. Do you have multi-gpu runner for quantization tests ? We can create multi-gpu tests if needed.

tests/models/test_modeling_common.py
SunMarc: `Would it make sense to make this test a part of ModelTesterMixin? Not strongly opinionated about it.`

It can make sense to add a test in ModelTesterMixin for the models that do have `_keep_in_fp32_modules` specified. However, for now, none of the models use this arg. Maybe in a follow-up PR when a model actually needs this ?

tests/models/test_modeling_common.py
SunMarc: done !

tests/models/test_modeling_common.py
SunMarc: > Let's make sure we also perform inference -- this helps us validate the effectiveness of _keep_modules_in_fp32 even better. WDYT?

fixed

src/diffusers/loaders/single_file_utils.py
```diff
@@ -1593,18 +1593,9 @@ def create_diffusers_clip_model_from_ldm(
...
-
-    if model._keys_to_ignore_on_load_unexpected is not None:
-        for pat in model._keys_to_ignore_on_load_unexpected:
-            unexpected_keys = [k for k in unexpected_keys if re.search(pat, k) is None]
-
-    if len(unexpected_keys) > 0:
-        logger.warning(
-            f"Some weights of the model checkpoint were not used when initializing {cls.__name__}: \n {[', '.join(unexpected_keys)]}"
-        )
+        model.load_state_dict(diffusers_format_checkpoint, strict=False)
```
sayakpaul, line 1598: If there are any unexpected or incompatible keys, how are we handling that?

src/diffusers/models/model_loading_utils.py
```diff
@@ -132,17 +134,46 @@ def _fetch_remapped_cls_from_config(config, old_class):
...
+        return "cpu"
+    else:
+        module_name = param_name
+        # find next higher level module that is defined in device_map:
+        # bert.lm_head.weight -> bert.lm_head -> bert -> ''
+        while len(module_name) > 0 and module_name not in device_map:
+            module_name = ".".join(module_name.split(".")[:-1])
+        if module_name == "" and "" not in device_map:
+            raise ValueError(f"{param_name} doesn't have any device set.")
+        return device_map[module_name]
```
sayakpaul, line 164: Perfect!

src/diffusers/models/model_loading_utils.py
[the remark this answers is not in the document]
sayakpaul: I meant previously we used to pass `dtype` to `set_module_tensor_to_device()` based on if it was supported in its call().

src/diffusers/models/model_loading_utils.py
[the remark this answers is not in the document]
sayakpaul: @hlky are we cool here?

src/diffusers/models/model_loading_utils.py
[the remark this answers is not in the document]
sayakpaul: Just an info is fine as it's not going to raise an error or cause harm?

src/diffusers/models/transformers/hunyuan_transformer_2d.py
```diff
@@ -280,9 +280,7 @@ def __init__(
             act_fn="silu_fp32",
         )
 
-        self.text_embedding_padding = nn.Parameter(
-            torch.randn(text_len + text_len_t5, cross_attention_dim, dtype=torch.float32)
-        )
+        self.text_embedding_padding = nn.Parameter(torch.randn(text_len + text_len_t5, cross_attention_dim))
```
sayakpaul, line 283: `dtype` is going away?

tests/models/test_modeling_common.py
```diff
@@ -334,6 +334,57 @@ def test_weight_overwrite(self):
...
+        Also ensures if inference works.
+        """
+        fp32_modules = SD3Transformer2DModel._keep_in_fp32_modules
+
+        for torch_dtype in [torch.bfloat16, torch.float16]:
+            SD3Transformer2DModel._keep_in_fp32_modules = ["proj_out"]
+
+            model = SD3Transformer2DModel.from_pretrained(
+                "hf-internal-testing/tiny-sd3-pipe", subfolder="transformer", torch_dtype=torch_dtype
+            ).to("cuda")
```
sayakpaul, line 349:
```suggestion
            ).to(torch_device)
```

tests/models/test_modeling_common.py
```diff
@@ -334,6 +334,57 @@ def test_weight_overwrite(self):
...
+
+            return {
+                "hidden_states": hidden_states,
+                "encoder_hidden_states": encoder_hidden_states,
+                "pooled_projections": pooled_prompt_embeds,
+                "timestep": timestep,
+            }
+
+        # test if inference works.
+        with torch.no_grad() and torch.amp.autocast("cuda", dtype=torch_dtype):
```
sayakpaul, line 378:
```suggestion
        with torch.no_grad() and torch.amp.autocast(torch_device, dtype=torch_dtype):
```

tests/quantization/bnb/test_4bit.py
sayakpaul: Okay let's do that in a separate PR. I can do that. DO you have a reference?

tests/quantization/bnb/test_4bit.py
```diff
@@ -527,6 +529,94 @@ def test_pipeline_cuda_placement_works_with_nf4(self):
...
+        quantized_model = FluxTransformer2DModel.from_pretrained(
+            "hf-internal-testing/tiny-flux-pipe",
+            subfolder="transformer",
+            quantization_config=quantization_config,
+            device_map="auto",
+            torch_dtype=torch.bfloat16,
+        )
+
+        weight = quantized_model.transformer_blocks[0].ff.net[2].weight
+        self.assertTrue(isinstance(weight, bnb.nn.modules.Params4bit))
```
sayakpaul, line 593: Maybe we measure the cosine similarity on the slices instead of a hard assertion like done in the other tests?

tests/quantization/torchao/test_torchao.py
```diff
@@ -301,54 +298,72 @@ def test_device_map(self):
...
+                -0.49609375,
+                0.4375,
+                -0.1640625,
+                -0.66015625,
+                0.43164062,
+            ]
+        )
+        expected_slice_offload = np.array(
+            [0.34375, -0.03515625, 0.0123291, -0.22753906, -0.49414062, 0.4375, -0.16308594, -0.66015625, 0.43554688]
+        )
```
sayakpaul, line 317: Are these changing because of device changes? Cc: @a-r-r-o-w for a double-check.

tests/quantization/torchao/test_torchao.py
```diff
@@ -301,54 +298,72 @@ def test_device_map(self):
         }
         device_maps = ["auto", custom_device_map_dict]
 
-        # inputs = self.get_dummy_tensor_inputs(torch_device)
-        # expected_slice = np.array([0.3457, -0.0366, 0.0105, -0.2275, -0.4941, 0.4395, -0.166, -0.6641, 0.4375])
-
+        inputs = self.get_dummy_tensor_inputs(torch_device)
+        expected_slice_auto = np.array(
```
sayakpaul, line 302: I am also wondering we need two different slices for offload and non-offload.

sayakpaul (approved):
Just some final set of questions and we should be good to go.

src/diffusers/models/model_loading_utils.py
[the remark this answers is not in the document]
SunMarc: I'll add that then

src/diffusers/models/transformers/hunyuan_transformer_2d.py
SunMarc: yeah there was a failing tests but I don't remember the details and also `text_embedding_padding` shouldn't need to have a fixed dtype as text_embedder don't have one either.

src/diffusers/models/model_loading_utils.py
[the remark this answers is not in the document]
hlky: The tests are passing so all good

tests/quantization/bnb/test_4bit.py
SunMarc: Okay let's do that in a separate PR. I can do that. DO you have a reference?

tests/quantization/torchao/test_torchao.py
SunMarc: for torchao, the offloaded modules are not quantized actually. So it means that the two model tested are a bit difference. Hence we get slightly different results

tests/quantization/bnb/test_4bit.py
SunMarc: I can change if needed

src/diffusers/loaders/single_file_utils.py
SunMarc: For unexpected keys, we will just display a warning saying that we didn't load these weights.

tests/quantization/torchao/test_torchao.py
sayakpaul: Okay, let's then add a note about that.

tests/quantization/bnb/test_4bit.py
SunMarc: done

tests/quantization/torchao/test_torchao.py
SunMarc: done

yiyixuxu (approved):
thanks!

sayakpaul:
Failing test is unrelated and are being fixed by @hlky!

PR #10611, opened 2025-01-20
Refactor gradient checkpointing

### LINKED ISSUE #10124: [models] allow specifying block numbers in `enable_gradient_checkpointing()` to speed up training
As brought up in https://github.com/huggingface/diffusers/pull/9982#issuecomment-2517826764 by @bghira, I think we could support this directly in the `enable_gradient_checkpointing()` method we expose for the models. 

Users could specify the block interval they want gradient checkpointing to be applied in and we take care of the rest. The code for this is simple and doesn't require any hacks. 

Gradient checkpointing is a crucial component for training/fine-tuning larger models and this technique allows for nice speed/memory trade-off. 

Cc: @a-r-r-o-w @hlky

a-r-r-o-w:
Fixes #10124.

When finetuning, we currently apply gradient checkpointing to each transformer block. This works wonders for saving memory but can lead to slower throughput. To improve throughput, at the cost of slightly higher memory usage, an acceptable compromise can be made by only checkpointing certain blocks, or by applying a different checkpointing strategy.

This PR will try to refactor how we do gradient checkpointing to enable users to use their own checkpointing functions/strategies. Currently, only LTXVideo has been updated to gather initial feedback on the changes made. If all looks well, will update all the other modeling implementations.

<details>

```python
import gc

import torch
from diffusers import LTXVideoTransformer3DModel


transformer = LTXVideoTransformer3DModel.from_pretrained("a-r-r-o-w/LTX-Video-diffusers", subfolder="transformer", torch_dtype=torch.bfloat16)

batch_size = 1
height = 512
width = 768
num_frames = 81

spatial_compression_ratio = 32
temporal_compression_ratio = 8

latent_height = height // spatial_compression_ratio
latent_width = width // spatial_compression_ratio
num_latent_frames = (num_frames - 1) // temporal_compression_ratio + 1


    t0 = benchmark.Timer(
        stmt="f(*args, **kwargs)",
    )
    torch.cuda.synchronize()
    return t0.blocked_autorange().mean


def clear_memory():
    gc.collect()
    torch.cuda.empty_cache()


def create_inputs(device: torch.device, dtype: torch.dtype):
    max_sequence_length = 128

    hidden_states = torch.randn(batch_size, num_latent_frames * latent_height * latent_width, transformer.config.in_channels, device=device, dtype=dtype)
    encoder_hidden_states = torch.randn(batch_size, max_sequence_length, transformer.config.caption_channels, device=device, dtype=dtype)
    timestep = torch.randint(0, 1000, device=device, dtype=torch.int64, size=(batch_size,))
    encoder_attention_mask = torch.randint(0, 2, device=device, dtype=torch.bool, size=(batch_size, max_sequence_length))

    return {
        "hidden_states": hidden_states,
        "encoder_hidden_states": encoder_hidden_states,
        "timestep": timestep,
        "encoder_attention_mask": encoder_attention_mask,
        "num_frames": num_latent_frames,
        "height": latent_height,
        "width": latent_width,
    }


def forward_and_backward(model, **inputs):
    output = model(**inputs)[0]
    output.mean().backward()


clear_memory()

device = torch.device("cuda")
dtype = torch.bfloat16
num_iterations = 5

transformer.to(device, dtype=dtype)
torch.cuda.synchronize()

clear_memory()
print(f"Model memory: {torch.cuda.max_memory_reserved() / 1024**3:.3f} GB")


# Warmup
inputs = create_inputs(device, dtype)
for _ in range(2):
    forward_and_backward(transformer, **inputs)


# Benchmark no gradient checkpointing
clear_memory()
inputs = create_inputs(device, dtype)
total_time = 0.0
for _ in range(num_iterations):
    time = benchmark_fn(forward_and_backward, transformer, **inputs)
    total_time += time
torch.cuda.synchronize()
print(f"No gradient checkpointing: memory={torch.cuda.max_memory_reserved() / 1024**3:.3f} GB, it/s={num_iterations / total_time:.3f}")


# Benchmark gradient checkpointing
clear_memory()
transformer.enable_gradient_checkpointing()
inputs = create_inputs(device, dtype)
total_time = 0.0
for _ in range(num_iterations):
    time = benchmark_fn(forward_and_backward, transformer, **inputs)
    total_time += time
torch.cuda.synchronize()
print(f"Gradient checkpointing: memory={torch.cuda.max_memory_reserved() / 1024**3:.3f} GB, it/s={num_iterations / total_time:.3f}")


# Benchmark gradient checkpointing with custom layers (every 2nd block)
clear_memory()
for index, layer in enumerate(transformer.transformer_blocks):
    layer.layer_index = index

def gradient_checkpointing_func(model, *args):
    if model.layer_index % 2 == 0:
        return torch.utils.checkpoint.checkpoint(model.__call__, *args, use_reentrant=False)
    return model(*args)

transformer.enable_gradient_checkpointing(gradient_checkpointing_func)
inputs = create_inputs(device, dtype)
total_time = 0.0
for _ in range(num_iterations):
    time = benchmark_fn(forward_and_backward, transformer, **inputs)
    total_time += time
torch.cuda.synchronize()
print(f"Gradient checkpointing (every 2nd block): memory={torch.cuda.max_memory_reserved() / 1024**3:.3f} GB, it/s={num_iterations / total_time:.3f}")


# Benchmark gradient checkpointing with custom layers (every 4th block)
clear_memory()
for index, layer in enumerate(transformer.transformer_blocks):
    layer.layer_index = index

def gradient_checkpointing_func(model, *args):
    if model.layer_index % 4 == 0:
        return torch.utils.checkpoint.checkpoint(model.__call__, *args, use_reentrant=False)
    return model(*args)

transformer.enable_gradient_checkpointing(gradient_checkpointing_func)
inputs = create_inputs(device, dtype)
total_time = 0.0
for _ in range(num_iterations):
    time = benchmark_fn(forward_and_backward, transformer, **inputs)
    total_time += time
torch.cuda.synchronize()
print(f"Gradient checkpointing (every 4th block): memory={torch.cuda.max_memory_reserved() / 1024**3:.3f} GB, it/s={num_iterations / total_time:.3f}")
```
</details>

```
Model memory: 4.025 GB
No gradient checkpointing: memory=26.719 GB, it/s=1.680
Gradient checkpointing: memory=9.316 GB, it/s=1.358
Gradient checkpointing (every 2nd block): memory=17.658 GB, it/s=1.539
Gradient checkpointing (every 4th block): memory=22.188 GB, it/s=1.658
```

Additonal context: https://github.com/huggingface/diffusers/pull/9982#issuecomment-2517826764

cc @bghira

### DIFF
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -27,6 +27,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import safetensors
 import torch
+import torch.utils.checkpoint
 from huggingface_hub import DDUFEntry, create_repo, split_torch_state_dict_into_shards
 from huggingface_hub.utils import validate_hf_hub_args
 from torch import Tensor, nn
@@ -154,6 +155,8 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
     def __init__(self):
         super().__init__()
 
+        self._gradient_checkpointing_func = None
+
     def __getattr__(self, name: str) -> Any:
         """The only reason we overwrite `getattr` here is to gracefully deprecate accessing
         config attributes directly. See https://github.com/huggingface/diffusers/pull/3129 We need to overwrite
@@ -179,14 +182,47 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
         """
         return any(hasattr(m, "gradient_checkpointing") and m.gradient_checkpointing for m in self.modules())
 
-    def enable_gradient_checkpointing(self) -> None:
+    def enable_gradient_checkpointing(
+        self,
+        gradient_checkpointing_func: Optional[Callable] = None,
+        gradient_checkpointing_kwargs: Optional[Dict[str, Any]] = None,
+    ) -> None:
         """
         Activates gradient checkpointing for the current model (may be referred to as *activation checkpointing* or
         *checkpoint activations* in other frameworks).
         """
         if not self._supports_gradient_checkpointing:
-            raise ValueError(f"{self.__class__.__name__} does not support gradient checkpointing.")
-        self.apply(partial(self._set_gradient_checkpointing, value=True))
+            raise ValueError(
+                f"{self.__class__.__name__} does not support gradient checkpointing. Please make sure to set the boolean attribute "
+                f"`_supports_gradient_checkpointing` to `True` in the class definition."
+            )
+
+        user_provided_gradient_checkpointing_func = gradient_checkpointing_func is not None
+        if gradient_checkpointing_func is None:
+
+            def _gradient_checkpointing_func(module, *args):
+                ckpt_kwargs = {"use_reentrant": False} if is_torch_version(">=", "1.11.0") else {}
+                return torch.utils.checkpoint.checkpoint(
+                    module.__call__,
+                    *args,
+                    **ckpt_kwargs,
+                )
+
+            gradient_checkpointing_func = _gradient_checkpointing_func
+
+        if gradient_checkpointing_kwargs is None:
+            gradient_checkpointing_kwargs = {}
+
+            if (
+                not user_provided_gradient_checkpointing_func
+                and is_torch_version(">=", "1.11.0")
+                and inspect.signature(gradient_checkpointing_func).parameters.get("use_reentrant") is not None
+            ):
+                gradient_checkpointing_kwargs["use_reentrant"] = False
+
+        gradient_checkpointing_func = partial(gradient_checkpointing_func, **gradient_checkpointing_kwargs)
+
+        self._set_gradient_checkpointing(enable=True, gradient_checkpointing_func=gradient_checkpointing_func)
 
     def disable_gradient_checkpointing(self) -> None:
         """
@@ -194,7 +230,7 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
         *checkpoint activations* in other frameworks).
         """
         if self._supports_gradient_checkpointing:
-            self.apply(partial(self._set_gradient_checkpointing, value=False))
+            self._set_gradient_checkpointing(enable=False)
 
     def set_use_npu_flash_attention(self, valid: bool) -> None:
         r"""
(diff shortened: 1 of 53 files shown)


bghira:
a thought i have had is that we can invert the condition so that we "dont checkpoint" instead of "do checkpoint" which allows increasing memory use more finely with larger intervals

a-r-r-o-w:
Yep, a number of variations to try out. To not limit how to apply checkpointing, it's best to provide that control to the user, so the idea you mention should be possible to use here too. LMK if you expected something different or want to implement with a better design

yiyixuxu:
are the numbers correct?
```
Model memory: 4.025 GB
No gradient checkpointing: memory=26.664 GB, it/s=0.582
Gradient checkpointing: memory=9.316 GB, it/s=1.339
Gradient checkpointing (every 2nd block): memory=17.658 GB, it/s=1.522
Gradient checkpointing (every 4th block): memory=22.188 GB, it/s=1.630
```
why is no gradient checkpoint has highest memory and lowest throughput?

a-r-r-o-w:
@yiyixuxu They were incorrect, my bad. I thought I was doing the warmup correctly for the first backward call, but it was not correct. Looking at the profiles revealed that the backward pass kernel launch was not warmed up. Updated the example so that it happens now before the "No Gradient Accumulation" benchmark, and we're now seeing the correct numbers

bghira:
yes this looks great, and will be a very useful addition. thank you.

yiyixuxu:
if we just want to allow user to skip certain layers, can we go with a simpler solution?
```python
transformer.enable_gradient_checkpointing(skip_layers= [])
```

then in the code 
```python

        for i, block in enumerate(self.transformer_blocks):
            if torch.is_grad_enabled() and self.gradient_checkpointing and i not in self.skip_layers:
                 ...
```

a-r-r-o-w:
Skipping layers is one use case. The idea is to not limit how the checkpointing is applied and just give the control to the user (with us also providing sensible defaults) - for example, using a different provider for checkpointing, such as deepspeed, instead of monkey patching the forward pass or using other intrusive solution, or allowing custom checkpoint implementations that can perform CPU offloading of stored inputs and retrieve them back when required for recomputation

src/diffusers/models/modeling_utils.py
```diff
@@ -179,22 +182,55 @@ def is_gradient_checkpointing(self) -> bool:
         """
         return any(hasattr(m, "gradient_checkpointing") and m.gradient_checkpointing for m in self.modules())
 
-    def enable_gradient_checkpointing(self) -> None:
+    def enable_gradient_checkpointing(
+        self,
+        gradient_checkpointing_func: Optional[Callable] = None,
+        gradient_checkpointing_kwargs: Optional[Dict[str, Any]] = None,
+    ) -> None:
```
yiyixuxu, line 189: can you give an example when they need to pass both func and kwargs? can't we just accept a  `gradient_checkpointing_func`?

yiyixuxu: ok looks good to me now :)

a-r-r-o-w:
@yiyixuxu LMK if you think the current changes look good and I'll propagate and add tests for all other models to make sure this works as expected

yiyixuxu (approved):
thanks @a-r-r-o-w 
very very nice:)

a-r-r-o-w:
Thanks YiYi! Hopefully this isn't backwards breaking in any way -- it is really nice to be able to customize the grad ckpt function this way imo. Will merge once CI is 🟢

PR #10620, opened 2025-01-21
Inplace sums, remove call to `F.pad`  and better memory count

bm-synth:
- perform in-place sums of `hidden_states` ie doing `a.add_(b)` instead of `a=a+b` to reduce temp memory allocation
- remove call to asymmetric padding in `F.pad` of non-`replicate` pad mode, and instead let padding be done by `Conv3d` for more efficient execution;
- computation of `memory_count` doesn't extend dimensions to allow `torch.compile` to do a better optimisation (?)

### DIFF
--- a/src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
+++ b/src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
@@ -41,9 +41,7 @@ class CogVideoXSafeConv3d(nn.Conv3d):
     """
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
-        memory_count = (
-            (input.shape[0] * input.shape[1] * input.shape[2] * input.shape[3] * input.shape[4]) * 2 / 1024**3
-        )
+        memory_count = torch.prod(torch.tensor(input.shape)) * 2 / 1024**3
 
         # Set to 2GB, suitable for CuDNN
         if memory_count > 2:
--- a/src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
+++ b/src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
@@ -303,7 +303,7 @@ class CogVideoXResnetBlock3D(nn.Module):
         hidden_states, new_conv_cache["conv1"] = self.conv1(hidden_states, conv_cache=conv_cache.get("conv1"))
 
         if temb is not None:
-            hidden_states = hidden_states + self.temb_proj(self.nonlinearity(temb))[:, :, None, None, None]
+            hidden_states.add_(self.temb_proj(self.nonlinearity(temb))[:, :, None, None, None])
 
         if zq is not None:
             hidden_states, new_conv_cache["norm2"] = self.norm2(hidden_states, zq, conv_cache=conv_cache.get("norm2"))
@@ -322,7 +322,7 @@ class CogVideoXResnetBlock3D(nn.Module):
             else:
                 inputs = self.conv_shortcut(inputs)
 
-        hidden_states = hidden_states + inputs
+        hidden_states.add_(inputs)
         return hidden_states, new_conv_cache
 
 
--- a/src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
+++ b/src/diffusers/models/autoencoders/autoencoder_kl_cogvideox.py
@@ -103,6 +103,7 @@ class CogVideoXCausalConv3d(nn.Module):
         self.width_pad = width_pad
         self.time_pad = time_pad
         self.time_causal_padding = (width_pad, width_pad, height_pad, height_pad, time_pad, 0)
+        self.const_padding_conv3d =  (0, self.width_pad, self.height_pad) 
 
         self.temporal_dim = 2
         self.time_kernel_size = time_kernel_size
@@ -115,6 +116,8 @@ class CogVideoXCausalConv3d(nn.Module):
             kernel_size=kernel_size,
             stride=stride,
             dilation=dilation,
+            padding = 0 if self.pad_mode == 'replicate' else self.const_padding_conv3d,
+            padding_mode = 'zeros',
         )
 
     def fake_context_parallel_forward(
@@ -135,9 +138,7 @@ class CogVideoXCausalConv3d(nn.Module):
         if self.pad_mode == "replicate":
             conv_cache = None
         else:
-            padding_2d = (self.width_pad, self.width_pad, self.height_pad, self.height_pad)
             conv_cache = inputs[:, :, -self.time_kernel_size + 1 :].clone()
-            inputs = F.pad(inputs, padding_2d, mode="constant", value=0)
 
         output = self.conv(inputs)
         return output, conv_cache


brunomaga:
Running the following `test_autoencoder.py`

```python
from diffusers.models.autoencoders.autoencoder_kl_cogvideox import CogVideoXCausalConv3d, CogVideoXResnetBlock3D

import time
import torch
import torch.nn as nn


def train(model: nn.Module, video_input: torch.Tensor):
    optimizer = torch.optim.Adam(model.parameters(), lr=0.001)
    loss_fn = torch.nn.MSELoss()
    model.train()
    start_train = time.time()
    for _ in range(100):  # Simulate 100 training iterations
        optimizer.zero_grad()
        output = model(video_input)[0] # returns output, conv_cache
        loss = loss_fn(output*1.1, output)
        loss.backward()
        optimizer.step()
        torch.cuda.synchronize()
    train_time = time.time() - start_train
    print("train_time", train_time, "secs")


def eval(model: nn.Module, video_input: torch.Tensor):
    model.eval()
    start_train = time.time()
    with torch.no_grad():
        for _ in range(300):  # Simulate 300 inference iterations
            output = model(video_input)
            torch.cuda.synchronize()
    eval_time = time.time() - start_train
    print("eval_time", eval_time, "secs")




if __name__=='__main__':
    device='cuda'
    video_input = torch.randn([1, 128, 8, 544, 960], device=device, dtype=torch.bfloat16) # Batch Channels Frames Height Width
    memory_count = torch.prod(torch.tensor(video_input.shape)) / 1024**3
    print("input size: ", memory_count.item(), "GBs")
    model = CogVideoXCausalConv3d(in_channels=video_input.shape[1], out_channels=512, kernel_size=3)
    model = model.to(device=video_input.device, dtype=video_input.dtype)
    eval(model, video_input)
    train(model, video_input)
    print("Max memory", torch.cuda.max_memory_allocated()/ 1024**3, "GBs")
```

calling the main branch gives:

```shell
$ PYTHONPATH=./diffusers_main/src/ python test_autoencoder.py
input size:  0.498046875 GBs
eval_time 33.06385564804077 secs
train_time 34.33984375 secs
Max memory 22.18018913269043 GBs
```

calling this PR branch gives:

```shell
$ PYTHONPATH=./diffusers_PR/src/ python test_autoencoder.py
input size:  0.498046875 GBs
eval_time 31.588099241256714 secs
train_time 34.1251916885376 secs
Max memory 22.17398452758789 GBs
```

I'll try to test more dimensions.

hlky:
@bm-synth Great, thanks. Would it also be possible to verify numerical accuracy between the two versions? For a change like this we would expect between 0 to 1e-6 difference.

brunomaga:
@hlky I updated the code above to fix a seed (`torch.manual_seed(42)`) and output the output tensor after 100 training iterations. Then I ran this to compare both `output_*.pt` files:

```python
if __name__=='__main__':
    output_main: torch.Tensor = torch.load("output_main.pt")
    output_PR: torch.Tensor = torch.load("output_PR.pt")
    print(output_main.mean().item(), "vs", output_PR.mean().item())
    print(output_main.std().item(), "vs", output_PR.std().item())
    print("max abs diff", (output_PR-output_main).diff().abs().max().item())
    assert torch.allclose(output_main, output_PR)
```

output:
```
mean: 1.7345882952213287e-08 vs 1.7345882952213287e-08
std: 0.0020904541015625 vs 0.0020904541015625
max abs diff: 0.0
```

bm-synth:
@hlky ping?

hlky:
Hi @bm-synth. We need to verify the accuracy of `CogVideoXCausalConv3d`, not the output from your trained model.

<details><summary>Code</summary>
<p>

```python
from typing import Optional, Tuple, Union

import torch
import torch.nn as nn
from diffusers.models.autoencoders.autoencoder_kl_cogvideox import CogVideoXCausalConv3d


class CogVideoXSafeConv3d_PR(nn.Conv3d):
    r"""
    A 3D convolution layer that splits the input tensor into smaller parts to avoid OOM in CogVideoX Model.
    """

    def forward(self, input: torch.Tensor) -> torch.Tensor:
        memory_count = torch.prod(torch.tensor(input.shape)) * 2 / 1024**3

        # Set to 2GB, suitable for CuDNN
        if memory_count > 2:
            kernel_size = self.kernel_size[0]
            part_num = int(memory_count / 2) + 1
            input_chunks = torch.chunk(input, part_num, dim=2)

            if kernel_size > 1:
                input_chunks = [input_chunks[0]] + [
                    torch.cat((input_chunks[i - 1][:, :, -kernel_size + 1 :], input_chunks[i]), dim=2)
                    for i in range(1, len(input_chunks))
                ]

            output_chunks = []
            for input_chunk in input_chunks:
                output_chunks.append(super().forward(input_chunk))
            output = torch.cat(output_chunks, dim=2)
            return output
        else:
            return super().forward(input)


class CogVideoXCausalConv3d_PR(nn.Module):
    r"""A 3D causal convolution layer that pads the input tensor to ensure causality in CogVideoX Model.

    Args:
        in_channels (`int`): Number of channels in the input tensor.
        out_channels (`int`): Number of output channels produced by the convolution.
        kernel_size (`int` or `Tuple[int, int, int]`): Kernel size of the convolutional kernel.
        stride (`int`, defaults to `1`): Stride of the convolution.
        dilation (`int`, defaults to `1`): Dilation rate of the convolution.
        pad_mode (`str`, defaults to `"constant"`): Padding mode.
    """

    def __init__(
        self,
        in_channels: int,
        out_channels: int,
        kernel_size: Union[int, Tuple[int, int, int]],
        stride: int = 1,
        dilation: int = 1,
        pad_mode: str = "constant",
    ):
        super().__init__()

        if isinstance(kernel_size, int):
            kernel_size = (kernel_size,) * 3

        time_kernel_size, height_kernel_size, width_kernel_size = kernel_size

        # TODO(aryan): configure calculation based on stride and dilation in the future.
        # Since CogVideoX does not use it, it is currently tailored to "just work" with Mochi
        time_pad = time_kernel_size - 1
        height_pad = (height_kernel_size - 1) // 2
        width_pad = (width_kernel_size - 1) // 2

        self.pad_mode = pad_mode
        self.height_pad = height_pad
        self.width_pad = width_pad
        self.time_pad = time_pad
        self.time_causal_padding = (width_pad, width_pad, height_pad, height_pad, time_pad, 0)
        self.const_padding_conv3d = (0, self.width_pad, self.height_pad)

        self.temporal_dim = 2
        self.time_kernel_size = time_kernel_size

        stride = stride if isinstance(stride, tuple) else (stride, 1, 1)
        dilation = (dilation, 1, 1)
        self.conv = CogVideoXSafeConv3d_PR(
            in_channels=in_channels,
            out_channels=out_channels,
            kernel_size=kernel_size,
            stride=stride,
            dilation=dilation,
            padding=0 if self.pad_mode == "replicate" else self.const_padding_conv3d,
            padding_mode="zeros",
        )

    def fake_context_parallel_forward(
        self, inputs: torch.Tensor, conv_cache: Optional[torch.Tensor] = None
    ) -> torch.Tensor:
        if self.pad_mode == "replicate":
            inputs = F.pad(inputs, self.time_causal_padding, mode="replicate")
        else:
            kernel_size = self.time_kernel_size
            if kernel_size > 1:
                cached_inputs = [conv_cache] if conv_cache is not None else [inputs[:, :, :1]] * (kernel_size - 1)
                inputs = torch.cat(cached_inputs + [inputs], dim=2)
        return inputs

    def forward(self, inputs: torch.Tensor, conv_cache: Optional[torch.Tensor] = None) -> torch.Tensor:
        inputs = self.fake_context_parallel_forward(inputs, conv_cache)

        if self.pad_mode == "replicate":
            conv_cache = None
        else:
            conv_cache = inputs[:, :, -self.time_kernel_size + 1 :].clone()

        output = self.conv(inputs)
        return output, conv_cache

model = CogVideoXCausalConv3d(in_channels=128, out_channels=512, kernel_size=3).eval()
with torch.no_grad():
    output = model(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]

with torch.no_grad():
    output_2 = model(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]

torch.testing.assert_close(output, output_2)

print((output - output_2).abs().max())

model_pr = CogVideoXCausalConv3d_PR(in_channels=128, out_channels=512, kernel_size=3).eval()
with torch.no_grad():
    output_pr = model_pr(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]
```

</p>
</details> 

```python
torch.testing.assert_close(output, output_pr)
Mismatched elements: 2139073042 / 2139095040 (100.0%)
Greatest absolute difference: 5.3313703536987305 at index (0, 421, 5, 286, 946) (up to 1e-05 allowed)
Greatest relative difference: 2893981952.0 at index (0, 348, 3, 142, 869) (up to 1.3e-06 allowed)
```

```python
print((output - output_pr).abs().max())
tensor(5.3314)
```

The first check here `torch.testing.assert_close(output, output_2)` shows that `CogVideoXCausalConv3d` is deterministic so we would only accept up to around 1e-6 difference but preferably less or no change.

Also note the code will run on CPU in float32, this is to avoid other source of non-determinism when testing although it will use a large amount of memory. Generally we choose the smallest possible shape and model configuration for tests.

bm-synth:
@hlky your code has an issue. the random state before initializing each `CogVideoXCausalConv3D` is different. Try to call `torch.manual_seed(42)` before each initialization, i.e.:

```python
if __name__=='__main__':

    print(f"diffusers version: {diffusers.__version__}")

    torch.manual_seed(42)  #### <---- added this
    model = CogVideoXCausalConv3d(in_channels=128, out_channels=512, kernel_size=3).eval()
    with torch.no_grad():
        output = model(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]

    with torch.no_grad():
        output_2 = model(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]

    torch.testing.assert_close(output, output_2)
    print("max abs difference (output, output_2):", (output - output_2).abs().max().item())
    print("number of different elements (output, output_2):", (output != output_2).sum().item())

    torch.manual_seed(42) ##### <---- added this
    model_pr = CogVideoXCausalConv3d_PR(in_channels=128, out_channels=512, kernel_size=3).eval()
    with torch.no_grad():
        output_pr = model_pr(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]
        
    torch.testing.assert_close(output, output_pr)
    print("max abs difference (output, output_pr):", (output - output_pr).abs().max().item())
    print("number of different elements (output, output_pr):", (output != output_pr).sum().item())
```  

output:

```
diffusers version: 0.32.2
abs difference (output, output_2): 0.0
number of different elements (output, output_2): 0
abs difference (output, output_pr): 0.0
number of different elements (output, output_pr): 0
```

yiyixuxu (approved):
thanks @bm-synth 
the refactor looks really nice. 

let's merge once @hlky confirm the results are identical

hlky:
I'm still concerned by the reproducibility, especially as it's with float32 on CPU. Using `torch.manual_seed` isn't something we normally require users to do yet with this change they would need to use `torch.manual_seed` to get the same result from VAE decode in addition to passing a `torch.Generator` to the pipeline to get the same generated latent, and the issue would be compounded by the expected sources of non-determinism i.e. CUDA.

bm-synth:
@hlky (cc @yiyixuxu )  your reproducibility concerns are not related to this PR. They are an issue in the production code. This PR gives the same behaviour as production. If you run the code below, where you only call your current CogVideoX module twice, without setting the random seed, it already fails. See below:

```python
if __name__=='__main__':

    print(f"diffusers version: {diffusers.__version__}")

    torch.manual_seed(42)
    model = CogVideoXCausalConv3d(in_channels=128, out_channels=512, kernel_size=3).eval()
    with torch.no_grad():
        output = model(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]

    with torch.no_grad():
        output_2 = model(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]

    torch.testing.assert_close(output, output_2)
    print("max abs difference (output, output_2):", (output - output_2).abs().max().item())
    print("number of different elements (output, output_2):", (output != output_2).sum().item())

    # torch.manual_seed(42)  ##### <--- THE SECOND CHECK ONLY PASSES IF YOU UNCOMMENT THIS!  
    model_pr = CogVideoXCausalConv3d(in_channels=128, out_channels=512, kernel_size=3).eval()
    with torch.no_grad():
        output_pr = model_pr(torch.randn([1, 128, 8, 544, 960], generator=torch.Generator().manual_seed(0)))[0]
        
    torch.testing.assert_close(output, output_pr)
    print("max abs difference (output, output_pr):", (output - output_pr).abs().max().item())
    print("number of different elements (output, output_pr):", (output != output_pr).sum().item())
```

hlky:
We get the same `output_2` as `output` with the same input (determined by `torch.Generator().manual_seed(0)`). We should expect the same output from PR version with the same input.

hlky:
Can you run more benchmarks? From https://github.com/huggingface/diffusers/pull/10620#issuecomment-2613100934 the benefit is not clear and it looks like the figures for memory only cover training. Follow our standard benchmarking methodology 

https://github.com/huggingface/diffusers/blob/1357931d74e5ec5b187ddaa1da118672dd004f21/benchmarks/utils.py#L52-L58
https://github.com/huggingface/diffusers/tree/main/benchmarks

https://github.com/huggingface/diffusers/pull/10620#issue-2801635501 also mentions improvement in torch.compile, can we demonstrate this? what specifically is the optimization it allows in torch.compile?

bm-synth:
> Can you run more benchmarks?

@hlky running this code that calls your `benchmark_fn` on 5 different shapes (feel free to change it):

```python
    t0 = benchmark.Timer( 
         stmt="f(*args, **kwargs)", 
    ) 
    return f"{(t0.blocked_autorange().mean):.5f}"

def train(model, inputs, targets, criterion, optimizer):
    model.train()
    optimizer.zero_grad()
    outputs = model(inputs)
    if isinstance(outputs, tuple):
        outputs = outputs[0] # only use the first tensor as output
    loss = criterion(outputs, targets)
    loss.backward()
    optimizer.step()
    return loss.item()

def inference(model, inputs):
    model.eval()
    with torch.no_grad():
        outputs = model(inputs)
    return outputs

if __name__=='__main__':
    print(f"diffusers version: {diffusers.__version__}")
    to_kw = dict(device="cuda", dtype=torch.bfloat16)

    torch.manual_seed(42)
    model_main = CogVideoXCausalConv3d(in_channels=128, out_channels=512, kernel_size=3).to(**to_kw)
    model_main = torch.compile(model_main, mode="max-autotune-no-cudagraphs")
    optimizer_main = torch.optim.Adam(model_main.parameters(), lr=1e-3)

    torch.manual_seed(42)
    model_pr = CogVideoXCausalConv3d(in_channels=128, out_channels=512, kernel_size=3).to(**to_kw)
    model_pr = torch.compile(model_pr, mode="max-autotune-no-cudagraphs")
    optimizer_pr = torch.optim.Adam(model_pr.parameters(), lr=1e-3)

    criterion = nn.MSELoss()

    # test 5 very different shape inputs
    for shape in [
        (1, 128, 8, 544, 960),
        (3, 128, 22, 12, 123),
        (5, 128, 2, 4, 324),
        (2, 128, 32, 128, 128),
        (9, 128, 64, 24, 123),
        ]:

        input = torch.randn(shape, generator=torch.Generator().manual_seed(0)).to(**to_kw)
        output_shape = ( shape[0], 512, shape[2], shape[3], shape[4] )
        target = torch.randn(output_shape, generator=torch.Generator().manual_seed(0)).to(**to_kw)
        
        # few iterations to warm up the model and compiler
        for _ in range(5):
            benchmark_fn( train, model_main, input, target, criterion, optimizer_main)
            benchmark_fn( train, model_pr, input, target, criterion, optimizer_pr)

        # branchmark training
        runtime = benchmark_fn( train, model_main, input, target, criterion, optimizer_main)
        print(f"shape {shape}, main runtime:", runtime)   

        runtime = benchmark_fn( train, model_pr, input, target, criterion, optimizer_pr)
        print(f"shape {shape}, PR runtime:", runtime)
        print()
```

output:

```
diffusers version: 0.32.2

shape (1, 128, 8, 544, 960), main runtime: 0.12767
shape (1, 128, 8, 544, 960), PR runtime: 0.12665

shape (3, 128, 22, 12, 123), main runtime: 0.00345
shape (3, 128, 22, 12, 123), PR runtime: 0.00338

shape (5, 128, 2, 4, 324), main runtime: 0.00104
shape (5, 128, 2, 4, 324), PR runtime: 0.00103

shape (2, 128, 32, 128, 128), main runtime: 0.03298
shape (2, 128, 32, 128, 128), PR runtime: 0.03290

shape (9, 128, 64, 24, 123), main runtime: 0.05253
shape (9, 128, 64, 24, 123), PR runtime: 0.05283
```

so you're just avoiding the extra 

```python
inputs = F.pad(inputs, padding_2d, mode="constant", value=0)
```

and having the Conv3d cuda kernel handilng that instead. Plus the cleaner code.

bm-synth:
> https://github.com/huggingface/diffusers/pull/10620#issue-2801635501 also mentions improvement in torch.compile, can we demonstrate this? what specifically is the optimization it allows in torch.compile?

i'll cc here my colleague @ic-synth that first flagged that and let him answer with more accuracy. Either way, it's a cleaner way to write a product over all shape dimensions.

hlky:
@bm-synth Can you benchmark inference? Training this autoencoder is not going to be a common use case. Timing is also worse in some cases and very minimal in others to the point it appears like random variation, the biggest difference is less than 1%. We can also include memory in benchmarking :)

bm-synth:
@hlky  cc @yiyixuxu here's the updated benchmark scritpt that includes inference:

```python

def train(model, inputs, targets, criterion, optimizer):
    model.train()
    optimizer.zero_grad()
    outputs = model(inputs)
    if isinstance(outputs, tuple):
        outputs = outputs[0] # only use the first tensor as output
    loss = criterion(outputs, targets)
    loss.backward()
    optimizer.step()
    return loss.item()

def inference(model, inputs):
    model.eval()
    with torch.no_grad():
        for _ in range(50):
            outputs = model(inputs)
    return outputs

if __name__=='__main__':

    print(f"diffusers version: {diffusers.__version__}")
    to_kw = dict(device="cuda", dtype=torch.bfloat16)

    torch.manual_seed(42)
    model_main = CogVideoXCausalConv3d(in_channels=128, out_channels=512, kernel_size=3).to(**to_kw)
    model_main = torch.compile(model_main, mode="max-autotune-no-cudagraphs")
    optimizer_main = torch.optim.Adam(model_main.parameters(), lr=1e-3)

    torch.manual_seed(42)
    model_pr = CogVideoXCausalConv3d_PR(in_channels=128, out_channels=512, kernel_size=3).to(**to_kw)
    model_pr = torch.compile(model_pr, mode="max-autotune-no-cudagraphs")
    optimizer_pr = torch.optim.Adam(model_pr.parameters(), lr=1e-3)

    criterion = nn.MSELoss()

    # test 5 very different shape inputs
    for shape in (
        (16, 128, 64, 64, 64),
        (8, 128, 128, 32, 32),
        (4, 128, 256, 128, 128),
        (4, 128, 512, 256, 256),
        (2, 128, 1024, 512, 512),
    ):

        input = torch.randn(shape, generator=torch.Generator().manual_seed(0)).to(**to_kw)
        output_shape = ( shape[0], 512, shape[2], shape[3], shape[4] )
        target = torch.randn(output_shape, generator=torch.Generator().manual_seed(0)).to(**to_kw)
        
        # few iterations of training to warm up the model and compiler
        model_main.train()
        model_pr.train()
        for _ in range(5):
            train(model_main, input, target, criterion, optimizer_main)
            train(model_pr, input, target, criterion, optimizer_pr)

        # benchmark training
        runtime_main = benchmark_fn( train, model_main, input, target, criterion, optimizer_main)
        runtime_pr = benchmark_fn( train, model_pr, input, target, criterion, optimizer_pr)
        print(f"shape {shape} train: main runtime {runtime_main} vs PR runtime {runtime_pr}")

        # few iterations of inference to warm up the model and compiler
        torch.compiler.reset()
        model_main.eval()
        model_pr.eval()
        with torch.inference_mode():
            for _ in range(5):
                # few iterations of inference to warm up the model and compiler
                inference(model_main, input)
                inference(model_pr, input)
            runtime_main = benchmark_fn( inference, model_main, input)
            runtime_pr = benchmark_fn( inference, model_pr, input)
            print(f"shape {shape} inference: main runtime {runtime_main} vs PR runtime {runtime_pr}")

        print()
```

to measure memory, i added to the `forward` of `CogVideoXCausalConv3d`:

```python
        print(f"MAIN Memory BEFORE conv:", torch.cuda.memory_allocated())
        output = self.conv(inputs)
        print(f"MAIN Memory AFTER conv:", torch.cuda.memory_allocated())
```

to both PR and main branch. **Memory usage for first shape**:

`main`, train:
- MAIN Memory BEFORE conv: 6595022848
- MAIN Memory AFTER conv: 10889990144

PR, train:
- PR Memory BEFORE conv: 6523719680
- PR Memory AFTER conv: 10818686976

`main`, inference:
- MAIN Memory BEFORE conv: 10916464640
- MAIN Memory AFTER conv: 15211431936

PR, inference:
- PR Memory BEFORE conv: 10845161472
- PR Memory AFTER conv: 15140128768

**Runtimes for first 2 shapes** (NVIDIA H200):

- shape (16, 128, 64, 64, 64) train: main runtime 0.13034 vs PR runtime 0.07435
- shape (16, 128, 64, 64, 64) inference: main runtime 1.49722 vs PR runtime 1.46837
- shape (8, 128, 128, 32, 32) train: main runtime 0.03417 vs PR runtime 0.01869
- shape (8, 128, 128, 32, 32) inference: main runtime 0.38165 vs PR runtime 0.36572

Feel free to run it on your side and correct this benchmark script if you find any issue.

ic-synth:
@bm-synth  i think there's been some misunderstanding: doing `torch.prod(torch.tensor(input.shape)) * 2 / 1024**3` is worse for compilation that writing down the product explicitly. i'd revert that change.

bm-synth:
thank you @hlky here are the new results with the updated branch (NVIDIA H200):

Script that measures runtimes for 2 shapes:

```
shape (16, 128, 64, 64, 64) train: main runtime 0.12332 vs PR runtime 0.07097
shape (16, 128, 64, 64, 64) inference: main runtime 1.48820 vs PR runtime 1.46182
shape (8, 128, 128, 32, 32) train: main runtime 0.03254 vs PR runtime 0.01825
shape (8, 128, 128, 32, 32) inference: main runtime 0.37884 vs PR runtime 0.36265
```

the other test for the comparison of numerical difference:

```
max abs difference (output, output_2): 0.0
number of different elements (output, output_2): 0
max abs difference (output, output_pr): 0.0
number of different elements (output, output_pr): 0
```

PR #10622, opened 2025-01-21
[docs] uv installation

stevhliu:
Proposes adding uv install instructions (see Slack [thread](https://huggingface.slack.com/archives/C01N44FJDHT/p1737377177709279) for more context) if you're also interested!

### DIFF
--- a/docs/source/en/installation.md
+++ b/docs/source/en/installation.md
@@ -23,12 +23,26 @@ You should install 🤗 Diffusers in a [virtual environment](https://docs.python
 If you're unfamiliar with Python virtual environments, take a look at this [guide](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/).
 A virtual environment makes it easier to manage different projects and avoid compatibility issues between dependencies.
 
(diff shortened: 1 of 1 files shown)


docs/source/en/installation.md
```diff
@@ -23,32 +23,46 @@ You should install 🤗 Diffusers in a [virtual environment](https://docs.python
 If you're unfamiliar with Python virtual environments, take a look at this [guide](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/).
 A virtual environment makes it easier to manage different projects and avoid compatibility issues between dependencies.
 
-Start by creating a virtual environment in your project directory:
+Create a virtual environment with [uv](https://docs.astral.sh/uv/) (refer to [Installation](https://docs.astral.sh/uv/getting-started/installation/) for installation instructions), a fast Rust-based Python package and project manager.
```
sayakpaul, line 26: There are still some cases where managing virtual envs with `uv` might be problematic. We could maybe keep two options `venv` and `uv`?

yiyixuxu (approved):
thanks @stevhliu!

Issue #10640, opened 2025-01-24
FluxInpaintPipeline error when im using DataParallel to use multiple gpus


Himasnhu-AT:
### Describe the bug

line 175, in forward
    for t in chain(self.module.parameters(), self.module.buffers()):
                   ^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/tensorflow/lib/python3.12/site-packages/diffusers/configuration_utils.py", line 143, in __getattr__
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'FluxFillPipeline' object has no attribute 'parameters'

### Reproduction

if torch.cuda.device_count() > 1:
     pipe = torch.nn.DataParallel(pipe)
pipe = pipe.to("cuda")

### Logs

```shell

```

### System Info

Ubuntu AWS deep learning AMI

g5.12xlarge
4A10 GPUS

a-r-r-o-w:
DP needs to be applied on torch modules. The pipelines are not torch modules, so it is expected to not work. Instead, you can initialise the modules separately, wrap with DP, then create a per-device pipeline by initialising with the modules.

Himasnhu-AT:
okay, is there a way to make it part of the pipeline itself as it'll be easier for other to use it too

a-r-r-o-w:
> okay, is there a way to make it part of the pipeline itself as it'll be easier for other to use it too

Not really. Data parallel should be applied per torch module, not on the container holding the modules. Pipeline objects should be created per process to avoid this error. The simple way of doing distributed inference with diffusers would look something like:

```python
import torch
import torch.distributed as dist
from diffusers import FluxPipeline

dist.init_process_group("nccl")
world_size = dist.get_world_size()
rank = dist.get_rank()

torch.cuda.set_device(rank)

pipe = FluxPipeline.from_pretrained(...).to("cuda")

with open("prompts.txt", "r") as file:
    prompts = [line.strip() for line in file.readlines()]

num_chunks = len(prompts) // world_size
prompts = prompts[rank * num_chunks : (rank + 1) * num_chunks]

for i, prompt in enumerate(prompts):
    output = pipe(prompt, ...).images[0]  # or, do batched inference per rank
    output.save(f"{rank}-{i}-{escape_special_characters(prompt)}.png")

dist.destroy_process_group()
```

Then run the code using:

```shell
torchrun --nodes 1 --nproc_per_node <NUM_GPUS> test.py
```

[...]

PR #10641, opened 2025-01-24
sample solution #10640

Himasnhu-AT:
#10640 

This pull request includes a new method to the `FluxPipeline` class in the `src/diffusers/pipelines/flux/pipeline_flux_fill.py` file. The addition of the `parameters` method allows for load distribution using `torch.nn.parallel.dataparallel`

* [`src/diffusers/pipelines/flux/pipeline_flux_fill.py`](diffhunk://#diff-d32149c583e73b53b57c6630e37207449699cb760944d0e64199424c4e4c890aR969-R971): Added a `parameters` method to return a list of parameters from `transformer`, `vae`, `text_encoder`, and `text_encoder_2`.

> [!IMPORTANT]
> Wasn't able to test using testcases because im almost out of mem on my PC

USECASE:
```py
pipe = FluxFillPipeline( ... )
pipe = torch.nn.DataParallel(pipe)
```

### DIFF
--- a/src/diffusers/pipelines/flux/pipeline_flux_fill.py
+++ b/src/diffusers/pipelines/flux/pipeline_flux_fill.py
@@ -966,3 +966,6 @@ class FluxFillPipeline(
             return (image,)
 
         return FluxPipelineOutput(images=image)
+
+    def parameters(self):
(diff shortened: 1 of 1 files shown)


a-r-r-o-w:
Closing as not planned since we want pipelines to eventually just be a container object without tying in information of models too closely. Explanation for how to do data parallel is [here](https://github.com/huggingface/diffusers/issues/10640#issuecomment-2661490718).

PR #10642, opened 2025-01-24
Add support for lumina2

zhuole1025:
# What does this PR do?

This PR will add the official Lumina-Image 2 to the diffusers. Lumina-Image 2.0 is the latest model in the Lumina family and will be released very soon (https://huggingface.co/Alpha-VLLM/Lumina-Image-2.0). It is a 2B parameter Diffusion Transformer that significantly improves instruction-following and generates higher-quality, more diverse images. Our paper will be released soon, and we have finished the diffuser pipeline for Lumina-Image 2.0.

Core library:

- Schedulers: @yiyixuxu


src/diffusers/models/normalization.py
```diff
@@ -211,25 +211,31 @@ class LuminaRMSNormZero(nn.Module):
         embedding_dim (`int`): The size of each embedding vector.
     """
 
-    def __init__(self, embedding_dim: int, norm_eps: float, norm_elementwise_affine: bool):
+    def __init__(self, embedding_dim: int, norm_eps: float, modulation: bool):
```
yiyixuxu, line 214: this seems like a breaking change, no? did we test the lumina model that currently using it?

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,415 @@
...
+            inner_dim=4 * dim,
+            multiple_of=multiple_of,
+            ffn_dim_multiplier=ffn_dim_multiplier,
+        )
+        
+        self.norm1 = LuminaRMSNormZero(
+            embedding_dim=dim,
+            norm_eps=norm_eps,
+            modulation=modulation,
+        )
```
yiyixuxu, line 92: instead of changing the LuminaRMSNormZero, maybe we can do this?
```suggestion
        self.norm1 = LuminaRMSNormZero() if modulation else RMSNorm(...)
```

src/diffusers/models/normalization.py
zhuole1025: oh, this is indeed a breaking change. Although norm_elementwise_affine is set to True in default so this parameter is cumbersome, but modulation should have a default value of False

src/diffusers/models/transformers/transformer_lumina2.py
zhuole1025: Thanks, this will solve the break

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,418 @@
...
+        img_freqs_cis_shape[1] = max_img_len
+        img_freqs_cis = torch.zeros(*img_freqs_cis_shape, device=device, dtype=freqs_cis.dtype)
+        
+        for i in range(bsz):
+            cap_len = l_effective_cap_len[i]
+            img_len = l_effective_img_len[i]
+            cap_freqs_cis[i, :cap_len] = freqs_cis[i, :cap_len]
+            img_freqs_cis[i, :img_len] = freqs_cis[i, cap_len:cap_len+img_len]
+        
+        for layer in self.context_refiner:
```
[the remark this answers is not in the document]
zhuole1025: Thanks for the suggestion! I have moved refiner into forward() for better readbility. As for the freqs_cis, we already have a class Lumina2PosEmbed.

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,432 @@
...
+        for i in range(bsz):
+            cap_len = l_effective_cap_len[i]
+            img_len = l_effective_img_len[i]
+            mask[i, :cap_len+img_len] = True
+            padded_full_embed[i, :cap_len] = cap_feats[i, :cap_len]
+            padded_full_embed[i, cap_len:cap_len+img_len] = padded_img_embed[i, :img_len]
+            
+        return padded_full_embed, mask
+
+    def forward(
```
yiyixuxu, line 373: I'm not exactly sure I understand the code entirely here (I think the code as it is is not able to run, no?)
so let me know if my suggestion does not make sense 

does it make sense to make a `Lumina2PatchEmbed` wrap all the logic we have here to create image_rotary_emb and cap_freqs_cis? (instead of the simple PosEmb that returns the freqs_cis based on position ids)

so the forward pass I have in mind is something like this 

```python
temb, encoder_hidden_states = self.time_caption_embed(timestep, encoder_hidden_states)

image_sizes = ...
list_text_seq_length = ..  
list_num_patches = ...
max_seq_len = ...

image_rotary_emb, text_rotary_emb = self.patch_embed(....)
for layer in self.context_refier:
   ...

padded_img_embed, padded_img_mask = self.patchify_image(..)

for layer in self.noise_refiner:
    ...

hidden_state, mask = .

for layer in self.layers:

...
```

zhuole1025: So sorry to cause you such confusion! Now I have updated the latest update, including making a Lumina2PatchEmbed following your suggestions. I have tested the code to correctly generate images.

src/diffusers/models/embeddings.py
```diff
@@ -1766,6 +1766,38 @@ def forward(self, timestep, encoder_hidden_states, image_meta_size, style, hidde
         return conditioning
 
 
+class Lumina2CombinedTimestepCaptionEmbedding(nn.Module):
+    def __init__(self, hidden_size=4096, cap_feat_dim=2048, frequency_embedding_size=256, norm_eps=1e-5):
+        super().__init__()
+        
+        from normalization import RMSNorm
```
rodjjo, line 1773: Is that suppose to be `from .normalization import RMSNorm` instead of `from normalization import RMSNorm`  ?

zhuole1025: thanks, this has been fixed

src/diffusers/models/attention_processor.py
```diff
@@ -4192,6 +4192,102 @@ def __call__(
...
+        self,
+        attn: Attention,
+        hidden_states: torch.Tensor,
+        encoder_hidden_states: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+        query_rotary_emb: Optional[torch.Tensor] = None,
+        key_rotary_emb: Optional[torch.Tensor] = None,
+        base_sequence_length: Optional[int] = None,
+    ) -> torch.Tensor:
+        from embeddings import apply_rotary_emb
```
rodjjo, line 4215: `from .embeddings import apply_rotary_emb`
I guess it should be a relative import to avoid import errors.

zhuole1025: fix that!

src/diffusers/models/attention.py
```diff
@@ -612,7 +612,6 @@ def __init__(
         ffn_dim_multiplier: Optional[float] = None,
     ):
         super().__init__()
-        inner_dim = int(2 * inner_dim / 3)
```
a-r-r-o-w, line 615: I think this is unintented/backwards-breaking change, yes? If so, let's revert.

Edit: oh okay, not backwards breaking. I see you've made the required changes in `lumina_nextdit2d.py`

src/diffusers/models/attention_processor.py
```diff
@@ -4192,6 +4192,102 @@ def __call__(
         return hidden_states
 
 
+class Lumina2AttnProcessor2_0:
```
a-r-r-o-w, line 4195: Let's move this attention processor implementation to the `transformer_lumina2.py` file, since all new models are now being added in single-file format

src/diffusers/models/attention_processor.py
```diff
@@ -4192,6 +4192,102 @@ def __call__(
...
+            raise ImportError("AttnProcessor2_0 requires PyTorch 2.0, to use it, please upgrade PyTorch to 2.0.")
+
+    def __call__(
+        self,
+        attn: Attention,
+        hidden_states: torch.Tensor,
+        encoder_hidden_states: torch.Tensor,
+        attention_mask: Optional[torch.Tensor] = None,
+        query_rotary_emb: Optional[torch.Tensor] = None,
+        key_rotary_emb: Optional[torch.Tensor] = None,
```
a-r-r-o-w, line 4212: I see that the `rotary_emb` being passed for both the query and key are the same. Not sure if I'm missing a case where both would have to be different in the implementation. If not supposed to be different, let's use a single `image_rotary_emb` parameter here

src/diffusers/models/attention_processor.py
```diff
@@ -4192,6 +4192,102 @@ def __call__(
...
+        base_sequence_length: Optional[int] = None,
+    ) -> torch.Tensor:
+        from .embeddings import apply_rotary_emb
+
+        input_ndim = hidden_states.ndim
+
+        if input_ndim == 4:
+            batch_size, channel, height, width = hidden_states.shape
+            hidden_states = hidden_states.view(batch_size, channel, height * width).transpose(1, 2)
```
a-r-r-o-w, line 4221:
```suggestion
```

I think we're passing ndim=3 tensors here, so can simplify this code and remove it

src/diffusers/models/embeddings.py
```diff
@@ -1766,6 +1766,38 @@ def forward(self, timestep, encoder_hidden_states, image_meta_size, style, hidde
         return conditioning
 
 
+class Lumina2CombinedTimestepCaptionEmbedding(nn.Module):
```
a-r-r-o-w, line 1769: Let's move this module to the `transformer_lumina2.py` file as well

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,451 @@
...
+                key_rotary_emb=rotary_emb,
+            )
+            hidden_states = hidden_states + self.norm2(attn_output)
+            mlp_output = self.feed_forward(self.ffn_norm1(hidden_states))
+            hidden_states = hidden_states + self.ffn_norm2(mlp_output)
+
+        return hidden_states
+    
+    
+class Lumina2PosEmbed(nn.Module):
```
a-r-r-o-w, line 144: Let's rename to `Lumina2RotaryPosEmbed` for naming consistency

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,451 @@
...
+        result = []
+        for i in range(len(self.axes_dim)):
+            freqs = self.freqs_cis[i].to(ids.device)
+            index = ids[:, :, i:i+1].repeat(1, 1, freqs.shape[-1]).to(torch.int64)
+            result.append(torch.gather(freqs.unsqueeze(0).repeat(index.shape[0], 1, 1), dim=1, index=index))
+        return torch.cat(result, dim=-1)
+    
+    def forward(
+        self,
+        x: list[torch.Tensor] | torch.Tensor,
```
a-r-r-o-w, line 175: For type hints, would prefer to use Union instead of `|` to be compatible with some older pytorch versions

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,451 @@
...
+            padded_img_mask[i, :l_effective_img_len[i]] = True
+            
+        return padded_img_embed, padded_img_mask, img_sizes, l_effective_cap_len, l_effective_img_len, freqs_cis, cap_freqs_cis, img_freqs_cis, max_seq_len
+
+
+class Lumina2Transformer2DModel(ModelMixin, ConfigMixin):
+    """
+    LuminaNextDiT: Diffusion model with a Transformer backbone.
+
+    Inherit ModelMixin and ConfigMixin to be compatible with the sampler StableDiffusionPipeline of diffusers.
```
a-r-r-o-w, line 245:
```suggestion
```

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,451 @@
...
+    ) -> None:
+        super().__init__()
+        self.sample_size = sample_size
+        self.patch_size = patch_size
+        self.in_channels = in_channels
+        self.out_channels = out_channels or in_channels
+        self.hidden_size = hidden_size
+        self.num_attention_heads = num_attention_heads
+        self.head_dim = hidden_size // num_attention_heads
+        self.scaling_factor = scaling_factor
```
a-r-r-o-w, line 307: These don't need to be saves as they can be accessed via `self.config.{param_name}`
```suggestion
```

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,451 @@
...
+            embedding_dim=hidden_size,
+            conditioning_embedding_dim=min(hidden_size, 1024),
+            elementwise_affine=False,
+            eps=1e-6,
+            bias=True,
+            out_dim=patch_size * patch_size * self.out_channels,
+        )
+        # self.final_layer = LuminaFinalLayer(hidden_size, patch_size, self.out_channels)
+
+        assert (hidden_size // num_attention_heads) % 4 == 0, "2d rope needs head dim to be divisible by 4"
```
a-r-r-o-w, line 381:
```suggestion
```

Let's do the assertion in the RoPE layer

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,451 @@
...
+            hidden_states[i, :cap_len] = encoder_hidden_states[i, :cap_len]
+            hidden_states[i, cap_len:cap_len+img_len] = padded_img_embed[i, :img_len]
+
+        for layer in self.layers:
+            hidden_states = layer(
+                hidden_states,
+                mask,
+                joint_rotary_emb,
+                temb=temb,
+            )
```
a-r-r-o-w, line 428: Needs gradient checkpointing to be enabled, similar to: https://github.com/huggingface/diffusers/blob/3e35f56b00d73bc3c2d3bb69615176d0909fab8a/src/diffusers/models/transformers/transformer_hunyuan_video.py#L734

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,825 @@
...
+    import ftfy
+
+EXAMPLE_DOC_STRING = """
+    Examples:
+        ```py
+        >>> import torch
+        >>> from diffusers import Lumina2Text2ImgPipeline
+
+        >>> pipe = Lumina2Text2ImgPipeline.from_pretrained(
+        ...     "Alpha-VLLM/Lumina-Next-SFT-diffusers", torch_dtype=torch.bfloat16
```
a-r-r-o-w, line 65: Just marking as a TODO for updating checkpoint id

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,825 @@
...
+        >>> # Enable memory optimizations.
+        >>> pipe.enable_model_cpu_offload()
+
+        >>> prompt = "Upper body of a young woman in a Victorian-era outfit with brass goggles and leather straps. Background shows an industrial revolution cityscape with smoky skies and tall, metal structures"
+        >>> image = pipe(prompt).images[0]
+        ```
+"""
+
+
+def calculate_shift(
```
a-r-r-o-w, line 76: Let's add a `# Copied from` here

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,825 @@
...
+                )
+            # Padding negative prompt to the same length with prompt
+            prompt_max_length = prompt_embeds.shape[1]
+            negative_text_inputs = self.tokenizer(
+                negative_prompt,
+                padding="max_length",
+                max_length=prompt_max_length,
+                truncation=True,
+                return_tensors="pt",
+            )
```
a-r-r-o-w, line 359: Could `_get_gemma_prompt_embeds` be leveraged here as well to compute negative embeds, or not really?

a-r-r-o-w (commented):
Thanks for the PR @zhuole1025! It looks great and almost ready to merge

We need to add tests and docs apart from addressing the comments/questions here. For both, [this](https://github.com/huggingface/diffusers/pull/10136/files) PR would serve as a good example of what files needs to be updated. In tests, a modeling test (`test_models_transformer_lumina2.py`) and a pipeline test (`test_lumina2.py`) will be needed. More than happy to help make any of the required changes to move the PR to completion 🤗

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,451 @@
...
+            freqs = self.freqs_cis[i].to(ids.device)
+            index = ids[:, :, i:i+1].repeat(1, 1, freqs.shape[-1]).to(torch.int64)
+            result.append(torch.gather(freqs.unsqueeze(0).repeat(index.shape[0], 1, 1), dim=1, index=index))
+        return torch.cat(result, dim=-1)
+    
+    def forward(
+        self,
+        x: list[torch.Tensor] | torch.Tensor,
+        cap_mask: torch.Tensor,
+        t: torch.Tensor,
```
a-r-r-o-w, line 177: Also, I think some of YiYi's comments are yet to be addressed (such as naming this to `temb`) and her questions. In general, I think you could change any occcurences of `x` to `hidden_states`, `cap_mask` to `attention_mask`, and similar naming conventions followed in latest diffusers implementations for consistency reasons

zhuole1025:
Thanks for all the suggestions! I have fix all of them and update a new version for review~ As for the tests, I am still writing them.

a-r-r-o-w:
@zhuole1025 @csuhan Let me know your thoughts about the changes in #10732. I've verified on my end that the inference works as expected, so we should be able to merge the PR soon after these changes are incorporated

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,798 @@
...
+        tokenizer (`AutoModel`):
+            Tokenizer of class
+            [AutoModel](https://huggingface.co/docs/transformers/model_doc/t5#transformers.AutoModel).
+        transformer ([`Transformer2DModel`]):
+            A text conditioned `Transformer2DModel` to denoise the encoded image latents.
+        scheduler ([`SchedulerMixin`]):
+            A scheduler to be used in combination with `transformer` to denoise the encoded image latents.
+    """
+
+    bad_punct_regex = re.compile(
```
yiyixuxu, line 167: is this used?

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,798 @@
...
+
+        self.image_processor = VaeImageProcessor(vae_scale_factor=self.vae_scale_factor * 2)
+
+        if getattr(self, "tokenizer", None) is not None:
+            self.tokenizer.padding_side = "right"
+
+    def _get_gemma_prompt_embeds(
+        self,
+        prompt: Union[str, List[str]],
+        num_images_per_prompt: int = 1,
```
yiyixuxu, line 221: I think we can maybe move the `num_images_pe_prompt` logic to `encode_prompt`
```suggestion
```

it is more practical to have a function that just encodes the prompt and does nothing else
we are working with @hlky to refactor all other `encode_prompt too`

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,798 @@
...
+
+        # 5. Prepare timesteps
+        sigmas = np.linspace(1.0, 1 / num_inference_steps, num_inference_steps) if sigmas is None else sigmas
+        image_seq_len = latents.shape[1]
+        mu = calculate_shift(
+            image_seq_len,
+            self.scheduler.config.get("base_image_seq_len", 256),
+            self.scheduler.config.get("max_image_seq_len", 4096),
+            self.scheduler.config.get("base_shift", 0.5),
+            self.scheduler.config.get("max_shift", 1.16),
```
yiyixuxu, line 697:
```suggestion
            self.scheduler.config.get("max_shift", 1.15),
```

https://huggingface.co/black-forest-labs/FLUX.1-dev/blob/main/scheduler/scheduler_config.json#L7
I think we have a typo in the flux code and should be 1.15 here (will fix in flux too)

yiyixuxu (approved):
thanks! I left some nits, feel free to merge!

docs/source/en/api/pipelines/lumina2.md
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License. -->
+
+# Lumina2
+
+[Lumina Image 2.0](https://huggingface.co/Alpha-VLLM/Lumina-Image-2.0) by Alpha-VLLM.
```
stevhliu, line 17: A little more detail here would be nice, you can even just copy it from the [model card](https://huggingface.co/Alpha-VLLM/Lumina-Image-2.0) :)

zhuole1025:
Thanks for the suggestions and update the final version!

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,787 @@
...
+        # 6. Denoising loop
+        with self.progress_bar(total=num_inference_steps) as progress_bar:
+            for i, t in enumerate(timesteps):
+                # expand the latents if we are doing classifier free guidance
+                latent_model_input = (
+                    torch.cat([latents] * 2)
+                    if do_classifier_free_guidance
+                    and 1 - t.item() / self.scheduler.config.num_train_timesteps < cfg_trunc_ratio
+                    else latents
+                )
```
a-r-r-o-w, line 707: We were profiling the Lumina2 implementation before merging the PR and found these parts to introduce slowdowns during inference. Each step takes about ~0.27-0.30 seconds extra due to a GPU->CPU stream synchronization operation. We investigate the issue more deeply internally, but I will try to keep the explanation brief here.

1. When we perform data dependent branching such that CPU depends on data from the GPU, it introduces a GPU->CPU stream sync operation. This makes it so that the CPU cannot asynchronously queue instructions for the GPU kernel launches until the data transfer from GPU->CPU is complete.
2. It is okay to have CPU->GPU syncs (and we usually have many for most integrations), but the opposite is much worse for inference time required -- in this case we have it after every inference step.

There are two instances where possible GPU->CPU synchronization occurs for Lumina2. The first is:
```python
latent_model_input = (
  torch.cat([latents] * 2)
  if do_classifier_free_guidance
  and 1 - t.item() / self.scheduler.config.num_train_timesteps < cfg_trunc_ratio
  else latents
)
```

The second is:

```python
if do_classifier_free_guidance and current_timestep[0] < cfg_trunc_ratio:
#                     this part -> -------------------------------------
```

3. In diffusers, the `scheduler.timesteps` existing on the GPU, so accessing via `.item()` for a CPU computation, or the CPU-required data dependant branching, introduce these sync points.
4. Unless we remove these synchronization, the inference will be much slower for users.

Our proposed fix to not lose any performance out-of-the-box is:
- Let's assume that `cfg_trunc_ratio` means the % of inference steps after which CFG is to be turned off. The % value will be represented between 0 and 1 (same as currently implemented).
- Instead of accessing the timesteps to determine if we've reached cutoff ratio, let's use the denoising loop variable `i`. So, let's do something like `(i + 1) / num_inference_steps < cfg_trunc_ratio`.
- We will have to update both occurrences mentioned above as they use timestep-based if statements that need to be run on CPU

Really sorry about the back and forth bothering you. We would like to merge the PR quickly too but different things keep coming up 😅 Any further improvements required can be done in follow-up PRs but let's try to do just this last one

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,787 @@
...
+                noise_pred = self.transformer(
+                    hidden_states=latent_model_input,
+                    timestep=current_timestep,
+                    encoder_hidden_states=prompt_embeds,
+                    attention_mask=prompt_attention_mask,
+                    return_dict=False,
+                )[0]
+
+                # perform normalization-based guidance scale on a truncated timestep interval
+                if do_classifier_free_guidance and current_timestep[0] < cfg_trunc_ratio:
```
a-r-r-o-w, line 739: Same comment as above for `current_timestep[0] < cfg_trunc_ratio`

a-r-r-o-w, line 739:
```suggestion
                if self.do_classifier_free_guidance and current_timestep[0] < cfg_trunc_ratio:
```

Since callbacks can also modify whether to perform CFG or not, we should use `self.do_classifier_free_guidance`

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
```diff
@@ -0,0 +1,787 @@
...
+                        dtype = torch.float32 if is_mps else torch.float64
+                    else:
+                        dtype = torch.int32 if is_mps else torch.int64
+                    current_timestep = torch.tensor(
+                        [current_timestep],
+                        dtype=dtype,
+                        device=latent_model_input.device,
+                    )
+                elif len(current_timestep.shape) == 0:
+                    current_timestep = current_timestep[None].to(latent_model_input.device)
```
a-r-r-o-w, line 723: We can simplify this logic a bit, or even remove completely if not required since current_timestep will always be a tensor with the latest scheduler implementation in diffusers

As a reference, the HunyuanVideo implementation is good:

https://github.com/huggingface/diffusers/blob/464374fb87610c53b2cf81e08d3df628fada3ce4/src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video.py#L652-L655

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
a-r-r-o-w: If done without synchronizations, we should hopefully be seeing close to 10 seconds or more speedup per 1024x1024 image at 50 denoising steps compared to current implementation on consumer GPUs

a-r-r-o-w (approved):

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
a-r-r-o-w: Here are some relevant images from the traces:

image

image

image

image

"inside trunc ratio" was the profiler annotation for computing the boolean condition using current timestep and cfg_trunc_ratio

Column01:
I just installed diffusers from the PR branch to try out lumina2, so far it works for normal use just fine, but enabling vae tiling or vae slicing causes the following error:

```
File "c:\Users\colin\Documents\_GithubRepos\AI-Function-Agent\functions\system\image_gen.py", line 92, in <module>
    gen_image(prompt)
  File "c:\Users\colin\Documents\_GithubRepos\AI-Function-Agent\functions\system\image_gen.py", line 40, in gen_image
    image = pipe(
            ^^^^^
  File "C:\Users\colin\Documents\_GithubRepos\AI-Function-Agent\venv\Lib\site-packages\torch\utils\_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\colin\Documents\_GithubRepos\AI-Function-Agent\venv\Lib\site-packages\diffusers\pipelines\lumina2\pipeline_lumina2.py", line 773, in __call__
    noise_pred_cond, noise_pred_uncond = torch.split(
                                         ^^^^^^^^^^^^
  File "C:\Users\colin\Documents\_GithubRepos\AI-Function-Agent\venv\Lib\site-packages\torch\functional.py", line 207, in split
    return tensor.split(split_size_or_sections, dim)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\colin\Documents\_GithubRepos\AI-Function-Agent\venv\Lib\site-packages\torch\_tensor.py", line 981, in split
    return torch._VF.split(self, split_size, dim)  # type: ignore[attr-defined]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: split_size can only be 0 if dimension size is 0, but got dimension size of 1
```

a-r-r-o-w:
@Column01 Could you share the minimal reproducible example of what you're trying please? It will help us understand where the issue comes from

nitinmukesh:
@Column01 

Remove this and it will work. This was working earlier but not after some changes.
cfg_trunc_ratio=0.25,
cfg_normalization=True,

Not sure if these parameters should be updatable or internal only.

nitinmukesh:
@a-r-r-o-w 

Here is the sample code to reproduce the issue

```
import torch
from diffusers import Lumina2Text2ImgPipeline, Lumina2Transformer2DModel

bfl_repo = "Alpha-VLLM/Lumina-Image-2.0"
dtype = torch.bfloat16
pipe = Lumina2Text2ImgPipeline.from_pretrained(
    bfl_repo,
    torch_dtype=torch.bfloat16
)
pipe.vae.enable_slicing()
pipe.vae.enable_tiling()
prompt = "Close-up portrait of a young woman with light brown hair, looking to the right, illuminated by warm, golden sunlight. Her hair is gently tousled, catching the light and creating a halo effect around her head. She wears a white garment with a V-neck, visible in the lower left of the frame. The background is dark and out of focus, enhancing the contrast between her illuminated face and the shadows. Soft, ethereal lighting, high contrast, warm color palette, shallow depth of field, natural backlighting, serene and contemplative mood, cinematic quality, intimate and visually striking composition"
image = pipe(
    prompt,
    width=576,
    height=1024,
    guidance_scale=4.0,
    num_inference_steps=30,
    cfg_trunc_ratio=0.25,
    cfg_normalization=True,
).images[0]
image.save("lumina_4.png")

```
Removing the following works (it was working earlier with these parameters). Not sure if these parameters are customizable (output quality...etc) or fixed values in code

cfg_trunc_ratio=0.25,
cfg_normalization=True,

Column01:
@nitinmukesh's sample is functionally equivalent to what I had to produce that error

src/diffusers/pipelines/lumina2/pipeline_lumina2.py
zhuole1025: really appreciate you in-depth analysis that will help us to optimize the inference efficiency! I have fixed them in the latest commit.

src/diffusers/models/transformers/transformer_lumina2.py
```diff
@@ -0,0 +1,539 @@
...
+        # perform Grouped-qurey Attention (GQA)
+        n_rep = attn.heads // kv_heads
+        if n_rep >= 1:
+            key = key.unsqueeze(3).repeat(1, 1, 1, n_rep, 1).flatten(2, 3)
+            value = value.unsqueeze(3).repeat(1, 1, 1, n_rep, 1).flatten(2, 3)
+
+        # scaled_dot_product_attention expects attention_mask shape to be
+        # (batch, heads, source_length, target_length)
+        attention_mask = attention_mask.bool().view(batch_size, 1, 1, -1)
+        attention_mask = attention_mask.expand(-1, attn.heads, sequence_length, -1)
```
hlky, line 133:
```suggestion
```
🚀

Column01:
Can confirm after pulling the latest changes I was unable to get the program to crash with vae slicing and tiling enabled, but further testing that I did **BEFORE** updating revealed that their enabling wasn't the problem and on some runs it would crash with just the below parameters included. After updating, it appears to no longer be an issue

```
cfg_trunc_ratio=0.25,
cfg_normalization=True,
```

src/diffusers/models/transformers/transformer_lumina2.py
zhuole1025: Since we are using GQA in lumina2, is it okay to remove this line?

nitinmukesh:
lumina_8

A futuristic AI android with a sleek metallic face, glowing circuits, and an ultra-modern sci-fi aesthetic. Cybernetic implants visible, digital HUD elements in the background.

Vargol:
THeres a few MPS incompatibilities currently a couple of things Torch on MPS doesn't support and doesn't fall back to CPU on, and a float64 reference.

```py
 in _get_freqs_cis
    result.append(torch.gather(freqs.unsqueeze(0).repeat(index.shape[0], 1, 1), dim=1, index=index))
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: repeat(): Not supported for complex yet!
```

```py
 in _get_freqs_cis
    result.append(torch.gather(xx.repeat(index.shape[0], 1, 1), dim=1, index=index))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: gather(): Yet not supported for complex
```

```py
  File "/Volumes/SSD2TB/AI/Lumina/lib/python3.11/site-packages/diffusers/models/transformers/transformer_lumina2.py", line 298, in forward
    cap_freqs_cis = torch.zeros(*cap_freqs_cis_shape, device=device, dtype=freqs_cis.dtype)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Cannot convert a float64 Tensor to MPS as the MPS framework doesn't support float64. Please use float32 instead.
```
The float64 reference is actually in `_precompute_freqs_cis`

```py
emb = get_1d_rotary_pos_embed(d, e, theta=self.theta, freqs_dtype=torch.float64)
```

yiyixuxu:
@Vargol thanks for reporting, can you test to see if it run on MPS now with the latest commit https://github.com/huggingface/diffusers/pull/10642/commits/9ea9bbebf6c6045b0ce5b3faf35ec23b3b6c8135?

Vargol:
It fixes the float64, but the torch unsupported bits are still there, not sure if you expected that.
I still had to hack `_get_freqs_cis` to 

```py
   def _get_freqs_cis(self, ids: torch.Tensor) -> torch.Tensor:
        result = []
        for i in range(len(self.axes_dim)):
            freqs = self.freqs_cis[i].to(ids.device)
            index = ids[:, :, i : i + 1].repeat(1, 1, freqs.shape[-1]).to('cpu', torch.int64)
            freq_unsqueezed = freqs.unsqueeze(0).to("cpu")
            result.append(torch.gather(freq_unsqueezed.repeat(index.shape[0], 1, 1), dim=1, index=index))
        return torch.cat(result, dim=-1).to('mps')
```

To get it working on MPS.

yiyixuxu:
@Vargol can you test on this PR? https://github.com/huggingface/diffusers/pull/10776

vladmandic:
quick q - if i look at reference implementation, it relies on `flash_attn`, but diffusers implementation does not?

a-r-r-o-w:
@vladmandic Diffusers only supports directly calling torch SDPA. Custom attention processors can be used for flash-attn or other backends. To enable flash attention with pytorch, you can set the torch backend to `SDPBackend.FLASH_ATTENTION`.

We briefly discussed a dispatching mechanism for different libraries/backends but no decisions on how to move forward yet. Maybe something to consider for the release after current schedule

yiyixuxu:
@asomoza 
I think not using mask affects the quality of shorter prompt quite a bit if using multiple prompts, can you verify?

```python
# test lumina2
import torch
from diffusers import Lumina2Text2ImgPipeline
import itertools
from pathlib import Path
import shutil

# branch = "refactor_lumina2"
branch = "main"
params = {
    'use_mask_in_transformer': [True, False],  
}

# Generate all combinations
param_combinations = list(itertools.product(*params.values()))

# Create output directory (remove if exists)
output_dir = Path(f"yiyi_test_6_outputs_{branch}")
if output_dir.exists():
    shutil.rmtree(output_dir)
output_dir.mkdir(exist_ok=True)

pipe = Lumina2Text2ImgPipeline.from_pretrained("Alpha-VLLM/Lumina-Image-2.0", torch_dtype=torch.bfloat16).to("cuda")

prompt = [
    "focused exterior view on a living room of a limestone rock regular blocks shaped villa with sliding windows and timber screens in Provence along the cliff facing the sea, with waterfalls from the roof to the pool, designed by Zaha Hadid, with rocky textures and form, made of regular giant rock blocks stacked each other with infinity edge pool in front of it, blends in with the surrounding nature. Regular rock blocks. Giant rock blocks shaping the space. The image to capture the infinity edge profile of the pool and the flow of water going down creating a waterfall effect. Adriatic Sea. The design is sustainable and semi prefab. The photo is shot on a canon 5D mark 4",
    "A capybara holding a sign that reads Hello World"
]

# Run test for each combination
for (mask,) in param_combinations:
    print(f"\nTesting combination:")
    print(f"  use_mask_in_transformer: {mask}")
    
    # Generate image
    generator = torch.Generator(device="cuda").manual_seed(0)
    images = pipe(
        prompt=prompt,
        num_inference_steps=25,
        use_mask_in_transformer=mask,
        generator=generator,
    ).images
    
    # Save images
        output_path = output_dir / f"output_mask{int(mask)}_prompt{i}.png"
        image.save(output_path)
        print(f"Saved to: {output_path}")
```

<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

asomoza:
There's something different in your version or env, I don't get that bad result, also there's a couple of things to take notice:

- You're not using a system prompt and I think it was trained to use them all the time.
- It's really not a real use case scenario to use a really short prompt.
- Text generation is really finicky everywhere and even if it's a cool research project and to play with it, I don't use it to benchmark because I can literally do the a better job in seconds with any image software and it's not really usable in production. And even with flux I sometimes get bad results or low quality texts.
- There's some quirks of the model that I haven't discovered how to use yet, for example, I can't get it to produce pose estimation images (with mask or no mask).

But to compare it to your results, this are my images without system prompt:

|mask|no mask|
|---|---|
|capybara_mask|capybara_no_mask|
|cat_mask|cat_no_mask|

with a system prompt and the capybara I get a more similar result to yours but without the bad text:

|mask|no mask|
|---|---|
|capybara_sp_mask|capybara_sp_no_mask|

There's a difference but I need a lor more tests to see if there's a loss in quality, sometimes I think they masked ones are better but sometimes not, also most people won't notice the difference unless you show them side by side.

I do test the models with texts and with all the stuff I see in the paper or technical report, and for lumina2, it's really hard to tell yet if there's a loss in quality for me, maybe we will see a more clear difference with controlnets and loras.

IMO we stiill ned to make the default `True` just in case even if it adds some seconds in the generation time.

yiyixuxu:
@asomoza 
it comes with default system prompt,  you don't have to explicitly provide one https://github.com/huggingface/diffusers/blob/28f48f4051e80082cbe97f2d62b365dbb01040ec/src/diffusers/pipelines/lumina2/pipeline_lumina2.py#L194

my results are generated when using a list of prompts as same time, I linked a test script in my previous comments here https://github.com/huggingface/diffusers/pull/10642#issuecomment-2654492988

asomoza:
oh I totally missed that, so I was using a double system prompt. But still, I don't get that bad result and without the system prompt it generates cartoon images for me. I'll take another look then.

asomoza:
with your code, I can see the difference, not that much as yours though because both are bad for me:

|mask|no mask|
|---|---|
|output_mask1_prompt1|output_mask0_prompt1|

@yiyixuxu  since this is happening with a list of prompts, isn't the issue with that instead of the mask or no mask? I don't use that option because it's not really usable for generating images, so I'll add it to my tests from now on, but the difference for the capybara it's really noticeable versus the single prompt ones.

yiyixuxu:
@asomoza 
I don't think you are using double, it's just if you don't provide one, it have a default, otherwise it will use yours
>oh I totally missed that, so I was using a double system prompt.

for the outputs, are you getting a different results with the same testing script? is this pipeline not deterministic?

asomoza:
some more tests:

### List of prompts with:

```python
prompt = [
    "A cat wearing sunglasses, in the back there's a sign that says: Beware of the cat!",
    "A capybara holding a sign that reads: I'm cute or what?",
]
```

|mask|no mask|
|---|---|
|output_mask1_prompt0|output_mask0_prompt0|
|output_mask1_prompt1|output_mask0_prompt1|

### Single prompts:

```python
prompt = "A cat wearing sunglasses, in the back there's a sign that says: Beware of the cat!"
```

|mask|no mask|
|---|---|
|cat_mask|cat_no_mask|
|capybara_mask|capybara_no_mask|

I think this is just random, we're getting some bad results and good results depending on the seed and the prompt probably.

> for the outputs, are you getting a different results with the same testing script? is this pipeline not deterministic?

no, the pipeline is deterministic, I got different results when I switched to use your script with a list of prompts, that's expected right? since we're using the same generator.

With a single prompt I can reproduce the exact same image all the time, BTW I'm always using a seed of `0` for these tests.

asomoza:
> I don't think you are using double, it's just if you don't provide one, it have a default, otherwise it will use yours

I tested this and you're correct, I get the same exact image with and without the system prompt.

asomoza:
so using a list, and switching the order:

long prompt first:

|mask|no mask|
|---|---|
|output_mask1_prompt0|output_mask1_prompt1|
|output_mask0_prompt0|output_mask0_prompt1|

short prompt first:

|mask|no mask|
|---|---|
|output_mask1_prompt0|output_mask1_prompt1|
|output_mask0_prompt0|output_mask0_prompt1|

Doing some more tests, the order and the longer prompt doesn't matter, the only one that changes a lot is the shorter prompt, so I'll post those only:

shorter prompt without text: "Photo of a capybara wearing sunglasses and a jacket"

|mask|no mask|
|---|---|
|output_mask1_prompt0|output_mask0_prompt0|

even shorter prompt without text: "a capybara"

|mask|no mask|
|---|---|
|output_mask1_prompt0|output_mask0_prompt0|

medium prompt with text: "photo of a cat wearing stylish black sunglasses. The cat has a light brown and white fur pattern with distinct stripes. Behind the cat, there is a wooden sign with the text: Beware of the cat!!! written in a playful, handwritten style."

|mask|no mask|
|---|---|
|output_mask1_prompt0|output_mask0_prompt0|

medium prompt without text: "photo of a cat wearing stylish black sunglasses and a leather jacket. The cat has a light brown and white fur pattern with distinct stripes. The overall scene has a humorous and whimsical tone, combining the cat's cool demeanor with human clothes and eyewear."

|mask|no mask|
|---|---|
|output_mask1_prompt0|output_mask0_prompt0|

Finished with the testings, this is a nice find @yiyixuxu , we can infer that without a mask and using a list of prompts where one is shorter, the text generation will take a quality hit, so this is important for web services or APIs that use this model with those.

Overall, I can't really see any other loss in quality with the exception that a shorter prompt, without details, will look worse without a mask because the model does the very basic with the prompt. For example, I noticed that without a mask and if you ask for a subject only, the model will most of the time generate the subject without a background which is really nice IMO.

asomoza:
also an example with just text:

```python
prompt = [
    "A sign that says: Lumina2 and diffusers are really good!!!",
    "A sign that says: We should make the capybara the official diffusers mascot!",
]
```

|mask|no mask|
|---|---|
|output_mask1_prompt0|output_mask0_prompt0|
|output_mask1_prompt1|output_mask0_prompt1|

PR #10185, opened 2024-12-11
Fix RePaint Scheduler

hlky:
# What does this PR do?

Algorithm 1 Line 5 https://arxiv.org/pdf/2201.09865.pdf

$$x_{t-1}^{\text{known}} = \sqrt{\bar{\alpha}_t} x_0 + (1 - \bar{\alpha}_t) \epsilon$$

Current code
```python
prev_known_part = (alpha_prod_t_prev**0.5) * original_image + ((1 - alpha_prod_t_prev) ** 0.5) * noise
```

Square root of `(1 - alpha_prod_t_prev)` is incorrect.

Fixes #9900

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

cc @yiyixuxu

PR #10644, opened 2025-01-24
Revert RePaint scheduler 'fix'

GiusCat:
# What does this PR do?

Reverts 'fix' introduced by pull request #10185; the PR adjusted the code accordingly to Algorithm 1 Line 5 of the RePaint paper ([https://arxiv.org/pdf/2201.09865.pdf](https://arxiv.org/pdf/2201.09865.pdf)), however that line is reported incorrectly in the paper itself.

Algorithm 1 Line 5 refers to formula 8a of the same paper

$$x_{t-1}^{\text{known}} \sim \mathcal{N}(\sqrt{\bar{\alpha}} x_0, (1 - \bar{\alpha}) \mathbf{I})$$

which tells to sample from a Gaussian distribution $\mathcal{N}(\mu, \sigma^2)$ with $\mu = \sqrt{\bar{\alpha}} x_0$ and $\sigma^2 = (1 - \bar{\alpha})$. In order to obtain this result, the standard Gaussian distribution $\epsilon \sim \mathcal{N}(\mathbf{0}, \mathbf{I})$ should be stretched by a factor $\sigma$ and translated by a factor $\mu$ (source: [Wikipedia](https://en.wikipedia.org/wiki/Normal_distribution#General_normal_distribution)). The correct operation should therefore be

$$x_{t-1}^{\text{known}} = \sqrt{\bar{\alpha}} x_0 + \sqrt{(1 - \bar{\alpha})} \epsilon$$

as it was indeed calculated before the change; Algorithm 1 Line 5 wrongly reports a stretching factor of $\sigma^2$. This PR restores the previous (correct) computation and adds some comments on the subject.

Fixes degraded performance using RePaint. Below are examples obtained using the example code in the [documentation](https://huggingface.co/docs/diffusers/main/en/api/pipelines/repaint).

### Diffusers v0.31.0 (before the 'fix')
inference_0-31-0

### Diffusers v0.32.1 (after the 'fix')
inference_0-32-1

## Who can review?

- Schedulers: @yiyixuxu

### DIFF
--- a/src/diffusers/schedulers/scheduling_repaint.py
+++ b/src/diffusers/schedulers/scheduling_repaint.py
@@ -319,7 +319,11 @@ class RePaintScheduler(SchedulerMixin, ConfigMixin):
         prev_unknown_part = alpha_prod_t_prev**0.5 * pred_original_sample + pred_sample_direction + variance
 
         # 8. Algorithm 1 Line 5 https://arxiv.org/pdf/2201.09865.pdf
-        prev_known_part = (alpha_prod_t_prev**0.5) * original_image + (1 - alpha_prod_t_prev) * noise
+        # The computation reported in Algorithm 1 Line 5 is incorrect. Line 5 refers to formula (8a) of the same paper,
+        # which tells to sample from a Gaussian distribution with mean "(alpha_prod_t_prev**0.5) * original_image"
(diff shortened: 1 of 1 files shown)


yiyixuxu:
thanks @GiusCat

yiyixuxu (approved):

yiyixuxu:
cc @hlky here too

hlky (approved):

Issue #10634, opened 2025-01-23
The huggingface repo need to be fixed for Sana 2K and 4K models


nitinmukesh:
### Describe the bug

Hello @lawrence-cj ,

I am using Sana using diffusers. The issue is applicable for both these repos and maybe for 512/1024 but not tested.

    if inference_type == "Sana 4K":
        model_path = "Efficient-Large-Model/Sana_1600M_4Kpx_BF16_diffusers"
    else:
        model_path = "Efficient-Large-Model/Sana_1600M_2Kpx_BF16_diffusers"

    pipe_sana = SanaPipeline.from_pretrained(
        pretrained_model_name_or_path=model_path,
        variant="bf16",
        torch_dtype=torch.bfloat16,
        use_safetensors=True,
    )  

When I specify bf16 and use_safetensors it should only download bf16 models and not 32 bit. It is working fine for text_encoder and vae but not for transformer.

```
C:\Users\nitin\.cache\huggingface\hub\models--Efficient-Large-Model--Sana_1600M_2Kpx_BF16_diffusers\snapshots\c096bbd4f6da0daf181f4fbce5e7505051b8c75c>tree /F
Folder PATH listing for volume Windows-SSD
Volume serial number is CE9F-A6AE
C:.
│   model_index.json
│
├───scheduler
│       scheduler_config.json
│
├───text_encoder
│       config.json
│       model.bf16-00001-of-00002.safetensors
│       model.bf16-00002-of-00002.safetensors
│       model.safetensors.index.bf16.json
│
├───tokenizer
│       special_tokens_map.json
│       tokenizer.json
│       tokenizer.model
│       tokenizer_config.json
│
├───transformer
│       config.json
│       diffusion_pytorch_model-00001-of-00002.safetensors
│       diffusion_pytorch_model-00002-of-00002.safetensors
│       diffusion_pytorch_model.bf16.safetensors
│       diffusion_pytorch_model.safetensors.index.json
│
└───vae
        config.json
        diffusion_pytorch_model.bf16.safetensors
```

### Reproduction

```
from diffusers import SanaPipeline
model_path = "Efficient-Large-Model/Sana_1600M_2Kpx_BF16_diffusers"
pipe_sana = SanaPipeline.from_pretrained(
        pretrained_model_name_or_path=model_path,
        variant="bf16",
        torch_dtype=torch.bfloat16,
        use_safetensors=True,
    )
pipe_sana.to("cuda")
[...]

PR #10646, opened 2025-01-24
Fix for fetching variants only

DN6:
# What does this PR do?
With PR: #9869 we fixed downloading sharded variants only and mixed variants, however we missed the case where a component might have sharded non-variant files and non-sharded variant file, which is the case with issue: #10634

This PR:
1. Adds a simpler `has_variant` check and removes the previous checks. This check first checks to see if we are in a component folder and then checks if any variants exist within the folder. If no folder exists then skip trying to add the additional non variant file. 
2. Adds additional tests to check for the condition mentioned in issue #10634. 

Since `usable_filenames` is always populated with variants,  we capture the necessary variant files and what we're trying to avoid is extra file downloads.    

The only edge case I can think of here where this would fail (which passes with the current implementation) is if the filenames are the following:

Non-Variant in the main dir and a variant in a subfolder. Although I think this an edge case that we probably can't load anyway? I can't think of any pipelines that would have this configuration. 

```python
filenames = ["diffusion_pytorch_model.safetensors", f"unet/diffusion_pytorch_model.{variant}.safetensors"]
```
 

Fixes # (issue)
#10634

## Before submitting

## Who can review?


lawrence-cj:
Thanks @DN6 for supporting!

src/diffusers/pipelines/pipeline_loading_utils.py
```diff
@@ -191,37 +191,21 @@ def variant_compatible_siblings(filenames, variant=None) -> Union[List[os.PathLi
...
         component = component + "/" if component else ""
-        variant_index_re = re.compile(
-            rf"{component}({'|'.join(weight_prefixes)})\.({'|'.join(weight_suffixs)})\.index\.{variant}\.json$"
-        )
-        return any(f for f in variant_filenames if variant_index_re.match(f) is not None)
 
-    for filename in non_variant_filenames:
-        if convert_to_variant(filename) in variant_filenames:
-            continue
+        return any(f.startswith(component) for f in variant_filenames)
```
yiyixuxu, line 204: what appends if like we only have a bf16.bin and this is a non-variant safetensors ? stuff like this

DN6: As in we are trying to fetch something like this? 

```
        variant = "fp16"
        filenames = [
                f"vae/diffusion_pytorch_model.{variant}.bin",
                f"text_encoder/model.{variant}.bin",
                f"unet/diffusion_pytorch_model.{variant}.bin",
        ]
        model_filenames, variant_filenames = variant_compatible_siblings(filenames, variant=None)
```

yiyixuxu: like this, I think we should fetch the non-variant safetensors in this case, no?         
        
        variant = "fp16"
        filenames = [
                f"vae/diffusion_pytorch_model.{variant}.bin",
                f"text_encoder/model.{variant}.bin",
                f"unet/diffusion_pytorch_model.{variant}.bin",
                f"vae/diffusion_pytorch_model.safetensors",
                f"text_encoder/model.safetensors",
                f"unet/diffusion_pytorch_model.safetensors",
        ]

DN6: Hmm currently the behaviour on main is to return all the files in that list (both bin and safetensors) as  `usable_filenames` and I think the ignore patterns would remove the bin files, resulting in just the safetensors being downloaded.  

With this change only the fp16.bin files would be downloaded. Which feels technically "correct" to me since they are the "variant" files of each component. IMO non-variants should only be downloaded if no variant exists (regardless of format)

But this case implies that the proposal here is a breaking change, so I'll update to account for it.

src/diffusers/pipelines/pipeline_loading_utils.py
```diff
@@ -104,7 +104,7 @@ def is_safetensors_compatible(filenames, passed_components=None, folder_names=No
       extension is replaced with ".safetensors"
     """
     passed_components = passed_components or []
-    if folder_names is not None:
+    if folder_names:
```
DN6, line 107: This change is needed to correct a weird bug that was caught with this test
`tests/pipelines/test_pipelines.py::CustomPipelineTests::test_load_custom_github`

What was basically happening was that `variant_compatible_siblings` was incorrectly returning both the bin and safetensors version of the checkpoint in this repo
https://huggingface.co/google/ddpm-cifar10-32/tree/main

In the DiffusionPipeline download method, if `model_folder_names` ends up being an empty set  
https://github.com/huggingface/diffusers/blob/33f936154db0bc7080960316b4ddb291e9555bf7/src/diffusers/pipelines/pipeline_utils.py#L1414 and is passed to `is_safetensors_compatible` the function will return false because filenames will end up being an empty set. But since  `variant_compatible_siblings` was asking to fetch both bin and safetensors versions of the checkpoint, the test passed because the bin version was downloaded.

DN6:
@yiyixuxu Had to do a bit more of a refactor to account for safetensors prioritization. But it should be much more robust to handle any number of repo file combinations. I've added a test for your case and a few others as well. I think they should cover all likely repo file layout scenarios, but if there are others I may have missed lmk.

nitinmukesh:
@DN6 

Here is another test case with fp16, Failing after using 2 updated files from 
https://github.com/huggingface/diffusers/pull/10646/files

```
import torch
from diffusers import AuraFlowPipeline

pipe = AuraFlowPipeline.from_pretrained(
    "fal/AuraFlow-v0.3",
    variant="fp16",
    torch_dtype=torch.float16,
    use_safetensors=True,
)
```

Warning: The repository contains sharded checkpoints for variant 'fp16' maybe in a deprecated format. Please check your files carefully:

- Correct format example: diffusion_pytorch_model.fp16-00003-of-00003.safetensors
- Deprecated format example: diffusion_pytorch_model-00001-of-00002.fp16.safetensors

If you find any files in the deprecated format:
1. Remove all existing checkpoint files for this variant.
2. Re-obtain the correct files by running `save_pretrained()`.

This will ensure you're using the most up-to-date and compatible checkpoint format.
model_index.json: 100%|██████████████████████████████████████████████████| 458/458 [00:00<?, ?B/s]

```
A mixture of fp16 and non-fp16 filenames will be loaded.
Loaded fp16 filenames:
[vae/diffusion_pytorch_model.fp16.safetensors, text_encoder/model.fp16.safetensors]
Loaded non-fp16 filenames:
[transformer/diffusion_pytorch_model-00001-of-00003.safetensors, transformer/diffusion_pytorch_model-00002-of-00003.safetensors, transformer/diffusion_pytorch_model-00003-of-00003.safetensors
If this behavior is not expected, please check your folder structure.
tokenizer.model: 100%|█████████████████████████████████████████| 500k/500k [00:00<00:00, 10.5MB/s]
tokenizer/tokenizer_config.json: 100%|███████████████████████| 21.0k/21.0k [00:00<00:00, 19.1MB/s]
tokenizer/added_tokens.json: 100%|████████████████████████████| 2.59k/2.59k [00:00<00:00, 573kB/s]
scheduler/scheduler_config.json: 100%|███████████████████████████████████| 142/142 [00:00<?, ?B/s]
tokenizer/special_tokens_map.json: 100%|█████████████████████████████| 2.68k/2.68k [00:00<?, ?B/s]
text_encoder/config.json: 100%|███████████████████████████████████| 949/949 [00:00<00:00, 951kB/s]
tokenizer/tokenizer.json: 100%|██████████████████████████████| 1.86M/1.86M [00:00<00:00, 7.64MB/s]
transformer/config.json: 100%|████████████████████████████████████| 379/379 [00:00<00:00, 376kB/s]
vae/config.json: 100%|███████████████████████████████████████████████████| 859/859 [00:00<?, ?B/s]
(…)ion_pytorch_model.safetensors.index.json: 100%|███████████| 36.0k/36.0k [00:00<00:00, 9.04MB/s]
diffusion_pytorch_model.fp16.safetensors: 100%|████████████████| 167M/167M [01:10<00:00, 2.36MB/s]
model.fp16.safetensors:   4%|█▍                               | 126M/2.95G [01:08<25:54, 1.82MB/s]
model.fp16.safetensors:   5%|█▊                               | 157M/2.95G [01:20<20:22, 2.28MB/s]
(…)pytorch_model-00001-of-00003.safetensors:   2%|▎           | 241M/9.99G [01:07<47:46, 3.40MB/s]
(…)pytorch_model-00001-of-00003.safetensors:   3%|▎           | 294M/9.99G [01:21<41:42, 3.88MB/s]
(…)pytorch_model-00002-of-00003.safetensors:   2%|▏         | 178M/9.89G [01:20<1:09:08, 2.34MB/s]
(…)pytorch_model-00003-of-00003.safetensors:   4%|▍           | 273M/7.55G [01:22<31:42, 3.83MB/s]

```


image

yiyixuxu:
it seems we are changing the behavior of `variant_compatible_siblings` a bit
currently, `variant_compatible_siblings` does not consider extensions, we address that in the `download` step https://github.com/huggingface/diffusers/blob/23bc56a02de3745e1eb19f5703280972e195ff56/src/diffusers/pipelines/pipeline_utils.py#L1332 (related arguments are `from_flax`, `use_onnx` and `use_safetensors`)

so my questions is:
1. is it more efficient to move the extension filter outside of `download` and earlier in the process
2. if so, should we apply it too all extensions? is there a reason to single out `safetensors`?

DN6:
> is it more efficient to move the extension filter outside of download and earlier in the process

Yes, I think it would be better to add an extension filter/check earlier and then filter variants after (without extensions checks). 

The current behaviour of decoupling the steps is a bit confusing and makes it harder to reason about which files end up getting downloaded. For instance in this case:

```
        filenames = [
            "text_encoder/model.bin",
            "unet/diffusion_pytorch_model.bin",
            "unet/diffusion_pytorch_model.safetensors",
        ]
```
`variant_compatible_siblings` will currently return all the files in this list, and the `is_safetensors_compatible` check will return False because not all components have safetensors. Which means `bin` files won't be filtered out in`_get_ignore_patterns` and all the `unet` files in this case will be downloaded. While we could consider this an edge case, I think it illustrates the point that the current file filtering mechanism is not easy to follow.   

I think the core issue is that we apply checks/filters currently over the full list of filenames while we should actually be checking per component in order to support mixed downloading properly, as well as things like downloading safetensors over bin files, non-sharded-variants for sharded-non-variants etc. 

We can either add additional checks in `variant_compatible_siblings` or refactor `download` a bit to make it more clear exactly which files will be used. I can update this PR to do that if that sounds good to you?  

> if so, should we apply it too all extensions? is there a reason to single out safetensors?

It was an oversight on my part. I was focusing on solving the use case you pointed out and forgot about the use_onnx and from_flax flags.

yiyixuxu:
>We can either add additional checks in variant_compatible_siblings or refactor download a bit to make it more clear exactly which files will be used. I can update this PR to do that if that sounds good to you?

either approach sounds good to me:)

DN6:
@yiyixuxu This is ready for another review. It should be able to handle all file format types. I made some changes in download so that `ignore_patterns` is created earlier. We then leverage that to filter out files and variants.

yiyixuxu (approved):
thanks @DN6 
I think it is a real nice refactor!!! (and really tricky too)

I left some small feedback on the PR

src/diffusers/pipelines/pipeline_loading_utils.py
```diff
@@ -177,54 +199,68 @@ def variant_compatible_siblings(filenames, variant=None) -> Union[List[os.PathLi
...
+            component_variants = filter_with_regex(component_filenames, variant_file_re)
+            component_variant_index_files = filter_with_regex(component_filenames, variant_index_re)
+
+            component_legacy_variants = filter_with_regex(component_filenames, legacy_variant_file_re)
+            component_legacy_variant_index_files = filter_with_regex(component_filenames, legacy_variant_index_re)
+
+        if component_variants:
+            variant_filenames.update(
+                component_variants | component_variant_index_files
+                if component_variants
```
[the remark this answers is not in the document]
DN6: Oh missed this. Great catch! Updated to account for this.

PR #10649, opened 2025-01-25
CogView4 (supports different length c and uc)

zRzRzRzRzRzRzR:
# What does this PR do?

This PR aims to support the adaptation of the CogView with c and uc different length  for diffusers.

We have reproduced the algorithm implementation, but this PR still requires further refinement. Currently, the output is pure green noise, so this PR remains in draft status and requires help from @a-r-r-o-w and @yiyixuxu.


a-r-r-o-w:
@zRzRzRzRzRzRzR @OleehyO Thanks for the PR! I'm going to take a look soon and try to help with debugging Megatron -> Diffusers. I see some additional changes made to the files that are not relevant to CogView (currently more than 200 files have been changed 😅). Could you revert those changes by doing something like `git restore -s main examples/ scripts/`

zRzRzRzRzRzRzR:
@zRzRzRzRzRzRzR @OleehyO Thanks for the pull request! I'll review it shortly and attempt to assist with debugging the transition from Megatron to Diffusers. I've noticed some modifications to files that aren't pertinent to CogView (it looks like over 200 files have been altered 😅). Could you undo those changes by running a command like `git restore -s main examples/ scripts/`?

This is my mistake, I have corrected it. I was trying to verify whether the new weights and the sat version are correctly aligned (q,k,v parts). It can be confirmed now that #[e239c3c](https://github.com/huggingface/diffusers/pull/10649/commits/e239c3cd54b3f609c14dbf8fa1d75028309bb6bb) is aligned, but I have not yet verified if it is aligned with megatron. I am working on this task.

check_same_sat_megatron_diffusers.py
```diff
@@ -0,0 +1,264 @@
+import torch
```
a-r-r-o-w, line 1: Since this is the comparison script to match outputs, maybe you can update the description of the PR with this code for future reference. But since it is not essential for integration, let's remove it from here

docs/source/en/api/models/cogview4_transformer2d.md
```diff
@@ -0,0 +1,30 @@
...
+specific language governing permissions and limitations under the License. -->
+
+# CogView4Transformer2DModel
+
+A Diffusion Transformer model for 2D data from [CogView4]()
+
+The model can be loaded with the following code snippet.
+
+```python
+from diffusers import CogView3PlusTransformer2DModel
```
a-r-r-o-w, line 19:
```suggestion
from diffusers import CogView4Transformer2DModel
```

docs/source/en/api/models/cogview4_transformer2d.md
```diff
@@ -0,0 +1,30 @@
...
+
+transformer = CogView3PlusTransformer2DModel.from_pretrained("THUDM/CogView4-6B", subfolder="transformer", torch_dtype=torch.bfloat16).to("cuda")
```
a-r-r-o-w, line 21:
```suggestion
transformer = CogView4Transformer2DModel.from_pretrained("THUDM/CogView4-6B", subfolder="transformer", torch_dtype=torch.bfloat16).to("cuda")
```

scripts/convert_cogview4_to_diffusers_megatron.py
```diff
@@ -0,0 +1,384 @@
...
+            "mid_block_add_attention": False,
+        }
+        converted_vae_state_dict = convert_cogview4_vae_checkpoint_to_diffusers(args.vae_checkpoint_path, vae_config)
+        vae = AutoencoderKL(**vae_config)
+        vae.load_state_dict(converted_vae_state_dict, strict=True)
+        if dtype is not None:
+            vae = vae.to(dtype=dtype)
+
+    # Load the text encoder and tokenizer
+    text_encoder_id = "/share/home/zyx/Models/glm-4-9b-hf"
```
a-r-r-o-w, line 339: I believe this should be `THUDM/glm-4-9b-hf` too

src/diffusers/models/attention_processor.py
```diff
@@ -2803,6 +2803,104 @@ def __call__(
             return hidden_states
 
 
+class CogView4AttnProcessor:
```
a-r-r-o-w, line 2806: From our DM, if I understand correctly, the CogView4 team has verified that the original implementation and CogView4 diffusers adaptation matched already. If so, I think we can remove the debug code and comments that are not necessary.

Let's also move this attention processor to the `transformer_cogview4.py` file because we're moving to a more single-file based implementation. You can take a look at the HunyuanVideo integration PR for an example (https://github.com/huggingface/diffusers/pull/10136)

src/diffusers/models/embeddings.py
```diff
@@ -1234,6 +1283,40 @@ def apply_1d_rope(tokens, pos, cos, sin):
     x = torch.cat([t, h, w], dim=-1)
     return x
 
+def apply_rotary_emb_megatron(x: torch.Tensor, freqs: torch.Tensor) -> torch.Tensor:
```
a-r-r-o-w, line 1286: This looks similar in logic to `apply_rotary_emb` logic, no? We can prepare the cos and sin values beforehand when RoPE frequencies are initialized so that this does not have to be computed for each block

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -0,0 +1,538 @@
...
+        self.ff = FeedForward(dim=dim, dim_out=dim, activation_fn="gelu-approximate")
+
+    def multi_modulate(self, hidden_states, encoder_hidden_states, factors) -> torch.Tensor:
+        n_sample, n_type, h = factors[0].shape
+        shift_factor, scale_factor = factors[0].view(-1, h), factors[1].view(-1, h)
+
+        shift_factor_hidden_states, shift_factor_encoder_hidden_states = shift_factor.chunk(2, dim=0)
+        scale_factor_hidden_states, scale_factor_encoder_hidden_states = scale_factor.chunk(2, dim=0)
+
+        hidden_states = torch.addcmul(shift_factor_hidden_states, hidden_states, (1 + scale_factor_hidden_states))
```
a-r-r-o-w, line 89: Let's directly write these as `hidden_states * (1 + scale) + shift`. This is to keep naming and implementation consistent with our model adaptations

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -0,0 +1,538 @@
...
+        batch_size, encoder_hidden_states_len, hidden_dim = encoder_hidden_states.shape
+        hidden_states = torch.cat([encoder_hidden_states, hidden_states], dim=1)
+
+        residual = hidden_states
+
+        # time_embedding embedding, [n_sample, h]
+        assert time_embedding is not None
+
+        layernorm_factor = (
+            self.adaln(time_embedding)
```
a-r-r-o-w, line 121: Normally, we don't alias an internal layer for direct use. I think some refactor might be required here - will help 🤗

src/diffusers/pipelines/__init__.py
```diff
@@ -154,7 +154,7 @@
         "CogVideoXFunControlPipeline",
     ]
     _import_structure["cogview3"] = ["CogView3PlusPipeline"]
-    _import_structure["consisid"] = ["ConsisIDPipeline"]
+    _import_structure["cogview4"] = ["CogView4Pipeline"]
```
a-r-r-o-w, line 157: I think ConsisID has been removed by mistake here. Could you undo this change?

src/diffusers/pipelines/__init__.py
```diff
@@ -497,7 +497,6 @@
             CogVideoXVideoToVideoPipeline,
         )
         from .cogview3 import CogView3PlusPipeline
-        from .consisid import ConsisIDPipeline
```
a-r-r-o-w, line 500: Similarly seems to have been removed by mistake here

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
```diff
@@ -0,0 +1,712 @@
...
+    b = base_shift - m * base_seq_len
+    mu = image_seq_len * m + b
+    return mu
+
+
+def time_shift(mu: float, shift_sigma: float, sigmas: torch.Tensor):
+    return math.exp(mu) / (math.exp(mu) + (1 / sigmas - 1) ** shift_sigma)
+
+
+# def retrieve_timesteps(
```
a-r-r-o-w, line 71: If this function is not used, let's remove it completely

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
```diff
@@ -0,0 +1,712 @@
...
+
+                ###########################
+                # Get the corresponding sigma value
+                # 这一部分应该放到schduler中（包括self.sigmas的计算也是）
+                # 最后应该调用self.scheduler.step()，只需要传入当前的t，返回下一步的latents即可
+                sigma = self.sigmas[i]
+                sigma_next = self.sigmas[i + 1]
+                dt = sigma_next - sigma
+
+                latents = latents + dt * noise_pred_guided
```
a-r-r-o-w, line 677: If I understand correctly, this is flow matching sampling. Could we directly utilize the `FlowMatchEulerDiscreteScheduler` here and use `scheduler.step()`?

src/diffusers/schedulers/scheduling_cogview.py
```diff
@@ -0,0 +1,332 @@
...
+import numpy as np
+import torch
+
+from ..configuration_utils import ConfigMixin, register_to_config
+from ..utils.torch_utils import randn_tensor
+from .scheduling_ddim import DDIMSchedulerOutput
+from .scheduling_utils import SchedulerMixin
+
+
+class CogViewScheduler(SchedulerMixin, ConfigMixin):
```
a-r-r-o-w, line 28: I'm having a bit of a confusion 😅

Is CogView4 based on Flow Matching? If so, we could directly use the existing flow matching implementation in diffusers. Will take a look at the original codebase too to try and understand if this scheduler is needed

src/diffusers/schedulers/scheduling_ddim_cogview4.py
```diff
@@ -0,0 +1,519 @@
+# Copyright 2024 Stanford University Team and The HuggingFace Team. All rights reserved.
```
a-r-r-o-w, line 1: Are `scheduling_cogview.py` and `scheduling_ddim_cogview.py` the same? 😅

a-r-r-o-w (commented):
Thanks for the awesome PR @zRzRzRzRzRzRzR @OleehyO! Great work 👍 

Just some minor comments to address, but the PR is looking good. Some minor refactors are needed to the modeling design. Since the day of integration is scheduled to be Feb 6th, I'll try to assist in making some final refactors of the transformer.

src/diffusers/pipelines/__init__.py
zRzRzRzRzRzRzR: put back here now

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
OleehyO: Yes, this part is the flow matching sampling. However, since we are still aligning the transformer forward part, we have not checked whether it can match the corresponding implementation of megatron. We will check if we can directly reuse the implementation in `FlowMatchEulerDiscreteScheduler` after confirming that this part is correct.

src/diffusers/schedulers/scheduling_ddim_cogview4.py
OleehyO: Yes, CogView4 is based on Flow Matching, but the implementation of this part has not been finalized yet, so the code from the ddim part was used first, and changes will be made later. After it is finalized, if we can reuse `FlowMatchEulerDiscreteScheduler`, we will remove this file.

src/diffusers/models/attention_processor.py
zRzRzRzRzRzRzR: We are still conducting tests because we cannot handle the tensor at every step, and we now need to correct the issue of generating all noise.
We tried to fix the text embed (a fixed embed) and the Image Rope.
Accuracy of VAE algorithm reconstruction.
The correctness of weight conversion, with an error of 0.
+ Correctness of the text_embed algorithm.
The correctness of the image_rope algorithm.
+ The correctness of sigma (this part is directly copied from the Megatron implementation)

We cannot correctly verify the alignment of tensors from the text_embed step to each subsequent step. Because their errors are all very large.
This part may require some communication with us tomorrow (February 6th)
We have not yet solved the problem of pure noise generated by the images in the diffusers library.

src/diffusers/schedulers/scheduling_cogview.py
zRzRzRzRzRzRzR: This place was just synchronized with my training colleagues, some changes need to be made, and we will strive to modify this by tomorrow.

src/diffusers/schedulers/scheduling_ddim_cogview4.py
zRzRzRzRzRzRzR: Same here

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
zRzRzRzRzRzRzR: `time_shift` is different; we will make the modification tomorrow, and it is different from `FlowMatchEulerDiscreteScheduler`. We will make the modification tomorrow; the colleagues responsible for training have optimized other mathematical algorithms, and this needs to be changed.

zRzRzRzRzRzRzR: This algorithm formula has been changed.

zRzRzRzRzRzRzR: The `time_shift` and `calculate_shift` have been changed now

a-r-r-o-w (approved):
The PR looks good to merge from the modeling and pipeline perspective. For scheduler review, will defer to @yiyixuxu. I think we don't need a specific scheduler for CogView4 from a quick look but YiYi has a better understanding of it.

<table align=center>
<tr>
<td align=center></td>
</tr>
</table>

The outputs don't match to the values (numerically. they still match visually) before I took over with the latest changes. The differences are because:
- RoPE in float32
- latents are always in float32 and casted to transformer dtype before forward pass. Previously, it was bfloat16 (latent creation + transformer forward) -> float32 (scheduler step) -> bfloat16 again (transformer forward)
- Changed how multi_modulate and multi_gate worked. This should practically have no difference in outputs, but seems like there's a difference on order of `1e-4` due to not performing the layer norm on concatenated `encoder_hidden_states` and `hidden_states`.

The reason for two forward passes in the pipeline is because the conditioning embeds and unconditioning embeds can have different sequence lengths.

The model for testing is available here (private): https://huggingface.co/ZP2HF/CogView4-6B-0125. This has the weights in the same format as the model that will be released next week. BUT, it is not the final weights based on my conversation with Yuxuan and an improved version will be released.

src/diffusers/schedulers/scheduling_cogview.py
yiyixuxu: this is still not resolved, no?

yiyixuxu (commented):
looks good to me! I think we still need to see if we can use existing scheduler though

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
```diff
@@ -98,13 +98,17 @@ def __init__(
         use_karras_sigmas: Optional[bool] = False,
         use_exponential_sigmas: Optional[bool] = False,
         use_beta_sigmas: Optional[bool] = False,
+        time_shift_type: str = "exponential",
```
a-r-r-o-w, line 101: We need this new parameter because CogView4 performs timestep shifting without first exponentiating

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
```diff
@@ -0,0 +1,661 @@
...
+        image_seq_len = ((height // self.vae_scale_factor) * (width // self.vae_scale_factor)) // (
+            self.transformer.config.patch_size**2
+        )
+
+        timesteps = (
+            np.linspace(self.scheduler.config.num_train_timesteps, 1.0, num_inference_steps)
+            if timesteps is None
+            else np.array(timesteps)
+        )
+        timesteps = timesteps.astype(np.int64)
```
a-r-r-o-w, line 578: This is a bit different from what we usually do. We don't use `self.scheduler.timesteps` returned from the call to `retrieve_timesteps` because those timesteps are from after resolution-based timestep shifting is applied. For CogView4, it seems like we need to use the timesteps from before applying shifting, but sigmas from after applying shifting.

scripts/convert_cogview4_to_diffusers_megatron.py
```diff
@@ -0,0 +1,366 @@
...
+        cache_dir=args.text_encoder_cache_dir,
+        torch_dtype=torch.bfloat16 if args.dtype == "bf16" else torch.float32,
+    )
+    for param in text_encoder.parameters():
+        param.data = param.data.contiguous()
+
+    # Initialize the scheduler
+    scheduler = FlowMatchEulerDiscreteScheduler(
+        base_shift=0.25, max_shift=0.75, base_image_seq_len=256, use_dynamic_shifting=True, time_shift_type="linear"
+    )
```
a-r-r-o-w, line 345: @zRzRzRzRzRzRzR The converted checkpoints will have to also update the scheduler config

scripts/convert_cogview4_to_diffusers.py
```diff
@@ -0,0 +1,243 @@
...
+        cache_dir=args.text_encoder_cache_dir,
+        torch_dtype=torch.bfloat16 if args.dtype == "bf16" else torch.float32,
+    )
+
+    for param in text_encoder.parameters():
+        param.data = param.data.contiguous()
+
+    scheduler = FlowMatchEulerDiscreteScheduler(
+        base_shift=0.25, max_shift=0.75, base_image_seq_len=256, use_dynamic_shifting=True, time_shift_type="linear"
+    )
```
a-r-r-o-w, line 227: @zRzRzRzRzRzRzR Same comment as above

a-r-r-o-w (approved):
@yiyixuxu I think we should be good to merge. Could you review the scheduler related changes? I've mentioned the relevant parts

With these changes, we match the implementation before I took over with the latest updates. Outputs are identical

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
a-r-r-o-w: cc @hlky here too maybe if you think there's a better way to handle this

hlky: This is ok IMO, will let @yiyixuxu comment as well. We'll have 3 types of shifting (non-dynamic, dynamic with exponential shift, dynamic with linear shift) after this, so it can be looked at with the scheduler refactor.

scripts/convert_cogview4_to_diffusers.py
zRzRzRzRzRzRzR: Based on the current algorithm comparison and the images produced in practice, this change seems to be functioning properly.

src/diffusers/models/embeddings.py
```diff
@@ -2611,3 +2611,78 @@ def forward(self, image_embeds: List[torch.Tensor]):
...
+
+        emb = torch.cat([h_emb, w_emb], dim=-1) * mask
+
+        assert emb.ndim == 2, f"expected emb to have 2 dimensions, got {emb.ndim}"
+        if not self.rotary_interleaved:
+            emb = torch.repeat_interleave(emb, 2, dim=0)
+        else:
+            emb = torch.repeat_interleave(emb, 2, dim=1)
+
+        return emb
```
yiyixuxu, line 2688:
```suggestion
```
I don't think we used this anywhere,no?

a-r-r-o-w: 🤔 🤔 🤔 

I remember removing this 🤔

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
```diff
@@ -0,0 +1,661 @@
...
+        self.register_modules(
+            tokenizer=tokenizer, text_encoder=text_encoder, vae=vae, transformer=transformer, scheduler=scheduler
+        )
+        self.vae_scale_factor = 2 ** (len(self.vae.config.block_out_channels) - 1) if getattr(self, "vae", None) else 8
+        self.image_processor = VaeImageProcessor(vae_scale_factor=self.vae_scale_factor)
+
+    def _get_glm_embeds(
+        self,
+        prompt: Union[str, List[str]] = None,
+        num_images_per_prompt: int = 1,
```
yiyixuxu, line 180: let's move the logic to duplicate the prompt into `encode_prompt` 
reasoning is that the user can recompute the prompt and then use it to generate multiple images per prompt

I know this is different from the pattern we currently use, but we will refactor them soon

a-r-r-o-w: Makes sense, updated! Also tested that `num_images_per_prompt > 1` works as expected with the latest change

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
yiyixuxu: that's really werid - but if it's on purpose (not a oversight or something), we need support it from `retrieve_timesteps` to accept both custom timesteps and custom sigmas, otherwise I don't think it would function correctly for img2img or training, where you do not start from the first timestep and need to search against self.scheduler.timesteps 
e.g. this function won't work https://github.com/huggingface/diffusers/blob/a0c22997fd45770fffd9b454625e9ab525fa2b16/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L299

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
yiyixuxu: looks good to me!

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
a-r-r-o-w: We don't have access to the original codebase yet, so it will be hard to check if it's an oversight. It is weird that we have to do it this way, but if we don't do it, the final outputs come out with some residual noise.

Also seems like in my latest update I just updated this by mistake to `np.float32` from some testing I was doing. Basically, we want integer timesteps here first, but then need float32 timesteps for our scheduler to not raise an error:

https://github.com/huggingface/diffusers/blob/a0c22997fd45770fffd9b454625e9ab525fa2b16/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L366-L372

So, it will have to be something like `timesteps.astype(np.int64).astype(np.float32)` to be consistent with the behaviour when we started updating the PR and to not error out in our scheduler

yiyixuxu: > it will have to be something like timesteps.astype(np.int64).astype(np.float32)

ok, it seems like custom timesteps might be the way to go because this logic here is just really custom (even if we calculate the timesteps without shifting, we also need to do this round up thing first)
basically, you need to:
1. remove the ValueError about passing sigma and timesteps at the same time 
2. add `timesteps` to `set_timesteps` https://github.com/huggingface/diffusers/blob/a0c22997fd45770fffd9b454625e9ab525fa2b16/src/diffusers/schedulers/scheduling_euler_discrete.py#L323

a-r-r-o-w: I'll do some more testing with a fresh mind in the morning to verify if we need differing timesteps vs sigmas here (to recheck if it is a possible oversight or not). I do think I did everything correctly when I tried earlier today, and as a result we might need to do what you mentioned, but wouldn't hurt to delay a little longer and verify again if it might help save us a bunch of changes.

@zRzRzRzRzRzRzR If it would be possible to share just the scheduler implemention related files with us, it would really help us understand if changes are required. No problem if not :) We can wait for the official release from THUDM and update our implementation

a-r-r-o-w: I think we're going to have to go forward with the update regarding both timesteps and sigmas being provided. Almost anything else I try seems to result in residual noise. I've pushed some changes regarding handling with some additional comments in `.step()` for better understanding. Okay to remove if it's not really required

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
```diff
@@ -249,49 +259,71 @@ def set_timesteps(
...
+                automatically.
         """
         if self.config.use_dynamic_shifting and mu is None:
             raise ValueError(" you have a pass a value for `mu` when `use_dynamic_shifting` is set to be `True`")
 
-        if sigmas is None:
-            timesteps = np.linspace(
-                self._sigma_to_t(self.sigma_max), self._sigma_to_t(self.sigma_min), num_inference_steps
-            )
+        self.num_inference_steps = num_inference_steps
```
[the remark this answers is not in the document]
a-r-r-o-w: Added the checks

yiyixuxu (approved):
left one last comment, look good to me otherwise!

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
joangava: Hello, drear 
Tanks for all you jobs . Am really  appreciated.

nitinmukesh:
Hello @zRzRzRzRzRzRzR 

The code was merged so I tried to run but models do not exists THUDM/CogView4-6B
Any plan to make the model available for testing.

vladmandic:
btw, whats the status of this model? afaik, it was supposed to be released by now...

Issue #10656, opened 2025-01-26
ControlNet union pipeline fails on multi-model


vladmandic:
### Describe the bug

All controlnet types are typically defined inside pipeline as below (example from `StableDiffusionXLControlNetPipeline`):

> controlnet: Union[ControlNetModel, List[ControlNetModel], Tuple[ControlNetModel], MultiControlNetModel],  

however, StableDiffusionXLControlNetUnionPipeline pipeline defines it simply as:

> controlnet: ControlNetUnionModel

which defeats one of the main advantages of union controlnet - to be able to perform multiple guidances using same model.
for reference, controlnetunion was added via pr #10131
any changes to txt2img pipeline should also be mirrored in img2img and inpaint pipelines.

### Reproduction

```py
control1 = ControlNetUnionModel.from_single_file(...)
control2 = ControlNetUnionModel.from_single_file(...)
pipe = StableDiffusionXLControlNetUnionPipeline.from_single_file(..., control=[control1, control2])
```

### Logs

```shell
│    256 │   │   if not isinstance(controlnet, ControlNetUnionModel):                                                                                                                                                                                                                                                                                                                                                         │
│ ❱  257 │   │   │   raise ValueError("Expected `controlnet` to be of type `ControlNetUnionModel`.")                                                                                                                                                                                                                                                                                                                          │
│    258                                                                                                                                                                                                                                                                                                                                                                                                                      │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
ValueError: Expected `controlnet` to be of type `ControlNetUnionModel`.
```

### System Info

diffusers==0.33.0.dev0

### Who can help?

@hlky @yiyixuxu @sayakpaul @DN6

hlky:
The advantage of controlnet union is that we don't need to use multiple controlnet models, no? We just use multiple control conditionings with the same controlnet union model. What is the use case here?

vladmandic:
lets say i want to run openpose and canny. how?
i'm guessing you're saying to specify `control_mode` as list instead of int and everything just works if i send two control_images as input list instead?

issue is that i need to match number of all params - control_image, controlnet_conditioning_scale, start, end, etc.
[...]

PR #10666, opened 2025-01-27
ControlNet Union controlnet_conditioning_scale for multiple control inputs

hlky:
# What does this PR do?

e.g. `'controlnet_conditioning_scale': [0.5, 0.8]`

Removes unneeded check that was copied from other controlnet as controlnet union supports multiple inputs with the same controlnet model.

## Who can review?

@vladmandic

### DIFF
--- a/src/diffusers/pipelines/controlnet/pipeline_controlnet_union_inpaint_sd_xl.py
+++ b/src/diffusers/pipelines/controlnet/pipeline_controlnet_union_inpaint_sd_xl.py
@@ -766,26 +766,6 @@ class StableDiffusionXLControlNetUnionInpaintPipeline(
         else:
             assert False
 
-        # Check `controlnet_conditioning_scale`
-        if (
-            isinstance(self.controlnet, ControlNetModel)
-            or is_compiled
-            and isinstance(self.controlnet._orig_mod, ControlNetModel)
-        ):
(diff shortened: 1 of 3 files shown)


yiyixuxu (approved):
thanks for the fix!

vladmandic:
ok, but this needs to be end-to-end for all args, now it fails with:

> float(i / len(timesteps) < control_guidance_start or (i + 1) / len(timesteps) > control_guidance_end)  
> TypeError: '<' not supported between instances of 'float' and 'list'

hlky:
@john09282922 Yes. As per the title the issue is with multiple controlnet_conditioning_scale values.

hlky:
@john09282922 After further testing we found this is a limitation of ControlNet Union - we are tracking the issue here https://github.com/huggingface/diffusers/issues/10656

There is some experimentation that could allow scale per control type otherwise we are planning on allowing multiple ControlNet Union models in the same manner that other ControlNet pipelines work however this would negate the benefit of ControlNet Union.

A single scale value is supported which should cover most use cases, the primary use case that is not supported would be multiple `control_guidance_start` and `control_guidance_end` as the underlying code uses `controlnet_conditioning_scale` to turn off the ControlNet, again a single value for `control_guidance_start` and `control_guidance_end` is currently supported.

PR #10670, opened 2025-01-28
Fix pipeline dtype unexpected change when using SDXL reference community pipelines in float16 mode

dimitribarbot:
# What does this PR do?

When using one of the SDXL reference pipelines, configuring the pipeline with the `torch_dtype` parameter set to `torch.float16` and using a VAE with the `force_upcast` configuration, an upcast of the VAE to `torch.float32` is performed during inference. However, it is never downcast to `torch.float16` which leads to side effects.

Indeed, the dtype of the pipeline is defined in `pipelines/pipeline_utils.py` as:
```
@property
def dtype(self) -> torch.dtype:
    r"""
    Returns:
        `torch.dtype`: The torch dtype on which the pipeline is located.
    """
    module_names, _ = self._get_signature_keys(self)
    modules = [getattr(self, n, None) for n in module_names]
    modules = [m for m in modules if isinstance(m, torch.nn.Module)]

    for module in modules:
        return module.dtype

    return torch.float32
```
When the VAE comes first, the entire pipeline dtype is no longer set to float16 but to float32.

In my case, it causes the following error when inferring the unet and using an IP adapter at the same time:
```
RuntimeError: mat1 and mat2 must have the same dtype, but got Float and Half
```

This PR aims to address this issue by downcasting the VAE to `float16` after computing the reference image latents.

Additionally, I'd like to mention that `module_names` in the above `dtype` property does not seam to be retrieved deterministically, despite using the same code to instantiate my pipeline. For instance, sometimes with the controlnet reference pipeline I have:
```
{'tokenizer_2', 'tokenizer', 'text_encoder_2', 'unet', 'controlnet', 'scheduler', 'vae', 'feature_extractor', 'text_encoder', 'image_encoder'}
```
And sometimes:
```
{'feature_extractor', 'scheduler', 'vae', 'tokenizer', 'unet', 'text_encoder_2', 'tokenizer_2', 'text_encoder', 'image_encoder', 'controlnet'}
```
In the first case, the pipeline `dtype` is not changed to `float32`. In the second case it is.
Bug reproducibility becomes complicated then.

Perhaps something could be done in this part of the code to remove this non-deterministic part?

## Before submitting

## Who can review?

### DIFF
--- a/examples/community/stable_diffusion_xl_controlnet_reference.py
+++ b/examples/community/stable_diffusion_xl_controlnet_reference.py
@@ -193,7 +193,8 @@ class StableDiffusionXLControlNetReferencePipeline(StableDiffusionXLControlNetPi
 
     def prepare_ref_latents(self, refimage, batch_size, dtype, device, generator, do_classifier_free_guidance):
         refimage = refimage.to(device=device)
-        if self.vae.dtype == torch.float16 and self.vae.config.force_upcast:
+        needs_upcasting = self.vae.dtype == torch.float16 and self.vae.config.force_upcast
+        if needs_upcasting:
             self.upcast_vae()
             refimage = refimage.to(next(iter(self.vae.post_quant_conv.parameters())).dtype)
         if refimage.dtype != self.vae.dtype:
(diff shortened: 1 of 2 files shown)


asomoza:
Hi, thanks. For the deterministic problem, if it is in the community pipelines you're more than welcome to open a new PR to fix it, if you find the same problem in the core, you can open an issue about it, but let's keep this PR simple with the vae fix only.

asomoza (approved):

Issue #10671, opened 2025-01-28
Deterministic issue in DiffusionPipeline when getting dtype or device


dimitribarbot:
### Describe the bug

This issue is a follow-up of this [PR](https://github.com/huggingface/diffusers/pull/10670).

The idea was to fix an issue leading to the following error message:
```
RuntimeError: mat1 and mat2 must have the same dtype, but got Float and Half
```

This was due to a dtype mismatch between image input encoded with a VAE using float32 precision and unet configured with float16 precision. 

Although this bug is fixed, I would like to mention how difficult it was to debug since it didn't occur every time. After debugging I found that the `dtype` property of the diffusion pipeline does not return a value deterministically (it seems the same applies to the device). It uses the dtype of the first pipeline component it finds, but the list of pipeline components is not sorted.

I think this may lead to other mistakes in the future and may require changes.

### Reproduction

To reproduce the deterministic issue, you can first modify the `DiffusionPipeline` `dtype` property in the `src/diffusers/pipelines/pipeline_utils.py` file to add some logs:
```
@property
def dtype(self) -> torch.dtype:
    r"""
    Returns:
        `torch.dtype`: The torch dtype on which the pipeline is located.
    """
    module_names, _ = self._get_signature_keys(self)
    modules = [getattr(self, n, None) for n in module_names]
    modules = [m for m in modules if isinstance(m, torch.nn.Module)]

    for module in modules:
        print(f"module.dtype is {module.dtype} using {type(module).__name__} from {module_names}.")  # <--- Add this line
        return module.dtype

    return torch.float32
```
Then create a new python script file with the following content:
```
from diffusers import StableDiffusionXLPipeline
from diffusers.schedulers import UniPCMultistepScheduler
import torch

# initialize the models and pipeline
pipe = StableDiffusionXLPipeline.from_pretrained(
    "stabilityai/stable-diffusion-xl-base-1.0",
    torch_dtype=torch.float16,
    use_safetensors=True,
    variant="fp16"
)
pipe.to(torch.device("cuda"))
pipe.scheduler = UniPCMultistepScheduler.from_config(pipe.scheduler.config)

# generate image
print("Pipeline dtype:", pipe.dtype)
image = pipe(prompt="a dog", num_inference_steps=20).images[0]
```
Run this script multiple times and observe the results (see "Logs" section below).

### Logs

```shell
1st script run gives:
[...]

PR #10696, opened 2025-01-31
Fix deterministic issue when getting pipeline dtype and device

dimitribarbot:
# What does this PR do?

This PR aims to fix a deterministic issue when getting the pipeline dtype or device across multiple runs of the same Python script.

Internally, the ordering issue comes from the use of `set` in the `_get_signature_keys` function in the `src/diffusers/pipelines/pipeline_utils.py` file. As stated in the Python documentation, `set` does not guarantee that the elements will have the same order across multiple runs of a Python script. This is due to the hashing algorithm which is initialized with a random salt each time the python interpreter is started. More information can be found [here](https://docs.python.org/3/reference/datamodel.html#object.__hash__) (especially in the second note at the end of the section).

Fixes [#10671](https://github.com/huggingface/diffusers/issues/10671).

## Before submitting

## Who can review?

@sayakpaul


src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1619,10 +1619,12 @@ def components(self) -> Dict[str, Any]:
             k: getattr(self, k) for k in self.config.keys() if not k.startswith("_") and k not in optional_parameters
         }
 
-        if set(components.keys()) != expected_modules:
+        actual = sorted(set(components.keys()))
+        expected = sorted(expected_modules)
```
dimitribarbot, line 1623: I'm not sure if using `sorted` is necessary here, since `expected_modules` comes from the `_get_signature_keys` call which already calls `sorted` internally. 

However, I leave it that way because it allows to be completely decoupled from the `_get_signature_keys` function. And I don't think it will add a big performance overhead (sorting of n elements, n=~10, elements already sorted).

DN6 (approved):

PR #10704, opened 2025-02-02
Fix Documentation about Image-to-Image Pipeline

### LINKED ISSUE #9566: Documentation about superresolution
### Describe the bug

Reading this [documentation](https://huggingface.co/docs/diffusers/en/using-diffusers/img2img#image-to-upscaler-to-super-resolution) which is not possible.
Indeed, [super resolution pipeline](https://huggingface.co/docs/diffusers/v0.16.0/en/api/pipelines/stable_diffusion/upscale#diffusers.StableDiffusionUpscalePipeline) works with an image as input in the image input parameter.
So it's not possible to use `output_type="latent"` for image_2. 
Otherwise you'll get an error :
ValueError: Incorrect configuration settings! The config of `pipeline.unet` expects 7 but received `num_channels_latents`: 4 + `num_channels_image`: 4  = 8. Please verify the config of `pipeline.unet` or your `image` input.
Because it's placing a latent instead of an image.

Please also note that `"variant="fp16"` does not work for stabilityai/sd-x2-latent-upscaler" for me. The error message I get is : 

```
ValueError: The deprecation tuple ('no variant default', '0.24.0', "You are trying to load the model files of the `variant=fp16`, but no such modeling files are available.The default model files: {'unet/diffusion_pytorch_model.bin', 'text_encoder/pytorch_model.bin', 'text_encoder/model.safetensors', 'vae/diffusion_pytorch_model.safetensors', 'vae/diffusion_pytorch_model.bin', 'unet/diffusion_pytorch_model.safetensors'} will be loaded instead. Make sure to not load from `variant=fp16`if such variant modeling files are not available. Doing so will lead to an error in v0.24.0 as defaulting to non-variantmodeling files is deprecated.") should be removed since diffusers' version 0.30.3 is >= 0.24.0
```

### Reproduction

simple snippet of code

### Logs

_No response_

### System Info

ubuntu

### Who can help?

@yiyixuxu @asomoza

ParagEkbote:
# What does this PR do?

As previously discussed, I have updated the documentation to ensure it works correctly. Please let me know if any improvements are needed and I will make the necessary changes.

Fixes #9566 

## Before submitting

## Who can review?

@stevhliu

### DIFF
--- a/docs/source/en/using-diffusers/img2img.md
+++ b/docs/source/en/using-diffusers/img2img.md
@@ -461,12 +461,12 @@ Chain it to an upscaler pipeline to increase the image resolution:
 from diffusers import StableDiffusionLatentUpscalePipeline
 
 upscaler = StableDiffusionLatentUpscalePipeline.from_pretrained(
-    "stabilityai/sd-x2-latent-upscaler", torch_dtype=torch.float16, variant="fp16", use_safetensors=True
+    "stabilityai/sd-x2-latent-upscaler", torch_dtype=torch.float16, use_safetensors=True
 )
 upscaler.enable_model_cpu_offload()
 upscaler.enable_xformers_memory_efficient_attention()
 
-image_2 = upscaler(prompt, image=image_1, output_type="latent").images[0]
+image_2 = upscaler(prompt, image=image_1).images[0]
 ```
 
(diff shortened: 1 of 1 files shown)


stevhliu (approved):
Thanks for the fix! 👍

Issue #10689, opened 2025-01-30
Support IPAdapter for all Flux pipelines- not only for txt2img


almog2065:
I see diffusers recently merged the IPAdapter for Flux pipelines, but only for txt2img pipeline. https://github.com/huggingface/diffusers/issues/9825
The feature request is about supporting IPAdapter to all Flux pipelines as img2img, sketch2img and more..

It will be great to have load_ip_adapter and unload_ip_adapter for all those pipelines:
(FluxControlNetPipeline, FluxImg2ImgPipeline, FluxInpaintPipeline)

For now this support works only for FluxPipeline(txt2img only)

I see the feature can be implemented by adding the FluxIPAdapterMixin to the pipeline class.

asomoza:
cc: @hlky

hlky:
Hi @almog2065. We are planning to add IPAdapter in other Flux pipelines, we can add this issue to the Roadmap for next release. If you'd like to contribute a PR for any of the Flux pipelines we are happy to assist, each needs similar [pipeline changes](https://github.com/huggingface/diffusers/pull/10261/files#diff-68d4ddab9d9846465f7d3b0a8583b256f7293df93e60a04eaeaced3b85de15aa) and changes for the [pipeline's tests](https://github.com/huggingface/diffusers/pull/10261/files#diff-0e5bda7c51cb41ae4ac58e1580b7b56fb56fc0437e970bc0ab89a932838b090c) - we don't necessarily need slow tests for all pipelines though it can be covered by `FluxIPAdapterTesterMixin`.

SahilCarterr:
PR #10708

hlky:
IPAdapter support has been added in the other most common pipelines: `flux_img2img`, `flux_inpaint`, `flux_controlnet`. We will consider this issue resolved so it can be moved to complete on the Roadmap for 0.33 release. Requests for other pipelines are welcome in a new issue.

PR #10708, opened 2025-02-03
Support IPAdapter for more Flux pipelines

hlky:
# What does this PR do?

Adds IPAdapter support to

- `flux_img2img`
- `flux_inpaint`
- `flux_controlnet`

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #10717, opened 2025-02-04
IP-Adapter for `FluxImg2ImgPipeline`

guiyrt:
# What does this PR do?

Adds image prompting support for `FluxImg2ImgPipeline` via `FluxIPAdapterMixin`, as part of #10689.

Example output:
astro_grid

I did some minor changes on `pipeline_flux.py` regarding documentation and argument order, for consistency.
I also adapted a bit the ip_adapter code for clarity, hope you find it simpler this way :)
I'm not sure if you prefer it this way, but I made a change on `FluxIPAdapterPipelineSlowTests` so that the rest of the pipelines can use it simply by adding input arguments and changing expected output. Let me know if subclassing like this is ok, or if you prefer having the full class definition for each pipeline test.

<details>
<summary>Inference code</summary>

```python
import torch

from diffusers import FluxImg2ImgPipeline

image_encoder_id = "openai/clip-vit-large-patch14"
ip_adapter_id = "XLabs-AI/flux-ip-adapter"

pipe = FluxImg2ImgPipeline.from_pretrained(
    model_id,
    torch_dtype=torch.bfloat16
)

# Load IP Adapter
pipe.load_ip_adapter(
    ip_adapter_id,
    weight_name="ip_adapter.safetensors",
    image_encoder_pretrained_model_name_or_path=image_encoder_id
)
pipe.set_ip_adapter_scale(1.0)

reference_img = load_image("astronaut.jpg")
ip_adapter_img = load_image("../reference_images/abstract.png")

image = pipe(
    image=reference_img,
    width=1024,
    height=1024,
    ip_adapter_image=ip_adapter_img,
    guidance_scale=15,
    num_inference_steps=40,
    strength=0.85,
    prompt="an astronaut in space"
).images[0]

image.save("result.jpg")
```

</details>

## Before submitting

## Who can review?
@hlky @yiyixuxu

### DIFF
--- a/src/diffusers/pipelines/flux/pipeline_flux_img2img.py
+++ b/src/diffusers/pipelines/flux/pipeline_flux_img2img.py
@@ -17,10 +17,17 @@ from typing import Any, Callable, Dict, List, Optional, Union
 
 import numpy as np
 import torch
-from transformers import CLIPTextModel, CLIPTokenizer, T5EncoderModel, T5TokenizerFast
+from transformers import (
+    CLIPImageProcessor,
+    CLIPTextModel,
+    CLIPTokenizer,
+    CLIPVisionModelWithProjection,
+    T5EncoderModel,
+    T5TokenizerFast,
+)
 
 from ...image_processor import PipelineImageInput, VaeImageProcessor
-from ...loaders import FluxLoraLoaderMixin, FromSingleFileMixin, TextualInversionLoaderMixin
+from ...loaders import FluxIPAdapterMixin, FluxLoraLoaderMixin, FromSingleFileMixin, TextualInversionLoaderMixin
 from ...models.autoencoders import AutoencoderKL
 from ...models.transformers import FluxTransformer2DModel
 from ...schedulers import FlowMatchEulerDiscreteScheduler
@@ -159,7 +166,7 @@ def retrieve_timesteps(
     return timesteps, num_inference_steps
 
 
-class FluxImg2ImgPipeline(DiffusionPipeline, FluxLoraLoaderMixin, FromSingleFileMixin):
+class FluxImg2ImgPipeline(DiffusionPipeline, FluxLoraLoaderMixin, FromSingleFileMixin, FluxIPAdapterMixin):
     r"""
     The Flux pipeline for image inpainting.
 
@@ -184,10 +191,14 @@ class FluxImg2ImgPipeline(DiffusionPipeline, FluxLoraLoaderMixin, FromSingleFile
         tokenizer_2 (`T5TokenizerFast`):
             Second Tokenizer of class
             [T5TokenizerFast](https://huggingface.co/docs/transformers/en/model_doc/t5#transformers.T5TokenizerFast).
+        image_encoder (`CLIPImageProcessor`, *optional*):
+            Pre-trained Vision Model for IP Adapter.
+        feature_extractor (`CLIPVisionModelWithProjection`, *optional*):
+            Image processor for IP Adapter.
     """
 
-    model_cpu_offload_seq = "text_encoder->text_encoder_2->transformer->vae"
-    _optional_components = []
+    model_cpu_offload_seq = "text_encoder->text_encoder_2->image_encoder->transformer->vae"
+    _optional_components = ["image_encoder", "feature_extractor"]
     _callback_tensor_inputs = ["latents", "prompt_embeds"]
 
     def __init__(
@@ -199,6 +210,8 @@ class FluxImg2ImgPipeline(DiffusionPipeline, FluxLoraLoaderMixin, FromSingleFile
         text_encoder_2: T5EncoderModel,
         tokenizer_2: T5TokenizerFast,
         transformer: FluxTransformer2DModel,
+        image_encoder: CLIPVisionModelWithProjection = None,
+        feature_extractor: CLIPImageProcessor = None,
     ):
         super().__init__()
 
@@ -210,6 +223,8 @@ class FluxImg2ImgPipeline(DiffusionPipeline, FluxLoraLoaderMixin, FromSingleFile
             tokenizer_2=tokenizer_2,
             transformer=transformer,
             scheduler=scheduler,
+            image_encoder=image_encoder,
+            feature_extractor=feature_extractor,
         )
         self.vae_scale_factor = 2 ** (len(self.vae.config.block_out_channels) - 1) if getattr(self, "vae", None) else 8
         # Flux latents are turned into 2x2 patches and packed. This means the latent width and height has to be divisible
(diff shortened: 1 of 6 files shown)


guiyrt:
While testing I also found that if you load the IP-Adapter, you must provide image prompt otherwise the pipeline fails. I will open a PR for that, but it's just a check for `ip_hidden_states is None` in `FluxIPAdapterJointAttnProcessor2_0`.

src/diffusers/pipelines/flux/pipeline_flux_img2img.py
```diff
@@ -791,12 +938,48 @@ def __call__(
...
+        negative_image_embeds = (
+            self.prepare_ip_adapter_image_embeds(
+                negative_ip_adapter_image,
+                negative_ip_adapter_image_embeds,
+                device,
+                batch_size * num_images_per_prompt,
+            )
+            if any_negative_ip_adapter_input
+            else None
+        )
```
hlky, line 972: Why is this changed, we can just copy from text-to-image pipeline, no?

tests/pipelines/flux/test_pipeline_flux_img2img.py
```diff
@@ -161,3 +164,53 @@ def test_flux_image_output_shape(self):
             image = pipe(**inputs).images[0]
             output_height, output_width, _ = image.shape
             assert (output_height, output_width) == (expected_height, expected_width)
+
+
+class FluxIPAdapterImg2ImgPipelineSlowTests(FluxIPAdapterPipelineSlowTests):
+    """Same test as in FluxIPAdapterPipelineSlowTests, only with inital `image` and `strength` parameters."""
```
hlky, line 170: I don't think we need slow tests for IP-Adapter in each pipeline variant, the general functionality is tested with the fast test `FluxIPAdapterTesterMixin`.

hlky (commented):
Thanks @guiyrt. I had opened #10708 to cover FluxImg2Img and others.

> While testing I also found that if you load the IP-Adapter, you must provide image prompt otherwise the pipeline fails.

Is this the same behavior as other IP-Adapter? I believe it may be expected that we call `unload_ip_adapter` if we no longer want to use `ip_adapter_image`.

guiyrt:
> Thanks @guiyrt. I had opened #10708 to cover FluxImg2Img and others.

Oh, I started this yesterday and didn't notice :cry:  bad timing, we can close this then

> > While testing I also found that if you load the IP-Adapter, you must provide image prompt otherwise the pipeline fails.
> 
> Is this the same behavior as other IP-Adapter? I believe it may be expected that we call `unload_ip_adapter` if we no longer want to use `ip_adapter_image`.

In SD3 you can load the IP-Adapter and not pass an image prompt, and it behaves like the regular SD3 pipeline. This can be useful when you don't control the pipeline args, such as in a gradio demo or just general model serving

guiyrt:
Thanks for still taking the time to review @hlky, is there anything else with priority open for the community? I can also check the roadmap, but if you have something in mind let me know :)

hlky:
Would you like to try `MultiControlNet`-like support in ControlNet Union as a companion to https://github.com/huggingface/diffusers/pull/10723? See https://github.com/huggingface/diffusers/issues/10656 for context.

We need to restore all the `MultiControlNet` related code using `StableDiffusionXLControlNetPipeline` as a reference.

https://github.com/huggingface/diffusers/blob/5b1dcd15848f6748c6cec978ef962db391c4e4cd/src/diffusers/pipelines/controlnet/pipeline_controlnet_sd_xl.py#L184

I'll handle integrating `MultiControlNet` and the experimental per control type scale support either in that PR or yours, depending on the merge order. For example this will be either `len(control_mode)` or `len(controlnet.nets)`.

https://github.com/huggingface/diffusers/blob/0914fa3901b464488442b37887f69610361a7b3f/src/diffusers/pipelines/controlnet/pipeline_controlnet_union_sd_xl.py#L1144

guiyrt:
> Would you like to try `MultiControlNet`-like support in ControlNet Union as a companion to #10723? See #10656 for context.

I'll get on it 🚀

guiyrt:
@hlky If I got it right, the idea is to make `MultiControlNetModel` accept `ControlNetUnionModel`, and change `StableDiffusionXLControlNetUnionPipeline` to have the controlnet property as `MultiControlNetModel`, so the interface is consistent, is this it?

I noticed the PR you have open regarding the experiments with per control type scale, but from your previous message I didn't get if you already started with the integration of `ControlNetUnionModel` into `MultiControlNetModel` and want me to handle changes to `StableDiffusionXLControlNetUnionPipeline`, or if my PR would include both.  Just wanted to clarify to avoid duplicated work :)

I didn't know about ControlNet union, very interesting work, thanks for pointing me towards this! :)

hlky:
Yes, so `MultiControlNetModel` and the variants like [`FluxMultiControlNetModel`](https://github.com/huggingface/diffusers/blob/464374fb87610c53b2cf81e08d3df628fada3ce4/src/diffusers/models/controlnets/controlnet_flux.py#L399) are designed to support multiple control types. When we pass `controlnet_conditioning_scale` for multiple control types it is handled by `MultiControlNetModel` and an individual scale value is passed to each `ControlNetModel`. When integrating `ControlNetUnion` I updated the original pipelines against the existing XL ControlNet but `ControlNetUnion` natively supports multiple control types so removed code related to `MultiControlNetModel`. However multiple scale values are not supported because we cannot apply multiple scale values in the [usual place](https://github.com/huggingface/diffusers/issues/10656#issuecomment-2616973573). My experimental PR applies the multiple scale values at a different location. We'd still like to support the `MultiControlNet` case for consistency.

In this case we need a `MultiControlNetUnionModel` as `ControlNetUnion` like `FluxControlNetModel` has a different interface than `ControlNetModel`.

`StableDiffusionXLControlNetUnionPipeline` (and other `ControlNetUnion` pipelines) will have the `controlnet` property like:

```python
controlnet: Union[
            ControlNetUnionModel, List[ControlNetUnionModel], Tuple[ControlNetUnionModel], MultiControlNetUnionModel
        ],
```

Essentially compare to [`StableDiffusionXLControlNetPipeline`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/controlnet/pipeline_controlnet_sd_xl.py) and add any `MultiControlNetModel` code but use `MultiControlNetUnionModel` and `ControlNetUnionModel` in place of `MultiControlNetModel` and `ControlNetModel`.

guiyrt:
> In this case we need a MultiControlNetUnionModel as ControlNetUnion like FluxControlNetModel has a different interface than ControlNetModel.

Yeah, this was my main concern if we tried to just glue together `ControlNetUnion ` into `MultiControlNetModel`.

Got it, much clearer now, thanks!

PR #10726, opened 2025-02-05
[modular] Refactor pipeline functions

hlky:
# What does this PR do?

https://github.com/huggingface/diffusers/pull/9672#pullrequestreview-2592804232

## Who can review?

@yiyixuxu

### DIFF
--- a/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_modular.py
+++ b/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_modular.py
@@ -221,15 +221,24 @@ class StableDiffusionXLIPAdapterStep(PipelineBlock):
             ip_adapter_image=data.ip_adapter_image,
             ip_adapter_image_embeds=None,
             device=data.device,
-            num_images_per_prompt=1,
-            do_classifier_free_guidance=data.do_classifier_free_guidance,
         )
+        
         if data.do_classifier_free_guidance:
-            data.negative_ip_adapter_embeds = []
-            for i, image_embeds in enumerate(data.ip_adapter_embeds):
-                negative_image_embeds, image_embeds = image_embeds.chunk(2)
-                data.negative_ip_adapter_embeds.append(negative_image_embeds)
-                data.ip_adapter_embeds[i] = image_embeds
+            output_hidden_states = [not isinstance(image_proj_layer, ImageProjection) for image_proj_layer in pipeline.unet.encoder_hid_proj.image_projection_layers]
+            negative_ip_adapter_embeds = []
+            for (idx, output_hidden_state), ip_adapter_embeds in zip(enumerate(output_hidden_states), data.ip_adapter_embeds):
+                if not output_hidden_state:
+                    negative_ip_adapter_embed = torch.zeros_like(ip_adapter_embeds)
+                else:
+                    ip_adapter_image = data.ip_adapter_image[idx] if isinstance(data.ip_adapter_image, list) else data.ip_adapter_image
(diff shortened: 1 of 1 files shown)


src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_modular.py
```diff
@@ -3254,20 +3261,13 @@ def prepare_control_image(
 
         return image
 
-    # Copied from diffusers.pipelines.stable_diffusion_xl.pipeline_stable_diffusion_xl.StableDiffusionXLPipeline.encode_prompt
```
yiyixuxu, line 3257: I was thinking that we could have a new `_encode_prompt` method ( or maybe better to have a public method for this, so use a different name),  and refactor the current `encode_prompt` to use that new method,  what do you think?

hlky: Yeah that works and I agree a public method is better for this, how about `encode_single_prompt`?

yiyixuxu: sounds good! and we can discuss with the team and change it later (it will be under this PR for now but I think we should change that for regular pipeline too)

hlky: I've added `encode_single_prompt` and `encode_prompt`.

src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_modular.py
```diff
@@ -3270,6 +3266,48 @@ def encode_prompt(
...
+                negative_prompt_2,
+                device,
+                prompt_embeds=negative_prompt_embeds,
+                pooled_prompt_embeds=negative_pooled_prompt_embeds,
+                lora_scale=lora_scale,
+                clip_skip=clip_skip,
+            )
+        return prompt_embeds, negative_prompt_embeds, pooled_prompt_embeds, negative_pooled_prompt_embeds
+
+    def encode_single_prompt(
```
yiyixuxu, line 3302: Nice! do you think we can maybe make this a class method while we are at this? 
you will need to pass the components, and lora_scale will not be used if it is None
i.e. 
```python
def encode_single_prompt(text_encoder, prompt, device, lora_scale, clip_skip)
```

(have this in mind that we want to potentially apply this pattern across all pipeline in the future, so maybe look over other encode_prompts to see if this makes seense)

hlky: Would we pass `text_encoder`, `text_encoder_2` etc. or would we call separately with each and have an additional method to handle concatenation?

yiyixuxu: maybe make a `encde_single_prompt_clip` that can take a list of text_encoders/tokenizers? we're already doing that for sd3/flux https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py#L288

I think class method would add convenience when using with our regular pipeline (i.e, we now support creating a pipeline without unet & vae very well but it's better not having to create one to begin with). But I think it does not matter much with modular diffusers, so we don't have to make class methods too. 

all are just thoughts here, I think it depends on the use case we want to support here (e.g. long prompt is an example, or any other customization users want to play with prompts). so let me know what you think

PR #10747, opened 2025-02-07
`MultiControlNetUnionModel` on SDXL

guiyrt:
# What does this PR do?

New `MultiControlNetUnionModel` wrapper class to handle multiple `ControlNetUnionModel`s, similarly to `MultiControlNetModel`. Addressed in #10656 to control start, end and scale of each condition image.

<details>
<summary>Input</summary>

<table>
<thead>
<tr>
<th>Segmentation</th>
<th>Pose</th>
</tr>
</thead>
<tbody>
<tr>
<td>

 </td>
<td>

</td>
</tr>
</tbody>
</table>
</details>

<details>
<summary>Inference code</summary>

```python
import torch

from diffusers import StableDiffusionXLControlNetUnionPipeline
from diffusers.models import ControlNetUnionModel, AutoencoderKL

model_id = "stabilityai/stable-diffusion-xl-base-1.0"
controlnet_id = "brad-twinkl/controlnet-union-sdxl-1.0-promax"

controlnet = ControlNetUnionModel.from_pretrained(
    "brad-twinkl/controlnet-union-sdxl-1.0-promax", torch_dtype=torch.float16
)

pipe = StableDiffusionXLControlNetUnionPipeline.from_pretrained(
    controlnet=[controlnet, controlnet],
    vae = AutoencoderKL.from_pretrained(
        "madebyollin/sdxl-vae-fp16-fix", torch_dtype=torch.float16
    ),
    torch_dtype=torch.float16,
    variant="fp16",
)

room_seg_img = load_image("https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/sd_controlnet/room_seg.png")
pose_img = load_image("https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/sd_controlnet/pose.png")


image = pipe(
    prompt="an astronaut in space",
    width=1024,
    height=1024,
    guidance_scale=5,
    num_inference_steps=50,
    control_image=[[pose_img], [room_seg_img]],
    control_mode=[[0], [5]]
).images[0]

image.save("result.jpg")
```
</details>

First, I ran the pipeline as before, using a single `ControlNetUnionModel` with pose, segmentation and pose+segmentation conditions, to have outputs to compare with.

### `ControlNetUnionModel`
<table>
<thead>
<tr>
<th>Segmentation</th>
<th>Pose</th>
<th>Segmentation + Pose</th>
</tr>
</thead>
<tbody>
<tr>
<td>

 </td>
<td>

</td>
<td>

</td>
</tr>
</tbody>
</table>

### `MultiControlNetUnionModel`
Two instances of `ControlNetUnionModel`, one got the segmentation conditioning and the other the pose. To compare the output, I set `controlnet_conditioning_scale` to [0.0, 1.0] and [0.0, 1.0] to compare with the output of single conditioning using `ControlNetUnionModel` above. As we see (and expect), these outputs are the same. The output is different when using both segmentation and pose conditioning, which I think is expected.

<table>
<thead>
<tr>
<th>Segmentation</th>
<th>Pose</th>
<th>Segmentation + Pose</th>
</tr>
</thead>
<tbody>
<tr>
<td>

 </td>
<td>

</td>
<td>

</td>
</tr>
</tbody>
</table>

## Before submitting

## Who can review?
@hlky @yiyixuxu @vladmandic @asomoza

### DIFF
--- a/src/diffusers/models/controlnets/multicontrolnet_union.py
+++ b/src/diffusers/models/controlnets/multicontrolnet_union.py
@@ -119,6 +119,7 @@ class MultiControlNetUnionModel(ModelMixin):
             )
 
     @classmethod
+    # Copied from diffusers.models.controlnets.multicontrolnet.MultiControlNetModel.from_pretrained with MultiControlNetModel->MultiControlNetUnionModel
     def from_pretrained(cls, pretrained_model_path: Optional[Union[str, os.PathLike]], **kwargs):
         r"""
         Instantiate a pretrained MultiControlNetUnion model from multiple pre-trained controlnet models.
--- a/src/diffusers/models/controlnets/multicontrolnet_union.py
+++ b/src/diffusers/models/controlnets/multicontrolnet_union.py
@@ -79,7 +79,7 @@ class MultiControlNetUnionModel(ModelMixin):
 
         return down_block_res_samples, mid_block_res_sample
 
-    # Copied from diffusers.models.controlnets.multicontrolnet.MultiControlNetModel.save_pretrained with MultiControlNetModel->MultiControlNetUnionModel
+    # Copied from diffusers.models.controlnets.multicontrolnet.MultiControlNetModel.save_pretrained with ControlNet->ControlNetUnion
     def save_pretrained(
         self,
         save_directory: Union[str, os.PathLike],
@@ -119,7 +119,7 @@ class MultiControlNetUnionModel(ModelMixin):
             )
 
     @classmethod
-    # Copied from diffusers.models.controlnets.multicontrolnet.MultiControlNetModel.from_pretrained with MultiControlNetModel->MultiControlNetUnionModel
+    # Copied from diffusers.models.controlnets.multicontrolnet.MultiControlNetModel.from_pretrained with ControlNet->ControlNetUnion
     def from_pretrained(cls, pretrained_model_path: Optional[Union[str, os.PathLike]], **kwargs):
         r"""
         Instantiate a pretrained MultiControlNetUnion model from multiple pre-trained controlnet models.
@@ -186,7 +186,7 @@ class MultiControlNetUnionModel(ModelMixin):
 
         if len(controlnets) == 0:
             raise ValueError(
-                f"No ControlNets found under {os.path.dirname(pretrained_model_path)}. Expected at least {pretrained_model_path + '_0'}."
+                f"No ControlNetUnions found under {os.path.dirname(pretrained_model_path)}. Expected at least {pretrained_model_path + '_0'}."
             )
 
         return cls(controlnets)
--- a/src/diffusers/models/controlnets/multicontrolnet_union.py
+++ b/src/diffusers/models/controlnets/multicontrolnet_union.py
@@ -79,6 +79,7 @@ class MultiControlNetUnionModel(ModelMixin):
 
         return down_block_res_samples, mid_block_res_sample
 
+    # Copied from diffusers.models.controlnets.multicontrolnet.MultiControlNetModel.save_pretrained with MultiControlNetModel->MultiControlNetUnionModel
     def save_pretrained(
         self,
         save_directory: Union[str, os.PathLike],
--- a/src/diffusers/models/controlnets/multicontrolnet_union.py
+++ b/src/diffusers/models/controlnets/multicontrolnet_union.py
@@ -1,0 +1,190 @@
+import os
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+import torch
+from torch import nn
+
+from ...models.controlnets.controlnet import ControlNetOutput
+from ...models.controlnets.controlnet_union import ControlNetUnionModel
+from ...models.modeling_utils import ModelMixin
+from ...utils import logging
+
+
+logger = logging.get_logger(__name__)
+
+
+class MultiControlNetUnionModel(ModelMixin):
+    r"""
+    Multiple `ControlNetUnionModel` wrapper class for Multi-ControlNet-Union.
+
+    This module is a wrapper for multiple instances of the `ControlNetUnionModel`. The `forward()` API is designed to
+    be compatible with `ControlNetUnionModel`.
+
+    Args:
+        controlnets (`List[ControlNetUnionModel]`):
+            Provides additional conditioning to the unet during the denoising process. You must set multiple
+            `ControlNetUnionModel` as a list.
+    """
+
+    def __init__(self, controlnets: Union[List[ControlNetUnionModel], Tuple[ControlNetUnionModel]]):
+        super().__init__()
+        self.nets = nn.ModuleList(controlnets)
+
+    def forward(
+        self,
+        sample: torch.Tensor,
+        timestep: Union[torch.Tensor, float, int],
+        encoder_hidden_states: torch.Tensor,
+        controlnet_cond: List[torch.tensor],
+        control_type: List[torch.Tensor],
+        control_type_idx: List[List[int]],
+        conditioning_scale: List[float],
+        class_labels: Optional[torch.Tensor] = None,
+        timestep_cond: Optional[torch.Tensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
+        added_cond_kwargs: Optional[Dict[str, torch.Tensor]] = None,
+        cross_attention_kwargs: Optional[Dict[str, Any]] = None,
+        guess_mode: bool = False,
+        return_dict: bool = True,
+    ) -> Union[ControlNetOutput, Tuple]:
+        for i, (image, ctype, ctype_idx, scale, controlnet) in enumerate(
+            zip(controlnet_cond, control_type, control_type_idx, conditioning_scale, self.nets)
+        ):
+            down_samples, mid_sample = controlnet(
+                sample=sample,
+                timestep=timestep,
+                encoder_hidden_states=encoder_hidden_states,
+                controlnet_cond=image,
+                control_type=ctype,
+                control_type_idx=ctype_idx,
+                conditioning_scale=scale,
+                class_labels=class_labels,
+                timestep_cond=timestep_cond,
+                attention_mask=attention_mask,
+                added_cond_kwargs=added_cond_kwargs,
+                cross_attention_kwargs=cross_attention_kwargs,
+                guess_mode=guess_mode,
+                return_dict=return_dict,
+            )
+
+            # merge samples
+            if i == 0:
+                down_block_res_samples, mid_block_res_sample = down_samples, mid_sample
+            else:
+                down_block_res_samples = [
+                    samples_prev + samples_curr
+                    for samples_prev, samples_curr in zip(down_block_res_samples, down_samples)
+                ]
+                mid_block_res_sample += mid_sample
+
+        return down_block_res_samples, mid_block_res_sample
+
+    def save_pretrained(
+        self,
+        save_directory: Union[str, os.PathLike],
+        is_main_process: bool = True,
+        save_function: Callable = None,
+        safe_serialization: bool = True,
+        variant: Optional[str] = None,
+    ):
+        """
+        Save a model and its configuration file to a directory, so that it can be re-loaded using the
+        `[`~models.controlnets.multicontrolnet.MultiControlNetUnionModel.from_pretrained`]` class method.
+
+        Arguments:
+            save_directory (`str` or `os.PathLike`):
+                Directory to which to save. Will be created if it doesn't exist.
+            is_main_process (`bool`, *optional*, defaults to `True`):
+                Whether the process calling this is the main process or not. Useful when in distributed training like
+                TPUs and need to call this function on all processes. In this case, set `is_main_process=True` only on
+                the main process to avoid race conditions.
+            save_function (`Callable`):
+                The function to use to save the state dictionary. Useful on distributed training like TPUs when one
+                need to replace `torch.save` by another method. Can be configured with the environment variable
+                `DIFFUSERS_SAVE_MODE`.
+            safe_serialization (`bool`, *optional*, defaults to `True`):
+                Whether to save the model using `safetensors` or the traditional PyTorch way (that uses `pickle`).
+            variant (`str`, *optional*):
+                If specified, weights are saved in the format pytorch_model.<variant>.bin.
+        """
+        for idx, controlnet in enumerate(self.nets):
+            suffix = "" if idx == 0 else f"_{idx}"
+            controlnet.save_pretrained(
+                save_directory + suffix,
+                is_main_process=is_main_process,
+                save_function=save_function,
+                safe_serialization=safe_serialization,
+                variant=variant,
+            )
+
+    @classmethod
+    def from_pretrained(cls, pretrained_model_path: Optional[Union[str, os.PathLike]], **kwargs):
+        r"""
+        Instantiate a pretrained MultiControlNetUnion model from multiple pre-trained controlnet models.
+
+        The model is set in evaluation mode by default using `model.eval()` (Dropout modules are deactivated). To train
+        the model, you should first set it back in training mode with `model.train()`.
+
+        The warning *Weights from XXX not initialized from pretrained model* means that the weights of XXX do not come
+        pretrained with the rest of the model. It is up to you to train those weights with a downstream fine-tuning
+        task.
+
+        The warning *Weights from XXX not used in YYY* means that the layer XXX is not used by YYY, therefore those
+        weights are discarded.
+
+        Parameters:
+            pretrained_model_path (`os.PathLike`):
+                A path to a *directory* containing model weights saved using
+                [`~models.controlnets.multicontrolnet.MultiControlNetUnionModel.save_pretrained`], e.g.,
+                `./my_model_directory/controlnet`.
+            torch_dtype (`str` or `torch.dtype`, *optional*):
+                Override the default `torch.dtype` and load the model under this dtype. If `"auto"` is passed the dtype
+                will be automatically derived from the model's weights.
+            output_loading_info(`bool`, *optional*, defaults to `False`):
+                Whether or not to also return a dictionary containing missing keys, unexpected keys and error messages.
+            device_map (`str` or `Dict[str, Union[int, str, torch.device]]`, *optional*):
+                A map that specifies where each submodule should go. It doesn't need to be refined to each
+                parameter/buffer name, once a given module name is inside, every submodule of it will be sent to the
+                same device.
+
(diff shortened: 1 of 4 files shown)


guiyrt:
Some points I had in my mind:

1. What's the standing on `control_image ` / `image` parameter naming? Is this something relevant to change here?
2. Should I write new tests for `MultiControlNetUnionModel`?
3. In the forward() of `ControlNetUnionModel`, we have the argument `control_type_idx` which is the multi-hot encoding of the conditions used. Moving this to inside the function would remove some code from the pipelines, as it is derived from `control_type`. And we wouldn't need to handle it differently for multi or single controlnet_union, as we do now. As we are updating the controlnet_union pipelines, we could easily change this, if you find it relevant :) https://github.com/guiyrt/diffusers/blob/090e3ca6391bdfae1816c0ed7f6d90e50753d334/src/diffusers/pipelines/controlnet/pipeline_controlnet_union_sd_xl.py#L1247-L1253

src/diffusers/pipelines/controlnet/pipeline_controlnet_union_sd_xl.py
```diff
@@ -253,8 +260,8 @@ def __init__(
     ):
         super().__init__()
 
-        if not isinstance(controlnet, ControlNetUnionModel):
-            raise ValueError("Expected `controlnet` to be of type `ControlNetUnionModel`.")
```
hlky, line 257: For context this was added to ensure ControlNetUnion type was used and not ControlNetModel, we've since added a warning for pipelines when the passed object is the incorrect type so this is ok to remove.

src/diffusers/models/controlnets/multicontrolnet_union.py
```diff
@@ -0,0 +1,190 @@
...
+            else:
+                down_block_res_samples = [
+                    samples_prev + samples_curr
+                    for samples_prev, samples_curr in zip(down_block_res_samples, down_samples)
+                ]
+                mid_block_res_sample += mid_sample
+
+        return down_block_res_samples, mid_block_res_sample
+
+    def save_pretrained(
```
hlky, line 82:
```suggestion
    # Copied from diffusers.models.controlnets.multicontrolnet.MultiControlNetModel.save_pretrained with MultiControlNetModel->MultiControlNetUnionModel
    def save_pretrained(
```

src/diffusers/models/controlnets/multicontrolnet_union.py
```diff
@@ -0,0 +1,190 @@
...
+            controlnet.save_pretrained(
+                save_directory + suffix,
+                is_main_process=is_main_process,
+                save_function=save_function,
+                safe_serialization=safe_serialization,
+                variant=variant,
+            )
+
+    @classmethod
+    def from_pretrained(cls, pretrained_model_path: Optional[Union[str, os.PathLike]], **kwargs):
```
hlky, line 121:
```suggestion
    # Copied from diffusers.models.controlnets.multicontrolnet.MultiControlNetModel.from_pretrained with MultiControlNetModel->MultiControlNetUnionModel
    def from_pretrained(cls, pretrained_model_path: Optional[Union[str, os.PathLike]], **kwargs):
```

hlky (commented):
1. `control_image` / `image` is the standard naming, we'd like to keep it consistent across pipelines
2. That would be great
3. We have that in the pipeline to avoid re-computing it for every sampling step. The original had it in the model and used `nonzero` which caused a cuda sync every step. We changed it to avoid the sync, using `scatter` here is nice though, looks like we can use [`scatter_` to be in-place](https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_.html).

src/diffusers/pipelines/controlnet/pipeline_controlnet_union_sd_xl.py
guiyrt: I removed it because we are checking the controlnet type in `check_input`, but maybe it's not that clear, as it just throws `assert False`. Is that warning you're referring to already in place for this pipeline?

hlky: Yes it's generic thing that applies to all pipelines https://github.com/huggingface/diffusers/commit/0ed09a17bbab784a78fb163b557b4827467b0468

guiyrt: I passed a `ControlNetModel` instead of the expected `ControlNetUnionModel`, and it didn't throw any warning. I digged a little and found that the argument `controlnet` wasn't type checked, but the rest of the arguments were. 

This is because `controlnet` is part of `missing_modules`, so it is not type checked and only added to `init_kwargs` just before instantiating the pipeline.
https://github.com/huggingface/diffusers/blob/464374fb87610c53b2cf81e08d3df628fada3ce4/src/diffusers/pipelines/pipeline_utils.py#L993-L1004

This should be a different PR, but I propose we move the type check to just before the pipeline instantiation, so we can check every argument. Also, yesterday I played around a bit with dynamic type checking and implemented a way to recursively fully type-check collection types. For example, `controlnet` can be `List[ControlNetUnionModel]`, but the current type check passes with any list, as it only checks the first level.

I will open a different PR for this and you can let me know what you think :)

hlky: Thanks @guiyrt, we'll take a look at this in a different PR. Just to note we will not check `scheduler` as it's too restrictive for some pipelines.

yiyixuxu:
@hlky feel free to merge once it looks good to you!

hlky:
@guiyrt Can you check the output of `ControlNetUnionModel` with `Segmentation + Pose` against `main`?

guiyrt:
> @guiyrt Can you check the output of `ControlNetUnionModel` with `Segmentation + Pose` against `main`?

Yep, I'll post it here in a sec.

I'm working on the tests, one of the issues is related to `enable_sequential_cpu_offload`, we have the same problem we had with the SigLip image encoder we used for SD3 IP-Adapter, not sure if you remember (
https://github.com/huggingface/diffusers/pull/9987#issuecomment-2553716852). `ControlNetUnionModel` uses `nn.MultiheadAttention`, which only works with `enable_sequential_cpu_offload` if you exclude it from offloading. Only then it passes the tests `test_sequential_cpu_offload_forward_pass`and `test_sequential_cpu_offload_forward_pass`.

guiyrt:
> @guiyrt, Hi, thanks for awesome work, How can I use multi-controlnet-union? is it code? and also can you give me example for each condition scale?

Hi @john09282922, example inference code is above in a dropdown of the PR description, but I'll past here again :). To change condition scale, start and end, you just need to pass them as you would normally, but now in a list. For example, you now pass `control_guidance_start` as [0.1, 0.0, 0.5], if you have 3 controlnet_unions and 0.1 goes to the first, 0.0 to the second and 0.5 to the third. Same for `control_guidance_end` and `controlnet_conditioning_scale`.

<details>
<summary>Inference code</summary>

```python
import torch

from diffusers import StableDiffusionXLControlNetUnionPipeline
from diffusers.models import ControlNetUnionModel, AutoencoderKL

model_id = "stabilityai/stable-diffusion-xl-base-1.0"
controlnet_id = "brad-twinkl/controlnet-union-sdxl-1.0-promax"

controlnet = ControlNetUnionModel.from_pretrained(
    "brad-twinkl/controlnet-union-sdxl-1.0-promax", torch_dtype=torch.float16
)

pipe = StableDiffusionXLControlNetUnionPipeline.from_pretrained(
    controlnet=[controlnet, controlnet],
    vae = AutoencoderKL.from_pretrained(
        "madebyollin/sdxl-vae-fp16-fix", torch_dtype=torch.float16
    ),
    torch_dtype=torch.float16,
    variant="fp16",
)

room_seg_img = load_image("https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/sd_controlnet/room_seg.png")
pose_img = load_image("https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/sd_controlnet/pose.png")


image = pipe(
    prompt="an astronaut in space",
    width=1024,
    height=1024,
    guidance_scale=5,
    num_inference_steps=50,
    control_image=[[pose_img], [room_seg_img]],
    control_mode=[[0], [5]]
).images[0]

image.save("result.jpg")
```
</details>

guiyrt:
> @guiyrt Can you check the output of `ControlNetUnionModel` with `Segmentation + Pose` against `main`?


<table>
<thead>
<tr>
<th>Main</th>
<th>34ab1af</th>
</tr>
</thead>
<tbody>
<tr>
<td>

 </td>
<td>

</td>
</tr>
</tbody>
</table>

<details>
<summary>Inference code</summary>

```python
import torch

from diffusers import StableDiffusionXLControlNetUnionPipeline
from diffusers.models import ControlNetUnionModel, AutoencoderKL

model_id = "stabilityai/stable-diffusion-xl-base-1.0"

controlnet = ControlNetUnionModel.from_pretrained(
    "brad-twinkl/controlnet-union-sdxl-1.0-promax", torch_dtype=torch.float16
)

pipe = StableDiffusionXLControlNetUnionPipeline.from_pretrained(
    controlnet=controlnet,
    vae = AutoencoderKL.from_pretrained(
        "madebyollin/sdxl-vae-fp16-fix", torch_dtype=torch.float16
    ),
    torch_dtype=torch.float16,
    variant="fp16",
)

room_seg_img = load_image("https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/sd_controlnet/room_seg.png")
pose_img = load_image("https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/sd_controlnet/pose.png")


image = pipe(
    prompt="an astronaut in a space station",
    width=1024,
    height=1024,
    guidance_scale=5,
    num_inference_steps=50,
    control_image=[pose_img, room_seg_img],
    control_mode=[0, 5],
).images[0]

image.save("result_main.jpg")
```

</details>

hlky:
Thanks @guiyrt. Just checking output hasn't changed, I think we should be expecting the output to be the same between `ControlNetUnionModel` and `MultiControlNetUnionModel`, but both outputs for `Segmentation + Pose` look a little off to me.

cc @asomoza we're using control images of different resolutions here, and not resizing to the generation resolution, in your testing of ControlNetUnion does this affect the result? Do you have any recommendations for two control images that are known to work together?

guiyrt:
> Thanks @guiyrt. Just checking output hasn't changed, I think we should be expecting the output to be the same between `ControlNetUnionModel` and `MultiControlNetUnionModel`, but both outputs for `Segmentation + Pose` look a little off to me.

It works well for each condition, but to get the best results using both might require trying different `controlnet_conditioning_scale`, for example. Also, the best results I saw had both conditions complementing each other, which is not the case here.

> cc @asomoza we're using control images of different resolutions here, and not resizing to the generation resolution, in your testing of ControlNetUnion does this affect the result?

The control images are passed by the `VaeImageProcessor` which resizes them, but aspect ratio is not conserved.

<table>
<thead>
<tr>
<th>Original</th>
<th>Processed</th>
</tr>
</thead>
<tbody>
<tr>
<td>

</td>
<td>

</td>
</tr>
<tr>
<td>

</td>
<td>

 </td>
</tr>
</tbody>
</table>

asomoza:
I'm not particularly fond of the auto resizing of the images for controlnet, the good controlnets are really affected by bad resizings or resolutions, so I would like it more if it throws an error instead of auto scaling the images specially if it messes up the aspect ratio, but that's another issue and not something for this PR.

Using a real scenario and with real images with the correct sizes and also with the correct resolution for the conditioning images (they have resolutions with the preprocessors too), one common scenario is to use a depth map with another one of the edges or lines ones, I lie to use teed with a special combination I learned from anyline. 

This combination makes it really easy to test because if you want a good image you'll need to lower the scales and the guidance ends, so you can tell when they're not working good together.

This is my test with it:

|depth|teed|both|
|---|---|---|
|depth|teed|both|

In this case, the depth map is the one that gives the overall scene composition and lighting and the teed one is the one that adds the details, specially for the river waves and the background.

How does it compare to a single controlnet in main? I can test it with specific conditioning scales and ends for each one, so testing with both conditions at 1.0:

|single controlnet multi conditions|multi controlnet|
|---|---|
|single|both_full|

They're different but in theory they should produce the same result. To test this, I tested each condition separately with the single controlnet union:

|single controlnet depth|single controlnet teed|
|---|---|
|single_depth|single_teed|

There's definitely something going on here, but if you ask me, I think and like more the result in the multicontrolnet from this PR than in the original single one. Also I mostly use it with multricontrolnets because I like to control each one with a different guidance scale and end.

To me it seems that the single controlnet with multiple conditions is not applying that much the depth condition and takes more into account the teed one.

hlky (approved):
Thanks @guiyrt

guiyrt:
> I'm not particularly fond of the auto resizing of the images for controlnet, the good controlnets are really affected by bad resizings or resolutions, so I would like it more if it throws an error instead of auto scaling the images specially if it messes up the aspect ratio, but that's another issue and not something for this PR.

I agree on this, is there something in place for the other controlnet pipelines regarding preserving aspect ratio? Otherwise, if you find value on this, I could work on something.

> They're different but in theory they should produce the same result. To test this, I tested each condition separately with the single controlnet union:

I don't think an equal result is expected from "single controlnet multi conditions" and "multi controlnet". From the standpoint of controlnet inputs, the control embedding is different, and the fused control condition is also different, right? The output should be similar, but not equal.

asomoza:
> I don't think an equal result is expected from "single controlnet multi conditions" and "multi controlnet". From the standpoint of controlnet inputs, the control embedding is different, and the fused control condition is also different, right? The output should be similar, but not equal.

yeah, probably word it bad, that's what I meant, they're different when comparing them, and as I wrote before, I don't use the multiple condition in the same controlnet union except for testing. I'll do some more testings later.

Also I wouldn't make this something that important or a blocker, controlnet union it's really weird and cool at the same time, you can even mix the conditions in one image and it will still work, also you can use conditions with other control types to get some interesting results and most of the time they will still work.

vladmandic:
```py
    control_image=[[pose_img], [room_seg_img]],
    control_mode=[[0], [5]]
```
why are those now list-of-lists?
this is not how any other multicontrolnets are as control_image is list-of-images?

vladmandic:
yes, exactly.
- its either single controlnet union with multi inputs (and then we deal with no independent scale/start/end) or
- its multi controlnet union, each with single input
doing multi-of-multi is not something that can be effectively used - and makes assembling correct params a complete nightmare and non-standard with any other controlnet.

guiyrt:
> * its either single controlnet union with multi inputs (and then we deal with no independent scale/start/end) or
> * its multi controlnet union, each with single input

I looked into the Flux implementation for context, and it is indeed implemented differently. From what I got, there is no `FluxControlNetUnionModel`, there is only `FluxControlNetModel` where its forward() has the argument `controlnet_mode` as None for normal controlnets, and with a **single** value for controlnet unions. This means you **cannot** perform multi-condition input on a single controlnet union execution. If you have multiple conditions, you instead use `FluxMultiControlNetModel` which calls your controlnet union for each condition input separately. In this case, you have reduced memory usage, as a single controlnet is loaded, but no reduced execution time.

> doing multi-of-multi is not something that can be effectively used - and makes assembling correct params a complete nightmare and non-standard with any other controlnet.

I agree with you that in the current state, the list-of-lists input is confusing if you have many conditions or many controlnets, but I also don't think we should disregard the usage of multi-condition single execution if a controlnet union supports it. After giving some thought, maybe the ideal scenario would be the following:

- `ControNetModel`: regular controlnets and controlnet unions in single-condition mode, like for Flux. Input is single condition image and single condition mode (can be None for normal contronets).
- `ControlNetUnionModel`: exclusively for controlnet unions that operate in multi-condition mode, like in current version. Input is list of condition images and list of condition modes. Executes controlnet union once, independently of number of condition images.
- `MultiControlNetModel`: multiple `ControlNetModel`, like in Flux. Input is list of condition images and list of condition modes. Executes controlnets multiple times, as many as condition images. Nets can be (like for Flux):
    -  a single controlnet union operating in single-condition mode. This way, you can control condition scale/start/end individually.
    - a mix of normal controlnets and controlnet unions in single-condition mode.

Comparing to current version, this would have the benefits of simplified interface for `MultiControlNetModel` as in `FluxMultiControlNetModel`, and you could still benefit of multi-condition single execution controlnet unions via `ControlNetUnionModel`. The drawback is that you cannot use `ControlNetUnionModel` in `MultiControlNetModel`. If you want to use normal controlnets and controlnet unions, it can still be done but in single-condition mode.

The confusing interface comes from `MultiControlNetUnionModel`, where you can pass a list of conditions for each of your controlnets. In this proposal that would be removed, and with that you then can't run multiple controlnet unions in multi-condition mode. Not sure if that is a common use case, but if you want this functionality, you always need to pass a list of lists for condition images and modes.

hlky:
@vladmandic You requested this use case, @guiyrt has very kindly taken up MultiControlNetUnion and I've done experiments to use scale per condition without MultiControlNetUnion. Yes, it is a different interface, it is a unique model, all we need to do to handle this in any integration is an simple if else.

vladmandic:
@hlky @guiyrt i'll make it work, i'm just worried about lack of standardization which significantly increases complexities for end user.

vladmandic:
i'm ok with leaving this as-is - i'll add a special case handler on my side.

hlky:
@john09282922 Issues should be raised [here](https://github.com/huggingface/diffusers/issues), include a reproduction and the traceback.

asomoza:
what do you mean? the pipeline has the [IPAdapterMixin](https://github.com/guiyrt/diffusers/blob/34ab1af83eed226a9cd29aba862b0ecf058a98c2/src/diffusers/pipelines/controlnet/pipeline_controlnet_union_sd_xl.py#L183), you can just use it like any other SDXL pipeline with IP Adapters, there's nothing that changes in the method to use it.

asomoza:
> Hi, I got an issue when using inpainting model. I just changed current pipeline to inpaint pipeline, and when using multi-controlnet_union, there is set_att_process issue in multicontrolunion class. they don't have the process.

I see, the problem is not with IP Adapters, this PR only introduces `MultiControlnetUnion` to the `StableDiffusionXLControlNetUnionPipeline` so it won't work with any other controlnet pipeline.

asomoza:
> okay, can you merge it to other controlnet pipeline?

We leave those kind of tasks to the community if they want to do it and there's a popular need for it, it should be relatively easy and you can open a `feature request` with it but I don't think that people will take it because of two points:

- It's redundant since controlnetunion has an inpainting mode.
- you get a lot better results with it than an inpainting pipeline and model.

PR #10761, opened 2025-02-11
make tensors contiguous before passing to safetensors

faaany:
# What does this PR do?
Since `safetensors` requires tensors to be contiguous (see [this ](https://github.com/huggingface/safetensors/pull/530) link), we need to make sure that tensors are contiguous before passing it to safetensors. Below is the error message I got: 

```bash
Configuration saved in path-to-save-model/vae/config.json
Model weights saved in path-to-save-model/vae/diffusion_pytorch_model.safetensors
Configuration saved in path-to-save-model/unet/config.json
[rank0]: Traceback (most recent call last):
[rank0]:   File "/home/sdp/fanli/diffusers/examples/dreambooth/train_dreambooth.py", line 1444, in <module>
[rank0]:     main(args)
[rank0]:   File "/home/sdp/fanli/diffusers/examples/dreambooth/train_dreambooth.py", line 1420, in main
[rank0]:     pipeline.save_pretrained(args.output_dir)
[rank0]:   File "/home/sdp/fanli/diffusers/src/diffusers/pipelines/pipeline_utils.py", line 308, in save_pretrained
[rank0]:     save_method(os.path.join(save_directory, pipeline_component_name), **save_kwargs)
[rank0]:   File "/home/sdp/fanli/diffusers/src/diffusers/models/modeling_utils.py", line 557, in save_pretrained
[rank0]:     safetensors.torch.save_file(shard, filepath, metadata={"format": "pt"})
[rank0]:   File "/home/sdp/.local/lib/python3.10/site-packages/safetensors/torch.py", line 286, in save_file
[rank0]:     serialize_file(_flatten(tensors), filename, metadata=metadata)
[rank0]:   File "/home/sdp/.local/lib/python3.10/site-packages/safetensors/torch.py", line 496, in _flatten
[rank0]:     return {
[rank0]:   File "/home/sdp/.local/lib/python3.10/site-packages/safetensors/torch.py", line 500, in <dictcomp>
[rank0]:     "data": _tobytes(v, k),
[rank0]:   File "/home/sdp/.local/lib/python3.10/site-packages/safetensors/torch.py", line 414, in _tobytes
[rank0]:     raise ValueError(
[rank0]: ValueError: You are trying to save a non contiguous tensor: `conv_in.weight` which is not allowed. It either means you are trying to save tensors which are reference of each other in which case it's recommended to save only the full tensors, and reslice at load time, or simply call `.contiguous()` on your tensor to pack it before saving.
```

 cc @sayakpaul


hlky:
@faaany It's unusual for `conv_in.weight` to be non contiguous so let's find out why that's happening. Can you share some further details, what are the arguments to `train_dreambooth.py` for example? is this multi-GPU?

faaany:
> @faaany It's unusual for `conv_in.weight` to be non contiguous so let's find out why that's happening. Can you share some further details, what are the arguments to `train_dreambooth.py` for example? is this multi-GPU?

yes, multi-GPU or multi-XPU in my case. I am running the dog toy example in this README: https://github.com/huggingface/diffusers/tree/main/examples/dreambooth. 

The command is: 
```bash
export MODEL_NAME="CompVis/stable-diffusion-v1-4"
export INSTANCE_DIR="dog"
export OUTPUT_DIR="path-to-save-model"

accelerate launch train_dreambooth.py \
  --pretrained_model_name_or_path=$MODEL_NAME  \
  --instance_data_dir=$INSTANCE_DIR \
  --output_dir=$OUTPUT_DIR \
  --instance_prompt="a photo of sks dog" \
  --resolution=512 \
  --train_batch_size=1 \
  --gradient_accumulation_steps=1 \
  --learning_rate=5e-6 \
  --lr_scheduler="constant" \
  --lr_warmup_steps=0 \
  --max_train_steps=400
```

And my accelerate config yaml is:
```bash
compute_environment: LOCAL_MACHINE
debug: false
distributed_type: MULTI_XPU
downcast_bf16: 'no'
enable_cpu_affinity: false
gpu_ids: 2,3
ipex_config:
  use_xpu: true
machine_rank: 0
main_training_function: main
mixed_precision: bf16
num_machines: 1
num_processes: 2
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

hlky:
Thanks for confirming @faaany.

@sayakpaul This is not expected in (CUDA) multi-GPU environment, right? Seems like it could be a PyTorch XPU kernel issue that results in strided tensors. `.contiguous()` should be a no-op if the tensor is already contiguous but can be expensive if it is not, I think we should also follow SafeTensors philosophy of *not* forcing contiguous on everything as it's hiding the issues that could potentially effect other areas.

hlky (approved):
Thanks @faaany. In that case it's fine.

sayakpaul:
> @sayakpaul This is not expected in (CUDA) multi-GPU environment, right?

No I have never faced this issue in those environments. 

> but can be expensive if it is not, I think we should also follow SafeTensors philosophy of not forcing contiguous on everything as it's hiding the issues that could potentially effect other areas.

Usually the most use of `.contiguous()` is in the examples. I think that is fine. For its use in the core modules, I think we could profile a bit an introduce the change. WDYT?

PR #10764, opened 2025-02-11
speedup hunyuan encoder causal mask generation

dabeschte:
# What does this PR do?

The original causal attention mask generation for the hunyuan encoder is very slow, especially when the tensor is created on the GPU, because it needs to make 1000s of calls (for loop over sequence length).

I tried to compile it...which works and makes it fast too, but compilation unfortunately also takes a long time when using a long sequence length.

This implementation is ~20-70x faster depending on the sequence lengths and since it is re-created for every SDPA, this accumulates to multiple seconds per step for larger videos

This PR is a duplicate of the PR to the original HunyuanVideo repository https://github.com/Tencent/HunyuanVideo/pull/208

## Before submitting
- [ X] Did you write any new necessary tests?

## Who can review?
not sure tbh

### DIFF
--- a/src/diffusers/models/autoencoders/autoencoder_kl_hunyuan_video.py
+++ b/src/diffusers/models/autoencoders/autoencoder_kl_hunyuan_video.py
@@ -36,11 +36,11 @@ logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
 def prepare_causal_attention_mask(
     num_frames: int, height_width: int, dtype: torch.dtype, device: torch.device, batch_size: int = None
 ) -> torch.Tensor:
-    seq_len = num_frames * height_width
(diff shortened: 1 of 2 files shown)


a-r-r-o-w (approved):
Hey, thank you so much! It was on my mind to address this after the integration as it looked weird, but I forgot about it :/ This is super cool

a-r-r-o-w:
@dabeschte Could you look into the failing test? We can merge after that :)

dabeschte:
thanks for the quick review. 
yeah, it really does look weird to loop over the sequence lengths - and even more so when you profile model :D

I had a small bug in how I used the testing framework (looking at you, Cursor ;))
I didn't want to install all dependencies to get the tests to run - but this time I tested it in isolation to ensure at least this test is working correctly.

PR #10771, opened 2025-02-11
Refactor OmniGen

a-r-r-o-w:
Refactors OmniGen added in #10148.

<table align=center>
<tr>
  <td align=center>Image Editing</td>
  <td align=center>Text-to-Image</td>
</tr>
<tr>
  <td align=center></td>
  <td align=center></td>
</tr>
</table>

I haven't tested all supported usecases extensively, but I believe the changes should be okay since they did not alter any existing logic (except for RoPE).

The summary of changes are:
- Instead of the custom `apply_rotary_emb` function, we use the Diffusers one
- Removed set_attn_processors and related functions. We're thinking of a better way of supporting this since it makes the modeling implementation bloated. For any new models, let's not add these methods. It will be handled better with an upcoming refactor before this release hopefully.
- Makes sure guidance and scheduler step is performed in float32, so that we don't perform unnecessary bfloat16 -> float32 -> bfloat16 dtype conversion at each step. Instead, latents are always in float32 and casted just before feeding into transformer to required dtype.
- Removes everything related to attention kwargs. This will be added again by @sayakpaul, or someone else, who will take up adding peft support.
- Marks non-forward methods in the transformer private, since it could be subject to change in more future refactors
- Cleans up docs a bit and removes any intermediate layer docs. The parameters are self-explanatory, or already documented in the OmniGenTransformer2DModel docstring
- Enables layerwise upcasting test


src/diffusers/models/transformers/transformer_omnigen.py
```diff
@@ -593,106 +414,55 @@ def forward(
...
-                logger.warning(
-                    "Passing `scale` via `attention_kwargs` when not using the PEFT backend is ineffective."
-                )
-        height, width = hidden_states.size()[-2:]
+        # 1. Patch & Timestep & Conditional Embedding
         hidden_states = self.patch_embedding(hidden_states, is_input_image=False)
         num_tokens_for_output_image = hidden_states.size(1)
 
         time_token = self.time_token(self.time_proj(timestep).to(hidden_states.dtype)).unsqueeze(1)
+        timestep_proj = self.time_proj(timestep)
```
yiyixuxu, line 428: `self.time_token` can also use `timestep_proj` as input now, no?

src/diffusers/pipelines/omnigen/pipeline_omnigen.py
```diff
@@ -324,10 +320,8 @@ def __call__(
         latents: Optional[torch.Tensor] = None,
         output_type: Optional[str] = "pil",
         return_dict: bool = True,
-        attention_kwargs: Optional[Dict[str, Any]] = None,
         callback_on_step_end: Optional[Callable[[int, int, Dict], None]] = None,
         callback_on_step_end_tensor_inputs: List[str] = ["latents"],
-        max_sequence_length: int = 120000,
```
yiyixuxu, line 330: I see this is not used, but let's confirm with the author if it's intended to be used or not

yiyixuxu (approved):
very nice! thank you!

yiyixuxu:
@staoxiao can you take a look to see if the changes are ok?

sayakpaul:
Thanks for the heads-up, Aryan! Appreciate it.

I will wait for this before to get merged before starting the LoRA support.

src/diffusers/pipelines/omnigen/pipeline_omnigen.py
staoxiao: This argument is simply copied from other pipelines, but is not used in OmniGen.

staoxiao:
@a-r-r-o-w @yiyixuxu , thanks for refactoring. I think these changes are ok.

src/diffusers/models/transformers/transformer_omnigen.py
a-r-r-o-w: Ah thanks, nice observation! I've updated it

a-r-r-o-w:
Thanks for the review @staoxiao! Will merge once CI is green

PR #9510, opened 2024-09-24
[LoRA] allow loras to be loaded with low_cpu_mem_usage.

sayakpaul:
# What does this PR do?

https://github.com/huggingface/peft/pull/1961 added the ability to set `low_cpu_mem_usage` while loading LoRAs. This can be quite helpful in speeding up the loading of LoRAs that are large and have many layers. 

https://github.com/huggingface/diffusers/issues/8953 is a good example where this feature could be beneficial. 

<details>
<summary>Benchmarking code</summary>

```python
from diffusers import DiffusionPipeline 
import torch 
import time 
import fire

def main(ckpt_id: str, lora_id: str, low_cpu_mem_usage: bool = False):
    pipeline = DiffusionPipeline.from_pretrained(ckpt_id, torch_dtype=torch.bfloat16).to("cuda")
    
    for _ in range(10):
        start_time = time.time()  
        pipeline.load_lora_weights(lora_id, low_cpu_mem_usage=low_cpu_mem_usage)
        end_time = time.time()  
        pipeline.unload_lora_weights()
        elapsed_time = end_time - start_time 
        
        print(f"Iteration {_ + 1}: Load Lora weights took {elapsed_time:.6f} seconds with {low_cpu_mem_usage=}")

if __name__ == "__main__":
    fire.Fire(main)
```

</details>

<details>
<summary>Results:</summary>

```bash
Iteration 1: Load Lora weights took 13.924374 seconds with low_cpu_mem_usage=True
Iteration 2: Load Lora weights took 1.621597 seconds with low_cpu_mem_usage=True
Iteration 3: Load Lora weights took 1.612010 seconds with low_cpu_mem_usage=True
Iteration 4: Load Lora weights took 1.670260 seconds with low_cpu_mem_usage=True
Iteration 5: Load Lora weights took 1.664858 seconds with low_cpu_mem_usage=True
Iteration 6: Load Lora weights took 1.482521 seconds with low_cpu_mem_usage=True
Iteration 7: Load Lora weights took 1.633697 seconds with low_cpu_mem_usage=True
Iteration 8: Load Lora weights took 1.593326 seconds with low_cpu_mem_usage=True
Iteration 9: Load Lora weights took 1.503672 seconds with low_cpu_mem_usage=True
Iteration 10: Load Lora weights took 1.566633 seconds with low_cpu_mem_usage=True
```

---

```bash
Iteration 1: Load Lora weights took 33.370373 seconds with low_cpu_mem_usage=False
[...]

PR #10777, opened 2025-02-12
Faster set_adapters

Luvata:
# What does this PR do?

The previous code iterated through `model.named_modules()` **for each adapter**, which can be very costly when the number of adapters reaches hundreds.

I've slightly changed the logic to iterate over `model.named_modules()` **only once**, setting the adapters for each submodule within that pass.

I haven't run extensive qualitative tests yet, but in my local experiment (Flux with 150+ adapters 😅) this change is significantly faster.

## Before submitting

## Who can review?

- PEFT: @sayakpaul @BenjaminBossan

### DIFF
--- a/src/diffusers/utils/peft_utils.py
+++ b/src/diffusers/utils/peft_utils.py
@@ -253,29 +253,20 @@ def set_weights_and_activate_adapters(model, adapter_names, weights):
         parts = module_name.split(".")
         # e.g. key = "down_blocks.1.attentions.0"
         key = f"{parts[0]}.{parts[1]}.attentions.{parts[3]}"
-        block_weight = weight_for_adapter.get(key, 1.0)
+        return weight_for_adapter.get(key, 1.0)
 
-        return block_weight
-
-    # iterate over each adapter, make it active and set the corresponding scaling weight
-    for adapter_name, weight in zip(adapter_names, weights):
-        for module_name, module in model.named_modules():
-            if isinstance(module, BaseTunerLayer):
-                # For backward compatbility with previous PEFT versions
-                if hasattr(module, "set_adapter"):
-                    module.set_adapter(adapter_name)
-                else:
-                    module.active_adapter = adapter_name
-                module.set_scale(adapter_name, get_module_weight(weight, module_name))
-
-    # set multiple active adapters
-    for module in model.modules():
+    # Iterate over modules once
+    for module_name, module in model.named_modules():
         if isinstance(module, BaseTunerLayer):
-            # For backward compatbility with previous PEFT versions
+            # For backward compatibility with previous PEFT versions, set multiple active adapters
             if hasattr(module, "set_adapter"):
                 module.set_adapter(adapter_names)
             else:
                 module.active_adapter = adapter_names
+
+            # Set the scaling weight for each adapter for this module
+            for adapter_name, weight in zip(adapter_names, weights):
+                module.set_scale(adapter_name, get_module_weight(weight, module_name))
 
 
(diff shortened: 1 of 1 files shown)


sayakpaul:
Thanks for this PR. Do you have any benchmarking numbers on this?

Luvata:
I made a small benchmark with Colab [here](https://colab.research.google.com/drive/1IcyCTPJT2LhjuDqizPH6kgPIEaw5E0vN?usp=sharing). 

```python
import time
from tqdm import tqdm
import torch

pipe = DiffusionPipeline.from_pretrained("stable-diffusion-v1-5/stable-diffusion-v1-5", torch_dtype=torch.float16).to("cuda")


def load_n_lora(n):
    pipe.unload_lora_weights()
    adapter_names = []
    for i in range(n):
        adapter_names.append(f"floor{i}")
        pipe.load_lora_weights("maria26/Floor_Plan_LoRA", adapter_name=f"floor{i}")
    return adapter_names

for n_lora in [1, 5, 10, 20, 50]:
    adapter_names = load_n_lora(n_lora)
    adapter_weights = [1./n_lora] * n_lora

    start = time.time()
    pipe.set_adapters(adapter_names, adapter_weights=adapter_weights)
    end = time.time()
    print(f"n_lora: {n_lora}, time: {end - start}")
```

|   n_lora |  main branch set_adapters | faster set_adapters | load_lora_weights |
|---------:|---------------:|---------------:|-------------|
|        1 |         0.0164 |         0.012  | 0.466 |
|        5 |         0.118 |         0.067 | 2.763 | 
|       10 |         0.699 |         0.206  | 6.082 |
|       20 |         1.762 |         1.283 | 14.064 | 
|       50 |        9.661 |         5.975 | 49.573 |

The difference will be more significant with bigger models (e.g Flux)

btw, `load_lora_weights` is also very slow as the number of LoRA increases, but since in my use case it runs only once, I didn't look into optimizing it.

sayakpaul:
Thanks for providing the benchmark!

> btw, load_lora_weights is also very slow as the number of LoRA increases, but since in my use case it runs only once at first to load all lora, I didn't look into optimizing it.

Can you ensure you're using `low_cpu_mem_usage=True`. By default it should be set to True based on your env but just double-checking.

sayakpaul:
Can you confirm if the machine you're using to benchmark is shared by other uses? Sometimes that can perturbate the results.

It's bit weird that you're experiencing such low load times (even though this issue should be in a different thread). We benchmarked it with `low_cpu_mem_usage` and you can check the results here: https://github.com/huggingface/diffusers/pull/9510.

Luvata:
Oh I see, I forgot the unload is still inside the timing, i'm rerunning the benchmark now
btw, for reference, my local experiment with Flux Schnell, loading LoRA weights with the default setting for 152 adapters (rank 4) takes around 4 minutes with no unload.

sayakpaul:
> btw, for reference, my local experiment with Flux Schnell, loading LoRA weights with the default setting for 152 adapters (rank 4) takes around 4 minutes with no unload.

What is the expected time you would like so here? 👀 4 mins for 152 LoRAs seem reasonable to me.

Luvata:
In the first 7 seconds, diffusers can load 20 LoRAs, so I expect it could load faster overall. nvm, I'll look into it more closely tomorrow since it's pretty late now

BenjaminBossan (approved):
The changes here LGTM, even if there was no speedup. Thanks.

Regarding the LoRA loading, I'd suggest to open another issue and check the underlying issue there.

PR #10789, opened 2025-02-13
Refactor CogVideoX transformer forward

### LINKED ISSUE #10566: Unnecessary operations in `CogVideoXTransformer3DModel.forward()`?
### Describe the bug

Here are few rows of codes in `CogVideoXTransformer3DModel.forward()` :
```py
        # 3. Transformer blocks
        ...

        if not self.config.use_rotary_positional_embeddings:
            # CogVideoX-2B
            hidden_states = self.norm_final(hidden_states)
        else:
            # CogVideoX-5B
            hidden_states = torch.cat([encoder_hidden_states, hidden_states], dim=1)
            hidden_states = self.norm_final(hidden_states)
            hidden_states = hidden_states[:, text_seq_length:]

        # 4. Final block
        ...
```

where `self.norm_final` is a `LayerNorm`  defined by:
```py
self.norm_final = nn.LayerNorm(inner_dim, norm_eps, norm_elementwise_affine)
```

Since the `normalized_shape` of  `self.norm_final` is 1-dimension which means only the last dimension will be normalized, it seems that **the "cat -> layernorm -> slice" logic on the 2nd dimension in CogVideoX-5B branch is unnecessary because it does the same thing with**
```py
hidden_states = self.norm_final(hidden_states)
```

These codes is imported via [PR#9203](https://github.com/huggingface/diffusers/pull/9203/files#diff-6e4d5c6638b71b7a0e7de21357c5b55ffd5ff6373dd1ced70070650855830173R469).  @zRzRzRzRzRzRzR @yiyixuxu  could you possibly walk me through why these changes were necessary? Thanks a lot for your help!

### Reproduction

.

### Logs

```shell

```

### System Info

.

### Who can help?

_No response_

a-r-r-o-w:
Fixes #10566.

I overlooked this part during the integration. Since we're operating on the embedding dimension, there's no need to concatenate across the sequence dimension

### DIFF
--- a/src/diffusers/models/transformers/cogvideox_transformer_3d.py
+++ b/src/diffusers/models/transformers/cogvideox_transformer_3d.py
@@ -503,14 +503,7 @@ class CogVideoXTransformer3DModel(ModelMixin, ConfigMixin, PeftAdapterMixin, Cac
                     attention_kwargs=attention_kwargs,
                 )
 
-        if not self.config.use_rotary_positional_embeddings:
-            # CogVideoX-2B
-            hidden_states = self.norm_final(hidden_states)
-        else:
-            # CogVideoX-5B
-            hidden_states = torch.cat([encoder_hidden_states, hidden_states], dim=1)
-            hidden_states = self.norm_final(hidden_states)
-            hidden_states = hidden_states[:, text_seq_length:]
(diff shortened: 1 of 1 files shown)


yiyixuxu (approved):
thanks!

Issue #10804, opened 2025-02-16
comfy-ui compatible FLUX1.dev LoRA fails to load


AmericanPresidentJimmyCarter:
### Describe the bug

https://civitai.com/models/677200?modelVersionId=758070

This LoRA fails to load. Specifically, `model.load_state_dict` fails with mismatched tensors when attempting to load in the LoRA state dict.

### Reproduction

```py
import torch
from diffusers import FluxPipeline

pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16).to('cuda')

pipe.load_lora_weights('/data/models/wow_details.safetensors', adapter_name="wow_details")

prompt = "a tiny astronaut hatching from an egg on the moon"
out = pipe(
    prompt=prompt,
    guidance_scale=3.5,
    height=1024,
    width=1024,
    num_inference_steps=25,
).images[0]
out.save("image.png")
```

### Logs

```shell
pipe.load_lora_weights(lora_fn, adapter_name=name, low_cpu_mem_usage=True)
env/lib/python3.11/site-packages/diffusers/loaders/lora_pipeline.py:1872: in loa
d_lora_weights                                                                  
    self.load_lora_into_transformer(                                            
env/lib/python3.11/site-packages/diffusers/loaders/lora_pipeline.py:1936: in loa
d_lora_into_transformer                                                         
    transformer.load_lora_adapter(                                              
env/lib/python3.11/site-packages/diffusers/loaders/peft.py:327: in load_lora_ada
pter                                                                            
    incompatible_keys = set_peft_model_state_dict(self, state_dict, adapter_name
, **peft_kwargs)                                                                
env/lib/python3.11/site-packages/peft/utils/save_and_load.py:432: in set_peft_mo
del_state_dict
    load_result = model.load_state_dict(peft_model_state_dict, strict=False)

  File "/root/app/env/lib/python3.11/site-packages/torch/nn/modules/module.py", line 2584, in load_state_dict                                        raise RuntimeError(                                                         RuntimeError: Error(s) in loading state_dict for FluxTransformer2DModel:        
        size mismatch for transformer_blocks.0.attn.to_q.lora_A.wow_details.weight: copying a param with shape torch.Size([24, 3072]) from checkpoint, the shape in current model is torch.Size([4, 3072]).                                     
[...]

PR #10808, opened 2025-02-17
[LoRA] restrict certain keys to be checked for peft config update.

sayakpaul:
# What does this PR do?

Fixes: https://github.com/huggingface/diffusers/issues/10804

Code:

```py
import torch

).to("cuda")

pipe.load_lora_weights(
    "sayakpaul/different-lora-from-civitai", 
    weight_name="wow_details.safetensors", 
    adapter_name="wow_details"
)

prompt = "a tiny astronaut hatching from an egg on the moon"
out = pipe(
    prompt=prompt,
    guidance_scale=3.5,
    height=1024,
    width=1024,
    num_inference_steps=25,
).images[0]
out.save("image.png")
```

### DIFF
--- a/src/diffusers/loaders/peft.py
+++ b/src/diffusers/loaders/peft.py
@@ -53,6 +53,7 @@ _SET_ADAPTER_SCALE_FN_MAPPING = {
     "LTXVideoTransformer3DModel": lambda model_cls, weights: weights,
     "SanaTransformer2DModel": lambda model_cls, weights: weights,
 }
+_NO_CONFIG_UPDATE_KEYS = ["to_k", "to_q", "to_v"]
 
 
 def _maybe_adjust_config(config):
@@ -67,6 +68,8 @@ def _maybe_adjust_config(config):
     original_r = config["r"]
 
     for key in list(rank_pattern.keys()):
+        if any(prefix in key for prefix in _NO_CONFIG_UPDATE_KEYS):
+            continue
         key_rank = rank_pattern[key]
 
         # try to detect ambiguity


sayakpaul:
Your deducations are correct, thank you!

Help me understand this:

> peft is treating rank_pattern as a pattern but we aren't using it as a pattern, when we say proj_out in the original config that's the whole key

We infer this in `get_peft_kwargs()`. There's no direct way for us to specify a `rank_pattern` which is why the adjustment bit was added during Control LoRA.

> If we change get_pattern_key to rf"^{key}$" then black-forest-labs/FLUX.1-Canny-dev-lora works without _maybe_adjust_config and the lora from this issue.

Feel free to push it directly to my branch and we can run the tests to confirm nothing broke.

sayakpaul:
> but it's the full key that we want to be rank 64 so we adjust the config essentially reversing default and rank pattern, defining full keys of everything we want to be a non-default rank. 

This was done with suggestions from @BenjaminBossan here:
https://github.com/huggingface/diffusers/pull/9985#issuecomment-2493840028

> So I'm wondering if this could be an option from peft to use the passed keys in rank_pattern as entire keys not suffixes, and how common it is to pass a suffix/actual pattern in rank_pattern.

I think it's a bit hard to gauge but I will let @BenjaminBossan comment further. In my experience, I would say this is fairly divided.

BenjaminBossan:
Always this cursèd problem of missing metadata :-/

> `r` is 24, problem keys `transformer_blocks.0.attn.to_q` etc get removed, and all other `target_modules` are added to `rank_pattern` with rank `4` which also restores `transformer_blocks.0.attn.to_q` etc with rank `4` on later iterations and causes the reported issue.

In general, I think we have to accept that without metadata, we cannot guarantee that the identical config is restored, but it should of course still lead to the same adapter model. Do we know why `transformer_blocks.0.attn.to_q` is being removed? If we can prevent that, would it solve the problem?

> peft is treating rank_pattern as a pattern but we aren't using it as a pattern, when we say proj_out in the original config that's the whole key
If we change get_pattern_key to rf"^{key}$" then black-forest-labs/FLUX.1-Canny-dev-lora works without _maybe_adjust_config and the lora from this issue.

We can't change the base logic here in PEFT, as it could potentially break existing code. Theoretically, we could think about having a flag on `LoraConfig` that determines how to parse the `rank_pattern` keys, but I think this would be overly specific. Therefore, I'm considering another option: Maybe we can add a special prefix to a key to "mark" it as being the full key. Something along the lines of:

```python
FULL_KEY_PREFIX = "FULL-KEY-"  # insert better name

def get_pattern_key(pattern_keys, key_to_match):
    """Match a substring of key_to_match in pattern keys"""
    if key.startswith(FULL_KEY_PREFIX):  # <== new logic
        return key_to_match[len(FULL_KEY_PREFIX):]
    # old logic
    return next(filter(lambda key: re.match(rf".*\.{key}$", key_to_match), pattern_keys), key_to_match)
```

Diffusers would have to be adjusted to use this new prefix. We would also need to ensure that the PEFT and diffusers version match, otherwise adding the prefix in diffusers would result in no match on the PEFT side. WDYT, would it solve the issue? I'm also open to different proposals.

sayakpaul:
Thanks for the discussions! Appreciate that. 

> Do we know why transformer_blocks.0.attn.to_q is being removed? If we can prevent that, would it solve the problem?

It's this line that is the culprit I think:
https://github.com/huggingface/diffusers/blob/c14057c8dbc32847bac9082bcc0ae00c9a19357d/src/diffusers/loaders/peft.py#L77

> Diffusers would have to be adjusted to use this new prefix. We would also need to ensure that the PEFT and diffusers version match, otherwise adding the prefix in diffusers would result in no match on the PEFT side. WDYT, would it solve the issue? I'm also open to different proposals.

This sounds like a bit too much of `diffusers`-specific changes on the `peft` side whereas a simple fix (for now) exists. Also, note that these changes are usually motivated by community checkpoints which we cannot always control, hence the guesswork.

BenjaminBossan:
> This sounds like a bit too much of `diffusers`-specific changes on the `peft` side whereas a simple fix (for now) exists. Also, note that these changes are usually motivated by community checkpoints which we cannot always control, hence the guesswork.

I wouldn't say it's diffusers-specific, there could be other model architectures where users may want to add a specific rank pattern but where the target module key is not unique. Therefore, I think a more general solution would be preferable, even if it's more complicated.

> _NO_CONFIG_UPDATE_KEYS = ["to_k", "to_q", "to_v"]

Isn't this error prone?

sayakpaul:
It is error-prone, yes. 

> FULLY_QUALIFIED_KEY_PREFIX

How do we decide a reasonable default for this variable? 

API-wise could you walk me through the changes that might needed on the `diffusers` side? Would like to gauge the changes a bit. `get_pattern_key()` isn't used in the `diffusers` codebase. Or maybe I am not fully understanding the changes on the `diffusers` side.

BenjaminBossan:
> API-wise could you walk me through the changes that might needed on the `diffusers` side? Would like to gauge the changes a bit. `get_pattern_key()` isn't used in the `diffusers` codebase. Or maybe I am not fully understanding the changes on the `diffusers` side.

If I'm not missing something, the change would be that instead of creating a `rank_pattern` dict like

`{key0, rank0, key1: rank1, ...}`

the dict would be

`{FULLY_QUALIFIED_KEY_PREFIX + key0, rank0, FULLY_QUALIFIED_KEY_PREFIX + key1: rank1, ...}`

This would signal to PEFT that we should consider this key to be the full key, not a pattern to match.

I'm not sure if this could lead to bad performance, given that `rank_pattern` could be quite large and `get_pattern_key` would be doing more work on the PEFT side, but I think not (it's not quadratic or worse).

BenjaminBossan:
If I understood the issue correctly, it is caused by trying to set a special rank pattern for a certain key but the same key (`to_q`) is also matching another key because PEFT interprets it as a pattern. Thus if diffusers would let PEFT know that it's actually the full name and not a pattern, the false match could be avoided.

BenjaminBossan:
To be clear, if you think the proposed solution in this PR is good enough to work for most users, I'd also be fine with going that way and not having to change PEFT, as it's less work overall :)

sayakpaul:
> If I understood the issue correctly, it is caused by trying to set a special rank pattern for a certain key but the same key (`to_q`) is also matching another key because PEFT interprets it as a pattern. Thus if diffusers would let PEFT know that it's actually the full name and not a pattern, the false match could be avoided.

I see. This can work. Will just have to figure out the right API calls inside of `diffusers` when the PEFT support has landed. I think this is much better as a solution.

> To be clear, if you think the proposed solution in this PR is good enough to work for most users, I'd also be fine with going that way and not having to change PEFT, as it's less work overall :)

My solution was admittedly a dirty hack. But after the discussions, I am leaning towards your proposal.

BenjaminBossan:
I created a draft PR to implement this: https://github.com/huggingface/peft/pull/2382. Let's check that, with the right adjustments on the diffusers code, this is solving the original issue before proceeding.

hlky:
> Always this cursèd problem of missing metadata :-/

There is no missing metadata, right? we're detecting the correct default rank and "fully qualified" keys to set as a different rank. Adding a magic key seems like a weird solution to me.

BenjaminBossan:
> There is no missing metadata, right? we're detecting the correct default rank and "fully qualified" keys to set as a different rank.

I just meant that in general, many LoRA training frameworks don't save any `LoraConfig`-equivalent metadata with the checkpoint, which means that it needs to be inferred.

> Adding a magic key seems like a weird solution to me.

It's not beautiful, but what would be your suggested solution?

hlky:
> it needs to be inferred

I think it's being inferred for both loras mentioned in this issue. Are there known circumstances when the metadata cannot be inferred correctly?

> what would be your suggested solution

Maybe just a flag in `LoraConfig` or a separate list for fully qualified keys. As far as I can tell we're always using fully qualified keys with `rank_pattern` (and `target_modules`) from `diffusers`.

sayakpaul:
> Are there known circumstances when the metadata cannot be inferred correctly?

As far as I can tell from my experience of dealing with several LoRA checkpoints from the community, metadata (rank, alpha, etc.) can always be inferred from the LoRA checkpoints. 

> As far as I can tell we're always using fully qualified keys with rank_pattern (and target_modules) from diffusers.

Yes, this is true, especially when we're loading non-diffusers LoRA checkpoints. However, I think @BenjaminBossan wanted a more general solution to deal with this in `peft`:

@BenjaminBossan, in https://github.com/huggingface/peft/pull/2382, what if we add a flag instead like @hlky suggested. What would be the downside of it?

hlky:
It seems there are 2 separate paths for peft/`LoraConfig` then, one for "training" where we may want to specify a pattern/suffix like `proj_out` and "inference" when we need to use inferred keys as-is. Enforcing a regex makes "training" somewhat easier but has limitations/[impossible configs](https://github.com/huggingface/peft/blob/9849e89fc88b5a2c3ca38431cec5b3b47e6cddad/tests/test_custom_models.py#L2047-L2062) and makes "inference" awkward. Could a future major version consider accepting a regex directly, that would support both cases, we'd pass `.*\.proj_out$` ourselves if we wanted `.proj_out` suffix and just `proj_out` if we wanted fully qualified key `proj_out`.

A flag would feel more natural but I have no major issues with a magic key prefix.

sayakpaul:
I can then base this PR off of https://github.com/huggingface/peft/pull/2382 until the regex flag is landed in the future. Is that fine by @hlky @BenjaminBossan?

sayakpaul:
@BenjaminBossan @hlky I have applied the changes. PTAL. I haven't intentionally version-guarded the PEFT versions yet.

hlky (approved):
LGTM

src/diffusers/loaders/peft.py
```diff
@@ -53,6 +53,8 @@
     "LTXVideoTransformer3DModel": lambda model_cls, weights: weights,
     "SanaTransformer2DModel": lambda model_cls, weights: weights,
 }
+_NO_CONFIG_UPDATE_KEYS = ["to_k", "to_q", "to_v"]
```
sayakpaul, line 56: Will be deleted.

src/diffusers/loaders/peft.py
```diff
@@ -53,6 +53,8 @@
...
+_FULL_NAME_PREFIX_FOR_PEFT = "FULL-NAME"
 
 
 def _maybe_adjust_config(config):
```
sayakpaul, line 60: This method will be removed altogether.

BenjaminBossan:
> Could a future major version consider accepting a regex directly, that would support both cases, we'd pass `.*\.proj_out$` ourselves if we wanted `.proj_out` suffix and just `proj_out` if we wanted fully qualified key `proj_out`.

I had also thought about that. Ideally, we could allow passing a string if we want exact match and a regex pattern (or callable) if we want more flexibility. The problem with this:

1. It would change the semantics of existing configs from pattern to exact match, so would be backwards incompatible.
2. We can't serialize regex pattern (or callables) to json. This could be avoided by special serialization logic, but it's no less complex than the magic string solution.

> A flag would feel more natural but I have no major issues with a magic key prefix.

My reasons to avoid the flag solution are:

1. I'd like to avoid having parameters on `LoraConfig` intended for changing internal implementation details instead of actually controlling the LoRA adapter. 
2. With the magic prefix, we have the option to mix pattern and strict matching in the same `rank_pattern`, so this gives us more flexibility.
4. It's not without precedent. We already have `target_modules="all-linear"` and `target_modules="dummy-target-modules"` as special cases.

One more thing that came to my mind is that if we make these changes, no matter if it's the magic prefix or the flag solution, could we run into trouble for other packages that load diffusers LoRA checkpoints and need to be aware of the change? I suppose not, as they probably use other formats than PEFT, but I just wanted to bring up the possibility.

hlky:
> would be backwards incompatible.

Of course, it could be planned for `v1.0.0` or later. Assuming `peft` will follow semver breaking changes would be allow with a major version increment.

> We can't serialize regex pattern

WDYM? We couldn't just do something like this?

```python
import re
import json
re.match(json.loads(json.dumps({"regex": ".*\.proj_out$"}))['regex'], "test.proj_out")
<re.Match object; span=(0, 13), match='test.proj_out'>
```

> I'd like to avoid having parameters on LoraConfig intended for changing internal implementation details instead of actually controlling the LoRA adapter.

If not a flag, an additional `fully_qualified_rank_keys` would work and support 2.

> With the magic prefix, we have the option to mix pattern and strict matching in the same rank_pattern, so this gives us more flexibility.

However, I don't think this will be a particularly common use case.

> for other packages that load diffusers LoRA checkpoints and need to be aware of the change

WDYM? It's not checkpoint specific

Anyway, the proposed solution is ok, it does what we need and allows us reduce complexity by removing `_maybe_adjust_config` (note: @sayakpaul a possible fix for the issue in that function is storing deleted keys to ensure they aren't re-added by other iterations)

sayakpaul:
@BenjaminBossan 

> One more thing that came to my mind is that if we make these changes, no matter if it's the magic prefix or the flag solution, could we run into trouble for other packages that load diffusers LoRA checkpoints and need to be aware of the change? I suppose not, as they probably use other formats than PEFT, but I just wanted to bring up the possibility.

Well, we're going through hurdles because the checkpoint under consideration is a non-diffusers and non-peft checkpoint. So, I think we're good for now.

LMK if the current changes are good for you.

@hlky 

> (note: @sayakpaul a possible fix for the issue in that function is storing deleted keys to ensure they aren't re-added by other iterations)

Thanks for the headsup, yeah I thought of about it too but didn't reflect on it here as we had already found a better solution.

BenjaminBossan:
> WDYM? We couldn't just do something like this?

What I meant is that it would be cool if we could do:

```python
config = LoraConfig(..., rank_pattern={"my-string-key": 8, re.compile("my-regex-key"): 16})
```

That way, we would inverse control so that users can decide how they want to match if the pass a regex as key. However, we would not be able to JSON-serialize this config.

> If not a flag, an additional `fully_qualified_rank_keys` would work and support 2.

You mean something like:

```python
config = LoraConfig(..., rank_pattern={"foo": 8, "bar": 16}, fully_qualified_rank_keys=["foo"])?
```

I'm not a big fan of this pattern, as it requires 2 independent data structures to be in sync. It's easier to make mistake as a user and PEFT would need to implement checks for consistency.

> However, I don't think this will be a particularly common use case.

I agree.

> WDYM? It's not checkpoint specific

I mean if this adapter is saved in the PEFT format, it will contain the magic string. If anyone wants to load this without PEFT, they need to be aware of that. As mentioned, I don't think it's a big deal but I just wanted to mention it in case I'm missing something.

> LMK if the current changes are good for you.

> `_FULL_NAME_PREFIX_FOR_PEFT = "FULL-NAME"`

In the PEFT PR, I went with `"FULL-NAME-"`. But we can also agree on a different prefix, I don't care too much. If we agree, I can merge the PEFT PR and the prefix can be imported anyway. On the diffusers side, I think we would need a PEFT version guard though, right?

hlky:
> it will contain the magic string

Oh, we will need to handle that then.

sayakpaul:
> In the PEFT PR, I went with "FULL-NAME-". But we can also agree on a different prefix, I don't care too much. If we agree, I can merge the PEFT PR and the prefix can be imported anyway. On the diffusers side, I think we would need a PEFT version guard though, right?

Yeah, I didn't version-guard yet. I think we may have to support both codepaths based on the `peft` version being used:

* Adjusting the LoRA configs 
* Current updates 

Do you see any easy way other than this?

BenjaminBossan:
Hmm, I would have added a check for PEFT > 0.14.0 and

- if True, add the prefix
- if False, either:
  - don't add the prefix and see what happens
  - don't add the prefix and use your original workaround

sayakpaul:
> don't add the prefix and see what happens

Well, it would certainly fail without `_maybe_adjust_lora_config()`. Is that what you meant?

BenjaminBossan:
> Well, it would certainly fail without `_maybe_adjust_lora_config()`. Is that what you meant?

Yes, it would mean that for the error to go away, users would need the next PEFT version. But we could also adopt the workaround for older PEFT versions and the new prefix-based solution if the installed PEFT version supports it.

sayakpaul:
> But we could also adopt the workaround for older PEFT versions and the new prefix-based solution if the installed PEFT version supports it.

Yeah for the older peft version, I will defer to what I had in this PR (constant the attention modules and checking against them in the config update method). Do you have any other ideas?

BenjaminBossan:
> Do you have any other ideas?

No, unfortunately not :(

sayakpaul:
@BenjaminBossan ready for another review.

src/diffusers/loaders/peft.py
```diff
@@ -54,6 +54,7 @@
     "SanaTransformer2DModel": lambda model_cls, weights: weights,
     "Lumina2Transformer2DModel": lambda model_cls, weights: weights,
 }
+_NO_CONFIG_UPDATE_KEYS = ["to_k", "to_q", "to_v"]
```
hlky, line 57:
```suggestion
```

src/diffusers/loaders/peft.py
```diff
@@ -68,6 +69,8 @@ def _maybe_adjust_config(config):
     original_r = config["r"]
 
     for key in list(rank_pattern.keys()):
+        if any(prefix in key for prefix in _NO_CONFIG_UPDATE_KEYS):
+            continue
```
hlky, line 73:
```suggestion
```

hlky (approved):
LGTM

src/diffusers/loaders/peft.py
sayakpaul: We need it when PEFT version doesn't contain the required prefix.

hlky: Got it

> a possible fix for the issue in that function is storing deleted keys to ensure they aren't re-added by other iterations

If this issue reoccurs with other keys before minimum PEFT version is increased this can be applied

sayakpaul: Check now?

hlky (approved):

joangava (approved):

src/diffusers/loaders/peft.py
```diff
@@ -71,30 +74,27 @@ def _maybe_adjust_config(config):
         key_rank = rank_pattern[key]
 
         # try to detect ambiguity
-        # `target_modules` can also be a str, in which case this loop would loop
```
BenjaminBossan, line 74: Any specific reason why this was removed?

BenjaminBossan (approved):
It's a bit of a messy situation, but I think this PR finds a nice solution to the issue. LGTM. If the test script passes both with and without the new feature in PEFT, it's good to be merged.

src/diffusers/loaders/peft.py
sayakpaul: Let me add this back.

sayakpaul:
@AmericanPresidentJimmyCarter can you check if this solves https://github.com/huggingface/diffusers/issues/10804

burgalon:
after install diffusers@b0550a66cc3c882a1b88470df7e26103208b13de and trying to load Comfy LoRA (https://civitai.com/models/631986/xlabs-flux-realism-lora) - still getting 
```
ValueError: Incompatible keys detected:

diffusion_model.double_blocks.0.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.0.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.0.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.0.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.0.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.0.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.0.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.0.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.1.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.1.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.1.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.1.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.1.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.1.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.1.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.1.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.10.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.10.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.10.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.10.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.10.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.10.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.10.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.10.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.11.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.11.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.11.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.11.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.11.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.11.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.11.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.11.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.12.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.12.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.12.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.12.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.12.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.12.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.12.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.12.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.13.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.13.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.13.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.13.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.13.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.13.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.13.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.13.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.14.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.14.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.14.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.14.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.14.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.14.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.14.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.14.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.15.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.15.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.15.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.15.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.15.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.15.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.15.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.15.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.16.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.16.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.16.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.16.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.16.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.16.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.16.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.16.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.17.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.17.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.17.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.17.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.17.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.17.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.17.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.17.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.18.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.18.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.18.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.18.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.18.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.18.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.18.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.18.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.2.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.2.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.2.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.2.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.2.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.2.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.2.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.2.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.3.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.3.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.3.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.3.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.3.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.3.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.3.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.3.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.4.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.4.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.4.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.4.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.4.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.4.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.4.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.4.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.5.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.5.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.5.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.5.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.5.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.5.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.5.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.5.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.6.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.6.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.6.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.6.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.6.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.6.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.6.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.6.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.7.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.7.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.7.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.7.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.7.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.7.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.7.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.7.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.8.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.8.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.8.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.8.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.8.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.8.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.8.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.8.txt_attn.qkv.lora_up.weight, diffusion_model.double_blocks.9.img_attn.proj.lora_down.weight, diffusion_model.double_blocks.9.img_attn.proj.lora_up.weight, diffusion_model.double_blocks.9.img_attn.qkv.lora_down.weight, diffusion_model.double_blocks.9.img_attn.qkv.lora_up.weight, diffusion_model.double_blocks.9.txt_attn.proj.lora_down.weight, diffusion_model.double_blocks.9.txt_attn.proj.lora_up.weight, diffusion_model.double_blocks.9.txt_attn.qkv.lora_down.weight, diffusion_model.double_blocks.9.txt_attn.qkv.lora_up.weight

....
```

sayakpaul:
Could you provide a new issue thread as I just ran the xlabs integration tests and they are passing:

https://github.com/huggingface/diffusers/blob/cc7b5b873a38ee68f35620e77f19427d1dd2b6ab/tests/lora/test_lora_layers_flux.py#L911

https://github.com/huggingface/diffusers/blob/cc7b5b873a38ee68f35620e77f19427d1dd2b6ab/tests/lora/test_lora_layers_flux.py#L933

(I ensured I pulled in the latest changes in `main`)

AmericanPresidentJimmyCarter:
> @AmericanPresidentJimmyCarter can you check if this solves #10804

Confirmed that it works.

PR #10815, opened 2025-02-18
Enhance-A-Video

a-r-r-o-w:
Adds support for Enhance-A-Video.

Paper: https://huggingface.co/papers/2502.07508
Project: https://oahzxl.github.io/Enhance_A_Video
Code: https://github.com/NUS-HPC-AI-Lab/Enhance-A-Video

The PR needs some rework in terms of user-facing API design. I'll need some reviews to gather thoughts on how best to implement this and make available with most, if not all, diffusers or diffusers-like video model implementations.

Currently, I've only tested with LTX Video.

```python
import torch
from diffusers.hooks import apply_enhance_a_video, EnhanceAVideoConfig

set_verbosity_debug()

pipe.to("cuda")

num_frames = 161
latent_num_frames = (num_frames - 1) // pipe.vae_temporal_compression_ratio + 1
config = EnhanceAVideoConfig(weight=1.0, num_frames_callback=lambda: latent_num_frames, _attention_type=1)
apply_enhance_a_video(pipe.transformer, config)

video = pipe(
    prompt=prompt,
    width=704,
    height=480,
    num_frames=161,
    num_inference_steps=50,
).frames[0]
export_to_video(video, "output.mp4", fps=24)
```

-----

The intended effect of Enhance-A-Video does not seem to be applied yet, as outputs with & without it are the same. I did some quick debugging and it seems like the enhance scores are always `1`. This leads to no effect on the `hidden_states * scores` that are returned from the attention block. Will need to investigate with authors if I'm doing something wrong.

cc @yangluo7 @oahzxl @kaiwang960112

### DIFF
--- a/src/diffusers/hooks/enhance_a_video.py
+++ b/src/diffusers/hooks/enhance_a_video.py
@@ -28,7 +28,6 @@ logger = get_logger(__name__)
 
 
 _ENHANCE_A_VIDEO = "enhance_a_video"
-_ENHANCE_A_VIDEO_SDPA = "enhance_a_video_sdpa"
 
 
 class _AttentionType(Enum):
@@ -188,4 +187,13 @@ def apply_enhance_a_video(module: torch.nn.Module, config: EnhanceAVideoConfig) 
             num_frames_callback=config.num_frames_callback,
             _attention_type=config._attention_type,
         )
-        hook_registry.register_hook(hook, _ENHANCE_A_VIDEO_SDPA)
+        hook_registry.register_hook(hook, _ENHANCE_A_VIDEO)
+
+
+def remove_enhance_a_video(module: torch.nn.Module) -> None:
+    for name, submodule in module.named_modules():
+        if not hasattr(submodule, "_diffusers_hook"):
+            continue
+        hook_registry = submodule._diffusers_hook
+        hook_registry.remove_hook(_ENHANCE_A_VIDEO, recurse=False)
+        logger.debug(f"Removed Enhance-A-Video from layer '{name}'")
--- a/src/diffusers/hooks/enhance_a_video.py
+++ b/src/diffusers/hooks/enhance_a_video.py
@@ -1,0 +1,191 @@
+# Copyright 2024 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Callable
+
+import torch
+import torch.overrides
+
+from ..utils import get_logger
+from ._common import _ATTENTION_CLASSES
+from .hooks import HookRegistry, ModelHook
+
+
+logger = get_logger(__name__)
+
+
+_ENHANCE_A_VIDEO = "enhance_a_video"
+_ENHANCE_A_VIDEO_SDPA = "enhance_a_video_sdpa"
+
+
+class _AttentionType(Enum):
+    SELF = 0
+    JOINT___LATENTS_FIRST = 1
+    JOINT___LATENTS_LAST = 2
+
+
+@dataclass
+class EnhanceAVideoConfig:
+    r"""
+    Configuration for [Enhance A Video](https://huggingface.co/papers/2502.07508).
+    """
+
+    weight: float = 1.0
+    num_frames_callback: Callable[[], int] = None
+    _attention_type: _AttentionType = _AttentionType.SELF
+
+
+class EnhanceAVideoAttentionState:
+    def __init__(self) -> None:
+        self.query: torch.Tensor = None
+        self.key: torch.Tensor = None
+        self.latents_sequence_length: int = None
+
+    def reset(self) -> None:
+        self.query = None
+        self.key = None
+        self.latents_sequence_length = None
+
+    def __repr__(self):
+        return f"EnhanceAVideoAttentionState(scores={self.scores}, latents_sequence_length={self.latents_sequence_length})"
+
+
+class EnhanceAVideoCaptureSDPAInputsFunctionMode(torch.overrides.TorchFunctionMode):
+    def __init__(self, query_key_save_callback: Callable[[torch.Tensor, torch.Tensor], None]) -> None:
+        super().__init__()
+
+        self.query_key_save_callback = query_key_save_callback
+
+    def __torch_function__(self, func, types, args=(), kwargs=None):
+        if kwargs is None:
+            kwargs = {}
+        if func is torch.nn.functional.scaled_dot_product_attention:
+            query = kwargs.get("query", None) or args[0]
+            key = kwargs.get("key", None) or args[1]
+            self.query_key_save_callback(query, key)
+        return func(*args, **(kwargs or {}))
+
+
+class EnhanceAVideoSDPAHook(ModelHook):
+    _is_stateful = True
+
+    def __init__(self, weight: float, num_frames_callback: Callable[[], int], _attention_type: _AttentionType) -> None:
+        self.weight = weight
+        self.num_frames_callback = num_frames_callback
+        self._attention_type = _attention_type
+
+    def initialize_hook(self, module):
+        self.state = EnhanceAVideoAttentionState()
+        return module
+
+    def new_forward(self, module, *args, **kwargs):
+        # Here, query and key have two shapes (considering the general diffusers-style model implementation):
+        #   1. [batch_size, attention_heads, latents_sequence_length, head_dim]
+        #   2. [batch_size, attention_heads, latents_sequence_length + encoder_sequence_length, head_dim]
+        #   3. [batch_size, attention_heads, encoder_sequence_length + latents_sequence_length, head_dim]
+        kwargs_hidden_states = kwargs.get("hidden_states", None)
+        hidden_states = kwargs_hidden_states if kwargs_hidden_states is not None else args[0]
+        self.state.latents_sequence_length = hidden_states.size(2)
+
+        # Capture query and key tensors to compute EnhanceAVideo scores
+        with EnhanceAVideoCaptureSDPAInputsFunctionMode(self._query_key_capture_callback):
+            return self.fn_ref.original_forward(*args, **kwargs)
+
+    def post_forward(self, module, output):
+        # For diffusers models, or ones that are implemented similar to our design, we either return:
+        #   1. A single output: `hidden_states`
+        #   2. A tuple of outputs: `(hidden_states, encoder_hidden_states)`.
+        # We need to handle both cases of applying EnhanceAVideo scores.
+        hidden_states = output[0] if isinstance(output, tuple) else output
+
+        def reshape_for_framewise_attention(tensor: torch.Tensor) -> torch.Tensor:
+            # This code assumes tensor is [B, N, S, C]. This should be true for most diffusers-style implementations.
+            # [B, N, S, C] -> [B, N, F, S, C] -> [B, S, N, F, C] -> [B * S, N, F, C]
+            return tensor.unflatten(2, (num_frames, -1)).permute(0, 3, 1, 2, 4).flatten(0, 1)
+
+        # Handle reshaping of query and key tensors
+        query, key = self.state.query, self.state.key
+        if self._attention_type == _AttentionType.SELF:
+            pass
+        elif self._attention_type == _AttentionType.JOINT___LATENTS_FIRST:
+            query = query[:, :, : self.state.latents_sequence_length]
+            key = key[:, :, : self.state.latents_sequence_length]
+        elif self._attention_type == _AttentionType.JOINT___LATENTS_LAST:
+            query = query[:, :, -self.state.latents_sequence_length :]
+            key = key[:, :, -self.state.latents_sequence_length :]
+
+        num_frames = self.num_frames_callback()
+        query = reshape_for_framewise_attention(query)
+        key = reshape_for_framewise_attention(key)
+        scores = enhance_a_video_score(query, key, num_frames, self.weight)
+        print("Applying scores:", scores)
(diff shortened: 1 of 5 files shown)


src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,191 @@
...
+    attn_wo_diag = attn_temp.masked_fill(diag_mask, 0)
+
+    # Calculate mean for each token's attention matrix
+    # Number of off-diagonal elements per matrix is n*n - n
+    num_off_diag = num_frames * num_frames - num_frames
+    mean_scores = attn_wo_diag.sum(dim=(1, 2)) / num_off_diag
+
+    scores = mean_scores.mean() * (num_frames + weight)
+    scores = scores.clamp(min=1)
+    return scores
```
a-r-r-o-w, line 176: @yangluo7 @oahzxl `scores` here is always 1 with many different inputs that I tried. I've copied this part from the original implementation: https://github.com/NUS-HPC-AI-Lab/Enhance-A-Video/blob/088d9e047b1738a45a253fd7cbe37fdf8526fb97/enhance_a_video/enhance.py

Am I doing something incorrect here or elsewhere? Thanks for your time!

src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # Zero out diagonal elements
+    attn_wo_diag = attn_temp.masked_fill(diag_mask, 0)
+
+    # Calculate mean for each token's attention matrix
+    # Number of off-diagonal elements per matrix is n*n - n
+    num_off_diag = num_frames * num_frames - num_frames
+    mean_scores = attn_wo_diag.sum(dim=(1, 2)) / num_off_diag
+
+    scores = mean_scores.mean() * (num_frames + weight)
```
a-r-r-o-w, line 174: Also, should the mean be taken across all dimensions? I think it might be incorrect since each batch of data should have a different score due to different conditioning. Since we concatenate both unconditional and conditional branches and run batched inference, I believe this should be `mean_scores.mean(list(range(1, mean_scores.ndim)))`. This will give us a tensor of shape `(B,)`, which will also be compatible for multiplication and seems more correct to me

src/diffusers/hooks/enhance_a_video.py
yangluo7: @a-r-r-o-w I noticed you set the enhance weight as 1 in "config = EnhanceAVideoConfig(weight=1.0, num_frames_callback=lambda: latent_num_frames, _attention_type=1)." Maybe it is too small to affect the final output. In our experiments, the weight is at least 5 for LTX-Video with the setting "width=768, height=512, num_frames=121." Thanks a lot!

a-r-r-o-w: Thank you! I see it taking effect now :)

yangluo7: The enhance weight is the only introduced parameter in our proposed method, it is affected by several factors including num_frames and prompts, so it needs to be further tuned based on them. We sincerely thank you for incorporating our method into diffusers, which makes our work more accessible to the community :)

src/diffusers/hooks/enhance_a_video.py
yangluo7: Thanks for your advice. We have tried this implementation before by calculating the mean score for each branch separately but found no obvious difference in the final output. As a result, we chose a more concise implementation by calculating the mean score together.

tin2tin:
I tested this. The first thing to note is that the models isn't released from memory when the inference has finished. 

I used this prompt, captioned by chatGPT, imitating the style of the prompt above:

> A young girl with curly brown hair sits on the floor, wearing a bright yellow dress. A small yellow hair clip holds back some of her curls. She is drawing on a large piece of cardboard propped up like an easel, creating a colorful scene of flowers, grass, and a sun with markers. Sunlight streams through a window behind her, casting a warm glow on the wooden floor. Her bare feet are tucked under her as she leans forward, focused on her artwork. Various markers are scattered around her. The setting appears to be a cozy living room with soft, natural lighting. The scene appears to be real-life footage.

I used the code above, but changed the weight value.

This is with no enhance:

https://github.com/user-attachments/assets/e066e708-f83f-4c12-8024-946a385245cf

Weight: 2

https://github.com/user-attachments/assets/1cb0e9e3-cadf-423b-8f78-56ac2197347c

Weight: 3

https://github.com/user-attachments/assets/b50ca353-af9f-4b21-b116-b205678fdf46

Weight: 5 

https://github.com/user-attachments/assets/999aeaff-98ef-4794-8b85-fd9139e25a3c

Weight: 7

https://github.com/user-attachments/assets/3a710bdd-976d-43e9-bea2-401e162a5fb2

Weight: 10

https://github.com/user-attachments/assets/badc11d7-9143-4463-9967-7609090f19ef

Weight: 15

https://github.com/user-attachments/assets/775f18a4-ce73-4c0d-8158-adacccd0a51d

They all have deformities, so it is kind of hard to conclude anything on (is 15 best, but the colors start to fry?), but maybe it's because of the LTX video being wonky from the start (personally, I find it very hard to get LTX to produce anything without severe deformities, when not using the default prompt).  So, it is clear with the current implementation, that the enhance function is doing something, but it is not super clear to me if it is enough to save poor input material?  Would it clean up the deformities it run twice?

yangluo7:
Thanks for the testing. Firstly, Enhance-A-Video improves the generated video's quality based on the foundation models' existing attention weights and makes moderate adjustments in the residual connection, so the final video quality still relies on the existing generative quality of the foundation model itself. If the original generated video quality is quite low, it is hard to generate an ideal video without improving the pre-training phase. Secondly, we can find that the video quality is improved with weights between 8-10, which demonstrates the effectiveness of Enhance-A-Video.

src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,199 @@
...
+        self.num_frames_callback = num_frames_callback
+        self._attention_type = _attention_type
+
+    def initialize_hook(self, module):
+        self.state = EnhanceAVideoAttentionState()
+        return module
+
+    def new_forward(self, module, *args, **kwargs):
+        # Here, query and key have two shapes (considering the general diffusers-style model implementation):
+        #   1. [batch_size, attention_heads, latents_sequence_length, head_dim]
```
yiyixuxu, line 118: can we start to automatically test and enforce this (make sure the case for all new models we implemented)?
OmniGen almost did not follow this, and it was not always easy to spot such things

src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,199 @@
...
+    def post_forward(self, module, output):
+        # For diffusers models, or ones that are implemented similar to our design, we either return:
+        #   1. A single output: `hidden_states`
+        #   2. A tuple of outputs: `(hidden_states, encoder_hidden_states)`.
+        # We need to handle both cases of applying EnhanceAVideo scores.
+        hidden_states = output[0] if isinstance(output, tuple) else output
+
+        def reshape_for_framewise_attention(tensor: torch.Tensor) -> torch.Tensor:
+            # This code assumes tensor is [B, N, S, C]. This should be true for most diffusers-style implementations.
+            # [B, N, S, C] -> [B, N, F, S, C] -> [B, S, N, F, C] -> [B * S, N, F, C]
```
yiyixuxu, line 138: same here
can we start to enforce this?

yiyixuxu (approved):
design looks good to me 
I think we should start to add test to enforce certain styles now we want to expand and encourage the usage of hooks

src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@dataclass
+class EnhanceAVideoConfig:
+    r"""
+    Configuration for [Enhance A Video](https://huggingface.co/papers/2502.07508).
+    """
+
+    weight: float = 1.0
+    num_frames_callback: Callable[[], int] = None
```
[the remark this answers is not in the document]
a-r-r-o-w: So... there's no easy way to determine this. Some models use dim=1 as frame dimension, whereas some do it in dim=2. Some models don't do this at all, for example LTX Video already flattens the FHW dimension before the transformer forward.

The information about number of latent frames is only available in the model transformer. Even then, sometimes it is modified by a patch embedding layer -- we don't know for sure, in general case, how to determine number of frames being used for inference.

In the Attention block where we attach hooks, the dimension of tensors are `[B, S, D]`, we don't have access to this info either.

The only source for accurately getting this information is the user :( I'm open to suggestions and holding on to the PR for longer if we can figure out better way to do this

src/diffusers/hooks/enhance_a_video.py
a-r-r-o-w: Sounds good. I can add some tests soon to enforce this on any model that is added

src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,302 @@
...
+                - `_AttentionType.JOINT___LATENTS_LAST`:
+                    The model uses joint attention layers (concatenated video and text stream data) with video tokens
+                    last.
+
+            This parameter is not backwards-compatible and may be subject to change in future versions.
+    """
+
+    weight: Union[float, Dict[str, float]] = 1.0
+    num_frames_callback: Callable[[], int] = None
+    _attention_type: _AttentionType = _AttentionType.SELF
```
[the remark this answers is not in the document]
a-r-r-o-w: We currently only support dynamically updating these values if user first removes all hooks by calling `remove_enhance_a_video` and then doing `apply_enhance_a_video` again. It's, uh, not really ideal but is a lightweight operation so we can get away with it.

Alternatively, to update dynamically, do you think we should do this:
- when user calls `apply_enhance_a_video`, we return them some kind of handle object that has knowledge about the hooks
- they can call a `set_weight` and `set_frames` method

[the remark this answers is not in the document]
a-r-r-o-w: Also, perhaps we don't need the `_attention_type` argument. I can define a simple dictionary in the `_common.py` file that categorizes each transformer into the three groups -- I think this is good info to have for some other methods that we could integrate soon

a-r-r-o-w:
@tin2tin Our LTX implementation is missing a few of the latest features implemented in the original repository to improve generation quality. This will be improved soon as I find some time to work on it. I added support for adding weight factor per block (you can specify a dictionary of regex pattern mapping to weight values), so you can play around a bit and see what layers are best suited for applying the method -- from my testing, applying on blocks between 5-20 seem to work best and do not modify predictions too much

a-r-r-o-w:
@yangluo7 Would you be able to give a final review as well for correctness check? The implementation will not change much after my latest commit, apart from docs/tests, so is more or less finalized. Thanks!

src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,302 @@
...
+
+    def initialize_hook(self, module):
+        self.state = EnhanceAVideoAttentionState()
+        return module
+
+    def new_forward(self, module, *args, **kwargs):
+        # Here, query and key have two shapes (considering the general diffusers-style model implementation):
+        #   1. [batch_size, attention_heads, latents_sequence_length, head_dim]
+        #   2. [batch_size, attention_heads, latents_sequence_length + encoder_sequence_length, head_dim]
+        #   3. [batch_size, attention_heads, encoder_sequence_length + latents_sequence_length, head_dim]
```
a-r-r-o-w, line 120:
```suggestion
        #   1. [batch_size, latents_sequence_length, embedding_dim]
        #   2. [batch_size, latents_sequence_length + encoder_sequence_length, embedding_dim]
        #   3. [batch_size, encoder_sequence_length + latents_sequence_length, embedding_dim]
```

src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,302 @@
...
+        self.weight = weight
+        self.num_frames_callback = num_frames_callback
+        self._attention_type = _attention_type
+
+    def initialize_hook(self, module):
+        self.state = EnhanceAVideoAttentionState()
+        return module
+
+    def new_forward(self, module, *args, **kwargs):
+        # Here, query and key have two shapes (considering the general diffusers-style model implementation):
```
a-r-r-o-w, line 117:
```suggestion
        # Here, hidden_states could have three shapes (considering the general diffusers-style model implementation):
```

src/diffusers/hooks/enhance_a_video.py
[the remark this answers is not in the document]
yiyixuxu: oh that's fine,
I was just wondering why it is a function, not a constant

src/diffusers/hooks/enhance_a_video.py
```diff
@@ -0,0 +1,302 @@
...
+        hook_registry = HookRegistry.check_if_exists_or_initialize(submodule)
+        hook = EnhanceAVideoSDPAHook(
+            weight=current_weight,
+            num_frames_callback=config.num_frames_callback,
+            _attention_type=config._attention_type,
+        )
+        hook_registry.register_hook(hook, _ENHANCE_A_VIDEO)
+
+
+def remove_enhance_a_video(module: torch.nn.Module) -> None:
```
yiyixuxu, line 279: do we have a method to call when we want to remove all the model hooks on a model?

src/diffusers/hooks/enhance_a_video.py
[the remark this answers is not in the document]
yiyixuxu: remove_enhance_a_video is fine I think,! 
we can wait for a couple more use cases to decide how to support 
(I don't like the set_weight and set_frame method because it's specific for each config name, I think we need something more generic)

src/diffusers/hooks/enhance_a_video.py
a-r-r-o-w: Not yet :( 

We can only remove hooks one at a time: https://github.com/huggingface/diffusers/blob/f8b54cf0373b031a72861bb99e0e3646a83cf31f/src/diffusers/hooks/hooks.py#L179 if the user knows the name (can be found by printing the registry)

I can add a method that allows removing all hooks if you'd like, LMK

DN6: Hmm might be good to add a methods such as 
`enable_hook`, `disable_hook`, `disable_all_hooks` etc to ModelMixin.

a-r-r-o-w: @DN6 Sounds good. I think we should add those first, so will hold off merging here and open a PR for that first

PR #10822, opened 2025-02-18
Fix weird interaction between disk offload and group offload tests

a-r-r-o-w:
Not quite sure what causes the following test failures: https://github.com/huggingface/diffusers/actions/runs/13384935248/job/37379809841#step:6:10497

From some debugging, the following seems to be happening:
- `test_disk_offload_without_safetensors` and `test_disk_offload_with_safetensors` runs first. This adds Accelerate hooks to handle device map correctly on the model
- When we run group offloading tests, for some reason, all the new instances of the models that are created contain Accelerate hooks as well.

This makes me believe Accelerate is applying hooks at the class-level instead of the instance-level (I'm not quite sure yet & will look into accelerate code as soon as I can).

I've added a new test (just for repro purposes) that shows the above behaviour consistently happens on some models. But for some models, it works without problems 🤷‍♂️ 

```shell
pytest -s tests/models/transformers/test_models_transformer_cogvideox.py -k test_error_when_disk_offload_run_together_with_group_offloading
```

```shell
FAILED tests/models/transformers/test_models_transformer_cogvideox.py::CogVideoX1_5TransformerTests::test_error_when_disk_offload_run_together_with_group_offloading - ValueError: Cannot apply group offloading to a module that is already applying an alternative offloading strategy from Accelerate. If you want to apply group offloading, please disable the existing offloading strategy first. Offending module: time_embedding.act (<class 'torch.nn.modules.activation.SiLU'>)
```

cc @SunMarc @DN6

### DIFF
--- a/tests/models/test_modeling_common.py
+++ b/tests/models/test_modeling_common.py
@@ -1507,6 +1507,49 @@ class ModelTesterMixin:
         self.assertTrue(torch.allclose(output_without_group_offloading, output_with_group_offloading3, atol=1e-5))
         self.assertTrue(torch.allclose(output_without_group_offloading, output_with_group_offloading4, atol=1e-5))
 
+    def test_error_when_disk_offload_run_together_with_group_offloading(self):
+        config, inputs_dict = self.prepare_init_args_and_inputs_for_common()
+        model1 = self.model_class(**config).eval()
+        model1 = model1.to(torch_device)
+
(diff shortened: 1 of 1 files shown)


SunMarc:
The above PR should fix your issue ! Feel free to merge this PR so that we don't have the same issue again.

PR #10835, opened 2025-02-19
Some consistency-related fixes for HunyuanVideo

a-r-r-o-w:


src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video.py
```diff
@@ -624,13 +624,12 @@ def __call__(
...
         # 5. Prepare latent variables
         num_channels_latents = self.transformer.config.in_channels
-        num_latent_frames = (num_frames - 1) // self.vae_scale_factor_temporal + 1
         latents = self.prepare_latents(
             batch_size * num_videos_per_prompt,
             num_channels_latents,
             height,
             width,
-            num_latent_frames,
+            num_frames,
```
yiyixuxu, line 632: is this change just for make code is less confusing? (we were passing latents_frame as num_frames before)

a-r-r-o-w: So, in prepare_latents, we calculate the latent height and latent width from the actual height/width BUT expect the num_frames to mean number of latent frames.

With this change, we make the interface uniform. prepare_latents now takes pixel-space width/height/num_frames and performs the latent related calculations in itself. We need to make these interfaces uniform a bit otherwise it leads to inconvenient behavior, so I will try to keep this behaviour consistent across all new implementations and gradually introduce changes to other pipelines

Issue #10797, opened 2025-02-15
Group_offloading with HunyuanVideoPipeline not working


nitinmukesh:
### Describe the bug

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument index in method wrapper_CUDA__index_select)

### Reproduction

```

import json
import time
import torch
import gc
from diffusers.models import HunyuanVideoTransformer3DModel
from diffusers.utils import export_to_video
from diffusers import HunyuanVideoPipeline
from diffusers import BitsAndBytesConfig
from transformers import LlamaModel
from diffusers.hooks import apply_group_offloading

model_id = "hunyuanvideo-community/HunyuanVideo"

quantization_config = BitsAndBytesConfig(load_in_4bit=True, bnb_4bit_quant_type="nf4", bnb_4bit_compute_dtype=torch.bfloat16)
text_encoder = LlamaModel.from_pretrained(
	model_id, 
	subfolder="text_encoder", 
	quantization_config=quantization_config,
	torch_dtype=torch.float16
)
transformer = HunyuanVideoTransformer3DModel.from_pretrained(
	model_id, 
	subfolder="transformer",
	quantization_config=quantization_config,
	torch_dtype=torch.bfloat16
)
pipe = HunyuanVideoPipeline.from_pretrained(
	model_id, 
        text_encoder=text_encoder,
	transformer=transformer, 
	torch_dtype=torch.float16, 
)

apply_group_offloading(
    pipe.transformer,
    offload_type="leaf_level",
    offload_device=torch.device("cpu"),
    onload_device=torch.device("cuda"),
)
apply_group_offloading(
    pipe.text_encoder, 
    offload_device=torch.device("cpu"),
    onload_device=torch.device("cuda"),
    offload_type="leaf_level"
)
apply_group_offloading(
    pipe.vae, 
    offload_device=torch.device("cpu"),
    onload_device=torch.device("cuda"),
    offload_type="leaf_level"
)

prompt="A cat wearing sunglasses and working as a lifeguard at pool."

generator = torch.Generator().manual_seed(181201)
output = pipe(
    prompt,
	width=512,
	height=320,
	num_frames=17,
	num_inference_steps=30,
    generator=generator
)[0]

export_to_video(output, "hunyuan_test.mp4", fps=8)

```

### Log

```
(venv) C:\aiOWN\diffuser_webui>python hunyuan_test.py
[...]

PR #10847, opened 2025-02-20
[docs] Flux group offload

stevhliu:
From discussion in #10840, this PR adds an example of group offloading to the Flux docs as well as a note on memory requirements.

### DIFF
--- a/docs/source/en/api/pipelines/flux.md
+++ b/docs/source/en/api/pipelines/flux.md
@@ -355,8 +355,65 @@ image.save('flux_ip_adapter_output.jpg')
     <figcaption class="mt-2 text-sm text-center text-gray-500">IP-Adapter examples with prompt "wearing sunglasses"</figcaption>
 </div>
 
+## Optimize
 
-## Running FP16 inference
+Flux is a very large model and requires ~50GB of RAM. Enable some of the optimizations below to lower the memory requirements.
+
+### Group offloading
+
+[Group offloading](../../optimization/memory#group-offloading) saves memory by offloading groups of internal layers rather than the whole model or weights. Use [`~hooks.apply_group_offloading`] on a model and you can optionally specify the `offload_type`. Setting it to `leaf_level` offloads the lowest leaf-level parameters to the CPU instead of offloading at the module-level.
+
+```py
+import torch
+from diffusers import FluxPipeline
+from diffusers.hooks import apply_group_offloading
+
+model_id = "black-forest-labs/FLUX.1-dev"
+dtype = torch.bfloat16
+pipe = FluxPipeline.from_pretrained(
+	model_id,
+	torch_dtype=dtype,
+)
+
+apply_group_offloading(
+    pipe.transformer,
+    offload_type="leaf_level",
+    offload_device=torch.device("cpu"),
+    onload_device=torch.device("cuda"),
+)
+apply_group_offloading(
+    pipe.text_encoder, 
+    offload_device=torch.device("cpu"),
+    onload_device=torch.device("cuda"),
+    offload_type="leaf_level"
+)
+apply_group_offloading(
+    pipe.text_encoder_2, 
+    offload_device=torch.device("cpu"),
+    onload_device=torch.device("cuda"),
+    offload_type="leaf_level"
+)
+apply_group_offloading(
+    pipe.vae, 
+    offload_device=torch.device("cpu"),
+    onload_device=torch.device("cuda"),
+    offload_type="leaf_level"
+)
+
+prompt="A cat wearing sunglasses and working as a lifeguard at pool."
+
+generator = torch.Generator().manual_seed(181201)
+image = pipe(
+    prompt,
+	width=576,
+	height=1024,
+	num_inference_steps=30,
+    generator=generator
+).images[0]
+image
+```
+
+### Running FP16 inference
 
 Flux can generate high-quality images with FP16 (i.e. to accelerate inference on Turing/Volta GPUs) but produces different outputs compared to FP32/BF16. The issue is that some activations in the text encoders have to be clipped when running in FP16, which affects the overall image. Forcing text encoders to run with FP32 inference thus removes this output difference. See [here](https://github.com/huggingface/diffusers/pull/9097#issuecomment-2272292516) for details.
 
@@ -385,7 +442,7 @@ out = pipe(
 out.save("image.png")
 ```
 
-## Quantization
+### Quantization
 
 Quantization helps reduce the memory requirements of very large models by storing model weights in a lower precision data type. However, quantization may have varying impact on video quality depending on the video model.
 
--- a/docs/source/en/api/pipelines/flux.md
+++ b/docs/source/en/api/pipelines/flux.md
@@ -357,11 +357,16 @@ image.save('flux_ip_adapter_output.jpg')
 
 ## Optimize
 
-Flux is a very large model and requires ~50GB of RAM. Enable some of the optimizations below to lower the memory requirements.
+Flux is a very large model and requires ~50GB of RAM/VRAM to load all the modeling components. Enable some of the optimizations below to lower the memory requirements.
 
 ### Group offloading
 
-[Group offloading](../../optimization/memory#group-offloading) saves memory by offloading groups of internal layers rather than the whole model or weights. Use [`~hooks.apply_group_offloading`] on a model and you can optionally specify the `offload_type`. Setting it to `leaf_level` offloads the lowest leaf-level parameters to the CPU instead of offloading at the module-level.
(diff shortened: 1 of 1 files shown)


docs/source/en/api/pipelines/flux.md
```diff
@@ -355,8 +355,65 @@ image.save('flux_ip_adapter_output.jpg')
     <figcaption class="mt-2 text-sm text-center text-gray-500">IP-Adapter examples with prompt "wearing sunglasses"</figcaption>
 </div>
 
+## Optimize
 
-## Running FP16 inference
+Flux is a very large model and requires ~50GB of RAM. Enable some of the optimizations below to lower the memory requirements.
```
asomoza, line 360: nice!, but the 50 GB of RAM are used when using group offloading not before, also @a-r-r-o-w was going to check if this is the real number or not, I get this but maybe there's something in my env that makes it go that high. It should use around 20GB for the transformer model in theory.

stevhliu: Ah ok, I'll update this number once we get a clearer value from @a-r-r-o-w!

nitinmukesh:
May I request to add all 3 examples as they cover different aspects of this feature.
1st example covers leaf_level
3rd example covers block_level + leaf_level (demonstrates mix is allowed for different components)
2nd example explains that you need to apply on all components otherwise there will be error " Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0!". Another solution is to use .to() for component for which you don't want to use apply_group_offloading. This scenario/issue is covered here  https://github.com/huggingface/diffusers/issues/10797

*component means text_encoder(s), transformer, vae

a-r-r-o-w:
@asomoza @stevhliu So, I looked into the CPU memory usage, and it is indeed higher for group offloading compared to model/sequential offloading. To improve, I'll need @SunMarc's help in understanding what's going on, since it looks like a component remains on disk (is not loaded on to the CPU) when not required -- otherwise memory usage should have been much higher from just loading model weights.

Code: https://gist.github.com/a-r-r-o-w/f5c9fb5c515d24f9a06001adb5c6cf18

| Configuration                 | Time (s) | CUDA Model Memory (GB) | CUDA Inference Memory (GB) | CPU Offload Memory (GB) | CPU Inference Memory (GB) |
|--------------------------------|---------|-------------------------|----------------------------|-------------------------|--------------------------|
| full_cuda                     | 25.77   | 31.45                   | 36.07                      | 0.86                    | 1.41                     |
| model_offload                 | 230.95  | 0.0                      | 23.22                      | 0.8                     | 10.51                    |
| sequential_offload            | 2660.6  | 0.0                      | 2.4                        | 0.92                    | 32.69                    |
| group_offload_block_1         | 306.01  | 0.17                     | 13.41                      | 0.91                    | 37.55                    |
| group_offload_leaf            | 375.29  | 0.17                     | 4.56                       | 0.92                    | 37.41                    |
| group_offload_block_1_stream  | 58.79   | 0.17                     | 14.49                      | 47.99                   | 57.84                    |
| group_offload_leaf_stream     | 55.26   | 0.17                     | 5.52                       | 47.99                   | 48.54                    |

<sup>For group offloading, we do not offload VAE since the `forward` is never invoked (which is required to trigger the hooks) as we call `decode`. So the numbers are not really a fair comparision, but I can benchmark properly again if someone needs.</sup>

So, as Alvaro pointed out, we do require a lot of RAM. There are a few ways we can reduce the RAM requirements in the near future:
- Applying disk offloading in combination with group offloading or whatever sorcery is happening in `model_offload` :eyes:
- Allowing loading in `torch.float8_*` types

We should definitely mention the limitations for now, but revisit once improvements have been made. There is slightly higher CPU usage compared to sequential offloading because we require pinned memory tensors to allow streams to be used. The [docs](https://pytorch.org/tutorials/intermediate/pinmem_nonblock.html) explain this:

> In general, the transfer is blocking on the device side (even if it isn’t on the host side): the copy on the device cannot occur while another operation is being executed. However, in some advanced scenarios, a copy and a kernel execution can be done simultaneously on the GPU side. As the following example will show, three requirements must be met to enable this:
>
> The device must have at least one free DMA (Direct Memory Access) engine. Modern GPU architectures such as Volterra, Tesla, or H100 devices have more than one DMA engine.
> 
> The transfer must be done on a separate, non-default cuda stream. In PyTorch, cuda streams can be handles using [Stream](https://pytorch.org/docs/stable/generated/torch.cuda.Stream.html#torch.cuda.Stream).
>
> The source data must be in pinned memory.

docs/source/en/api/pipelines/flux.md
a-r-r-o-w, line 360:
```suggestion
Flux is a very large model and requires ~50GB of RAM/VRAM to load all modeling components. Enable some of the optimizations below to lower the memory requirements.
```

docs/source/en/api/pipelines/flux.md
```diff
@@ -355,8 +355,65 @@ image.save('flux_ip_adapter_output.jpg')
...
+
+### Group offloading
+
+[Group offloading](../../optimization/memory#group-offloading) saves memory by offloading groups of internal layers rather than the whole model or weights. Use [`~hooks.apply_group_offloading`] on a model and you can optionally specify the `offload_type`. Setting it to `leaf_level` offloads the lowest leaf-level parameters to the CPU instead of offloading at the module-level.
```
a-r-r-o-w, line 364:
```suggestion
[Group offloading](../../optimization/memory#group-offloading) lowers VRAM usage by offloading groups of internal layers rather than the whole model or weights. Use [`~hooks.apply_group_offloading`] on a model and you can optionally specify the `offload_type`. Setting it to `leaf_level` offloads the lowest leaf-level parameters to the CPU instead of offloading at the module-level.
```

docs/source/en/api/pipelines/flux.md
```diff
@@ -355,8 +355,65 @@ image.save('flux_ip_adapter_output.jpg')
...
+from diffusers.hooks import apply_group_offloading
+
+model_id = "black-forest-labs/FLUX.1-dev"
+dtype = torch.bfloat16
+pipe = FluxPipeline.from_pretrained(
+	model_id,
+	torch_dtype=dtype,
+)
+
+apply_group_offloading(
```
a-r-r-o-w, line 378: Group offloading in itself is not better than sequential cpu offloading from accelerate. Both are slow but I'm not sure why I got sequential offload time numbers to be much higher than group leaf-level. They are the same in theory so I will cc @SunMarc to help take a look as well

a-r-r-o-w: If we're showcasing group offloading here, I think there's a strong case to show the usage of `use_stream=True`, or mention it somewhere. It speeds up inference quite a lot!

docs/source/en/api/pipelines/flux.md
```diff
@@ -355,8 +355,65 @@ image.save('flux_ip_adapter_output.jpg')
...
+
+prompt="A cat wearing sunglasses and working as a lifeguard at pool."
+
+generator = torch.Generator().manual_seed(181201)
+image = pipe(
+    prompt,
+	width=576,
+	height=1024,
+	num_inference_steps=30,
+    generator=generator
```
a-r-r-o-w, line 411:
```suggestion
image = pipe(
    prompt,
	  width=576,
	  height=1024,
	  num_inference_steps=30,
    generator=generator
```

a-r-r-o-w (approved):
Thanks @stevhliu! Just some thoughts

shethaadit (approved):

stevhliu:
> May I request to add all 3 examples as they cover different aspects of this feature.

I added a note that should cover the other 2 examples (mixing `block` and `leaf` offloading, and using `apply_group_offloading` to all components). I think adding the full code examples for the other 2 might be a bit much when you're only making small changes to the code.

I'll update the exact RAM/VRAM requirements later in a separate PR pending @a-r-r-o-w's investigation :)

a-r-r-o-w:
@stevhliu It is indeed 50 GB at the moment for group offloading as mentioned in the table in my previous comment :( I'll work with Marc on improving this by understanding what accelerate does for model offloading + we might be directly supporting this in accelerate

docs/source/en/api/pipelines/flux.md
```diff
@@ -355,8 +355,74 @@ image.save('flux_ip_adapter_output.jpg')
     <figcaption class="mt-2 text-sm text-center text-gray-500">IP-Adapter examples with prompt "wearing sunglasses"</figcaption>
 </div>
 
+## Optimize
 
-## Running FP16 inference
+Flux is a very large model and requires ~50GB of RAM/VRAM to load all the modeling components. Enable some of the optimizations below to lower the memory requirements.
+
+### Group offloading
```
sayakpaul, line 362: @stevhliu it might be a good idea to consider making it more generally available to all major pipelines with high usage.

PR #10850, opened 2025-02-21
Add helper utility methods for interacting with diffusers hooks

a-r-r-o-w:
Context: https://github.com/huggingface/diffusers/pull/10815#discussion_r1963988999


a-r-r-o-w:
Or would you alternatively prefer a context manager solution? Something like:

```python
with model.disable_hooks([...]):
    model(...)
```

If that, we can probably support both context manager and method-based approach in the same code flow

Issue #10863, opened 2025-02-21
Error loading locally safetensors files


elismasilva:
### Describe the bug

Some models are getting error on load safetensors directly from .safetensors file.

### Reproduction

```python
StableDiffusionXLControlTilingPipeline.from_single_file("F:\\models\\Stable-diffusion\\RealVisXL v5.0 Lightning.safetensors",...)
```

@SunMarc - > metadata is returning {} len = 0 and this condition is not in this check:
![Image](https://github.com/user-attachments/assets/79f33664-b5aa-4f0b-a27f-a678261f8079)

### Logs

```shell
Traceback (most recent call last):
  File "C:\Users\pelol\.pyenv\pyenv-win\versions\3.10.11\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\pelol\.pyenv\pyenv-win\versions\3.10.11\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "c:\Users\pelol\.vscode\extensions\ms-python.debugpy-2025.1.2025021701-win32-x64\bundled\libs\debugpy\__main__.py", line 71, in <module>
    cli.main()
  File "c:\Users\pelol\.vscode\extensions\ms-python.debugpy-2025.1.2025021701-win32-x64\bundled\libs\debugpy/..\debugpy\server\cli.py", line 501, in main
    run()
  File "c:\Users\pelol\.vscode\extensions\ms-python.debugpy-2025.1.2025021701-win32-x64\bundled\libs\debugpy/..\debugpy\server\cli.py", line 351, in run_file
    runpy.run_path(target, run_name="__main__")
  File "c:\Users\pelol\.vscode\extensions\ms-python.debugpy-2025.1.2025021701-win32-x64\bundled\libs\debugpy\_vendored\pydevd\_pydevd_bundle\pydevd_runpy.py", line 310, in run_path
    return _run_module_code(code, init_globals, run_name, pkg_name=pkg_name, script_name=fname)
  File "c:\Users\pelol\.vscode\extensions\ms-python.debugpy-2025.1.2025021701-win32-x64\bundled\libs\debugpy\_vendored\pydevd\_pydevd_bundle\pydevd_runpy.py", line 127, in _run_module_code
    _run_code(code, mod_globals, init_globals, mod_name, mod_spec, pkg_name, script_name)
[...]

PR #10864, opened 2025-02-21
remove format check for safetensors file

SunMarc:
# What does this PR do?

FIxes https://github.com/huggingface/diffusers/issues/10863
This PR removes the format check that was recently introduced as it is not might not be compatible with single file format. For diffusers format there should be no issues. If you think it is still worth to add it for diffusers checkpoints, I can modify a bit the PR.


SunMarc:
Merging this first as this is might impact lots of files

PR #10881, opened 2025-02-23
[refactor] Remove additional Flux code

a-r-r-o-w:
Some minor refactors to make implementations consistent.

- We already check for `F.scaled_dot_product_attention` in the attention processor init method (see [here](https://github.com/huggingface/diffusers/blob/9c7e205176c30b27c5f44ec7650a8dfcc12dde86/src/diffusers/models/attention_processor.py#L2265))
- Removes docstrings from intermediate blocks since they are explained in the main transformer implementation
- Additionally should we consider [this](https://github.com/huggingface/diffusers/pull/9751#issuecomment-2677129361) comment and remove the NPU related logic in FluxSingleTransformerBlock? It might be backwards breaking though

### DIFF
--- a/src/diffusers/models/transformers/transformer_flux.py
+++ b/src/diffusers/models/transformers/transformer_flux.py
@@ -31,7 +31,7 @@ from ...models.attention_processor import (
 )
 from ...models.modeling_utils import ModelMixin
 from ...models.normalization import AdaLayerNormContinuous, AdaLayerNormZero, AdaLayerNormZeroSingle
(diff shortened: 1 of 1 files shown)


yiyixuxu:
> Additionally should we consider https://github.com/huggingface/diffusers/pull/9751#issuecomment-2677129361 comment and remove the NPU related logic in FluxSingleTransformerBlock? It might be backwards breaking though

let's remove it in this PR! 
we could do deprecate if needed, but only for short-time, e.g like  remove in 0.34.

yiyixuxu:
other changes looks good to me!

yiyixuxu (approved):
perfect!

PR #10882, opened 2025-02-23
[refactor] SD3 docs & remove additional code

a-r-r-o-w:
Refactors SD3:
- Removes `F.scaled_dot_product_attention` check which is already done in the processor (see [here](https://github.com/huggingface/diffusers/blob/9c7e205176c30b27c5f44ec7650a8dfcc12dde86/src/diffusers/models/attention_processor.py#L1410))
- Accesses parameters directly instead of via `self.config` for init methods
- Updates docs
- Updates type hints from FloatTensor -> Tensor since the former is deprecated

### DIFF
--- a/src/diffusers/models/transformers/transformer_sd3.py
+++ b/src/diffusers/models/transformers/transformer_sd3.py
@@ -316,24 +316,24 @@ class SD3Transformer2DModel(
 
     def forward(
         self,
-        hidden_states: torch.FloatTensor,
-        encoder_hidden_states: torch.FloatTensor = None,
-        pooled_projections: torch.FloatTensor = None,
+        hidden_states: torch.Tensor,
+        encoder_hidden_states: torch.Tensor = None,
+        pooled_projections: torch.Tensor = None,
         timestep: torch.LongTensor = None,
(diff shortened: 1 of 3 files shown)


yiyixuxu (approved):
thanks!

src/diffusers/models/transformers/transformer_sd3.py
```diff
@@ -39,17 +38,6 @@
 
 @maybe_allow_in_graph
 class SD3SingleTransformerBlock(nn.Module):
-    r"""
```
yiyixuxu, line 42: is this on purpose?

a-r-r-o-w: Indeed, yes! All the latest integrations don't document the internal blocks. This is because of two reasons:
- the parameters are already documented for main transformer, so copying for each internal block is redundant and almost always has introduced mismatched explanations (from doing the refactoring, i sometimes see parameters that don't exist are documented, and sometimes the explanations are just wrong)
- The internal blocks are not user-facing API or linked in main diffusers documentation, so they will not show up there anyway

PR #10908, opened 2025-02-26
[style bot] improve security for the stylebot.

sayakpaul:
As per the internal discussions.


.github/workflows/pr_style_bot.yml
```diff
@@ -64,18 +64,33 @@ jobs:
         run: |
           pip install .[quality]
 
-      - name: Download Makefile from main branch
+      - name: Download necessary files from main branch of Diffusers
         run: |
           curl -o main_Makefile https://raw.githubusercontent.com/huggingface/diffusers/main/Makefile
+          curl -o main_setup.py https://raw.githubusercontent.com/huggingface/diffusers/refs/heads/main/setup.py
+          curl -o main_check_doc_toc.py https://raw.githubusercontent.com/huggingface/diffusers/refs/heads/main/utils/check_doc_toc.py
```
sayakpaul, line 71: We don't leverage any other files during the styling process.

DN6: I think it might also make sense to validate inputs in this step, before running echo. 
https://github.com/huggingface/diffusers/blob/6a478569f0f41b47690fff8bc6c7f6fd6a171591/.github/workflows/pr_style_bot.yml#L50-L58

e.g. 
```
    if [[ ! "$HEADREPOFULLNAME" =~ ^[a-zA-Z0-9_.-]+/[a-zA-Z0-9_.-]+$ ]]; then
      echo "Invalid repository format: $HEADREPOFULLNAME"
      exit 1
    fi

    # check for unusual characters
    if [[ ! "$HEADREF" =~ ^[a-zA-Z0-9_/-]+$ ]]; then
      echo "Invalid branch name: $HEADREF"
      exit 1
    fi
```

DN6: Also is there any risk of the Github Token leaking here? e.g. The step fails and the error trace show the token since it's injected into the URL? 

https://github.com/huggingface/diffusers/blob/9a8e8db79f4752bc32fb3c61d2feed46cebd166e/.github/workflows/pr_style_bot.yml#L98

cc: @glegendre01

DN6 (approved):

Issue #10679, opened 2025-01-29
possibly to avoid `from_single_file` loading in fp32 to save RAM


asomoza:
### Describe the bug

When loading a model using `from_single_file()`, the RAM usage is really high possibly because the weights are loaded in FP32 before conversion.

### Reproduction

```python
import threading
import time

import psutil
import torch
from huggingface_hub import hf_hub_download

from diffusers import UNet2DConditionModel


filename = hf_hub_download("stable-diffusion-v1-5/stable-diffusion-v1-5", filename="v1-5-pruned-emaonly.safetensors")

stop_monitoring = False


def log_memory_usage():
    process = psutil.Process()
    mem_info = process.memory_info()
    return mem_info.rss / (1024**2)  # Convert to MB


def monitor_memory(interval, peak_memory):
    while not stop_monitoring:
        current_memory = log_memory_usage()
        peak_memory[0] = max(peak_memory[0], current_memory)
        time.sleep(interval)


def load_model(filename, dtype):
    global stop_monitoring

    peak_memory = [0]  # Use a list to store peak memory so it can be updated in the thread
    initial_memory = log_memory_usage()
    print(f"Initial memory usage: {initial_memory:.2f} MB")

    monitor_thread = threading.Thread(target=monitor_memory, args=(0.01, peak_memory))
    monitor_thread.start()

    start_time = time.time()
    UNet2DConditionModel.from_single_file(filename, torch_dtype=dtype)
    end_time = time.time()

    stop_monitoring = True
    monitor_thread.join()  # Wait for the monitoring thread to finish

    print(f"Peak memory usage: {peak_memory[0]:.2f} MB")
    print(f"Time taken: {end_time - start_time:.2f} seconds")
    final_memory = log_memory_usage()
    print(f"Final memory usage: {final_memory:.2f} MB")


load_model(filename, torch.float8_e4m3fn)
```

### Logs

```shell
`
Initial memory usage: 737.19 MB
Peak memory usage: 4867.43 MB
Time taken: 0.92 seconds
Final memory usage: 1578.99 MB
`
```

### System Info

not relevant here

### Who can help?

@DN6

yiyixuxu:
@asomoza can you test under this PR? https://github.com/huggingface/diffusers/pull/10604

asomoza:
@yiyixuxu changed to windows and a mobile 4090 since the original issue was with windows.

Without the PR:
```shell
Initial memory usage: 548.29 MB
[...]

PR #10948, opened 2025-03-03
Improve load_ip_adapter RAM Usage

CyberVy:
Loading a model with ```torch_dtype``` is ```None``` using ```transformers.modeling_utils.PreTrainedModel``` will result in additional RAM usage because of data type conversion.
There is no conversion if the data type of the model weights is the same as ```torch_dtype```.
And different conversions have different impacts on RAM usage which is mentioned in #10679.
When ```torch_dtype``` is ```None```, ```PreTrainedModel``` will regard it as ```torch.float32```, which causes a lot more RAM usage.

This PR is to improve this issue.

@asomoza


asomoza:
nice! do you measure the ram savings?

hlky:
@bot /style

CyberVy:
> nice! do you measure the ram savings?

Yep! The peak RAM usage decreases from 8.9G to 2.1G when I load an FP16 SDXL IP Adapter.

CyberVy:
It's very easy to measure RAM savings in Colab.

**Now**
```python
from transformers import CLIPVisionModelWithProjection
import torch
model = CLIPVisionModelWithProjection.from_pretrained("eramth/ip-adapter",subfolder="sdxl_models/image_encoder",torch_dtype=torch.float16)
# The CPU RAM usage is about 2.3G.
```


```python
model = model.to("cuda")
# The CPU RAM usage is still about 2.3G. And the VRAM usage is about 3.8G
```


```python
# The below code can return the released CPU RAM to the system, so that we can observe the CPU RAM usage easily.
import ctypes
ctypes.CDLL("libc.so.6").malloc_trim(0)
# The CPU RAM usage is about 2.1G.
```

**Before**
```python
from transformers import CLIPVisionModelWithProjection
import torch
model = CLIPVisionModelWithProjection.from_pretrained("eramth/ip-adapter",subfolder="sdxl_models/image_encoder")
# The CPU RAM usage is about 8.9G.
```

```python
model = model.to("cuda",dtype=torch.float16)
# The CPU RAM usage is still about 6.4G. And the VRAM usage is about 3.8G
```

```python
# The below code can return the released CPU RAM to the system, so that we can observe the CPU RAM usage easily.
import ctypes
ctypes.CDLL("libc.so.6").malloc_trim(0)
# The CPU RAM usage is about 2.2G.
```

asomoza:
thanks a lot, I don't have a low RAM system so this is hard to catch for me at least. We have some errors in the tests which I believe aren't related to this PR but I'll wait until we can resolve them (they seems to happen because of transformers)

asomoza (approved):

hlky (approved):
Thanks @CyberVy. Nice improvement for low RAM users.

Failing tests are unrelated.

PR #10927, opened 2025-02-28
[Quantization]  support pass MappingType for TorchAoConfig

a120092009:
# What does this PR do?

Passing `MappingType` for `TorchAoConfig` is invalid

```
  File "/workspace/volume/zty/diffusers/src/diffusers/configuration_utils.py", line 123, in register_to_config
    logger.debug(f"Updating config from {previous_dict} to {internal_dict}")
                                                           ^^^^^^^^^^^^^^^
  File "/workspace/volume/zty/diffusers/src/diffusers/quantizers/quantization_config.py", line 671, in __repr__
    return f"{self.__class__.__name__} {json.dumps(config_dict, indent=2, sort_keys=True)}\n"
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 202, in encode
    chunks = list(chunks)
             ^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 432, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.12/json/encoder.py", line 406, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.12/json/encoder.py", line 406, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.12/json/encoder.py", line 439, in _iterencode
    o = _default(o)
        ^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type MappingType is not JSON serializable
```

Add JSON encoder for supporting `MappingType`

Test result:

image

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/diffusers/blob/main/CONTRIBUTING.md)?
- [x] Did you read our [philosophy doc](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md) (important for complex PRs)?
[...]

PR #10961, opened 2025-03-04
Fix the missing parentheses when calling is_torchao_available in quantization_config.py.

CyberVy:
Missing parentheses when calling ```is_torchao_available()``` will result in an import error in systems without ```torchao```.

### DIFF
--- a/src/diffusers/quantizers/quantization_config.py
+++ b/src/diffusers/quantizers/quantization_config.py
@@ -47,7 +47,7 @@ class QuantizationMethod(str, Enum):
     TORCHAO = "torchao"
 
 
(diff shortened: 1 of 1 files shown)


a-r-r-o-w (approved):
Oh this is my bad! I should've tested with/without TorchAO in #10927 but only did with TorchAO and missed this.

cc @DN6 Are the lazy imports not lazy? 😅

asomoza (approved):

asomoza:
Thanks!

a120092009:
Thank you for fixing the code! I just noticed this issue. @CyberVy

Issue #10962, opened 2025-03-04
Cogview4 pipeline not accepting prompt embeds, due to shape issues .


Vargol:
### Describe the bug

I've trying to run CogView4 using separate pipelines to encode text and generate the image in order to save memory (Unified Memory so I can't use offloading) with the aim of doing multiple prompts

e.g.

```py
te_pipe = CogView4Pipeline.from_pretrained("THUDM/CogView4-6B",
                                           transformer=None,
                                           vae=None,
                                           torch_dtype=torch.bfloat16).to("mps")

with torch.no_grad():
    prompt_embeds, negative_prompt_embeds = te_pipe.encode_prompt(
        prompt,
        negative_prompt,
        num_images_per_prompt=num_images_per_prompt,
    )

del te_pipe

pipe = CogView4Pipeline.from_pretrained("THUDM/CogView4-6B", text_encoder=None, tokenizer=None, torch_dtype=torch.bfloat16).to("mps")

```

and I get a failure with the following error

```py
ValueError: `prompt_embeds` and `negative_prompt_embeds` must have the same shape when passed directly, but got: `prompt_embeds` torch.Size([1, 144, 4096]) != `negative_prompt_embeds` torch.Size([1, 48, 4096]).
```

I'm using the encode function used by the pipe and can't see why the embeds would the any different to those used internally, and if so I'm not sure why the size check is needed and assuming either the check is a bug or the size of the embeddings the encode_prompt function generates is a bug.

If I try to skip the negative embeds the code tries to generate negative prompt embeddings which fails the new pipe doesn't have an encoder.

### Reproduction

```py
from diffusers import CogView4Pipeline
import torch
import gc

te_pipe = CogView4Pipeline.from_pretrained("THUDM/CogView4-6B",
                                           transformer=None,
                                           vae=None,
                                           torch_dtype=torch.bfloat16).to("mps")


prompt = "A vibrant cherry red sports car sits proudly under the gleaming sun, its polished exterior smooth and flawless, casting a mirror-like reflection. The car features a low, aerodynamic body, angular headlights that gaze forward like predatory eyes, and a set of black, high-gloss racing rims that contrast starkly with the red. A subtle hint of chrome embellishes the grille and exhaust, while the tinted windows suggest a luxurious and private interior. The scene conveys a sense of speed and elegance, the car appearing as if it's about to burst into a sprint along a coastal road, with the ocean's azure waves crashing in the background."

[...]

PR #10966, opened 2025-03-04
update check_input for cogview4

yiyixuxu:
https://github.com/huggingface/diffusers/issues/10962

allow prompt_embeds and negative_prompt_embeds to have different sequence length


a-r-r-o-w (approved):
Awesome, thanks! 🔥

PR #10974, opened 2025-03-05
Fix Flux Controlnet Pipeline _callback_tensor_inputs Missing Some Elements

CyberVy:
```._callback_tensor_inputs``` of Flux controlnet related pipelines are missing some important elements.

This PR is to fix it.

I'll be very appreciated if you can have a check.
@asomoza


CyberVy:
I've also accidently noticed that Flux related pipeline classes are missing some base classes.

I'll fix it in another PR.

asomoza:
yeah, lets keep it uniform with the rest of the API, even if it's not used right now.

CyberVy:
> yeah, lets keep it uniform with the rest of the API, even if it's not used right now.

👌
And I'll open another PR to add ```mask``` and ```masked_image_latents``` for ```StableDiffusionControlNetInpaintPipeline```. I didn't noticed it in last fix. 😭

asomoza (approved):

asomoza:
thanks!

PR #10968, opened 2025-03-05
LTX 0.9.5

a-r-r-o-w:
See https://huggingface.slack.com/archives/C08275HSG8J/p1741091747532049?thread_ts=1738246363.413529&cid=C08275HSG8J

## testing scripts 

<details>
<summary>test1: image condition</summary>

```python
# test ltx image conditioning
import torch
from diffusers import LTXConditionPipeline
from diffusers.pipelines.ltx.pipeline_ltx_condition import LTXVideoCondition
from diffusers.utils import export_to_video, load_video, load_image


device = "cuda:2"
dtype = torch.bfloat16
repo = "YiYiXu/ltx-95"

# Initialize the pipeline
pipe = LTXConditionPipeline.from_pretrained(repo, torch_dtype=dtype)
pipe.to(device)


prompt = "The video depicts a long, straight highway stretching into the distance, flanked by metal guardrails. The road is divided into multiple lanes, with a few vehicles visible in the far distance. The surrounding landscape features dry, grassy fields on one side and rolling hills on the other. The sky is mostly clear with a few scattered clouds, suggesting a bright, sunny day."
negative_prompt='worst quality, inconsistent motion, blurry, jittery, distorted'
image = load_image(
    "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/cosmos/cosmos-video2world-input.jpg"
)

# use `conditions` input
condition = LTXVideoCondition(
    image=image,
)

generator = torch.Generator(device=device).manual_seed(0)
video = pipe(
    conditions=[condition],
    prompt=prompt,
    negative_prompt=negative_prompt,
    width=768,
    height=512,
    num_frames=161,
    num_inference_steps=40,
    generator=generator,
).frames[0]

# Export the video
export_to_video(video, "yiyi_test_1_output_1.mp4", fps=24)


#  pass `image` input directly
generator = torch.Generator(device=device).manual_seed(0)
video = pipe(
    image=image,
    prompt=prompt,
    negative_prompt=negative_prompt,
    width=768,
    height=512,
    num_frames=161,
    num_inference_steps=40,
    generator=generator,
).frames[0]

# Export the video
export_to_video(video, "yiyi_test_1_output_2.mp4", fps=24)
```
</details>

<details>
<summary>test2: video condition</summary>

```python
# ltx video conditioning
import torch
from diffusers.pipelines.ltx.pipeline_ltx_condition import LTXConditionPipeline, LTXVideoCondition
[...]

PR #10976, opened 2025-03-06
[WIP] test prepare_latents for ltx0.95

yiyixuxu:
for https://github.com/huggingface/diffusers/pull/10968

### DIFF
--- a/src/diffusers/models/autoencoders/autoencoder_kl_ltx.py
+++ b/src/diffusers/models/autoencoders/autoencoder_kl_ltx.py
@@ -1105,6 +1105,8 @@ class AutoencoderKLLTXVideo(ModelMixin, ConfigMixin, FromOriginalModelMixin):
         scaling_factor: float = 1.0,
         encoder_causal: bool = True,
         decoder_causal: bool = False,
+        spatial_compression_ratio: int = None,
+        temporal_compression_ratio: int = None,
     ) -> None:
         super().__init__()
 
@@ -1142,8 +1144,8 @@ class AutoencoderKLLTXVideo(ModelMixin, ConfigMixin, FromOriginalModelMixin):
         self.register_buffer("latents_mean", latents_mean, persistent=True)
         self.register_buffer("latents_std", latents_std, persistent=True)
 
-        self.spatial_compression_ratio = patch_size * 2 ** sum(spatio_temporal_scaling)
-        self.temporal_compression_ratio = patch_size_t * 2 ** sum(spatio_temporal_scaling)
+        self.spatial_compression_ratio = patch_size * 2 ** sum(spatio_temporal_scaling) if spatial_compression_ratio is None else spatial_compression_ratio
+        self.temporal_compression_ratio = patch_size_t * 2 ** sum(spatio_temporal_scaling) if temporal_compression_ratio is None else temporal_compression_ratio
 
         # When decoding a batch of video latents at a time, one can save memory by slicing across the batch dimension
         # to perform decoding of a single video latent at a time.
--- a/src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
+++ b/src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
@@ -437,8 +437,8 @@ class LTXConditionPipeline(DiffusionPipeline, FromSingleFileMixin, LTXVideoLoraL
                 )
 
     @staticmethod
-    # Copied from diffusers.pipelines.ltx.pipeline_ltx.LTXPipeline._pack_latents
-    def _pack_latents(latents: torch.Tensor, patch_size: int = 1, patch_size_t: int = 1) -> torch.Tensor:
+    # adapted from diffusers.pipelines.ltx.pipeline_ltx.LTXPipeline._pack_latents
+    def _pack_latents(latents: torch.Tensor, patch_size: int = 1, patch_size_t: int = 1, device: torch.device = None) -> torch.Tensor:
         # Unpacked latents of shape are [B, C, F, H, W] are patched into tokens of shape [B, C, F // p_t, p_t, H // p, p, W // p, p].
         # The patch dimensions are then permuted and collapsed into the channel dimension of shape:
         # [B, F // p_t * H // p * W // p, C * p_t * p * p] (an ndim=3 tensor).
@@ -447,6 +447,16 @@ class LTXConditionPipeline(DiffusionPipeline, FromSingleFileMixin, LTXVideoLoraL
         post_patch_num_frames = num_frames // patch_size_t
         post_patch_height = height // patch_size
         post_patch_width = width // patch_size
+
+        latent_sample_coords = torch.meshgrid(
+            torch.arange(0, num_frames, patch_size_t, device=device),
+            torch.arange(0, height, patch_size, device=device),
+            torch.arange(0, width, patch_size, device=device),
+        )
+        latent_sample_coords = torch.stack(latent_sample_coords, dim=0)
+        latent_coords = latent_sample_coords.unsqueeze(0).repeat(batch_size, 1, 1, 1, 1)
+        latent_coords = latent_coords.reshape(batch_size, -1, num_frames * height * width)
+
         latents = latents.reshape(
             batch_size,
             -1,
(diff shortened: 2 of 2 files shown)


src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
```diff
@@ -447,6 +447,16 @@ def _pack_latents(latents: torch.Tensor, patch_size: int = 1, patch_size_t: int
         post_patch_num_frames = num_frames // patch_size_t
         post_patch_height = height // patch_size
         post_patch_width = width // patch_size
+
+        latent_sample_coords = torch.meshgrid(
+            torch.arange(0, num_frames, patch_size_t, device=device),
+            torch.arange(0, height, patch_size, device=device),
+            torch.arange(0, width, patch_size, device=device),
+        )
```
hlky, line 455:
```suggestion
        latent_sample_coords = torch.meshgrid(
            torch.arange(0, num_frames, patch_size_t, device=device),
            torch.arange(0, height, patch_size, device=device),
            torch.arange(0, width, patch_size, device=device),
            indexing="ij",
        )
```

https://pytorch.org/docs/stable/generated/torch.meshgrid.html

Default `indexing` in PyTorch will change from `ij` (same as numpy) to `xy`, there's a warning if we don't specify `indexing="ij"`.

src/diffusers/models/autoencoders/autoencoder_kl_ltx.py
```diff
@@ -1105,6 +1105,8 @@ def __init__(
         scaling_factor: float = 1.0,
         encoder_causal: bool = True,
         decoder_causal: bool = False,
+        spatial_compression_ratio: int = None,
```
yiyixuxu, line 1108: @a-r-r-o-w 
need to add  these vae config for ltx0.95   
```python 
{
   "spatial_compression_ratio": 32,
    "temporal_compression_ratio": 8,
}
```

a-r-r-o-w: Thanks, I'll update it when doing the final conversion

src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
```diff
@@ -617,47 +652,39 @@ def prepare_latents(
...
+                        condition_latents, self.transformer_spatial_patch_size, self.transformer_temporal_patch_size, device
                     )
+
+                    rope_interpolation_scale = (
+                        rope_interpolation_scale * 
+                        torch.tensor([self.vae_temporal_compression_ratio, self.vae_spatial_compression_ratio, self.vae_spatial_compression_ratio], device=rope_interpolation_scale.device)[None, :, None]
+                    )
+                    rope_interpolation_scale[:, 0] = (rope_interpolation_scale[:, 0] + 1 - self.vae_temporal_compression_ratio).clamp(min=0)               
+                    rope_interpolation_scale[:, 0] += condition.frame_index
+
```
a-r-r-o-w, line 665: I don't think this is compatible with what we do in LTXRotaryPosEmbed layer... We prepare the meshgrid there and only pass the interpolation scales from the pipeline. It seems like here we are preparing the meshgrid beforehand, which will be incorrect. I think we would have to do one of the following:
- Make sure to only pass multiplicative interpolation scale without first multiplying with the latent_coords (the screenshot below shows how I handled it in the other PR)
- If we're passing latent_coords, we will have to handle it differently in the transformer for LTX v0.9.0/v0.9.1 vs v0.9.5

image

src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
```diff
@@ -864,7 +891,7 @@ def __call__(
             frame_rate,
             generator,
             device,
-            torch.float32,
+            prompt_embeds.dtype,
```
a-r-r-o-w, line 894: We could use float32 here and then typecast before sending into transformer, no? That way there won't be a downcast/upcast for CFG

hlky (approved):
Thanks @yiyixuxu

src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
```diff
@@ -495,178 +553,176 @@ def _denormalize_latents(
...
-            conditions = [conditions]
-
         num_latent_frames = (num_frames - 1) // self.vae_temporal_compression_ratio + 1
         latent_height = height // self.vae_spatial_compression_ratio
         latent_width = width // self.vae_spatial_compression_ratio
 
         shape = (batch_size, num_channels_latents, num_latent_frames, latent_height, latent_width)
         latents = randn_tensor(shape, generator=generator, device=device, dtype=dtype)
+        # YiYi TODO: testing only, remove this change before merging
+        # latents = torch.load("/raid/yiyi/LTX-Video/init_latents.pt").to(device, dtype=dtype)
```
a-r-r-o-w, line 623:
```suggestion
```

src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
```diff
@@ -495,178 +553,176 @@ def _denormalize_latents(
...
+                            latents[:, :, start_frame:end_frame],
+                            condition_latents[:, :, num_prefix_latent_frames:],
+                            strength,
+                        )
+                        condition_latent_frames_mask[:, start_frame:end_frame] = strength
+                        condition_latents = condition_latents[:, :, :num_prefix_latent_frames]
+
+                noise = randn_tensor(condition_latents.shape, generator=generator, device=device, dtype=dtype)
+                # YiYi TODO: testing only, remove this change before merging
+                # noise = torch.load("/raid/yiyi/LTX-Video/noise.pt").to(device, dtype=dtype)
```
a-r-r-o-w, line 666:
```suggestion
```

src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
```diff
@@ -495,178 +553,176 @@ def _denormalize_latents(
...
-                        condition_latents.shape[:2], condition.strength, device=device, dtype=dtype
-                    )
-
-                    rope_interpolation_scale = [
-                        # TODO!!! This is incorrect: the frame index needs to added AFTER multiplying the interpolation
-                        # scale with the grid.
-                        (self.vae_temporal_compression_ratio + condition.frame_index) / frame_rate,
-                        self.vae_spatial_compression_ratio,
-                        self.vae_spatial_compression_ratio,
-                    ]
```
a-r-r-o-w, line 633: @yiyixuxu Pardon my stupidity, but I can't seem to find if we're handling this `+ condition.frame_index` part. Is this missing by any chance, or was I mistaken in trying to handle this here?

In the original code, this is what I was meaning to handle: https://github.com/Lightricks/LTX-Video/blob/496dc5058f4408dcb777282f3fb6377fb2da08e6/ltx_video/pipelines/pipeline_ltx_video.py#L1285

a-r-r-o-w (approved):
Thanks so much for taking this up @yiyixuxu! LGTM but I had one doubt

src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
```diff
@@ -436,6 +455,45 @@ def check_inputs(
...
+        scale_factor_t: int = 8,
+        frame_index: int = 0,
+        device: torch.device = None,
+    ) -> torch.Tensor:
+        scaled_latent_coords = (
+            video_ids
+            * torch.tensor([scale_factor_t, scale_factor, scale_factor], device=video_ids.device)[None, :, None]
+        )
+        scaled_latent_coords[:, 0] = (scaled_latent_coords[:, 0] + 1 - scale_factor_t).clamp(min=0)
+        scaled_latent_coords[:, 0] += frame_index
```
yiyixuxu, line 493: @a-r-r-o-w it's here!

src/diffusers/pipelines/ltx/pipeline_ltx_condition.py
yiyixuxu: @a-r-r-o-w 

I think we have different ways to handle rope in our current code base, in general, I think it's more convenient/natural to prepare position ids (e.g. the image_ids, text_ids in flux or the grid here for ltx) at same time when we patchify the latents (e.g. pack_latent for ltx or flux). flux and ltx do this in pipeline and other models like lumina handle both together inside transformer with a patch embed

I think it is ok to have this flexibility for rope since it's something that slows us down for each integration. maybe a general rule is to try to follow closer to the original code base and fit it into one of the patterns that's easier for us to maintain.

PR #10980, opened 2025-03-06
[CI] remove synchornized.

sayakpaul:
# What does this PR do?

Otherwise, the PRs will only trigger tests when there are changes on the PRs not when the PRs are opened


hlky (approved):
Interesting, thanks for the fix.

DN6 (approved):

PR #10981, opened 2025-03-06
[LoRA] CogView4

a-r-r-o-w:


tests/lora/test_lora_layers_cogview4.py
```diff
@@ -0,0 +1,146 @@
...
+    @staticmethod
+    def from_pretrained(*args, **kwargs):
+        return AutoTokenizer.from_pretrained("THUDM/glm-4-9b-chat", trust_remote_code=True)
+
+
+class TextEncoderWrapper:
+    @staticmethod
+    def from_pretrained(*args, **kwargs):
+        config = GlmConfig(hidden_size=32, intermediate_size=8, num_hidden_layers=2, num_attention_heads=4, head_dim=8)
+        return GlmModel(config)
```
a-r-r-o-w, line 40: I had to do this because I'm not quite sure how to make the `trust_remote_code=True` part compatible. But it seems like a security hole...

@sayakpaul What are your suggestions for handling this? Is this okay or do we create the dummy checkpoints? Creating dummy checkpoint for text encoder is no problem, but I'm not sure how to handle the tokenizer here -- do I just copy the entire repo for tokenization to our hf-internal-testing org?

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
```diff
@@ -627,6 +635,7 @@ def __call__(
                         original_size=original_size,
                         target_size=target_size,
                         crop_coords=crops_coords_top_left,
+                        attention_kwargs=attention_kwargs,
```
sayakpaul, line 638: Let's add this to docstrings.

tests/lora/test_lora_layers_cogview4.py
sayakpaul: Thanks for investigating. How about:

1. Create a dummy checkpoint for the text encoder and host under internal testing.
2. Copy over the tokenizer files (they are small enough) to that same repo while crediting the source in the model card.

tests/lora/test_lora_layers_cogview4.py
```diff
@@ -0,0 +1,146 @@
...
+        super().test_simple_inference_with_text_lora_denoiser_fused_multi(expected_atol=9e-3)
+
+    def test_simple_inference_with_text_denoiser_lora_unfused(self):
+        super().test_simple_inference_with_text_denoiser_lora_unfused(expected_atol=9e-3)
+
+    @unittest.skip(
+        "Needs additional debugging. OSError: Incorrect path_or_model_id: ''. Please provide either the path to a local folder or the repo_id of a model on the Hub."
+    )
+    def test_simple_inference_save_pretrained(self):
+        pass
```
sayakpaul, line 114: This seems like an important test -- should we debug further? I can help look into it after the dummy checkpoints for the text encoder and tokenizer are created. I won't mind passing additional `pretrained_kwargs` to tokenizer and text encoder if needed.

sayakpaul (approved):
LGTM after the test related comments have been addressed. 

It's also great that `finetrainers` now supports image models like CogView4. I think that demands some communications to the communities. Let's tackle that later.

tests/lora/test_lora_layers_cogview4.py
a-r-r-o-w: Done here: https://huggingface.co/hf-internal-testing/tiny-random-cogview4

src/diffusers/pipelines/cogview4/pipeline_cogview4.py
a-r-r-o-w: It's already in the docstrings from before (it wasn't removed during integration)

sayakpaul (approved):
Nice learning in accommodating the tokenizer. Good work!

PR #10982, opened 2025-03-06
fix default values of Flux guidance_scale in docstrings

catwell:
Note: I just fixed the docstrings so they match the actual params.

I did not check what the actual values should be for each pipeline. Also there is discrepancy with the values used in the examples at the top of the files.


a-r-r-o-w (approved):
Thanks! I think the values you mentioned is perfect and not the ones at the top of the file example

PR #11004, opened 2025-03-07
[LoRA] remove full key prefix from peft.

sayakpaul:
# What does this PR do?

There's some uncertainty around https://github.com/huggingface/peft/pull/2382. So, we decided to remove the full-key prefix thingy from `diffusers` until https://github.com/huggingface/peft/pull/2382 is merged with a specific direction.

### DIFF
--- a/src/diffusers/loaders/peft.py
+++ b/src/diffusers/loaders/peft.py
@@ -192,11 +192,6 @@ class PeftAdapterMixin:
         from peft import LoraConfig, inject_adapter_in_model, set_peft_model_state_dict
         from peft.tuners.tuners_utils import BaseTunerLayer
 
-        try:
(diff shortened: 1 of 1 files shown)


hlky (approved):
Thanks! Should we try to other proposed fix " storing deleted keys to ensure they aren't re-added by other iterations"

sayakpaul:
> hould we try to other proposed fix " storing deleted keys to ensure they aren't re-added by other iterations"


Did you mean?

https://github.com/huggingface/diffusers/blob/363d1ab7e24c5ed6c190abb00df66d9edb74383b/src/diffusers/loaders/peft.py#L89-L91

hlky:
Oh :) all good then!

BenjaminBossan (approved):
Thanks for the quick adjustments and sorry again for the inconvenience.

sayakpaul:
No problem at all!

PR #11011, opened 2025-03-08
Fix missing **kwargs in lora_pipeline.py

CyberVy:
Methods ```super().fuse_lora``` and ```super().unfuse_lora``` in  ```lora_pipeline.py``` are all missing ```**kwargs```.

This PR is to fix this issue.

@asomoza


hlky:
@bot /style

hlky:
@bot /style

hlky:
@bot /style

CyberVy:
Thanks for style fix. 🥰

sayakpaul (commented):
Thanks for the PR. But do we need the `kwargs` here? IIRC, the `kwargs` was added to provide backwards compatibility when we refactored the LoRA loader classes.

CyberVy:
@sayakpaul Hi!
I think ```**kwargs``` should be passed to ```super().fuse_lora``` and ```super().unfuse_lora```, because there are some ```deprecate``` functions which require ```**kwargs``` in these two methods. Without **kwargs, users won't know if they have passed any deprecated arguments.

hlky (approved):
Thanks @CyberVy. Deprecations are set for 1.0 so useful to ensure the warnings are raised to the user.

sayakpaul (approved):
Thanks for catching this!

PR #11030, opened 2025-03-11
Use `output_size` in `repeat_interleave`

hlky:
# What does this PR do?

https://pytorch.org/docs/stable/generated/torch.repeat_interleave.html

> output_size ([int](https://docs.python.org/3/library/functions.html#int), optional) – Total output size for the given axis ( e.g. sum of repeats). If given, it will avoid stream synchronization needed to calculate output shape of the tensor.

## Who can review?

### DIFF
--- a/src/diffusers/models/attention_processor.py
+++ b/src/diffusers/models/attention_processor.py
@@ -741,10 +741,14 @@ class Attention(nn.Module):
 
         if out_dim == 3:
             if attention_mask.shape[0] < batch_size * head_size:
-                attention_mask = attention_mask.repeat_interleave(head_size, dim=0)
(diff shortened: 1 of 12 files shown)


yiyixuxu (approved):
thanks! awesome!

a-r-r-o-w (approved):
All occurrences look correctly updated to me! Also super cool to learn that specifying output_size removes a sync!

hlky:
> do we see a performance difference?

I've only tested Flux pipeline which uses `repeat_interleave` via `apply_rotary_emb`, no noticeable difference there, will test the others.

I plan on doing some further investigation into syncs and graph breaks (with torch.compile).

PR #11017, opened 2025-03-10
[hybrid inference 🍯🐝] Add VAE encode

hlky:
# What does this PR do?

- [x] Create endpoints, run slow tests (internal: https://github.com/huggingface/honey/pull/12 https://huggingface.co/diffusers/autoencoder_kl.encoder)
- [x] VAE encode docs page (https://huggingface.co/datasets/huggingface/documentation-images/discussions/449 https://huggingface.co/datasets/huggingface/documentation-images/discussions/448)
    - [x] Benchmarks
    - [x] Use cases (training, img2img)
- [x] Changelog on Overview page

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #11031, opened 2025-03-11
[hybrid inference 🍯🐝] Wan 2.1 decode

hlky:
# What does this PR do?

Rebase after #11017 

## Who can review?


tests/remote/test_remote_decode.py
```diff
@@ -467,6 +457,31 @@ class RemoteAutoencoderKLHunyuanVideoTests(
...
+        3,
+        40,
+        64,
+    )
+    out_hw = (
+        320,
+        512,
+    )
+    endpoint = DECODE_ENDPOINT_WAN_2_1
+    dtype = torch.float16
```
hlky, line 476: cc @yiyixuxu Currently the endpoint is running in float16 and output seems ok (on random latent), the examples use float32 but we noticed in the original code that everything is under bfloat16 autocast context. Can we check with the authors regarding the use of float32 for VAE?

tests/remote/test_remote_decode.py
```diff
@@ -176,18 +177,6 @@ def test_output_type_pt_partial_postprocess_return_type_pt(self):
             f"{output_slice}",
         )
 
-    def test_do_scaling_deprecation(self):
```
hlky, line 179: Wan VAE doesn't have a `scaling_factor` that we could pass, so this deprecation/removing `do_scaling` doesn't work, we will keep it.

tests/remote/test_remote_decode.py
yiyixuxu: i haven't heard back from them, 
maybe let's generate some sample comparison and ask @asomoza to help decide if there is a quality difference

PR #11037, opened 2025-03-12
[Tests] add requires peft decorator.

sayakpaul:


.github/workflows/nightly_tests.yml
```diff
@@ -437,6 +437,9 @@ jobs:
           python -m venv /opt/venv && export PATH="/opt/venv/bin:$PATH"
           python -m uv pip install -e [quality,test]
           python -m uv pip install -U ${{ matrix.config.backend }}
+          if [ "${{ matrix.config.backend }}" = "bitsandbytes" ]; then
```
DN6, line 440: Hmm since we're using a matrix here, we can add a field to the matrix e.g `additional_deps` 
```
          - backend: "bitsandbytes"
            test_location: "bnb"
            additional_deps: ["peft"]
          - backend: "gguf"
            test_location: "gguf"
            additional_deps: []
```
then install via

```
          if [ ! -z "${{ join(matrix.config.additional_deps, ' ') }}" ]; then
            python -m uv pip install ${{ join(matrix.config.additional_deps, ' ') }}
          fi
```
That way if other quant backend need additional test dependencies it's easy to add.

sayakpaul: Done: https://github.com/huggingface/diffusers/pull/11037/commits/6f85919dcf08fad0e13475a19e0c076ea932a146

DN6 (approved):

PR #11039, opened 2025-03-12
Don't override `torch_dtype` and don't use when `quantization_config` is set

hlky:
# What does this PR do?

https://github.com/huggingface/diffusers/actions/runs/13800894621/job/38602946014#step:7:1904

## Who can review?


CyberVy:
@hlky Hi!

This is a good PR for me. Sometimes I forget to set ```torch_dtype``` when loading a model, which can cause some issues, because the loader defaults to```torch.float32```. 👍

src/diffusers/models/modeling_utils.py
```diff
@@ -883,12 +883,15 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
         disable_mmap = kwargs.pop("disable_mmap", False)
 
-        if not isinstance(torch_dtype, torch.dtype):
+        if torch_dtype is not None and not isinstance(torch_dtype, torch.dtype):
             torch_dtype = torch.float32
             logger.warning(
                 f"Passed `torch_dtype` {torch_dtype} is not a `torch.dtype`. Defaulting to `torch.float32`."
             )
 
+        if quantization_config is not None and torch_dtype is not None:
```
DN6, line 892: In this case, if we were trying to set the dtype of the model e.g. `FluxTransformer(quantization_config=BnBConfig, torch_dtype=torch.bfloat16)` Wouldn't the dtype be overwritten and then get set to the BnB default `float16`?

hlky: Yes looks like it, I think it's ok without this part.

src/diffusers/models/modeling_utils.py
```diff
@@ -883,12 +883,15 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
+            torch_dtype = None
+
```
hlky, line 894:
```suggestion
```

src/diffusers/loaders/single_file_model.py
```diff
@@ -255,17 +255,20 @@ def from_single_file(cls, pretrained_model_link_or_path_or_dict: Optional[str] =
...
-        if not isinstance(torch_dtype, torch.dtype):
+        if torch_dtype is not None and not isinstance(torch_dtype, torch.dtype):
             torch_dtype = torch.float32
             logger.warning(
                 f"Passed `torch_dtype` {torch_dtype} is not a `torch.dtype`. Defaulting to `torch.float32`."
             )
 
+        if quantization_config is not None and torch_dtype is not None:
+            torch_dtype = None
+
```
hlky, line 271:
```suggestion
```

DN6 (approved):

PR #11054, opened 2025-03-13
Remove forced float64 from onnx stable diffusion pipelines

lostdisc:
These commits change "np.float64(self.scheduler.init_noise_sigma)" into just "self.scheduler.init_noise_sigma" for the onnx stable diffusion pipelines.  The forced float64 was affecting the latents' dtypes too, and causing an error with onnxruntime, which wanted float16.

I encountered the error while running test-inferences with [Olive's stable diffusion example](https://github.com/microsoft/Olive/tree/main/examples/stable_diffusion), and eventually tracked down the source of the float64 typing to these lines of code.  Simply removing the forced float64 resolved the error and let image-generation finish successfully.


yiyixuxu (approved):
thanks!

PR #11058, opened 2025-03-14
[tests] make cuda only tests device-agnostic 

faaany:
## What does this PR do?
This PR is ready for review. 

cc @hlky


hlky (approved):
Thanks @faaany.

May we ask which Intel GPUs you are using internally? We have been looking at getting B570/B580 to benchmark performance and would be interested in anything you can share in that regard. cc @sayakpaul

faaany:
> Thanks @faaany.
> 
> May we ask which Intel GPUs you are using internally? We have been looking at getting B570/B580 to benchmark performance and would be interested in anything you can share in that regard. cc @sayakpaul

We are using Intel(R) Data Center GPU Max 1550.

PR #11059, opened 2025-03-14
[LoRA] feat: support non-diffusers wan t2v loras.

sayakpaul:
# What does this PR do?

https://github.com/huggingface/diffusers//issues/11025 added support for the 14B I2V non-diffusers LoRAs. This adds support for 1.3B T2V non-diffusers LoRAs.


yiyixuxu (approved):
thanks!

PR #11066, opened 2025-03-15
New HunyuanVideo-I2V

a-r-r-o-w:
WIP

### DIFF
--- a/src/diffusers/models/transformers/transformer_hunyuan_video.py
+++ b/src/diffusers/models/transformers/transformer_hunyuan_video.py
@@ -29,11 +29,14 @@ from ..cache_utils import CacheMixin
 from ..embeddings import (
     CombinedTimestepGuidanceTextProjEmbeddings,
     CombinedTimestepTextProjEmbeddings,
+    PixArtAlphaTextProjection,
+    TimestepEmbedding,
+    Timesteps,
     get_1d_rotary_pos_embed,
 )
 from ..modeling_outputs import Transformer2DModelOutput
 from ..modeling_utils import ModelMixin
-from ..normalization import AdaLayerNormContinuous, AdaLayerNormZero, AdaLayerNormZeroSingle
+from ..normalization import AdaLayerNormContinuous, AdaLayerNormZero, AdaLayerNormZeroSingle, FP32LayerNorm
 
 
 logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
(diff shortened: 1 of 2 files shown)


src/diffusers/models/transformers/transformer_hunyuan_video.py
```diff
@@ -503,6 +585,101 @@ def forward(
         return hidden_states, encoder_hidden_states
 
 
+class HunyuanVideoTokenReplaceTransformerBlock(nn.Module):
```
a-r-r-o-w, line 588: @yiyixuxu Decided that it was easier to maintain separate transformer block implementations instead of packing the different logics into the forward (which would introduce a lot of if-else). We never know how many more variants the team will decide to release, and the logic could get really complicate really fast (for example with LTX Video). LMK your thoughts on this (although the full PR is not ready for review yet)

src/diffusers/models/transformers/transformer_hunyuan_video.py
```diff
@@ -390,6 +527,8 @@ def forward(
         temb: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
         image_rotary_emb: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
+        *args,
```
yiyixuxu, line 530: what's these used for?

src/diffusers/models/transformers/transformer_hunyuan_video.py
```diff
@@ -468,6 +607,8 @@ def forward(
         temb: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
         freqs_cis: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
+        *args,
```
yiyixuxu, line 610: same here

src/diffusers/models/transformers/transformer_hunyuan_video.py
```diff
@@ -570,8 +890,10 @@ def __init__(
         pooled_projection_dim: int = 768,
         rope_theta: float = 256.0,
         rope_axes_dim: Tuple[int] = (16, 56, 56),
+        image_condition_type: Optional[str] = None,
     ) -> None:
         super().__init__()
+        assert image_condition_type is None or image_condition_type in ["latent_concat", "token_replace"]
```
yiyixuxu, line 896: can we raise a ValueError instead here?

src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
```diff
@@ -497,10 +506,11 @@ def prepare_latents(
         image = image.unsqueeze(2)  # [B, C, 1, H, W]
         if isinstance(generator, list):
             image_latents = [
-                retrieve_latents(self.vae.encode(image[i].unsqueeze(0)), generator[i]) for i in range(batch_size)
+                retrieve_latents(self.vae.encode(image[i].unsqueeze(0)), generator[i], "argmax")
+                for i in range(batch_size)
```
yiyixuxu, line 510: is this not backward breaking?

src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
```diff
@@ -598,6 +611,7 @@ def __call__(
         callback_on_step_end_tensor_inputs: List[str] = ["latents"],
         prompt_template: Dict[str, Any] = DEFAULT_PROMPT_TEMPLATE,
         max_sequence_length: int = 256,
+        image_embed_interleave: Optional[int] = None,
```
yiyixuxu, line 614: what is this argument?

src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
```diff
@@ -706,10 +720,18 @@ def __call__(
             prompt_template,
         )
 
+        image_condition_type = self.transformer.config.image_condition_type
```
yiyixuxu, line 723: can we set a default value in case transformer is not loaded?

src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
```diff
@@ -782,10 +812,17 @@ def __call__(
             negative_prompt_attention_mask = negative_prompt_attention_mask.to(transformer_dtype)
             negative_pooled_prompt_embeds = negative_pooled_prompt_embeds.to(transformer_dtype)
 
-        # 4. Prepare timesteps
+        # 5. Prepare timesteps
         sigmas = np.linspace(1.0, 0.0, num_inference_steps + 1)[:-1] if sigmas is None else sigmas
         timesteps, num_inference_steps = retrieve_timesteps(self.scheduler, num_inference_steps, device, sigmas=sigmas)
 
+        # 6. Prepare guidance condition
+        guidance = None
```
yiyixuxu, line 820: oh should we add a check now for true_cfg & guidance? I think we cannot use cfg with the guidance embeds, no?

src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
a-r-r-o-w: Nope, the `latent_concat` I2V model also uses the mode of the distribution instead of sample. I missed it previously, but now looking at their code again, they do `latent_dist.mode()`

src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
a-r-r-o-w: This is the __call__ function, so transformer must be loaded

src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
a-r-r-o-w: I believe it controls the amount of image embedding sequence tokens seen by the transformer. Previously, we had this defaulted to 2 (in encode_prompt). For the new I2V model, they make this user-configurable and it's different for 33-ch vs 16-ch model

src/diffusers/models/transformers/transformer_hunyuan_video.py
a-r-r-o-w: In the main transformer, we pass `token_replace_emb` and `first_frame_num_tokens`. The original single and double transformer blocks don't use those arguments, so we need to discard them. The args/kwargs is not used, so they are essentially discarded.

The extra arguments are only used by the new token-replace single and double blocks.

We will have to do a lot of if-else in the main transformer model forward otherwise:

image

src/diffusers/models/transformers/transformer_hunyuan_video.py
a-r-r-o-w: We raise ValueError now

src/diffusers/models/transformers/transformer_hunyuan_video.py
a-r-r-o-w: Do you prefer something like this?

```
block_args = [hidden_states, encoder_hidden_states, temb, attention_mask, image_rotary_emb]

if self.config.image_condition_type == "token_replace":
    block_args.extend([token_replace_emb, first_frame_num_tokens])

for block in self.transformer_blocks:
    block_args[0], block_args[1] = block(*block_args)
```

a-r-r-o-w:
Failing tests seem unrelated

src/diffusers/models/transformers/transformer_hunyuan_video.py
yiyixuxu: ok, sound good! (no need to change)

yiyixuxu (approved):
thanks!

src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
```diff
@@ -704,12 +727,22 @@ def __call__(
             prompt_embeds,
             callback_on_step_end_tensor_inputs,
             prompt_template,
+            true_cfg_scale,
+            guidance_scale,
         )
 
+        image_condition_type = self.transformer.config.image_condition_type
```
yiyixuxu, line 734: testing failure related to here when `transformer.config.image_condition_type` is None

a-r-r-o-w: Oh I see. `image_condition_type` must take a value of "latent_concat" or "token_replace" but I didn't update for ImageToVideoPipeline test. Done now, so we can merge once CI is green

src/diffusers/models/transformers/transformer_hunyuan_video.py
```diff
@@ -570,9 +890,16 @@ def __init__(
         pooled_projection_dim: int = 768,
         rope_theta: float = 256.0,
         rope_axes_dim: Tuple[int] = (16, 56, 56),
+        image_condition_type: Optional[str] = None,
```
yiyixuxu, line 893: but here it defaults to be `None`

a-r-r-o-w: Yes, that should be None by default since None is required for Text-to-Video model. The issue was with the image-to-video pipeline test not setting the correct image_condition_type, no? The failing test related to UnboundLocalVariable are not happening now

a-r-r-o-w:
There needs to be multiple memory saving optimizations enabled for it to run on 24gb for this one :(

cc @asomoza as he'll be able to help you in this regard with examples. The latest diffusers main provides group offloading (+an option to lower cpu usage) and layerwise casting, which may be helpful for you to run the model without too much additional time overhead: https://huggingface.co/docs/diffusers/main/en/optimization/memory

eppaneamd:
https://github.com/huggingface/diffusers/issues/12273

PR #11076, opened 2025-03-17
change name from SanaSCMPipeline to SanaSprintPipeline.

lawrence-cj:
This PR change pipeline name from SanaSCMPipeline to SanaSprintPipeline.

Cc: @yiyixuxu @sayakpaul


yiyixuxu (approved):
thanks!

PR #11079, opened 2025-03-17
[fix bug] PixArt inference_steps=1

lawrence-cj:
This PR fix the bug when PixArt inference with `num_inference_steps=1`

### DIFF
--- a/src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
+++ b/src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
@@ -941,8 +941,12 @@ class PixArtAlphaPipeline(DiffusionPipeline):
 
                 # compute previous image: x_t -> x_t-1
                 if num_inference_steps == 1:
-                    # For DMD one step sampling: https://arxiv.org/abs/2311.18828
-                    latents = self.scheduler.step(noise_pred, t, latents, **extra_step_kwargs).pred_original_sample
+                    try:
+                        # For LCM one step sampling
+                        latents = self.scheduler.step(noise_pred, t, latents, **extra_step_kwargs).denoised
+                    except:
+                        # For DMD one step sampling: https://arxiv.org/abs/2311.18828
+                        latents = self.scheduler.step(noise_pred, t, latents, **extra_step_kwargs).pred_original_sample
                 else:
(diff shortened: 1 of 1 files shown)


hlky:
@bot /style

hlky:
@bot /style

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
```diff
@@ -941,8 +941,12 @@ def __call__(
 
                 # compute previous image: x_t -> x_t-1
                 if num_inference_steps == 1:
-                    # For DMD one step sampling: https://arxiv.org/abs/2311.18828
-                    latents = self.scheduler.step(noise_pred, t, latents, **extra_step_kwargs).pred_original_sample
+                    try:
+                        # For LCM one step sampling
+                        latents = self.scheduler.step(noise_pred, t, latents, **extra_step_kwargs).denoised
+                    except:
```
hlky, line 947:
```suggestion
                    except Exception:
```
At minimum due to [`ruff`](https://github.com/huggingface/diffusers/actions/runs/13895274934/job/38874447777). What's the exception type raised here? Is there anything else we can check instead of relying on try/except?

lawrence-cj: This one is due to the Output difference between
https://github.com/huggingface/diffusers/blob/100142586f82a9410f5bc393b9eb06c12d771006/src/diffusers/schedulers/scheduling_lcm.py#L48-L49
and 
https://github.com/huggingface/diffusers/blob/100142586f82a9410f5bc393b9eb06c12d771006/src/diffusers/schedulers/scheduling_ddim.py#L46-L47

What do you think is better to write here?

lawrence-cj: gentle ping @hlky

hlky: With `return_dict=False` a `tuple` is returned, both `denoised` and `pred_original_sample` are at index 1 so we can do
```python
latents = self.scheduler.step(noise_pred, t, latents, **extra_step_kwargs, return_dict=False)[1]
```

lawrence-cj: That makes sense. updated.

PR #11080, opened 2025-03-17
[BUG] fixes in kadinsky pipeline

### LINKED ISSUE #11060: `prepare_image` in Kandinsky pipelines doesn't support `torch.Tensor`
Hi, I want to report a bug in Kandinsky pipelines.

https://github.com/huggingface/diffusers/blob/2f0f281b0d808c05bc7a974e68d298a006dd120a/src/diffusers/pipelines/kandinsky/pipeline_kandinsky_img2img.py#L413-L420

According to the above contents, elements in `image` can be either `PIL.Image.Image` or `torch.Tensor`.

https://github.com/huggingface/diffusers/blob/2f0f281b0d808c05bc7a974e68d298a006dd120a/src/diffusers/pipelines/kandinsky/pipeline_kandinsky_img2img.py#L98-L104

However, the `prepare_image` function is only for `PIL.Image.Image`, and does not support `torch.Tensor`.

Can you resolve this problem by implementing an image resize function for `torch.Tensor`?

ishan-modi:
# What does this PR do?

Fixes #11060

## Who can review?

@DN6

### DIFF
--- a/src/diffusers/pipelines/kandinsky2_2/pipeline_kandinsky2_2_controlnet_img2img.py
+++ b/src/diffusers/pipelines/kandinsky2_2/pipeline_kandinsky2_2_controlnet_img2img.py
@@ -27,7 +27,7 @@ from ...utils import (
 )
 from ...utils.torch_utils import randn_tensor
 from ..pipeline_utils import DiffusionPipeline, ImagePipelineOutput
-
+from ...image_processor import VaeImageProcessor
 
 if is_torch_xla_available():
     import torch_xla.core.xla_model as xm
@@ -105,27 +105,6 @@ EXAMPLE_DOC_STRING = """
 """
 
 
-# Copied from diffusers.pipelines.kandinsky2_2.pipeline_kandinsky2_2.downscale_height_and_width
-def downscale_height_and_width(height, width, scale_factor=8):
-    new_height = height // scale_factor**2
-    if height % scale_factor**2 != 0:
-        new_height += 1
-    new_width = width // scale_factor**2
-    if width % scale_factor**2 != 0:
-        new_width += 1
-    return new_height * scale_factor, new_width * scale_factor
-
-
-# Copied from diffusers.pipelines.kandinsky.pipeline_kandinsky_img2img.prepare_image
-def prepare_image(pil_image, w=512, h=512):
-    pil_image = pil_image.resize((w, h), resample=Image.BICUBIC, reducing_gap=1)
-    arr = np.array(pil_image.convert("RGB"))
-    arr = arr.astype(np.float32) / 127.5 - 1
-    arr = np.transpose(arr, [2, 0, 1])
-    image = torch.from_numpy(arr).unsqueeze(0)
-    return image
-
-
 class KandinskyV22ControlnetImg2ImgPipeline(DiffusionPipeline):
     """
     Pipeline for image-to-image generation using Kandinsky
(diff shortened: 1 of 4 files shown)


yiyixuxu:
thanks @ishan-modi !
are you able to run the docstring examples for these pipelines and see if the outputs are same in this branch vs on main?

ishan-modi:
I tried running both branches for kadinsky3 and following are the results

they are slightly different potentially due to usage of quantization/balanced_strategy(device_map) because of limited GPU

diff 
main
fixes-issue-11060

yiyixuxu (approved):
thanks!

yiyixuxu:
I think there are some kandinsky test failure that's potentially related https://github.com/huggingface/diffusers/actions/runs/14330695758/job/40203664079?pr=11080#step:6:33932

ishan-modi:
Thanks for the review, fixed the tests !

src/diffusers/pipelines/kandinsky2_2/pipeline_kandinsky2_2_controlnet_img2img.py
```diff
@@ -135,13 +135,11 @@ def __init__(
...
             movq=movq,
         )
-        movq_scale_factor = 2 ** (len(self.movq.config.block_out_channels) - 1)
-        self.image_processor = VaeImageProcessor(
-            vae_scale_factor=movq_scale_factor,
-            vae_latent_channels=self.movq.config.latent_channels,
-            resample="bicubic",
-            reducing_gap=1,
-        )
+        kwargs = {}
```
yiyixuxu, line 138: ohh we can just give them a default value like this

```python
movq_scale_factor = 2 ** (len(self.movq.config.block_out_channels) - 1) if getattr(self, "movq", None) else ..
movq_latent_channels = self.movq.config.latent_channels if getattr(self, "movq", None) else ..
```

to be consistent with how it is handled in other pipelines, for example https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py#L218

src/diffusers/pipelines/kandinsky3/pipeline_kandinsky3_img2img.py
```diff
@@ -630,22 +620,11 @@ def __call__(
...
-                if output_type in ["np", "pil"]:
-                    image = image * 0.5 + 0.5
-                    image = image.clamp(0, 1)
-                    image = image.cpu().permute(0, 2, 3, 1).float().numpy()
-
-                if output_type == "pil":
-                    image = self.numpy_to_pil(image)
             else:
                 image = latents
+            image = self.image_processor.postprocess(image, output_type)
```
hlky, line 627: This breaks `output_type == "latent"`. `image_processor.postprocess` should not be applied to latent output.

hlky (changes requested):
Same comment for all pipelines.

src/diffusers/pipelines/kandinsky3/pipeline_kandinsky3_img2img.py
ishan-modi: hmm, `postprocess` function doesn't do anything if the output_type is "latent", [see here](https://github.com/ishan-modi/diffusers/blob/fixes-issue-11060/src/diffusers/image_processor.py#L777-L778).

Let me know if I am missing anything

hlky: Even so, the code style does not match other pipelines, please make the requested changes to keep consistency with code styling.

ishan-modi: Alright, made the change let me know if it looks good

src/diffusers/pipelines/kandinsky2_2/pipeline_kandinsky2_2_img2img.py
```diff
@@ -367,11 +367,16 @@ def __call__(
             if XLA_AVAILABLE:
                 xm.mark_step()
 
+        if output_type not in ["pt", "np", "pil", "latent"]:
+            raise ValueError(
+                f"Only the output types `pt`, `pil`, `np` and `latent` are supported not output_type={output_type}"
+            )
```
hlky, line 373: This can be removed. Refer to other pipelines for an example.

https://github.com/huggingface/diffusers/blob/31c4f24fc18b7b451bd186e2132fa8bb3610374a/src/diffusers/pipelines/flux/pipeline_flux.py#L974-L981

hlky (changes requested):

ishan-modi:
@hlky gentle ping

yiyixuxu (approved):

yiyixuxu:
thanks a lot @ishan-modi

PR #11082, opened 2025-03-17
update conversion script for SANA-1.5 and SANA-Sprint

lawrence-cj:
This PR update the diffusers checkpoint conversion scripts.


lawrence-cj:
Adding SANA1.5 config into conversion script which also need qknorm. @yiyixuxu

scripts/convert_sana_to_diffusers.py
```diff
@@ -232,11 +258,9 @@ def main(args):
             "norm_elementwise_affine": False,
             "norm_eps": 1e-6,
             "interpolation_scale": interpolation_scale[args.image_size],
-        }
```
asomoza, line 235: deleted by mistake?

lawrence-cj: OMG, Thank you for noticing.

PR #11087, opened 2025-03-17
fix_wan_i2v_quality

chengzeyi:


src/diffusers/pipelines/wan/pipeline_wan_i2v.py
```diff
@@ -113,9 +113,11 @@ def retrieve_latents(
     latents_mean: torch.Tensor,
     latents_std: torch.Tensor,
     generator: Optional[torch.Generator] = None,
-    sample_mode: str = "sample",
+    sample_mode: str = "none",
```
yiyixuxu, line 116: I think this is actually should be same as `sample_mode == "argmax"` , no?
https://github.com/huggingface/diffusers/blob/100142586f82a9410f5bc393b9eb06c12d771006/src/diffusers/models/autoencoders/vae.py#L743

chengzeyi: Yeah, I think it is just without any extra computation cost.

src/diffusers/pipelines/wan/pipeline_wan_i2v.py
```diff
@@ -113,9 +113,11 @@ def retrieve_latents(
...
 ):
-    if hasattr(encoder_output, "latent_dist") and sample_mode == "sample":
+    if hasattr(encoder_output, "latent_dist") and sample_mode == "none":
+        return (encoder_output.latent_dist.mean - latents_mean) * latents_std
+    elif hasattr(encoder_output, "latent_dist") and sample_mode == "sample":
```
yiyixuxu, line 120:
```suggestion
    else:
           raise AttributeError("Could not access latents of provided encoder_output")
```
we can remove rest of code here

yiyixuxu:
cc @hlky here, I think we can refactor &simplify this code a bit here just for wan, this is not a `#Copied from` method anymore, it does not have to have same pattern as the original one

hlky (commented):
I think now `retrieve_latents` can be `Copied from`.

```python
# Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion_img2img.retrieve_latents
def retrieve_latents(
    encoder_output: torch.Tensor, generator: Optional[torch.Generator] = None, sample_mode: str = "sample"
):
    if hasattr(encoder_output, "latent_dist") and sample_mode == "sample":
        return encoder_output.latent_dist.sample(generator)
    elif hasattr(encoder_output, "latent_dist") and sample_mode == "argmax":
        return encoder_output.latent_dist.mode()
    elif hasattr(encoder_output, "latents"):
        return encoder_output.latents
    else:
        raise AttributeError("Could not access latents of provided encoder_output")
```

We pass `sample_mode="argmax"` then do the scaling

```python
            latent_condition = [
                retrieve_latents(self.vae.encode(video_condition), latents_mean, latents_std, g, sample_mode="argmax") for g in generator
            ]
            latent_condition = torch.cat(latent_condition)
        else:
            latent_condition = retrieve_latents(self.vae.encode(video_condition), latents_mean, latents_std, generator, sample_mode="argmax")
            latent_condition = latent_condition.repeat(batch_size, 1, 1, 1, 1)

        latent_condition = (latent_condition - latents_mean) * latents_std
```

hlky:
Output is good with https://github.com/huggingface/diffusers/pull/11087#pullrequestreview-2690321683

https://github.com/user-attachments/assets/0dd370db-afef-4a2a-9a8f-89e6e9f06421

hlky:
@chengzeyi Can we apply these changes https://github.com/huggingface/diffusers/pull/11087#pullrequestreview-2690321683 instead

chengzeyi:
> @chengzeyi Can we apply these changes [#11087 (review)](https://github.com/huggingface/diffusers/pull/11087#pullrequestreview-2690321683) instead

From my part I think these changes are OK.👌
And obviously no need to pass the generator in such cases.

hlky:
The changes are ok but the other changes allow us to keep `copied from` for `retrieve_latents` and only add 1 line `latent_condition = (latent_condition - latents_mean) * latents_std` to `prepare_latents` and modify 2 others to add `sample_mode="argmax"`, generator can also be removed from the function call as you mentioned.

hlky (approved):

Issue #11083, opened 2025-03-17
Running diffusers with torch-DDP leads to warning on stride mismatch


jinc7461:
### Describe the bug

This issue has already been reported at https://github.com/huggingface/diffusers/issues/3809 Through an analysis of the mismatched network parameters, it was found that modifying line 369 in `diffusers/models/resnet.py` from:  

```python
input_tensor = self.conv_shortcut(input_tensor)
```
to  
```python
input_tensor = self.conv_shortcut(input_tensor.contiguous())
```
resolves the problem.

### Reproduction

This issue has already been reported at https://github.com/huggingface/diffusers/issues/3809.

### Logs

```shell

```

### System Info

Diffusers: diffusers==0.31.0
Torch: torch==2.6.0
Python: Python 3.10.15

### Who can help?

@DN6 @yiyixuxu @sayakpaul

hlky:
Hi @jinc7461. Thanks for raising this and providing the fix, would you like to open a PR?

PR #11098, opened 2025-03-18
Resolve stride mismatch in UNet's ResNet to support Torch DDP

jinc7461:
Fixes https://github.com/huggingface/diffusers/issues/11083

explanation:
When using Torch's distributed training, the Diffusers code will display a "stride mismatch" warning. Simply adding `.contiguous()` in the code can resolve this issue.


hlky (approved):
Thanks @jinc7461

PR #11100, opened 2025-03-18
[👋] deprecate blip.

sayakpaul:
# What does this PR do?

As discussed internally.


src/diffusers/pipelines/__init__.py
```diff
@@ -105,6 +105,7 @@
             "VQDiffusionPipeline",
             "AltDiffusionPipeline",
             "AltDiffusionImg2ImgPipeline",
+            "BlipDiffusionPipeline" "BlipDiffusionControlNetPipeline",
```
hlky, line 108:
```suggestion
            "BlipDiffusionPipeline",
            "BlipDiffusionControlNetPipeline",
```

hlky (approved):
Thanks @sayakpaul

sayakpaul:
Will keep the PR for a bit as we will need to figure out a robust way to deprecate pipelines that use custom modules like Blip Diffusion does.

Right now, we cannot do:

```py
BlipDiffusionPipeline.from_pretrrained(...)
```

on the pretrained checkpoint.

Cc: @DN6

sayakpaul:
Closing in view of https://github.com/huggingface/diffusers/pull/11354.

PR #11105, opened 2025-03-18
Add gradient checkpointing support for AutoencoderKLWan

### LINKED ISSUE #11071: AutoencoderKLWan - support grandient_checkpointing
Do you have a plan for supporting gradient checkpointing for AutoencoderKLWan?

Thank you for always working hard for open source :pray::pray:

victolee0:
# What does this PR do?

Fixes #11071  (issue)

## Before submitting

## Who can review?

@a-r-r-o-w

### DIFF
--- a/src/diffusers/models/autoencoders/autoencoder_kl_wan.py
+++ b/src/diffusers/models/autoencoders/autoencoder_kl_wan.py
@@ -351,14 +351,24 @@ class WanMidBlock(nn.Module):
 
     def forward(self, x, feat_cache=None, feat_idx=[0]):
         # First residual block
-        x = self.resnets[0](x, feat_cache, feat_idx)
+        if torch.is_grad_enabled() and self.gradient_checkpointing:
+            x = self._gradient_checkpointing_func(self.resnets[0], x, feat_cache, feat_idx)
 
-        # Process through attention and residual blocks
-        for attn, resnet in zip(self.attentions, self.resnets[1:]):
-            if attn is not None:
-                x = attn(x)
+            for attn, resnet in zip(self.attentions, self.resnets[1:]):
+                if attn is not None:
+                    x = attn(x)
 
-            x = resnet(x, feat_cache, feat_idx)
+                x = self._gradient_checkpointing_func(resnet, x, feat_cache, feat_idx)
+
+        else:
+            x = self.resnets[0](x, feat_cache, feat_idx)
+
+            # Process through attention and residual blocks
+            for attn, resnet in zip(self.attentions, self.resnets[1:]):
+                if attn is not None:
+                    x = attn(x)
+
+                x = resnet(x, feat_cache, feat_idx)
 
         return x
 
@@ -443,15 +453,26 @@ class WanEncoder3d(nn.Module):
         else:
             x = self.conv_in(x)
 
-        ## downsamples
-        for layer in self.down_blocks:
-            if feat_cache is not None:
-                x = layer(x, feat_cache, feat_idx)
-            else:
-                x = layer(x)
+        if torch.is_grad_enabled() and self.gradient_checkpointing:
+            ## downsamples
+            for layer in self.down_blocks:
+                if feat_cache is not None:
+                    x = self._gradient_checkpointing_func(layer, x, feat_cache, feat_idx)
+                else:
+                    x = self._gradient_checkpointing_func(layer, x)
 
-        ## middle
-        x = self.mid_block(x, feat_cache, feat_idx)
+            ## middle
+            x = self._gradient_checkpointing_func(self.mid_block, x, feat_cache, feat_idx)
+        else:
+            ## downsamples
+            for layer in self.down_blocks:
+                if feat_cache is not None:
+                    x = layer(x, feat_cache, feat_idx)
+                else:
+                    x = layer(x)
+
+            ## middle
+            x = self.mid_block(x, feat_cache, feat_idx)
 
         ## head
         x = self.norm_out(x)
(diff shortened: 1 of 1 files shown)


a-r-r-o-w (approved):
Thanks, very nice!

a-r-r-o-w:
@bot /style

victolee0:
> Thanks, very nice!

My PR failed to pass the test code. Would you like to look at this [link](https://github.com/huggingface/diffusers/issues/11071#issuecomment-2733375368)?

quickdahuk:
@victolee0 I implemented it the same way you did. It works fine for the forward pass, but during the backward pass, the cache index gets mixed up and goes out of bounds. I might need to use a dictionary for the cache mechanism instead of an index-based list.

victolee0:
> @victolee0 I implemented it the same way you did. It works fine for the forward pass, but during the backward pass, the cache index gets mixed up and goes out of bounds. I might need to use a dictionary for the cache mechanism instead of an index-based list.

When I add exception handling as shown below, I get a different error.

Code
```
def forward(self, x, feat_cache=None, feat_idx=[0]):
        # Apply shortcut connection
        h = self.conv_shortcut(x)

        # First normalization and activation
        x = self.norm1(x)
        x = self.nonlinearity(x)

        # exception handling
        if feat_cache is not None and len(feat_cache) < feat_idx[0]:
            idx = feat_idx[0]
            cache_x = x[:, :, -CACHE_T:, :, :].clone()
            if cache_x.shape[2] < 2 and feat_cache[idx] is not None:
                cache_x = torch.cat([feat_cache[idx][:, :, -1, :, :].unsqueeze(2).to(cache_x.device), cache_x], dim=2)

            x = self.conv1(x, feat_cache[idx])
            feat_cache[idx] = cache_x
            feat_idx[0] += 1
        else:
            x = self.conv1(x)

        # Second normalization and activation
        x = self.norm2(x)
        x = self.nonlinearity(x)

        # Dropout
        x = self.dropout(x)

        # exception handling
        if feat_cache is not None and len(feat_cache) < feat_idx[0]:
            idx = feat_idx[0]
            cache_x = x[:, :, -CACHE_T:, :, :].clone()
            if cache_x.shape[2] < 2 and feat_cache[idx] is not None:
                cache_x = torch.cat([feat_cache[idx][:, :, -1, :, :].unsqueeze(2).to(cache_x.device), cache_x], dim=2)

            x = self.conv2(x, feat_cache[idx])
            feat_cache[idx] = cache_x
            feat_idx[0] += 1
        else:
            x = self.conv2(x)

        # Add residual connection
        return x + h
```

test case error
```
E           torch.utils.checkpoint.CheckpointError: torch.utils.checkpoint: A different number of tensors was saved during the original forward and recomputation.
E           Number of tensors saved during forward: 8
E           Number of tensors saved during recomputation: 6
```

quickdahuk:
@victolee0 I've implemented gradient checkpointing for the decoder. I don't need it for the encoder now. The training is working fine for me. I implemented checkpointing for each frame but didn't put it inside decoder operations.
[I've the code here](https://gist.github.com/quickdahuk/cdada5050dc0a01892d338f353171ae6).

victolee0:
@quickdahuk 
After copying and pasting the code you provided, I still get errors when running the test code on the same test.
I'm including the error message here:
```
a = tensor([-4.6806e-05, -2.9469e-05,  1.2341e-04]), b = tensor([-3.7164e-06, -4.1561e-06,  9.7381e-06]), args = (), kwargs = {'atol': 5e-05}

    def torch_all_close(a, b, *args, **kwargs):
        if not is_torch_available():
            raise ValueError("PyTorch needs to be installed to use this function.")
        if not torch.allclose(a, b, *args, **kwargs):
>           assert False, f"Max diff is absolute {(a - b).abs().max()}. Diff tensor is {(a - b).abs()}."
E           AssertionError: Max diff is absolute 0.00011366714170435444. Diff tensor is tensor([4.3089e-05, 2.5313e-05, 1.1367e-04]).

src/diffusers/utils/testing_utils.py:111: AssertionError
```

quickdahuk:
@victolee0 The gradient calculated (<= 1.1367e-04) differs slightly. This difference may not be significant, but it's better to have a much lower difference. @a-r-r-o-w, Do you see anything suspicious in our implementation?

quickdahuk:
@victolee0 I just found that if I used _use_reentry=False_, then it matches perfectly. I've updated the [code](https://gist.github.com/quickdahuk/cdada5050dc0a01892d338f353171ae6).

a-r-r-o-w:
The differing result is definitely suspicious. I would actually prefer having a refactor of the VAE so that we don't have to work with cache indexing in the way it's done here (and rather have it behave similar to what's done in CogVideoX and Mochi). If it's not possible to do it without indexing, we can consider removing the cache completely too..

The speed difference from using cache vs not is minimal from my past benchmarks with CogVideoX VAE and removing it, given it's complicated to do and probably doesn't save much time here, is a tradeoff that we could possibly make. cc @hlky @yiyixuxu Would you be able to take a look?

victolee0:
**quickdahuk's updated code passes the test code successfully. Should I create a new commit based on this code? I plan to commit by replacing the torch.utils.checkpoint.checkpoint function with self._gradient_checkpointing_func. This code applies gradient checkpointing only to the decoder.
If you're planning to refactor AutoEncoderKLWan to remove the cache mechanism, then a new commit might not be necessary, so please let me know.

@a-r-r-o-w

a-r-r-o-w:
If the test passes just for the decoder, let's add support for just the decoder now. We can revisit if someone wants encoder support too. I'm not sure if anyone on the team has the bandwidth to refactor the implementation at the moment, so it may take time until that's possible.

victolee0:
I've made the commit

src/diffusers/models/autoencoders/autoencoder_kl_wan.py
```diff
@@ -618,7 +618,14 @@ def __init__(
...
         self.gradient_checkpointing = False
 
-    def forward(self, x, feat_cache=None, feat_idx=[0]):
+    def forward(self, x, feat_cache=None, feat_idx=[0], latent=0):
+        if torch.is_grad_enabled() and self.gradient_checkpointing:
+            return self._gradient_checkpointing_func(self._decode, x, feat_cache, feat_idx, latent)
+        else:
+            return self._decode(x, feat_cache, feat_idx)
+
+    def _decode(self, x, in_cache=None, feat_idx=[0], latent=0):
```
yiyixuxu, line 627: what's this new argument `latent`?  it is not used here?

yiyixuxu (approved):
thanks!

a-r-r-o-w:
@yiyixuxu Any reason why we've not merged this yet?

yiyixuxu:
@bot /style

PR #11107, opened 2025-03-18
[Wan LoRAs] make T2V LoRAs compatible with Wan I2V

linoytsaban:
This PR was authored by @hlky to make T2V Wan LoRAs compatible with Wan I2V 🙌🏻

### DIFF
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -4255,22 +4255,23 @@ class WanLoraLoaderMixin(LoraBaseMixin):
         transformer: torch.nn.Module,
         state_dict,
     ):
-        num_blocks = len({k.split("blocks.")[1].split(".")[0] for k in state_dict})
-        is_i2v_lora = any("k_img" in k for k in state_dict) and any("v_img" in k for k in state_dict)
-        if not is_i2v_lora:
-            return state_dict
+        if any(k.startswith("blocks.") for k in state_dict):
+            num_blocks = len({k.split("blocks.")[1].split(".")[0] for k in state_dict})
+            is_i2v_lora = any("k_img" in k for k in state_dict) and any("v_img" in k for k in state_dict)
+            if not is_i2v_lora:
+                return state_dict
 
-        if transformer.config.image_dim is None:
-            return state_dict
+            if transformer.config.image_dim is None:
+                return state_dict
 
-        for i in range(num_blocks):
-            for o, c in zip(["k_img", "v_img"], ["add_k_proj", "add_v_proj"]):
-                state_dict[f"blocks.{i}.attn2.{c}.lora_A.weight"] = torch.zeros_like(
-                    state_dict[f"blocks.{i}.attn2.{o.replace('_img', '')}.lora_A.weight"]
-                )
-                state_dict[f"blocks.{i}.attn2.{c}.lora_B.weight"] = torch.zeros_like(
-                    state_dict[f"blocks.{i}.attn2.{o.replace('_img', '')}.lora_B.weight"]
-                )
+            for i in range(num_blocks):
+                for o, c in zip(["k_img", "v_img"], ["add_k_proj", "add_v_proj"]):
+                    state_dict[f"blocks.{i}.attn2.{c}.lora_A.weight"] = torch.zeros_like(
+                        state_dict[f"blocks.{i}.attn2.{o.replace('_img', '')}.lora_A.weight"]
+                    )
+                    state_dict[f"blocks.{i}.attn2.{c}.lora_B.weight"] = torch.zeros_like(
+                        state_dict[f"blocks.{i}.attn2.{o.replace('_img', '')}.lora_B.weight"]
+                    )
 
         return state_dict
 
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -4275,7 +4275,6 @@ class WanLoraLoaderMixin(LoraBaseMixin):
 
         return state_dict
 
-    # Copied from diffusers.loaders.lora_pipeline.CogVideoXLoraLoaderMixin.load_lora_weights with T2V LoRA->I2V LoRA option
     def load_lora_weights(
         self, pretrained_model_name_or_path_or_dict: Union[str, Dict[str, torch.Tensor]], adapter_name=None, **kwargs
     ):
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -4275,7 +4275,7 @@ class WanLoraLoaderMixin(LoraBaseMixin):
 
         return state_dict
 
-    # Copied from diffusers.loaders.lora_pipeline.CogVideoXLoraLoaderMixin.load_lora_weights
+    # Copied from diffusers.loaders.lora_pipeline.CogVideoXLoraLoaderMixin.load_lora_weights with T2V LoRA->I2V LoRA option
     def load_lora_weights(
         self, pretrained_model_name_or_path_or_dict: Union[str, Dict[str, torch.Tensor]], adapter_name=None, **kwargs
     ):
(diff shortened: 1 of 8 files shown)


hlky:
@bot /style

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -4249,7 +4249,33 @@ def lora_state_dict(
...
+                    state_dict[f"blocks.{i}.attn2.{c}.lora_A.weight"] = torch.zeros_like(
+                        state_dict[f"blocks.{i}.attn2.{o.replace('_img', '')}.lora_A.weight"]
+                    )
+                    state_dict[f"blocks.{i}.attn2.{c}.lora_B.weight"] = torch.zeros_like(
+                        state_dict[f"blocks.{i}.attn2.{o.replace('_img', '')}.lora_B.weight"]
+                    )
+
+        return state_dict
+
+    # Copied from diffusers.loaders.lora_pipeline.CogVideoXLoraLoaderMixin.load_lora_weights with T2V LoRA->I2V LoRA option
```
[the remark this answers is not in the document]
linoytsaban: oops yeah

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -4249,7 +4249,32 @@ def lora_state_dict(
...
-    # Copied from diffusers.loaders.lora_pipeline.CogVideoXLoraLoaderMixin.load_lora_weights
+    @classmethod
+    def _maybe_expand_t2v_lora_for_i2v(
+        cls,
+        transformer: torch.nn.Module,
+        state_dict,
+    ):
+        if any(k.startswith("blocks.") for k in state_dict):
+            num_blocks = len({k.split("blocks.")[1].split(".")[0] for k in state_dict})
+            is_i2v_lora = any("k_img" in k for k in state_dict) and any("v_img" in k for k in state_dict)
```
hlky, line 4260:
```suggestion
            is_i2v_lora = any("add_k_proj" in k for k in state_dict) and any("add_v_proj" in k for k in state_dict)
```
`k_img` and `v_img` are the original keys

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -4249,7 +4249,32 @@ def lora_state_dict(
...
+            if not is_i2v_lora:
```
hlky, line 4261:
```suggestion
            if is_i2v_lora:
```
Other way round

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -4249,7 +4249,32 @@ def lora_state_dict(
...
+                return state_dict
+
+            for i in range(num_blocks):
+                for o, c in zip(["k_img", "v_img"], ["add_k_proj", "add_v_proj"]):
+                    state_dict[f"blocks.{i}.attn2.{c}.lora_A.weight"] = torch.zeros_like(
+                        state_dict[f"blocks.{i}.attn2.{o.replace('_img', '')}.lora_A.weight"]
+                    )
+                    state_dict[f"blocks.{i}.attn2.{c}.lora_B.weight"] = torch.zeros_like(
+                        state_dict[f"blocks.{i}.attn2.{o.replace('_img', '')}.lora_B.weight"]
+                    )
```
hlky, line 4274:
```suggestion
                for o, c in zip(["k_img", "v_img"], ["add_k_proj", "add_v_proj"]):
                    state_dict[f"transformer.blocks.{i}.attn2.{c}.lora_A.weight"] = torch.zeros_like(
                        state_dict[f"transformer.blocks.{i}.attn2.to_k.lora_A.weight"]
                    )
                    state_dict[f"transformer.blocks.{i}.attn2.{c}.lora_B.weight"] = torch.zeros_like(
                        state_dict[f"transformer.blocks.{i}.attn2.to_k.lora_B.weight"]
                    )
```
`{o.replace('_img', '')}` would work on unconverted checkpoint, I think we can just use any key to get the shape, rank of this is probably not important? maybe `zeros` with any rank and `transformer.config.added_kv_proj_dim`.
Also, `transformer.` prefix is required.

hlky (approved):
Thanks @linoytsaban. I've left some comments with changes we discussed here https://huggingface.slack.com/archives/D08GXKYKGAK/p1742301801696229

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -4249,7 +4249,32 @@ def lora_state_dict(
...
+    def _maybe_expand_t2v_lora_for_i2v(
+        cls,
+        transformer: torch.nn.Module,
+        state_dict,
+    ):
+        if any(k.startswith("blocks.") for k in state_dict):
+            num_blocks = len({k.split("blocks.")[1].split(".")[0] for k in state_dict})
+            is_i2v_lora = any("k_img" in k for k in state_dict) and any("v_img" in k for k in state_dict)
+            if not is_i2v_lora:
+                return state_dict
```
sayakpaul, line 4262: We don't perform any extra operation if it's `is_i2v_lora`?

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -4249,7 +4249,32 @@ def lora_state_dict(
...
+
+            if transformer.config.image_dim is None:
+                return state_dict
```
sayakpaul, line 4265: This could be moved out at the top of this function.

src/diffusers/loaders/lora_pipeline.py
hlky: This is for loading T2V lora into I2V model, so if it's already I2V lora we return the state dict as-is.

src/diffusers/loaders/lora_pipeline.py
hlky: Might be slightly faster than checking the keys first, this is checking whether the transformer is I2V. T2V transformer config has `image_dim` as `None`.

src/diffusers/loaders/lora_pipeline.py
linoytsaban: Not sure I follow, `k_img` and `v_img` are the extra keys we have in I2V LoRAs, so if we already have these keys we just return the state dict 

for example (squish I2v lora, vs flat color t2v lora)
Screenshot 2025-03-19 at 15 30 20

Screenshot 2025-03-19 at 15 30 11

hlky: https://github.com/huggingface/diffusers/blob/fc28791fc8f4df6ac74afd1d0a0e0f1cea67aeec/src/diffusers/loaders/lora_conversion_utils.py#L1380-L1386

`k_img` and `v_img` are the non-Diffusers key, they are mapped to Diffusers keys `add_k_proj` and `add_v_proj` in `_convert_non_diffusers_wan_lora_to_diffusers` which happens during `lora_state_dict`

https://github.com/huggingface/diffusers/blob/fc28791fc8f4df6ac74afd1d0a0e0f1cea67aeec/src/diffusers/loaders/lora_pipeline.py#L4241-L4242

src/diffusers/loaders/lora_pipeline.py
hlky: T2V has [`transformer.config.image_dim` = `None`](https://huggingface.co/Wan-AI/Wan2.1-T2V-1.3B-Diffusers/blob/db5905750367e63c271ae4ea34e6ebfd71fba443/transformer/config.json#L10)

We have T2V loaded -> we are loading T2V lora -> `if transformer.config.image_dim is None` -> no state dict modification required -> return as-is

We have I2V loaded -> we are loading T2V lora -> `if transformer.config.image_dim is None` does not match as `image_dim` is not `None`  -> continue with state dict modifications

linoytsaban: you're absolutely right🙌🏻

linoytsaban:
@bot /style

hlky:
@bot /style

hlky:
https://github.com/huggingface/diffusers/actions/runs/13948751559/job/39042360914
You don't have `admin` permissions for some reason @linoytsaban

hlky (approved):

sayakpaul (approved):
Sleek! 

I would suggest adding a snippet about this in our Wan [docs](https://huggingface.co/docs/diffusers/main/en/api/pipelines/wan).

Should be good to merge after that!

linoytsaban:
> I would suggest adding a snippet about this in our Wan [docs](https://huggingface.co/docs/diffusers/main/en/api/pipelines/wan).

@sayakpaul  Yeah we should definitely have some documentation for it, but I'm thinking maybe in a separate PR to add it somewhere else? I think for other pipelines we didnt include lora examples in the pipeline page so maybe makes more sense as part of LoRA documentation dedicated to videos?

hlky:
Yeah I think docs can be added in a follow up as we'll be overhauling all the pipeline doc pages similar to #11112 anyway and we don't have a great example for this right now.

sayakpaul:
I respectfully disagree here. Just a snippet and saying that T2V LoRAs are compatible with I2V LoRAs won't be too much. 

It is quite easy to forget it if we were to wait till the overhaul to complete. 

> we don't have a great example for this right now.

Well, if we don't know with examples if it works as expected, then it could wait until we do I guess. If we're matching the Replicate reference in terms of quality, I am fine with that.

PR #11115, opened 2025-03-19
AutoModel

### LINKED ISSUE #10059: [core] introduce `AutoModel`
Now that we have quite a few model classes (and transformer model classes), I wonder if it makes sense to have an AutoModel class. Oftentimes, users initialize the model separately and pass it to a `DiffusionPipeline`:

```py
transformer = "..."
pipeline = DiffusionPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", transformer=transformer)
```

It's common to see this happening, especially in the realm of quantization. And in these cases, an AutoModel class could be quite nice with the following developer experience:

```py
transformer = AutoModel.from_pretrained("...", subfolder="...")
```

hlky:
# What does this PR do?

Fixes #10059

## Who can review?


sayakpaul (commented):
Nice! I think this will be so much better as an UX improvement. Cc: @vladmandic you might like this :) 

Let's add docs and tests after @DN6 reviews.

ParagEkbote:
I anticipate this functionality to be quite helpful in a future version of diffusers. Does it make sense to add this PR to the future release section of the diffusers roadmap?

hlky:
@vladmandic Do you have an example pseudo-code of your expected usage?

vladmandic:
> @vladmandic Do you have an example pseudo-code of your expected usage?

sorry, i was too fast with my comment. if using auto-pipeline, we dont know what's the model ahead of time.
my comment was more general, if we do use specific pipeline, there should be a default model specified.

one thing with automodel i'd love to see is the model type.
e.g., i want to list available txt2img pipelines so i can display to user which models are supported. right now everything is a flat list without any way to differentiate them.

src/diffusers/models/auto_model.py
```diff
@@ -0,0 +1,311 @@
...
+            "token": token,
+            "local_files_only": local_files_only,
+            "revision": revision,
+            "subfolder": subfolder,
+        }
+
+        config = cls.load_config(pretrained_model_or_path, **load_config_kwargs)
+        orig_class_name = config["_class_name"]
+
+        model_cls = _get_task_class(AUTO_MODEL_MAPPING, orig_class_name)
```
DN6, line 308: Rather than maintain this big mapping can't we use importlib? All the models are available in the main diffusers init right? Similar to how we do it here?

https://github.com/huggingface/diffusers/blob/c51b6bd83749faaa163064668f4d6dd5c9d2aac9/src/diffusers/pipelines/pipeline_loading_utils.py#L354-L358

yiyixuxu: yeah, agree, probably don't need a mapping here,  could use the same logic as in `from_pretrained`

yiyixuxu: if we can get rid of mapping, everything will be supported automatically too I think

src/diffusers/utils/dummy_pt_objects.py
```diff
@@ -895,6 +910,21 @@ def from_pretrained(cls, *args, **kwargs):
         requires_backends(cls, ["torch"])
 
 
+class TransformerTemporalModel(metaclass=DummyObject):
```
yiyixuxu, line 913: why this?

hlky: It was missing, needed to be added for the test to pass.

DN6 (approved):
Nice work 👍🏽

PR #11125, opened 2025-03-20
Modify the implementation of retrieve_timesteps in CogView4-Control.

zRzRzRzRzRzRzR:
Use CogView4 Models one


src/diffusers/pipelines/cogview4/pipeline_cogview4_control.py
```diff
@@ -100,10 +100,19 @@ def retrieve_timesteps(
         `Tuple[torch.Tensor, int]`: A tuple where the first element is the timestep schedule from the scheduler and the
         second element is the number of inference steps.
     """
+    accepts_timesteps = "timesteps" in set(inspect.signature(scheduler.set_timesteps).parameters.keys())
```
a-r-r-o-w, line 103: Let's update the `# Copied from` comment to point to CogView4 pipeline file

zRzRzRzRzRzRzR: Is that how I understand it?

src/diffusers/pipelines/cogview4/pipeline_cogview4_control.py
```diff
@@ -68,7 +68,7 @@ def calculate_shift(
     return mu
 
 
-# Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion.retrieve_timesteps
+# Copied from diffusers.pipelines.cogview4.pipeline_cogview4.retrieve_timesteps
```
a-r-r-o-w, line 71: Just the change here is correct @zRzRzRzRzRzRzR

src/diffusers/pipelines/cogview4/pipeline_cogview4_control.py
```diff
@@ -100,10 +100,20 @@ def retrieve_timesteps(
         `Tuple[torch.Tensor, int]`: A tuple where the first element is the timestep schedule from the scheduler and the
         second element is the number of inference steps.
     """
+    # Copied from diffusers.pipelines.cogview4.pipeline_cogview4.retrieve_timesteps
```
a-r-r-o-w, line 103:
```suggestion
```

This is not needed. Please also run `make fix-copies` after removing this to ensure that both the implementation in this file and in pipeline_cogview4.py is the same 🤗

zRzRzRzRzRzRzR: I have completed the verification, running make fix-copies did not produce any errors or other improvement suggestions

zRzRzRzRzRzRzR:
I have completed the modification. @a-r-r-o-w can check if there are any other places that need to be modified.

a-r-r-o-w (approved):

src/diffusers/pipelines/cogview4/pipeline_cogview4_control.py
```diff
@@ -533,8 +541,6 @@ def __call__(
             max_sequence_length (`int`, defaults to `224`):
                 Maximum sequence length in encoded prompt. Can be set to other values but may lead to poorer results.
 
-        Examples:
-
```
a-r-r-o-w, line 537: @zRzRzRzRzRzRzR These should not be removed, otherwise the tests will fail: https://github.com/huggingface/diffusers/actions/runs/14009637640/job/39228136089?pr=11125#step:15:64

Let's add this back and we should be good to merge

zRzRzRzRzRzRzR: Done

a-r-r-o-w:
@bot /style

PR #11126, opened 2025-03-20
remove F.rms_norm for now

yiyixuxu:


chengzeyi:
Code to reproduce:

```python
# diffusers 2c1ed50fc57f154768364e4506d7bab9daebf83d
# torch 2.5.0
# Recompilation caused by calling torch.nn.functional.rms_norm(), which calls aten.size instead of aten.sym_size

import os

os.environ["TORCH_LOGS"] = "recompiles,dynamic"

import torch

pipe.transformer = torch.compile(pipe.transformer, dynamic=True, backend="eager")

for height, width in [(1536, 1024), (1024, 512)]:
    image = pipe(
        prompt,
        height=1024,
        width=1024,
        guidance_scale=3.5,
        num_inference_steps=50,
        max_sequence_length=512,
    ).images[0]
    image.save(f"flux-dev_{height}x{width}.png")
```

PR #11132, opened 2025-03-21
[bistandbytes] improve replacement warnings for bnb

sayakpaul:


ukaprch:
I do have a question. Why can't we use ".to(device=device)" for bitsandbytes load 8 bit = True as we do for load 4 bit = True?

sayakpaul:
8bit models don't support that placement.

sayakpaul:
@DN6 a gentle ping

PR #11150, opened 2025-03-25
add a timestep scale for sana-sprint teacher model

lawrence-cj:
This PR add a `timestep_scale` for SANA-Sprint's teacher model
This timestep_scale will be applied to flow matching timesteps and rescales them form [1~1000] to [0~1]

Cc: @yiyixuxu


yiyixuxu (approved):
ohh that's all it takes? awesome!

PR #11152, opened 2025-03-25
fix flux controlnet bug

free001style:
# What does this PR do?

Before this if txt_ids was 3d tensor, line with txt_ids[:1] concat txt_ids by batch dim. Now we first check that txt_ids is 2d tensor (or take first batch element) and then concat by token dim

### DIFF
--- a/src/diffusers/models/controlnets/controlnet_flux.py
+++ b/src/diffusers/models/controlnets/controlnet_flux.py
@@ -298,15 +298,6 @@ class FluxControlNetModel(ModelMixin, ConfigMixin, PeftAdapterMixin):
         )
         encoder_hidden_states = self.context_embedder(encoder_hidden_states)
 
-        if self.union:
-            # union mode
-            if controlnet_mode is None:
-                raise ValueError("`controlnet_mode` cannot be `None` when applying ControlNet-Union")
-            # union mode emb
-            controlnet_mode_emb = self.controlnet_mode_embedder(controlnet_mode)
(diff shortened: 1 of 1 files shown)


lordsoffallen:
Still getting the error even with your fix:

RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 1 but got size 2 for tensor number 1 in the list.

I am not sure where the issue lies but controlnet inpainting is not working at all...

free001style:
> RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 1 but got size 2 for tensor number 1 in the list.

Are you getting error in this line? txt_ids = torch.cat([txt_ids[:1], txt_ids], dim=0)
Because I didn't face with this issues(((

lordsoffallen:
I am getting it in this line: encoder_hidden_states = torch.cat([controlnet_mode_emb, encoder_hidden_states], dim=1)

It happens when I use multicontrol code but only have one control defined. Reverting to use single control i made it work. I don't really know where the error comes from honestly, too many different stuff going on.

hlky (commented):
Hi @free001style. Is there a case where `txt_ids` is still being passed as 3d?

@lordsoffallen Can you create an issue for the multicontrolnet case with a reproducer?

free001style:
@hlky both `txt_ids`  and `img_ids` can be 3d (if user feed them as 3d in `forward`) but in that case I leave only first batch element. I don't know why original implementation need only 2d tensors (I think it's better to fix this in order to use 3d tensors).

Issue #11148, opened 2025-03-24
[PEFT] Loading a LoRA in PeftAdapterMixin does not set _hf_peft_config_loaded to True; causes issues with {enable,disable}_adapters


kentdan3msu:
### Describe the bug

Certain adapter functions in PeftAdapterMixin such as `enable_adapters` and `disable_adapters` do not work if the only adapters that were added were LoRAs loaded with `load_lora_adapter`, instead failing with a ValueError message indicating that no adapters are loaded:
```
>>> transformer.enable_adapters()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/<redacted>/diffusers/loaders/peft.py", line 604, in enable_adapters
    raise ValueError("No adapter loaded. Please load an adapter first.")
ValueError: No adapter loaded. Please load an adapter first.

```

I encountered this bug when I was trying to apply a LoRA to a `SD3Transformer2DModel` instead of a `StableDiffusion3Pipeline`, as the training script I'm using does not initialize a `StableDiffusion3Pipeline` during training.

Based on a cursory glance of the code, it looks like the issue is that `_hf_peft_config_loaded` is not set to True when `load_lora_adapter` is called, before `inject_adapter_in_model`: https://github.com/huggingface/diffusers/blob/7aac77affa17b6b504b0a406aacb471c5226b36d/src/diffusers/loaders/peft.py#L301-L322

Compare to the corresponding code in `add_adapter`: https://github.com/huggingface/diffusers/blob/7aac77affa17b6b504b0a406aacb471c5226b36d/src/diffusers/loaders/peft.py#L504-L520

Setting the `_hf_peft_config_loaded` flag to True works around the issue.

I intend to submit a PR unless there's a reason why the code is set up this way.

### Reproduction

```
from diffusers import SD3Transformer2DModel, StableDiffusion3Pipeline
from peft import LoraConfig
from peft.utils import get_peft_model_state_dict
import os
sd3_path='/path/to/stable-diffusion-3-medium-diffusers'

# Prerequisite: we need to have a LoRA to test with.
# In lieu of finding a premade LoRA, we can just make a fresh dummy LoRA from scratch
transformer = SD3Transformer2DModel.from_pretrained(sd3_path, subfolder='transformer')
[...]

PR #11155, opened 2025-03-25
Set self._hf_peft_config_loaded to True when LoRA is loaded using `load_lora_adapter` in PeftAdapterMixin class

kentdan3msu:
# What does this PR do?
This PR changes the behavior of the PeftAdapterMixin interface class so that LoRA modules loaded with `load_lora_adapter` set the `self._hf_peft_config_loaded` flag to True after successful injection. This change fixes issue #11148.

Previously, the `self._hf_peft_config_loaded` flag was not set in `load_lora_adapter()`, only in `add_adapter()` and `add_adapters()`. Some functions within the PEFT code such as `{enable,disable}_adapters()` depend on this flag being set before they will function properly. I believe this change should bring parity between the `add_adapter()`/`add_adapters()` and `load_lora_adapter()` functions.

Since the PeftAdapterMixin class is used by many classes, such as SD3Transformer2DModel (which I was using when I encountered the original issue), this could potentially fix LoRA loading/enabling/disabling issues in many classes across the diffusers library.

Small note: in `add_adapter()`/`add_adapters()`, the flag is set before the module is actually injected. In my proposed PR, I end up doing it afterwards, in case there is an issue with the injection process. If there is a good reason to set the flag earlier, I can alter the PR accordingly.

## Before submitting

## Who can review?
@sayakpaul

### DIFF
--- a/src/diffusers/loaders/peft.py
+++ b/src/diffusers/loaders/peft.py
@@ -307,6 +307,9 @@ class PeftAdapterMixin:
             try:
                 inject_adapter_in_model(lora_config, self, adapter_name=adapter_name, **peft_kwargs)
                 incompatible_keys = set_peft_model_state_dict(self, state_dict, adapter_name, **peft_kwargs)
+                # Set peft config loaded flag to True if module has been successfully injected and incompatible keys retrieved
+                if not self._hf_peft_config_loaded:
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):

sayakpaul:
Thanks very much!

PR #11161, opened 2025-03-27
[Refactor] Minor Improvement for import utils

ishan-modi:
# What does this PR do?

At times pkg_name (used for import) and dist_name (used for install) are different. The `_is_package_available` function doesnt work for that see [here](https://github.com/huggingface/diffusers/blob/main/src/diffusers/utils/import_utils.py#L191-L197). This PR helps resolve this through `_is_package_available` function.

## Who can review?

@DN6 @hlky


src/diffusers/utils/import_utils.py
```diff
@@ -54,12 +54,17 @@
 _is_google_colab = "google.colab" in sys.modules or any(k.startswith("COLAB_") for k in os.environ)
 
 
-def _is_package_available(pkg_name: str):
+def _is_package_available(pkg_name: str, get_dist_name: bool = False) -> tuple[bool, str]:
```
hlky, line 57:
```suggestion
def _is_package_available(pkg_name: str, get_dist_name: bool = False) -> Tuple[bool, str]:
```
And `from typing import Tuple` if not already imported, we need to still support 3.8

hlky (changes requested):

hlky:
@bot /style

hlky (commented):
Thanks for the iteration. Looks like `packages_distributions` was only added in 3.10
https://docs.python.org/3/library/importlib.metadata.html#importlib.metadata.packages_distributions
Can you look into alternatives?

ishan-modi:
Couldn't find alternatives, but pulled out some code from packages_distribution from `python-3.10` and added to `_is_package_available` function

```
```

hlky:
LGTM, thanks!

hlky (approved):

src/diffusers/utils/import_utils.py
```diff
@@ -186,15 +211,7 @@ def _is_package_available(pkg_name: str):
 _gguf_available, _gguf_version = _is_package_available("gguf")
 _torchao_available, _torchao_version = _is_package_available("torchao")
 _bitsandbytes_available, _bitsandbytes_version = _is_package_available("bitsandbytes")
-_torchao_available, _torchao_version = _is_package_available("torchao")
```
yiyixuxu, line 189: is this intended?

hlky: It's duplicate

https://github.com/huggingface/diffusers/blob/68bb9553d3f774a2f539f6eedd0221e9f5bdb3ce/src/diffusers/utils/import_utils.py#L212

yiyixuxu (approved):
thanks!

PR #11180, opened 2025-03-31
First Block Cache

a-r-r-o-w:
FBC Reference: https://github.com/chengzeyi/ParaAttention

### Minimal example

```python
import torch
from diffusers import CogView4Pipeline
from diffusers.hooks import apply_first_block_cache, FirstBlockCacheConfig

pipe = CogView4Pipeline.from_pretrained("THUDM/CogView4-6B", torch_dtype=torch.bfloat16)
pipe.to("cuda")

apply_first_block_cache(pipe.transformer, FirstBlockCacheConfig(threshold=0.3))

prompt = "A photo of an astronaut riding a horse on mars"
image = pipe(prompt, generator=torch.Generator().manual_seed(42)).images[0]
image.save("output.png")
```

### Benchmark scripts

Threshold vs Generation time (in seconds) for each model. In general, values below 0.2 work well for First Block Cache depending on the model. Higher values leads to blurring and artifacting

| Threshold | CogView4 | HunyuanVideo | LTX Video | Wan  | Flux |
|-----------|-------------|------------------|---------------|----------|----------|
| 0.00      | 40.51       | 121.85           | 33.64         | 222.17   | 16.47    |
| 0.03      | -           | -                | 27.14         | -        | -        |
| 0.05      | 24.08       | 62.47            | 21.73         | 139.26   | 12.63    |
| 0.10      | 17.55       | 41.84            | 15.10         | 89.99    | 9.27     |
| 0.20      | 12.99       | 28.11            | 10.29         | 57.01    | 5.91     |
| 0.40      | -           | 18.93            | 7.25          | -        | 3.70     |
| 0.50      | -           | 16.65            | -             | -        | 3.13     |

<details>
<summary> CogView4 </summary>

```python
import argparse
import pathlib

import torch
from diffusers import CogView4Pipeline
from diffusers.hooks import FirstBlockCacheConfig


def main(args):
    output_dir = pathlib.Path(args.output_dir)
    output_dir.mkdir(parents=True, exist_ok=True)

    set_verbosity_debug()
    pipe = CogView4Pipeline.from_pretrained("THUDM/CogView4-6B", torch_dtype=torch.bfloat16)
    pipe.to("cuda")

    prompt = "A photo of the eye of agomotto, a mystical object from the Marvel universe, with intricate details and a glowing effect, in a fantasy style. The eye sits at the center of a mystical landscape surrounded by green trees, vibrant flowers and orbs of light. The scene is illuminated by a soft, ethereal glow, creating a magical atmosphere. The eye itself is detailed with swirling patterns and a radiant light"
    negative_prompt = "bad anatomy, ugly, blurry, out of focus, low quality, worst quality, normal quality, jpeg artifacts, signature, watermark, username, artist name, text, error, missing fingers, extra digit, fewer digits, cropped, worst quality, low quality, normal quality, jpeg artifacts, signature, watermark, username, artist name, text, error, missing fingers, extra digit, fewer digits, cropped"

    # Warmup
    pipe(prompt=prompt, negative_prompt=negative_prompt, num_inference_steps=2, generator=torch.Generator().manual_seed(42)).images[0]

    # Benchmark
    for threshold in [0, 0.05, 0.1, 0.2, 0.4, 0.5]:
        print(f"Using threshold: {threshold}")
        filename = f"output_{threshold:.5f}.png"

        start.record()
        if threshold == 0:
            image = pipe(prompt=prompt, negative_prompt=negative_prompt, generator=torch.Generator().manual_seed(42)).images[0]
        else:
            pipe.transformer.enable_cache(FirstBlockCacheConfig(threshold=threshold))
            image = pipe(prompt=prompt, negative_prompt=negative_prompt, generator=torch.Generator().manual_seed(42)).images[0]
            pipe.transformer.disable_cache()
        end.record()
        torch.cuda.synchronize()
        
        elapsed_time = start.elapsed_time(end)
        print(f"Elapsed time: {elapsed_time / 1000:.2f}s")
        print(f"Output saved to {output_dir / filename}")
        
        image.save((output_dir / filename).as_posix())


def get_args():
    parser.add_argument("--output_dir", type=str, default="output_cogview4")
    return parser.parse_args()


if __name__ == "__main__":
    args = get_args()
    main(args)
```
</details>

<details>
<summary> HunyuanVideo </summary>

```python
import argparse
import pathlib

import torch
from diffusers.hooks import FirstBlockCacheConfig


def main(args):
    output_dir = pathlib.Path(args.output_dir)
    output_dir.mkdir(parents=True, exist_ok=True)

    set_verbosity_debug()
    )
    pipe.vae.enable_tiling()
    pipe.to("cuda")

    prompt = "A cat walks on the grass, realistic"

    # Warmup
    pipe(prompt=prompt, height=320, width=512, num_frames=61, num_inference_steps=2, generator=torch.Generator().manual_seed(42)).frames[0]

    # Benchmark
    for threshold in [0, 0.05, 0.1, 0.2, 0.4, 0.5]:
        print(f"Using threshold: {threshold}")
        filename = f"output_{threshold:.5f}.mp4"

        start.record()
        if threshold == 0:
            video = pipe(prompt=prompt, height=320, width=512, num_frames=61, generator=torch.Generator().manual_seed(42)).frames[0]
        else:
            pipe.transformer.enable_cache(FirstBlockCacheConfig(threshold=threshold))
            video = pipe(prompt=prompt, height=320, width=512, num_frames=61, generator=torch.Generator().manual_seed(42)).frames[0]
            pipe.transformer.disable_cache()
        end.record()
        torch.cuda.synchronize()
        
        elapsed_time = start.elapsed_time(end)
        print(f"Elapsed time: {elapsed_time / 1000:.2f}s")
        print(f"Output saved to {output_dir / filename}")
        
        export_to_video(video, (output_dir / filename).as_posix(), fps=16)


def get_args():
    parser.add_argument("--output_dir", type=str, default="output_hunyuanvideo")
    return parser.parse_args()


if __name__ == "__main__":
    args = get_args()
    main(args)
```
</details>

<details>
<summary> LTX Video </summary>

```python
import argparse
import pathlib

import torch
from diffusers.hooks import FirstBlockCacheConfig


def main(args):
    output_dir = pathlib.Path(args.output_dir)
    output_dir.mkdir(parents=True, exist_ok=True)

    set_verbosity_debug()
    pipe.to("cuda")

    # Warmup
    pipe(prompt=prompt, negative_prompt=negative_prompt, height=480, width=704, num_frames=161, num_inference_steps=2, generator=torch.Generator().manual_seed(42)).frames[0]

    # Benchmark
    for threshold in [0, 0.03, 0.05, 0.1, 0.2, 0.4]:
        print(f"Using threshold: {threshold}")
        filename = f"output_{threshold:.5f}.mp4"

        start.record()
        if threshold == 0:
            video = pipe(prompt=prompt, negative_prompt=negative_prompt, height=480, width=704, num_frames=161, generator=torch.Generator().manual_seed(42)).frames[0]
        else:
            pipe.transformer.enable_cache(FirstBlockCacheConfig(threshold=threshold))
            video = pipe(prompt=prompt, negative_prompt=negative_prompt, height=480, width=704, num_frames=161, generator=torch.Generator().manual_seed(42)).frames[0]
            pipe.transformer.disable_cache()
        end.record()
        torch.cuda.synchronize()
        
        elapsed_time = start.elapsed_time(end)
        print(f"Elapsed time: {elapsed_time / 1000:.2f}s")
        print(f"Output saved to {output_dir / filename}")
        
        export_to_video(video, (output_dir / filename).as_posix(), fps=24)


def get_args():
    parser.add_argument("--output_dir", type=str, default="output_hunyuanvideo")
    return parser.parse_args()


if __name__ == "__main__":
    args = get_args()
    main(args)
```
</details>

<details>
<summary> Wan </summary>

```python
import argparse
import pathlib

import torch
from diffusers import AutoencoderKLWan, WanPipeline
from diffusers.hooks import FirstBlockCacheConfig


def main(args):
    output_dir = pathlib.Path(args.output_dir)
    output_dir.mkdir(parents=True, exist_ok=True)

    set_verbosity_debug()
    model_id = "Wan-AI/Wan2.1-T2V-1.3B-Diffusers"
    vae = AutoencoderKLWan.from_pretrained(model_id, subfolder="vae", torch_dtype=torch.float32)
    pipe = WanPipeline.from_pretrained(model_id, vae=vae, torch_dtype=torch.bfloat16)
    pipe.to("cuda")

    prompt = "A cat and a dog baking a cake together in a kitchen. The cat is carefully measuring flour, while the dog is stirring the batter with a wooden spoon. The kitchen is cozy, with sunlight streaming through the window."
    negative_prompt = "Bright tones, overexposed, static, blurred details, subtitles, style, works, paintings, images, static, overall gray, worst quality, low quality, JPEG compression residue, ugly, incomplete, extra fingers, poorly drawn hands, poorly drawn faces, deformed, disfigured, misshapen limbs, fused fingers, still picture, messy background, three legs, many people in the background, walking backwards"

    # Warmup
    pipe(prompt=prompt, negative_prompt=negative_prompt, num_inference_steps=2, generator=torch.Generator().manual_seed(42)).frames[0]

    # Benchmark
    for threshold in [0, 0.05, 0.1, 0.2]:
        print(f"Using threshold: {threshold}")
        filename = f"output_{threshold:.5f}.mp4"

        start.record()
        if threshold == 0:
            video = pipe(prompt=prompt, negative_prompt=negative_prompt, generator=torch.Generator().manual_seed(42)).frames[0]
        else:
            pipe.transformer.enable_cache(FirstBlockCacheConfig(threshold=threshold))
            video = pipe(prompt=prompt, negative_prompt=negative_prompt, generator=torch.Generator().manual_seed(42)).frames[0]
            pipe.transformer.disable_cache()
        end.record()
        torch.cuda.synchronize()
        
        elapsed_time = start.elapsed_time(end)
        print(f"Elapsed time: {elapsed_time / 1000:.2f}s")
        print(f"Output saved to {output_dir / filename}")
        
        export_to_video(video, (output_dir / filename).as_posix(), fps=16)


def get_args():
    parser.add_argument("--output_dir", type=str, default="output_wan")
    return parser.parse_args()


if __name__ == "__main__":
    args = get_args()
    main(args)
```
</details>

<details>
<summary> Flux </summary>

```python
import argparse
import pathlib

import torch
from diffusers.hooks import FirstBlockCacheConfig


def main(args):
    output_dir = pathlib.Path(args.output_dir)
    output_dir.mkdir(parents=True, exist_ok=True)

    set_verbosity_debug()
    pipe = FluxPipeline.from_pretrained("black-forest-labs/Flux.1-Dev", torch_dtype=torch.bfloat16, cache_dir="/raid/.cache/huggingface")
    pipe.to("cuda")

    prompt = "A photo of the eye of agomotto, a mystical object from the Marvel universe, with intricate details and a glowing effect, in a fantasy style. The eye sits at the center of a mystical landscape surrounded by green trees, vibrant flowers and orbs of light. The scene is illuminated by a soft, ethereal glow, creating a magical atmosphere. The eye itself is detailed with swirling patterns and a radiant light"

    # Warmup
    pipe(prompt=prompt, num_inference_steps=2, generator=torch.Generator().manual_seed(42)).images[0]

    # Benchmark
    for threshold in [0, 0.05, 0.1, 0.2, 0.4, 0.5]:
        print(f"Using threshold: {threshold}")
        filename = f"output_{threshold:.5f}.png"

        start.record()
        if threshold == 0:
            image = pipe(prompt=prompt, generator=torch.Generator().manual_seed(42)).images[0]
        else:
            pipe.transformer.enable_cache(FirstBlockCacheConfig(threshold=threshold))
            image = pipe(prompt=prompt, generator=torch.Generator().manual_seed(42)).images[0]
            pipe.transformer.disable_cache()
        end.record()
        torch.cuda.synchronize()
        
        elapsed_time = start.elapsed_time(end)
        print(f"Elapsed time: {elapsed_time / 1000:.2f}s")
        print(f"Output saved to {output_dir / filename}")
        
        image.save((output_dir / filename).as_posix())


def get_args():
    parser.add_argument("--output_dir", type=str, default="output_flux")
    return parser.parse_args()


if __name__ == "__main__":
    args = get_args()
    main(args)

```
</details>

### Visual result comparison

<details>
<summary> CogView4 </summary>

<table align=center>
<tr>
  <th align=center>threshold=0.00</th>
  <th align=center>threshold=0.05</th>
</tr>
<tr>
  <td align=center></td>
  <td align=center></td>
</tr>

<tr>
  <th align=center>threshold=0.10</th>
  <th align=center>threshold=0.20</th>
</tr>
<tr>
  <td align=center></td>
  <td align=center></td>
</tr>
</table>
</details>

<details>
<summary> Hunyuan Video </summary>

<table align=center>
<tr>
  <th align=center>threshold=0.00</th>
  <th align=center>threshold=0.05</th>
</tr>
<tr>
  <td align=center><video src="https://github.com/user-attachments/assets/ec1091cc-b0a7-4b4d-9740-a258cc5f48c4"> Your browser does not support the video tag. </video></td>
  <td align=center><video src="https://github.com/user-attachments/assets/221f833a-151f-43bc-a51e-7a6680efad16"> Your browser does not support the video tag. </video></td>
</tr>

<tr>
  <th align=center>threshold=0.10</th>
  <th align=center>threshold=0.20</th>
</tr>
<tr>
  <td align=center><video src="https://github.com/user-attachments/assets/d523812f-f6eb-40a5-9cb6-2cdf999906f4"> Your browser does not support the video tag. </video></td>
  <td align=center><video src="https://github.com/user-attachments/assets/778dabb7-ac0e-4431-8ef3-ce5a204543ff"> Your browser does not support the video tag. </video></td>
</tr>

<tr>
  <th align=center>threshold=0.40</th>
  <th align=center>threshold=0.50</th>
</tr>
<tr>
  <td align=center><video src="https://github.com/user-attachments/assets/3854f5a3-e781-49fc-8130-5cc26cb16e76"> Your browser does not support the video tag. </video></td>
  <td align=center><video src="https://github.com/user-attachments/assets/5c880037-641e-48fc-95bd-c4cc1552b658"> Your browser does not support the video tag. </video></td>
</tr>
</table>
</details>

<details>
<summary> LTX Video </summary>

<table align=center>
<tr>
  <th align=center>threshold=0.00</th>
  <th align=center>threshold=0.05</th>
</tr>
<tr>
  <td align=center><video src="https://github.com/user-attachments/assets/805525e1-736f-4372-8f49-c0572d94bb05"> Your browser does not support the video tag. </video></td>
  <td align=center><video src="https://github.com/user-attachments/assets/0703eaab-6074-47f4-a55b-41c88bf7ef21"> Your browser does not support the video tag. </video></td>
</tr>

<tr>
  <th align=center>threshold=0.10</th>
  <th align=center>threshold=0.20</th>
</tr>
<tr>
  <td align=center><video src="https://github.com/user-attachments/assets/37e6af8b-dd94-4439-83c2-67ccba34010a"> Your browser does not support the video tag. </video></td>
  <td align=center><video src="https://github.com/user-attachments/assets/0b2033a1-8c5c-4e9a-83f7-d1e0246a731a"> Your browser does not support the video tag. </video></td>
</tr>

<tr>
  <th align=center>threshold=0.40</th>
  <th align=center>threshold=0.50</th>
</tr>
<tr>
  <td align=center><video src="https://github.com/user-attachments/assets/16eda9ee-5977-47aa-8d46-af0e1346d979"> Your browser does not support the video tag. </video></td>
  <td align=center><video src="https://github.com/user-attachments/assets/cf169d23-8eee-4cf4-9621-19d0e289df59"> Your browser does not support the video tag. </video></td>
</tr>
</table>
</details>

<details>
<summary> Wan </summary>

<table align=center>
<tr>
  <th align=center>threshold=0.00</th>
  <th align=center>threshold=0.05</th>
</tr>
<tr>
  <td align=center><video src="https://github.com/user-attachments/assets/96b9e387-2780-43e1-bf6b-b29af0e74a49"> Your browser does not support the video tag. </video></td>
  <td align=center><video src="https://github.com/user-attachments/assets/c1077977-3ece-4284-ba33-cc57e78ec229"> Your browser does not support the video tag. </video></td>
</tr>

<tr>
  <th align=center>threshold=0.10</th>
  <th align=center>threshold=0.20</th>
</tr>
<tr>
  <td align=center><video src="https://github.com/user-attachments/assets/fd6f6f3a-0ebf-4247-ab36-ef141b043e1c"> Your browser does not support the video tag. </video></td>
  <td align=center><video src="https://github.com/user-attachments/assets/83522b6a-b533-4a97-860d-6e32a66830a0"> Your browser does not support the video tag. </video></td>
</tr>
</table>
</details>

<details>
<summary> Flux </summary>

<table align=center>
<tr>
  <th align=center>threshold=0.00</th>
  <th align=center>threshold=0.05</th>
</tr>
<tr>
  <td align=center></td>
  <td align=center></td>
</tr>

<tr>
  <th align=center>threshold=0.10</th>
  <th align=center>threshold=0.20</th>
</tr>
<tr>
  <td align=center></td>
  <td align=center></td>
</tr>

<tr>
  <th align=center>threshold=0.40</th>
  <th align=center>threshold=0.50</th>
</tr>
<tr>
  <td align=center></td>
  <td align=center></td>
</tr>
</table>
</details>

### Using with `torch.compile`

- There is a forced graph break for the data-dependant control flow branching. This portion of code will always run in eager mode
- There are a few recompilations triggered at a weird/unexpected location - the attention processor invocation. This only happens when using hooks, so I believe the current hook implementation is making torch.compile tracing add some unnecessary id/type guards. This will be tackled in the future since I haven't been able to make much progress into rewriting it


src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -79,10 +79,14 @@ def __init__(self, dim: int, num_attention_heads: int, attention_head_dim: int,
     def forward(
```
a-r-r-o-w, line 79: cc @yiyixuxu for reviewing changes to the transformer here. The changes were made to simplify some of the code required to make cache techniques work somewhat more easily without even more if-else branching.

Ideally, if we stick to implementing models such that all blocks take in both hidden_states and encoder_hidden_states, and always return `(hidden_states, encoder_hidden_states)` from the block, a lot of design choices in the hook-based code can be simplified.

For now, I think these changes should be safe and come without any significant overhead to generation time (I haven't benchmarked though).

src/diffusers/hooks/first_block_cache.py
```diff
@@ -0,0 +1,222 @@
+# Copyright 2024 The HuggingFace Team. All rights reserved.
```
a-r-r-o-w, line 1: cc @chengzeyi Would be super cool if you could give this PR a review since we're trying to integrate FBC to work with all supported models!

Currently, I've only done limited testing on few models but it should be easily extendable to all

src/diffusers/hooks/_helpers.py
```diff
@@ -0,0 +1,199 @@
...
+        cls._registry[model_class] = metadata
+
+    @classmethod
+    def get(cls, model_class: Type) -> TransformerBlockMetadata:
+        if model_class not in cls._registry:
+            raise ValueError(f"Model class {model_class} not registered.")
+        return cls._registry[model_class]
+
+
+def _register_transformer_blocks_metadata():
```
a-r-r-o-w, line 53: cc @DN6 For now, this PR only adds metadata for transformer blocks. The information here will be filled up over time for more models to simplify the assumptions made in hook-based code and make it cleaner to work with. More metadata needs to be maintained for transformer implementations to simplify cache methods like FasterCache, which I'll cover in future PRs

src/diffusers/hooks/hooks.py
```diff
@@ -23,6 +23,70 @@
...
 
 
+class BaseState:
+    def reset(self, *args, **kwargs) -> None:
+        raise NotImplementedError(
+            "BaseState::reset is not implemented. Please implement this method in the derived class."
+        )
+
+
+class BaseMarkedState(BaseState):
```
a-r-r-o-w, line 33: To explain simply, a "marked" state is a copies of a state object for different batches of data. In our pipelines, we do the following:
- concatenate unconditional and conditional batch and perform single forward pass through transformer
- perform individual forward passes for conditional and unconditional batch

The state variables must track values specific to each batch of data over all inference steps, otherwise you might end up in a situation where the state variable for conditional batch is used for unconditional batch, or vice versa.

src/diffusers/pipelines/flux/pipeline_flux.py
```diff
@@ -917,6 +917,7 @@ def __call__(
                 # broadcast to batch dimension in a way that's compatible with ONNX/Core ML
                 timestep = t.expand(latents.shape[0]).to(latents.dtype)
 
+                cc.mark_state("cond")
```
a-r-r-o-w, line 920: Doing it this way helps us distinguish different batches of data. I believe this design will fit well with the upcoming "guiders" to support multiple guidance methods. As guidance methods may use 1 (guidance-distilled) or 2 (CFG) or 3 (PAG) or more latent batches, we can call "mark state" any number of times to distinguish between calls to `transformer.forward` with different batches of data for the same inference step.

tests/pipelines/test_pipelines_common.py
```diff
@@ -2715,6 +2716,55 @@ def faster_cache_state_check_callback(pipe, i, t, kwargs):
                 self.assertTrue(state.cache is None, "Cache should be reset to None.")
 
 
+# TODO(aryan, dhruv): the cache tester mixins should probably be rewritten so that more models can be tested out
+# of the box once there is better cache support/implementation
+class FirstBlockCacheTesterMixin:
```
a-r-r-o-w, line 2721: Adding new cache tester mixins each time and extending the list of parent mixins for each pipeline test is probably not going to be a clean way of testing. We can refactor this in the future and consolidate all cache methods into a single tester once they are better supported/implemented for most models

docs/source/en/api/cache.md
```diff
@@ -11,6 +11,50 @@ specific language governing permissions and limitations under the License. -->
...
+
+```python
+import torch
+from diffusers import CogVideoXPipeline, FirstBlockCacheConfig
+
+pipe = CogVideoXPipeline.from_pretrained("THUDM/CogVideoX-5b", torch_dtype=torch.bfloat16)
+pipe.to("cuda")
+
+# Increasing the threshold may lead to faster inference speeds, but may also lead to poorer quality of generated videos.
+# Smaller values between 0.02-2.0 are recommended based on the model being used. The default value is 0.05.
```
a-r-r-o-w, line 53:
```suggestion
# Smaller values between 0.02-0.20 are recommended based on the model being used. The default value is 0.05.
```

src/diffusers/hooks/hooks.py
```diff
@@ -18,11 +18,76 @@
...
+        self._mark_name = None
+        self._state_cache = {}
+
+    def get_current_state(self) -> "BaseMarkedState":
+        if self._mark_name is None:
+            # If no mark name is set, simply return a dummy object since we're not going to be using it
+            return self
+        if self._mark_name not in self._state_cache.keys():
+            self._state_cache[self._mark_name] = self.__class__(*self._init_args, **self._init_kwargs)
+        return self._state_cache[self._mark_name]
```
DN6, line 49: I think it might make it more clear if this is placed under a method that makes its function a bit more explicit. e.g. `_create_state_instance_for_mark`

src/diffusers/hooks/hooks.py
```diff
@@ -18,11 +18,76 @@
...
 
 
+class BaseState:
+    def reset(self, *args, **kwargs) -> None:
+        raise NotImplementedError(
+            "BaseState::reset is not implemented. Please implement this method in the derived class."
+        )
+
+
+class BaseMarkedState(BaseState):
```
DN6, line 34: Something to consider. Should we call this `ContextAwareState`? And rename the rest of the terms in a similar way?

Might just be my personal preference. 

```python
_mark_name -> _current_context
def _mark_state -> _set_context()

```

src/diffusers/hooks/hooks.py
```diff
@@ -18,11 +18,76 @@
...
+
+    def __getattribute__(self, name):
+        if name in (
+            "get_current_state",
+            "mark_state",
+            "reset",
+            "_init_args",
+            "_init_kwargs",
+            "_mark_name",
+            "_state_cache",
```
DN6, line 68: Nit. 
```suggestion
        direct_attrs = (
            "get_current_state",
            "mark_state",
            "reset",
            "_init_args",
            "_init_kwargs",
            "_mark_name",
            "_state_cache",)
        if name in direct_attrs or _is_dunder_method(name):
```

src/diffusers/hooks/_helpers.py
```diff
@@ -0,0 +1,199 @@
...
+        metadata=TransformerBlockMetadata(
+            skip_block_output_fn=_skip_block_output_fn_WanTransformerBlock,
+            return_hidden_states_index=0,
+            return_encoder_hidden_states_index=None,
+        ),
+    )
+
+
+# fmt: off
+def _skip_block_output_fn___hidden_states_0___ret___hidden_states(self, *args, **kwargs):
```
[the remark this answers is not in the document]
a-r-r-o-w: Not really. It just spells at what argument index hidden_states is and what it returns. Do you have any particular recommendation?

src/diffusers/models/cache_utils.py
```diff
@@ -106,3 +116,21 @@ def _reset_stateful_cache(self, recurse: bool = True) -> None:
...
+class _CacheContextManager:
+    def __init__(self, model: CacheMixin):
+        self.model = model
+
+    def mark_state(self, name: str) -> None:
+        from ..hooks import HookRegistry
+
+        if self.model.is_cache_enabled:
+            registry = HookRegistry.check_if_exists_or_initialize(self.model)
+            registry._mark_state(name)
```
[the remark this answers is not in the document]
a-r-r-o-w: We're not yet anticipating any other methods yet (probably `enable` and `disable` for on-the-fly modifications for front-end use cases eventually).

Changes sounds good!

src/diffusers/hooks/_helpers.py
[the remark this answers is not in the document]
a-r-r-o-w: `skip_block_output_fn` is not confined to just FBC. You can take a look at other use cases in support different guiders: #11205 and #11311. The registries maintain information about what each of the following return:
- attention processors
- transformer blocks
- (eventually, transformer themselves when required)

I'm not sure what's the best approach yet too since we're implementing things as they come. There will probably be a refactor needed eventually. How about we keep the centralized registries for now (until guiders and FBC is merged) and then I can look into better refactor? This will help me work on the other PRs that are using the same changes without first refactoring both PRs

src/diffusers/hooks/hooks.py
a-r-r-o-w: Sounds good!

src/diffusers/hooks/hooks.py
a-r-r-o-w: Yes, this is much better, thanks I'll rename

a-r-r-o-w:
@DN6 Addressed the review comments. Could you give it another look?

nitinmukesh:
Please could this feature also be implemented for LTX-Video, LTXConditionPipeline (0.9.5 and above), if not already covered.

a-r-r-o-w:
@nitinmukesh It's already covered I believe. But I didn't test very rigorously, so it would be super helpful if you wanted to give it a try :hugs:

a-r-r-o-w:
Gentle ping @DN6

src/diffusers/hooks/_helpers.py
```diff
@@ -0,0 +1,271 @@
...
+        metadata=TransformerBlockMetadata(
+            skip_block_output_fn=_skip_block_output_fn_WanTransformerBlock,
+            return_hidden_states_index=0,
+            return_encoder_hidden_states_index=None,
+        ),
+    )
+
+
+# fmt: off
+def _skip_attention___ret___hidden_states(self, *args, **kwargs):
```
DN6, line 207: These functions look like they're only used internally in the hook no? 

Can't we define a function that extracts the hidden_states and encoder_hidden_states and returns them as a dictionary? Then inside the hook, we construct the return tuple based on the return_hidden_index metadata?  So then for encoder_hidden_states/hidden_state return type you just have a single function?

src/diffusers/hooks/_helpers.py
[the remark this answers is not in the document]
DN6: Would prefer we add metadata to the block in the same module where the block is defined. Not a merge blocker, but I'm not 100% sold on using a registry.

src/diffusers/hooks/_helpers.py
```diff
@@ -0,0 +1,271 @@
...
+        hidden_states = args[0]
+    return hidden_states
+
+
+def _skip_block_output_fn___hidden_states_0___encoder_hidden_states_1___ret___hidden_states___encoder_hidden_states(self, *args, **kwargs):
+    hidden_states = kwargs.get("hidden_states", None)
+    encoder_hidden_states = kwargs.get("encoder_hidden_states", None)
+    if hidden_states is None and len(args) > 0:
+        hidden_states = args[0]
+    if encoder_hidden_states is None and len(args) > 1:
```
DN6, line 240: Just a thought, could we refactor all the blocks to always use kwargs? And just enforce that? Would take a lot of guess work out of building future features like this?

src/diffusers/hooks/hooks.py
```diff
@@ -18,11 +18,72 @@
...
 
 
+class BaseState:
+    def reset(self, *args, **kwargs) -> None:
+        raise NotImplementedError(
+            "BaseState::reset is not implemented. Please implement this method in the derived class."
+        )
+
+
+class ContextAwareState(BaseState):
```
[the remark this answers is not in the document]
a-r-r-o-w: SGTM

src/diffusers/hooks/_helpers.py
a-r-r-o-w: It would be no problem for us to enforce passing kwargs. But, for outside implementations built on diffusers, if we want OOTB compatibility and make it easy to work with custom implementations (for example, ComfyUI-maintained original modeling implementations, custom research repo wanting to use some cache implementation for demo, ...), we should support args-index based identification. So, in the metadata class for transformer & intermediate blocks, I would say we should maintain this info

a-r-r-o-w: Done

src/diffusers/hooks/_helpers.py
a-r-r-o-w: Done

DN6 (approved):
Nicely done! 👍🏽

a-r-r-o-w:
Failing tests should hopefully be fixed now. Caused due to a divergence in behaviour from the refactor where, previously, a cache context was not really necessary as the default state object would have been used, which is no longer the case. The current implementation is the correct behaviour

src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -100,10 +111,17 @@ def forward(
         if hidden_states.dtype == torch.float16:
             hidden_states = hidden_states.clip(-65504, 65504)
 
-        return hidden_states
+        encoder_hidden_states, hidden_states = hidden_states[:, :text_seq_len], hidden_states[:, text_seq_len:]
+        return encoder_hidden_states, hidden_states
 
 
 @maybe_allow_in_graph
+@TransformerBlockRegistry.register(
```
yiyixuxu, line 119: ohhh cannt we just make a `xxTransformerBlockOutput`? could be a named tuple or something else, depends on what type of info you need now and could need in future

yiyixuxu (commented):
thanks @a-r-r-o-w 
I left some comments, I just want to brainstorm if we can find a more flexible and extensible ways to make transformers work better with these techniques

src/diffusers/models/transformers/transformer_flux.py
a-r-r-o-w: Reason for not changing what is returned is that it might be a big breaking change for people that import and use the blocks. Also, in general as I've been implementing new methods, there comes requirements that I cannot anticipate beforehand - being able to add some metadata information quickly would be much more convenient compared to committing to a dedicated approach imo. Once we support a large number of techniques, we could look into refactoring/better design, wdyt?

We can refactor this a bit for now though. Dhruv suggested replacing the centralized registry with attributes instead. So, similar to how we have `_no_split_modules`, etc. at ModelMixin level, we can maintain properties at the blocks too

glide-the:
This is a great branch, but I encountered this problem when testing it. The first method will fail, perhaps because some logic is not implemented.

model: CogVIew4-6B
image

Detailed Error
```

apply_first_block_cache(pipe.transformer, FirstBlockCacheConfig(threshold=0.3))

prompt = "A photo of an astronaut riding a horse on mars"
image = pipe(prompt, generator=torch.Generator().manual_seed(42)).images[0]
image.save("output.png")

  0%|                                                                                                                            | 0/50 [00:00<?, ?it/s]
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[12], line 4
      1 apply_first_block_cache(pipe.transformer, FirstBlockCacheConfig(threshold=0.3))
      3 prompt = "A photo of an astronaut riding a horse on mars"
----> 4 image = pipe(prompt, generator=torch.Generator().manual_seed(42)).images[0]
      5 image.save("output.png")

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/torch/utils/_contextlib.py:116, in context_decorator.<locals>.decorate_context(*args, **kwargs)
    113 @functools.wraps(func)
    114 def decorate_context(*args, **kwargs):
    115     with ctx_factory():
--> 116         return func(*args, **kwargs)

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/diffusers/pipelines/cogview4/pipeline_cogview4.py:623, in CogView4Pipeline.__call__(self, prompt, negative_prompt, height, width, num_inference_steps, timesteps, sigmas, guidance_scale, num_images_per_prompt, generator, latents, prompt_embeds, negative_prompt_embeds, original_size, crops_coords_top_left, output_type, return_dict, attention_kwargs, callback_on_step_end, callback_on_step_end_tensor_inputs, max_sequence_length)
    620 timestep = t.expand(latents.shape[0])
    622 with self.transformer.cache_context("cond"):
--> 623     noise_pred_cond = self.transformer(
    624         hidden_states=latent_model_input,
    625         encoder_hidden_states=prompt_embeds,
    626         timestep=timestep,
    627         original_size=original_size,
    628         target_size=target_size,
    629         crop_coords=crops_coords_top_left,
    630         attention_kwargs=attention_kwargs,
    631         return_dict=False,
    632     )[0]
    634 # perform guidance
    635 if self.do_classifier_free_guidance:

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/torch/nn/modules/module.py:1553, in Module._wrapped_call_impl(self, *args, **kwargs)
   1551     return self._compiled_call_impl(*args, **kwargs)  # type: ignore[misc]
   1552 else:
-> 1553     return self._call_impl(*args, **kwargs)

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/torch/nn/modules/module.py:1562, in Module._call_impl(self, *args, **kwargs)
   1557 # If we don't have any hooks, we want to skip the rest of the logic in
   1558 # this function, and just call forward.
   1559 if not (self._backward_hooks or self._backward_pre_hooks or self._forward_hooks or self._forward_pre_hooks
   1560         or _global_backward_pre_hooks or _global_backward_hooks
   1561         or _global_forward_hooks or _global_forward_pre_hooks):
-> 1562     return forward_call(*args, **kwargs)
   1564 try:
   1565     result = None

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/diffusers/models/transformers/transformer_cogview4.py:740, in CogView4Transformer2DModel.forward(self, hidden_states, encoder_hidden_states, timestep, original_size, target_size, crop_coords, attention_kwargs, return_dict, attention_mask, image_rotary_emb)
    730         hidden_states, encoder_hidden_states = self._gradient_checkpointing_func(
    731             block,
    732             hidden_states,
   (...)
    737             attention_kwargs,
    738         )
    739     else:
--> 740         hidden_states, encoder_hidden_states = block(
    741             hidden_states,
    742             encoder_hidden_states,
    743             temb,
    744             image_rotary_emb,
    745             attention_mask,
    746             attention_kwargs,
    747         )
    749 # 4. Output norm & projection
    750 hidden_states = self.norm_out(hidden_states, temb)

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/torch/nn/modules/module.py:1553, in Module._wrapped_call_impl(self, *args, **kwargs)
   1551     return self._compiled_call_impl(*args, **kwargs)  # type: ignore[misc]
   1552 else:
-> 1553     return self._call_impl(*args, **kwargs)

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/torch/nn/modules/module.py:1562, in Module._call_impl(self, *args, **kwargs)
   1557 # If we don't have any hooks, we want to skip the rest of the logic in
   1558 # this function, and just call forward.
   1559 if not (self._backward_hooks or self._backward_pre_hooks or self._forward_hooks or self._forward_pre_hooks
   1560         or _global_backward_pre_hooks or _global_backward_hooks
   1561         or _global_forward_hooks or _global_forward_pre_hooks):
-> 1562     return forward_call(*args, **kwargs)
   1564 try:
   1565     result = None

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/diffusers/hooks/hooks.py:189, in HookRegistry.register_hook.<locals>.create_new_forward.<locals>.new_forward(module, *args, **kwargs)
    187 def new_forward(module, *args, **kwargs):
    188     args, kwargs = function_reference.pre_forward(module, *args, **kwargs)
--> 189     output = function_reference.forward(*args, **kwargs)
    190     return function_reference.post_forward(module, output)

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/diffusers/hooks/first_block_cache.py:89, in FBCHeadBlockHook.new_forward(self, module, *args, **kwargs)
     86 else:
     87     hidden_states_residual = output - original_hidden_states
---> 89 shared_state: FBCSharedBlockState = self.state_manager.get_state()
     90 hidden_states = encoder_hidden_states = None
     91 should_compute = self._should_compute_remaining_blocks(hidden_states_residual)

File /mnt/ceph/develop/jiawei/conda_env/diffuser/lib/python3.10/site-packages/diffusers/hooks/hooks.py:44, in StateManager.get_state(self)
     42 def get_state(self):
     43     if self._current_context is None:
---> 44         raise ValueError("No context is set. Please set a context before retrieving the state.")
     45     if self._current_context not in self._state_cache.keys():
     46         self._state_cache[self._current_context] = self._state_cls(*self._init_args, **self._init_kwargs)

ValueError: No context is set. Please set a context before retrieving the state.
```

a-r-r-o-w:
@glide-the Thanks for testing and reporting the issue!

After the latest refactor, it seems that applying cache with `apply_first_block_cache` was broken but worked as expected with `pipe.transformer.enable_cache()`. This has now been fixed and both approaches of enabling cache should be usable

src/diffusers/models/transformers/transformer_flux.py
a-r-r-o-w: Okay I'll update.

How do you suggest we return the output from the cache hooks? If xxTransformerBlockOutput means maintaining an output class per transformer block type, the cache hook should also return the same output class. Do we instead register what the output class should be to the block metadata so that the cache hook can return the correct object? Or do we use a generic TransformerBlockOutput class for all models?

yiyixuxu: I just looked at the code so my understanding could be completely wrong, let me know if that's the case

I think you would not need to register anything, no? if you use this new output you will get the new output object in your hook and you will know which class it is and also all the other info you need (e.g which value is which) - would this not be the case? 
```python  

def new_forward(self, module: torch.nn.Module, *args, **kwargs):
    original_hidden_states = self._metadata._get_parameter_from_args_kwargs("hidden_states", args, kwargs)

    output = self.fn_ref.original_forward(*args, **kwargs)
    
    encoder_hidden_states = fn_to_update(output.encoder_hidden_states)
    
    hidden_states = fn_to_update(output.hidden_states)

    new_output = output.__class__(
        encoder_hidden_states=encoder_hidden_states, 
        hidden_states=hidden_states
    )
    
    return new_output
```

a-r-r-o-w: @yiyixuxu Thanks for the suggestions. Based on it, I've opened https://github.com/huggingface/diffusers/pull/11594. In that branch, only Wan works at the moment since the changes need to be propagated to all models.

- There are still branches of code that don't have access to the output object/class (the branch that decides when calling the original forward method should be skipped), but due to how the cache mechanism works, we can get access to the output class on a previous inference step and save it as an attribute of the hook.
- For accessing original hidden states and encoder hidden states at each block, we use the `inspect` module. The current code requires the `@register_transformer_block` call so that the results from inspect can be cached for reuse, otherwise there's a minor penalty to performance. For this reason, I've retained the registration but with an empty object to show that your idea works, but we will have to somehow cache the attributes of `forward` method. This does not need to be done in the hook or metadata, and can instead be done automatically when any model is instantiated by iterating through all submodules and creating a mapping of Layer->Attributes_in_Forward.

If you could try to rewrite with what you have in mind, feel free to push changes.

a-r-r-o-w: Gentle ping @yiyixuxu. This is a very anticipated feature

yiyixuxu: @a-r-r-o-w overall the refactor looks good, still not too clear why we need an empty registry though? do you mean that we current code needs it but we could refactor out in the future?

a-r-r-o-w: @yiyixuxu On second thought, I was thinking: let's go forth with current implementation as-is. These are internal details that we can change at any time because the user will never have to interact with any of it. The reason why I think we should do this is because I want to first work on adding a couple of different techniques like SmoothCache, TaylorSeer, Radial Attention, etc. and see what kind of information we need to maintain in blocks/processors/elsewhere. Will help the iteration speed a lot IMO

WDYT?

yiyixuxu: I really don't like adding a meta data registry about the return element indeex to all the model classes 
maybe you keep a hardcoded map with this info for now so we can keep the changes in the hooks folder and get this merged now?

a-r-r-o-w: Okay thanks, I'll go with the hardcoded map approach that we previously had and make the relevant updates

a-r-r-o-w: @yiyixuxu Updated! Please take another look

yiyixuxu (approved):
thanks @a-r-r-o-w  :)

a-r-r-o-w:
Failing tests seem to be unrelated. The VACE lora tests are very flaky and documentation tests are being investigated by Sayak.

I think we're good to go here. Once we have a couple of the latest techniques, I will look into refactoring our test suite and the internal implementations/registration part since we established we don't want centralized registries. I'll open followups for more models supporting caching in the coming days after testing

src/diffusers/models/transformers/transformer_flux.py
burgalon: Please See https://github.com/huggingface/diffusers/issues/12131
This change is causing slowness. Copilot :

This diff changes the way tensors are processed and returned in the forward pass of the transformer_flux model in the HuggingFace Diffusers library. The main performance issue comes from the repeated use of torch.cat and splitting tensors, which can introduce significant overhead, especially on large tensors or in high-throughput scenarios.

## Key Reasons for Slowness
1. Frequent Concatenation and Splitting
Before: hidden_states was used as a single tensor. Forward methods returned just hidden_states.
After:
At the start of each block, you concatenate encoder_hidden_states and hidden_states:
Python
text_seq_len = encoder_hidden_states.shape[1]
hidden_states = torch.cat([encoder_hidden_states, hidden_states], dim=1)
At the end of each block, you split them back:
Python
encoder_hidden_states, hidden_states = hidden_states[:, :text_seq_len], hidden_states[:, text_seq_len:]
Why it’s slow:
torch.cat allocates new memory and copies both tensors every time it’s called.
Slicing (hidden_states[:, :text_seq_len]) also allocates new memory for the split tensors.
If this is done in every block (as in your loop), the cost compounds quickly.
2. Increased Memory Footprint
Every concat/split operation increases peak memory usage, which can slow down training and inference (especially on GPU with limited bandwidth and memory).
3. Potential for Duplicate Computation
If encoder_hidden_states doesn’t change between blocks, you’re repeatedly concatenating the same tensor, which is wasteful.
Why Was This Change Made?
This pattern of concatenating then splitting often occurs in models that want cross-attention between different sources (e.g., text and image). It can make implementation simpler, but at the cost of speed and memory.

## How to Fix or Mitigate
Avoid repeated concat/split:
Concatenate only once per forward pass if possible, not inside every block.
Pass both tensors separately and only combine when strictly necessary for attention.
Profile with torch.autograd.profiler or nvprof:
Confirm that torch.cat and slicing are taking significant time.
Summary:
This diff causes slowness because every block in your transformer repeatedly concatenates and splits large tensors, leading to high memory usage and computational overhead. Optimize by minimizing these operations and only combining tensors when essential.

PR #11186, opened 2025-04-01
[tests] no hard-coded cuda 

faaany:
## What does this PR do?
As the title suggest, this PR enables the case to run on XPU. test tesult is shown as below:

```bash
PASSED tests/quantization/bnb/test_mixed_int8.py::BnB8bitBasicTests::test_device_and_dtype_assignment
```

cc @hlky


hlky (approved):
Thanks @faaany

PR #11190, opened 2025-04-02
fix autocast

jiqing-feng:
Hi @sayakpaul . To enable other device tests like XPU, we need to change the specific cuda autocast to general autocast to apply different devices. Please review this PR. Thanks!


hlky (approved):
Hi @jiqing-feng. Thanks for the PR!

hlky:
Failing tests are unrelated.

PR #11197, opened 2025-04-02
[tests] HunyuanDiTControlNetPipeline inference precision issue on XPU

faaany:
## What does this PR do?
By changing the torch_dtyoe to torch.float32, the outputs are equal with that on CUDA. So this is a hardware specific numerical precision issue. This PR adds xpu-specific check to make the test pass. 
@hlky


hlky (approved):
Hi @faaany. Thanks for the PR!

PR #11198, opened 2025-04-02
Change KolorsPipeline LoRA Loader to StableDiffusion

BasileLewan:
The KolorsPipeline inherits from `StableDiffusionXLLoraLoaderMixin` which expects the pipeline to have two text encoders, whereas Kolors only uses one. This causes an error when trying to load LoRA weights on the pipeline.

Falling back to `StableDiffusionLoraLoaderMixin` fixes the issue.

(1st contribution here, sorry if I did something wrong)

### DIFF
--- a/src/diffusers/pipelines/kolors/pipeline_kolors.py
+++ b/src/diffusers/pipelines/kolors/pipeline_kolors.py
@@ -19,7 +19,7 @@ from transformers import CLIPImageProcessor, CLIPVisionModelWithProjection
 
 from ...callbacks import MultiPipelineCallbacks, PipelineCallback
 from ...image_processor import PipelineImageInput, VaeImageProcessor
-from ...loaders import IPAdapterMixin, StableDiffusionXLLoraLoaderMixin
+from ...loaders import IPAdapterMixin, StableDiffusionLoraLoaderMixin
 from ...models import AutoencoderKL, ImageProjection, UNet2DConditionModel
 from ...models.attention_processor import AttnProcessor2_0, FusedAttnProcessor2_0, XFormersAttnProcessor
 from ...schedulers import KarrasDiffusionSchedulers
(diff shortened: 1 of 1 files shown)


asomoza:
oh nice, I forgot about that since at the time there weren't any kolors loras to test it. Can you please share the lora and an example of it?

BasileLewan:
I'm not sure to get what you mean, this is the case for any LoRA.
If you want a minimal reproduction, here it is:

```python
import torch
from diffusers import KolorsPipeline
from diffusers.utils import convert_state_dict_to_diffusers
from peft import LoraConfig
from peft.utils import get_peft_model_state_dict

pipe = KolorsPipeline.from_pretrained("Kwai-Kolors/Kolors-diffusers", variant='fp16')
model = pipe.unet
cfg = LoraConfig(target_modules=["to_k", "to_q", "to_v"])

model.add_adapter(cfg)
lora_state_dict = convert_state_dict_to_diffusers(get_peft_model_state_dict(model))
pipe.save_lora_weights("lora_path", lora_state_dict)
pipe.load_lora_weights("lora_path/pytorch_lora_weights.safetensors")

# AttributeError: 'KolorsPipeline' object has no attribute 'text_encoder_2'

```

asomoza:
that's example is nice but what I meant was a real lora trained that people use, the reason why this wasn't noticed by anyone until now except for you is that there wasn't any more people that used a Kolors Lora that has text encoders keys, even as unet loras, there's just a few and actually I never seen anyone using them.

Still doesn't matter, I was just curious, we still need to merge this in case that ever happens.

asomoza:
I don't think there's a need to create a specific lora loader mixin for kolors since there are just a handful of them but this will enable people to use them if they want.

asomoza:
thanks a lot!

BasileLewan:
No problem, thanks for the quick review!

PR #11200, opened 2025-04-02
Fix Single File loading for LTX VAE 

DN6:
# What does this PR do?
Changes to the LTX VAE for 0.9.5 caused single file loading to break for the 0.9.1 version VAE. This is because we didn't have a `timestep_scale_multiplier` attribute in the decoder and removed it from the checkpoint during conversion. Although I think this might be a mistake because the parameter does exist in the 0.9.1 checkpoint and the [VAE config](https://huggingface.co/diffusers/LTX-Video-0.9.1/blob/a248b564f0d86b8537474efe2b9f0e05a63e087c/vae/config.json#L54) seems to imply that it should be used. 

The changes for 0.9.5 added the `timestep_scale_multiplier` attribute but since we were removing it in the conversion function, the model loading fails when trying to load the 0.9.1 checkpoint. This PR just skips removing the key from the checkpoint during conversion.    

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/loaders/single_file_utils.py
+++ b/src/diffusers/loaders/single_file_utils.py
@@ -2406,7 +2406,6 @@ def convert_ltx_vae_checkpoint_to_diffusers(checkpoint, **kwargs):
         "per_channel_statistics.channel": remove_keys_,
         "per_channel_statistics.mean-of-means": remove_keys_,
         "per_channel_statistics.mean-of-stds": remove_keys_,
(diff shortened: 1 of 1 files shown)


a-r-r-o-w (approved):
Thanks! We updated the VAE weights to correctly use timestep_scale_multiplier (which was a learned parameter instead of fixed value 1000.0) when LTX 0.9.5 was merged

PR #11227, opened 2025-04-08
introduce compute arch specific expectations and fix test_sd3_img2img_inference failure

yao-matrix:
@hlky , as you know, different compute arch can lead to different expected numeric for same case.
One example is: `pytest -rA tests -k "StableDiffusion3Img2ImgPipelineSlowTests and test_sd3_img2img_inference" ` will fail on A100 and XPU since current `expected_slice` is designed for T4 (CUDA 7).

By borrowing the design from `transformers` for the same case(as @ivarflakstad introduced in https://github.com/huggingface/transformers/pull/36569), I am proposing to use the same mechanism in diffusers tests.

I copied the implementations from transformers to avoid unnecessary versioning dependency, and let me know if you think directly import from transformers is more preferred from you.

W. this PR, test_sd3_img2img_inference case passed both pass on A100 and XPU 1550; before both failed.

Thx.


yao-matrix:
@faaany , if this approved and merged finally, we can follow same design to handle other numerical expectation related issues later.

src/diffusers/utils/testing_utils.py
```diff
@@ -14,10 +14,11 @@
 import time
 import unittest
 import urllib.parse
+from collections import UserDict
 from contextlib import contextmanager
 from io import BytesIO, StringIO
 from pathlib import Path
-from typing import Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Union
```
hlky, line 21:
```suggestion
from typing import Any, Callable, Dict, List, Optional, Tuple, Union, TYPE_CHECKING
```

src/diffusers/utils/testing_utils.py
```diff
@@ -1275,3 +1287,87 @@ def update_mapping_from_spec(device_fn_dict: Dict[str, Callable], attribute_name
         update_mapping_from_spec(BACKEND_RESET_PEAK_MEMORY_STATS, "RESET_PEAK_MEMORY_STATS_FN")
         update_mapping_from_spec(BACKEND_RESET_MAX_MEMORY_ALLOCATED, "RESET_MAX_MEMORY_ALLOCATED_FN")
         update_mapping_from_spec(BACKEND_MAX_MEMORY_ALLOCATED, "MAX_MEMORY_ALLOCATED_FN")
+
+
+# below codes are copied from https://github.com/huggingface/transformers/blob/main/src/transformers/testing_utils.py#L3090
+
+# Type definition of key used in `Expectations` class.
+DeviceProperties = tuple[Union[str, None], Union[int, None]]
```
hlky, line 1295:
```suggestion
DeviceProperties = Tuple[Union[str, None], Union[int, None]]
```

src/diffusers/utils/testing_utils.py
```diff
@@ -1275,3 +1287,87 @@ def update_mapping_from_spec(device_fn_dict: Dict[str, Callable], attribute_name
...
+
+@functools.cache
```
hlky, line 1297:
```suggestion
@functools.lru_cache
```

src/diffusers/utils/testing_utils.py
```diff
@@ -1275,3 +1287,87 @@ def update_mapping_from_spec(device_fn_dict: Dict[str, Callable], attribute_name
...
+        # To get more info of the architecture meaning and bit allocation, refer to https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/ext/oneapi/experimental/device_architecture.def
+        arch = torch.xpu.get_device_capability()["architecture"]
+        gen_mask = 0x000000FF00000000
+        gen = (arch & gen_mask) >> 32
+        return ("xpu", gen)
+    else:
+        return (torch_device, None)
+
+
+class Expectations(UserDict[DeviceProperties, Any]):
```
hlky, line 1322:
```suggestion
if TYPE_CHECKING:
    DevicePropertiesUserDict = UserDict[DeviceProperties, Any]
else:
    DevicePropertiesUserDict = UserDict

class Expectations(DevicePropertiesUserDict):
```

hlky (approved):
Hi @yao-matrix. Thanks for making this PR, nice improvement. Copying from `transformers` is ok in this case, `diffusers` is still supporting Python 3.8, we will hopefully drop it by 0.34, the suggested changes here are all for that purpose. After those changes are applied and a `make style && make quality`, which I can apply via our bot, this should be good to go.

cc @yiyixuxu for awareness

hlky:
@bot /style

sayakpaul (approved):
Looks really really good! Thanks for solving this longstanding problem.

src/diffusers/utils/testing_utils.py
```diff
@@ -1275,3 +1287,93 @@ def update_mapping_from_spec(device_fn_dict: Dict[str, Callable], attribute_name
         update_mapping_from_spec(BACKEND_RESET_PEAK_MEMORY_STATS, "RESET_PEAK_MEMORY_STATS_FN")
         update_mapping_from_spec(BACKEND_RESET_MAX_MEMORY_ALLOCATED, "RESET_MAX_MEMORY_ALLOCATED_FN")
         update_mapping_from_spec(BACKEND_MAX_MEMORY_ALLOCATED, "MAX_MEMORY_ALLOCATED_FN")
+
+
+# below codes are copied from https://github.com/huggingface/transformers/blob/main/src/transformers/testing_utils.py#L3090
```
hlky, line 1292:
```suggestion
# Modified from https://github.com/huggingface/transformers/blob/cdfb018d0300fef3b07d9220f3efe9c2a9974662/src/transformers/testing_utils.py#L3090
```

src/diffusers/utils/testing_utils.py
```diff
@@ -1275,3 +1287,93 @@ def update_mapping_from_spec(device_fn_dict: Dict[str, Callable], attribute_name
...
+
+# Type definition of key used in `Expectations` class.
+DeviceProperties = Tuple[Union[str, None], Union[int, None]]
+
+
+@functools.lru_cache
+def get_device_properties() -> DeviceProperties:
```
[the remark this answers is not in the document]
hlky: Let's raise it internally with transformers team as it's best to keep this in sync with their code (except Python 3.8 changes)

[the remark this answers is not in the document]
sayakpaul: Fine by me. Cc: @ivarflakstad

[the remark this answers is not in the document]
ivarflakstad: Hey!
Honestly any type that can be used as a key is fine by me.
Originally it was a dataclass with fields, but we wanted to keep it as simple as possible while still solving the problem it's addressing.

Feel free to upstream any improvements you can think of ☺️

PR #11236, opened 2025-04-09
minor update to sana sprint docs.

sayakpaul:
# What does this PR do?

Just a minor update to change the title.

@yiyixuxu FYI.


sayakpaul:
Merging as this is safe to do so.

Issue #11225, opened 2025-04-07
SD3 Controlnet Train Example, run out of memory on validation step


Bex0n:
### Describe the bug

On default settings provided in SD3 controlnet example, with 2 validation images training will error out with out of memory during validation on single A100 80GB.


```
04/07/2025 21:15:15 - INFO - __main__ - ***** Running training *****    
04/07/2025 21:15:15 - INFO - __main__ -   Num examples = 10000                 
04/07/2025 21:15:15 - INFO - __main__ -   Num batches each epoch = 10000       
04/07/2025 21:15:15 - INFO - __main__ -   Num Epochs = 2                                                                                                      
04/07/2025 21:15:15 - INFO - __main__ -   Instantaneous batch size per device = 1                                                                             
04/07/2025 21:15:15 - INFO - __main__ -   Total train batch size (w. parallel, distributed & accumulation) = 4               
04/07/2025 21:15:15 - INFO - __main__ -   Gradient Accumulation steps = 4      
04/07/2025 21:15:15 - INFO - __main__ -   Total optimization steps = 4000                                                                                     
Steps:   0%|          | 5/4000 [00:21<4:38:36,  4.18s/it, loss=0.00669, lr=1e-5]04/07/2025 21:15:36 - INFO - __main__ - Running validation... 
{'controlnet', 'image_encoder', 'feature_extractor'} was not found in config. Values will be initialized to default values.                                   Keyword arguments {'safety_checker': None} are not expected by StableDiffusion3ControlNetPipeline and will be ignored.                                                                                                             Loaded tokenizer as CLIPTokenizer from `tokenizer` subfolder of stabilityai/stable-diffusion-3.5-medium.e components...:   0%|          | 0/9 [00:00<?, ?it/s]                                                                                         {'invert_sigmas', 'base_shift', 'base_image_seq_len', 'use_dynamic_shifting', 'shift_terminal', 'time_shift_type', 'use_exponential_sigmas', 'max_shift', 'use
_karras_sigmas', 'max_image_seq_len', 'use_beta_sigmas'} was not found in config. Values will be initialized to default values.
Loaded scheduler as FlowMatchEulerDiscreteScheduler from `scheduler` subfolder of stabilityai/stable-diffusion-3.5-medium.                
Instantiating AutoencoderKL model under default dtype torch.float32.                                                                                          
All model checkpoint weights were used when initializing AutoencoderKL.                                                                                       
                                                                                                                                                              
[...]

PR #11238, opened 2025-04-09
fix: SD3 ControlNet validation so that it runs on a A100.

sayakpaul:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/issues/11225

### DIFF
--- a/examples/controlnet/train_controlnet_sd3.py
+++ b/examples/controlnet/train_controlnet_sd3.py
@@ -53,6 +53,7 @@ from diffusers.optimization import get_scheduler
 from diffusers.training_utils import compute_density_for_timestep_sampling, compute_loss_weighting_for_sd3, free_memory
 from diffusers.utils import check_min_version, is_wandb_available, make_image_grid
 from diffusers.utils.hub_utils import load_or_create_model_card, populate_model_card
+from diffusers.utils.testing_utils import backend_empty_cache
 from diffusers.utils.torch_utils import is_compiled_module
 
 
@@ -118,7 +119,7 @@ def log_validation(controlnet, args, accelerator, weight_dtype, step, is_final_v
 
     del pipeline
     gc.collect()
-    torch.cuda.empty_cache()
+    backend_empty_cache(accelerator.device.type)
 
     pipeline = StableDiffusion3ControlNetPipeline.from_pretrained(
         args.pretrained_model_name_or_path,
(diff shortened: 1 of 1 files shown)


examples/controlnet/train_controlnet_sd3.py
```diff
@@ -102,18 +104,55 @@ def log_validation(controlnet, args, accelerator, weight_dtype, step, is_final_v
...
+    with torch.no_grad():
+        (
+            prompt_embeds,
+            negative_prompt_embeds,
+            pooled_prompt_embeds,
+            negative_pooled_prompt_embeds,
+        ) = pipeline.encode_prompt(
+            validation_prompts,
+            prompt_2=None,
+            prompt_3=None,
```
hlky, line 116: Not a blocker for this PR but looks like `prompt_2` and `prompt_3` should be made `Optional` in the pipeline.

examples/controlnet/train_controlnet_sd3.py
```diff
@@ -102,18 +104,55 @@ def log_validation(controlnet, args, accelerator, weight_dtype, step, is_final_v
...
+        )
+
+    del pipeline
+    gc.collect()
+    torch.cuda.empty_cache()
```
hlky, line 121: Can we use the backend agnostic empty cache here?

examples/controlnet/train_controlnet_sd3.py
```diff
@@ -102,18 +104,55 @@ def log_validation(controlnet, args, accelerator, weight_dtype, step, is_final_v
...
+        controlnet=controlnet,
+        safety_checker=None,
+        text_encoder=None,
+        text_encoder_2=None,
+        text_encoder_3=None,
+        revision=args.revision,
+        variant=args.variant,
+        torch_dtype=weight_dtype,
+    )
+    pipeline.enable_model_cpu_offload()
```
hlky, line 134: We should also pass `device` (and optionally `gpu_id`) here, the default `device` is `cuda`.

examples/controlnet/train_controlnet_sd3.py
```diff
@@ -655,6 +694,7 @@ def make_train_dataset(args, tokenizer_one, tokenizer_two, tokenizer_three, acce
             dataset = load_dataset(
                 args.train_data_dir,
                 cache_dir=args.cache_dir,
+                trust_remote_code=True,
```
hlky, line 697: Nice, could be an argument but more convenient like this especially as the example dataset requires it. Can be replicated across training scripts.
For future reference we should look at the `num_proc` option which should help speed up processing.

hlky (approved):
Left a few comments. Thanks @sayakpaul

hlky (approved):

Issue #8384, opened 2024-06-03
[Community] Help us fix the LR schedulers when `num_train_epochs` is passed in a distributed training env


sayakpaul:
## Context

Refer to https://github.com/huggingface/diffusers/pull/8312 for the full context. The changes introduced in the PR should be propagated to the following scripts, too: 

- advanced_diffusion_training
    - [x] train_dreambooth_lora_sd15_advanced.py
    - [x] train_dreambooth_lora_sdxl_advanced.py
- consistency_distillation
    - [x] train_lcm_distill_lora_sdxl.py
- controlnet
    - [x] train_controlnet.py
    - [x] train_controlnet_sdxl.py
- custom_diffusion
    - [x] train_custom_diffusion.py
- dreambooth
    - [x] train_dreambooth.py
    - [x] train_dreambooth_lora.py
    - [x] train_dreambooth_lora_sdxl.py
- instruct_pix2pix
    - [x] train_instruct_pix2pix.py
    - [ ] rain_instruct_pix2pix_sdxl.py
- kandinsky2_2/text_to_image
    - [ ] train_text_to_image_decoder.py
    - [ ] train_text_to_image_prior.py
    - [ ] train_text_to_image_lora_decoder.py
    - [ ] train_text_to_image_lora_prior.py

- t2i_adapter
    - [ ] train_t2i_adapter_sdxl.py
- text_to_image
    - [x] train_text_to_image.py
    - [ ] train_text_to_image_sdxl.py
    - [x] train_text_to_image_lora.py
    - [ ] train_text_to_image_lora_sdxl.py
- textual_inversion
    - [ ] textual_inversion.py
    - [ ] textual_inversion_sdxl.py
- unconditional_image_generation
    - [ ] train_unconditional.py
- wuerstchen
    - [ ] text_to_image/train_text_to_image_prior.py
    - [ ] text_to_image/train_text_to_image_lora_prior.py
- research_projects (low-priority) 
    - [ ] consistency_training/train_cm_ct_unconditional.py
    - [ ] diffusion_dpo/train_diffusion_dpo.py
    - [ ] diffusion_dpo/train_diffusion_dpo_sdxl.py
    - [ ] diffusion_orpo/train_diffusion_orpo_sdxl_lora.py
    - [ ] dreambooth_inpaint/train_dreambooth_inpaint.py
    - [ ] dreambooth_inpaint/train_dreambooth_inpaint_lora.py
[...]

PR #11242, opened 2025-04-09
[Flux LoRAs] fix lr scheduler bug in distributed scenarios

linoytsaban:
address #8384 for flux scripts


sayakpaul (approved):
Thanks!

sayakpaul:
@bot /style

sayakpaul:
For future merges, we should trigger the CI ourselves (by pushing an empty commit) after the style-bot has pushed commits as that doesn't currently trigger the CI.

PR #11245, opened 2025-04-09
fix FluxReduxSlowTests::test_flux_redux_inference case failure on XPU

yao-matrix:
`pytest -rA tests/pipelines/flux/test_pipeline_flux_redux.py::FluxReduxSlowTests::test_flux_redux_inference` fail on XPU.

After comparing the output btw XPU and A100, we can confirm the image is similar perceptually.

XPU:
XPU

A100:
A100

So, I changed the XPU expectation to make it pass.

I also update the model per legacy comments.

test status:

A100: before this PR **pass**, after this PR **pass**
XPU: before this PR **fail**, after this PR **pass**

### DIFF
--- a/tests/pipelines/test_pipelines_common.py
+++ b/tests/pipelines/test_pipelines_common.py
@@ -1347,7 +1347,7 @@ class PipelineTesterMixin:
 
     @unittest.skipIf(torch_device not in ["cuda", "xpu"], reason="float16 requires CUDA or XPU")
     @require_accelerator
-    def test_float16_inference(self, expected_max_diff=5e-2):
+    def test_float16_inference(self, expected_max_diff=6e-2):
         components = self.get_dummy_components()
         pipe = self.pipeline_class(**components)
         for component in pipe.components.values():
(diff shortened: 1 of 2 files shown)


yao-matrix:
@hlky, pls help review, thx.

tests/pipelines/test_pipelines_common.py
```diff
@@ -1347,7 +1347,7 @@ def test_components_function(self):
 
     @unittest.skipIf(torch_device not in ["cuda", "xpu"], reason="float16 requires CUDA or XPU")
     @require_accelerator
-    def test_float16_inference(self, expected_max_diff=5e-2):
+    def test_float16_inference(self, expected_max_diff=6e-2):
```
hlky, line 1350: `6e-2` vs `5e-2` is ok, what's the context of this change though?

hlky (approved):
Thanks @yao-matrix. Left a comment but good to go when tests pass.

hlky:
Failing tests are unrelated.

tests/pipelines/test_pipelines_common.py
yao-matrix: sorry, it's an un-deleted experimental change, will change back in next PR. Sorry for unnecessary confusion

PR #11246, opened 2025-04-09
Use `ignore_cleanup_errors=True` in `TemporaryDirectory` for Windows compatibility

hlky:
# What does this PR do?

Allows tests using `tempfile.TemporaryDirectory` to run on Windows, the parameter was introduced in 3.10 so we include our own `TemporaryDirectory` based on the 3.10 version.

## Who can review?

### DIFF
--- a/src/diffusers/utils/testing_utils.py
+++ b/src/diffusers/utils/testing_utils.py
@@ -1388,15 +1388,13 @@ GenericAlias = List[int]
 
 
 class TemporaryDirectory:
-    """Create and return a temporary directory.  This has the same
-    behavior as mkdtemp but can be used as a context manager.  For
-    example:
+    """Create and return a temporary directory. This has the same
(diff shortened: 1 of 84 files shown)


hlky:
@bot /style

src/diffusers/utils/testing_utils.py
```diff
@@ -1377,3 +1380,82 @@ def find_expectation(self, key: DeviceProperties = (None, None)) -> Any:
...
     def __repr__(self):
         return f"{self.data}"
+
+
+# This is for Python 3.8/3.9 support of ignore_cleanup_errors
+
+GenericAlias = List[int]
+
+
+class TemporaryDirectory:
```
hlky, line 1390: @sayakpaul See here. This is the 3.10 version of `TemporaryDirectory` with a few changes for 3.8 support. We are using this over `tempfile.TemporaryDirectory`

PR #11253, opened 2025-04-09
Opencv dummy objects

hlky:
# What does this PR do?

https://github.com/huggingface/diffusers/actions/runs/14357087053/job/40248981968#step:7:252

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #11254, opened 2025-04-09
fix consisid imports

sayakpaul:
Maybe supersedes https://github.com/huggingface/diffusers/pull/11253


src/diffusers/dependency_versions_table.py
```diff
@@ -49,4 +49,5 @@
     "urllib3": "urllib3<=2.0.0",
     "black": "black",
     "phonemizer": "phonemizer",
+    "opencv-python": "opencv-python",
```
hlky, line 52:
```suggestion
    "cv2": "opencv-python",
```

tests/others/test_dependencies.py
```diff
@@ -37,6 +37,8 @@ def test_backend_registration(self):
                         backend = "k-diffusion"
                     elif backend == "invisible_watermark":
                         backend = "invisible-watermark"
+                    elif backend == "opencv":
+                        backend = "cv2"
```
hlky, line 41:
```suggestion
                        backend = "opencv-python"
```

hlky (approved):

PR #11263, opened 2025-04-09
fix test_vanilla_funetuning failure on XPU and A100

yao-matrix:
make `pytest -rA tests/lora/test_lora_layers_sd.py::LoraIntegrationTests::test_vanilla_funetuning` pass on XPU and A100.

before this PR: XPU **fail**, A100 **fail**
after this PR: XPU **pass**, A100 **pass**

I've checked the output of 2 device, since this case's step is 2, so basically they are still noise at this timestep. But still can see perceptually similar.

@hlky , pls help review, thx. and also I changed the `test_float16_inference`'s `expected_max_diff` back to `5e-2`.


hlky (approved):
Thanks @yao-matrix, the additional slice for A100 is appreciated!

PR #11270, opened 2025-04-10
[bitsandbytes] improve dtype mismatch handling for bnb + lora.

sayakpaul:
# What does this PR do?

If we try to do:

```py
from diffusers import DiffusionPipeline, FluxControlPipeline
from PIL import Image
import torch

pipe = FluxControlPipeline.from_pretrained("eramth/flux-4bit", torch_dtype=torch.bfloat16).to("cuda")
pipe.load_lora_weights("black-forest-labs/FLUX.1-Canny-dev-lora")

pipe("a dog", control_image=Image.new(mode="RGB", size=(256, 256)))
```

we will run into

<details>
<summary>Error</summary>

```bash
  File "/fsx/sayak/diffusers/bnb_torch_dtype.py", line 8, in <module>
    pipe("a dog", control_image=Image.new(mode="RGB", size=(256, 256)))
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/fsx/sayak/diffusers/src/diffusers/pipelines/flux/pipeline_flux_control.py", line 835, in __call__
    noise_pred = self.transformer(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
  File "/fsx/sayak/diffusers/src/diffusers/models/transformers/transformer_flux.py", line 445, in forward
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/peft/tuners/lora/layer.py", line 712, in forward
    result = self.base_layer(x, *args, **kwargs)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/linear.py", line 125, in forward
    return F.linear(input, self.weight, self.bias)
RuntimeError: self and mat2 must have the same dtype, but got BFloat16 and Half
```

</details>

This is confusing. This PR handles it by raising an error when it finds that the dtype of the dequatized weight will not match the model dtype. The error is raised after the `load_lora_weights()` call so users won't have to wait till the pipeline is called saving time.

```bash
Model is in torch.bfloat16 dtype while the current module weight will be dequantized to torch.float16 dtype. Please pass torch.float16 as `torch_dtype` in `from_pretrained()`.
```

### DIFF
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -94,6 +94,11 @@ def _maybe_dequantize_weight_for_expanded_lora(model, module):
         weight_on_cpu = True
 
     if is_bnb_4bit_quantized:
+        if module.weight.quant_state.dtype != model.dtype:
+            raise ValueError(
+                f"Model is in {model.dtype} dtype while the current module weight will be dequantized to {module.weight.quant_state.dtype} dtype. "
+                f"Please pass {module.weight.quant_state.dtype} as `torch_dtype` in `from_pretrained()`."
+            )
         module_weight = dequantize_bnb_weight(
             module.weight.cuda() if weight_on_cpu else module.weight,
             state=module.weight.quant_state,
--- a/tests/quantization/bnb/test_4bit.py
+++ b/tests/quantization/bnb/test_4bit.py
@@ -752,6 +752,20 @@ class SlowBnb4BitFluxControlWithLoraTests(Base4bitTests):
         max_diff = numpy_cosine_similarity_distance(expected_slice, out_slice)
         self.assertTrue(max_diff < 1e-3, msg=f"{out_slice=} != {expected_slice=}")
 
+    def test_loading_lora_with_incorrect_dtype_raises_error(self):
+        self.tearDown()
+        model_dtype = torch.bfloat16
+        #  https://huggingface.co/eramth/flux-4bit/blob/main/transformer/config.json#L23
+        actual_dtype = torch.float16
+        self.pipeline_4bit = FluxControlPipeline.from_pretrained("eramth/flux-4bit", torch_dtype=torch.bfloat16)
+        self.pipeline_4bit.enable_model_cpu_offload()
+        with self.assertRaises(ValueError) as err_context:
+            self.pipeline_4bit.load_lora_weights("black-forest-labs/FLUX.1-Canny-dev-lora")
+        assert (
+            f"Model is in {model_dtype} dtype while the current module weight will be dequantized to {actual_dtype} dtype."
+            in str(err_context.exception)
+        )
+
 
 @slow
 class BaseBnb4BitSerializationTests(Base4bitTests):


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -95,6 +95,11 @@ def _maybe_dequantize_weight_for_expanded_lora(model, module):
         weight_on_cpu = True
 
     if is_bnb_4bit_quantized:
+        if module.weight.quant_state.dtype != model.dtype:
+            raise ValueError(
+                f"Model is in {model.dtype} dtype while the current module weight will be dequantized to {module.weight.quant_state.dtype} dtype. "
+                f"Please pass {module.weight.quant_state.dtype} as `torch_dtype` in `from_pretrained()`."
+            )
         module_weight = dequantize_bnb_weight(
```
SunMarc, line 103: since we specified dtype = model.dtype in the `dequantize_bnb_weight`, won't the `module_weights` have the same dtype as model ?

SunMarc (approved):
LGTM ! Just a small question

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Yeah. We don't really have any special treatment to handle LoRA param dtype. Ccing @BenjaminBossan here.

> dequantized module_weight using dtype from model.dtype (so we are not using module.weight.quant_state.dtype actually no ?). model.dtype value comes from torch_dtype.

Well, we use the quant_state:
https://github.com/huggingface/diffusers/blob/ea5a6a8b7cee3a0cd1dea691f4219a562c0ea3bd/src/diffusers/quantizers/bitsandbytes/utils.py#L173

But then we also perform another type-casting:
https://github.com/huggingface/diffusers/blob/ea5a6a8b7cee3a0cd1dea691f4219a562c0ea3bd/src/diffusers/quantizers/bitsandbytes/utils.py#L189

BenjaminBossan: Just to clarify: This is unrelated to the LoRA parameters. Instead, what happens is that a PEFT `LoraLayer` wraps the base layer and calls `self.base_layer(x)`, which should just be the result from the original layer. Due to the change in dtype, we will encounter the dtype mismatch there.

sayakpaul:
@BenjaminBossan @SunMarc instead of the current error raising proposal, this solves the issue:

```diff
diff --git a/src/diffusers/loaders/lora_pipeline.py b/src/diffusers/loaders/lora_pipeline.py
index 2e241bc9f..080559357 100644
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -2367,7 +2367,7 @@ class FluxLoraLoaderMixin(LoraBaseMixin):
                     # TODO: consider if this layer needs to be a quantized layer as well if `is_quantized` is True.
                     with torch.device("meta"):
                         expanded_module = torch.nn.Linear(
-                            in_features, out_features, bias=bias, dtype=module_weight.dtype
+                            in_features, out_features, bias=bias, dtype=transformer.dtype
                         )
                     # Only weights are expanded and biases are not. This is because only the input dimensions
                     # are changed while the output dimensions remain the same. The shape of the weight tensor
```

Does this work for you?

Currently, we keep the expanded module as `nn.Linear` even when the underlying model is quantized (such as the case here). But we eventually plan to move to using the respective quantized linear layer (determined by the quantization backend being used).

https://github.com/huggingface/diffusers/blob/ea5a6a8b7cee3a0cd1dea691f4219a562c0ea3bd/src/diffusers/loaders/lora_pipeline.py#L2369

BenjaminBossan:
> Does this work for you?

If the current situation is just temporary and the proposed change solves the initial issue, then that's fine with me. I wonder if we can always rely on the `dtype` attribute on the model, there can be more than one dtype, right?

sayakpaul:
> I wonder if we can always rely on the dtype attribute on the model, there can be more than one dtype, right?

Eventually, we will resort to the `module_weight.dtype` solution as that is more precise. But this is a special case for now.

BenjaminBossan:
Wait, I'm confused now, I had understood [your comment](https://github.com/huggingface/diffusers/pull/11270#issuecomment-2796153160) to suggest that you would rather do that instead of raising an error. But the PR currently still raises the error. :confounded:

sayakpaul:
My comment suggests the use of `model.dtype` not `module_weight.dtype`. For now, that should be okay for the case we're covering. I definitely want to follow what I suggested in https://github.com/huggingface/diffusers/pull/11270#issuecomment-2796153160 and eventually move to (through a future PR): 

> But we eventually plan to move to using the respective quantized linear layer (determined by the quantization backend being used).

This will include

> we will resort to the module_weight.dtype solution as that is more precise

Is this clearer now?

BenjaminBossan (approved):
Got it now. PR LGTM.

DN6:
Sorry, a bit late here, but my understanding is that the casting issue happens here
https://github.com/huggingface/diffusers/blob/ea5a6a8b7cee3a0cd1dea691f4219a562c0ea3bd/src/diffusers/quantizers/bitsandbytes/utils.py#L173-L177

4bit is always dequantized to FP16 and immediately returned so the dtype casting at the end of `dequantize_bnb_weight` is never called. Adding casting right after this step should fix the issue no? 
https://github.com/huggingface/diffusers/blob/8b7ef9caed1fe06ce9dcf76b8e685d2b7af56a41/src/diffusers/loaders/lora_pipeline.py#L98-L102

sayakpaul:
@DN6 good catch! Before that change let's also ask if the following would be more sensible: 

@matthewdouglas should the following

https://github.com/huggingface/diffusers/blob/ea5a6a8b7cee3a0cd1dea691f4219a562c0ea3bd/src/diffusers/quantizers/bitsandbytes/utils.py#L189C5-L190C44

be included inside

https://github.com/huggingface/diffusers/blob/ea5a6a8b7cee3a0cd1dea691f4219a562c0ea3bd/src/diffusers/quantizers/bitsandbytes/utils.py#L177

matthewdouglas:
@sayakpaul @DN6 Good catch there. Yes, I think that change would make sense!

PR #11274, opened 2025-04-10
[LoRA] improve LoRA fusion tests

sayakpaul:
# What does this PR do?

This PR improves our LoRA fusion tests by:

* Additionally checking the `pipe.num_fused_loras` argument. 
* Correctly determines `num_fused_loras` within `lora_base.py`.
* Adding a test `test_lora_scale_kwargs_match_fusion` that checks LoRA + scale outputs match with LoRA fusion with the same scale. 

@BenjaminBossan if I could pick your brain, the following tests are failing on a GPU but passing on a CPU:

```bash
FAILED tests/lora/test_lora_layers_sdxl.py::StableDiffusionXLLoRATests::test_simple_inference_with_text_denoiser_lora_unfused - AssertionError: False is not true : Fused lora should not change the output
FAILED tests/lora/test_lora_layers_sdxl.py::StableDiffusionXLLoRATests::test_simple_inference_with_text_lora_denoiser_fused_multi - AssertionError: False is not true : Fused lora should not change the output
FAILED tests/lora/test_lora_layers_sdxl.py::StableDiffusionXLLoRATests::test_lora_scale_kwargs_match_fusion_0 - AssertionError: False is not true : Fused lora should not change the output
FAILED tests/lora/test_lora_layers_sdxl.py::StableDiffusionXLLoRATests::test_lora_scale_kwargs_match_fusion_1 - AssertionError: False is not true : Fused lora should not change the output
```

Would you have any pointers?

### DIFF
--- a/src/diffusers/loaders/lora_base.py
+++ b/src/diffusers/loaders/lora_base.py
@@ -607,7 +607,10 @@ class LoraBaseMixin:
                         model, lora_scale=lora_scale, safe_fusing=safe_fusing, adapter_names=adapter_names
                     )
 
-        self.num_fused_loras += 1
+        if adapter_names is None:
+            self.num_fused_loras += 1
+        elif isinstance(adapter_names, list):
+            self.num_fused_loras += len(adapter_names)
 
     def unfuse_lora(self, components: List[str] = [], **kwargs):
         r"""
@@ -663,7 +666,7 @@ class LoraBaseMixin:
                         if isinstance(module, BaseTunerLayer):
                             module.unmerge()
 
-        self.num_fused_loras -= 1
+        self.num_fused_loras = 0
 
     def set_adapters(
         self,
--- a/src/diffusers/loaders/lora_base.py
+++ b/src/diffusers/loaders/lora_base.py
@@ -592,6 +592,7 @@ class LoraBaseMixin:
         if len(components) == 0:
             raise ValueError("`components` cannot be an empty list.")
 
+        merged_adapters = set()
         for fuse_component in components:
             if fuse_component not in self._lora_loadable_modules:
                 raise ValueError(f"{fuse_component} is not found in {self._lora_loadable_modules=}.")
@@ -601,16 +602,19 @@ class LoraBaseMixin:
                 # check if diffusers model
                 if issubclass(model.__class__, ModelMixin):
                     model.fuse_lora(lora_scale, safe_fusing=safe_fusing, adapter_names=adapter_names)
+                    for module in model.modules():
+                        if isinstance(module, BaseTunerLayer):
+                            merged_adapters.update(set(module.merged_adapters))
                 # handle transformers models.
                 if issubclass(model.__class__, PreTrainedModel):
                     fuse_text_encoder_lora(
                         model, lora_scale=lora_scale, safe_fusing=safe_fusing, adapter_names=adapter_names
                     )
+                    for module in model.modules():
+                        if isinstance(module, BaseTunerLayer):
+                            merged_adapters.update(set(module.merged_adapters))
 
-        if adapter_names is None:
-            self.num_fused_loras += 1
-        elif isinstance(adapter_names, list):
-            self.num_fused_loras += len(adapter_names)
+        self.num_fused_loras += len(merged_adapters)
 
     def unfuse_lora(self, components: List[str] = [], **kwargs):
         r"""
--- a/tests/lora/utils.py
+++ b/tests/lora/utils.py
@@ -22,6 +22,7 @@ from itertools import product
 import numpy as np
 import pytest
 import torch
+from parameterized import parameterized
 
 from diffusers import (
     AutoencoderKL,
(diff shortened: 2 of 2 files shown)


sayakpaul:
@BenjaminBossan could you share your `diffusers-cli env`?

Also, feel free to review the rest of PR for the changes in `lora_base.py` and tests

BenjaminBossan (commented):
The PR generally LGTM. I have one comment, but it's probably an edge case.

Regarding the failing tests, I don't see why they would start failing now (the ones that were already there), since the changes shouldn't really affect them. Are you sure they pass without the changes, all else being equal? As mentioned, I think tweaking the tolerances might be a solution (albeit not a nice one).

sayakpaul:
> Are you sure they pass without the changes, all else being equal? As mentioned, I think tweaking the tolerances might be a solution (albeit not a nice one).

@BenjaminBossan I should have been clear. The first two tests fail without the PR changes (the latter two being added in this PR).

src/diffusers/loaders/lora_base.py
```diff
@@ -601,13 +602,19 @@ def fuse_lora(
...
                 if issubclass(model.__class__, PreTrainedModel):
                     fuse_text_encoder_lora(
                         model, lora_scale=lora_scale, safe_fusing=safe_fusing, adapter_names=adapter_names
                     )
+                    for module in model.modules():
+                        if isinstance(module, BaseTunerLayer):
+                            merged_adapters.update(set(module.merged_adapters))
 
-        self.num_fused_loras += 1
+        self.num_fused_loras += len(merged_adapters)
```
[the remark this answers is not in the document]
sayakpaul: Good catch. How are you proposing to populate:

```
self._merged_adapters
```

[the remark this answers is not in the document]
BenjaminBossan: Basically replace `merged_adapters.update(set(module.merged_adapters))` you added in this PR with `self._merged_adapters(set(module.merged_adapters))`

sayakpaul:
@BenjaminBossan I decided to just relax the tolerance for GPU. I requested for a clarification here: https://github.com/huggingface/diffusers/pull/11274#discussion_r2037086189

BenjaminBossan:
> I decided to just relax the tolerance for GPU

Makes sense.

> I requested for a clarification here: [#11274 (comment)](https://github.com/huggingface/diffusers/pull/11274#discussion_r2037086189)

I already replied :thinking:

src/diffusers/loaders/lora_base.py
```diff
@@ -662,8 +666,11 @@ def unfuse_lora(self, components: List[str] = [], **kwargs):
                     for module in model.modules():
                         if isinstance(module, BaseTunerLayer):
                             module.unmerge()
+                            self._merged_adapters.update(set(module.merged_adapters))
```
BenjaminBossan, line 669: For unfusing, should this not remove instead?

src/diffusers/loaders/lora_base.py
```diff
@@ -505,6 +504,12 @@ def _best_guess_weight_name(cls, *args, **kwargs):
         deprecate("_best_guess_weight_name", "0.35.0", deprecation_message)
         return _best_guess_weight_name(*args, **kwargs)
 
+    @property
+    def _merged_adapters(self):
+        if "_merged_adapters" not in self.__dict__:
+            self.__dict__["_merged_adapters"] = set()
+        return self.__dict__["_merged_adapters"]
```
sayakpaul, line 511: Since it's a `_merged_adapters` would have been class attribute otherwise, it would have side-effects. 

So,

```py
pipe = DiffusionPipeline(...)
print(pipe.fused_loras) # set()
pipe.unet.add_adapter(...)
pipe.fuse_lora(...)

pipe = DiffusionPipeline(...)
print(pipe.fused_loras) # {"default"}
```

We see that even for a new instantiation of the pipeline, `fused_loras` aren't automatically reset. This is undeseriable. The other solution would have been to always call `unfuse_lora()` after `fuse_lora()` but I didn't prefer that.

BenjaminBossan: We can it not be an instance attribute set during `__init__`?

sayakpaul: Mixin classes are not supposed to have an `__init__()`.

sayakpaul: Yeah we cannot really do that. Hence I had to resort to the current solution.

BenjaminBossan: I think a cleaner solution would be to re-assign the attribute instead of mutating it. That way, no `__dict__` shenanigans are required and there are no side effects:

```python
In [1]: class Foo:
   ...:     _merged_adapters = set()
   ...:     def fuse_lora(self, adapters):
   ...:         self._merged_adapters = self._merged_adapters | set(adapters)  # <= re-assign
   ...: 

In [2]: foo0 = Foo()

In [3]: foo0.fuse_lora(["foo", "bar"])

In [4]: foo0._merged_adapters
Out[4]: {'bar', 'foo'}

In [5]: foo1 = Foo()

In [6]: foo1._merged_adapters
Out[6]: set()

In [7]: foo1.fuse_lora(["bar", "baz"])

In [8]: foo1._merged_adapters
Out[8]: {'bar', 'baz'}

In [9]: foo0._merged_adapters
Out[9]: {'bar', 'foo'}
```

sayakpaul: Do the recent changes work for you? 

If not, I would appreciate either a direct PR to this branch or suggestions that work in the full context of this PR.

src/diffusers/loaders/lora_base.py
```diff
@@ -661,9 +670,18 @@ def unfuse_lora(self, components: List[str] = [], **kwargs):
                 if issubclass(model.__class__, (ModelMixin, PreTrainedModel)):
                     for module in model.modules():
                         if isinstance(module, BaseTunerLayer):
+                            for adapter in set(module.merged_adapters):
+                                if adapter and adapter in self._merged_adapters:
+                                    self._merged_adapters.remove(adapter)
```
BenjaminBossan, line 675: Same idea as in `fuse_lora` to avoid mutating a class attribute. It's probably not strictly necessary because `unfuse_lora` is never called before `fuse_lora` but IMO it's better to be rigorous here.

```suggestion
                                    self._merged_adapters = self._merged_adapters - {adapter}
```

BenjaminBossan (approved):
Overall LGTM, thanks. I made one small suggestion for a possible edge case, up to you if you want to fix it.

sayakpaul:
@DN6 could you give the changes under `lora_base.py` a look too?

sayakpaul:
@DN6 a gentle ping.

PR #11276, opened 2025-04-10
[Tests] Cleanup lora tests utils

sayakpaul:
# What does this PR do?

Removes repetitive code for checks and uses a function instead. As a result:

```
53 additions and 249 deletions.
```

### DIFF
--- a/tests/lora/utils.py
+++ b/tests/lora/utils.py
@@ -658,12 +658,12 @@ class PeftLoraLoaderMixinTests:
                 "Removing adapters should change the output",
             )
 
-    def test_simple_inference_save_pretrained(self):
+    def test_simple_inference_save_pretrained_with_text_lora(self):
         """
         Tests a simple usecase where users could use saving utilities for LoRA through save_pretrained
         """
         for scheduler_cls in self.scheduler_classes:
-            components, text_lora_config, denoiser_lora_config = self.get_dummy_components(scheduler_cls)
+            components, text_lora_config, _ = self.get_dummy_components(scheduler_cls)
             pipe = self.pipeline_class(**components)
             pipe = pipe.to(torch_device)
             pipe.set_progress_bar_config(disable=None)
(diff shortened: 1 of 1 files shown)


tests/lora/utils.py
```diff
@@ -655,8 +616,11 @@ def test_simple_inference_with_partial_text_lora(self):
             output_no_lora = pipe(**inputs, generator=torch.manual_seed(0))[0]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
-            pipe.text_encoder.add_adapter(text_lora_config)
-            self.assertTrue(check_if_lora_correctly_set(pipe.text_encoder), "Lora not correctly set in text encoder")
+            if "text_encoder" in self.pipeline_class._lora_loadable_modules:
+                pipe.text_encoder.add_adapter(text_lora_config)
+                self.assertTrue(
+                    check_if_lora_correctly_set(pipe.text_encoder), "Lora not correctly set in text encoder"
+                )
```
hlky, line 623: Can this be

```python
pipe, _ = self.check_if_adapters_added_correctly(pipe, text_lora_config, denoiser_lora_config=None)
```

tests/lora/utils.py
```diff
@@ -726,10 +681,11 @@ def test_simple_inference_save_pretrained(self):
...
 
-            self.assertTrue(
-                check_if_lora_correctly_set(pipe_from_pretrained.text_encoder),
-                "Lora not correctly set in text encoder",
-            )
+            if "text_encoder" in self.pipeline_class._lora_loadable_modules:
+                self.assertTrue(
+                    check_if_lora_correctly_set(pipe_from_pretrained.text_encoder),
+                    "Lora not correctly set in text encoder",
+                )
```
hlky, line 688: Same here

hlky (approved):
Left a couple comments, LGTM other than that, thanks @sayakpaul

tests/lora/utils.py
sayakpaul: This one's a bit different as we're not going to be passing any config. So, we can leave it.

hlky (approved):

PR #11281, opened 2025-04-10
[LoRA] add LoRA support to HiDream and fine-tuning script

linoytsaban:
add lora training for Hi Dream Image

- trains transformer layers (text encoder training not supported at the moment)
- MoE training disabled for LoRA fine-tuning (potentially re-visited in the future) 

memory optimizations: 
1. cpu offloading support for vae and text encoders
2. latent caching 
3. vae loaded in mixed precision

example - 
```
--pretrained_model_name_or_path=HiDream-ai/HiDream-I1-Full 
--dataset_name=Norod78/Yarn-art-style 
--output_dir=trained-hidream-lora 
--mixed_precision=bf16 
--instance_prompt="a dog, yarn art style" 
--caption_column=text 
--resolution=512 
--train_batch_size=1 
--gradient_accumulation_steps=1 
--optimizer=prodigy 
--rank=16 
--learning_rate=1.0 
--report_to=wandb 
--lr_scheduler=constant 
--lr_warmup_steps=0 
--max_train_steps=1000 
--validation_epochs=25 
--validation_prompt="yoda, yarn art style" 
--seed=0
```
Screenshot 2025-04-21 at 10 42 52

### DIFF
--- a/examples/dreambooth/README_hidream.md
+++ b/examples/dreambooth/README_hidream.md
@@ -87,7 +87,8 @@ accelerate launch train_dreambooth_lora_hidream.py \
   --train_batch_size=1 \
   --gradient_accumulation_steps=4 \
   --use_8bit_adam \
-  --learning_rate=1e-4 \
+  --rank=16 \
+  --learning_rate=2e-4 \
   --report_to="wandb" \
   --lr_scheduler="constant" \
   --lr_warmup_steps=0 \
--- a/examples/dreambooth/README_hidream.md
+++ b/examples/dreambooth/README_hidream.md
@@ -129,4 +129,5 @@ We provide several options for optimizing memory optimization:
 * `cache_latents`: When enabled, we will pre-compute the latents from the input images with the VAE and remove the VAE from memory once done.
 * `--use_8bit_adam`: When enabled, we will use the 8bit version of AdamW provided by the `bitsandbytes` library.
 * `--instance_prompt` and no `--caption_column`: when only an instance prompt is provided, we will pre-compute the text embeddings and remove the text encoders from memory once done.
+
 Refer to the [official documentation](https://huggingface.co/docs/diffusers/main/en/api/pipelines/) of the `HiDreamImagePipeline` to know more about the model.
--- a/examples/dreambooth/README_hidream.md
+++ b/examples/dreambooth/README_hidream.md
@@ -72,7 +72,8 @@ This will also allow us to push the trained LoRA parameters to the Hugging Face 
 
 Now, we can launch training using:
 > [!NOTE]
-> The following training configuration prioritizes lower memory consumption by using gradient checkpointing, 8-bit Adam, latent caching and no validation. 
+> The following training configuration prioritizes lower memory consumption by using gradient checkpointing, 
+> 8-bit Adam optimizer, latent caching, offloading, no validation. 
 > Additionally, when provided with 'instance_prompt' only and no 'caption_column' (used for custom prompts for each image)
 > text embeddings are pre-computed to save memory.
 
@@ -127,5 +128,5 @@ We provide several options for optimizing memory optimization:
 * `--offload`: When enabled, we will offload the text encoder and VAE to CPU, when they are not used.
 * `cache_latents`: When enabled, we will pre-compute the latents from the input images with the VAE and remove the VAE from memory once done.
 * `--use_8bit_adam`: When enabled, we will use the 8bit version of AdamW provided by the `bitsandbytes` library.
-
+* `--instance_prompt` and no `--caption_column`: when only an instance prompt is provided, we will pre-compute the text embeddings and remove the text encoders from memory once done.
 Refer to the [official documentation](https://huggingface.co/docs/diffusers/main/en/api/pipelines/) of the `HiDreamImagePipeline` to know more about the model.
--- a/examples/dreambooth/README_hidream.md
+++ b/examples/dreambooth/README_hidream.md
@@ -71,6 +71,10 @@ snapshot_download(
 This will also allow us to push the trained LoRA parameters to the Hugging Face Hub platform.
 
 Now, we can launch training using:
+> [!NOTE]
+> The following training configuration prioritizes lower memory consumption by using gradient checkpointing, 8-bit Adam, latent caching and no validation. 
+> Additionally, when provided with 'instance_prompt' only and no 'caption_column' (used for custom prompts for each image)
+> text embeddings are pre-computed to save memory.
 
 ```bash
 export MODEL_NAME="HiDream-ai/HiDream-I1-Dev"
@@ -92,8 +96,9 @@ accelerate launch train_dreambooth_lora_hidream.py \
   --report_to="wandb" \
   --lr_scheduler="constant" \
   --lr_warmup_steps=0 \
-  --max_train_steps=500 \
-  --validation_prompt="A photo of sks dog in a bucket" \
+  --max_train_steps=1000 \
+  --cache_latents \
+  --gradient_checkpointing \
   --validation_epochs=25 \
   --seed="0" \
   --push_to_hub
@@ -115,9 +120,7 @@ To better track our training experiments, we're using the following flags in the
 Additionally, we welcome you to explore the following CLI arguments:
 
 * `--lora_layers`: The transformer modules to apply LoRA training on. Please specify the layers in a comma seperated. E.g. - "to_k,to_q,to_v" will result in lora training of attention layers only.
-* `--system_prompt`: A custom system prompt to provide additional personality to the model.
-* `--max_sequence_length`: Maximum sequence length to use for text embeddings.
-
+* `--rank`: The rank of the LoRA layers. The higher the rank, the more parameters are trained. The default is 16.
 
 We provide several options for optimizing memory optimization:
 
--- a/examples/dreambooth/train_dreambooth_lora_hidream.py
+++ b/examples/dreambooth/train_dreambooth_lora_hidream.py
@@ -1665,7 +1665,7 @@ def main(args):
                     img_ids[..., 1] = img_ids[..., 1] + torch.arange(pH)[:, None]
                     img_ids[..., 2] = img_ids[..., 2] + torch.arange(pW)[None, :]
                     img_ids = img_ids.reshape(pH * pW, -1)
-                    img_ids_pad = torch.zeros(self.transformer.max_seq, 3)
+                    img_ids_pad = torch.zeros(transformer.max_seq, 3)
                     img_ids_pad[: pH * pW, :] = img_ids
 
                     img_sizes = img_sizes.unsqueeze(0).to(model_input.device)
--- a/examples/dreambooth/train_dreambooth_lora_hidream.py
+++ b/examples/dreambooth/train_dreambooth_lora_hidream.py
@@ -1097,7 +1097,10 @@ def main(args):
             "Mixed precision training with bfloat16 is not supported on MPS. Please use fp16 (recommended) or fp32 instead."
         )
 
-    vae.to(dtype=weight_dtype)
+    if not args.offload:
+         vae.to(dtype=weight_dtype, device=accelerator.device)
+    else:
+        vae.to(dtype=weight_dtype)
     transformer.to(accelerator.device, dtype=weight_dtype)
     text_encoder_one.to(accelerator.device, dtype=weight_dtype)
     text_encoder_two.to(accelerator.device, dtype=weight_dtype)
--- a/examples/dreambooth/train_dreambooth_lora_hidream.py
+++ b/examples/dreambooth/train_dreambooth_lora_hidream.py
@@ -1699,15 +1699,14 @@ def main(args):
                 model_pred = (
                     transformer(
                         hidden_states=noisy_model_input,
-                        encoder_hidden_states_t5=prompt_embeds_t5,
-                        encoder_hidden_states_llama3=prompt_embeds_llama3,
+                        encoder_hidden_states_t5=t5_prompt_embeds,
+                        encoder_hidden_states_llama3=llama3_prompt_embeds,
                         pooled_embeds=pooled_prompt_embeds,
                         timesteps=timesteps,
                         return_dict=False,
                     )[0]
                     * -1
                 )
-                # print("model_pred", model_pred.shape)
                 # these weighting schemes use a uniform timestep sampling
                 # and instead post-weight the loss
                 weighting = compute_loss_weighting_for_sd3(weighting_scheme=args.weighting_scheme, sigmas=sigmas)
--- a/examples/dreambooth/train_dreambooth_lora_hidream.py
+++ b/examples/dreambooth/train_dreambooth_lora_hidream.py
@@ -184,7 +184,7 @@ def load_text_encoders(class_one, class_two, class_three):
         args.pretrained_model_name_or_path, subfolder="text_encoder_3", revision=args.revision, variant=args.variant
     )
     text_encoder_four = LlamaForCausalLM.from_pretrained(
-        "meta-llama/Meta-Llama-3.1-8B-Instruct",
+        args.pretrained_text_encoder_4_name_or_path,
         output_hidden_states=True,
         output_attentions=True,
         torch_dtype=torch.bfloat16,
@@ -285,6 +285,18 @@ def parse_args(input_args=None):
         type=str,
         default=None,
         required=True,
+        help="Path to pretrained model or model identifier from huggingface.co/models.",
+    )
+    parser.add_argument(
+        "--pretrained_tokenizer_4_name_or_path",
+        type=str,
+        default="meta-llama/Meta-Llama-3.1-8B-Instruct",
+        help="Path to pretrained model or model identifier from huggingface.co/models.",
+    )
+    parser.add_argument(
+        "--pretrained_text_encoder_4_name_or_path",
+        type=str,
+        default="meta-llama/Meta-Llama-3.1-8B-Instruct",
         help="Path to pretrained model or model identifier from huggingface.co/models.",
     )
     parser.add_argument(
@@ -1036,7 +1048,7 @@ def main(args):
     )
 
     tokenizer_four = AutoTokenizer.from_pretrained(
-        "meta-llama/Meta-Llama-3.1-8B-Instruct",
+        args.pretrained_tokenizer_4_name_or_path,
         revision=args.revision,
     )
     tokenizer_four.pad_token = tokenizer_four.eos_token
@@ -1646,6 +1658,7 @@ def main(args):
         # Final inference
         # Load previous pipeline
         tokenizer_4 = AutoTokenizer.from_pretrained("meta-llama/Meta-Llama-3.1-8B-Instruct")
+        tokenizer_4.pad_token = tokenizer_4.eos_token
         text_encoder_4 = LlamaForCausalLM.from_pretrained(
             "meta-llama/Meta-Llama-3.1-8B-Instruct",
             output_hidden_states=True,
--- a/examples/dreambooth/train_dreambooth_lora_hidream.py
+++ b/examples/dreambooth/train_dreambooth_lora_hidream.py
@@ -1357,7 +1357,7 @@ def main(args):
     vae_config_shift_factor = vae.config.shift_factor
     if args.cache_latents:
         latents_cache = []
-        if not args.offload:
+        if args.offload:
             vae = vae.to(accelerator.device)
         for batch in tqdm(train_dataloader, desc="Caching latents"):
             with torch.no_grad():
--- a/examples/dreambooth/train_dreambooth_lora_hidream.py
+++ b/examples/dreambooth/train_dreambooth_lora_hidream.py
@@ -1643,7 +1643,7 @@ def main(args):
                     img_ids[..., 1] = img_ids[..., 1] + torch.arange(pH)[:, None]
                     img_ids[..., 2] = img_ids[..., 2] + torch.arange(pW)[None, :]
                     img_ids = img_ids.reshape(pH * pW, -1)
-                    img_ids_pad = torch.zeros(self.transformer.max_seq, 3)
+                    img_ids_pad = torch.zeros(transformer.max_seq, 3)
                     img_ids_pad[: pH * pW, :] = img_ids
 
                     img_sizes = img_sizes.unsqueeze(0).to(model_input.device)
--- a/examples/dreambooth/train_dreambooth_lora_hidream.py
+++ b/examples/dreambooth/train_dreambooth_lora_hidream.py
@@ -196,22 +196,35 @@ def log_validation(
     accelerator,
     pipeline_args,
     epoch,
+    torch_dtype,
     is_final_validation=False,
 ):
     logger.info(
         f"Running validation... \n Generating {args.num_validation_images} images with prompt:"
         f" {args.validation_prompt}."
     )
-
-    pipeline = pipeline.to(accelerator.device)
+    pipeline = pipeline.to(accelerator.device, dtype=torch_dtype)
     pipeline.set_progress_bar_config(disable=True)
 
     # run inference
     generator = torch.Generator(device=accelerator.device).manual_seed(args.seed) if args.seed is not None else None
     autocast_ctx = torch.autocast(accelerator.device.type) if not is_final_validation else nullcontext()
 
-    with autocast_ctx:
-        images = [pipeline(**pipeline_args, generator=generator).images[0] for _ in range(args.num_validation_images)]
+    # pre-calculate  prompt embeds, pooled prompt embeds, text ids because t5 does not support autocast
+    with torch.no_grad():
+        prompt_embeds, negative_prompt_embeds, pooled_prompt_embeds, negative_pooled_prompt_embeds= pipeline.encode_prompt(
+            pipeline_args["prompt"], prompt_2=pipeline_args["prompt"], prompt_3=pipeline_args["prompt"], prompt_4=pipeline_args["prompt"]
+        )
+    images = []
+    for _ in range(args.num_validation_images):
+        with autocast_ctx:
+            image = pipeline(
+                prompt_embeds=prompt_embeds,
+                negative_prompt_embeds=negative_prompt_embeds,
+                pooled_prompt_embeds=pooled_prompt_embeds,
+                negative_pooled_prompt_embeds=negative_pooled_prompt_embeds, generator=generator
+            ).images[0]
+            images.append(image)
 
     for tracker in accelerator.trackers:
         phase_name = "test" if is_final_validation else "validation"
@@ -222,7 +235,7 @@ def log_validation(
             tracker.log(
                 {
                     phase_name: [
-                        wandb.Image(image, caption=f"{i}: {pipeline_args['prompt']}") for i, image in enumerate(images)
+                        wandb.Image(image, caption=f"{i}: {args.validation_prompt}") for i, image in enumerate(images)
                     ]
                 }
             )
@@ -1481,7 +1494,7 @@ def main(args):
     if not train_dataset.custom_instance_prompts:
         # delete tokenizers and text ecnoders except for llama (tokenizer & te four)
         # as it's needed for inference with pipeline
-        del text_encoder_one, text_encoder_two, text_encoder_three, tokenizer_one, tokenizer_two,tokenizer_three
+        del text_encoder_one, text_encoder_two, text_encoder_three, tokenizer_one, tokenizer_two, tokenizer_three
         if not args.validation_prompt:
             del tokenizer_four, text_encoder_four
         free_memory()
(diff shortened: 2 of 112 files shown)


examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1829 @@
...
+
+                # Add noise according to flow matching.
+                # zt = (1 - texp) * x + texp * z1
+                sigmas = get_sigmas(timesteps, n_dim=model_input.ndim, dtype=model_input.dtype)
+                noisy_model_input = (1.0 - sigmas) * model_input + sigmas * noise
+
+                # Predict the noise residual
+                model_pred = transformer(
+                    hidden_states=noisy_model_input,
+                    encoder_hidden_states=prompt_embeds.repeat(len(prompts), 1, 1)
```
sayakpaul, line 1649: Let's move this to a variable for easier readability?

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1829 @@
...
+                    img_ids[..., 2] = img_ids[..., 2] + torch.arange(pW)[None, :]
+                    img_ids = img_ids.reshape(pH * pW, -1)
+                    img_ids_pad = torch.zeros(self.transformer.max_seq, 3)
+                    img_ids_pad[: pH * pW, :] = img_ids
+
+                    img_sizes = img_sizes.unsqueeze(0).to(model_input.device)
+                    img_ids = img_ids_pad.unsqueeze(0).to(model_input.device)
+
+                else:
+                    img_sizes = img_ids = None
```
sayakpaul, line 1623: Would add a comment explaining what is happening and move the `img_ids` calculation in a function.

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1829 @@
...
+    tokenizers = [tokenizer_one, tokenizer_two, tokenizer_three, tokenizer_four]
+    text_encoders = [text_encoder_one, text_encoder_two, text_encoder_three, text_encoder_four]
+    def compute_text_embeddings(prompt, text_encoders, tokenizers):
+        with torch.no_grad():
+            prompt_embeds, pooled_prompt_embeds = encode_prompt(
+                text_encoders, tokenizers, prompt, args.max_sequence_length
+            )
+            prompt_embeds = prompt_embeds.to(accelerator.device)
+            pooled_prompt_embeds = pooled_prompt_embeds.to(accelerator.device)
+        return prompt_embeds, pooled_prompt_embeds
```
sayakpaul, line 1441: Hundred percent not a blocker for this PR. But we could allow users to drop out the other text encoders by zeroing those embeds.

sayakpaul (commented):
Thanks a lot for this!

I have left some clarification questions. LMK if they make sense.

examples/dreambooth/train_dreambooth_lora_hidream.py
bghira: it is a single caption used for every image?

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1829 @@
...
+                    img_sizes=img_sizes,
+                    img_ids=img_ids,
+                    return_dict=False,
+                )[0]
+
+                # these weighting schemes use a uniform timestep sampling
+                # and instead post-weight the loss
+                weighting = compute_loss_weighting_for_sd3(weighting_scheme=args.weighting_scheme, sigmas=sigmas)
+
+                # flow matching loss (reversed)
```
bghira, line 1665:
```suggestion
                # flow matching loss (reverse ODE, inverted flow field)
```

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1829 @@
...
+            truncation=True,
+            return_tensors="pt",
+        )
+
+        text_input_ids = text_inputs.input_ids
+    else:
+        if text_input_ids is None:
+            raise ValueError("text_input_ids must be provided when the tokenizer is not specified")
+
+    prompt_embeds = text_encoder(text_input_ids.to(device), output_hidden_states=True)
```
hlky, line 994:
```suggestion
    prompt_embeds = text_encoder(text_input_ids.to(device))
```
Reduced memory requirements by not returning all `hidden_states` as they aren't being used, missed it on integration.

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1829 @@
...
+
+    vae = AutoencoderKL.from_pretrained(
+        args.pretrained_model_name_or_path,
+        subfolder="vae",
+        revision=args.revision,
+        variant=args.variant,
+    )
+    transformer = Lumina2Transformer2DModel.from_pretrained(
+        args.pretrained_model_name_or_path, subfolder="transformer", revision=args.revision, variant=args.variant
+    )
```
hlky, line 1221:
```suggestion
    transformer = HiDreamImageTransformer2DModel.from_pretrained(
        args.pretrained_model_name_or_path, subfolder="transformer", revision=args.revision, variant=args.variant
    )
```

examples/dreambooth/train_dreambooth_lora_hidream.py
hlky: This was moved into the `transformer` so should be ok to remove

https://github.com/huggingface/diffusers/blob/a8f5134c113da402a93580ef7a021557e816c98d/src/diffusers/models/transformers/transformer_hidream_image.py#L743-L756

hlky (commented):
Just a few comments, will review again when out of draft, thanks @linoytsaban!

linoytsaban:
https://github.com/HiDream-ai/HiDream-I1/issues/15#issuecomment-2798694028 noticing this behavior too
mentioning here for documentation

@bghira from you experiments, it seems training still works as expected?

bghira:
nope, its pretty limited in what can be introduced via LoRA (no human likeness) and full tuning doesn't work even w/ DeepSpeed (model collapses presumably because of MoE training) and the only way so far that one person managed likeness is via lycoris-lora library and use_scalar=True for LoKr. overall, HiDream doesn't seem "worthwhile" to train, it's so slow after, and the model degrades.

image

linoytsaban:
@sayakpaul 
bringing our discussion here for future reference - it's probably best to disable by default the aux-moe-loss computation for loras triggered by [`transformer.training`](https://github.com/huggingface/diffusers/blob/7ecfe2916044e4d1f2e5b7d8f2b519d6c4677e77/src/diffusers/models/transformers/transformer_hidream_image.py#L312) (at least as a first step and expand to that later)

bghira:
if you check my implementation (it only matters for FFT and certain LoKr config tho) you can use the updated MoEGate for training, but it's going to be a pain, since it has to be updated in the main modeling impl first.

linoytsaban:
example [run](https://wandb.ai/linoy/dreambooth-hidream-lora/runs/6mhzezb4?nw=nwuserlinoy):
(w/ _moe training disabled_)
```
import os
os.environ['MODEL_NAME'] = "HiDream-ai/HiDream-I1-Full"
os.environ['DATASET_NAME'] ="dog"
os.environ['OUTPUT_DIR'] = "trained-hidream-lora"

!accelerate launch train_dreambooth_lora_hidream.py \
  --pretrained_model_name_or_path=$MODEL_NAME  \
  --instance_data_dir=$DATASET_NAME \
  --output_dir=$OUTPUT_DIR \
  --mixed_precision="bf16" \
  --instance_prompt="a photo of sks dog" \
  --resolution=512 \
  --train_batch_size=1 \
  --gradient_accumulation_steps=1 \
  --optimizer="prodigy"\
  --learning_rate=1.0 \
  --report_to="wandb" \
  --rank=16 \
  --lr_scheduler="constant" \
  --lr_warmup_steps=0 \
  --max_train_steps=2000 \
  --validation_epochs=25 \
  --validation_prompt="A photo of sks dog in a bucket" \
  --seed="0" \
  --push_to_hub
```
Screenshot 2025-04-17 at 11 53 23

ojasaar:
@bghira I achieved promising results for likeness training with SimpleTuner, Lycoris LoRA

Source of truth:
ChatGPT Image Apr 5, 2025 at 02_41_44 PM

Examples:
output_002

output_torch

Since loss was discussed:
Screenshot 2025-04-17 at 14 14 21

linoytsaban:
@ojasaar thanks for sharing. is this with the aux loss?

bghira:
lora algo in lycoris lib or lokr? i think lokr works when set to use scalar but i hadnt seen any tests with lora yet. thank you, good results.

ojasaar:
@linoytsaban Probably @bghira can answer that, I did not dig into the SimpleTuner code yet.
And yes, lora algo in lycoris lib

Configs I used for reference, since it may be helpful

`lycoris_config.json`
```json
{
    "algo": "lora",
    "multiplier": 1.0,
    "linear_dim": 8,
    "linear_alpha": 8,
    "apply_preset": {
        "target_module": [
            "Attention",
            "FeedForward"
        ],
        "module_algo_map": {
            "Attention": {
                "factor": 8
            },
            "FeedForward": {
                "factor": 8
            }
        }
    }
}
```
`config.json`
```json
{
    "--resume_from_checkpoint": "latest",
    "--data_backend_config": "config/multidatabackend.json",
    "--aspect_bucket_rounding": 2,
    "--seed": 42,
    "--minimum_image_size": 0,
    "--disable_benchmark": false,
    "--output_dir": "output/models",
    "--lora_type": "lycoris",
    "--lycoris_config": "config/lycoris_config.json",
    "--max_train_steps": 1000,
    "--num_train_epochs": 0,
    "--checkpointing_steps": 400,
    "--checkpoints_total_limit": 5,
    "--attention_mechanism": "diffusers",
    "--tracker_project_name": "hidream-lora",
    "--tracker_run_name": "simpletuner-lora",
    "--report_to": "wandb",
    "--model_type": "lora",
    "--model_family": "hidream",
    "--model_flavour": "full",
    "--train_batch_size": 1,
    "--gradient_checkpointing": "true",
    "--caption_dropout_probability": 0.05,
    "--mixed_precision": "bf16",
    "--resolution_type": "pixel_area",
    "--resolution": 1024,
    "--validation_seed": 42,
    "--validation_steps": "100",
    "--validation_resolution": "512x512",
    "--validation_guidance": 3.0,
    "--validation_guidance_rescale": "0.0",
    "--validation_num_inference_steps": "30",
    "--validation_noise_scheduler": "unipc",
    "--validation_prompt": "a high quality portrait of asi1",
    "--user_prompt_library": "config/user_prompt_library.json",
    "--optimizer": "optimi-lion",
    "--learning_rate": "4e-5",
    "--use_ema": "true",
    "--lr_scheduler": "constant",
    "--lr_warmup_steps": 100,
    "--validation_torch_compile": "false"
}
```

sayakpaul:
@ojasaar thanks for sharing!

@bghira how are the FeedForwards targeted in `SimpleTuner`?

bghira:
actually for HiDream the `FeedForward` is ignored unless you specifically target the swiglu feedforward module name. the reason `Attention` works is because i named its attention processor that, and its modules are named simply `Attention`.

whether it would target them w/ LoRA in Lycoris even if the module is correctly identified for inclusion i'm not sure. but i purposely left it ignored because there were some bugs when training them that i'd not wanted to really dig into.

@linoytsaban he didn't apply auxiliary loss, from the config shared (the loss values would be **even higher**)

bghira:
it's worth noting that the original LoRA paper authors originally did not find value in FF/MLP training because there was an error in their test methodology. in a later note they added, they do in fact get better results on LoRA with attn+ff instead of just attn.

sayakpaul:
Thanks for the notes!

linoytsaban:
Thanks @bghira!

src/diffusers/models/transformers/transformer_hidream_image.py
```diff
@@ -663,12 +663,24 @@ def __init__(
...
+            # This might require padding/truncating x if training uses fixed max_seq
+            current_S = x.shape[1]
+            if current_S > expected_S:
+                x = x[:, :expected_S, :]  # Use only the relevant part of the sequence
+            elif current_S < expected_S:
+                # This case is less likely if padding happens earlier, but handle defensively
+                raise ValueError(
+                    f"Sequence length {current_S} is less than expected {expected_S} ({pH}x{pW}) during unpatchify.")
+
+            # Corrected line using einops and H, W:
```
sayakpaul, line 678: Just for the record, we cannot be using `einops` here.

bghira: i think it is inherited from the original codebase, easy to overlook after installing einops for testing and then it works everywhere after :see_no_evil:

linoytsaban:
scheduler config was incorrect - loaded with shift==1, fixed to 3.0.  
converges much faster, results look as expected: 
Screenshot 2025-04-17 at 17 45 13

linoytsaban:
@sayakpaul I think of removing `--max_sequence_length` as I'm not sure it makes sense to make it configurable here. WDYT?

vladmandic:
> @sayakpaul I think of removing `--max_sequence_length` as I'm not sure it makes sense to make it configurable here. WDYT?

default max-sequence-length of 128 is pretty bad. i've tested with larger and there are no sideeffects and results are in general much better. imo, absolutely leave it as configurable.

bghira:
for training @vladmandic the results are worse, the transformer models heavily bias seq len unless you are throwing huuuge levels of data and compute at it. I've hidden the seq len modifying options behind `--i_know_what_i_am_doing` because i don't want to field bug reports of "my model look bad" when they try it without this awareness

bghira:
the reason higher sequence lengths work during inference is due to the nature of the positional embeds.

linoytsaban:
let's keep it configurable for now and I'll add a note to the readme

linoytsaban:
@bot /style

linoytsaban:
@bot /style

linoytsaban:
@bot /style

linoytsaban:
@bot /style

linoytsaban:
@bot /style

examples/dreambooth/README_hidream.md
```diff
@@ -0,0 +1,128 @@
...
+  --rank=16 \
+  --learning_rate=2e-4 \
+  --report_to="wandb" \
+  --lr_scheduler="constant" \
+  --lr_warmup_steps=0 \
+  --max_train_steps=500 \
+  --validation_prompt="A photo of sks dog in a bucket" \
+  --validation_epochs=25 \
+  --seed="0" \
+  --push_to_hub
```
sayakpaul, line 99: Let's keep this command relatively less memory-hungry? 

* `cache_latents`
* `gradient_checkpointing`
* No validation
* Precomputation of the text embeddings

Then we can make a note that the example command prioritizes lower memory consumption and the options available to the users. WDYT?

examples/dreambooth/README_hidream.md
```diff
@@ -0,0 +1,128 @@
...
+
+* `report_to="wandb` will ensure the training runs are tracked on [Weights and Biases](https://wandb.ai/site). To use it, be sure to install `wandb` with `pip install wandb`. Don't forget to call `wandb login <your_api_key>` before training if you haven't done it before.
+* `validation_prompt` and `validation_epochs` to allow the script to do a few validation inference runs. This allows us to qualitatively check if the training is progressing as expected.
+
+## Notes
+
+Additionally, we welcome you to explore the following CLI arguments:
+
+* `--lora_layers`: The transformer modules to apply LoRA training on. Please specify the layers in a comma seperated. E.g. - "to_k,to_q,to_v" will result in lora training of attention layers only.
+* `--system_prompt`: A custom system prompt to provide additional personality to the model.
```
sayakpaul, line 118: I don't see an option to specify `system_prompt` in:
https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py

examples/dreambooth/README_hidream.md
```diff
@@ -0,0 +1,128 @@
...
+* `--max_sequence_length`: Maximum sequence length to use for text embeddings.
+
+
+We provide several options for optimizing memory optimization:
+
+* `--offload`: When enabled, we will offload the text encoder and VAE to CPU, when they are not used.
+* `cache_latents`: When enabled, we will pre-compute the latents from the input images with the VAE and remove the VAE from memory once done.
+* `--use_8bit_adam`: When enabled, we will use the 8bit version of AdamW provided by the `bitsandbytes` library.
```
sayakpaul, line 126: We could also clarify to the users that when `instance_prompt` is used and once the text embeddings are computed the memory is freed by removing the text encoders?

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1708 @@
...
+
+
+```
+
+For more details, including weighting, merging and fusing LoRAs, check the [documentation on loading LoRAs in diffusers](https://huggingface.co/docs/diffusers/main/en/using-diffusers/loading_adapters)
+"""
+    model_card = load_or_create_model_card(
+        repo_id_or_path=repo_id,
+        from_training=True,
+        license="apache-2.0",
```
sayakpaul, line 156: Small nit: It's MIT license.

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1708 @@
...
+            negative_prompt_embeds_t5,
+            prompt_embeds_llama3,
+            negative_prompt_embeds_llama3,
+            pooled_prompt_embeds,
+            negative_pooled_prompt_embeds,
+        ) = pipeline.encode_prompt(
+            pipeline_args["prompt"],
+            prompt_2=pipeline_args["prompt"],
+            prompt_3=pipeline_args["prompt"],
+            prompt_4=pipeline_args["prompt"],
```
sayakpaul, line 228: We can just set them to `None` no?

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1708 @@
...
+                {
+                    phase_name: [
+                        wandb.Image(image, caption=f"{i}: {args.validation_prompt}") for i, image in enumerate(images)
+                    ]
+                }
+            )
+
+    del pipeline
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
```
sayakpaul, line 260: Have been seeing in my tests that doing just this isn't apparently enough. Just making a note that manually placing the components we don't need on CPU solves the problem at least in my tests. 

So, something like:

```py
for name, component in pipeline.components.items():
     if "text_encoder" in name:
         component.cpu()
         del component
```

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1708 @@
...
+                        timesteps=timesteps,
+                        return_dict=False,
+                    )[0]
+                    * -1
+                )
+                # these weighting schemes use a uniform timestep sampling
+                # and instead post-weight the loss
+                weighting = compute_loss_weighting_for_sd3(weighting_scheme=args.weighting_scheme, sigmas=sigmas)
+
+                # flow matching loss via reverse ODE (HiDream uses an inverted flow field as compared to the more typical `noise - model_input` target)
```
sayakpaul, line 1532: It's exactly `noise - model_input` in the code, so maybe remove / rephrase?

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1708 @@
...
+                    transformer(
+                        hidden_states=noisy_model_input,
+                        encoder_hidden_states_t5=t5_prompt_embeds,
+                        encoder_hidden_states_llama3=llama3_prompt_embeds,
+                        pooled_embeds=pooled_prompt_embeds,
+                        timesteps=timesteps,
+                        return_dict=False,
+                    )[0]
+                    * -1
+                )
```
sayakpaul, line 1527: Maybe with a comment as well?

```suggestion
                # Predict the noise residual
                model_pred = transformer(
                    hidden_states=noisy_model_input,
                    encoder_hidden_states_t5=t5_prompt_embeds,
                    encoder_hidden_states_llama3=llama3_prompt_embeds,
                    pooled_embeds=pooled_prompt_embeds,
                    timesteps=timesteps,
                    return_dict=False,
                )[0]
                model_pred = model_pred * -1
```

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1708 @@
...
+                    )
+                else:
+                    t5_prompt_embeds = t5_prompt_embeds.repeat(len(prompts), 1, 1)
+                    llama3_prompt_embeds = llama3_prompt_embeds.repeat(1, len(prompts), 1, 1)
+                    pooled_prompt_embeds = pooled_prompt_embeds.repeat(len(prompts), 1)
+                # Convert images to latent space
+                if args.cache_latents:
+                    model_input = latents_cache[step].sample()
+                else:
+                    vae = vae.to(accelerator.device)
```
sayakpaul, line 1487: Otherwise it should already be on the device no?

```suggestion
                    if args.offload: vae = vae.to(accelerator.device)
```

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1708 @@
...
+    elif accelerator.mixed_precision == "bf16":
+        weight_dtype = torch.bfloat16
+
+    if torch.backends.mps.is_available() and weight_dtype == torch.bfloat16:
+        # due to pytorch#99272, MPS does not yet support bfloat16.
+        raise ValueError(
+            "Mixed precision training with bfloat16 is not supported on MPS. Please use fp16 (recommended) or fp32 instead."
+        )
+
+    vae.to(dtype=weight_dtype)
```
sayakpaul, line 1103:
```suggestion
    if not args.offload:
         vae.to(dtype=weight_dtype, device=accelerator.device)
    else:
        vae.to(dtype=weight_dtype)
```

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1708 @@
...
+        transformer=None,
+        text_encoder=text_encoder_one,
+        tokenizer=tokenizer_one,
+        text_encoder_2=text_encoder_two,
+        tokenizer_2=tokenizer_two,
+        text_encoder_3=text_encoder_three,
+        tokenizer_3=tokenizer_three,
+        text_encoder_4=text_encoder_four,
+        tokenizer_4=tokenizer_four,
+    )
```
sayakpaul, line 1123: I think the `text_encoding_pipeline` won't be on a CPU here because we're already moving the text encoders to `accelerator.device` just a few lines above.

sayakpaul (commented):
Thank you!

I left some comments. LMK if that makes sense. 

Would also be nice to update the description of the PR with some successful runs.

examples/dreambooth/README_hidream.md
linoytsaban: sounds good, just for clarification in this command text embedding are pre-computed since it using `instance_prompt` and custom prompts are not provided (re: our slack conversation)

examples/dreambooth/README_hidream.md
linoytsaban: yes good point

examples/dreambooth/train_dreambooth_lora_hidream.py
linoytsaban: oops yeah I forgot to remove the comment

examples/dreambooth/train_dreambooth_lora_hidream.py
linoytsaban: it wasn't already on device since it was initially `vae.to(dtype=weight_dtype)` for all cases, but now I pushed the change to have it (since it's more clear that way):
```
    if not args.offload:
        vae.to(dtype=weight_dtype, device=accelerator.device)
    else:
        vae.to(dtype=weight_dtype)
```
so this condition indeed needs to be changes

examples/dreambooth/train_dreambooth_lora_hidream.py
linoytsaban: yeah let me fix that

linoytsaban: how about now?

linoytsaban:
@bot /style

examples/dreambooth/train_dreambooth_lora_hidream.py
bghira: move it to the negation then. it's important to leave this note in because it's so non-standard.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -5360,6 +5360,329 @@ def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
...
+            warn_msg = "It seems like you are using a DoRA checkpoint that is not compatible in Diffusers at the moment. So, we are going to filter out the keys associated to 'dora_scale` from the state dict. If you think this is a mistake please open an issue https://github.com/huggingface/diffusers/issues/new."
+            logger.warning(warn_msg)
+            state_dict = {k: v for k, v in state_dict.items() if "dora_scale" not in k}
+
+        # conversion.
+        non_diffusers = any(k.startswith("diffusion_model.") for k in state_dict)
+        if non_diffusers:
+            state_dict = _convert_non_diffusers_lumina2_lora_to_diffusers(state_dict)
+
+        return state_dict
```
sayakpaul, line 5471: We don't need it no? And we can use the one from 
https://github.com/huggingface/diffusers/blob/5873377a660dac60a6bd86ef9b4fdfc385305977/src/diffusers/loaders/lora_pipeline.py#L1465

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -5360,6 +5360,329 @@ def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
...
+            state_dict,
+            transformer=getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer,
+            adapter_name=adapter_name,
+            _pipeline=self,
+            low_cpu_mem_usage=low_cpu_mem_usage,
+            hotswap=hotswap,
+        )
+
+    @classmethod
+    # Copied from diffusers.loaders.lora_pipeline.SD3LoraLoaderMixin.load_lora_into_transformer with SD3Transformer2DModel->Lumina2Transformer2DModel
```
sayakpaul, line 5532: Should not be Lumina2.

sayakpaul (commented):
Looking good! I think in `lora_pipeline.py`, the changes from Lumina2 shouldn't be included.

examples/dreambooth/train_dreambooth_lora_hidream.py
```diff
@@ -0,0 +1,1711 @@
...
+        text_encoder_one.to(dtype=weight_dtype, device=accelerator.device)
+        text_encoder_two.to(dtype=weight_dtype, device=accelerator.device)
+        text_encoder_three.to(dtype=weight_dtype, device=accelerator.device)
+        text_encoder_four.to(dtype=weight_dtype, device=accelerator.device)
+    else:
+        vae.to(dtype=weight_dtype)
+        text_encoder_one.to(dtype=weight_dtype)
+        text_encoder_two.to(dtype=weight_dtype)
+        text_encoder_three.to(dtype=weight_dtype)
+        text_encoder_four.to(dtype=weight_dtype)
```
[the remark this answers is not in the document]
linoytsaban: nice!

src/diffusers/loaders/lora_pipeline.py
linoytsaban: for ostris's hidream lora I saw the naming convention is the same as in the lumina case: 
https://github.com/ostris/ai-toolkit/blob/79c87701e7fc6f47f1ea8adc75834284b007d3ab/extensions_built_in/diffusion_models/hidream/hidream_model.py#L430

linoytsaban:
Hi @Nerogar! 
For context - the `self.training`/ `if training` conditions in the `HiDreamImageTransformer2DModel` implementation come from the original implementation and refer to the case where aux loss is used and the experts are trained as well. 
For LoRA training in this PR we currently do not support aux loss and MoE training, and hence we added 
`--fore_inference_output `https://github.com/huggingface/diffusers/blob/73ab2017f13b74c3a3a92212d015792adfad2b95/src/diffusers/models/transformers/transformer_hidream_image.py#L320

Specifically we encountered an [issue](https://github.com/HiDream-ai/HiDream-I1/issues/35#issue-2996119909) with `unpatchify` when initially working on this PR, with the resulting shape 
of the output not as we'd expect, resulting with an error when trying to calculate the loss due to mismatch of the shapes between `target` and `model_pred`

linoytsaban:
@bot /style

sayakpaul (approved):
Thank you!

vladmandic:
follow-up issue ##11383

PR #11282, opened 2025-04-10
Mark test_layerwise_upcasting as is_flaky

hlky:
# What does this PR do?

https://github.com/huggingface/diffusers/actions/runs/14371854670/job/40296313951?pr=11268#step:7:8122

## Who can review?

### DIFF
--- a/tests/models/test_modeling_common.py
+++ b/tests/models/test_modeling_common.py
@@ -61,6 +61,7 @@ from diffusers.utils.testing_utils import (
     backend_empty_cache,
     floats_tensor,
     get_python_version,
+    is_flaky,
     is_torch_compile,
     numpy_cosine_similarity_distance,
     require_peft_backend,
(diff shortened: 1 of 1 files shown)


tests/models/test_modeling_common.py
```diff
@@ -1436,6 +1437,7 @@ def test_fn(storage_dtype, compute_dtype):
         test_fn(torch.float8_e5m2, torch.float32)
         test_fn(torch.float8_e4m3fn, torch.bfloat16)
 
+    @is_flaky
```
sayakpaul, line 1440: If it's only for `test_models_autoencoder_hunyuan_video.py`, I would consider subclassing it inside `AutoencoderKLHunyuanVideoTests` and decorating it with `is_flaky`.

hlky: These tests haven't been running on CI for a while due to the unclip issue so it is potentially flaky for other models too.

sayakpaul: Would prefer confirming that first before marking all tests as flaky when they appear to be not.

PR #11284, opened 2025-04-10
[single file] enable telemetry for single file loading when using GGUF.

sayakpaul:
# What does this PR do?

Otherwise, we won't be able to collect stats on the usage of quantization methods like GGUF.

Cc: @SunMarc


SunMarc (approved):
LGTM

sayakpaul:
Failing tests are unrelated.

PR #11288, opened 2025-04-11
fix CPU offloading related fail cases on XPU

yao-matrix:
**Symptom**
As of now, below 5 CPU related offloading cases fail on XPU

> tests/pipelines/kandinsky/test_kandinsky_img2img.py::KandinskyImg2ImgPipelineFastTests::test_offloads
   tests/lora/test_lora_layers_sd.py::LoraIntegrationTests::test_a1111_with_model_cpu_offload
   tests/lora/test_lora_layers_sd.py::LoraIntegrationTests::test_a1111_with_sequential_cpu_offload
   tests/lora/test_lora_layers_sd.py::LoraIntegrationTests::test_sd_load_civitai_empty_network_alpha
tests/pipelines/stable_diffusion/test_stable_diffusion.py::StableDiffusionPipelineSlowTests::test_stable_diffusion_textual_inversion_with_model_cpu_offload 

**Proposal**
change `device` arg's default value from `"cuda"` to `None` in `enable_model_cpu_offload` and `enable_sequential_cpu_offload`, and add an automatically device detection logic in these 2 functions to detect accelerator if not specified.

**Possible Questions**
**Q1: why not just change the test case code to specify the device?**
   [A1]: this 2 utility functions is heavily used in docs and jupyter notebooks, and also be used in internal lora loader(as [here](https://github.com/yao-matrix/diffusers/blob/ea5a6a8b7cee3a0cd1dea691f4219a562c0ea3bd/src/diffusers/loaders/lora_base.py#L424)  ) and ti loader(as [here](https://github.com/yao-matrix/diffusers/blob/ea5a6a8b7cee3a0cd1dea691f4219a562c0ea3bd/src/diffusers/loaders/textual_inversion.py#L451)) with the assumption that if not specified, the arg should be accelerator(before this PR, this narrowly means "cuda"). We need inherit this good assumption and generalize the accelerator from cuda to all types of accelerators, so diffusers can support multiple devices with zero code change.

**Q2: why not add "npu" and other devices while `get_device`?**
   [A2]: I don't have these device access, so I prefer those guys who have access do that, so they can validate before merging PRs. And it's simple to add a if-branch in `get_device`, most of the efforts are validation efforts.

**Backward Compatibility**
Full backward compatibility.

- If CUDA device and non arg passed: before this PR, will use cuda device; after this PR, will detect the system and if cuda detected, will use cuda device.
- If CUDA device and cuda arg passed: before this PR, will use cuda device; after this PR, will use cuda device.
- If XPU device and cuda arg passed: before this PR, will crash and report confusing log "Torch not compiled with CUDA enabled"; after this PR, will automatically detect xpu device and use xpu device.
- If XPU device and xpu arg passed: before this PR, will use xpu device; after this PR, will use xpu device.


yao-matrix:
@hlky , pls help review and comments, thx very much.

hlky (approved):
Thanks @yao-matrix. Changes look good to me and save updating a lot of other code/docs/tests

```
docs/   enable_model_cpu_offload() 169 results in 40 files
docs/   enable_sequential_cpu_offload()  15 results in 8 files
src/    enable_model_cpu_offload()  `_load_lora_into_text_encoder`, `load_lora_adapter`, `load_attn_procs`, and 43 results in 35 files for examples
src/    enable_sequential_cpu_offload()  `_load_lora_into_text_encoder`, `load_lora_adapter`, `load_attn_procs`
tests/  enable_model_cpu_offload()  47 results in 21 files
tests/  enable_sequential_cpu_offload()  14 results in 9 files
```

There is one related fast test failing (others are transient Hub issues):
https://github.com/huggingface/diffusers/blob/f1f38ffbeed793d684684e00e6e1213bcaf494d6/tests/pipelines/test_pipelines.py#L1802-L1822

This test could either be moved to `PipelineSlowTests`, or if it's already passing when ran on a non-CUDA system we could simply add `device="cuda"` as the test is not actually offloading anything and only checking `_offload_gpu_id` is set correctly.

https://github.com/huggingface/diffusers/blob/f1f38ffbeed793d684684e00e6e1213bcaf494d6/tests/pipelines/test_pipelines.py#L1819

Can you confirm whether this test passes on XPU system, @yao-matrix?

Let's also get a second opinion, cc @sayakpaul

hlky:
@sayak This is a fast test that runs on CPU, if we decorate with `require_torch_gpu` it will be skipped. The test itself is not offloading anything and therefore does not actually require a torch gpu and is only checking _offload_gpu_id is set correctly.

sayakpaul:
My bad. Then all good.

yao-matrix:
> Thanks @yao-matrix. Changes look good to me and save updating a lot of other code/docs/tests
> 
> ```
> docs/   enable_model_cpu_offload() 169 results in 40 files
> docs/   enable_sequential_cpu_offload()  15 results in 8 files
> src/    enable_model_cpu_offload()  `_load_lora_into_text_encoder`, `load_lora_adapter`, `load_attn_procs`, and 43 results in 35 files for examples
> src/    enable_sequential_cpu_offload()  `_load_lora_into_text_encoder`, `load_lora_adapter`, `load_attn_procs`
> tests/  enable_model_cpu_offload()  47 results in 21 files
> tests/  enable_sequential_cpu_offload()  14 results in 9 files
> ```
> 
> There is one related fast test failing (others are transient Hub issues):
> 
> https://github.com/huggingface/diffusers/blob/f1f38ffbeed793d684684e00e6e1213bcaf494d6/tests/pipelines/test_pipelines.py#L1802-L1822
> 
> This test could either be moved to `PipelineSlowTests`, or if it's already passing when ran on a non-CUDA system we could simply add `device="cuda"` as the test is not actually offloading anything and only checking `_offload_gpu_id` is set correctly.
> 
> https://github.com/huggingface/diffusers/blob/f1f38ffbeed793d684684e00e6e1213bcaf494d6/tests/pipelines/test_pipelines.py#L1819
> 
> Can you confirm whether this test passes on XPU system, @yao-matrix?
> 
> Let's also get a second opinion, cc @sayakpaul

@hlky, this case passes on XPU, I pasted the log as below:
"PASSED tests/pipelines/test_pipelines.py::PipelineFastTests::test_pipe_same_device_id_offload"

The reason is in the `enable_model_cpu_offload`, it will convert `gpu_id` to device w/ below code:
(`device = torch.device(f"{device_type}:{self._offload_gpu_id}")`)[https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/pipeline_utils.py#L1134C9-L1134C71]
and before that, we've already detected the device correctly and assign the correct `device_type`.

Thx.

Issue #11062, opened 2025-03-14
Error in loading Civit AI Lora: LCMTurboMix_Euler_A_fix


ali-afridi26:
### Describe the bug

[This CIVITAI Lora](https://civitai.com/models/216190/lora) has over 20k downloads and doesn't work with SDXL Pipeline. It is giving `lora_unet_down_blocks_0_downsamplers_0_conv.alpha` not supported error. I have uploaded the model on hugging face. Error appears on `load_lora_weights()` function

### Reproduction

```
from diffusers import DiffusionPipeline

pipe = DiffusionPipeline.from_pretrained("stabilityai/stable-diffusion-xl-base-1.0")
pipe.load_lora_weights("RhaegarKhan/LCMTurboMix_Euler_A_fix")

prompt = "<lora:LCMTurboMix2fix:1>,abstract portrait of 1girl,undefined gender,fragmented visual style,red and black color palette,evokes feelings of rebellion,passion,and freedom,blurred boundaries,high resolution,aesthetic,"
image = pipe(prompt).images[0]
```

### Logs

```shell
Loading pipeline components...: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 7/7 [00:02<00:00,  3.35it/s]
LCMTurboMix_Euler_A_fix.safetensors: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████| 13.0M/13.0M [00:00<00:00, 38.8MB/s]
Traceback (most recent call last):
  File "/home/user/runware/Ali/sd-base-api/lora.py", line 4, in <module>
    pipe.load_lora_weights("RhaegarKhan/LCMTurboMix_Euler_A_fix")
  File "/home/user/runware/shehzad/temp/sd-base-api/diffusers/src/diffusers/loaders/lora_pipeline.py", line 545, in load_lora_weights
    state_dict, network_alphas = self.lora_state_dict(
                                 ^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/anaconda3/envs/3.7/lib/python3.12/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/user/runware/shehzad/temp/sd-base-api/diffusers/src/diffusers/loaders/lora_pipeline.py", line 695, in lora_state_dict
[...]

PR #11292, opened 2025-04-11
[LoRA] support more SDXL loras.

sayakpaul:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/issues/11062

@ali-afridi26 could you give this PR a try?

(note to self: run all respective integration tests before merging)


hlky (changes requested):
The LoRA contains both "sgm" and "not_sgm" keys, if we remove "not_sgm" keys then any LoRA with only "not_sgm" keys will be empty. Consider the following code:

```python
from safetensors.torch import load_file

file = hf_hub_download("RhaegarKhan/LCMTurboMix_Euler_A_fix", filename="LCMTurboMix_Euler_A_fix.safetensors")
state_dict = load_file(file)

all_keys = list(state_dict.keys())
sgm_patterns = ["input_blocks", "middle_block", "output_blocks"]
not_sgm_patterns = ["down_blocks", "mid_block", "up_blocks"]

not_sgm = {}

for block in not_sgm_patterns:
    for k in all_keys:
        if block in k:
            not_sgm[k] = state_dict.pop(k)

pipe = DiffusionPipeline.from_pretrained("stabilityai/stable-diffusion-xl-base-1.0")
pipe.load_lora_weights(not_sgm)
```

(logging added to the PR's code)
```
removing k='lora_unet_down_blocks_1_attentions_0_proj_in.lora_down.weight'
removing k='lora_unet_down_blocks_1_attentions_0_proj_in.lora_up.weight'
removing k='lora_unet_down_blocks_1_attentions_0_proj_out.alpha'
...
is_in_sgm_format=False
No LoRA keys associated to UNet2DConditionModel found with the prefix='unet'. This is safe to ignore if LoRA state dict didn't originally have any UNet2DConditionModel related params. You can also try specifying `prefix=None` to resolve the warning. Otherwise, open an issue if you think it's unexpected: https://github.com/huggingface/diffusers/issues/new
```

The resulting `state_dict` is empty.

On `main`, the above `not_sgm` state_dict will load correctly.

For a LoRA like this containing both types of keys we should prefer `not_sgm` to avoid unnecessary conversion. We can do something like this:

```python

    # check if state_dict contains both patterns
    contains_sgm_patterns = False
    contains_not_sgm_patterns = False
    for key in all_keys:
        if any(p in key for p in sgm_patterns):
            contains_sgm_patterns = True
        elif any(p in key for p in not_sgm_patterns):
            contains_not_sgm_patterns = True

    # if state_dict contains both patterns, remove sgm
    # we can then return state_dict immediately
    if contains_sgm_patterns and contains_not_sgm_patterns:
        for key in all_keys:
            if any(p in key for p in sgm_patterns):
                state_dict.pop(key)
        return state_dict
```

This works for all 3 cases, `sgm` state_dict, `not_sgm` state_dict and state_dict with both.

sayakpaul:
Nice catch. This is why I have a:

> (note to self: run all respective integration tests before merging)

In the PR description :D 

Do you want to push directly to the PR?

src/diffusers/loaders/lora_conversion_utils.py
```diff
@@ -126,7 +144,7 @@ def _maybe_map_sgm_blocks_to_diffusers(state_dict, unet_config, delimiter="_", b
             )
             new_state_dict[new_key] = state_dict.pop(key)
 
-    if len(state_dict) > 0:
+    if state_dict:
```
hlky, line 147: Just to note this is a good change, I'd have to find the source again but it should be marginally faster due to Python internals.

hlky (approved):
Thanks for working on this @sayakpaul. Out of curiosity do we know which training UI is producing state_dicts with both types of keys?

src/diffusers/loaders/lora_conversion_utils.py
sayakpaul: Sadly I don't know :/ 

Also, thanks for catching the edge case and promptly fixing it!

sayakpaul:
Everything passing. All good.

src/diffusers/loaders/lora_conversion_utils.py
asomoza: probably is not a training library but a merge made by someone, judging by the name of that lora it seems to be the case.

Issue #11118, opened 2025-03-19
HunyuanVideoImageToVideoPipeline failures


vladmandic:
### Describe the bug

pipline `HunyuanVideoImageToVideoPipeline` fails with latest combination of `diffusers` and `transformers` libraries

first, minor issue is with offloading - this snipped updates `pipeline_hunyuan_video_image2video.py` to add explicit `.to(device)` so two torch.cat operations do not fail.
```py
            if last_double_return_token_indices.shape[0] == 3:
                # in case the prompt is too long
                last_double_return_token_indices = torch.cat(
                    (last_double_return_token_indices, torch.tensor([text_input_ids.shape[-1]], device=last_double_return_token_indices.device))
                )
                batch_indices = torch.cat((batch_indices, torch.tensor([0], device=batch_indices.device)))
```

bigger issue is that `transformers` updated how image embeds work in `LlavaForConditionalGeneration`,  
so function `_get_llama_prompt_embeds` in `HunyuanVideoImageToVideoPipeline` needs an update  
(last version of transformers that works is `transformers==4.47.1`)  

specifically, it returns `prompt_embeds` and `prompt_attention_mask` which don't have the same length due to way that cropping is implemented, so later cannot be combined in `HunyuanVideoTokenRefiner`:

### Reproduction

see #10983 for simple example

### Logs

```shell
│ /home/vlado/dev/sdnext/venv/lib/python3.12/site-packages/diffusers/models/transformers/transformer_hunyuan_video.py:312 in forward                                                                                                                                                                                                                                                                                                    │
│                                                                                                                                                                                                                                                                                                                                                                                                                                       │
│   311 │   │   │   mask_float = attention_mask.float().unsqueeze(-1)                                                                                                                                                                                                                                                                                                                                                                   │
│ ❱ 312 │   │   │   pooled_projections = (hidden_states * mask_float).sum(dim=1) / mask_float.sum(dim=1)                                                                                                                                                                                                                                                                                                                                │
│   313 │   │   │   pooled_projections = pooled_projections.to(original_dtype)                                                                                                                                                                                                                                                                                                                                                          │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
[...]

PR #11293, opened 2025-04-11
Fix Hunyuan I2V for `transformers>4.47.1`  

DN6:
# What does this PR do?
We need to manually expand the `text_input_ids` to include the token ids for the image. The function responsible for it in transformers has been deprecated, and the logic in the current `LLaVAProcessor` isn't compatible with the way `max_sequence_length` should be applied in this model.  

This PR
1. Adds an `_expand_input_ids_with_image_tokens` function to replicate some of the behaviour in the LLava model in transformers<4.47.1
https://github.com/huggingface/transformers/blob/241c04d36867259cdf11dbb4e9d9a60f9cb65ebc/src/transformers/models/llava/modeling_llava.py#L302
  
This should keep backwards compatibility and work with the newer transformers versions, provided the prompt template and token indices for the image token do not change (seems unlikely?) 

Fixes # (issue)
#11118

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
+++ b/src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
@@ -24,7 +24,6 @@ from transformers import (
     CLIPTokenizer,
     LlamaTokenizerFast,
     LlavaForConditionalGeneration,
-    LlavaProcessor,
 )
 
 from ...callbacks import MultiPipelineCallbacks, PipelineCallback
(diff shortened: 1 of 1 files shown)


src/diffusers/pipelines/hunyuan_video/pipeline_hunyuan_video_image2video.py
```diff
@@ -100,6 +100,50 @@
 }
 
 
+def _expand_input_ids_with_image_tokens(
```
a-r-r-o-w, line 103: Maybe could point to a specific commit by adding a comment about where this was adapted from (assuming if it was)

a-r-r-o-w (approved):
LGTM if the outputs match before/after these changes! Prompt template and token indices are very unlikely to change by end users, but even if they do, they need to provide the correct indices to work, so I would assume it should be okay on our end

DN6:
> LGTM if the outputs match before/after these changes! 

Hmm so prompt embeds do change, because the way the input embeds are created in LLava `<=4.47.1` is different from the current version. Old version, the input embeds ended up being zero padded because of this line
https://github.com/huggingface/transformers/blob/241c04d36867259cdf11dbb4e9d9a60f9cb65ebc/src/transformers/models/llava/modeling_llava.py#L325

With the current changes, I think the padding token embedding is used to pad to the end of the sequence. Since the attention mask is passed in both cases, all the differences in the embeddings appear to be where the pad tokens are supposed to be, visually the pipeline outputs look pretty much the same, but numerically the prompt embeds are different. 

Old:
https://github.com/user-attachments/assets/77917cc9-a52e-41bf-9ec5-4831fd2f61a4

Current Fix:
https://github.com/user-attachments/assets/2af19aa0-23c4-410e-9638-40d100cd6700

DN6:
@vladmandic Do you have some time to check if this fixes the issue you were facing?

vladmandic:
quick `HunyuanVideoImageToVideoPipeline` with
- `diffusers==hunyuan-i2v-update`
- `transformers==4.51.1`

lgtm!

PR #11294, opened 2025-04-11
Fix incorrect tile_latent_min_width calculation in AutoencoderKLMochi

### LINKED ISSUE #11291: Inconsistent variable usage in `tile_latent_min_width` computation in `AutoencoderKLMochi` decoder
### Describe the bug

There seems to be an inconsistency in the calculation of `tile_latent_min_width` inside the decoder of the AutoencoderKLMochi model. In the [following line of code](https://github.com/huggingface/diffusers/blob/7054a34978e68bc2b7241378c07d938066c1aa64/src/diffusers/models/autoencoders/autoencoder_kl_mochi.py#L912), `tile_latent_min_width` is calculated using:

```python
tile_latent_min_width = self.tile_sample_stride_width // self.spatial_compression_ratio
```

However, directly above it, `tile_latent_min_height` is calculated as:

```python
tile_latent_min_height = self.tile_sample_min_height // self.spatial_compression_ratio
```

This discrepancy appears to be unintended. Logically, the width counterpart should follow the same pattern. The correct computation for `tile_latent_min_width` should likely be:

```python
tile_latent_min_width = self.tile_sample_min_width // self.spatial_compression_ratio
```

Further evidence supporting this interpretation can be found in the [tiled decoding section at line 1057](https://github.com/huggingface/diffusers/blob/7054a34978e68bc2b7241378c07d938066c1aa64/src/diffusers/models/autoencoders/autoencoder_kl_mochi.py#L1057), where `tile_latent_min_width` is indeed correctly computed using `self.tile_sample_min_width`

This reinforces the idea that the line at L912 is likely a copy-paste or naming oversight.

While this issue may not cause runtime errors under standard conditions—such as when generating 848x480 videos with mochi, where the check at [L914](https://github.com/huggingface/diffusers/blob/7054a34978e68bc2b7241378c07d938066c1aa64/src/diffusers/models/autoencoders/autoencoder_kl_mochi.py#L914) still passes—it can affect edge cases (e.g., for profiling, debugging with custom video dimensions), potentially leading to inconsistent or unintended behavior.

If the maintainers agree this is indeed a bug, I’d be happy to submit a PR to fix it.

### Reproduction

This issue does not trigger an immediate runtime error but leads to an incorrect value being calculated, which may result in unintended behavior when tiling is enabled in the VAE.

To reproduce the problem, simply run the Mochi pipeline with the following configuration:

```python
width = 848
height = 480
vae_tiling = True  # Enable VAE tiling
```

With this setup and default parameters, the AutoencoderKLMochi._decode function will receive the following relevant attributes:

```python
width = 106 (downsampled by spatial_compression_ratio)
height = 60 (downsampled by spatial_compression_ratio)
self.tile_sample_min_height = 256
self.tile_sample_min_width  = 256
self.tile_sample_stride_width = 192
self.spatial_compression_ratio = 8
```

From the current code at [L912](https://github.com/huggingface/diffusers/blob/7054a34978e68bc2b7241378c07d938066c1aa64/src/diffusers/models/autoencoders/autoencoder_kl_mochi.py#L912), the computed value becomes:
```python
tile_latent_min_width = self.tile_sample_stride_width // self.spatial_compression_ratio
                       = 192 // 8
                       = 24
```

Instead, it should have been:

```python
tile_latent_min_width = self.tile_sample_min_width // self.spatial_compression_ratio
                       = 256 // 8
                       = 32
```

As a result, the conditional check:

```python
if width > tile_latent_min_width:
```

incorrectly evaluates to True for `32<=width<24`, even though it should evaluate to False with the correct `tile_latent_min_width = 32`.

### Logs

```shell

```

### System Info

I realized this on my environment with diffusers 0.32.1 but it's still present in the master branch.

### Who can help?

@DN6

kuantuna:
# What does this PR do?
This PR fixes an inconsistency in the calculation of `tile_latent_min_width` inside the AutoencoderKLMochi decoder.

Previously, `tile_latent_min_width` was incorrectly calculated using `self.tile_sample_stride_width`:

```python
tile_latent_min_width = self.tile_sample_stride_width // self.spatial_compression_ratio
```

This led to a mismatch with the corresponding height calculation, which correctly used `self.tile_sample_min_height`. The corrected line now uses `self.tile_sample_min_width`.

This discrepancy did not raise errors under default settings (e.g., 848x480 resolution), but could cause incorrect conditional evaluations in edge cases, such as when profiling or testing different resolutions.

Fixes #11291.

## Who can review?

@a-r-r-o-w

### DIFF
--- a/src/diffusers/models/autoencoders/autoencoder_kl_mochi.py
+++ b/src/diffusers/models/autoencoders/autoencoder_kl_mochi.py
@@ -909,7 +909,7 @@ class AutoencoderKLMochi(ModelMixin, ConfigMixin):
     def _decode(self, z: torch.Tensor, return_dict: bool = True) -> Union[DecoderOutput, torch.Tensor]:
         batch_size, num_channels, num_frames, height, width = z.shape
         tile_latent_min_height = self.tile_sample_min_height // self.spatial_compression_ratio
-        tile_latent_min_width = self.tile_sample_stride_width // self.spatial_compression_ratio
+        tile_latent_min_width = self.tile_sample_min_width // self.spatial_compression_ratio
 
         if self.use_tiling and (width > tile_latent_min_width or height > tile_latent_min_height):
             return self.tiled_decode(z, return_dict=return_dict)


a-r-r-o-w (approved):
Awesome, thank you! 🎉

kuantuna:
Thanks for the review! Let me know if there's anything else needed from my side 🙌

PR #11296, opened 2025-04-12
[Hi Dream] follow-up

yiyixuxu:
```python
import torch
from transformers import PreTrainedTokenizerFast, LlamaForCausalLM
from diffusers import (
    UniPCMultistepScheduler,
    HiDreamImagePipeline,
    HiDreamImageTransformer2DModel,
)


# Get more detailed memory stats
def print_detailed_memory(step_name):
    print(f"\n=== CUDA Memory Stats {step_name} ===")
    print(f"Current allocated: {torch.cuda.memory_allocated() / 1024**3:.2f} GB")
    print(f"Max allocated: {torch.cuda.max_memory_allocated() / 1024**3:.2f} GB")
    print(f"Current reserved: {torch.cuda.memory_reserved() / 1024**3:.2f} GB")
    print(f"Max reserved: {torch.cuda.max_memory_reserved() / 1024**3:.2f} GB")


device = torch.device("cuda:0")

# Clear CUDA cache
if torch.cuda.is_available():
    torch.cuda.empty_cache()

# Initial memory usage
print_detailed_memory("start")


llama_repo = "unsloth/Meta-Llama-3.1-8B-Instruct"
# llama_repo = "meta-llama/Meta-Llama-3.1-8B-Instruct"

tokenizer_4 = PreTrainedTokenizerFast.from_pretrained(
    llama_repo,
)

text_encoder_4 = LlamaForCausalLM.from_pretrained(
    llama_repo,
    output_hidden_states=True,
    output_attentions=True,
    torch_dtype=torch.bfloat16,
)



pipe = HiDreamImagePipeline.from_pretrained(
    "HiDream-ai/HiDream-I1-Full",
    scheduler=None,
    tokenizer_4=tokenizer_4,
    text_encoder_4=text_encoder_4,
    transformer=None,
    vae=None,
    torch_dtype=torch.bfloat16,
)


print_detailed_memory("Before encode prompt")

with torch.no_grad():
    prompt_embeds, negative_prompt_embeds, pooled_prompt_embeds, negative_pooled_prompt_embeds = pipe.encode_prompt(
        'A cat holding a sign that says "Hi-Dreams.ai".',
        negative_prompt="bad quality, low quality",
        device=device,
        dtype=torch.bfloat16,
    )

print_detailed_memory("After encode prompt")
```


src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py
```diff
@@ -339,118 +339,104 @@ def encode_prompt(
...
-                self.tokenizer, self.text_encoder, prompt, max_sequence_length, device, dtype
-            )
             pooled_prompt_embeds_2 = self._get_clip_prompt_embeds(
                 self.tokenizer_2, self.text_encoder_2, prompt_2, max_sequence_length, device, dtype
             )
+
+        if do_classifier_free_guidance and negative_pooled_prompt_embeds is None:
+            negative_prompt_2 = negative_prompt_2 or ""
+            negative_prompt_2 = (
+                batch_size * [negative_prompt_2] if isinstance(negative_prompt_2, str) else negative_prompt_2
```
a-r-r-o-w, line 367: Could we rewrite this such that if you have the same prompt, you don't have to repeat by batch size and encode the same thing again and again? It should be much faster to encode once and then repeat by batch size

yiyixuxu: ohh but it is encoded with different text_encoder though, no?

a-r-r-o-w: I'm referring to the `batch_size * [prompt_]` parts. If batch size is > 1, you will do a slower forward pass through the text encoders as opposed to doing a faster forward pass by not repeating the prompt before text encoder but instead do it afterwards.

bghira:
why did you carry forward the weird thing where prompt_embeds is a list of tensors from their original implementation? it's unnecessary complexity and confuses people. please split them between t5_prompt_embeds and llama_prompt_embeds as in [this implementation](https://github.com/bghira/SimpleTuner/pull/1380/files#diff-33261dd95b06ecdb73f54b109fba251d39cce5ff11eacc17b0506df904682939)

you'll also want to rename their attention modules so that block name matching works correctly in downstream consumers like Lycoris.

i'm surprised this went in so haphazardly, can we step back and not rush this stuff ? there's no need to.

bghira:
this version also has really poor memory efficiency during training with the MoEGate tracking gradients through the whole network unnecessarily :cry:

sayakpaul:
The `check_inputs()` function could benefit from:

```diff
diff --git a/src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py b/src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py
index aa1de849e..ecca4548a 100644
--- a/src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py
+++ b/src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py
@@ -545,12 +545,14 @@ class HiDreamImagePipeline(DiffusionPipeline):
             )
 
         if prompt_embeds is not None and negative_prompt_embeds is not None:
-            if prompt_embeds.shape != negative_prompt_embeds.shape:
-                raise ValueError(
-                    "`prompt_embeds` and `negative_prompt_embeds` must have the same shape when passed directly, but"
-                    f" got: `prompt_embeds` {prompt_embeds.shape} != `negative_prompt_embeds`"
-                    f" {negative_prompt_embeds.shape}."
-                )
+            if isinstance(prompt_embeds, list) and isinstance(negative_prompt_embeds, list):
+                for i, (p, n) in enumerate(zip(prompt_embeds, negative_prompt_embeds)):
+                    if p.shape != n.shape:
+                        raise ValueError(
+                            "`prompt_embeds` and `negative_prompt_embeds` must have the same shape when passed directly, but"
+                            f" got: `prompt_embeds[{i}]` {p.shape} != `negative_prompt_embeds[{i}]`"
+                            f" {n.shape}."
+                        )
 
         if prompt_embeds is not None and pooled_prompt_embeds is None:
             raise ValueError(

```

This is required when the prompt embeds are passed.

yiyixuxu:
ohh yes, I think we should have 2 prompt_embeds, (t5 and llama) like @bghira suggested, instead of list, in that case, we should accept them as inputs too

sayakpaul:
Oh indeed!

bghira:
there's an issue when trying to make image with pre-encoded embeds and moving text encoder to `meta` device. it complains about tensor device misalignment. however, we can set `self._execution_device = self.transformer.device` manually inside `__call__` and it then works :partying_face:

yiyixuxu:
@a-r-r-o-w can you look into the comment here, if it makes sense, we can edit in a separate PR (since it is training related)
https://github.com/huggingface/diffusers/pull/11296#issuecomment-2798737662

yiyixuxu:
@bghira 
I think this is expected,  pre-encoded embeds works out-of-box with `enable_model_cpu_offload`though - we put hooks on the models and they are moved to device when forward pass is called. Since you are using pre-encoded embeds, text_encoders are not used so they stay on cpu entire time, and it won't affect `self._execution_device`

>there's an issue when trying to make image with pre-encoded embeds and moving text encoder to meta device. it complains about tensor device misalignment. however, we can set self._execution_device = self.transformer.device manually inside __call__ and it then works 🥳

a-r-r-o-w (approved):
LGTM, just one comment.

Thanks for the suggestions @bghira!

I'll look into rewriting the MoE related code. You could probably just make the MoEGate behave the same way as it'd do during inference. This simplification should work for LoRA-related training since most low-rank training is done on the QKVO layers. 

The additional gradient checkpointing does not make sense to me in the MoEGate layer though. It saves tracking gradients through a `b * s * top_k` tensor (some constant x `num_layers`) times, and that region is not really compute intensive, so that should not be a whole lot of memory compared to what happens in other layers. Yet to test training so I'll try to understand this better.

bghira:
> LGTM, just one comment.
> 
> Thanks for the suggestions @bghira!
> 
> I'll look into rewriting the MoE related code. You could probably just make the MoEGate behave the same way as it'd do during inference. This simplification should work for LoRA-related training since most low-rank training is done on the QKVO layers.
> 
> The additional gradient checkpointing does not make sense to me in the MoEGate layer though. It saves tracking gradients through a `b * s * top_k` tensor (some constant x `num_layers`) times, and that region is not really compute intensive, so that should not be a whole lot of memory compared to what happens in other layers. Yet to test training so I'll try to understand this better.

believe me, we hit >100G VRAM use without that change. 

but we're not doing LoRA training for the most part, we're doing Lycoris algorithms which hit these issues easily.

sayakpaul:
@yiyixuxu sorry for the late request. Do you think the following block could be turned into a static method like we [have](https://github.com/huggingface/diffusers/blob/7edace9a0578dd699aadeb1799164d790b87ce6b/src/diffusers/pipelines/flux/pipeline_flux.py#L528) in FluxPipeline?

https://github.com/huggingface/diffusers/blob/72c96676e13122dfaa5dbd829d99451d336bfb24/src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py#L634-L652

This would help training quite a bit.

ShuyUSTC:
We plan to update the scheduler config in our Hugging Face model repository. After this change, all `HiDream-I1` pipelines (`Full`/`Dev`/`Fast`) can be initialized as follows:
```python
import torch
from transformers import PreTrainedTokenizerFast, LlamaForCausalLM
from diffusers import UniPCMultistepScheduler, HiDreamImagePipeline

tokenizer_4 = PreTrainedTokenizerFast.from_pretrained("meta-llama/Meta-Llama-3.1-8B-Instruct")
text_encoder_4 = LlamaForCausalLM.from_pretrained(
    "meta-llama/Meta-Llama-3.1-8B-Instruct",
    output_hidden_states=True,
    output_attentions=True,
    torch_dtype=torch.bfloat16,
)

pipe = HiDreamImagePipeline.from_pretrained(
    "HiDream-ai/HiDream-I1-Full",  # "HiDream-ai/HiDream-I1-Dev" | "HiDream-ai/HiDream-I1-Fast"
    scheduler=scheduler,
    tokenizer_4=tokenizer_4,
    text_encoder_4=text_encoder_4,
    torch_dtype=torch.bfloat16,
)
```

There is no longer a need to explicitly initialize a `scheduler` outside of `HiDreamImagePipeline`. For more details, please refer to this pr: https://huggingface.co/HiDream-ai/HiDream-I1-Full/discussions/19#67ff4f0cdf848a9f25c82e95

Would appreciate any feedback on this update.

src/diffusers/models/transformers/transformer_hidream_image.py
```diff
@@ -686,46 +686,107 @@ def unpatchify(self, x: torch.Tensor, img_sizes: List[Tuple[int, int]], is_train
...
+    def patchify(self, hidden_states):
+        """
+        Convert input tensor to patches and create corresponding image IDs and masks.
+
+        Args:
+            hidden_states: Input tensor of shape [B, C, H, W]
+        Returns:
+            hidden_states: Patchified hidden states hidden_states_masks: Masks for hidden states img_sizes: Image sizes
+            img_ids: Image token positional IDs
+        """
```
yiyixuxu, line 698:
```suggestion
```

yiyixuxu:
@ShuyUSTC oh thanks, we can update our doc examples once you updated the repo!

sayakpaul:
Thanks a lot!

ShuyUSTC:
> @ShuyUSTC oh thanks, we can update our doc examples once you updated the repo!

Thanks very much!

vladmandic:
just imo, updating default scheduler to something that doesn't exist upstream in released diffusers is not the way to go as anyone that had working hidream yesterday will have immediate failure today without any clear explanation that they need to update to latest diffusers main codebase.

bghira:
ah that explains why all of my users are like "why do you keep breaking HiDream", and I have to let them know, "that is not me".

ShuyUSTC:
@vladmandic @bghira Thanks for the feedback! I'd like to clarify a few points regarding the scheduler update:

1. This change does not alter the inference behavior of any **HiDream-I1** models. The update is purely aimed at improving the pipeline's initialization experience — users can still manually initialize the scheduler externally, just like before. This update simply makes the default usage more streamlined for new users.

2. The default schedulers for **HiDream-I1-Fast** and **HiDream-I1-Dev** have always been `FlowMatchLCMScheduler`. The reason we hadn't yet provided examples for them in the diffusers docs was that this scheduler wasn't available in the upstream diffusers library until recently.

3. The `FlashFlowMatchEulerDiscreteScheduler` used in our [GitHub repository](https://github.com/HiDream-ai/HiDream-I1/blob/main/inference.py) is just a local implementation of FlowMatchLCMScheduler. Now that it's officially available in diffusers, we're aligning everything accordingly.

We understand the concern around breaking changes — but in this case, there’s no functional regression. Existing workflows that use externally defined schedulers will continue to work as expected.

Hopefully this clears up the confusion — and saves us all a few user support messages 😅

bghira:
no, it literally complains when starting up that the scheduler is not available in Diffusers because the model config was updated in the repository, which is where the path to the scheduler is used from.

people were using UniPC for flow matching and other options before, which now error.

ShuyUSTC:
@bghira which model do you use? For **HiDream-I1-Dev** and **HiDream-I1-Fast**, the original scheduler in Hugging Face model repository, i.e., `FlowMatchEulerDiscreteScheduler`, is not the official scheduler. Our original inference code need to create scheduler manually. It is recommanded to use `FlowMatchLCMScheduler` recently merged into Diffusers main branch, which can be created within `HiDreamImagePipeline` automated now.

yiyixuxu:
@bghira 
i added a short deprecation cycle for all the changes in this PR (until next release)
can let you test before merge though

src/diffusers/models/transformers/transformer_hidream_image.py
```diff
@@ -745,42 +805,18 @@ def forward(
...
-                dtype=hidden_states.dtype,
-                device=hidden_states.device,
-            )
-            hidden_states = hidden_states.reshape(B, C, pH, patch_size, pW, patch_size)
-            hidden_states = hidden_states.permute(0, 1, 2, 4, 3, 5)
-            hidden_states = hidden_states.reshape(B, C, pH * pW, patch_size * patch_size)
-            out[:, :, 0 : pH * pW] = hidden_states
-            hidden_states = out
+        # Patchify the input
+        hidden_states, hidden_states_masks, img_sizes, img_ids = self.patchify(hidden_states)
```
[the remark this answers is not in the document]
yiyixuxu: thanks! I made it so that the `patchify` step can be skipped, i.e. for samples with different aspect ratios, user would have to prepare patchified `hidden_states`, `hidden_states_mask`, `img_sizes` and `img_ids` outside of the model and pass them as inputs, I think it is easier this way:
1.  this library is mainly for inference and fine tune, I think training with different aspect ratios is most for pre-trainning, no? 
2. they would need to prepare `latents` differently, we currently do not support that anyway

bghira:
finetuning on multi-aspect bucketing is fine and common; but i'm confused what's meant by mixing shapes. making use of nested_tensor to purposely mix aspect bucketed data in a single batch?

yiyixuxu:
@ShuyUSTC @YehLi 
I found out that without CFG the it's not able to generate for squared images, but it's able to do so for non-squared images (not a good job).  is this expecterd? 


```python
import torch
from transformers import PreTrainedTokenizerFast, LlamaForCausalLM
from diffusers import (
    HiDreamImagePipeline,
    HiDreamImageTransformer2DModel,
)

device = "cuda:3"
testing_branch = "main"

llama_repo = "unsloth/Meta-Llama-3.1-8B-Instruct"
# llama_repo = "meta-llama/Meta-Llama-3.1-8B-Instruct"

tokenizer_4 = PreTrainedTokenizerFast.from_pretrained(
    llama_repo,
)

text_encoder_4 = LlamaForCausalLM.from_pretrained(
    llama_repo,
    output_hidden_states=True,
    output_attentions=True,
    torch_dtype=torch.bfloat16,
)

transformer = HiDreamImageTransformer2DModel.from_pretrained(
    "HiDream-ai/HiDream-I1-Full", subfolder="transformer", torch_dtype=torch.bfloat16
)


pipe = HiDreamImagePipeline.from_pretrained(
    "HiDream-ai/HiDream-I1-Full",
    tokenizer_4=tokenizer_4,
    text_encoder_4=text_encoder_4,
    transformer=transformer,
    torch_dtype=torch.bfloat16,
)



test_guidance = [0.0]
test_negative_prompt = [None]
test_width = [1024, 768]
num_inference_steps = 50

for guidance in test_guidance:
    for negative_prompt in test_negative_prompt:
        for width in test_width:
            print(f" --------------------------------------------------------------")
            print(f"testing guidance: {guidance}, negative_prompt: {negative_prompt}, width: {width}")
            image = pipe(
                'A cat holding a sign that says "Hi-Dreams.ai".',
                height=1024,
                width=width,
                guidance_scale=guidance,
                generator=torch.Generator("cuda").manual_seed(0),
            ).images[0]

            image.save(f"yiyi_test_2_out_{testing_branch}_{guidance}_{negative_prompt}_{width}.png")
            print(f"saved image to yiyi_test_2_out_{testing_branch}_{guidance}_{negative_prompt}_{width}.png")
            print(f" --------------------------------------------------------------")

```

 1024 x 1024
yiyi_test_2_out_main_0 0_None_1024

1024 x 768
yiyi_test_2_out_main_0 0_None_768

ShuyUSTC:
@yiyixuxu We recommend enabling **cfg** in **HiDream-I1-Full** model. The quality cannot be guaranteed without using cfg.

src/diffusers/models/transformers/transformer_hidream_image.py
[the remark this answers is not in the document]
YehLi: 1. During pre-trainning, the input images are not cropped by default. Thus, it is recommended that users do not apply any cropping for images and prepare patchified hidden_states, img_sizes and img_ids outside of the model during fine-tuning.  It is fine to finetune the model on multi-aspect bucketing like SDXL. The main purpose for img_sizes and img_ids is to provide users more options if they do not want to crop the images.
2. They can prepare latents on dataloader with shape (B, C, H*W / patch_size / patch_size, patch_size*patch_size) and then pad them into the same length (B, C, S, patch_size*patch_size)

src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py
```diff
@@ -332,120 +330,177 @@ def encode_prompt(
...
+            )
 
-        return prompt_embeds, pooled_prompt_embeds
+        return (
+            prompt_embeds_t5,
+            prompt_embeds_llama3,
+            negative_prompt_embeds_t5,
+            negative_prompt_embeds_llama3,
+            pooled_prompt_embeds,
+            negative_pooled_prompt_embeds,
```
yiyixuxu, line 502:
```suggestion
            prompt_embeds_t5,
            negative_prompt_embeds_t5,
            prompt_embeds_llama3,
            negative_prompt_embeds_llama3,
            pooled_prompt_embeds,
            negative_pooled_prompt_embeds,
```

src/diffusers/pipelines/hidream_image/pipeline_hidream_image.py
```diff
@@ -667,17 +869,18 @@ def __call__(
...
 
+        # 3. Encode prompt
         lora_scale = self.attention_kwargs.get("scale", None) if self.attention_kwargs is not None else None
         (
-            prompt_embeds,
-            negative_prompt_embeds,
+            prompt_embeds_t5,
+            prompt_embeds_llama3,
+            negative_prompt_embeds_t5,
+            negative_prompt_embeds_llama3,
```
yiyixuxu, line 883:
```suggestion
            prompt_embeds_t5,
            negative_prompt_embeds_t5,
            prompt_embeds_llama3,
            negative_prompt_embeds_llama3,
```

src/diffusers/models/transformers/transformer_hidream_image.py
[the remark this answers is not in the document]
yiyixuxu: cc @a-r-r-o-w @sayakpaul @linoytsaban here in case we want to try this out for training script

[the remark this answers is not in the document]
yiyixuxu: thanks @YehLi !

vladmandic:
deprecation change completely breaks hidream on diffusers==main
```log
│ /home/vlado/dev/sdnext/venv/lib/python3.12/site-packages/diffusers/models/transformers/transformer_hidream_image.py:794 in forward                                                                                                                                                                                                                                                                                               │
│                                                                                                                                                                                                                                                                                                                                                                                                                                  │
│   793 │   │   │   deprecation_message = "The `encoder_hidden_states` argument is deprecated. Please use `encoder_hidden_states_t5` and `encoder_hidden_states_llama3` instead."                                                                                                                                                                                                                                                  │
│ ❱ 794 │   │   │   deprecate("encoder_hidden_states", "0.34.0", deprecation_message)                                                                                                                                                                                                                                                                                                                                              │
│   795 │   │   │   encoder_hidden_states_t5 = encoder_hidden_states[0]                                                                                                                                                                                                                                                                                                                                                            │
│                                                                                                                                                                                                                                                                                                                                                                                                                                  │
│ /home/vlado/dev/sdnext/venv/lib/python3.12/site-packages/diffusers/utils/deprecation_utils.py:18 in deprecate                                                                                                                                                                                                                                                                                                                    │
│                                                                                                                                                                                                                                                                                                                                                                                                                                  │
│   17 │   │   if version.parse(version.parse(__version__).base_version) >= version.parse(version_name):                                                                                                                                                                                                                                                                                                                           │
│ ❱ 18 │   │   │   raise ValueError(                                                                                                                                                                                                                                                                                                                                                                                               │
│   19 │   │   │   │   f"The deprecation tuple {(attribute, version_name, message)} should be removed since diffusers'"     
```

PR #11305, opened 2025-04-13
Fix incorrect tile_latent_min_width calculations

kuantuna:
# What does this PR do?
This PR fixes the same issue that was recently addressed in AutoencoderKLMochi, where `tile_latent_min_width` was incorrectly calculated using `tile_sample_stride_width` instead of `tile_sample_min_width`.

The same mistake was found in the following autoencoders:

- [AutoencoderKLMagvit](https://github.com/huggingface/diffusers/blob/bc261058eee74aa6f574e840af9295b78b379f51/src/diffusers/models/autoencoders/autoencoder_kl_magvit.py#L890)
- [AutoencoderKLLTXVideo](https://github.com/huggingface/diffusers/blob/bc261058eee74aa6f574e840af9295b78b379f51/src/diffusers/models/autoencoders/autoencoder_kl_ltx.py#L1288)
- [AutoencoderKLHunyuanVideo](https://github.com/huggingface/diffusers/blob/bc261058eee74aa6f574e840af9295b78b379f51/src/diffusers/models/autoencoders/autoencoder_kl_hunyuan_video.py#L832)

This update ensures consistent and correct behavior across all affected models when VAE tiling is enabled.

Fixes: #11291. 

## Who can review?

@a-r-r-o-w


a-r-r-o-w (approved):
Tysm for the awesome catch! 👁️

PR #11308, opened 2025-04-14
make KandinskyV22PipelineInpaintCombinedFastTests::test_float16_inference pass on XPU

yao-matrix:
loose `expected_max_diff` from `5e-1` to `8e-1` to make KandinskyV22PipelineInpaintCombinedFastTests::test_float16_inference pass on XPU.


yao-matrix:
@hlky , pls help review, thx

hlky (approved):
Thanks @yao-matrix. This pipeline is not commonly used and already requires high tolerance so no issue increasing it further. If in the future we find some larger difference in required tolerances like 1e-4 on one backend and 1e-3 on another, then we could consider something similar to the expected slice changes where the `expected_max_diff` is set per backend type.

yao-matrix:
sure, will co-work w/ you to make it once needed

PR #11313, opened 2025-04-14
make `KolorsPipelineFastTests::test_inference_batch_single_identical` pass on XPU

faaany:
## What does this PR do?
Since there is no difference can be seen from the images generated with single and batched inputs, I adjusted the expected maximum difference from 5e-4 to 5e-3 to make the test pass on both XPU and CUDA.

pls help review it. Thanks! @hlky

### DIFF
--- a/tests/pipelines/kolors/test_kolors.py
+++ b/tests/pipelines/kolors/test_kolors.py
@@ -145,4 +145,4 @@ class KolorsPipelineFastTests(PipelineTesterMixin, unittest.TestCase):
         super().test_save_load_float16(expected_max_diff=2e-1)
 
     def test_inference_batch_single_identical(self):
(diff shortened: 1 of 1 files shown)


hlky (approved):
Thanks @faaany. Same comment as in https://github.com/huggingface/diffusers/pull/11308#pullrequestreview-2763295924 but I will add that having `expected_max_diff` per backend is certainly not a priority and the tolerances mentioned in that comment are not the best example of when this should be a priority. Essentially we wouldn't want to increase `expected_max_diff` to the point where changes affecting precision may not be detectable, however there haven't been many of these changes required and only for specific tests, `test_inference_batch_single_identical` in this case, so it would likely be picked up in another test, and we generally confirm outputs are visually the same when making any changes to a pipeline/model anyway.

PR #11317, opened 2025-04-14
[HiDream] code example

linoytsaban:


hlky (approved):
Thanks @linoytsaban 

For context, this was previously required as `transformer` was not in the `model_index.json` so doing `from_pretrained` would not include it. The model authors have since updated `model_index.json`.

asomoza:
WDYT about promoting the reuse of the pipeline for dev and fast? AFAIK you just need to switch the transformer model and the scheduler, this would prevent the redownload of the T5 which is not small.

Also we can promote the `AutoModel`

linoytsaban:
@asomoza sorry missed this. That's a good idea!

PR #11318, opened 2025-04-14
import for FlowMatchLCMScheduler

asomoza:
# What does this PR do?

use 

```python
from diffusers import FlowMatchLCMScheduler
```

instead of 

```python
from diffusers.schedulers.scheduling_flow_match_lcm import FlowMatchLCMScheduler
```


yiyixuxu (approved):
thanks!

PR #11322, opened 2025-04-15
[Tests] Enable more general testing for `torch.compile()` with LoRA hotswapping

sayakpaul:
# What does this PR do?

@BenjaminBossan shipped a great feature in https://github.com/huggingface/diffusers/pull/9453/. We decided to add tests only for the `Unet2DConditionModel`. However, we support LoRA hotswapping with `torch.compile()` for all models subclassed with `PeftAdapterMixin`. 

This PR refactors the `TestLoraHotSwappingForModel` class to make it more generally available to the model test classes. I believe this is nice to have as this feature is for power users and for users who use `diffusers` in production environments. 

Currently, this PR adds `LoraHotSwappingForModelTesterMixin` (revised name) to the test classes of `Unet2DConditionModel` and `FluxTransformer2DModel` as I think those are the most popular ones having significant usage. 

I think we can open the rest of the models for community contributions. 

@yiyixuxu FYI.


tests/models/test_modeling_common.py
```diff
@@ -1877,58 +1859,86 @@ def test_hotswapping_compiled_model_linear(self, rank0, rank1):
     @parameterized.expand([(11, 11), (7, 13), (13, 7)])  # important to test small to large and vice versa
     def test_hotswapping_compiled_model_conv2d(self, rank0, rank1):
         # It's important to add this context to raise an error on recompilation
+        if "unet" not in self.model_class.__name__.lower():
```
BenjaminBossan, line 1862: The comment above this line should move below this line, as it relates to the context manager.

tests/models/test_modeling_common.py
```diff
@@ -1877,58 +1859,86 @@ def test_hotswapping_compiled_model_linear(self, rank0, rank1):
...
+            return
+
         target_modules = ["conv", "conv1", "conv2"]
         with torch._dynamo.config.patch(error_on_recompile=True):
             self.check_model_hotswap(do_compile=True, rank0=rank0, rank1=rank1, target_modules0=target_modules)
 
     @parameterized.expand([(11, 11), (7, 13), (13, 7)])  # important to test small to large and vice versa
     def test_hotswapping_compiled_model_both_linear_and_conv2d(self, rank0, rank1):
         # It's important to add this context to raise an error on recompilation
+        if "unet" not in self.model_class.__name__.lower():
```
BenjaminBossan, line 1872: Same argument

tests/models/test_modeling_common.py
```diff
@@ -1877,58 +1859,86 @@ def test_hotswapping_compiled_model_linear(self, rank0, rank1):
...
         with torch._dynamo.config.patch(error_on_recompile=True):
             self.check_model_hotswap(do_compile=True, rank0=rank0, rank1=rank1, target_modules0=target_modules)
 
+    @parameterized.expand([(11, 11), (7, 13), (13, 7)])  # important to test small to large and vice versa
+    def test_hotswapping_compiled_model_both_linear_and_other(self, rank0, rank1):
+        # In `test_hotswapping_compiled_model_both_linear_and_conv2d()`, we check if we can do hotswapping
+        # with `torch.compile()` for models that have both linear and conv layers. In this test, we check
+        # if we can target a linear layer from the transformer blocks and another linear layer from non-attention
+        # block.
+        # It's important to add this context to raise an error on recompilation
```
BenjaminBossan, line 1885: Same argument

tests/models/test_modeling_common.py
```diff
@@ -1877,58 +1859,86 @@ def test_hotswapping_compiled_model_linear(self, rank0, rank1):
...
+        target_modules = ["to_q"]
+        init_dict, _ = self.prepare_init_args_and_inputs_for_common()
+        model = self.model_class(**init_dict)
+
+        target_modules.append(self.get_linear_module_name_other_than_attn(model))
+        del model
```
BenjaminBossan, line 1891: It looks a bit wasteful to create the model only to immediately discard it again, but it's probably not a big deal in overall runtime.

BenjaminBossan (approved):
Thanks for refactoring the tests to make them modular and align them better with the diffusers testing style. I don't have anything to add, just a few nits. I was not sure if the tests are run in the CI so I ran them locally and they all passed.

sayakpaul:
> I was not sure if the tests are run in the CI so I ran them locally and they all passed (they took 2 min 7 sec).

Just added it to our nightly CI: https://github.com/huggingface/diffusers/pull/11322/commits/4b11ab2901c8134594c7c50e5719d8f74adf2ca5.

tests/models/test_modeling_common.py
sayakpaul: Yeah the models are small. I think this should be okay.

DN6:
Just a question. Why not place these under the LoRA tests in the nightlys? The compile image isn't actually needed anymore. It was only a workaround for issues we were facing with Python 3.8 and torch.compile. Since the base images now using 3.10, we should be okay to have them run there.

sayakpaul:
@DN6 good point. Done.

sayakpaul:
@DN6 a gentle ping.

PR #11324, opened 2025-04-15
[LoRA] parse metadata from LoRA and save metadata

sayakpaul:
# What does this PR do?

I know we have revisited this over and over again but this is becoming increasingly important. So, we should consider this on priority. 

@a-r-r-o-w brought this issue to me while we were debugging something in Wan LoRA training. So, I started by just modifying the Wan LoRA loader (eventually, the changes will be propagated to other loaders too). Aryan, could you check if this change fixes the issue we were facing? 

Admittedly, the PR can be cleaned and modularized a bit but I wanted to get something up quickly to get feedback on the direction. 

## TODOs

- [x] Docs
- [ ] Modularize where possible (prefer to do it in a different PR)
- [ ] Follow up to show how the training scripts should be changed to faciliate this change.
- [x] Propagate to other pipelines


BenjaminBossan (commented):
I remember there were debates in the past, but I don't remember what the arguments for and against were. Could you quickly sketch why this is becoming increasingly important?

Implementation-wise, it generally LGTM. I was a bit confused about the `lora_adapter_config` key vs the `lora_metadata` key, what is the difference?

sayakpaul:
> I remember there were debates in the past, but I don't remember what the arguments for and against were. Could you quickly sketch why this is becoming increasingly important?

Folks are using different ranks and alphas and finding it to be very effective in practice.

> I was a bit confused about the lora_adapter_config key vs the lora_metadata key, what is the difference?

I will rename it to `lora_adapter_metadata_key`.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -5192,6 +5195,7 @@ def load_lora_weights(
             raise ValueError("PEFT backend is required for this method.")
 
         low_cpu_mem_usage = kwargs.pop("low_cpu_mem_usage", _LOW_CPU_MEM_USAGE_DEFAULT_LORA)
+        load_with_metdata = kwargs.get("load_with_metdata", False)
```
a-r-r-o-w, line 5198:
```suggestion
        load_with_metadata = kwargs.get("load_with_metadata", False)
```

a-r-r-o-w (commented):
Thanks for reviving this conversation again @sayakpaul!

I see some TODOs marked so will take a better look once you let us know when the PR is ready, but changes look good to me and seem sensible from user end. Maybe it might make sense to allow user to also specify the metadata key, or us automatically detect some common names (are there any that you know of?)

@BenjaminBossan To provide some more context, my discussion with Sayak involved training "Control" LoRAs.

<details>
<summary> Expand for mine and Sayak's conversation </summary>

<br />

Hello curious 🐱 In short:

- Expand the input projection layer's (patch embedding) input channels. Say we want to train with Canny conditioning on a 64-input channel model (Flux, for example), we'd expand the input projection linear to 128 channels with zeroed weights.
- For the input latent to match the expanded linear proj, you concatenate the latents with the latents of the control condition.
- In the "normal" latent stream (base layer), since we expanded with zeroed out weights, the control condition is effectively not added.
- In the "lora" latent stream, the noisy latents and control latents are projected to the expected inner dim of the model. The lora latent stream outputs are added to the normal latent stream, effectively adding some amount of conditioning information from the new Canny condition
- Now, in Flux, the linear projection goes from `64 channels -> 4096 channels`. With the new conditioning related expansion, it is `128 channels -> 4096 channels`. This is the "normal" latent stream.
- In the "lora" latent stream, you have `128 channels -> "rank" channels -> 4096 channels`. If the rank for the input projection layer is less than 128 channels, you might lose some information from the control conditioning (since the lora layer then acts as a 1-layer autoencoder).
- I've noticed from a limited set of experiments that training with a low input projection rank (does not matter what the rank for rest of the network is) almost always produces worse results and contains random noise in the generations.
- Due to this, in finetrainers, I made it so that the input projection layer has the same rank as `inner_dim`. So, effectively, in the "lora" latent stream, you have `128 channels -> 4096 channels -> 4096 channels`. This results in much faster convergence (conclusion made from a very limited set of experiments and it intuitively made sense to me if you compare this low-rank version against a full-rank Control model).

Here are the relevant pieces of code for a quick look:
- Expanding input channels: [this](https://github.com/a-r-r-o-w/finetrainers/blob/15c490e1be51617da84751506f84d2a4f773201a/finetrainers/models/utils.py#L66)
- Creating the LoRA adapter with custom rank for input projection layer vs rest of the network: [this](https://github.com/a-r-r-o-w/finetrainers/blob/15c490e1be51617da84751506f84d2a4f773201a/finetrainers/trainer/control_trainer/trainer.py#L157)
- Channel-wise input concatenation with control latents: [this](https://github.com/a-r-r-o-w/finetrainers/blob/15c490e1be51617da84751506f84d2a4f773201a/finetrainers/models/cogview4/control_specification.py#L321)

<table align=center>
<tr>
  <th> Convert to impasto style painting </th>
</tr>
<tr>
  <td align=center></td>
</tr>
<tr>
  <th>Canny to Image</th>
</tr>
<tr>
  <td align=center></td>
</tr>
</table>
</details>

These results come with just 2500 training steps on a `rank=64` lora (except `rank=4096` for the input projection).

For this control lora training settings, every `lora_alpha` is the same as `rank`. So, the `lora_alpha` for the input projection is `4096`, while for remaining layers is `64`. The size of LoRA weights is between 300-1000 MB in most cases, which is much less compared to a full control conditioning model.

The problem comes when trying to load such a LoRA with diffusers. Since we only know the ranks, but not the lora_alpha, diffusers config uses the same `lora_alpha` as the inferred `rank` (which is `4096` since the input projection layer is the first lora layer in the state dict too). As you can imagine, setting an alpha to 4096 on all layers (even the ones that originally had `rank = lora_alpha = 64` will result in random noise.

This is a more general problem because a lot of loras on CivitAI and from different trainers are trained with different alpha configurations from the rank. So making the assumption that `lora_alpha=rank` (which is our current default behaviour) is incorrect, and having this metadata information will be really helpful.

In order to solve the problem of being able train and run validation in `finetrainers`, we just directly serialize the lora adapter config directly into the safetensors metadata. See [this](https://github.com/a-r-r-o-w/finetrainers/blob/15c490e1be51617da84751506f84d2a4f773201a/finetrainers/trainer/control_trainer/trainer.py#L358) and [this](https://github.com/a-r-r-o-w/finetrainers/blob/15c490e1be51617da84751506f84d2a4f773201a/finetrainers/utils/serialization.py#L6), and we use a custom function to load the lora weights. _Really big kudos to whoever implemented the `save_function` allowing custom ones to be provided!_

sayakpaul:
> I see some TODOs marked so will take a better look once you let us know when the PR is ready, but changes look good to me and seem sensible from user end. Maybe it might make sense to allow user to also specify the metadata key, or us automatically detect some common names (are there any that you know of?)

Well, I wanted to check if the changes help solve the issue of `finetrainers`. If possible could you let me know about that? 

Regarding detecting metadata in non-diffusers checkpoints, we already either infer alphas or directly scale the LoRA weights. So, that should already be covered.

BenjaminBossan:
Thanks for explaining this further @a-r-r-o-w 

> diffusers config uses the same lora_alpha as the inferred rank (which is 4096 since the input projection layer is the first lora layer in the state dict too)

There could be better heuristics, like using the most common rank, but I imagine changing that now would be backwards incompatible.

> This is a more general problem because a lot of loras on CivitAI and from different trainers are trained with different alpha configurations from the rank. So making the assumption that `lora_alpha=rank` (which is our current default behaviour) is incorrect, and having this metadata information will be really helpful.

Any idea who other libraries/apps deal with this lack of info? I thought there was a bit of a standard to multiply the scale into the weights before saving, so that the assumption that `scale==1` holds true at the end, but I'm not knowledgeable enough about common practices in this area.

sayakpaul:
If no objections, I will button up the PR and make it generally available to the other loaders. But LMK if you have anything to share or have any concerns over this direction. 

@BenjaminBossan @a-r-r-o-w

sayakpaul:
>  just wondering if anything needs to be discussed (again) with the safetensors folks.

Well, previously there was no problem with them IIRC. It was about time and the complexity. But not we have enough evidence to ship it maintaining the single-file format LoRA checkpoints.

a-r-r-o-w:
@sayakpaul I've verified it to work, thanks for working on this. My understanding was that the lora initialization config would also be automatically saved into the exported safetensors as metadata, but it needs to be passed manually to `save_lora_weights`. If not passed, it leads to the following error ([this](https://github.com/a-r-r-o-w/finetrainers/tree/test/diffusers-metadata-info) example uses a different rank/alpha for the patch layer as described above):

```
RuntimeError: Error(s) in loading state_dict for WanTransformer3DModel:
        size mismatch for patch_embedding.lora_A.default_0.weight: copying a param with shape torch.Size([1536, 32, 1, 2, 2]) from checkpoint, the shape in current model is torch.Size([128, 32, 1, 2, 2]).
        size mismatch for patch_embedding.lora_B.default_0.weight: copying a param with shape torch.Size([1536, 1536, 1, 1, 1]) from checkpoint, the shape in current model is torch.Size([1536, 128, 1, 1, 1]).
```

I think it might be good if we serialize this info automatically in save_lora_weights, if you think there's an easy way to do it.

sayakpaul:
Thanks for trying! Well, we're serializing the the entire `LoraConfig()` when `save_lora_weights()` is called with a non-none `transformer_lora_adapter_metadata`:
https://github.com/huggingface/diffusers/blob/d390d4de0b8d16b9be51ca79b35ee6fb7d8b722c/src/diffusers/loaders/lora_pipeline.py#L5308

The test below also confirms that:
https://github.com/huggingface/diffusers/blob/d390d4de0b8d16b9be51ca79b35ee6fb7d8b722c/tests/lora/test_lora_layers_wan.py#L143

If you could provide a minimal reproducer for the issue I would be happy to look into it on priority.

a-r-r-o-w:
Yes, what you mentioned works as expected: we need to pass `transformer_lora_adapter_metadata` ourselves with the lora config when calling `save_lora_weights`. This is already quite convenient. I was asking if it would be possible/make sense to do this behind-the-scenes without needing to pass the metadata explicitly (since the model already "knows" what lora layers it has and what it was initialized with). Or is it not really ideal?

sayakpaul:
>  I was asking if it would be possible/make sense to do this behind-the-scenes without needing to pass the metadata explicitly (since the model already "knows" what lora layers it has and what it was initialized with). Or is it not really ideal?

Ah I see. I might have an idea. Will update soon. Thanks again for trying it out quickly.

bghira:
why can't we add alpha to the state dict like kohya trainer does?

sayakpaul:
> why can't we add alpha to the state dict like kohya trainer does?

The code path for `diffusers` state dict would have to be modified a lot more + the current solution being proposed in this PR is easier than those changes.

bghira:
so because "its harder" we will end up with a solution that requires other tools like comfyUI to implement special support to read these attributes and set them? 

i expected we would be adding compatibility points and not adding more tech debt. please reconsider just adding alpha to the state dict or leave the work to the community if it is too hard?

this kinda thing really takes an eternity to be implemented by tools like Swarm and Forge and ComfyUI while the alpha attribute in dict key would Just Work even with AUTOMATIC1111. please tell me you see the value in this..

bghira:
also it wasnt arrow who brought this up first, it was me, because simpletuner has the option to set rank differently to alpha, which peft has always supported but diffusers does not. setting alpha to 1 always would allow learning rates to stay the same across every rank. the initial requeat was just to write the alpha into the state dict key.

this proposal is heavy handed and requires too much effort from everybody.

sayakpaul:
> so because "its harder" we will end up with a solution that requires other tools like comfyUI to implement special support to read these attributes and set them?
> 
> i expected we would be adding compatibility points and not adding more tech debt. please reconsider just adding alpha to the state dict or leave the work to the community if it is too hard?
> 
> this kinda thing really takes an eternity to be implemented by tools like Swarm and Forge and ComfyUI while the alpha attribute in dict key would Just Work even with AUTOMATIC1111. please tell me you see the value in this..

We value other formats and hence we support most of them directly within the library.

At the same time, I also think it makes sense to be lenient towards our libraries and think about long-term maintainability. With the current proposal, we can serialize additional useful things like `rank_pattern`, `alpha_pattern`, etc. (supported by `LoraConfig`) which are becoming important to tweak from what I understand. 

I am not sure that for `diffusers`-format LoRAs, it's sufficient to just write `alpha` keys and the other community tools would just work correctly with them.

> also it wasnt arrow who brought this up first, it was me, because simpletuner has the option to set rank differently to alpha, which peft has always supported but diffusers does not.

Corrected my issue description but please note what my description starts with:

> **_I know we have revisited this over and over again_**

> setting alpha to 1

We set `alpha=rank` in our official scripts currently and also for `diffusers`-format state dicts. 

---

I would like to take a second opinion from @DN6, especially for maintainability. To summarize, the issue is that we just serialize the LoRA state dict in `diffusers` but no other info from `LoraConfig` is provided in the serialized state dict when training. So, if someone trains with say, an `alpha` that doesn't equal `rank`, we won't be inferring it from the state dict as there's no info in the state dict about it. This will make inference wrong.

This PR proposes to serialize the `LoraConfig` as a part of the metadata of the state dict as `safetensors` allows for that. This will allow us to parse the entire `LoraConfig` from the state dict easily. If it were only for `alpha` values, we could have simply serialized them as a part of the state dict (not metadata) but with the current PR proposal we have the flexibility to leverage full specturm of `LoraConfig` (additional things like `alpha_pattern`, `rank_pattern`, etc.).

sayakpaul:
We can still just serialize `alpha` keys in the `diffusers`-format LoRA state dicts (in retrospect, I think it will be easier to do actually), considering what's passed to the `LoraConfig`. Then make changes to the loader as needed. But before that I just want to be sure this is the design we want to have. So, will let @DN6 also comment here.

bghira:
i agree the config is needed for other more intricate patterns and whatever Diffusers does there can even become the standard depending on how @kohya-ss and @comfyanonymous feel about it, i'd rather something "common" happen (cc @Nerogar ) but it is a hard path to go down to try and get everybody on the same page before things start happening/becoming committed to.

Nerogar:
I feel like there is some background information I'm missing, I haven't used the diffusers LoRA implementation yet, so don't know about the state dict and metadata format.

That said, this sounds like you are trying to create yet another format where the alpha is stored in the model metadata instead of the state dict. I've never seen a lora where this was done. All files I've seen so far store the alpha in the state dict with the suffix `.alpha`.

Over the last few months I was talking mostly to kohya about a more standardized LoRA format. We've now agreed on something similar to the currently supported format. Just with a few small adjustments. I've already written conversion code for this format, and you can find example files [here](https://huggingface.co/Nerogar/omi_lora/tree/main).

If you are now changing the format for your LoRA output, I strongly suggest following that format instead of re-inventing something well established again, just because it's easier to implement.

sayakpaul:
For the record, the structure (keys, for example) of the state dict won't change. It will just be populated with metadata.

a-r-r-o-w:
To provide more context, this is not a new format and no existing non-diffusers code for lora loading needs to be changed. All this PR does is serialize a simple string dict into the safetensors file format ([docs](https://huggingface.co/docs/safetensors/main/en/api/torch#safetensors.torch.save_model.metadata)). The state dict formats remain unaffected.

This may actually help improve loading diffusers implementation state dicts more easily because everything needed for loading the peft-related parts becomes part of the file:
- rank, alpha
- rank_pattern (for any layers that have a different rank than `rank`)
- alpha_pattern (for any layers that have a different alpha than `alpha`)
- info like weight init and other configs for bookkeeping (basically what's documented [here](https://huggingface.co/docs/peft/en/package_reference/lora#peft.LoraConfig))

Knowing this info and using the native model implementations, the lora loading code becomes:
```python
with safe_open(model_file, framework="pt") as f:
   metadata = json.loads(f.metadata()["lora_metadata"])
   state_dict = ...
config = LoraConfig(**metadata)
adapter_name = "my-awesome-lora"
inject_adapter_in_model(config, model, adapter_name=adapter_name, low_cpu_mem_usage=True)
set_peft_model_state_dict(model, state_dict, adapter_name=adapter_name, ignore_mismatched_sizes=False, low_cpu_mem_usage=True)
```

No changes are needed in any existing code that supports loading diffusers loras, since the state dict remains completely unaltered. However, if another line or two are added to read the metadata and create the peft-init config based on that, the exact ranks/alphas used in training will be used.

bghira:
the rank can be different from the parameter shape?

bghira:
> OneTrainer always assumes alpha=1 as a default value.

can't do that with Diffusers models. they are `alpha=rank`.

a-r-r-o-w:
> the rank can be different from the parameter shape?

Not sure what you mean. What I mean is that different layers can have different ranks and alphas (see example below). We will now store this information as part of the safetensors metadata (which one may choose to use or not).

Supporting existing compatibility with storing the alphas in the state dict is a separate issue and unrelated to what's happening here (we can look into exporting loras that way if there's an ask)

Example model:

```
proj_in: in_features=4, out_features=64
layer_1: in_features=64, out_features=64
proj_out: in_features=64, out_features=4
```

Now, let's assume we want a lora with `rank=X`, `alpha=X`, `rank_pattern={"layer_1": Y}`, `alpha_pattern={"layer_1": Z}`

The equivalent peft init config is:

```
config = LoraConfig(rank=X, lora_alpha=X, rank_pattern={"layer_1": Y}, alpha_pattern={"layer_2": Z})
```

```
proj_in: lora_A (in_features=4, out_features=X), alpha=X, lora_B (in_features=X, out_features=64)
layer_1: lora_A (in_features=64, out_features=Y), alpha=Z, lora_B(in_features=Y, out_features=64)
proj_out: lora_A (in_features=64, out_features=X), alpha=X, lora_B (in_features=X, out_features=4)
```

The extra metadata that will be stored in the safetensors metadata (emphasizing again that it's not the state dict, so no existing code needs to be modified) is the serialized `config`

Does this make sense?

src/diffusers/loaders/lora_base.py
```diff
@@ -882,16 +888,31 @@ def write_lora_layers(
         weight_name: str,
         save_function: Callable,
         safe_serialization: bool,
+        lora_adapter_metadata: dict = None,
```
BenjaminBossan, line 891:
```suggestion
        lora_adapter_metadata: Optional[dict] = None,
```

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -1703,6 +1708,7 @@ def save_lora_weights(
             weight_name=weight_name,
             save_function=save_function,
             safe_serialization=safe_serialization,
+            lora_adapter_metadata=lora_adapter_metadata,
```
BenjaminBossan, line 1711: There are some instances of `cls.write_lora_layers` in this file where this new argument has not been propagated, I'm not sure if it's intended or not.

src/diffusers/loaders/lora_base.py
```diff
@@ -882,16 +888,31 @@ def write_lora_layers(
...
+        if lora_adapter_metadata and not isinstance(lora_adapter_metadata, dict):
+            raise ValueError("`lora_adapter_metadata` must be of type `dict`.")
+
         if save_function is None:
             if safe_serialization:
 
                 def save_function(weights, filename):
-                    return safetensors.torch.save_file(weights, filename, metadata={"format": "pt"})
+                    # We need to be able to serialize the NoneTypes too, otherwise we run into
+                    # 'NoneType' object cannot be converted to 'PyString'
```
BenjaminBossan, line 907: To me, it's not clear what this comment is referring too, as there seems to be no special treatment for None in the code below.

src/diffusers/loaders/peft.py
```diff
@@ -230,8 +238,12 @@ def load_lora_adapter(
             raise ValueError("`network_alphas` cannot be None when `prefix` is None.")
 
         if prefix is not None:
+            metadata = state_dict.pop(LORA_ADAPTER_METADATA_KEY, None)
             state_dict = {k[len(f"{prefix}.") :]: v for k, v in state_dict.items() if k.startswith(f"{prefix}.")}
 
+            if metadata is not None:
+                state_dict[LORA_ADAPTER_METADATA_KEY] = metadata
+
```
BenjaminBossan, line 246: Would it not be easier to add `and if k != LORA_ADAPTER_METADATA_KEY` to the dict comprehension instead of popping and re-adding the meta data?

src/diffusers/utils/peft_utils.py
```diff
@@ -147,7 +147,22 @@ def unscale_lora_layers(model, weight: Optional[float] = None):
...
+    is_unet=True,
+    prefix=None,
+):
+    from ..loaders.lora_base import LORA_ADAPTER_METADATA_KEY
+
+    if LORA_ADAPTER_METADATA_KEY in peft_state_dict:
+        metadata = peft_state_dict[LORA_ADAPTER_METADATA_KEY]
+        if metadata:
+            if prefix is not None:
+                metadata = {k.replace(f"{prefix}.", ""): v for k, v in metadata.items()}
```
BenjaminBossan, line 163: Could `k.removeprefix(prefix)` be used here? The current code would also remove the prefix if it's actually an infix, so the change would not have the same effect. I think this is intended only for actual prefixes.

src/diffusers/loaders/lora_base.py
```diff
@@ -882,16 +888,31 @@ def write_lora_layers(
...
+                    metadata = {"format": "pt"}
```
BenjaminBossan, line 908: At least for me, it would help to have a comment why metadata is initialized as `{"format": "pt"}`, probably it's required for safetensors?

src/diffusers/utils/testing_utils.py
```diff
@@ -132,6 +132,33 @@ def numpy_cosine_similarity_distance(a, b):
...
+        if dict1[key] != dict2[key]:
+            raise ValueError(
+                f"Difference found at key '{key}': first dictionary has {dict1[key]}, second dictionary has {dict2[key]}."
+            )
+
+    for key in dict2:
+        if key not in dict1:
+            raise ValueError(
+                f"Key '{key}' is missing in the first dictionary. Its value in the second dictionary is {dict2[key]}."
+            )
```
BenjaminBossan, line 157: You could create sets of the keys and check the `symmetric_difference`.

src/diffusers/utils/testing_utils.py
```diff
@@ -132,6 +132,33 @@ def numpy_cosine_similarity_distance(a, b):
     return distance
 
 
+def check_if_dicts_are_equal(dict1, dict2):
+    for key, value in dict1.items():
```
BenjaminBossan, line 136: Maybe add: `dict1, dict2 = dict1.copy(), dict2.copy()` to avoid mutating the dicts, which could make debugging of the tests harder.

src/diffusers/utils/testing_utils.py
```diff
@@ -132,6 +132,33 @@ def numpy_cosine_similarity_distance(a, b):
...
+        if isinstance(value, set):
+            dict1[key] = sorted(value)
+    for key, value in dict2.items():
+        if isinstance(value, set):
+            dict2[key] = sorted(value)
+
+    for key in dict1:
+        if key not in dict2:
+            raise ValueError(
```
BenjaminBossan, line 145: I would either return `False` here and let the caller assert that `check_if_dicts_are_equal` is `True`, or raise an `AssertionError` and not return `True` at the end.

tests/lora/test_lora_layers_wan.py
```diff
@@ -141,3 +146,64 @@ def test_simple_inference_with_text_lora_fused(self):
     @unittest.skip("Text encoder LoRA is not supported in Wan.")
     def test_simple_inference_with_text_lora_save_load(self):
         pass
+
+    def test_adapter_metadata_is_loaded_correctly(self):
+        # Will write the test in utils.py eventually.
```
BenjaminBossan, line 151:
```suggestion
        # TODO Will write the test in utils.py eventually.
```

BenjaminBossan (approved):
Thanks for the PR. Generally LGTM, my comments are just nits.

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Yeah that is intentional as I have not propagated that to every other pipeline.

src/diffusers/loaders/peft.py
sayakpaul: Rewrote this a bit as the metadata related bits can be moved out of `if prefix is not None`.

src/diffusers/utils/peft_utils.py
sayakpaul: I don't understand. The `prefix` is always meant to be actual (e.g., `transformer.r`, `transformer.lora_alpha`, etc.). So, I don't understand:

> The current code would also remove the prefix if it's actually an infix, so the change would not have the same effect.

src/diffusers/utils/testing_utils.py
sayakpaul: Going for the other suggestion on returning False as the current way of checking keys without `symmetric_difference` is more explicit to me and more readable.

src/diffusers/utils/peft_utils.py
BenjaminBossan: I mean this difference:

```python
>>> metadata = {"foo.bar": 0, "transformer.foo": 1, "bar.transformer.foo": 2}
>>> prefix = "transformer"
>>> {k.replace(f"{prefix}.", ""): v for k, v in metadata.items()}   # <= current code
{'foo.bar': 0, 'foo': 1, 'bar.foo': 2}
>>> {k.removeprefix(prefix + "."): v for k, v in metadata.items()}  # <= suggestion
{'foo.bar': 0, 'foo': 1, 'bar.transformer.foo': 2}
```

src/diffusers/utils/testing_utils.py
BenjaminBossan: To me, the advantage of `symmetric_difference` is that it immediately returns all differing keys, whereas the code as is first checks for missing keys in `dict2`, then when the user fixes it, there can be another error with keys missing in `dict1`. But I agree that some readers might not know about some of these set methods.

src/diffusers/utils/peft_utils.py
sayakpaul: Oh that is a fine suggestion. Thanks! Let me apply this to the codebase wide.

src/diffusers/loaders/peft.py
```diff
@@ -443,24 +465,22 @@ def save_lora_adapter(
                 underlying model has multiple adapters loaded.
             upcast_before_saving (`bool`, defaults to `False`):
                 Whether to cast the underlying model to `torch.float32` before serialization.
-            save_function (`Callable`):
-                The function to use to save the state dictionary. Useful during distributed training when you need to
-                replace `torch.save` with another method. Can be configured with the environment variable
-                `DIFFUSERS_SAVE_MODE`.
```
sayakpaul, line 449: Because it's not used.

src/diffusers/loaders/peft.py
```diff
@@ -470,7 +490,15 @@ def save_lora_adapter(
...
 
             def save_function(weights, filename):
-                return safetensors.torch.save_file(weights, filename, metadata={"format": "pt"})
+                # Inject framework format.
+                metadata = {"format": "pt"}
+                if lora_adapter_metadata is not None:
+                    for key, value in lora_adapter_metadata.items():
+                        if isinstance(value, set):
+                            lora_adapter_metadata[key] = list(value)
+                    metadata[LORA_ADAPTER_METADATA_KEY] = json.dumps(lora_adapter_metadata, indent=2, sort_keys=True)
```
sayakpaul, line 499: Automatically serialize the metadata when available. Cc: @a-r-r-o-w

sayakpaul:
@DN6 @a-r-r-o-w this is ready for a review.

sayakpaul:
Cc @linoytsaban once this is merged (I think close) let's apply the `lora_alpha` changes as initial planned. I would target one/two scripts and open the rest to the community. WDYT?

sayakpaul:
@a-r-r-o-w @DN6 a gentle ping on this one.

src/diffusers/utils/state_dict_utils.py
```diff
@@ -347,3 +348,23 @@ def state_dict_all_zero(state_dict, filter_str=None):
...
+
+    metadata_key = LORA_ADAPTER_METADATA_KEY
+    with safetensors.torch.safe_open(model_file, framework="pt", device="cpu") as f:
+        if hasattr(f, "metadata"):
+            metadata = f.metadata()
+            if metadata is not None:
+                metadata_keys = list(metadata.keys())
+                if not (len(metadata_keys) == 1 and metadata_keys[0] == "format"):
+                    peft_metadata = {k: v for k, v in metadata.items() if k != "format"}
+                    state_dict["lora_adapter_metadata"] = json.loads(peft_metadata[metadata_key])
```
a-r-r-o-w, line 369: I don't think we should mix the state dict and peft init metadata here. We should limit the state dict to being just weights because otherwise some parts of the code feel like they magically extract some variable.

Instead, we should probably return the state dict and metadata as tuple, and propagate to other methods as required.

src/diffusers/loaders/lora_base.py
```diff
@@ -347,8 +353,13 @@ def _load_lora_into_text_encoder(
         raise ValueError("At the moment, hotswapping is not supported for text encoders, please pass `hotswap=False`.")
 
     # Load the layers corresponding to text encoder and make necessary adjustments.
+    metadata = None
+    if LORA_ADAPTER_METADATA_KEY in state_dict:
+        metadata = state_dict[LORA_ADAPTER_METADATA_KEY]
     if prefix is not None:
         state_dict = {k.removeprefix(f"{prefix}."): v for k, v in state_dict.items() if k.startswith(f"{prefix}.")}
+    if metadata is not None:
+        state_dict[LORA_ADAPTER_METADATA_KEY] = metadata
```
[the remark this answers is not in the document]
sayakpaul: When serializing the state dict, we are injecting the state dict with the metadata. It's keys or values are not changing. So, we're not changing the format in any way.

src/diffusers/utils/peft_utils.py
```diff
@@ -147,7 +147,24 @@ def unscale_lora_layers(model, weight: Optional[float] = None):
...
+    from ..loaders.lora_base import LORA_ADAPTER_METADATA_KEY
+
+    if LORA_ADAPTER_METADATA_KEY in peft_state_dict:
+        metadata = peft_state_dict[LORA_ADAPTER_METADATA_KEY]
+    else:
+        metadata = None
+    if metadata:
+        if prefix is not None:
+            metadata = {k.removeprefix(f"{prefix}."): v for k, v in metadata.items() if k.startswith(f"{prefix}.")}
+        return metadata
```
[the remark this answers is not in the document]
sayakpaul: There's no state dict in the metadata.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -1376,13 +1425,23 @@ def save_lora_weights(
         if text_encoder_2_lora_layers:
             state_dict.update(cls.pack_weights(text_encoder_2_lora_layers, "text_encoder_2"))
 
+        if transformer_lora_adapter_metadata is not None:
+            lora_adapter_metadata.update(cls.pack_weights(transformer_lora_adapter_metadata, cls.transformer_name))
```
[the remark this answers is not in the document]
a-r-r-o-w: Oh okay, my mistake was in thinking cls.pack_weights returns the state dict. Only does so when `layers` is a torch module.

```python
@staticmethod
    def pack_weights(layers, prefix):
        layers_weights = layers.state_dict() if isinstance(layers, torch.nn.Module) else layers
        layers_state_dict = {f"{prefix}.{module_name}": param for module_name, param in layers_weights.items()}
        return layers_state_dict
```

It's a bit confusing why a function who's name implies it's for packaging weights to specific format is being used here (because unless you look at internal implementation, it looks like a dict of str/int/float/list is being passed for weight packing somehow). Anyway, internal implementation detail so can be refactored in future so that the naming makes more sense without having to look into actual implementation

sayakpaul:
@a-r-r-o-w we're not serializing state dict in the metadata or anything like that. We inject the metadata in the state dict safetensors. So, there's that separation. 

After loading the state dict and processing it, we make the metadata a part of the state dict (standard keys and values) to ease the rest of the process. I think this will answer most of your doubts.

> I don't think we should mix the state dict and peft init metadata here. We should limit the state dict to being just weights because otherwise some parts of the code feel like they magically extract some variable.

> Instead, we should probably return the state dict and metadata as tuple, and propagate to other methods as required.

I wanted to avoid that as I felt this might introduce breaking changes but this makes sense as it will make things less confusing. I will work on the changes accordingly.

src/diffusers/loaders/lora_base.py
```diff
@@ -261,10 +269,11 @@ def _fetch_state_dict(
                 user_agent=user_agent,
             )
             state_dict = load_state_dict(model_file)
+            metadata = None
     else:
         state_dict = pretrained_model_name_or_path_or_dict
 
-    return state_dict
+    return state_dict, metadata
```
sayakpaul, line 276: Should be okay as this is a private method.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -197,7 +198,8 @@ def load_lora_weights(
             pretrained_model_name_or_path_or_dict = pretrained_model_name_or_path_or_dict.copy()
 
         # First, ensure that the checkpoint is a compatible one and can be successfully loaded.
-        state_dict, network_alphas = self.lora_state_dict(pretrained_model_name_or_path_or_dict, **kwargs)
+        kwargs["return_lora_metadata"] = True
```
sayakpaul, line 201: I think this is okay to do otherwise we end up introducing another argument to `load_lora_weights()`.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -277,6 +281,7 @@ def lora_state_dict(
                 The subfolder location of a model file within a larger model repository on the Hub or locally.
             weight_name (`str`, *optional*, defaults to None):
                 Name of the serialized state dict file.
+            return_lora_metadata: TODO
```
sayakpaul, line 284: To not break backwards-compatibility. In case the state dict doesn't contain any metadata it will be None despite `return_lora_metadata=True`.

src/diffusers/loaders/lora_base.py
```diff
@@ -349,6 +367,8 @@ def _load_lora_into_text_encoder(
     # Load the layers corresponding to text encoder and make necessary adjustments.
     if prefix is not None:
         state_dict = {k.removeprefix(f"{prefix}."): v for k, v in state_dict.items() if k.startswith(f"{prefix}.")}
+        if metadata is not None:
```
sayakpaul, line 370: For future: `_load_lora_into_text_encoder()` and `load_lora_adapter()` share a bunch of common logic which I will refactor in a future PR.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -37,6 +37,7 @@
     LoraBaseMixin,
     _fetch_state_dict,
     _load_lora_into_text_encoder,
+    _pack_sd_with_prefix,
```
sayakpaul, line 40: The changes in this file look large because of the docstrings.

sayakpaul:
@a-r-r-o-w  (when you come back) I think I have resolved your concerns so, this should be ready to be reviewed. I will bulk update the training scripts with @linoytsaban in a near future PR.

sayakpaul:
@DN6 @a-r-r-o-w a gentle ping here.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -368,6 +373,7 @@ def load_lora_into_unet(
             adapter_name (`str`, *optional*):
                 Adapter name to be used for referencing the loaded adapter model. If not specified, it will use
                 `default_{i}` where i is the total number of adapters being loaded.
+            metadata: TODO
```
a-r-r-o-w: need to update the multiple todos

a-r-r-o-w (commented):
All the changes look good to me!

Handling metadata for each of the different components seems to require a lot of branching and different code paths though (I realize it is essentially just one or two changes repeated multiple times because each lora loader per model needs the changes). I'm trying to think if there's maybe a cleaner way to handle it (not a blocker though).

Would just like to note again that `pack_weights` is a confusing name when it's called in the context of metadata. That part should probably be refactored/renamed soon (follow-up PR or however you prefer).

sayakpaul:
> Handling metadata for each of the different components seems to require a lot of branching and different code paths though (I realize it is essentially just one or two changes repeated multiple times because each lora loader per model needs the changes). I'm trying to think if there's maybe a cleaner way to handle it (not a blocker though).

@a-r-r-o-w I hear you and I feel the same. This is why in the OP of this PR, I have said

> Modularize where possible (prefer to do it in a different PR)

I have also discussed this a bit with @DN6. So, this will be cleaned quite a bit now that we have sense of what needs to be factored out. Hope that makes sense.

> Would just like to note again that pack_weights is a confusing name when it's called in the context of metadata. That part should probably be refactored/renamed soon (follow-up PR or however you prefer).

Yes, that is because I didn't update all the `save_lora_weights()` methods like this:

https://github.com/huggingface/diffusers/blob/72b489d68cb55792a3b4bd6450edf909ad330da5/src/diffusers/loaders/lora_pipeline.py#L513C9-L519C14

I guess having the naming this way is less confusing/satisfactory?

Will take care of the TODOs.

sayakpaul:
@BenjaminBossan we decided to change the course of how we wanna support metadata a bit in this PR. Could you please review one more time? Sorry about the inconvenience.

src/diffusers/loaders/lora_base.py
```diff
@@ -206,6 +209,7 @@ def _fetch_state_dict(
     subfolder,
     user_agent,
     allow_pickle,
+    metadata=None,
```
BenjaminBossan, line 212: Why is it passed here?

src/diffusers/loaders/lora_base.py
BenjaminBossan: Instead of changing the signature of this function, the caller could also directly call `_load_sft_state_dict_metadata` after calling `_fetch_state_dict`.

src/diffusers/loaders/lora_base.py
```diff
@@ -314,13 +327,17 @@ def _load_lora_into_text_encoder(
...
     _pipeline=None,
     low_cpu_mem_usage=False,
     hotswap: bool = False,
 ):
     if not USE_PEFT_BACKEND:
         raise ValueError("PEFT backend is required for this method.")
 
+    if network_alphas and metadata:
+        raise ValueError("Both `network_alphas` and `metadata` cannot be specified.")
```
BenjaminBossan: IMO the current wording could be misinterpreted.

```suggestion
        raise ValueError("`network_alphas` and `metadata` cannot be specified both at the same time.")
```

src/diffusers/loaders/lora_base.py
```diff
@@ -398,7 +421,10 @@ def _load_lora_into_text_encoder(
                 if is_peft_version("<=", "0.13.2"):
                     lora_config_kwargs.pop("lora_bias")
 
-        lora_config = LoraConfig(**lora_config_kwargs)
+            try:
+                lora_config = LoraConfig(**lora_config_kwargs)
+            except TypeError as e:
+                raise TypeError(f"`LoraConfig` class could not be instantiated:\n{e}.")
```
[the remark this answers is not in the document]
BenjaminBossan: Hmm, how is this precaution, it just catches the error and changes the error message, right? Also, I would prefer to re-raise:
```suggestion
            try:
                lora_config = LoraConfig(**lora_config_kwargs)
            except TypeError as e:
                raise TypeError("`LoraConfig` class could not be instantiated.") from e
```

src/diffusers/loaders/lora_base.py
```diff
@@ -900,16 +924,32 @@ def write_lora_layers(
...
+        lora_adapter_metadata: Optional[dict] = None,
     ):
         if os.path.isfile(save_directory):
             logger.error(f"Provided path ({save_directory}) should be a directory, not a file")
             return
 
+        if lora_adapter_metadata and not safe_serialization:
+            raise ValueError("`lora_adapter_metadata` cannot be specified when not using `safe_serialization`.")
+        if lora_adapter_metadata and not isinstance(lora_adapter_metadata, dict):
+            raise ValueError("`lora_adapter_metadata` must be of type `dict`.")
```
BenjaminBossan:
```suggestion
            raise TypeError("`lora_adapter_metadata` must be of type `dict`.")
```

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -1916,8 +2055,13 @@ def lora_state_dict(
...
 
-        if return_alphas:
-            return state_dict, network_alphas
+        if return_alphas or return_lora_metadata:
+            outputs = [state_dict]
+            if return_alphas:
+                outputs.append(network_alphas)
+            if return_lora_metadata:
+                outputs.append(metadata)
+            return tuple(outputs)
```
BenjaminBossan, line 2069: Just FYI, there is no need to convert to list and then back to tuple, you can "append" to tuples like so: `my_tuple += (item,)`. This creates a new tuple, as tuples are immutable. But it's fine like it is too.

src/diffusers/loaders/peft.py
```diff
@@ -289,7 +303,11 @@ def load_lora_adapter(
                     if is_peft_version("<=", "0.13.2"):
                         lora_config_kwargs.pop("lora_bias")
 
-            lora_config = LoraConfig(**lora_config_kwargs)
+            try:
+                lora_config = LoraConfig(**lora_config_kwargs)
+            except TypeError as e:
+                raise TypeError(f"`LoraConfig` class could not be instantiated:\n{e}.")
+
```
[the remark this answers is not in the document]
BenjaminBossan: Same comment as I made earlier about re-raising.

examples/community/ip_adapter_face_id.py
```diff
@@ -282,10 +282,7 @@ def load_ip_adapter_face_id(self, pretrained_model_name_or_path_or_dict, weight_
         revision = kwargs.pop("revision", None)
         subfolder = kwargs.pop("subfolder", None)
 
-        user_agent = {
-            "file_type": "attn_procs_weights",
-            "framework": "pytorch",
-        }
+        user_agent = {"file_type": "attn_procs_weights", "framework": "pytorch"}
```
BenjaminBossan, line 285: Is this because of a change in ruff?

BenjaminBossan (approved):
Thanks for your continued work on this. I don't see any big issues, I just made some suggestions that may improve the PR.

examples/community/ip_adapter_face_id.py
sayakpaul: Yes.

src/diffusers/loaders/lora_base.py
sayakpaul: Because we pass `metadata` while loading the adapter:

https://github.com/huggingface/diffusers/blob/47cad58caade48ec9b6b7b66ec4e3c16dc4bdc56/src/diffusers/loaders/lora_pipeline.py#L213

which is, in turn, passed to this method if you step through it.

src/diffusers/loaders/lora_base.py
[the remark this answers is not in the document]
sayakpaul: Pardon my choice of English word.

src/diffusers/loaders/lora_base.py
a-r-r-o-w: Hmm, I don't have any strong feelings on doing it either way, so your call

a-r-r-o-w (approved):
Looks ready to merge to me, if Ben's comments have been addressed! Not sure if adding `metadata=None` param right after `adapter_name` instead of after `hotswap` is a breaking change (is it expected that user code may call `load_lora_into_transformer` directly? If so, I think definitely breaking for users that just pass arguments directly)

joeflack4:
Forgive my ignorance. Could anyone answer these questions, and maybe update the OP (and ideally the [docs](https://huggingface.co/docs/diffusers/index), if applicable), w/ the answers?

1. Does this apply to theoretically any LoRA from anywhere?
2. Are there situations in which the metadata may not be readable because it is not set correctly / in the expected standard way? (bonus for elaboration)
3. Is this feature something intended for users to use easily, directly, and specifically? (e.g. a CLI command, `parse-metadata PATH` or a func `parse_metadata(PATH)`?

Context: I'm an A1111 / ComfyUI user, using mostly Civit LoRAs, looking for an alternative / easy way to get metadata, ideally directly from LoRA binaries themselves.

I searched the [docs](https://huggingface.co/docs/diffusers/index) for "metadata", but onl code snippets were showing up, and they were all of the pattern `... export const metadata ...`. Most of the results that I clicked were broken, taking me to a page that said "The documentation page API/DIFFUSION_PIPELINE doesn’t exist in v0.33.1, but exists on the main version. Click [here](https://huggingface.co/docs/diffusers/main/en/api/diffusion_pipeline) to redirect to the main version of the documentation.".

sayakpaul:
Great questions!

1. No, it doesn't. It's currently specific to `diffusers`-formatted LoRAs. Although if your LoRAs are in the `safetensors` format, you can easily inject the metadata. Here is a reference:

https://github.com/huggingface/diffusers/blob/67603002024efdab26bc6f70a72fbf278e300100/src/diffusers/loaders/lora_base.py#L902-L913

2. I don't think that will be the case. That is because if the metadata was injected using the above-mentioned way, it should always be loadable. Within the library, we load it like so:

https://github.com/huggingface/diffusers/blob/67603002024efdab26bc6f70a72fbf278e300100/src/diffusers/utils/state_dict_utils.py#L358

3. The metadata injection is supposed to be done automatically from our training scripts that have been updated to use this feature ([example](https://github.com/huggingface/diffusers/blob/main/examples/dreambooth/train_dreambooth_lora_flux.py)).

We can definitely do a better job of documenting this. Pinging @stevhliu for ideas.

joeflack4:
Great answers! Thanks!

PR #11330, opened 2025-04-15
another fix for FlowMatchLCMScheduler forgotten import

asomoza:
I had to switch systems to do the `fix-copies` and somehow this was not included to the original PR.

@yiyixuxu


asomoza:
I noticed that we don't have tests for any of the newer schedulers, maybe we can open them to the community? If not, I can add it to my TODOs since I think this is important, my mistake should've a failed test and I didn't manually check this because I was confident that there were some tests for the imports.

cc: @sayakpaul

yiyixuxu (approved):

PR #11340, opened 2025-04-16
Add basic implementation of AuraFlowImg2ImgPipeline

AstraliteHeart:
# What does this PR do?

Adds a very mechanical conversion of other img2img pipelines (mostly SD3/Flux) to support AuraFlow. Seems to require a bit more strength (0.75+) compared to SDXL (my only point of reference that I've used a lot for I2I) but works fine and does not complain about GGUF (still need to check compilation). 

Fixes # (issue)

## Before submitting

## Who can review?

@cloneofsimo @sayakpaul
@yiyixuxu @asomoza


sayakpaul:
Thanks for yet another contribution! Could you post a snippet and some results?

AstraliteHeart:
Apologies, had to clean up things and make the tests actually work.

The docstrings included in the CL should be a good snippet, i.e.

```python
import torch
from diffusers import AuraFlowImg2ImgPipeline
import requests
from PIL import Image
from io import BytesIO

# download an initial image
url = "https://raw.githubusercontent.com/CompVis/stable-diffusion/main/assets/stable-samples/img2img/sketch-mountains-input.jpg"
response = requests.get(url)
init_image = Image.open(BytesIO(response.content)).convert("RGB")
init_image = init_image.resize((768, 512))

pipe = AuraFlowImg2ImgPipeline.from_pretrained("fal/AuraFlow-v0.3", torch_dtype=torch.float16)
pipe = pipe.to("cuda")
prompt = "A fantasy landscape, trending on artstation"
image = pipe(prompt=prompt, image=init_image, strength=0.75, num_inference_steps=50).images[0]
image.save("aura_flow_img2img.png")
```

Unfortunately seems that my math may be wrong somehow? 

With strength 0.75

image

with strength 0.85

image

with strength 0.95

image
 
but with 0.9

image

@DN6, any ideas?

src/diffusers/pipelines/aura_flow/pipeline_aura_flow_img2img.py
```diff
@@ -0,0 +1,761 @@
...
+                # Single generator - use its device if it has one
+                generator_device = getattr(generator, "device", "cpu") if generator is not None else "cpu"
+                noise = torch.randn(mean.shape, generator=generator, device=generator_device, dtype=mean.dtype)
+                sample = noise.to(mean.device)
+
+            # Compute latents
+            latents = mean + std * sample
+
+            # Scale latents
+            latents = latents * self.vae.config.scaling_factor
```
sayakpaul, line 426: I _think_ it should be:

```py
init_latents = (init_latents - mean) * self.vae.config.scaling_factor / latents_std
```

Following
https://github.com/huggingface/diffusers/blob/eef3d6595456e69a48989c2cc44c739792341e07/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_img2img.py#L751C17-L751C29

src/diffusers/pipelines/aura_flow/pipeline_aura_flow_img2img.py
```diff
@@ -0,0 +1,761 @@
...
+
+        # Get the original timestep using init_timestep
+        init_timestep = min(num_inference_steps * strength, num_inference_steps)
+
+        t_start = int(max(num_inference_steps - init_timestep, 0))
+        timesteps = self.scheduler.timesteps[t_start:]
+
+        return timesteps, num_inference_steps - t_start
+
+    def prepare_img2img_latents(
```
DN6, line 356: Can this be placed under `prepare_latents` like the other Img2Img pipelines
https://github.com/huggingface/diffusers/blob/bbd0c161b55ba2234304f1e6325832dd69c60565/src/diffusers/pipelines/flux/pipeline_flux_img2img.py#L610

src/diffusers/pipelines/aura_flow/pipeline_aura_flow_img2img.py
```diff
@@ -0,0 +1,761 @@
...
+            prompt_attention_mask=prompt_attention_mask,
+            negative_prompt_attention_mask=negative_prompt_attention_mask,
+            max_sequence_length=max_sequence_length,
+        )
+
+        if do_classifier_free_guidance:
+            prompt_embeds = torch.cat([negative_prompt_embeds, prompt_embeds], dim=0)
+
+        # 5. Prepare timesteps
+        timesteps, num_inference_steps = self.get_timesteps(
```
DN6, line 657: Think timesteps need to be adjusted for strength and shift no?
https://github.com/huggingface/diffusers/blob/bbd0c161b55ba2234304f1e6325832dd69c60565/src/diffusers/pipelines/flux/pipeline_flux_img2img.py#L876-L892

DN6 (commented):
Looking good 👍🏽 I think some parts of the implementation have to be adjusted to look more similar to the existing Flux pipelines.

src/diffusers/pipelines/aura_flow/pipeline_aura_flow_img2img.py
AstraliteHeart: The VAE config has `('latents_mean', None), ('latents_std', None)` so I believe the code would be a noop but implemented.

src/diffusers/pipelines/aura_flow/pipeline_aura_flow_img2img.py
AstraliteHeart: Not sure if this is what you intended, but done?

src/diffusers/pipelines/aura_flow/pipeline_aura_flow_img2img.py
AstraliteHeart: done, probably :)

AstraliteHeart:
Unfortunately still seeing noise at some values of strength so something in my math must be wrong.

src/diffusers/pipelines/aura_flow/pipeline_aura_flow_img2img.py
sayakpaul: Oh okay then `init_latents *= self.vae.config.scaling_factor` should be just fine. We can safely remove the other things.

sayakpaul:
Do those values generally tend to be higher?

AstraliteHeart:
image

sayakpaul:
A bit hard to see sorry

AstraliteHeart:
Weird, I can click on the image to get the full sized one (with an extra click)
It's the 0.88 - 0.94 range that is visual noise, and I think images above 0.94 not even using initial image.

AstraliteHeart:
I still have no idea what is going on but I think the code is correct yet AF may require some special work done. Here are my observations:

1.  When using certain input images, the VAE encoder, responsible for creating the initial latent representation (`x0`) of the input image, produces a latent distribution (`latent_dist`) where the standard deviation (`std`) component consistently collapses to effectively zero (e.g., `std=0.0000`, corresponding to a highly negative `logvar`).

2.  This variance collapse is observed **even when ensuring the VAE was loaded and operated entirely in `float32` precision**. This confirms the issue is not merely an `fp16` underflow problem during VAE computation but rather suggests the SDXL VAE predicts near-zero variance for this type of input (I was aware of different issues with SDXL VAE but not this specific behavior).

3.  Because the predicted `std` is zero, the subsequent step of sampling the initial latent variable `x0` from this distribution (`mean + std * noise`) becomes deterministic, effectively yielding only the `mean` component (`x0 = mean`).

4.  To address the lack of variance in `x0`, I attempted an experiment in which the `logvar` output by the VAE was manually "clamped" to a minimum value (tested `min_logvar = -10.0` resulting in `std ≈ 0.0067`, and `min_logvar = -4.0` resulting in `std ≈ 0.135`) before sampling `x0`. This successfully introduced non-zero variance into the initial latent state. *At this point my assumption was just - the issue is in VAE.*

5.  **Despite** successfully injecting variance into `x0` via clamping, the pipeline **still produced noise/corrupted images when run at high `strength` values (e.g., `strength=0.9`)**.

6.  Crucially, the pipeline **works reasonably well at lower `strength` values (e.g., `strength=0.7`)**, producing recognizable image outputs that incorporate the initial image structure.

The core issue no longer seems to be *only* the deterministic `x0` caused by the initial VAE variance collapse (as fixing that didn't solve the high-strength problem). Instead, the failure at high strength (`0.9`) **may stem from an instability in the denoising process itself** when initiated from the very high noise levels corresponding to these high strengths. The process is stable when starting from the lower noise levels associated with moderate strength (`0.7`).

sayakpaul:
Thanks for the detailed analysis. Do these vary from `AuraFlow` and `AuraFlow0.3`?

AstraliteHeart:
Yes, I am seeing same behavior in 0.2.

I focused too much on VAE in the last comment - I don't think it's the root cause (after all SDXL works just fine) and perhaps the real issue is some kind numerical instability we are facing?

I've included 3 videos generated taking a snapshot of the model state each 5 frames - before, at the moment of the issue and after. At least looking at them I can't notice anything weird that can explain the problem

https://github.com/user-attachments/assets/471f3572-5552-4f4e-8632-f459675ce44d

https://github.com/user-attachments/assets/e4acd117-7b89-44e8-819f-e1370026e13a

https://github.com/user-attachments/assets/2ce0fe29-6c47-445d-9baa-dd5c9bbc22d3

I've also attempted to affect the problematic strength range by changing number of steps or guidance scale but it had no effect. Interestingly enabling `use_karras_sigmas=True` on the scheduler seems to "fix" the issue as I no longer can hit the noise output but I still experience an very sharp change from "low strength i2i" to "just t2i" at around 0.98 strength. Feels like I am missing something super obvious here.

src/diffusers/pipelines/aura_flow/pipeline_aura_flow_img2img.py
```diff
@@ -0,0 +1,783 @@
...
+                )
+            else:
+                # Single generator - use its device if it has one
+                noise = randn_tensor(
+                    latents.shape, generator=generator, device=latents.device, dtype=latents.dtype
+                )
+
+            latents = self.scheduler.scale_noise(latents, init_timestep, noise)
+
+        return latents
```
bghira, line 460:
```suggestion
    def prepare_latents(
        self,
        image,
        timestep,
        batch_size,
        num_images_per_prompt,
        dtype,
        device,
        generator=None,
    ):
        if not isinstance(image, (torch.Tensor, PIL.Image.Image, list)):
            raise ValueError(
                f"`image` must be `torch.Tensor`, `PIL.Image.Image` or list, got {type(image)}"
            )

        image = image.to(device=device, dtype=dtype)
        batch_size = batch_size * num_images_per_prompt

        if image.shape[1] == 4:
            latents_0 = image
        else:
            # VAE ⇢ latents  (ALWAYS on fp32 for numerical stability)
            orig_dtype = self.vae.dtype
            if orig_dtype != torch.float32:
                self.vae.to(dtype=torch.float32)

            latent_dist = self.vae.encode(image.to(dtype=torch.float32)).latent_dist
            latents_0  = latent_dist.mean                      # ❶ deterministic!

            if orig_dtype != torch.float32:
                self.vae.to(dtype=orig_dtype)

            # scale
            latents_0 = latents_0 * self.vae.config.scaling_factor

        # replicate to match `batch_size`
        if latents_0.shape[0] != batch_size:
            if batch_size % latents_0.shape[0] != 0:
                raise ValueError(
                    f"Cannot duplicate image batch of size {latents_0.shape[0]} "
                    f"to effective batch size {batch_size}."
                )
            repeats   = batch_size // latents_0.shape[0]
            latents_0 = latents_0.repeat(repeats, 1, 1, 1)

        noise = randn_tensor(
            latents_0.shape,
            generator=generator,
            device=latents_0.device,
            dtype=latents_0.dtype,
        )

        # make sure `timestep` is 1-D and matches batch
        if isinstance(timestep, (int, float)):
            timestep = torch.tensor([timestep], device=latents_0.device, dtype=latents_0.dtype)
        timestep = timestep.expand(latents_0.shape[0])

        latents = self.scheduler.scale_noise(latents_0, timestep, noise)
        return latents

```

bghira: you're injecting noise twice here and overwriting the init state

AstraliteHeart:
Thank you @bghira, your code makes much more sense! Things are definitely looking better now and lower strengths produce reasonably results, but the noise issue remains and after running a few experiments I continue to be confused. The banding pattern is interesting, perhaps wrong dimensions somewhere?

image

bghira:
can you try with square samples?

bghira:
another thing i've discovered lately but wasn't sure whether i should share it or not, as, withholding it means my own platform's image results are better :sweat_smile: but if we decouple the strength from the number of steps / starting step for denoising, we can modulate the img2img by the sigmas and get a more smooth trajectory on the denoising updates.

PR #11349, opened 2025-04-17
[cogview4][feat] Support attention mechanism with variable-length support and batch packing

OleehyO:
# What does this PR do?

This PR is primarily aimed at adding native multi-resolution + packing training support to CogView4 to better meet user needs. We have conducted relevant tests in [this PR](https://github.com/THUDM/CogKit/pull/26).

## Before submitting

## Who can review?

@a-r-r-o-w 
@zRzRzRzRzRzRzR


src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -73,8 +73,9 @@ def __init__(self, embedding_dim: int, dim: int) -> None:
     def forward(
         self, hidden_states: torch.Tensor, encoder_hidden_states: torch.Tensor, temb: torch.Tensor
     ) -> Tuple[torch.Tensor, torch.Tensor]:
-        norm_hidden_states = self.norm(hidden_states)
-        norm_encoder_hidden_states = self.norm_context(encoder_hidden_states)
+        dtype = hidden_states.dtype
```
yiyixuxu, line 76: ohh, would these norm layers change input dtype otherwise? (these are layernorms that do not come with weights I think)

yiyixuxu:
let me know what you think too @a-r-r-o-w 
I wonder if it would be easier to make a new attention processor for this?

src/diffusers/models/transformers/transformer_cogview4.py
OleehyO: These layernorms indeed have no weights, but we found that they turn into fp32 after the forward pass

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -124,62 +125,263 @@ def __call__(
         attn: Attention,
         hidden_states: torch.Tensor,
         encoder_hidden_states: torch.Tensor,
-        attention_mask: Optional[torch.Tensor] = None,
-        image_rotary_emb: Optional[torch.Tensor] = None,
+        attention_mask: Optional[Dict[str, torch.Tensor]] = None,
+        image_rotary_emb: Optional[tuple[torch.Tensor, torch.Tensor] | list[tuple[torch.Tensor, torch.Tensor]]] = None,
     ) -> torch.Tensor:
+        """
```
a-r-r-o-w, line 131: Let's move all this into in another attention processor since it complicates reading through the current inference specific code. For training, you can then replace the attetion processor in your training codebase when initializing the models. You can maybe call it `CogView4TrainingAttnProcessor`

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -124,62 +125,263 @@ def __call__(
         attn: Attention,
         hidden_states: torch.Tensor,
         encoder_hidden_states: torch.Tensor,
-        attention_mask: Optional[torch.Tensor] = None,
-        image_rotary_emb: Optional[torch.Tensor] = None,
+        attention_mask: Optional[Dict[str, torch.Tensor]] = None,
+        image_rotary_emb: Optional[tuple[torch.Tensor, torch.Tensor] | list[tuple[torch.Tensor, torch.Tensor]]] = None,
```
a-r-r-o-w, line 129: Diffusers does not use the modern way for type hinting yet. Could you change these occurences to use `Tuple`, `List` and `Union` instead?

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -422,7 +625,10 @@ def forward(
         batch_size, num_channels, height, width = hidden_states.shape
 
         # 1. RoPE
-        image_rotary_emb = self.rope(hidden_states)
+        if image_rotary_emb is None:
+            image_rotary_emb = self.rope(hidden_states)
+        else:
+            image_rotary_emb = image_rotary_emb
```
a-r-r-o-w, line 631:
```suggestion
```

no-op

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -402,6 +604,7 @@ def forward(
         attention_kwargs: Optional[Dict[str, Any]] = None,
         return_dict: bool = True,
         attention_mask: Optional[torch.Tensor] = None,
+        image_rotary_emb: Optional[tuple[torch.Tensor, torch.Tensor] | list[tuple[torch.Tensor, torch.Tensor]]] = None,
```
a-r-r-o-w, line 607: Same comment about type hint here

src/diffusers/models/transformers/transformer_cogview4.py
OleehyO: Make sense, I'll update this part of the code

src/diffusers/models/transformers/transformer_cogview4.py
OleehyO: Will do

OleehyO:
@a-r-r-o-w It has been renamed to `CogView4TrainingAttnProcessor`, and a bug in the original `CogView4AttnProcessor` has been fixed. We use the same naming format in both `CogView4AttnProcessor` and `CogView4TrainingAttnProcessor`. Please take a look

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -355,6 +647,7 @@ def __init__(
         pos_embed_max_size: int = 128,
         sample_size: int = 128,
         rope_axes_dim: Tuple[int, int] = (256, 256),
+        attn_processor: Union[CogView4AttnProcessor, CogView4TrainingAttnProcessor] = CogView4AttnProcessor(),
```
a-r-r-o-w, line 650: Let's revert all these `attn_processor` related changes please because it is not really in diffusers-style. All blocks will end up sharing the same object instance.

The way you would set attention processors manually in training code would be something like:

```python
for name, submodule in transformer.named_modules():
    if isinstance(submodule, Attention):
        submodule.processor = CogView4TrainingAttnProcessor()
```

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -401,7 +694,8 @@ def forward(
         crop_coords: torch.Tensor,
         attention_kwargs: Optional[Dict[str, Any]] = None,
         return_dict: bool = True,
-        attention_mask: Optional[torch.Tensor] = None,
+        attention_mask: Optional[Dict[str, torch.Tensor]] = None,
+        image_rotary_emb: Optional[Tuple[torch.Tensor, torch.Tensor] | List[Tuple[torch.Tensor, torch.Tensor]]] = None,
```
a-r-r-o-w, line 698: Type hint needs to be modified here to use Union, instead of `|`

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -159,13 +166,14 @@ def __call__(
...
         # 4. Attention
         if attention_mask is not None:
-            text_attention_mask = attention_mask.float().to(query.device)
-            actual_text_seq_length = text_attention_mask.size(1)
-            new_attention_mask = torch.zeros((batch_size, text_seq_length + image_seq_length), device=query.device)
-            new_attention_mask[:, :actual_text_seq_length] = text_attention_mask
-            new_attention_mask = new_attention_mask.unsqueeze(2)
-            attention_mask_matrix = new_attention_mask @ new_attention_mask.transpose(1, 2)
-            attention_mask = (attention_mask_matrix > 0).unsqueeze(1).to(query.dtype)
+            text_attn_mask = attention_mask["text_attn_mask"]
```
a-r-r-o-w, line 169: Let's not change this to an attention mask dict, since otherwise this might break for users already passing attention_mask as a tensor.

Can we instead accept a tensor? For this attention processor, let's make sure there are no changes and only add the new attention processor.

Instead of passing dict here, let's add all required params to the training attention processor arguments (`text_attn_mask`, `latent_attn_mask`, `batch_flag` should all be separate params). In the transformer, we accept `**kwargs`. You can propagate those values to the custom attention processor instead of adding these 3 parameters in all function's params

OleehyO:
I have changed back to using `attention_kwargs` in `CogView4Transformer2DModel` to pass attention parameters.

a-r-r-o-w (approved):
Thanks for addressing the comments!

yiyixuxu:
@bot /style

src/diffusers/models/transformers/transformer_cogview4.py
```diff
@@ -125,8 +129,11 @@ def __call__(
         hidden_states: torch.Tensor,
         encoder_hidden_states: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
-        image_rotary_emb: Optional[torch.Tensor] = None,
-    ) -> torch.Tensor:
+        image_rotary_emb: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        dtype = encoder_hidden_states.dtype
+        device = encoder_hidden_states.device
```
yiyixuxu, line 135:
```suggestion
```

yiyixuxu:
@bot /style

PR #11350, opened 2025-04-17
Update controlnet_flux.py

haofanwang:
Add support for [FLUX.1-dev-ControlNet-Union-Pro-2.0](https://huggingface.co/Shakker-Labs/FLUX.1-dev-ControlNet-Union-Pro-2.0) which removes mode embedding and uses only double blocks. Compatible with the previous [FLUX.1-dev-ControlNet-Union-Pro](https://huggingface.co/Shakker-Labs/FLUX.1-dev-ControlNet-Union-Pro).


yiyixuxu (approved):
thanks!

PR #11371, opened 2025-04-20
Fix Wan I2V prepare_latents dtype

a-r-r-o-w:
`dtype` is set to `torch.float32` when `prepare_latents` is called. This works if the VAE is loaded in `torch.float32` (as done in the docstring). But, loading the VAE in any other dtype fails with a dtype mismatch error.


yiyixuxu (approved):
thanks!

Issue #11376, opened 2025-04-21
Offloading behaviour for HiDream seems broken


sayakpaul:
When using `enable_model_cpu_offload()` I would expect the final component in the offloading string chain to be offloaded to the CPU to realize memory savings. But it's not likely the case. 

Consider the script below:

<details>
<summary>Code</summary>

```py
from transformers import PreTrainedTokenizerFast, LlamaForCausalLM
from diffusers import DiffusionPipeline
import torch

def print_detailed_memory(step_name):
    print(f"\n=== CUDA Memory Stats {step_name} ===")
    torch.cuda.reset_peak_memory_stats()
    print(f"Current allocated: {torch.cuda.memory_allocated() / 1024**3:.2f} GB")
    print(f"Max allocated: {torch.cuda.max_memory_allocated() / 1024**3:.2f} GB")
    print(f"Current reserved: {torch.cuda.memory_reserved() / 1024**3:.2f} GB")
    print(f"Max reserved: {torch.cuda.max_memory_reserved() / 1024**3:.2f} GB")

tokenizer_4 = PreTrainedTokenizerFast.from_pretrained("meta-llama/Meta-Llama-3.1-8B-Instruct")
text_encoder_4 = LlamaForCausalLM.from_pretrained(
    "meta-llama/Meta-Llama-3.1-8B-Instruct",
    output_hidden_states=True,
    output_attentions=True,
    torch_dtype=torch.bfloat16,
)
repo_id = "HiDream-ai/HiDream-I1-Full"
pipe = DiffusionPipeline.from_pretrained(
    repo_id, 
    transformer=None, 
    vae=None, 
    tokenizer_4=tokenizer_4, 
    text_encoder_4=text_encoder_4, 
    torch_dtype=torch.bfloat16
)
pipe.enable_model_cpu_offload()
with torch.no_grad():
    (
        prompt_embeds_t5,
        negative_prompt_embeds_t5,
        prompt_embeds_llama3,
        negative_prompt_embeds_llama3,
        pooled_prompt_embeds,
        negative_pooled_prompt_embeds,
    ) = pipe.encode_prompt(prompt="hello")

print_detailed_memory("after_prompt_encoding")

for name, component in pipe.components.items():
    if component is not None and isinstance(component, torch.nn.Module):
        print(f"{name=}{component.device=}")
        del component 
    gc.collect()
torch.cuda.empty_cache()
print_detailed_memory("after_prompt_encoding_cleanup")
```

</details>

[...]

PR #11375, opened 2025-04-21
Fix group offloading with block_level and use_stream=True

a-r-r-o-w:
Fixes #11307 

The previous implementation assumed that the layers were instantiated in order of invocation. This is not true for HiDream (caption projection layers are instantiated after transformer layers).

The new implementation makes sure to first capture invocation order and then apply group offloading. In the case of `use_stream=True`, it does not really make sense to onload more than 1 block at a time, so we also now raise an error if `num_blocks_per_group != 1` when `use_stream=True`

Another possible fix is to simply move the initialization of the caption layers above the transformer blocks.

@sayakpaul @asomoza Could you verify if this fixes it for you?

PR #11385, opened 2025-04-22
[docs] Memory optims

stevhliu:
Refactors the memory optimization docs and combines it with working with big models (distributed setups).

Let me know if I'm missing anything!


Heasterian:
AutoencoderKLWan and AsymmetricAutoencoderKL does not support tiling or slicing (asymmetric have just unused flags), it most likely should be mentioned.

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,231 @@ specific language governing permissions and limitations under the License.
...
-image = pipe(prompt).images[0]
+Diffusers uses the maxmium memory of all devices by default, but if they don't fit on the GPUs, then you'll need to use a single GPU and offload to the CPU with the methods below.
+
+- [`~DiffusionPipeline.enable_model_cpu_offload`] only works on a single GPU but a very large model may not fit on it
+- [`~DiffusionPipeline.enable_sequential_cpu_offload`] may work but it is extremely slow and also limited to a single GPU
+
+Use the [`~DiffusionPipeline.reset_device_map`] method to reset the `device_map`. This is necessary if you want to use methods like `.to()`, [`~DiffusionPipeline.enable_sequential_cpu_offload`], and [`~DiffusionPipeline.enable_model_cpu_offload`] on a pipeline that was device-mapped.
+
+```py
+pipeline.reset_device_map
```
sayakpaul, line 106:
```suggestion
pipeline.reset_device_map()
```

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,231 @@ specific language governing permissions and limitations under the License.
...
+).images[0]
+print(f"Max memory reserved: {torch.cuda.max_memory_allocated() / 1024**3:.2f} GB")
 ```
 
-<Tip warning={true}>
+## Model offloading
+
+Model offloading moves entire models to the GPU instead of selectively moving *some* layers or model components. One of the main pipeline models, usually the text encoder, UNet, and VAE, is placed on the GPU while the other components are held on the CPU. Components like the UNet that run multiple times stays on the GPU until its completely finished and no longer needed. This eliminates the communication overhead of [CPU offloading](#cpu-offloading) and makes model offloading a faster alternative. The tradeoff is memory savings won't be as large.
+
+> [!WARNING]
```
sayakpaul, line 185: Do we want to add a `warning` after `enable_sequential_cpu_offload()` that it's terribly slow and can often appear to be impractical?

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,231 @@ specific language governing permissions and limitations under the License.
...
+> To properly offload models after they're called, it is required to run the entire pipeline and models in the expected order. Keep this in mind if models are reused outside the pipeline context after hooks have been installed (see [Removing Hooks](https://huggingface.co/docs/accelerate/en/package_reference/big_modeling#accelerate.hooks.remove_hook_from_module) for more details). This is a stateful operation that installs hooks on the model.
```
sayakpaul, line 186: This reads a bit incomplete:

> Keep this in mind if models are reused outside the pipeline context after hooks have been installed

sayakpaul: Also, not sure if it would make sense to include but users can still benefit from `pipeline.enable_model_cpu_offload()` when doing stuff like `pipeline.encode_prompt()`. See https://github.com/huggingface/diffusers/issues/11376

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,231 @@ specific language governing permissions and limitations under the License.
...
+Group offloading moves groups of internal layers ([torch.nn.ModuleList](https://pytorch.org/docs/stable/generated/torch.nn.ModuleList.html) or [torch.nn.Sequential](https://pytorch.org/docs/stable/generated/torch.nn.Sequential.html)) to the CPU. It uses less memory than [model offloading](#model-offloading) and it is faster than [CPU offloading](#cpu-offloading) because it reduces communication overhead.
 
-To enable group offloading, call the [`~ModelMixin.enable_group_offload`] method on the model if it is a Diffusers model implementation. For any other model implementation, use [`~hooks.group_offloading.apply_group_offloading`]:
+> [!WARNING]
+> Group offloading may not work with all models if the forward implementation contains weight-dependent device casting of inputs because it may clash with group offloading's device casting mechanism.
```
sayakpaul, line 215: Providing some example models would be helpful here. Cc: @a-r-r-o-w

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,231 @@ specific language governing permissions and limitations under the License.
...
-# pipe.transformer.enable_group_offload(onload_device=onload_device, offload_device=offload_device, offload_type="leaf_level", use_stream=True, record_stream=True)
+# Use the enable_group_offload method for Diffusers model implementations
+pipeline.transformer.enable_group_offload(onload_device=onload_device, offload_device=offload_device, offload_type="leaf_level")
 
-# For any other model implementations, the apply_group_offloading function can be used
-apply_group_offloading(pipe.text_encoder, onload_device=onload_device, offload_type="block_level", num_blocks_per_group=2)
-apply_group_offloading(pipe.vae, onload_device=onload_device, offload_type="leaf_level")
+# Use the apply_group_offloading method for other model components
+apply_group_offloading(pipeline.text_encoder, onload_device=onload_device, offload_type="block_level", num_blocks_per_group=2)
+apply_group_offloading(pipeline.vae, onload_device=onload_device, offload_type="leaf_level")
```
sayakpaul, line 239: `vae` is a subclass of `ModelMixin` so, we should be able to use `enable_group_offload()`.

docs/source/en/optimization/memory.md
```diff
@@ -193,48 +246,53 @@ prompt = (
...
-</Tip>
+```py
+pipeline.transformer.enable_group_offload(onload_device=onload_device, offload_device=offload_device, offload_type="leaf_level", use_stream=True, record_stream=True)
+```
 
-## FP8 layerwise weight-casting
+The `low_cpu_mem_usage` parameter can be set to `True` to reduce CPU memory usage when using streams during group offloading. It is best for `leaf_level` offloading and when CPU memory is bottlenecked. Memory is saved by creating pinned tensors on the fly instead of pre-pinning them. However, this may increase overall execution time.
 
-PyTorch supports `torch.float8_e4m3fn` and `torch.float8_e5m2` as weight storage dtypes, but they can't be used for computation in many different tensor operations due to unimplemented kernel support. However, you can use these dtypes to store model weights in fp8 precision and upcast them on-the-fly when the layers are used in the forward pass. This is known as layerwise weight-casting.
+## FP8 layerwise casting
```
sayakpaul, line 269:
```suggestion
## Layerwise casting
```

As it's not specific to FP8.

docs/source/en/optimization/memory.md
```diff
@@ -193,48 +246,53 @@ prompt = (
...
 
-Typically, inference on most models is done with `torch.float16` or `torch.bfloat16` weight/computation precision. Layerwise weight-casting cuts down the memory footprint of the model weights by approximately half.
+Layerwise casting stores weights in a smaller data format (`torch.float8_e4m3fn` and `torch.float8_e5m2`) to use less memory and upcasts those weights to `torch.float16` or `torch.bfloat16` for computation. Certain layers (normalization and modulation related weights) are skipped because storing them in fp8 can degrade generation quality.
```
sayakpaul, line 271:
```suggestion
Layerwise casting stores weights in a smaller data format (for example: `torch.float8_e4m3fn` and `torch.float8_e5m2`) to use less memory and upcasts those weights to a higher precision, e.g., `torch.float16` or `torch.bfloat16` for computation. Certain layers (normalization and modulation related weights) are skipped because storing them in fp8 can degrade generation quality.
```

sayakpaul (commented):
Thanks for the initiative! I left some minor comments, let me know if they make sense.

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,231 @@ specific language governing permissions and limitations under the License.
...
+
+```py
+pipeline.reset_device_map
 ```
 
-CPU offloading works on submodules rather than whole models. This is the best way to minimize memory consumption, but inference is much slower due to the iterative nature of the diffusion process. The UNet component of the pipeline runs several times (as many as `num_inference_steps`); each time, the different UNet submodules are sequentially onloaded and offloaded as needed, resulting in a large number of memory transfers.
+## Sliced VAE
 
-<Tip>
+Sliced VAE saves memory by processing an image in smaller non-overlapping "slices" instead of processing the entire image at once. This reduces peak memory usage because the GPU is only processing a small slice at a time.
```
a-r-r-o-w, line 111: VAE slicing refers to splitting a big batch of input into single batch of data, and processing each batch separately. Typically, users generate one image at a time and in that case this option does not save any memory. If say a user generates 4 images at once using multiple prompts or `num_images_per_prompt > 1`, then decoding with the VAE would increase the peak activation memory by roughly 4x. Slicing would make it so you decode 1 image at a time instead of all 4 together

Here "an image in smaller non-overlapping 'slices' instead of processing the entire image at once" looks incorrect

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,231 @@ specific language governing permissions and limitations under the License.
...
+pipeline.enable_vae_slicing()
+pipeline(["An astronaut riding a horse on Mars"]*32).images[0]
+print(f"Max memory reserved: {torch.cuda.max_memory_allocated() / 1024**3:.2f} GB")
+```
 
-When using [`~StableDiffusionPipeline.enable_sequential_cpu_offload`], don't move the pipeline to CUDA beforehand or else the gain in memory consumption will only be minimal (see this [issue](https://github.com/huggingface/diffusers/issues/1934) for more information).
+## VAE tiling
 
-[`~StableDiffusionPipeline.enable_sequential_cpu_offload`] is a stateful operation that installs hooks on the models.
+VAE tiling saves memory by dividing an image into smaller overlapping tiles instead of processing the entire image at once. This also reduces peak memory usage because the GPU is only processing a tile at a time. Unlike sliced VAE, tiled VAE maintains some context between tiles because they overlap which can generate more coherent images.
```
a-r-r-o-w, line 130: Sliced VAE is simply breaking a larger batch of data into batch_size=1 data and sequentially processing it. The comparison here about slicesd VAE vs tiled VAE maintaing context between tiles seems incorrect

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,231 @@ specific language governing permissions and limitations under the License.
...
 
-</Tip>
+Call [`~StableDiffusionPipeline.enable_vae_tiling`] to enable VAE tiling. The generated image may have some tone variation from tile-to-tile because they're decoded separately, but there shouldn't be any obvious seams between the tiles. Tiling is disabled for images that are 512x512 or smaller.
```
a-r-r-o-w, line 132: "disabled for images that are 512x512 or smaller"

This should be true in most cases but is not always true. Maybe we could write this as "Tiling is disabled for resolutions lower than a pre-specified (but configurable) limit, for example 512x512 for the VAE used by StableDiffusionPipeline" or similar?

docs/source/en/optimization/memory.md
a-r-r-o-w: Not sure I recall any official model implementation in transformers/diffusers off the top of my head. Basically, if you cast inputs by peeking into the device of a particular weight layer in a model, it might fail. I'll try to find/remember an example

docs/source/en/optimization/memory.md
stevhliu: I have several sentences about this in the `## CPU offloading` section and I bolded that it is extremely slow. I think that should do the trick :)

docs/source/en/optimization/memory.md
stevhliu: Let me know when you remember them and we can update in a later PR!

docs/source/en/optimization/memory.md
stevhliu: Removed comparison!

stevhliu:
Thanks for the feedback on the memory doc! I also updated the inference speed doc, so please feel free to check it out and leave some feedback ❤️

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,231 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
-| fp16     | 3.51s   | x1.50    |
-| combined | 3.41s   | x1.54    |
+The precision and data type of the model weights affect inference speed because a higher precision requires more memory to load and more time to perform the computations. PyTorch loads model weights in float32 or full precision by default, so changing the data type is a simple way to quickly get faster inference.
 
-## TensorFloat-32
+<hfoptions id="dtypes">
+<hfoption id="bfloat16">
 
-On Ampere and later CUDA devices, matrix multiplications and convolutions can use the [TensorFloat-32 (tf32)](https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/) mode for faster, but slightly less accurate computations. By default, PyTorch enables tf32 mode for convolutions but not matrix multiplications. Unless your network requires full float32 precision, we recommend enabling tf32 for matrix multiplications. It can significantly speed up computations with typically negligible loss in numerical accuracy.
+bfloat16 is similar to float16 but it is more robust to numerical errors. Hardware support for bfloat16 varies, but most modern GPUs are capable of supporting bfloat16.
```
sayakpaul, line 28: Do you think it makes sense to provide a link to the GPUs that support bfloat16?
https://developer.nvidia.com/cuda-gpus

(Nevermind, it's clarified later)

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,231 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
+prompt = "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k"
+pipeline(prompt, num_inference_steps=30).images[0]
 ```
 
-> [!WARNING]
-> Don't use [torch.autocast](https://pytorch.org/docs/stable/amp.html#torch.autocast) in any of the pipelines as it can lead to black images and is always slower than pure float16 precision.
+</hfoption>
+<hfoption id="TensorFloat-32">
 
-## Distilled model
```
sayakpaul, line 60: Good step.

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,231 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
-    <figcaption class="mt-2 text-center text-sm text-gray-500">original Stable Diffusion</figcaption>
-  </div>
-  <div>
-    <img class="rounded-xl" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/distilled_sd.png"/>
-    <figcaption class="mt-2 text-center text-sm text-gray-500">distilled Stable Diffusion</figcaption>
-  </div>
-</div>
+## torch.compile
 
-### Tiny AutoEncoder
```
sayakpaul, line 101: Good step.

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,231 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
+    "stabilityai/stable-diffusion-xl-base-1.0", torch_dtype=torch.float16
 ).to("cuda")
-distilled.vae = AutoencoderTiny.from_pretrained(
-    "sayakpaul/taesd-diffusers", torch_dtype=torch.float16, use_safetensors=True,
+pipeline.unet.to(memory_format=torch.channels_last)
+pipeline.vae.to(memory_format=torch.channels_last)
+pipeline.unet = torch.compile(pipeline.unet,
+    mode="max-autotune",
+    fullgraph=True
+)
```
sayakpaul, line 138:
```suggestion
pipeline.unet = torch.compile(
    pipeline.unet, mode="max-autotune", fullgraph=True
)
```

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,231 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
 specific language governing permissions and limitations under the License.
 -->
 
-# Speed up inference
+# Accelerate inference
```
sayakpaul, line 13: I think we might want to set the expectations of the users right saying that the steps discussed in this page might need to be modified depending on the model being used. For example, for a pipeline that uses a transformer-based diffusion model may not benefit equally. 

Also, `docs/source/en/optimization/fp16.md` doesn't go well with "Accelerate interface" no? Should we change this?

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,231 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
 
-prompt = "a golden vase with different flowers"
-generator = torch.manual_seed(2023)
-image = distilled("a golden vase with different flowers", num_inference_steps=25, generator=generator).images[0]
-image
+apply_dynamic_quant(pipeline.unet, dynamic_quant_filter_fn)
+apply_dynamic_quant(pipeline.vae, dynamic_quant_filter_fn)
+
+prompt = "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k"
+pipeline(prompt, num_inference_steps=30).images[0]
```
sayakpaul, line 203: Could mention the `torchao` integration we have.

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,231 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
+> [!WARNING]
+> The [fuse_qkv_projections](https://github.com/huggingface/diffusers/blob/58431f102cf39c3c8a569f32d71b2ea8caa461e1/src/diffusers/pipelines/pipeline_utils.py#L2034) method is experimental and support is limited to mostly Stable Diffusion pipelines. Take a look at this [PR](https://github.com/huggingface/diffusers/pull/6179) to learn more about how to enable it for other pipelines
+
+An input is projected into three subspaces, represented by the projection matrices Q, K, and V, in an attention block. These projections are typically calculated separately, but you can horizontally combine these into a single matrix and perform the projection in a single step. It increases the size of the matrix multiplications of the input projections and also improves the impact of quantization.
+
+```py
+pipeline.fuse_qkv_projections()
+```
+
+## Distilled models
```
sayakpaul, line 217: We can remove these sections I believe. They don't feel relevant to me at least. WDYT?

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,231 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
+)
+pipeline.vae = AutoencoderTiny.from_pretrained(
+    "madebyollin/taesdxl", torch_dtype=torch.float16
+)
+pipeline = pipeline.to("cuda")
 
-More tiny autoencoder models for other Stable Diffusion models, like Stable Diffusion 3, are available from [madebyollin](https://huggingface.co/madebyollin).
\ No newline at end of file
+prompt = "slice of delicious New York-style cheesecake topped with berries, mint, chocolate crumble"
+pipeline(prompt, num_inference_steps=50).images[0]
```
sayakpaul, line 239: I think we should also have an advanced guide on accelerating inference that make use of other techniques like SAGE, multi-GPU inference (context-parallel), etc. @a-r-r-o-w WDYT?

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,241 @@ specific language governing permissions and limitations under the License.
...
 ```
 
-You may see a small performance boost in VAE decoding on multi-image batches, and there should be no performance impact on single-image batches.
-
-## Tiled VAE
+### Device placement
 
-Tiled VAE processing also enables working with large images on limited VRAM (for example, generating 4k images on 8GB of VRAM) by splitting the image into overlapping tiles, decoding the tiles, and then blending the outputs together to compose the final image. You should also used tiled VAE with [`~ModelMixin.enable_xformers_memory_efficient_attention`] to reduce memory use further if you have xFormers installed.
+> [!WARNING]
+> Device placement is an experimental feature and the API may change. Only the `balanced` strategy is supported at the moment. We plan to support additional mapping strategies in the future.
```
sayakpaul, line 61: Do we want to also talk about the model-level `device_map` support?
https://huggingface.co/docs/diffusers/en/training/distributed_inference

docs/source/en/optimization/memory.md
sayakpaul: SGTM but I think this won't block the merge.

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,241 @@ specific language governing permissions and limitations under the License.
...
+pipeline.vae.enable_group_offload(onload_device=onload_device, offload_type="leaf_level")
 
-# Uncomment the following to also allow recording the current streams.
-# pipe.transformer.enable_group_offload(onload_device=onload_device, offload_device=offload_device, offload_type="leaf_level", use_stream=True, record_stream=True)
-
-# For any other model implementations, the apply_group_offloading function can be used
-apply_group_offloading(pipe.text_encoder, onload_device=onload_device, offload_type="block_level", num_blocks_per_group=2)
-apply_group_offloading(pipe.vae, onload_device=onload_device, offload_type="leaf_level")
+# Use the apply_group_offloading method for other model components
+apply_group_offloading(pipeline.text_encoder, onload_device=onload_device, offload_type="block_level", num_blocks_per_group=2)
```
sayakpaul, line 249: @a-r-r-o-w should we add a note here noting that `num_blocks_per_group` should be 1 when using `use_stream=True`? Reference: https://github.com/huggingface/diffusers/pull/11375/files

docs/source/en/optimization/memory.md
```diff
@@ -386,48 +458,36 @@ class TracedUNet(torch.nn.Module):
...
 
-
-pipe.unet = TracedUNet()
+pipeline.unet = TracedUNet()
 
 with torch.inference_mode():
     image = pipe([prompt] * 1, num_inference_steps=50).images[0]
 ```
 
 ## Memory-efficient attention
```
sayakpaul, line 467: I think this only contributes to memory but also latency. Could we make a note? I think it might be a bit confusing for readers given we mention SDPA in the inference acceleration guide and mention memory-efficient attention here.

sayakpaul (commented):
Thank you for the further updates. Left some further comments. LMK if they make sense.

docs/source/en/optimization/fp16.md
stevhliu: Added note to manage expectations!

I think the title "Accelerate inference" should be good but I'm open to other suggestions as well :)

docs/source/en/optimization/fp16.md
stevhliu: Sounds good to me 👍

docs/source/en/optimization/memory.md
stevhliu: Added a note!

docs/source/en/optimization/fp16.md
stevhliu: Happy to add in a separate PR!

docs/source/en/optimization/fp16.md
```diff
@@ -10,120 +10,236 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
 
-<div class="flex justify-center">
-  <div>
-    <img class="rounded-xl" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/distilled_sd_vae.png" />
-    <figcaption class="mt-2 text-center text-sm text-gray-500">distilled Stable Diffusion + Tiny AutoEncoder</figcaption>
-  </div>
-</div>
+## Distilled models
+
+Another option for accelerating inference is to use a smaller distilled model if it's available. During distillation, many of the UNet's residual and attention blocks are discarded to reduce model size and improve latency. A distilled model is faster and uses less memory without compromising quality compared to a full-sized model.
```
sayakpaul, line 224: @stevhliu I thought we were removing these sections?

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,258 @@ specific language governing permissions and limitations under the License.
...
-pipe = pipe.to("cuda")
-prompt = "a beautiful landscape photograph"
-pipe.enable_vae_tiling()
-#pipe.enable_xformers_memory_efficient_attention()
-
-image = pipe([prompt], width=3840, height=2224, num_inference_steps=20).images[0]
 ```
 
-The output image has some tile-to-tile tone variation because the tiles are decoded separately, but you shouldn't see any sharp and obvious seams between the tiles. Tiling is turned off for images that are 512x512 or smaller.
+The `device_map` parameter also works on the model-level. This is useful for loading large models, such as the Flux diffusion transformer which has 12.5B parameters. Instead of `balanced`, set it to `"auto"` to automatically distribute a model across the fastest device first before moving to slower devices. Refer to the
```
sayakpaul, line 79: "Refer to the " this sounds incomplete?

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,258 @@ specific language governing permissions and limitations under the License.
...
-Offloading the weights to the CPU and only loading them on the GPU when performing the forward pass can also save memory. Often, this technique can reduce memory consumption to less than 3GB.
+transformer = AutoModel.from_pretrained(
+    "black-forest-labs/FLUX.1-dev", 
+    subfolder="transformer",
+    device_map="auto",
+    torch_dtype=torch.bfloat16
+)
+```
+
+You can inspect a pipeline's device map with `hf_device_map`.
```
sayakpaul, line 93: I think it would be better to move this print section above and move the model-level `device_map` note below it. WDYT?

sayakpaul (approved):
Looks crisp!

docs/source/en/optimization/fp16.md
a-r-r-o-w: SGTM. Not sure how we plan to show context parallel + Sage because the implementation is based on PR in finetrainers. Do we want to showcase that in the docs?

Or do you want me to take just the relevant parts and create a minimal attention processor to showcase the example? I can send over a rough writeup to Steven for whichever we choose to do

docs/source/en/optimization/memory.md
```diff
@@ -12,178 +12,258 @@ specific language governing permissions and limitations under the License.
...
-To enable group offloading, call the [`~ModelMixin.enable_group_offload`] method on the model if it is a Diffusers model implementation. For any other model implementation, use [`~hooks.group_offloading.apply_group_offloading`]:
+> [!WARNING]
+> Group offloading may not work with all models if the forward implementation contains weight-dependent device casting of inputs because it may clash with group offloading's device casting mechanism.
 
-```python
+Call [`~ModelMixin.enable_group_offload`] to enable it for standard Diffusers model components that inherit from [`ModelMixin`]. For other model components that don't inherit from [`ModelMixin`], such as a generic [torch.nn.Module](https://pytorch.org/docs/stable/generated/torch.nn.Module.html), use [`~hooks.apply_group_offloading`] instead.
+
+The `offload_type` parameter can be set to `block_level` or `leaf_level`.
+
+- `block_level` offloads groups of layers based on the `num_blocks_per_group` parameter. For example, if `num_blocks_per_group=2` on a model with 40 layers, 2 layers are onloaded and offloaded at a time (2o total onloads/offloads). This drastically reduces memory requirements.
```
a-r-r-o-w, line 248:
```suggestion
- `block_level` offloads groups of layers based on the `num_blocks_per_group` parameter. For example, if `num_blocks_per_group=2` on a model with 40 layers, 2 layers are onloaded and offloaded at a time (20 total onloads/offloads). This drastically reduces memory requirements.
```

docs/source/en/optimization/memory.md
a-r-r-o-w: Marking as resolved

docs/source/en/optimization/memory.md
```diff
@@ -193,48 +273,55 @@ prompt = (
...
-- When using `use_stream=True`, users can additionally specify `record_stream=True` to get better speedups at the expense of slightly increased memory usage. Refer to the [official PyTorch docs](https://pytorch.org/docs/stable/generated/torch.Tensor.record_stream.html) to know more about this.
+Set `record_stream=True` for more of a speedup at the cost of slightly increased memory usage. Refer to the [torch.Tensor.record_stream](https://pytorch.org/docs/stable/generated/torch.Tensor.record_stream.html) docs to learn more.
 
-For more information about available parameters and an explanation of how group offloading works, refer to [`~hooks.group_offloading.apply_group_offloading`].
+> [!TIP]
+> When `use_stream=True` on VAEs with tiling enabled, make sure to do a dummy forward pass (possible with dummy inputs as well) before inference to avoid device mismatch errors. This may not work on all implementations, so feel free to open an issue if you encounter any problems.
 
-</Tip>
+The `num_blocks_per_group` parameter should be set to `1` if `use_stream` is enabled.
```
a-r-r-o-w, line 290:
```suggestion
If using `block_level` group offloading with `use_stream` enabled, the `num_blocks_per_group` parameter should be set to `1` otherwise a warning will be raised.
```

a-r-r-o-w (approved):
Thanks, my earlier comments have been addressed so LGTM apart from remaining Sayak's comments!

docs/source/en/optimization/fp16.md
sayakpaul: Hmm got it. Sorry for being unclear. 

If it's not too much would it be possible to send over a write-up covering both cases? And we can then further brainstorm how to best disseminate the information to the readers? But really no rush here. Do it whenever you get time.

docs/source/en/optimization/fp16.md
stevhliu: My bad, thought I already had! Should be gone now :)

docs/source/en/optimization/memory.md
stevhliu: Lol good eyes! 👀

PR #11401, opened 2025-04-23
[Feature] AutoModel can load components using model_index.json

### LINKED ISSUE #11388: extend AutoModel to be able to load transformer models & custom diffusers models
current AutoModel implementation only support models that importable from diffusers https://github.com/huggingface/diffusers/blob/448c72a230ceb31ae60e70e3b62e327314928a5e/src/diffusers/models/auto_model.py#L164

it uses this field in model config.json https://huggingface.co/HiDream-ai/HiDream-I1-Full/blob/main/transformer/config.json#L2

we can use info in model_index.json instead, https://huggingface.co/HiDream-ai/HiDream-I1-Full/blob/main/model_index.json
this way we can load anything that can be load in `from_pretrained`, include transformer models & custom diffusers moduels (not directly importable from top level)

can reference the code/logic in `from_pretrained` https://github.com/huggingface/diffusers/blob/448c72a230ceb31ae60e70e3b62e327314928a5e/src/diffusers/pipelines/pipeline_utils.py#L893 

I will work on this shortly, but feel free to pick this up if anyone in the community is interested!

ishan-modi:
# What does this PR do?

Allows `AutoModel` to load components using `model_index.json`. If `model_index.json` is not found `config.json`(existing) is used as a fallback

Fixes #11388 

## Who can review?

@yiyixuxu

### DIFF
--- a/src/diffusers/models/auto_model.py
+++ b/src/diffusers/models/auto_model.py
@@ -156,10 +156,14 @@ class AutoModel(ConfigMixin):
             "subfolder": subfolder,
         }
 
-        config = cls.load_config(pretrained_model_or_path, **load_config_kwargs)
-        orig_class_name = config["_class_name"]
-
-        library = importlib.import_module("diffusers")
+        try:
+            config = cls.load_config(os.path.join(pretrained_model_or_path, "model_index.json"), **load_config_kwargs)
+            library, orig_class_name = config["subfolder"]
+        except Exception:
+            # Fallback to loading the config from the config.json file
+            config = cls.load_config(os.path.join(pretrained_model_or_path, cls.config_name), **load_config_kwargs)
+            library = importlib.import_module("diffusers")
+            orig_class_name = config["_class_name"]
 
         model_cls = getattr(library, orig_class_name, None)
         if model_cls is None:


yiyixuxu:
thanks @ishan-modi !
can you run on the testing script here https://github.com/huggingface/diffusers/issues/11388#issuecomment-2825382670

ishan-modi:
On it

src/diffusers/models/auto_model.py
```diff
@@ -153,13 +153,17 @@ def from_pretrained(cls, pretrained_model_or_path: Optional[Union[str, os.PathLi
...
-            "subfolder": subfolder,
         }
 
-        config = cls.load_config(pretrained_model_or_path, **load_config_kwargs)
-        orig_class_name = config["_class_name"]
-
-        library = importlib.import_module("diffusers")
+        try:
+            config = cls.load_config(os.path.join(pretrained_model_or_path, "model_index.json"), **load_config_kwargs)
+            library, orig_class_name = config["subfolder"]
```
yiyixuxu, line 160: probably this?
```suggestion
            library, orig_class_name = config[subfolder]
```

ishan-modi:
should be working now !

src/diffusers/models/auto_model.py
```diff
@@ -22,7 +22,7 @@
 
 
 class AutoModel(ConfigMixin):
-    config_name = "config.json"
+    config_name = "model_index.json"
```
yiyixuxu, line 25:
```suggestion
```
this should not changed, the model config should still be saved in a `config.json`
maybe, you can try to use this to download model_index https://github.com/huggingface/diffusers/blob/main/src/diffusers/loaders/single_file.py#L242

yiyixuxu (commented):
currently test1 and 3 still fail

src/diffusers/models/auto_model.py
ishan-modi: My bad, completely missed it. Let me know if the testcases work now. I think they were failing earlier because I didn't have latest changes in the branch

src/diffusers/models/auto_model.py
```diff
@@ -156,10 +157,18 @@ def from_pretrained(cls, pretrained_model_or_path: Optional[Union[str, os.PathLi
...
-        orig_class_name = config["_class_name"]
-
-        library = importlib.import_module("diffusers")
+        try:
+            mindex_kwargs = {k: v for k, v in load_config_kwargs.items() if k != "subfolder"}
+            mindex_kwargs["filename"] = "model_index.json"
+            config_path = hf_hub_download(pretrained_model_or_path, **mindex_kwargs)
+            config = cls.load_config(config_path, **load_config_kwargs)
+            library, orig_class_name = config[subfolder]
+            library = importlib.import_module(library)
```
yiyixuxu, line 166: does this work for custom modules? like this

```python
# test3: load a custom diffusers model 
# https://huggingface.co/Kwai-Kolors/Kolors-diffusers/blob/main/model_index.json#L9
try:
    model = AutoModel.from_pretrained("Kwai-Kolors/Kolors-diffusers", subfolder="text_encoder", torch_dtype=torch.bfloat16)
    print(f"test3 passed!")
except Exception as e:
    print(f"test3 failed: {e}")
```


you can probably use this https://github.com/huggingface/diffusers/blob/4b60f4b60293bff0f132179a4d68ba585b360a41/src/diffusers/pipelines/pipeline_loading_utils.py#L334

ishan-modi:
Thanks for the patience, all test cases work now

<details>
<summary> <b>Script</b></summary>

```python
from diffusers import AutoModel 
import torch


# test1: load a diffusers model
try:
    model = AutoModel.from_pretrained(
        "Efficient-Large-Model/Sana_Sprint_0.6B_1024px_diffusers", subfolder="transformer", torch_dtype=torch.bfloat16, 
    )
    print(f"test1 passed!")
except Exception as e:
    print(f"test1 failed: {e}")


# test2: load a non-diffusers model
try:
    model = AutoModel.from_pretrained(
        "HiDream-ai/HiDream-I1-Full", subfolder="text_encoder", torch_dtype=torch.bfloat16, 
    )
    print(f"test2 passed!")
except Exception as e:
    print(f"test2 failed: {e}")


# test3: load a custom diffusers model 
# https://huggingface.co/Kwai-Kolors/Kolors-diffusers/blob/main/model_index.json#L9
try:
    model = AutoModel.from_pretrained("Kwai-Kolors/Kolors-diffusers", subfolder="text_encoder", variant="fp16")
    print(f"test3 passed!")
except Exception as e:
    print(f"test3 failed: {e}")


# test4: load a model directly (not subfolder)
controlnet_repo = "InstantX/SD3-Controlnet-Canny"
try:
    controlnet_model = AutoModel.from_pretrained(
        controlnet_repo, revision="refs/pr/12"
    )
    print(f"test4 passed!")
except Exception as e:
    print(f"test4 failed: {e}")
```
</details>

Please let me know if anything seems incorrect !

src/diffusers/pipelines/pipeline_loading_utils.py
```diff
@@ -332,7 +332,7 @@ def maybe_raise_or_warn(
 
 
 def get_class_obj_and_candidates(
-    library_name, class_name, importable_classes, pipelines, is_pipeline_module, component_name=None, cache_dir=None
+    library_name, class_name, importable_classes, pipelines, is_pipeline_module, component_name, cache_dir
```
yiyixuxu, line 335: umm I think we should update the code itself so that component_name and cache_dir can be `None` - they are only needed for custom code, which we don't support yet with `AutoModel`

yiyixuxu (commented):
I left some comments
but can confirm all tests pass now!

src/diffusers/models/auto_model.py
```diff
@@ -156,12 +159,28 @@ def from_pretrained(cls, pretrained_model_or_path: Optional[Union[str, os.PathLi
...
+            model_cls, _ = get_class_obj_and_candidates(
+                library_name=library,
+                class_name=orig_class_name,
+                importable_classes=ALL_IMPORTABLE_CLASSES,
+                pipelines=pipelines,
+                is_pipeline_module=hasattr(pipelines, library),
+                component_name=subfolder,
+                cache_dir=constants.HF_HUB_CACHE,
+            )
+        except Exception:
```
sayakpaul, line 177: We should catch the specific Exception here instead of making it generic. This will help eliminate other side-effects.

sayakpaul (commented):
Looking pretty cool. Let's also do a test for this.

src/diffusers/models/auto_model.py
```diff
@@ -16,9 +16,12 @@
 import os
 from typing import Optional, Union
 
+from huggingface_hub import constants, hf_hub_download
 from huggingface_hub.utils import validate_hf_hub_args
 
+from .. import pipelines
```
sayakpaul, line 22: I think if we move this import inside the `try` block, we should be able to get rid of the circular import problem: https://github.com/huggingface/diffusers/actions/runs/14787632114/job/41518909088?pr=11401#step:15:68

src/diffusers/pipelines/pipeline_loading_utils.py
sayakpaul: I just tried setting `component_name` and `cache_dir` to None values and ran the tests from https://github.com/huggingface/diffusers/pull/11401#issuecomment-2837944821 and they ran fine for me.

yiyixuxu: what do you mean? the first line would already throw an error, no? 
`os.path.join(None,None)`

sayakpaul: I meant when I changed to the original code:
https://github.com/huggingface/diffusers/blob/e23705e5577387872dd55ebf6db81bd59df928f1/src/diffusers/pipelines/pipeline_loading_utils.py#L335

It didn't error out for me 👀

ishan-modi: Oh, they should run fine even without it, because of [this line](https://github.com/ishan-modi/diffusers/blob/fixes-issue-11388/src/diffusers/models/auto_model.py#L168-L176)

src/diffusers/models/auto_model.py
```diff
@@ -156,12 +159,28 @@ def from_pretrained(cls, pretrained_model_or_path: Optional[Union[str, os.PathLi
...
+            config_path = hf_hub_download(pretrained_model_or_path, **mindex_kwargs)
+            config = cls.load_config(config_path, **load_config_kwargs)
+            library, orig_class_name = config[subfolder]
+            model_cls, _ = get_class_obj_and_candidates(
+                library_name=library,
+                class_name=orig_class_name,
+                importable_classes=ALL_IMPORTABLE_CLASSES,
+                pipelines=pipelines,
+                is_pipeline_module=hasattr(pipelines, library),
+                component_name=subfolder,
```
yiyixuxu, line 174:
```suggestion
                component_name=None,
```

src/diffusers/models/auto_model.py
```diff
@@ -156,12 +159,28 @@ def from_pretrained(cls, pretrained_model_or_path: Optional[Union[str, os.PathLi
...
+                cache_dir=constants.HF_HUB_CACHE,
```
yiyixuxu, line 175:
```suggestion
                cache_dir=None,
```
let's jjust pass None here since we ddon't actually have a cached_dir so not very meaningful
and update the other function so it works with cache_dir=None and component_name=NOne

src/diffusers/models/auto_model.py
```diff
@@ -156,12 +159,28 @@ def from_pretrained(cls, pretrained_model_or_path: Optional[Union[str, os.PathLi
             "subfolder": subfolder,
         }
 
-        config = cls.load_config(pretrained_model_or_path, **load_config_kwargs)
-        orig_class_name = config["_class_name"]
+        try:
+            mindex_kwargs = {k: v for k, v in load_config_kwargs.items() if k != "subfolder"}
```
yiyixuxu, line 163: move this section under a `if subfolder is not None`:

```python
if subfolder is not None:
     try: 
          ...
```

I think we are not supporting local path for `pretrained_model_or_path` here, are we? it would be a bit more complex if we do

src/diffusers/pipelines/pipeline_loading_utils.py
yiyixuxu: well, it is wrapped around a try/exception.....
but what I meant we should update this function so that componenet_name and cache_dir aree optional argument (it is meant to be, they aree only needed for custom code)

src/diffusers/models/auto_model.py
ishan-modi: What would be the complexity related to supporting local path here ? shouldn't we also be able to load models from structure where there is no folder heirarchy using this code

ishan-modi: ohk, so we dont want to support flat repo's like [this](https://huggingface.co/ishan24/Sana_600M_1024px_ControlNet_diffusers/tree/main) to load models

Because doing the following would mean that we dont support above case  
```python

if subfolder is not None:
     try: 
          ...
```

yiyixuxu: ohh no we need to support flat repos, we need to all repos (we don't need to support some edge cases with local path when there is a subfolder is included in the path, not as `subfolder`, you cannot do that with hub repo ids I think)

so basically here
if subfolder is not None -> we try model_index.json approach first, if that fails, we try the config.json approach
if subfolder is None -> we try the config.json approach (I think it is not meaningful to use model_index.json here because you need use subfolder as name to locate the info there, no?)

ishan-modi: got it, made the change

src/diffusers/models/auto_model.py
```diff
@@ -156,12 +158,38 @@ def from_pretrained(cls, pretrained_model_or_path: Optional[Union[str, os.PathLi
             "subfolder": subfolder,
         }
 
-        config = cls.load_config(pretrained_model_or_path, **load_config_kwargs)
```
DN6, line 159: Would avoid using exceptions for control flow and simplify this a bit

```python
        load_config_kwargs = {
            "cache_dir": cache_dir,
            "force_download": force_download,
            "proxies": proxies,
            "token": token,
            "local_files_only": local_files_only,
            "revision": revision,
        }

        library = None
        orig_class_name = None
        from diffusers import pipelines

        # Always attempt to fetch model_index.json first
        try:
            cls.config_name = "model_index.json"
            config = cls.load_config(pretrained_model_or_path, **load_config_kwargs)

            if subfolder is not None and subfolder in config:
                library, orig_class_name = config[subfolder]

        except EntryNotFoundError as e:
            logger.debug(e)

        # Unable to load from model_index.json so fallback to loading from config
        if library is None and orig_class_name is None:
            cls.config_name = "config.json"
            load_config_kwargs.update({"subfolder": subfolder})

            config = cls.load_config(pretrained_model_or_path, **load_config_kwargs)
            orig_class_name = config["_class_name"]
            library = "diffusers"

        model_cls, _ = get_class_obj_and_candidates(
            library_name=library,
            class_name=orig_class_name,
            importable_classes=ALL_IMPORTABLE_CLASSES,
            pipelines=pipelines,
            is_pipeline_module=hasattr(pipelines, library), 
        )
```

DN6:
@yiyixuxu Loading from model_index.json would introduce a slightly weird behaviour where Diffusers AutoModel can load transformer models for repos with model_index.json files but not directly. 

Perhaps we also raise a warning that Diffusers AutoModel cannot loading transformers directly? Or attempt to add logic to do that?

ishan-modi:
> Perhaps we also raise a warning that Diffusers AutoModel cannot loading transformers directly? Or attempt to add logic to do that?

Good point and great suggestions !

src/diffusers/models/auto_model.py
```diff
@@ -153,15 +156,60 @@ def from_pretrained(cls, pretrained_model_or_path: Optional[Union[str, os.PathLi
...
+                        logger.warning(
+                            f"Found multiple architectures in {pretrained_model_or_path}. Using the first one: {config['architectures'][0]}"
+                        )
+                    orig_class_name = config["architectures"][0]
+                    library = "transformers"
+                    load_config_kwargs.update({"subfolder": "" if subfolder is None else subfolder})
+                else:
+                    raise ValueError(
+                        f"Couldn't find model associated with the config file at {pretrained_model_or_path}."
+                    )
```
yiyixuxu, line 203:
```suggestion
            elif "model_type" in config:
                logger.warning(
                    f"Loading {config[model_type]} as a transformer model from {pretrained_model_or_path}."
                )
                from transformers import AutoModel
                # we can use the AutoModel from tranformers here I think? 
                ....
             else:
                    raise ValueError(...)
```

ishan-modi: thanks @yiyixuxu, using `AutoModel` will allow us to only load from [https://github.com/huggingface/transformers/blob/main/src/transformers/models/auto/modeling_auto.py#L32](this mapping). There are many other mappings in the file that we might want to load from hence I am using the architecture.

for example,

**AutoModelForCausalLM** - helps us import architecture like `MarianForCausalLM` while AutoModel doesn't

Let me know if we want to ignore all the other mappings and just use AutoModel for import ?

yiyixuxu: cc @DN6  can you also let me know do you think here?

my opinion is that we should not implement our own logic to load transformer models, so ok to either dispatch to their AutoModel or throw a warning for not supporting 

but open to other thoughts :)

ishan-modi: Made the change !

yiyixuxu (approved):
thanks!

yiyixuxu:
I think we have a circular import issue though

ishan-modi:
@bot /style

PR #11434, opened 2025-04-28
[tests] fix import.

sayakpaul:
# What does this PR do?

Not really sure how it was passed.

### DIFF
--- a/tests/models/test_modeling_common.py
+++ b/tests/models/test_modeling_common.py
@@ -71,6 +71,7 @@ from diffusers.utils.testing_utils import (
     require_torch_2,
     require_torch_accelerator,
     require_torch_accelerator_with_training,
+    require_torch_gpu,
     require_torch_multi_accelerator,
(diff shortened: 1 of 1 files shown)


sayakpaul:
Should be a simple import change. Merging.

yao-matrix:
the reason is:
In PR https://github.com/huggingface/diffusers/pull/11085, new `TorchCompileTesterMixin` cases were added w/ `require_torch_gpu` not not merged.
Later PR https://github.com/huggingface/diffusers/pull/11406 was submitted to enable all gpu only cases to work on XPU, so using `require_torch_accelerator` and remove `require_torch_gpu` per ruff since it's not referenced anymore.

But there is a time gap btw these 2 cases merging(2 hours) and it's no conflict btw these PRs, since the later PR didn't change `TorchCompileTesterMixin`(even not aware of it), and static diff tool cannot detect such situation, since it's a runtime dependency loss.

Issue #11443, opened 2025-04-29
model_cpu_offload failed in unidiffusers pipeline


yao-matrix:
### Describe the bug

unidiffusers cpu_offload failed with the log in Reproduction column.

I took a deeper look, it seems that in this case, [self.text_decoder.encode](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/unidiffuser/pipeline_unidiffuser.py#L1288) will be called after `text_encoder` and before `image_encoder`. The thing is this is just a submodule of `text_decoder` and not in [`model_cpu_offload_seq`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/unidiffuser/pipeline_unidiffuser.py#L100), so didn't register hook while [`enable_model_cpu_offload`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/pipeline_utils.py#L1092). It became an orphan. I don't have a good idea to fix it, since it's an embeded submodule in a sub-model and whether to trigger it is a runtime decision based on [`reduce_text_emb_dim`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/unidiffuser/pipeline_unidiffuser.py#L1243). But I'm willing to contribute to the fix of it.

@yiyixuxu @sayakpaul @DN6 

### Reproduction

`pytest -rA tests/pipelines/unidiffuser/test_unidiffuser.py::UniDiffuserPipelineFastTests::test_model_cpu_offload_forward_pass` you can see below error log. The same issue happens on CUDA too.

> 
self = Linear(in_features=32, out_features=32, bias=True)
input = tensor([[[-0.8407, -0.3964, -0.6832,  ..., -0.2908,  0.1523, -1.0043],
         [-0.8155, -0.1579,  0.6659,  ...,  1.4...375, -0.4626, -0.3352],
         [-1.2005, -0.1820,  0.4218,  ..., -0.3822, -0.5105, -0.2234]]],
       device='xpu:0')

    def forward(self, input: Tensor) -> Tensor:
        # print(f"input.device: {input.device}, weight device: {self.weight.device}")
>       return F.linear(input, self.weight, self.bias)
E       RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and xpu:0! (when checking argument for argument mat1 in method wrapper_XPU_addmm)

### Logs

```shell

```

### System Info

N/A

sayakpaul:
[...]

PR #9399, opened 2024-09-09
make `pipelines` tests device-agnostic (part1)

faaany:
## What does this PR do?
Below are some evidences:

```bash
PASSED tests/pipelines/animatediff/test_animatediff.py::AnimateDiffPipelineFastTests::test_to_device
PASSED tests/pipelines/amused/test_amused.py::AmusedPipelineFastTests::test_model_cpu_offload_forward_pass
PASSED tests/pipelines/amused/test_amused.py::AmusedPipelineFastTests::test_cpu_offload_
forward_pass_twice
PASSED tests/pipelines/amused/test_amused_inpaint.py::AmusedInpaintPipelineFastTests::te
st_to_device
```
@yiyixuxu

PR #11444, opened 2025-04-29
enable unidiffuser test cases on xpu

yao-matrix:
w/ `pytest -rA tests/pipelines/unidiffuser/test_unidiffuser.py`, all cases passed except below 2:

> tests/pipelines/unidiffuser/test_unidiffuser.py::UniDiffuserPipelineFastTests::test_cpu_offload_forward_pass_twice
    tests/pipelines/unidiffuser/test_unidiffuser.py::UniDiffuserPipelineFastTests::test_model_cpu_offload_forward_pass

these 2 cases fail both on cuda and xpu, track through this issue https://github.com/huggingface/diffusers/issues/11443.

### DIFF
--- a/tests/pipelines/test_pipelines_common.py
+++ b/tests/pipelines/test_pipelines_common.py
@@ -1485,8 +1485,8 @@ class PipelineTesterMixin:
         model_devices = [component.device.type for component in components.values() if hasattr(component, "device")]
         self.assertTrue(all(device == torch_device for device in model_devices))
 
-        output_cuda = pipe(**self.get_dummy_inputs(torch_device))[0]
-        self.assertTrue(np.isnan(to_np(output_cuda)).sum() == 0)
+        output_device = pipe(**self.get_dummy_inputs(torch_device))[0]
+        self.assertTrue(np.isnan(to_np(output_device)).sum() == 0)
 
     def test_to_dtype(self):
         components = self.get_dummy_components()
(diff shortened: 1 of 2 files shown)


yao-matrix:
@a-r-r-o-w @DN6 , pls help review, thx.

a-r-r-o-w (approved):
Thanks!

yao-matrix:
the failed cases seem not related to my changes.

tests/pipelines/test_pipelines_common.py
```diff
@@ -1677,11 +1677,11 @@ def test_cpu_offload_forward_pass_twice(self, expected_max_diff=2e-4):
 
         pipe.set_progress_bar_config(disable=None)
 
-        pipe.enable_model_cpu_offload(device=torch_device)
+        pipe.enable_model_cpu_offload()
```
a-r-r-o-w, line 1680: Was going to merge but just realized this was removed. Change looks good, but is there a particular reason for this to be removed for XPU case? The `enable_model_cpu_offload` method correctly gets passed torch_device="xpu" here, no?

yao-matrix: the pass value is correct. I changed it just because this is introduced in this PR https://github.com/huggingface/diffusers/pull/9399 to make put explicit `torch_device` to `enable_model_cpu_offload` to try to make it work on device other than CUDA(which is XPU in that case). But it doesn't work because diffusers internal logic also call `enable_model_cpu_offload` which cannot fixed by only setting device in application code, I fixed the issue in this PR https://github.com/huggingface/diffusers/pull/11288.

So, with the  https://github.com/huggingface/diffusers/pull/11288, we don't need application/test code to explicitly set the device, so I changed back such changes in this test back to original. And another motivation I changed back is: I can see this usage is the recommended way in diffusers docs, so to align w/ the recommendation and test the most common used pattern.

To be short, in current codebase, both(with `torch_device` and without `torch_device`) both work, I changed back the way in original test code. 

Hope it explains. thx.

yao-matrix:
@a-r-r-o-w , pls feel free to let me know if you need more clarification or you have more comments, thx

tests/pipelines/test_pipelines_common.py
a-r-r-o-w: Thanks for the explanation. I've triggered the tests and will merge once they pass

yao-matrix:
seems my rebase to main interrupted your test, sorry, pls help re-trigger.

PR #11450, opened 2025-04-29
[Refactor] Initialize class members inside init method

### LINKED ISSUE #11002: Any chance class members like self._interrupt could be defined in __init__ across pipelines?
### Describe the bug

I think there is no benefit to late initializing here and it puts a burden on the library user that could be easily avoided. Also leads to some confusion as it is uncommon, code inspection flags this. Let me know if I'm missing something.

### Reproduction

```
class WanImageToVideoPipeline:
	def __init__(self):
		pass
	
	def __call__(self, *args, **kwargs):
		self._interrupt = False
		return 23

	@property
	def interrupt(self):
		return self._interrupt
	
pipe = WanImageToVideoPipeline()

def on_async_user_abort_call_me_any_time():
	# check if already interupted but mid step
	print(pipe.interrupt)


on_async_user_abort_call_me_any_time()
```

### Logs

```shell
AttributeError: 'WanImageToVideoPipeline' object has no attribute '_interrupt'. Did you mean: 'interrupt'?
```

### System Info

Diffusers 0.33.0.dev0, Linux, Python 3.10

### Who can help?

@yiyixuxu @DN6

ishan-modi:
# What does this PR do?

Fixes #11002 

## Who can review?

@DN6 @yiyixuxu

### DIFF
--- a/src/diffusers/pipelines/allegro/pipeline_allegro.py
+++ b/src/diffusers/pipelines/allegro/pipeline_allegro.py
@@ -198,6 +198,10 @@ class AllegroPipeline(DiffusionPipeline):
         scheduler: KarrasDiffusionSchedulers,
     ):
         super().__init__()
+        self._guidance_scale=1.0
+        self._num_timesteps=0
+        self._current_timestep=None
+        self._interrupt=False
 
         self.register_modules(
             tokenizer=tokenizer, text_encoder=text_encoder, vae=vae, transformer=transformer, scheduler=scheduler
@@ -819,8 +823,6 @@ class AllegroPipeline(DiffusionPipeline):
             negative_prompt_attention_mask,
         )
         self._guidance_scale = guidance_scale
-        self._current_timestep = None
-        self._interrupt = False
 
         # 2. Default height and width to transformer
         if prompt is not None and isinstance(prompt, str):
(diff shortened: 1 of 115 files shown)


src/diffusers/pipelines/allegro/pipeline_allegro.py
```diff
@@ -198,6 +198,10 @@ def __init__(
         scheduler: KarrasDiffusionSchedulers,
     ):
         super().__init__()
+        self._guidance_scale = 1.0
```
yiyixuxu: i would suggest:

1 .default _guidance_scale to be same as each pipeline __call__ default, e.g. 7.5 here

2 _num_stesps can be None here since it is not yet known 


but let me know all your thoughts @DN6 @ishan-modi

yiyixuxu:
hi @ishan-modi 
are you still available on working on this? if not, we can open this to the community as well

let me know!

ishan-modi:
Hi @yiyixuxu, thanks for reviving this MR, as requested I have updated **guidance_scale** in `__init__` to match `__call__` defaults

let me know if you think it requires any other change

yiyixuxu:
can you also take a look here? @DN6

ishan-modi:
@yiyixuxu, gentle ping

PR #11452, opened 2025-04-29
[quant] add __repr__ for better printing of configs.

sayakpaul:
# What does this PR do?

Adds repr for GGUF and Quanto config classes.


sayakpaul:
@DN6

PR #11453, opened 2025-04-29
cache packages_distributions

vladmandic:
pr <https://github.com/huggingface/diffusers/pull/11161> refactored `_is_package_available`,  
but it introduced a major performance regression.    
diffusers load now takes up to **~5sec** while it was taking **~0.5sec** earlier.  

```py
import time
t0 = time.time()

import torch # preload torch so its not part of diffusers import time
t1 = time.time()
print(f'torch: {t1-t0:.3f}')

import diffusers.utils.import_utils
t2 = time.time()
print(f'utils: {t2-t1:.3f}')
```

this is due to very expensive call to `importlib_metadata.packages_distributions()` which is done for every single installed package (each call can be ~0.2sec and it quickly adds up).

this pr simply changes the logic so the call is done once and cached.

result is that diffusers startup is back to 0.5sec as usual.

cc @yiyixuxu @sayakpaul @a-r-r-o-w


yiyixuxu (approved):
thanks!

yiyixuxu:
@DN6 could you take a look here too?

DN6 (approved):
Small nit. Could we make `package_map` and internal variable `_package_map`. Good to merge after.

vladmandic:
@DN6 can you make the edits as you like? I'm really not sure on exactly what do you want as that introduces second variable that is not accessed anywhere else.

PR #11481, opened 2025-05-02
[tests] chore: rename lora model-level tests.

sayakpaul:
# What does this PR do?

Just renames model-level LoRA tests so that they get picked up in our PR CI:

https://github.com/huggingface/diffusers/blob/e23705e5577387872dd55ebf6db81bd59df928f1/.github/workflows/pr_tests.yml#L285-L288


sayakpaul:
The failing test passes locally for me. Quite possibly flaky.

PR #11493, opened 2025-05-05
[LoRA] use `removeprefix` to preserve sanity.

sayakpaul:
# What does this PR do?

We have multiple recurrences of using the following patterns in our LoRA codebase:

* `k.replace(f"{prefix}.", "")`
* `k[len(f"{prefix}.") :]`

It's just mind-twisting sometimes and I think `removeprefix()` just makes it easier.

I have run the integration tests and ensured they are passing.

### DIFF
--- a/src/diffusers/loaders/lora_base.py
+++ b/src/diffusers/loaders/lora_base.py
@@ -348,7 +348,7 @@ def _load_lora_into_text_encoder(
 
     # Load the layers corresponding to text encoder and make necessary adjustments.
     if prefix is not None:
-        state_dict = {k[len(f"{prefix}.") :]: v for k, v in state_dict.items() if k.startswith(f"{prefix}.")}
+        state_dict = {k.removeprefix(prefix + "."): v for k, v in state_dict.items() if k.startswith(f"{prefix}.")}
 
     if len(state_dict) > 0:
         logger.info(f"Loading {prefix}.")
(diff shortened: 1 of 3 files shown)


src/diffusers/loaders/lora_base.py
```diff
@@ -348,7 +348,7 @@ def _load_lora_into_text_encoder(
 
     # Load the layers corresponding to text encoder and make necessary adjustments.
     if prefix is not None:
-        state_dict = {k[len(f"{prefix}.") :]: v for k, v in state_dict.items() if k.startswith(f"{prefix}.")}
+        state_dict = {k.removeprefix(prefix + "."): v for k, v in state_dict.items() if k.startswith(f"{prefix}.")}
```
BenjaminBossan, line 351: Previous code used `f"{prefix}."` instead of `prefix + "."`, either works of course, just mentioning it.

BenjaminBossan (approved):
LGTM, thanks.

Just for other reviewers: The change is not equivalent, as this will strictly check for prefixes, i.e. the new code corresponds to `k[len(f"{prefix}.") :] if k.startswith(prefix) else k`. Also, `removeprefix` was introduced in Python 3.9, but 3.8 is EOL so this should be fine.

src/diffusers/loaders/lora_base.py
sayakpaul: Yes, correct. Should be pretty safe know?

sayakpaul:
> The change is not equivalent, as this will strictly check for prefixes, i.e. the new code corresponds to k[len(f"{prefix}.") :] if k.startswith(prefix) else k. 

@BenjaminBossan also to be clear that in our codebase `prefix` is quite exclusively used to handle what you meant. So, IMO, `removeprefix()` is a better alternative to tackle that. Am I mistaken to think that?

src/diffusers/loaders/lora_base.py
a-r-r-o-w: (nit) Would prefer the f-string here instead of `+`, because it's the same as before that way and we use f-string at the end of this line anyway

a-r-r-o-w: (nit) we don't need to check startswith if using removeprefix, no?

a-r-r-o-w (approved):
Just some nits. Personal preference is to use f-strings but either is okay

sayakpaul:
Fix for the failing tests https://github.com/huggingface/diffusers/pull/11504.

PR #11498, opened 2025-05-05
Fix huggingface-hub package name.

rasmi:
This checks for the actual Python package name, `huggingface-hub`.


src/diffusers/utils/testing_utils.py
```diff
@@ -555,10 +555,10 @@ def decorator(test_case):
 def require_hf_hub_version_greater(hf_hub_version):
     def decorator(test_case):
         correct_hf_hub_version = version.parse(
-            version.parse(importlib.metadata.version("huggingface_hub")).base_version
```
yiyixuxu, line 558: is there any advantage to use the "huggingface-hub" here?

PR #11503, opened 2025-05-06
enable 7 cases on XPU

yao-matrix:
@a-r-r-o-w  @DN6 , pls help review, thx.

### DIFF
--- a/tests/pipelines/consisid/test_consisid.py
+++ b/tests/pipelines/consisid/test_consisid.py
@@ -24,9 +24,10 @@ from transformers import AutoTokenizer, T5EncoderModel
 from diffusers import AutoencoderKLCogVideoX, ConsisIDPipeline, ConsisIDTransformer3DModel, DDIMScheduler
 from diffusers.utils import load_image
 from diffusers.utils.testing_utils import (
+    backend_empty_cache,
     enable_full_determinism,
     numpy_cosine_similarity_distance,
-    require_torch_gpu,
+    require_torch_accelerator,
     slow,
     torch_device,
 )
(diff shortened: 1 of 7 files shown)


tests/pipelines/consisid/test_consisid.py
```diff
@@ -338,8 +339,8 @@ def test_consisid(self):
 
         prompt = self.prompt
         image = load_image("https://github.com/PKU-YuanGroup/ConsisID/blob/main/asserts/example_images/2.png?raw=true")
-        id_vit_hidden = [torch.ones([1, 2, 2])] * 1
-        id_cond = torch.ones(1, 2)
+        id_vit_hidden = [torch.ones([1, 577, 1024])] * 5
+        id_cond = torch.ones(1, 1280)
```
yao-matrix, line 343: the size of the `id_vit_hidden` and `id_cond` is wrong, both A100 and XPU will report below error

>     def forward(self, input: Tensor) -> Tensor:
>       return F.linear(input, self.weight, self.bias)
>         E       RuntimeError: mat1 and mat2 shapes cannot be multiplied (1x2 and 1280x1024)

I checked the needed size for these 2 tensors and use the correct tensor

tests/pipelines/consisid/test_consisid.py
```diff
@@ -357,5 +358,5 @@ def test_consisid(self):
         video = videos[0]
         expected_video = torch.randn(1, 16, 480, 720, 3).numpy()
 
-        max_diff = numpy_cosine_similarity_distance(video, expected_video)
+        max_diff = numpy_cosine_similarity_distance(video.cpu(), expected_video)
```
yao-matrix, line 361: need to cpu, since `expected_video` is in cpu,  `numpy_cosine_similarity_distance` only supports 2 tensors in same device.

yao-matrix:
@DN6 @a-r-r-o-w , pls help review and comment, thx very much.

a-r-r-o-w (approved):
thanks, lgtm!

PR #11517, opened 2025-05-07
Feat/multispectral dataloader 5ch

snwtz:
- new modules: dataloader and dataloader test (local, not optimized for GPU)
- multispectral data loading and preprocessing
- TODO: explicit caching and order validation

- adapted main training script for MS data loading


yiyixuxu:
can you explain what's the use case for this?

PR #11523, opened 2025-05-08
enable dit integration cases on xpu

yao-matrix:
for `test_dit_512` case, both A100 and XPU don't pass with current pixel delta criteria(yes, as we noticed many times, there 2 have similar numerical behavior), I propose to use cosine similarity as some of other cases used(like `controlnet_impaint`). With proposed changes, both XPU and A100 passed.

@DN6 @a-r-r-o-w , pls help review, thx.


a-r-r-o-w (approved):
thanks!

PR #11534, opened 2025-05-09
Hunyuan Video Framepack F1

a-r-r-o-w:
Example:

```python
import torch
from diffusers import HunyuanVideoFramepackPipeline, HunyuanVideoFramepackTransformer3DModel
from diffusers.utils import export_to_video, load_image
from transformers import SiglipImageProcessor, SiglipVisionModel

transformer = HunyuanVideoFramepackTransformer3DModel.from_pretrained(
    "lllyasviel/FramePack_F1_I2V_HY_20250503", torch_dtype=torch.bfloat16
)
feature_extractor = SiglipImageProcessor.from_pretrained(
    "lllyasviel/flux_redux_bfl", subfolder="feature_extractor"
)
image_encoder = SiglipVisionModel.from_pretrained(
    "lllyasviel/flux_redux_bfl", subfolder="image_encoder", torch_dtype=torch.float16
)
pipe = HunyuanVideoFramepackPipeline.from_pretrained(
    "hunyuanvideo-community/HunyuanVideo",
    transformer=transformer,
    image_encoder=image_encoder,
    torch_dtype=torch.float16,
)

list(map(
    lambda x: apply_group_offloading(x, onload_device, offload_device, offload_type="leaf_level", use_stream=True, low_cpu_mem_usage=True),
    [pipe.text_encoder, pipe.text_encoder_2, pipe.transformer]
))
pipe.image_encoder.to(onload_device)
pipe.vae.to(onload_device)
pipe.vae.enable_tiling()

image = load_image(
    "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/penguin.png"
)
output = pipe(
    image=image,
    prompt="A penguin dancing in the snow",
    height=832,
    width=480,
    num_frames=91,
    num_inference_steps=30,
    guidance_scale=9.0,
    generator=torch.Generator().manual_seed(0),
    sampling_type="vanilla",
).frames[0]
print(f"Max memory: {torch.cuda.max_memory_allocated() / 1024**3:.3f} GB")
export_to_video(output, "output.mp4", fps=30)
```

https://github.com/user-attachments/assets/b15fe88e-439e-453b-91c1-525738a9d8bf

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -457,6 +457,8 @@
       title: Flux
     - local: api/pipelines/control_flux_inpaint
       title: FluxControlInpaint
+    - local: api/pipelines/framepack
+      title: Framepack
     - local: api/pipelines/hidream
       title: HiDream-I1
     - local: api/pipelines/hunyuandit
--- a/docs/source/en/api/pipelines/framepack.md
+++ b/docs/source/en/api/pipelines/framepack.md
@@ -1,0 +1,209 @@
+<!-- Copyright 2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License. -->
+
+# Framepack
+
+<div class="flex flex-wrap space-x-1">
+  <img alt="LoRA" src="https://img.shields.io/badge/LoRA-d8b4fe?style=flat"/>
+</div>
+
+[Packing Input Frame Context in Next-Frame Prediction Models for Video Generation](https://arxiv.org/abs/2504.12626) by Lvmin Zhang and Maneesh Agrawala.
+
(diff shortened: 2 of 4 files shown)


nitinmukesh:
@a-r-r-o-w 

As mentioned in code
apply_group_offloading

Does this still need a lot of RAM? I remember last in one of the topics there was discussion to reduce RAM requirements. Sorry but a little behind.

Is it possible to separate components and then do inference, will this require less VRAM.

text_encoder > prompt_embeds (with apply_group_offloading)

transformer > latents  (with apply_group_offloading)

vae > video output  (without apply_group_offloading)

Just some random thought it may be stupid though.

nitinmukesh:
@a-r-r-o-w 

Thank you. I am gonna try this and see the feasibility to use this feature on low VRAM. I hope this works on quantized model as well.

yiyixuxu (approved):
thanks! @a-r-r-o-w

a-r-r-o-w:
@tin2tin The first model that was released supported inverted anti-drifting (which means, it generates the last frame first and then fills in the blanks between the first and last frame) sampling. For this model, the community found that providing the end frame as target latent worked well in using first-last-frame interpolation case. With the new model (F1), it is based on the vanilla forward technique. Here, I don't think there's a trivial way you can inject last frame control. So, until further investigation, end frame support does not exist with the F1 model but should work as expected for the first model

tin2tin:
Is it the scheduler which tells diffusers which way to handle the model(forwards or backwards)?

a-r-r-o-w:
It's the new `sampling_type` parameter. Schedulers are different in that they only handle the denoising aspect i.e. computing a less noisy sample from a more noisy sample given the model prediction.

PR #11556, opened 2025-05-15
[docs] Regional compilation docs

sayakpaul:
# What does this PR do?

https://github.com/huggingface/accelerate/pull/3529

### DIFF
--- a/docs/source/en/optimization/torch2.0.md
+++ b/docs/source/en/optimization/torch2.0.md
@@ -83,8 +83,8 @@ For more information and different options about `torch.compile`, refer to the [
 Compiling the whole model usually has a big problem space for optimization. Models are often composed of multiple repeated blocks. [Regional compilation](https://pytorch.org/tutorials/recipes/regional_compilation.html) compiles the repeated block first (a transformer encoder block, for example), so that the Torch compiler would re-use its cached/optimized generated code for the other blocks, reducing (often massively) the cold start compilation time observed on the first inference call.
 
 Enabling regional compilation might require simple yet intrusive changes to the
-modeling code. However, 🤗 Accelerate provides a utility [`compile_regions()`](https://huggingface.co/docs/accelerate/main/en/usage_guides/compilation#how-to-use-regional-compilation) which automatically _only_ compiles
-the repeated blocks of the provided `nn.Module`.
+modeling code. However, 🤗 Accelerate provides a utility [`compile_regions()`](https://huggingface.co/docs/accelerate/main/en/usage_guides/compilation#how-to-use-regional-compilation) which automatically compiles
+the repeated blocks of the provided `nn.Module` along with other parts of it that are non-repeating. This helps with not only just cold start time but also the inference latency.
 
 ```py
 # Make sure you're on the latest `accelerate`: `pip install -U accelerate`.
@@ -93,8 +93,7 @@ from accelerate.utils import compile_regions
 pipe.unet = compile_regions(pipe.unet, mode="reduce-overhead", fullgraph=True)
 ```
 
-As you may have noticed `compile_regions()` takes the same arguments as `torch.compile()`, allowing
-flexibility.
+As you may have noticed `compile_regions()` takes the same arguments as `torch.compile()`, allowing flexibility.
 
 ## Benchmark
 
--- a/docs/source/en/optimization/torch2.0.md
+++ b/docs/source/en/optimization/torch2.0.md
@@ -78,6 +78,24 @@ For more information and different options about `torch.compile`, refer to the [
 > [!TIP]
 > Learn more about other ways PyTorch 2.0 can help optimize your model in the [Accelerate inference of text-to-image diffusion models](../tutorials/fast_diffusion) tutorial.
 
+### Regional compilation
+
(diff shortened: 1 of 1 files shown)


docs/source/en/optimization/torch2.0.md
```diff
@@ -78,6 +78,24 @@ For more information and different options about `torch.compile`, refer to the [
...
+### Regional compilation
+
+Compiling the whole model usually has a big problem space for optimization. Models are often composed of multiple repeated blocks. [Regional compilation](https://pytorch.org/tutorials/recipes/regional_compilation.html) compiles the repeated block first (a transformer encoder block, for example), so that the Torch compiler would re-use its cached/optimized generated code for the other blocks, reducing (often massively) the cold start compilation time observed on the first inference call.
+
+Enabling regional compilation might require simple yet intrusive changes to the
+modeling code. However, 🤗 Accelerate provides a utility [`compile_regions()`](https://huggingface.co/docs/accelerate/main/en/usage_guides/compilation#how-to-use-regional-compilation) which automatically _only_ compiles
+the repeated blocks of the provided `nn.Module`.
+
+```py
+# Make sure you're on the latest `accelerate`: `pip install -U accelerate`.
```
sayakpaul, line 90: Merge after `accelerate` new version is released this week.

docs/source/en/optimization/torch2.0.md
```diff
@@ -78,6 +78,24 @@ For more information and different options about `torch.compile`, refer to the [
 > [!TIP]
 > Learn more about other ways PyTorch 2.0 can help optimize your model in the [Accelerate inference of text-to-image diffusion models](../tutorials/fast_diffusion) tutorial.
 
+### Regional compilation
+
+Compiling the whole model usually has a big problem space for optimization. Models are often composed of multiple repeated blocks. [Regional compilation](https://pytorch.org/tutorials/recipes/regional_compilation.html) compiles the repeated block first (a transformer encoder block, for example), so that the Torch compiler would re-use its cached/optimized generated code for the other blocks, reducing (often massively) the cold start compilation time observed on the first inference call.
+
+Enabling regional compilation might require simple yet intrusive changes to the
+modeling code. However, 🤗 Accelerate provides a utility [`compile_regions()`](https://huggingface.co/docs/accelerate/main/en/usage_guides/compilation#how-to-use-regional-compilation) which automatically _only_ compiles
+the repeated blocks of the provided `nn.Module`.
```
IlyasMoutawwakil, line 87: no we actually compile the rest of the model as well 😅 I found out in my post that some people thought only the encoder/decoder block will be compiled in regional, which is not true.
I changed the docs to be more explicit https://github.com/huggingface/accelerate/pull/3572#discussion_r2089449059

sayakpaul: 👁️ But https://docs.pytorch.org/tutorials/recipes/regional_compilation.html suggests a completely different recipe no? No full compilation but only regional and I always thought that is what should be done. 

What am I missing?

IlyasMoutawwakil: regional compilation is simply: cut into regions and then compile them. I didn't compare the two approaches but I believe in the context of the pytorch tutorial they were simply trying to reduce cold start, not caring about how to keep inference optimized as well (they didn't benchamrk inference).

sayakpaul: So 

1. inference latency of compiling full model >= inference latency of regionally compiling repeated blocks + compiling additional blocks in a model
2. cold start time of compiling full model >> cold start time of regionally compiling repeated blocks + compiling additional blocks in a model

Is my understanding right or is it still fragmented?

IlyasMoutawwakil: yes that is how it works !

> Do you think providing an option to NOT compile the rest of the blocks could still make sense?

doesn't sense for me personally, since you will miss on the tuning of the task-specific head. Do you have any specific cases where we don't want to compile the rest of the model ?

docs/source/en/optimization/torch2.0.md
```diff
@@ -78,6 +78,23 @@ For more information and different options about `torch.compile`, refer to the [
 > [!TIP]
 > Learn more about other ways PyTorch 2.0 can help optimize your model in the [Accelerate inference of text-to-image diffusion models](../tutorials/fast_diffusion) tutorial.
 
+### Regional compilation
+
+Compiling the whole model usually has a big problem space for optimization. Models are often composed of multiple repeated blocks. [Regional compilation](https://pytorch.org/tutorials/recipes/regional_compilation.html) compiles the repeated block first (a transformer encoder block, for example), so that the Torch compiler would re-use its cached/optimized generated code for the other blocks, reducing (often massively) the cold start compilation time observed on the first inference call.
+
+Enabling regional compilation might require simple yet intrusive changes to the
+modeling code. However, 🤗 Accelerate provides a utility [`compile_regions()`](https://huggingface.co/docs/accelerate/main/en/usage_guides/compilation#how-to-use-regional-compilation) which automatically compiles
+the repeated blocks of the provided `nn.Module` along with other parts of it that are non-repeating. This helps with not only just cold start time but also the inference latency.
```
IlyasMoutawwakil, line 87: @sayakpaul what do you think ?

```suggestion
the repeated blocks of the provided `nn.Module` sequentially, and the rest of the model separately. This helps with reducing cold start time (by hitting the cache) while keeping most (if not all) of the speedup you would get from full compilation.
```

PR #11558, opened 2025-05-15
[tests] add tests for combining layerwise upcasting and groupoffloading.

sayakpaul:


tests/models/test_modeling_common.py
```diff
@@ -1580,6 +1580,45 @@ def run_forward(model):
...
+        _ = model(**inputs_dict)[0]
+
+        torch.manual_seed(0)
+        init_dict, inputs_dict = self.prepare_init_args_and_inputs_for_common()
+        inputs_dict = cast_maybe_tensor_dtype(inputs_dict, torch.float32, compute_dtype)
+        model = self.model_class(**init_dict)
+        model.eval()
+        model.enable_group_offload(
+            torch_device,
+            offload_type="block_level",
```
a-r-r-o-w, line 1614: I think it would make more sense to parameterize test on offload_type instead of record_stream

tests/models/test_modeling_common.py
```diff
@@ -1580,6 +1580,45 @@ def run_forward(model):
...
+        model.enable_group_offload(torch_device, offload_type="block_level", num_blocks_per_group=1)
+        model.enable_layerwise_casting(storage_dtype=storage_dtype, compute_dtype=compute_dtype)
+        _ = model(**inputs_dict)[0]
+
+        torch.manual_seed(0)
+        init_dict, inputs_dict = self.prepare_init_args_and_inputs_for_common()
+        inputs_dict = cast_maybe_tensor_dtype(inputs_dict, torch.float32, compute_dtype)
+        model = self.model_class(**init_dict)
+        model.eval()
+        model.enable_group_offload(
```
a-r-r-o-w, line 1612: The test is named saying both layerwise casting and group offloading are tested, which is done in the first part of this test, but this second part only checks group offloading. Should it be removed?

a-r-r-o-w (approved):
Thanks, LGTM

tests/models/test_modeling_common.py
sayakpaul: My bad. The second part can be removed.

tests/models/test_modeling_common.py
sayakpaul: Agree. Will update.

PR #11564, opened 2025-05-16
add prodigy deps

crapthings:
add missing optimizer deps for beginner


yiyixuxu:
already addressed so closing :)

PR #11565, opened 2025-05-16
[benchmarks] overhaul benchmarks

sayakpaul:
# What does this PR do?

This PR considerably simplifies how we do benchmarks. Instead of using entire pipeline-level benchmarks across different tasks, we will now ONLY benchmark the diffusion network that is the most compute-intensive part in a standard diffusion workflow.

To make the estimates more realistic, we will make use of pre-trained checkpoints and dummy inputs with reasonable dimensionalities. 

I ran `benchmarking_flux.py` on an 80GB A100 on a batch size of 1 and got the following results:

image

Analyze the results in this Space: https://huggingface.co/spaces/diffusers/benchmark-analyzer

By default, all benchmarks will use a batch size of 1, eliminating CFG.

## How to add your benchmark?

Adding benchmarks for a new model class (`SanaTransformer2DModel`, for example) boils down to the following:

1. Define the dummy inputs of the model.
2. Define the benchmarking scenarios we should run the benchmark on.

This is what `benchmarking_flux.py` does. More modularization can be shipped afterward.

Idea would be to merge this PR with pre-configured benchmarks for a few popular models and open others to the community.

## TODOs

Utilities:

- [x] To fire the execution of the individual model-level benchmarks sequentially.
- [x] To combine CSVs from multiple different model classes.
- [x] Central dataset update and Slack notification.

@DN6 could you give the approach a quick look? I can then work on resolving the TODOs.


benchmarks/benchmarking_utils.py
```diff
@@ -0,0 +1,157 @@
...
+
+from diffusers.models.modeling_utils import ModelMixin
+from diffusers.utils import logging
+from diffusers.utils.testing_utils import require_torch_gpu, torch_device
+
+
+logger = logging.get_logger(__name__)
+
+
+def benchmark_fn(f, *args, **kwargs):
```
sayakpaul, line 18: This automatically warms up the model. No need to do it explicitly.

benchmarks/benchmarking_flux.py
```diff
@@ -0,0 +1,97 @@
...
+        "img_ids": image_ids,
+        "txt_ids": text_ids,
+        "pooled_projections": pooled_prompt_embeds,
+        "timestep": timestep,
+        "guidance": guidance,
+    }
+
+
+if __name__ == "__main__":
+    scenarios = [
```
sayakpaul, line 36: Covered the following scenarios:

* Regular BF16 with compilation
* NF4
* Layerwise upcasting
* Group offloading

sayakpaul:
Added SDXL, Wan (14B), and LTX (13B):

<details>
<summary>Results</summary>

|    |   index | scenario                                                     | model_cls                  |   num_params_M |     flops_M |   time_plain_s |   mem_plain_GB |   time_compile_s |   mem_compile_GB |   fullgraph | mode    |
|---:|--------:|:-------------------------------------------------------------|:---------------------------|---------------:|------------:|---------------:|---------------:|-----------------:|-----------------:|------------:|:--------|
|  0 |       0 | Wan-AI/Wan2.1-T2V-14B-Diffusers-bf16                         | WanTransformer3DModel      |       14288.5  | 7.85612e+08 |         10.666 |          31.17 |            9.125 |            31.77 |           1 | default |
|  1 |       1 | Wan-AI/Wan2.1-T2V-14B-Diffusers-layerwise-upcasting          | WanTransformer3DModel      |       14288.5  | 7.85612e+08 |         10.67  |          26.78 |          nan     |           nan    |         nan | nan     |
|  2 |       2 | Wan-AI/Wan2.1-T2V-14B-Diffusers-group-offload-leaf           | WanTransformer3DModel      |       14288.5  | 7.85612e+08 |         10.786 |           4.48 |          nan     |           nan    |         nan | nan     |
|  3 |       0 | stabilityai/stable-diffusion-xl-base-1.0-bf16                | UNet2DConditionModel       |        2567.46 | 5.9791e+06  |          0.088 |           5.05 |            0.058 |             5.39 |           1 | default |
|  4 |       1 | stabilityai/stable-diffusion-xl-base-1.0-layerwise-upcasting | UNet2DConditionModel       |        2567.46 | 5.9791e+06  |          0.17  |           4.89 |          nan     |           nan    |         nan | nan     |
|  5 |       2 | stabilityai/stable-diffusion-xl-base-1.0-group-offload-leaf  | UNet2DConditionModel       |        2567.46 | 5.9791e+06  |          0.386 |           0.2  |          nan     |           nan    |         nan | nan     |
|  6 |       0 | Lightricks/LTX-Video-0.9.7-dev-bf16                          | LTXVideoTransformer3DModel |       13042.6  | 1.67583e+08 |          1.439 |          25.21 |            1.134 |            25.63 |           1 | default |
|  7 |       1 | Lightricks/LTX-Video-0.9.7-dev-layerwise-upcasting           | LTXVideoTransformer3DModel |       13042.6  | 1.67583e+08 |          1.522 |          24.38 |          nan     |           nan    |         nan | nan     |
|  8 |       2 | Lightricks/LTX-Video-0.9.7-dev-group-offload-leaf            | LTXVideoTransformer3DModel |       13042.6  | 1.67583e+08 |          1.806 |           1.04 |          nan     |           nan    |         nan | nan     |

</details>

a-r-r-o-w:
> Cc: @a-r-r-o-w if you want to add some caching benchmarks (in a later PR), I think that would be really great!

Sounds good, I'll take it up in near future once this PR is in

.github/workflows/benchmark.yml
```diff
@@ -13,15 +13,15 @@ env:
...
     env:
       SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL_BENCHMARK }}
-    name: Torch Core Pipelines CUDA Benchmarking Tests
+    name: Torch Core Models CUDA Benchmarking Tests
     strategy:
       fail-fast: false
       max-parallel: 1
     runs-on:
-      group: aws-g6-4xlarge-plus
+      group: aws-g6e-xlarge-plus
```
sayakpaul, line 24: Beefier machine.

sayakpaul:
@DN6 a gentle ping.

sayakpaul:
@McPatate I have updated this PR to add a DB population script and also adjusted the workflow that will trigger it. 

I decided to create a new table because the benchmarking workflow differs from the one used in `transformers` and as such, it has served us well. So, we would like to stick to this format.

LMK what you think.

McPatate:
> I decided to create a new table because the benchmarking workflow differs from the one used in transformers

Looking at your sql queries, everything you need is already present in the current db schema. I'd reuse it rather than create a new table. I understand the structure is a bit different but it shouldn't be too hard to make it fit:
- create a benchmark (with `INSERT INTO ... RETURNING benchmark_id`)
- create two sets of rows, one for device measurements and one for model measurements (I guess you don't even need device measurements as I don't see you measuring these values)
- for each measurement, create a dict with:
```
        scenario,
        model_cls,
        num_params_M,
        flops_M,
        time_plain_s,
        mem_plain_GB,
        time_compile_s,
        mem_compile_GB,
        fullgraph,
        mode,
```
- add the query to the batch:
```py
"""
INSERT INTO model_measurements (
  benchmark_id,
  measurements
) VALUES (%s, %s)
"""
```
- profit 😄 

In my benchmark script for transformers [I use argparse](https://github.com/huggingface/transformers/blob/8c59cdb3f81f9f4b8e7b8d05b92d40c2e9413788/benchmark/benchmarks_entrypoint.py#L88C1-L121C1) to get some args and [pass them from the workflow](https://github.com/huggingface/transformers/blob/8c59cdb3f81f9f4b8e7b8d05b92d40c2e9413788/.github/workflows/benchmark.yml#L67)

EDIT: don't forget to set your repository's name to differentiate from transformers, I didn't put `NOT NULL` constraints everywhere 😅)

sayakpaul:
> create two sets of rows, one for device measurements and one for model measurements (I guess you don't even need device measurements as I don't see you measuring these values)

Yeah the current numbers suffice for the purpose. 

Could you elaborate on the hesitation behind not creating a new table? This way, there's no chance to mess up with the transformers benchmarks and we maintain a clear separation.

For something that is working already well for observability purposes, I really don't want to introduce intrusive changes to the workflow unless there's a good reason to. You mentioned profit as well, whereas I fail to see it. So, please enlighten :D

McPatate:
> Could you elaborate on the hesitation behind not creating a new table? This way, there's no chance to mess up with the transformers benchmarks and we maintain a clear separation.

I don't think there needs to be a separation of tables, it'll create redundancy, especially so given you'll converge towards a similar data model (benchmark table containing metadata with relation 1->N measurements table).

It does create a dependency between the two repos due to said data model, but I don't really see this being an issue since the measurements table is already a JSONB field and the benchmarks table shouldn't really change that often, if at all.

I understand the concerns about messing the data of the other repo and that separating things might be useful in a protective manner. Personally I definitely think that having a shared data model helps regarding re-using dashboards and having a consistent structure of benchmarks between repositories makes it easier to reason about. If you really feel like it's necessary, I'd create a duplicate table, `diffusers_measurements` or something of the kind but still use the `benchmarks` table, so a hybrid approach.

Overall, I don't think there's much risk: database isn't updated that often, so data levels will always be low / small. Db is backed up so if something bad happens we can restore with minimal data loss. Data loss isn't a huge deal, you can always re-run benchmarks at a given sha. So this is where I'm coming from when saying re-use existing schema, it'll make for less code maintenance in diffusers (while also potentially making it possible to re-use dashboards in grafana, although I don't really see that happening yet).

> I really don't want to introduce intrusive changes to the workflow

Nothing intrusive about building upon existing work :)

sayakpaul:
Okay will roll with it.

.github/workflows/benchmark.yml
```diff
@@ -3,25 +3,26 @@ name: Benchmarking tests
 on:
   workflow_dispatch:
   schedule:
-    - cron: "30 1 1,15 * *" # every 2 weeks on the 1st and the 15th of every month at 1:30 AM
+    - cron: "0 0 * * *" # every day at midnight
```
DN6: Don't think we need to run every day at midnight no? Our optimizations don't change at a pace that requires running daily. If we want to increase frequency I would suggest running this weekly.

.github/workflows/benchmark.yml
```diff
@@ -37,18 +38,16 @@ jobs:
...
       - name: Diffusers Benchmarking
         env:
-            HF_TOKEN: ${{ secrets.DIFFUSERS_BOT_TOKEN }}
-            BASE_PATH: benchmark_outputs
+          HF_TOKEN: ${{ secrets.DIFFUSERS_BOT_TOKEN }}
         run: |
-          export TOTAL_GPU_MEMORY=$(python -c "import torch; print(torch.cuda.get_device_properties(0).total_memory / (1024**3))")
-          cd benchmarks && mkdir ${BASE_PATH} && python run_all.py && python push_results.py
+          cd benchmarks && python run_all.py && python push_results.py
+          mkdir ${BASE_PATH} && mv *.csv ${BASE_PATH}
```
DN6: Change to use env variable 
```suggestion
          mkdir $BASE_PATH && mv *.csv $BASE_PATH
```

.github/workflows/benchmark.yml
```diff
@@ -3,25 +3,26 @@ name: Benchmarking tests
...
     env:
       SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL_BENCHMARK }}
-    name: Torch Core Pipelines CUDA Benchmarking Tests
+    name: Torch Core Models CUDA Benchmarking Tests
     strategy:
       fail-fast: false
       max-parallel: 1
     runs-on:
-      group: aws-g6-4xlarge-plus
+      group: aws-g6e-xlarge-plus
```
DN6: Can be done in a follow up PR but it would be good to run a benchmark for more resource constrained environments too with a small GPU.

sayakpaul: Indeed. Some tests might need a bigger GPU and I think we can catch and report them appropriately. Do you want to take a stab at this in a follow-up?

benchmarks/populate_into_db.py
```diff
@@ -0,0 +1,114 @@
+import datetime
```
sayakpaul: @McPatate see if this works for you.

benchmarks/populate_into_db.py
```diff
@@ -0,0 +1,114 @@
...
+    try:
+        conn = psycopg2.connect(
+            host=os.getenv("PGHOST"),
+            database=os.getenv("PGDATABASE"),
+            user=os.getenv("PGUSER"),
+            password=os.getenv("PGPASSWORD"),
+        )
+        print("DB connection established successfully.")
+    except Exception:
+        raise
```
McPatate: Won't this obfuscate the error message?

benchmarks/populate_into_db.py
```diff
@@ -0,0 +1,114 @@
...
+            # If "github_sha" column exists in the CSV, cast it; else default to None
+            if "github_sha" in df.columns:
+                github_sha = _cast_value(row.get("github_sha"), "text")
+            else:
+                github_sha = None
+
+            if github_sha:
+                benchmark_id = f"{model_cls}-{scenario}-{github_sha}"
+            else:
+                benchmark_id = f"{model_cls}-{scenario}-{id_for_benchmark}"
```
McPatate: `benchmark_id` is the primary key of the row in the `benchmarks` table. You need to insert a new entry in this table first and fetch its id, like so:

https://github.com/huggingface/transformers/blob/afdb821318e06e670c7238a9059e7e031065e319/benchmark/benchmarks_entrypoint.py#L31

benchmarks/populate_into_db.py
```diff
@@ -0,0 +1,114 @@
...
+import uuid
+
+import pandas as pd
+import psycopg2
+import psycopg2.extras
+
+
+FINAL_CSV_FILENAME = "collated_results.csv"
+# https://github.com/huggingface/transformers/blob/593e29c5e2a9b17baec010e8dc7c1431fed6e841/benchmark/init_db.sql#L27
+TABLE_NAME = "model_measurements"
```
McPatate:
```suggestion
BENCHMARKS_TABLE_NAME = "benchmarks"
MEASUREMENTS_TABLE_NAME = "model_measurements"
```

benchmarks/populate_into_db.py
```diff
@@ -0,0 +1,114 @@
...
+                "mem_compile_GB": mem_compile_GB,
+                "fullgraph": fullgraph,
+                "mode": mode,
+                "github_sha": github_sha,
+            }
+            rows_to_insert.append((benchmark_id, measurements))
+
+        # Batch-insert all rows
+        insert_sql = f"""
+        INSERT INTO {TABLE_NAME} (
```
McPatate:
```suggestion
        INSERT INTO {MEASUREMENTS_TABLE_NAME} (
```

benchmarks/README.md
```diff
@@ -0,0 +1,69 @@
...
+Welcome to Diffusers Benchmarks. These benchmarks are use to obtain latency and memory information of the most popular models across different scenarios such as:
+
+* Base case i.e., when using `torch.bfloat16` and `torch.nn.functional.scaled_dot_product_attention`.
+* Base + `torch.compile()`
+* NF4 quantization
+* Layerwise upcasting
+
+Instead of full diffusion pipelines, only the forward pass of the respective model classes (such as `FluxTransformer2DModel`) is tested with the real checkpoints (such as `"black-forest-labs/FLUX.1-dev"`). 
+
+The entrypoint to running all the currently available benchmarks is in `run_all.py`. However, one can run the individual benchmarks, too, i.e., `python benchmarking_flux.py`. It should produce a CSV file containing various information about the benchmarks run.
```
sayakpaul:
```suggestion
The entrypoint to running all the currently available benchmarks is in `run_all.py`. However, one can run the individual benchmarks, too, e.g., `python benchmarking_flux.py`. It should produce a CSV file containing various information about the benchmarks run.
```

sayakpaul:
@DN6 a gentle ping.

DN6 (approved):

.github/workflows/benchmark.yml
```diff
@@ -3,25 +3,26 @@ name: Benchmarking tests
 on:
   workflow_dispatch:
   schedule:
-    - cron: "30 1 1,15 * *" # every 2 weeks on the 1st and the 15th of every month at 1:30 AM
+    - cron: "0 17 * * 1" # every monday at 5 PM.
```
DN6: Not a blocker. But why run every week? Is a monthly benchmark not sufficient?

sayakpaul: True. Changing to bi-weekly.

sayakpaul:
@anijain2305 just a ping to let you know that we're merging this PR which will run the benchmarking suite bi-weekly and report the results here: https://huggingface.co/datasets/diffusers/benchmarks/blob/main/collated_results.csv

anijain2305:
Thanks for setting this up. This will be really helpful for tracking progress and identifying regression

benchmarks/populate_into_db.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+FINAL_CSV_FILENAME = "collated_results.csv"
+# https://github.com/huggingface/transformers/blob/593e29c5e2a9b17baec010e8dc7c1431fed6e841/benchmark/init_db.sql#L27
+BENCHMARKS_TABLE_NAME = "benchmarks"
+MEASUREMENTS_TABLE_NAME = "model_measurements"
+
+
+def _init_benchmark(conn, branch, commit_id, commit_msg):
+    metadata = {"metadata": f"diffusers-{commit_id}"}
```
sayakpaul: @McPatate getting: https://github.com/huggingface/diffusers/actions/runs/16043615394/job/45269816792#step:10:22

[From](https://github.com/huggingface/transformers/blob/8178c43112295bf8c4ef04c667efbbbfd34b8bca/benchmark/llama.py#L48) what I understand, I _think_ I am doing things as expected here. Could you comment?

McPatate: You need to do the following:
```py
from psycopg2.extensions import register_adapter
from psycopg2.extras import Json

register_adapter(dict, Json)
```

McPatate: What is the purpose of the field? Doesn't bring much value imo, you can remove it. The metadata field is more to store extra info, in transformers we store gpu name and model id for instance.

sayakpaul: Indeed. Can we keep it null?

sayakpaul:
Since everything is passing now, will merge this PR :) 
https://github.com/huggingface/diffusers/actions/runs/16065987231/job/45340516693

sayakpaul:
Also cc @a-r-r-o-w for https://github.com/huggingface/diffusers/pull/11565#issuecomment-2893942224 (not urgent, when you get time).

benchmarks/populate_into_db.py
McPatate: The field isn't nullable 😅 
Guess you can put an empty dict though.

PR #11587, opened 2025-05-20
Fix bug when `variant` and `safetensor` file does not match

kaixuanliu:


kaixuanliu:
This PR fixes bug under `is_safetensors_compatible` function: when `variant` and `safetensor` file does not match, this function should return False. For example, in this model: [stable-diffusion-v1-5/stable-diffusion-inpainting](https://huggingface.co/stable-diffusion-v1-5/stable-diffusion-inpainting/tree/main/unet), there is file `diffusion_pytorch_model.fp16.safetensors`, but when `variant` is None, it should return `False` instead of `True`

DN6:
Thanks @kaixuanliu! I think we would need to add the variant check when there are no component folders as well. 

```python
    if not components:
        return any(".safetensors" in filename for filename in filenames)
``` 
And I think the changes will break the current `is_safetensors_compatible` tests here
https://github.com/huggingface/diffusers/blob/05c8b42b750ed4a7c0e66df6254e3649be80afa2/tests/pipelines/test_pipeline_utils.py#L25

kaixuanliu:
@DN6 , Hi, thanks for your advice. Have updated the code and test cases

kaixuanliu:
@bot /style

DN6:
@bot /style

kaixuanliu:
@DN6 Hi, I have checked all the CI tests offline. Can you help trigger the CI workflow again?

PR #11591, opened 2025-05-20
Torchao int4 serialization

SunMarc:
# What does this PR do?

This PR fixes torchao int4 checkpoint loading. We need to load the checkpoint directly on the device we quantized it. We make the assumption that we are loading the model on the right device at the start. 

Needed for this model https://huggingface.co/diffusers/FLUX.1-dev-torchao-int4

### DIFF
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -1194,7 +1194,9 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
             ):
                 map_location = torch.device([d for d in device_map.values() if d not in ["cpu", "disk"]][0])
             # Time to load the checkpoint
-            state_dict = load_state_dict(resolved_model_file[0], disable_mmap=disable_mmap, dduf_entries=dduf_entries, map_location=map_location)
+            state_dict = load_state_dict(
+                resolved_model_file[0], disable_mmap=disable_mmap, dduf_entries=dduf_entries, map_location=map_location
+            )
             # We only fix it for non sharded checkpoints as we don't need it yet for sharded one.
             model._fix_state_dict_keys_on_load(state_dict)
 


src/diffusers/models/modeling_utils.py
```diff
@@ -1185,8 +1185,18 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
         state_dict = None
         if not is_sharded:
+            map_location = "cpu"
+            if (
+                device_map is not None
+                and hf_quantizer is not None
+                and hf_quantizer.quantization_config.quant_method == QuantizationMethod.TORCHAO
+                and hf_quantizer.quantization_config.quant_type in ["int4_weight_only", "autoquant"]
+            ):
+                map_location = torch.device([d for d in device_map.values() if d not in ["cpu", "disk"]][0])
```
sayakpaul, line 1195: Sufficiently safe to say that we would always have a non-None `device_map`?

Also, what happens if the `device_map` has multiple CUDA devices specified? Would the indexing make sense there? 

Okay for this PR but we could potentially have a `resolve_map_location()` per quantizer class, maybe.

sayakpaul (commented):
Thanks! Some questions.

sayakpaul:
Once the PR is close to merging, let's also add a test.

src/diffusers/models/modeling_utils.py
SunMarc: > Sufficiently safe to say that we would always have a non-None device_map?

I check that the device_map is not None. Also this should be safe enough. I took that from transformers. There shouldn't be an issue with the indexing, in any case we will move again the tensors if they are multiple index. 

Yeah I can switch to `update_map_location`.

SunMarc:
Will add a test ! 

Note that in general, I wouldn't recommend saving int4 models with torchao as this is hardware dependent between `cpu` and `cuda`.

sayakpaul:
> Note that in general, I wouldn't recommend saving int4 models with torchao as this is hardware dependent between cpu and cuda.

Indeed. Then let's also add a note in the docs

a-r-r-o-w:
@SunMarc @sayakpaul @DN6 Gentle ping in case this PR is still relevant

src/diffusers/models/modeling_utils.py
sayakpaul: Thanks! Let's move to `update_map_location()`.

Also, let's remove "autoquant" as we don't support it.

sayakpaul:
Any reason why it's closed?

SunMarc:
many things changed in torchao especially regarding serialization, so it is better to do a new pr when safetensors serialization is supported which should be the case in their next release !

PR #11603, opened 2025-05-22
save_pretrained for serializing config.

sayakpaul:
# What does this PR do?

Follow-up of https://github.com/huggingface/diffusers/pull/11539. 

Test code:

```py
from diffusers.modular_pipelines import PipelineBlock

block = PipelineBlock.from_pretrained(
    "diffusers-internal-dev/modular-depth-block-sayak", trust_remote_code=True
)
bloc.save_pretrained("dummy")
```

Currently, it serializes the `config.json`. I think we agreed to manually save the block module that implements the custom block. It looks like so:

```json
{
  "_class_name": "DepthProcessorBlock",
  "_component_names": [
    "depth_processor"
  ],
  "_diffusers_version": "0.34.0.dev0",
  "auto_map": {
    "PipelineBlock": "block.DepthProcessorBlock"
  }
}
```

I know that we are not serializing out-of-the-ordinary info that is very non-trivial. So, I would be completely okay if we decided to serialize the `config.json` manually.

### DIFF
--- a/src/diffusers/modular_pipelines/modular_pipeline.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline.py
@@ -413,7 +413,7 @@ class ModularPipelineMixin(ConfigMixin):
         self._progress_bar_config = kwargs
 
 
-class PipelineBlock(ModularPipelineMixin, PushToHubMixin):
+class PipelineBlock(ModularPipelineMixin):
     
     model_name = None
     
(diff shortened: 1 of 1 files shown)


sayakpaul:
@DN6 I pushed an initial version of the `diffusers-cli` command we chatted about yesterday. Could you PTAL? 

Provided you have implemented a `block.py` in a directory called `custom_blocks`, run:

```bash
diffusers-cli custom_blocks
```

And then it should spit out:

```bash
config.json
requirements.txt
block.py (should already be present)
```

This is a rough proposal, happy to iterate further.

yiyixuxu:
thanks @sayakpaul, ok if I merge in first?  I think erializes the config.json is sufficient for now

sayakpaul:
@yiyixuxu let's go.

PR #11610, opened 2025-05-25
Make SDXL treat the "optional" components as really optional

ppbrown:
# What does this PR do?

This makes the "optional" text modules in StableDiffusionXLPipeline actually function as optional, instead of mandatory.

## Before submitting
- [x ] Did you read the [contributor guideline](https://github.com/huggingface/diffusers/blob/main/CONTRIBUTING.md)?
- [ x] Did you read our [philosophy doc](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md) (important for complex PRs)?

## Who can review?

Core library:

### DIFF
--- a/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
+++ b/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
@@ -16,6 +16,7 @@ import inspect
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import torch
+import torch.nn.functional as F  # required for padding shorter sequences
 from transformers import (
     CLIPImageProcessor,
     CLIPTextModel,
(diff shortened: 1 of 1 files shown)


asomoza:
@ppbrown thanks for your contribution.

We usually only add additional code to the official pipelines and the core  when there's a fully working and released model that it's also popular, to avoid adding more complexity and making it a huge library with unused models.

For what I can see in your [model repo](https://huggingface.co/opendiffusionai/sdxlone):

> So, I created this version to experimentally prove the better way.

and in the [other model](https://huggingface.co/opendiffusionai/sdxl-longcliponly#known-problems) there is this part:

> This raw version of the model seems to work great with 3-5 word prompts... but then decays after that. So I'm working on a finetuned version.

both texts suggest that this is just and experiment that needs more finetuning, also it's not that popular right now.

I suggest that we add this as a community pipeline for the moment so people can test your model and we can see then how this goes for the community, also it would make it a lot faster to merge.

asomoza:
thanks, pinging @DN6 or @sayakpaul. Was the intention with the original SDXL implementation to make the second text encoder optional and skippable?

sayakpaul:
Well, that was because you could first use `pipeline.encode_prompt()` to compute the prompt embeddings and pass those to the pipeline without having to have the text encoders loaded. This is common practice these days especially with models like HiDream, making use of four text encoders.

This is something we already test in:
https://github.com/huggingface/diffusers/blob/df55f0535829e2a1348062f85c7483fb51197102/tests/pipelines/test_pipelines_common.py#L2040

PR #11616, opened 2025-05-27
[training docs] smol update to README files 

linoytsaban:
address #11564 by adding notes to README files to install prodigy


sayakpaul (approved):
Thanks!

PR #9133, opened 2024-08-09
Update transformer_flux.py. Change float64 to float32 

jsmidt:
# What does this PR do?
dtype=torch.float64 is overkill, and float64 is not defined for certain devices such as Apple Silicon mps. This change enables the flux pipeline to be run on certain devices such as Apple Silicon mps without negative consequences.

PR #11643, opened 2025-06-02
Use float32 RoPE freqs in Wan with MPS backends

hvaara:
An issue was reported in PyTorch regarding dtype compat in the Wan model (see https://github.com/pytorch/pytorch/issues/148670) when using MPS as the backend. The issue is that Wan's RoPE implementation uses float64, which isn't a supported dtype on MPS and results in `TypeError: Trying to convert ComplexDouble to the MPS backend but it does not have support for that dtype.`

This PR proposes to downcast freqs on MPS backends. Looking at prior PRs (#10776, #9133, etc) this seems to be the preferred method, but please let me know if there's a different approach you'd like me to take and I'll amend as needed.

FYI; There's a separate issue with Wan when using MPS backends outlined in https://github.com/pytorch/pytorch/issues/148670#issuecomment-2928499989. My hunch is that this bug is a correctness issue in the MPS codebase in PyTorch, although that's not verified yet.

cc @a-r-r-o-w @asomoza @yiyixuxu

### DIFF
--- a/src/diffusers/models/transformers/transformer_wan.py
+++ b/src/diffusers/models/transformers/transformer_wan.py
@@ -72,7 +72,8 @@ class WanAttnProcessor2_0:
         if rotary_emb is not None:
 
             def apply_rotary_emb(hidden_states: torch.Tensor, freqs: torch.Tensor):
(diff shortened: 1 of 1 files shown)


a-r-r-o-w (approved):
Thanks, looks good!

hvaara:
Thanks for the quick review! 😄

PR #11645, opened 2025-06-02
Set _torch_version to N/A if torch is disabled.

rasmi:
# What does this PR do?

Set a default value for _torch_version if torch is disabled. This fixes an ImportError in hub_utils.py when it tries to import [_torch_version](https://github.com/huggingface/diffusers/blob/main/src/diffusers/utils/hub_utils.py#L61). This is not riased for Jax and Flax because they already set `_jax_version = "N/A"` as a default.

```
ImportError: cannot import name '_torch_version' from 'diffusers.utils.import_utils'
```

## Who can review?

Core library:


DN6 (approved):

rasmi:
Thanks @DN6 -- anything else needed from me to merge, or will you take it from here?

PR #11652, opened 2025-06-03
Allow remote code repo names to contain "."

akasharidas:
# What does this PR do?

When the repo path for a model containing custom code contains `.`, the code fails to import it correctly because `.` is interpreted by the Python import machinery as a module separator. This issue was reported in Transformers (https://github.com/huggingface/transformers/issues/28919) and was fixed in https://github.com/huggingface/transformers/pull/29175. The same issue exists in Diffusers, so this PR brings over that fix with minor changes to account for how Diffusers names the cache directory.

## Before submitting
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/diffusers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/diffusers/tree/main/docs#writing-source-documentation).

## Who can review?
@yiyixuxu @sayakpaul @asomoza

### DIFF
--- a/src/diffusers/utils/dynamic_modules_utils.py
+++ b/src/diffusers/utils/dynamic_modules_utils.py
@@ -154,7 +154,7 @@ def check_imports(filename):
     return get_relative_imports(filename)
 
 
-def get_class_in_module(pretrained_model_name_or_path, class_name, module_path):
+def get_class_in_module(class_name, module_path, pretrained_model_name_or_path=None):
     """
     Import a module on the cache directory for modules and extract a class from it.
     """
@@ -165,7 +165,8 @@ def get_class_in_module(pretrained_model_name_or_path, class_name, module_path):
         # This can happen when the repo id contains ".", which Python's import machinery interprets as a directory
         # separator. We do a bit of monkey patching to detect and fix this case.
         if not (
-            "." in pretrained_model_name_or_path
+            pretrained_model_name_or_path is not None
+            and "." in pretrained_model_name_or_path
             and module_path.startswith("diffusers_modules")
             and pretrained_model_name_or_path.replace("/", "--") in module_path
         ):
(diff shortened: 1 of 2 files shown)


src/diffusers/utils/dynamic_modules_utils.py
```diff
@@ -154,12 +154,29 @@ def check_imports(filename):
     return get_relative_imports(filename)
 
 
-def get_class_in_module(class_name, module_path):
+def get_class_in_module(pretrained_model_name_or_path, class_name, module_path):
```
sayakpaul: Could we do:

```py
def get_class_in_module(class_name, module_path, pretrained_model_name_or_path=None):
```

This will ensure the change is not backward breaking.

akasharidas:
> Looks pretty good! Should we add a test for it here?
> 
> https://github.com/huggingface/diffusers/blob/c934720629837257b15fd84d27e8eddaa52b76e6/tests/pipelines/test_pipelines.py#L1009

To test this we need a repo that both has a dot in the name and has custom remote code. Currently I'm testing it on a private repo. @sayakpaul could you help by creating a new repo on `hf-internal-testing` for this? You could just duplicate `hf-internal-testing/tiny-sdxl-custom-all` into something like `hf-internal-testing/tiny-sdxl-custom-all-1.0` and then I can write a test similar to `test_remote_auto_custom_pipe`.

src/diffusers/utils/dynamic_modules_utils.py
akasharidas: Thanks for the feedback, I've made the proposed change

akasharidas:
Added a test for it in `tests/pipelines/test_pipelines.py`. Currently it depends on [this test repo](https://huggingface.co/akasharidas/ddpm-cifar10-32-dot.in.name) in my account. Feel free to move it into `hf-internal-testing`.

sayakpaul:
@DN6 could you give this a look?

PR #11654, opened 2025-06-04
enable torchao test cases on XPU and switch to device agnostic APIs for test cases

yao-matrix:
1. enable `torchao` cases on XPU, torchao has official XPU support from `0.11`. Tested in local, same pass rate as CUDA
2. switch to use device-agnostic APIs for all test cases.

@a-r-r-o-w @DN6 , pls help review, thx very much


tests/quantization/torchao/test_torchao.py
```diff
@@ -269,6 +271,7 @@ def test_int4wo_quant_bfloat16_conversion(self):
             subfolder="transformer",
             quantization_config=quantization_config,
             torch_dtype=torch.bfloat16,
+            device_map=f"{torch_device}:0",
```
yao-matrix, line 274: if load model to CPU, will meet "NotImplementedError: Could not run 'aten::_convert_weight_to_int4pack' with arguments from the 'CPU' backend." error, this both happens in XPU and CUDA, so directly load model to accelerator here.

tests/quantization/torchao/test_torchao.py
```diff
@@ -79,7 +81,7 @@ def test_post_init_check(self):
         Test kwargs validations in TorchAoConfig
         """
         _ = TorchAoConfig("int4_weight_only")
-        with self.assertRaisesRegex(ValueError, "is not supported yet"):
+        with self.assertRaisesRegex(ValueError, "is not supported"):
```
yao-matrix, line 84: xpu's error message doesn't have "yet", so just match "is not supported" for both CUDA and XPU

src/diffusers/utils/testing_utils.py
```diff
@@ -300,9 +300,7 @@ def require_torch_gpu(test_case):
 
 def require_torch_cuda_compatibility(expected_compute_capability):
     def decorator(test_case):
-        if not torch.cuda.is_available():
-            return unittest.skip(test_case)
-        else:
+        if torch.cuda.is_available():
```
yao-matrix, line 303: only check cuda device compatibility, if non-cuda device, just pass. For non-cuda device which needs compatibility, should check by themselves.

src/diffusers/quantizers/quantization_config.py
```diff
@@ -656,13 +656,13 @@ def generate_fpx_quantization_types(bits: int):
...
-        if major == 8:
-            return minor >= 9
-        return major >= 9
+        if torch.cuda.is_available():
+            major, minor = torch.cuda.get_device_capability()
+            if major == 8:
+                return minor >= 9
+            return major >= 9
+        else:
+            return True
```
yao-matrix, line 668: only check device capability when it's cuda; for non-cuda device, should check in separate utilities. In this case, non-cuda device(like XPU)'s case will be skipped by original implementation.

a-r-r-o-w: We should probably still raise an error if torchao is being used with mps or other devices, otherwise it leads to an obscure error somewhere deep in the code that common users will not understand

tests/quantization/torchao/test_torchao.py
a-r-r-o-w: Makes sense. IIRC this test is failing in our cuda CI too since the error message does not "yet" any more, so this should be okay

tests/quantization/torchao/test_torchao.py
a-r-r-o-w: Makes sense, we recently got this error on our CI too

a-r-r-o-w (approved):
Thanks, changes look good! Would just make sure that we error out early for devices that are not cuda/xpu since otherwise the errors are much more complicated for users to understand

src/diffusers/quantizers/quantization_config.py
yao-matrix: @a-r-r-o-w , i enhanced this utility per your comments, pls help review again, thx.

yao-matrix:
@DN6 , could you help review? Thx very much.

PR #11663, opened 2025-06-05
use deterministic to get stable result

jiqing-feng:
The test cannot always pass due to result unstable, so I use torch.use_deterministic_algorithms(True) just like [here](https://github.com/huggingface/diffusers/blob/c934720629837257b15fd84d27e8eddaa52b76e6/tests/models/test_modeling_common.py#L529) to get stable result so the test can always pass.

### DIFF
--- a/tests/quantization/bnb/test_4bit.py
+++ b/tests/quantization/bnb/test_4bit.py
@@ -476,6 +476,7 @@ class SlowBnb4BitTests(Base4bitTests):
         r"""
         Test that loading the model and unquantize it produce correct results.
         """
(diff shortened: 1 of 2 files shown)


jiqing-feng:
Hi @sayakpaul , could you please review this PR?

sayakpaul (commented):
Should we not use the same in `test_mixed_int8.py` as well?

jiqing-feng:
> Should we not use the same in `test_mixed_int8.py` as well?

Right, thanks for your reminder.

sayakpaul (approved):

PR #11669, opened 2025-06-06
Fix wrong param types, docs, and handles noise=None in scale_noise of FlowMatching schedulers

Promisery:
# What does this PR do?

In the `scale_noise` functions of all flow matching schedulers, the definitions and docs are wrong:

* `timestep` should be a `torch.FloatTensor`, not a float or int
* `noise` is not included in the document

Also, `noise=None` is not handled in the functions

## Before submitting

## Who can review?

@yiyixuxu

### DIFF
--- a/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
+++ b/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
@@ -171,7 +171,7 @@ class FlowMatchEulerDiscreteScheduler(SchedulerMixin, ConfigMixin):
     def scale_noise(
         self,
         sample: torch.FloatTensor,
-        timestep: Union[float, torch.FloatTensor],
+        timestep: torch.FloatTensor,
         noise: Optional[torch.FloatTensor] = None,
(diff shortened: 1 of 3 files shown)


src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
```diff
@@ -212,6 +214,9 @@ def scale_noise(
         while len(sigma.shape) < len(sample.shape):
             sigma = sigma.unsqueeze(-1)
 
+        if noise is None:
```
yiyixuxu: can we `noise` a required arg instead? 
to properly generate noise inside scale_noise we need to pass a generator too, I don't think we need to do that

yiyixuxu (commented):
thanks for the PR
I left one feedback

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
Promisery: Agreed

PR #11605, opened 2025-05-23
Make group offloading compatible with torch.compile()

sayakpaul:
# What does this PR do?

On H100 with Wan 14B, we get:

```bash
Compile: False offloading: False
Latency: 4601.803 ms (median over 1 runs)

Compile: True offloading: False
Latency: 3766.335 ms (median over 1 runs)

Compile: False offloading: True
Latency: 4918.042 ms (median over 1 runs)

Compile: True offloading: True
Latency: 4109.500 ms (median over 1 runs)
```

On RTX 4090, we get:

```bash
Compile: False offloading: True
Latency: 13658.121 ms (median over 1 runs)

Compile: True offloading: True
Latency: 11583.754 ms (median over 1 runs)
```

Code:

<details>
<summary>Expand</summary>

```py
from diffusers import AutoModel
import torch 
torch.set_grad_enabled(False)
from torch.utils import benchmark
import argparse

torch._dynamo.config.cache_size_limit = 10000

def get_input_dict(**device_dtype_kwargs):
    # height: 480
    # width: 832
    # num_frames: 81
    # max_sequence_length: 512
    hidden_states = torch.randn(1, 16, 21, 60, 104, **device_dtype_kwargs)
    encoder_hidden_states = torch.randn(1, 512, 4096, **device_dtype_kwargs)
    timestep = torch.tensor([1.0], **device_dtype_kwargs)

    return {"hidden_states": hidden_states, "encoder_hidden_states": encoder_hidden_states, "timestep": timestep}

def get_parser():
    parser = argparse.ArgumentParser()
    parser.add_argument("--compile", action="store_true")
    parser.add_argument("--go", action="store_true")
    return parser.parse_args()

if __name__ == "__main__":
    args = get_parser()
    transformer = AutoModel.from_pretrained(
        "Wan-AI/Wan2.1-T2V-14B-Diffusers", subfolder="transformer", torch_dtype=torch.bfloat16
    )
    if not args.go:
        transformer.cuda()
    else:
        group_offload_kwargs={
            "onload_device": "cuda",
            "offload_device": "cpu",
            "offload_type": "block_level",
            "num_blocks_per_group": 1,
            "use_stream": True,
            "non_blocking": True,
        }
        transformer.enable_group_offload(**group_offload_kwargs)
    if args.compile:
        transformer.compile()
    
    input_kwargs = {"dtype": torch.bfloat16, "device": "cuda"} if not args.go else {"dtype": torch.bfloat16}
[...]

PR #11670, opened 2025-06-06
[tests] add test for torch.compile + group offloading

sayakpaul:
# What does this PR do?

Only makes sense to have one after https://github.com/huggingface/diffusers/pull/11605. To run: 

`RUN_SLOW=1 RUN_COMPILE=1 pytest tests/models/ -k "test_compile_with_group_offloading"`.

Cc: @stevhliu do you think we should document this thing in our `torch.compile()` documentation section?

### DIFF
--- a/tests/models/test_modeling_common.py
+++ b/tests/models/test_modeling_common.py
@@ -1744,6 +1744,10 @@ class ModelPushToHubTester(unittest.TestCase):
         delete_repo(self.repo_id, token=TOKEN)
 
 
+@require_torch_gpu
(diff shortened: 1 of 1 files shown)


DN6:
Looks good. We wouldn't need this specific command since the current `-k "compile"` will pick up the test since compile is in the name. 

```
RUN_SLOW=1 RUN_COMPILE=1 pytest tests/models/ -k "test_compile_with_group_offloading"
```

DN6 (approved):

PR #11671, opened 2025-06-06
enable cpu offloading of new pipelines on XPU & use device agnostic empty to make pipelines work on XPU

yao-matrix:


src/diffusers/pipelines/kandinsky/pipeline_kandinsky_combined.py
```diff
@@ -193,7 +193,7 @@ def __init__(
     def enable_xformers_memory_efficient_attention(self, attention_op: Optional[Callable] = None):
         self.decoder_pipe.enable_xformers_memory_efficient_attention(attention_op)
 
-    def enable_sequential_cpu_offload(self, gpu_id: Optional[int] = None, device: Union[torch.device, str] = "cuda"):
+    def enable_sequential_cpu_offload(self, gpu_id: Optional[int] = None, device: Union[torch.device, str] = None):
```
yao-matrix, line 196: per discussion in this PR https://github.com/huggingface/diffusers/pull/11288, we change the default to `None`, so cpu_offloading can work on other accelerators like XPU w/ application code change.

yao-matrix: with such changes, cases like `tests/pipelines/wuerstchen/test_wuerstchen_combined.py::WuerstchenCombinedPipelineFastTests::test_cpu_offload_forward_pass_twice`, `tests/pipelines/kandinsky2_2/test_kandinsky_combined.py::KandinskyV22PipelineImg2ImgCombinedFastTests::test_cpu_offload_forward_pass_twice` can pass on XPU

yao-matrix:
@a-r-r-o-w @DN6 , pls help review, thx very much.

DN6:
@bot /style

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1313,10 +1311,9 @@ def enable_sequential_cpu_offload(self, gpu_id: Optional[int] = None, device: Un
         self._offload_device = device
 
         if self.device.type != "cpu":
+            orig_device_type = self.device.type
             self.to("cpu", silence_dtype_warnings=True)
-            device_mod = getattr(torch, self.device.type, None)
-            if hasattr(device_mod, "empty_cache") and device_mod.is_available():
-                device_mod.empty_cache()  # otherwise we don't see the memory savings (but they probably exist)
+            empty_device_cache(orig_device_type)
```
yao-matrix, line 1317: @DN6 , the original code has a bug, the `empty_cache` will not be executed.
The logic is like: check module's device type w/  `self.device.type`, at this time, it's "cuda" or "xpu", then it goes to the `if` scope, then it put module to cpu w/ `to`, after this, the `self.device.type` will be "cpu". So, "device_mod" will be `torch.cpu`, it has no `empty_cache`, so the following check will not pass, so `empty_cache` will not be called, so no `empty_cache` behavior.

I changed the code to make it will `empty_cache` of device, pls review in case that's not what in your design. 

PS. I attached the PR which changes to current code here: https://github.com/huggingface/diffusers/pull/4191

yao-matrix:
@DN6 , pls help review again, thx.

DN6 (approved):

PR #11672, opened 2025-06-06
[wip][tests] start adding tests for compilation + quantization

sayakpaul:
# What does this PR do?

WIP.

### DIFF
--- a/tests/quantization/bnb/test_4bit.py
+++ b/tests/quantization/bnb/test_4bit.py
@@ -28,9 +28,9 @@ from diffusers import (
     DiffusionPipeline,
     FluxControlPipeline,
     FluxTransformer2DModel,
-    PipelineQuantizationConfig,
     SD3Transformer2DModel,
 )
+from diffusers.quantizers import PipelineQuantizationConfig
 from diffusers.utils import is_accelerate_version, logging
 from diffusers.utils.testing_utils import (
     CaptureLogger,
@@ -46,7 +46,7 @@ from diffusers.utils.testing_utils import (
     require_torch,
     require_torch_accelerator,
     require_torch_gpu,
-    require_torch_version_greater_equal,
+    require_torch_version_greater,
     require_transformers_version_greater,
     slow,
     torch_device,
@@ -875,17 +875,26 @@ class Bnb4BitCompileTests(unittest.TestCase):
         backend_empty_cache(torch_device)
         torch.compiler.reset()
 
-    @require_torch_version_greater_equal("2.8")
+    @require_torch_version_greater("2.7.1")
     def test_torch_compile_4bit(self):
+        torch._dynamo.config.capture_dynamic_output_shape_ops = True
+
         quantization_config = PipelineQuantizationConfig(
             quant_backend="bitsandbytes_4bit",
-            quant_kwargs={"load_in_4bit": True},
+            quant_kwargs={
+                "load_in_4bit": True,
+                "bnb_4bit_quant_type": "nf4",
+                "bnb_4bit_compute_dtype": torch.bfloat16,
+            },
             components_to_quantize=["transformer"],
         )
         pipe = DiffusionPipeline.from_pretrained(
-            "hf-internal-testing/tiny-flux-pipe", quantization_config=quantization_config, torch_dtype=torch.bfloat16
+            "stabilityai/stable-diffusion-3-medium-diffusers",
+            quantization_config=quantization_config,
+            torch_dtype=torch.bfloat16,
         ).to("cuda")
         pipe.transformer.compile(fullgraph=True)
 
         for _ in range(2):
-            pipe("a dog", num_inference_steps=4, max_sequence_length=16)
+            # with torch._dynamo.config.patch(error_on_recompile=True):
+            pipe("a dog", num_inference_steps=4, max_sequence_length=16, height=256, width=256)
--- a/tests/quantization/bnb/test_4bit.py
+++ b/tests/quantization/bnb/test_4bit.py
@@ -28,6 +28,7 @@ from diffusers import (
     DiffusionPipeline,
     FluxControlPipeline,
     FluxTransformer2DModel,
+    PipelineQuantizationConfig,
     SD3Transformer2DModel,
 )
 from diffusers.utils import is_accelerate_version, logging
@@ -44,6 +45,8 @@ from diffusers.utils.testing_utils import (
     require_peft_backend,
     require_torch,
     require_torch_accelerator,
+    require_torch_gpu,
+    require_torch_version_greater_equal,
     require_transformers_version_greater,
     slow,
     torch_device,
@@ -855,3 +858,34 @@ class ExtendedSerializationTest(BaseBnb4BitSerializationTests):
 
     def test_fp4_double_safe(self):
         self.test_serialization(quant_type="fp4", double_quant=True, safe_serialization=True)
+
+
+@require_torch_gpu
+@slow
+class Bnb4BitCompileTests(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        gc.collect()
+        backend_empty_cache(torch_device)
+        torch.compiler.reset()
+
+    def tearDown(self):
+        super().tearDown()
+        gc.collect()
+        backend_empty_cache(torch_device)
+        torch.compiler.reset()
+
+    @require_torch_version_greater_equal("2.8")
+    def test_torch_compile_4bit(self):
+        quantization_config = PipelineQuantizationConfig(
+            quant_backend="bitsandbytes_4bit",
+            quant_kwargs={"load_in_4bit": True},
+            components_to_quantize=["transformer"],
+        )
+        pipe = DiffusionPipeline.from_pretrained(
+            "hf-internal-testing/tiny-flux-pipe", quantization_config=quantization_config, torch_dtype=torch.bfloat16
+        ).to("cuda")
+        pipe.transformer.compile(fullgraph=True)
+
+        for _ in range(2):
+            pipe("a dog", num_inference_steps=4, max_sequence_length=16)
--- a/src/diffusers/utils/testing_utils.py
+++ b/src/diffusers/utils/testing_utils.py
@@ -291,6 +291,18 @@ def require_torch_version_greater_equal(torch_version):
     return decorator
 
 
+def require_torch_version_greater(torch_version):
+    """Decorator marking a test that requires torch with a specific version greater."""
+
+    def decorator(test_case):
+        correct_torch_version = is_torch_available() and is_torch_version(">", torch_version)
+        return unittest.skipUnless(
+            correct_torch_version, f"test requires torch with the version greater than {torch_version}"
+        )(test_case)
+
+    return decorator
+
+
 def require_torch_gpu(test_case):
     """Decorator marking a test that requires CUDA and PyTorch."""
     return unittest.skipUnless(is_torch_available() and torch_device == "cuda", "test requires PyTorch+CUDA")(


tests/quantization/bnb/test_4bit.py
```diff
@@ -855,3 +858,43 @@ def test_fp4_double_unsafe(self):
...
+        )
+        pipe = DiffusionPipeline.from_pretrained(
+            "stabilityai/stable-diffusion-3-medium-diffusers",
+            quantization_config=quantization_config,
+            torch_dtype=torch.bfloat16,
+        ).to("cuda")
+        pipe.transformer.compile(fullgraph=True)
+
+        for _ in range(2):
+            # with torch._dynamo.config.patch(error_on_recompile=True):
```
sayakpaul: @matthewdouglas is this expected to have recompilations? If you uncomment this line and run the tests, it will fail with:

```bash
E               torch._dynamo.exc.RecompileError: Recompiling function forward in /fsx/sayak/diffusers/src/diffusers/models/transformers/transformer_sd3.py:317
E                   triggered by the following guard failure(s):
E                   - 0/0: self._modules['norm_out']._modules['linear'].compute_type_is_set == False

../miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/_dynamo/guards.py:3491: RecompileError
```

sayakpaul: Oh it turns out we can make the following changes for it to work:

```diff
diff --git a/tests/quantization/bnb/test_4bit.py b/tests/quantization/bnb/test_4bit.py
index 0567e63bf..1e880dcd9 100644
--- a/tests/quantization/bnb/test_4bit.py
+++ b/tests/quantization/bnb/test_4bit.py
@@ -896,5 +896,6 @@ class Bnb4BitCompileTests(unittest.TestCase):
         pipe.transformer.compile(fullgraph=True)
 
         for _ in range(2):
-            # with torch._dynamo.config.patch(error_on_recompile=True):
             pipe("a dog", num_inference_steps=4, max_sequence_length=16, height=256, width=256)
+            with torch._dynamo.config.patch(error_on_recompile=True), torch.compiler.set_stance(skip_guard_eval_unsafe=True):
+                pipe("a dog", num_inference_steps=4, max_sequence_length=16, height=256, width=256)
```

(Referring to @anijain2305's https://torchcompile-guards.hashnode.dev/inside-torchcompile-guards-how-they-work-what-they-cost-and-ways-to-optimize. Animesh WDYT about the changes?)

I also benchmarked this change with the following code:

<details>
<summary>Expand</summary>

```py
from diffusers import BitsAndBytesConfig as DiffusersBitsAndBytesConfig
from transformers import BitsAndBytesConfig as TransformersBitsAndBytesConfig
from diffusers import AutoModel, FluxPipeline
import argparse
import json
from contextlib import nullcontext
import torch

def get_bnb_config(torch_dtype=torch.bfloat16):
    quant_kwargs = {"load_in_4bit": True, "bnb_4bit_compute_dtype": torch_dtype, "bnb_4bit_quant_type": "nf4"}
    text_encoder_2_quant_config = TransformersBitsAndBytesConfig(**quant_kwargs)
    dit_quant_config = DiffusersBitsAndBytesConfig(**quant_kwargs)
    return text_encoder_2_quant_config, dit_quant_config

@torch.no_grad()
def initialize_pipeline(quant_config=None, torch_dtype=torch.bfloat16, compile=False):
    if quant_config is not None:
        text_encoder_2_quant_config, dit_quant_config = quant_config
    else:
        text_encoder_2_quant_config = dit_quant_config = None
    ckpt_id = "black-forest-labs/FLUX.1-dev"
        ckpt_id,
        subfolder="text_encoder_2",
        quantization_config=text_encoder_2_quant_config,
        torch_dtype=torch_dtype,
    )
    transformer = AutoModel.from_pretrained(
        ckpt_id,
        subfolder="transformer",
        quantization_config=dit_quant_config,
        torch_dtype=torch_dtype,
    )
        ckpt_id,
        transformer=transformer,
        torch_dtype=torch_dtype,
    ).to("cuda")

    if quant_config is not None:
        torch._dynamo.config.capture_dynamic_output_shape_ops = True
    if compile:
        pipe.transformer = torch.compile(pipe.transformer, mode="max-autotune", fullgraph=True)

    return pipe

    t0 = benchmark.Timer(
        stmt="f(*args, **kwargs)",
    )


def get_pipe_kwargs():
    pipe_kwargs = {
        "prompt": "A cat holding a sign that says hello world",
        "height": 1024,
        "width": 1024,
        "guidance_scale": 3.5,
        "num_inference_steps": 50,
        "max_sequence_length": 512,
    }
    return pipe_kwargs

def run_inference(pipe, pipe_kwargs):
    is_quant_compile = pipe_kwargs.pop("quant_compile", False)
    context = torch.compiler.set_stance(skip_guard_eval_unsafe=True) if is_quant_compile else nullcontext()
    with context:
        _ = pipe(**pipe_kwargs)


def main(args):
    torch.cuda.empty_cache()
    torch.cuda.reset_peak_memory_stats("cuda")
    torch.cuda.reset_accumulated_memory_stats("cuda")

    quant_config = get_bnb_config() if args.do_quant else None
    pipe = initialize_pipeline(quant_config=quant_config, compile=args.compile)
    pipe_kwargs = get_pipe_kwargs()
    if args.do_quant and args.compile:
        for _ in range(2):
            run_inference(pipe, pipe_kwargs)
            pipe_kwargs.update({"quant_compile": True})

    time = benchmark_fn(run_inference, pipe, pipe_kwargs)
    inference_memory = torch.cuda.max_memory_allocated() / (1024 ** 3)
    
    context = torch.compiler.set_stance(skip_guard_eval_unsafe=True) if args.do_quant and args.compile else nullcontext()
    with context:
        image = pipe(**get_pipe_kwargs()).images[0]
    artifact_dict = {"time": time, "memory": inference_memory}
    artifact_dict.update(vars(args))
    file_prefix = f"compile@{args.compile}-quant@{args.do_quant}"
    image.save(f"{file_prefix}.png")
    with open(f"{file_prefix}.json", "w") as f:
        json.dump(artifact_dict, f)

if __name__ == "__main__":
    parser.add_argument("--do_quant", action="store_true")
    parser.add_argument("--compile", action="store_true")
    args = parser.parse_args()
    main(args)
```

</details>

Results (H100):

```bash
{"time": "12.695", "memory": 14.968644142150879, "do_quant": true, "compile": false}
{"time": "8.674", "memory": 14.968156814575195, "do_quant": true, "compile": true}
```

Doesn't impact results:

image

tests/quantization/bnb/test_4bit.py
```diff
@@ -855,3 +858,43 @@ def test_fp4_double_unsafe(self):
...
+
+    def tearDown(self):
+        super().tearDown()
+        gc.collect()
+        backend_empty_cache(torch_device)
+        torch.compiler.reset()
+
+    @require_torch_version_greater("2.7.1")
+    def test_torch_compile_4bit(self):
+        torch._dynamo.config.capture_dynamic_output_shape_ops = True
```
matthewdouglas, line 877: We should only need this for 8bit, but I don't think it hurts to have it here either.

tests/quantization/bnb/test_4bit.py
matthewdouglas: Regarding that recompile, we've actually just merged the change described in bitsandbytes-foundation/bitsandbytes#1623 to main on bitsandbytes and will include it in the next 0.47 release. If I'm understanding correctly, that should help with the guard there as this condition would always be `True`. As it is right now on the first forward() it's `False`, and then `True` afterwards.

```E                   - 0/0: self._modules['norm_out']._modules['linear'].compute_type_is_set == False```

But in this circumstance the output isn't going to be changed so I think it's completely safe to skip the guard on it as a workaround right now.

Comment (sayakpaul) [reply tests/quantization/bnb/test_4bit.py]: Noted. I will flip that back when the release is ready then? 

If the test looks good to you, I will propagate it to 8bit test suite as well.

**Review by matthewdouglas — APPROVED** [review]

File: tests/quantization/bnb/test_4bit.py (hunk 1 above)
Comment (anijain2305) [reply tests/quantization/bnb/test_4bit.py]: I don't think this would work.

```
with torch._dynamo.config.patch(error_on_recompile=True), torch.compiler.set_stance(skip_guard_eval_unsafe=True):
```

In the sense, this does not give you the guarantee that an exception will be thrown on a recompilation.


`torch.compiler.set_stance(skip_guard_eval_unsafe=True)` can only be used when a user can guarantee that there are no going to be more recompilations. It drops large number of guards because of that assumption, and it is totally possible that `compute_type_is_set` guard gets dropped and you miss the recompilation.

Now in your case, it does not matter if `compute_type_is_set` is True or False as noted by @matthewdouglas, so it might be OK. But in general, you can't rely on cascading these 2 managers to fail on recompilations.


Aside from that, it is good to see the perf improvement. This means guard overhead is high. I am happy to go over a  performance profile to see whats going on.

Comment (sayakpaul) [reply tests/quantization/bnb/test_4bit.py]: > But in general, you can't rely on cascading these 2 managers to fail on recompilations.

What would be a good way to catch the failures then?

> This means guard overhead is high. I am happy to go over a performance profile to see whats going on.

Yes please. LMK if you want me to run something and I will do it. Will be a good lesson to note for the future similar scenarios.

Comment (anijain2305) [reply tests/quantization/bnb/test_4bit.py]: I mean you can use the config patch error_on_recompile, that's totally works! But you cannot use the skip_guard_eval_unsafe stance along with it. The latter requires a guarantee from the user that there has been a warmup, and there are no further recompilations.  If you can't guarantee, then the skip_guard_eval_unsafe should not be used.

Also is the comparison compared to no compile - 12.6 (no compile) vs 8.6 (compile)? Can you give one number ? Compile with the skip_guard_eval_unsafe stance and compile without the flag (the default)?

Comment (sayakpaul) [reply tests/quantization/bnb/test_4bit.py]: > I mean you can use the config patch error_on_recompile, that's totally works! But you cannot use the skip_guard_eval_unsafe stance along with it. The latter requires a guarantee from the user that there has been a warmup, and there are no further recompilations. If you can't guarantee, then the skip_guard_eval_unsafe should not be used.

Ah makes sense. I will modify the test accordingly then by removing the recompilation 
manager.

> Also is the comparison compared to no compile - 12.6 (no compile) vs 8.6 (compile)? Can you give one number ? Compile with the skip_guard_eval_unsafe stance and compile without the flag (the default)?

Here are the numbers:

```bash
{"time": "12.698", "memory": 14.968644142150879, "do_quant": true, "compile": false, "use_guard_stance": false}
{"time": "8.644", "memory": 14.968156814575195, "do_quant": true, "compile": true, "use_guard_stance": false}
{"time": "8.678", "memory": 14.968156814575195, "do_quant": true, "compile": true, "use_guard_stance": true}

```

Hmm now it seems like the guard overhead isn't actually too high, no? In my earlier benchmarks with `bitsandbytes` nightly and PyTorch nightly, the performance with and without `fullgraph=True` while compilation wasn't that significant, like it is now. So, very likely, there was some regression that went uncaught. Apologies for the false alarm.

Benchmarking code: https://gist.github.com/sayakpaul/0db9d8eeeb3d2a0e5ed7cf0d9ca19b7

When using the stance, I ensure some warmup runs:
https://gist.github.com/sayakpaul/0db9d8eeeb3d2a0e5ed7cf0d9ca19b7d#file-bnb_compile_benchmark-py-L100

anijain2305: Yes, guards do not seem to be taking too much time here. This checks out because diffusion models have large latency, and guards end up being a tiny percentage

sayakpaul:
@DN6 LMK what you think of the test suite. The combinations target consumer GPU where using quantization is beneficial (cc: @asomoza). Would you be able to add this for GGUF, too? 

Also cc: @stevhliu I think we should try to document these combos of optims in an easy manner now that we know they work (I can help get latency and memory numbers).

tests/quantization/test_torch_compile_utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+import torch
+
+from diffusers import DiffusionPipeline
+from diffusers.utils.testing_utils import backend_empty_cache, require_torch_gpu, slow, torch_device
+
+
+@require_torch_gpu
+@slow
+class QuantCompileMiscTests(unittest.TestCase):
```
DN6: Can just call it QuantCompileTests
```suggestion
class QuantCompileTests(unittest.TestCase):
```

sayakpaul:
Also cc @a-r-r-o-w see if any of these tests can be propagated to TorchAO as well. LMK if there's anything I can help with.

PR #11681, opened 2025-06-09
[tests] model-level `device_map` clarifications

sayakpaul:
# What does this PR do?

Our `device_map` (model-level) documentation could be improved a bit. We could also test accepted values for `device_map`. This PR does it.

The PR relies on https://github.com/huggingface/diffusers/pull/11680 to get merged.

@Birch-san is it better?

### DIFF
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -986,14 +986,6 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
                 )
             else:
                 device_map = {"": device_map}
-        # {"": device} case.
-        elif isinstance(device_map, dict) and len(device_map) == 1:
-            device_value = list(device_map.values())[0]
-            if isinstance(device_value, str):
-                try:
-                    device_map = {"": torch.device(device_value)}
-                except RuntimeError:
-                    raise ValueError(f"Invalid value ({device_value}) specified in the {device_map=}.")
 
         if device_map is not None:
             if low_cpu_mem_usage is None:
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -986,6 +986,14 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
                 )
             else:
                 device_map = {"": device_map}
+        # {"": device} case.
+        elif isinstance(device_map, dict) and len(dict) == 1:
+            device_value = list(device_map.values())[0]
+            if isinstance(device_value, str):
+                try:
+                    device_map = {"": torch.device(device_value)}
+                except RuntimeError:
+                    raise ValueError(f"Invalid value ({device_value}) specified in the {device_map=}.")
 
         if device_map is not None:
             if low_cpu_mem_usage is None:
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -814,10 +814,30 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
                 Mirror source to resolve accessibility issues if you're downloading a model in China. We do not
                 guarantee the timeliness or safety of the source, and you should refer to the mirror site for more
                 information.
-            device_map (`str` or `Dict[str, Union[int, str, torch.device]]`, *optional*):
+            device_map (`Union[int, str, torch.Device]` or `Dict[str, Union[int, str, torch.device]]`, *optional*):
                 A map that specifies where each submodule should go. It doesn't need to be defined for each
                 parameter/buffer name; once a given module name is inside, every submodule of it will be sent to the
                 same device. Defaults to `None`, meaning that the model will be loaded on CPU.
+
+                Examples:
+
+                ```py
+                >>> from diffusers import AutoModel
+                >>> import torch
+
+                >>> # This works.
+                >>> model = AutoModel.from_pretrained(
+                ...     "stabilityai/stable-diffusion-xl-base-1.0", subfolder="unet", device_map="cuda"
+                ... )
+                >>> # This also works (integer accelerator device ID).
+                >>> model = AutoModel.from_pretrained(
+                ...     "stabilityai/stable-diffusion-xl-base-1.0", subfolder="unet", device_map=0
+                ... )
+                >>> # Specify a supported offloading strategy like "auto" also works.
+                >>> model = AutoModel.from_pretrained(
+                ...     "stabilityai/stable-diffusion-xl-base-1.0", subfolder="unet", device_map="auto"
+                ... )
+                ```
 
                 Set `device_map="auto"` to have 🤗 Accelerate automatically compute the most optimized `device_map`. For
                 more information about each option see [designing a device
@@ -1390,7 +1410,7 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
         low_cpu_mem_usage: bool = True,
         dtype: Optional[Union[str, torch.dtype]] = None,
         keep_in_fp32_modules: Optional[List[str]] = None,
-        device_map: Dict[str, Union[int, str, torch.device]] = None,
+        device_map: Dict[Union[str, int, torch.device], Union[int, str, torch.device]] = None,
         offload_state_dict: Optional[bool] = None,
         offload_folder: Optional[Union[str, os.PathLike]] = None,
         dduf_entries: Optional[Dict[str, DDUFEntry]] = None,
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -987,7 +987,7 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
             else:
                 device_map = {"": device_map}
         # {"": device} case.
-        elif isinstance(device_map, dict) and len(dict) == 1:
+        elif isinstance(device_map, dict) and len(device_map) == 1:
             device_value = list(device_map.values())[0]
             if isinstance(device_value, str):
                 try:
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -814,7 +814,7 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
                 Mirror source to resolve accessibility issues if you're downloading a model in China. We do not
                 guarantee the timeliness or safety of the source, and you should refer to the mirror site for more
                 information.
-            device_map (`Union[int, str, torch.Device]` or `Dict[str, Union[int, str, torch.device]]`, *optional*):
+            device_map (`Union[int, str, torch.device]` or `Dict[str, Union[int, str, torch.device]]`, *optional*):
                 A map that specifies where each submodule should go. It doesn't need to be defined for each
                 parameter/buffer name; once a given module name is inside, every submodule of it will be sent to the
                 same device. Defaults to `None`, meaning that the model will be loaded on CPU.
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -1410,7 +1410,7 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
         low_cpu_mem_usage: bool = True,
         dtype: Optional[Union[str, torch.dtype]] = None,
         keep_in_fp32_modules: Optional[List[str]] = None,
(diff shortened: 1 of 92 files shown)


GrigoryEvko:
Hi guys, I tinkered a bit with device_map code with Opus, replaced diffusers internal usage of device_map in favor of accelerate functions. I will test now on multi gpu machines, please share your thoughts! 
 https://github.com/huggingface/diffusers/commit/3635b4146abbccf0201513bd1f6e158e1f1f57c2
 
Next step would be making diffusers modeling_utils on par with transformers modeling_utils to allow Torch TP usage, Deepspeed and FSDP/FSDP2 usage, etc.

sayakpaul:
> I tinkered a bit with device_map code with Opus, replaced diffusers internal usage of device_map in favor of accelerate functions.

We already make use of accelerate fun tions to facilitate device mapping. So, please open a PR as that is easier for the team to navigate through. 

> Next step would be making diffusers modeling_utils on par with transformers modeling_utils to allow Torch TP usage, Deepspeed and FSDP/FSDP2 usage, etc.

This is in the works already by @a-r-r-o-w. Again, it is much better to converse about it through PRs and issues. 

This PR is exactly what the description of the PR says. So, let’s switch to other threads for discussing related things further.

GrigoryEvko:
>We already make use of accelerate fun tions to facilitate device mapping. So, please open a PR as that is easier for the team to navigate through.

Sure, I will clean up, test and then submit the PR. I just noticed the team started to work on the issue and decided to comment here.

src/diffusers/models/modeling_utils.py
```diff
@@ -1387,7 +1410,7 @@ def _load_pretrained_model(
         low_cpu_mem_usage: bool = True,
         dtype: Optional[Union[str, torch.dtype]] = None,
         keep_in_fp32_modules: Optional[List[str]] = None,
-        device_map: Dict[str, Union[int, str, torch.device]] = None,
+        device_map: Dict[Union[str, int, torch.device], Union[int, str, torch.device]] = None,
```
a-r-r-o-w: Just confirming if this change is correct. Can the dictionary key be a str/int/torch.device? I thought only the values can be that

tests/models/unets/test_models_unet_2d_condition.py
```diff
@@ -1084,6 +1084,25 @@ def test_load_sharded_checkpoint_device_map_from_hub_local_subfolder(self):
...
+            _ = self.model_class.from_pretrained(
+                "hf-internal-testing/unet2d-sharded-dummy-subfolder", subfolder="unet", device_map=-1
+            )
+
+        msg_substring = "You can't pass device_map as a negative int"
+        assert msg_substring in str(err_ctx.exception)
+
+    @parameterized.expand([0, "cuda", torch.device("cuda"), torch.device("cuda:0")])
+    @require_torch_gpu
+    def test_passing_non_dict_device_map_works(self, device_map):
```
a-r-r-o-w, line 1103: Can we test some more cases like: `{"": torch.device("meta"), "decoder": torch.device("cuda")}`? For example, if this was a VAE, the intention here is to not load the encoder weights, but directly load the decoder weights to device.

Additionally, we should probably run device map tests for all models IMO (can be taken up in future PR)

tests/models/unets/test_models_unet_2d_condition.py
```diff
@@ -1084,6 +1084,25 @@ def test_load_sharded_checkpoint_device_map_from_hub_local_subfolder(self):
         assert loaded_model
         assert new_output.sample.shape == (4, 4, 16, 16)
 
+    def test_wrong_device_map_raises_error(self):
```
a-r-r-o-w: Testing more of the failure code paths (if there are any; I am not fully aware of the relevant parts of the codebase) could be nice

a-r-r-o-w:
@GrigoryEvko re: FSDP/FSDP2; This is already supported via accelerate, and the intented way for usage is to enable it via the accelerate config and modify some parts of the codebase based on accelerate documentation. Could you elaborate what it would mean to support this natively?

re: TP; TP does not really help much with e2e speed with current diffusion model sizes IME. I will be prioritizing supporting CP natively instead in the coming weeks. For TP, it's something we could consider once CP is supported.

re: DeepSpeed; This is also something that should be enabled via accelerate in your training codebase

tests/models/unets/test_models_unet_2d_condition.py
Birch-san: I agree this would be fantastic, but we can probably tackle that in a separate PR, and leave the scope of this one to tests/docs/bugfixes/assertions.

src/diffusers/models/modeling_utils.py
```diff
@@ -814,14 +814,37 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
                 Mirror source to resolve accessibility issues if you're downloading a model in China. We do not
                 guarantee the timeliness or safety of the source, and you should refer to the mirror site for more
                 information.
-            device_map (`str` or `Dict[str, Union[int, str, torch.device]]`, *optional*):
+            device_map (`Union[int, str, torch.Device]` or `Dict[str, Union[int, str, torch.device]]`, *optional*):
```
Birch-san: should be cased as `torch.device` for consistency with the dict value's type hint and the class name as written in torch

Birch-san:
does `device_map=device` work (where the `device` variable is a torch.device)? the docs say it should work, but when I tried this, I got a "device cuda is invalid" error.

likewise, does `device_map={'': device}` work? the docs say it should work, but when I tried this, I got a "device cuda is invalid" error.

```python
from diffusers.models.unets.unet_2d_condition import UNet2DConditionModel
import torch

device = torch.device('cuda')
unet: UNet2DConditionModel = UNet2DConditionModel.from_pretrained(
    'NovelAI/nai-anime-v1-full',
    torch_dtype=torch.float16,
    use_safetensors=True,
    subfolder='unet',
    device_map={'': device},
).eval()
```

I'm using diffusers==0.32.2 and accelerate==1.4.0 if it matters.

src/diffusers/models/modeling_utils.py
a-r-r-o-w: From the examples added, it looks like this should be: `Union[str, int, torch.device, Dict[str, [str, int, torch.device]]]`

src/diffusers/models/modeling_utils.py
```diff
@@ -814,14 +814,37 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
                 A map that specifies where each submodule should go. It doesn't need to be defined for each
                 parameter/buffer name; once a given module name is inside, every submodule of it will be sent to the
                 same device. Defaults to `None`, meaning that the model will be loaded on CPU.
 
+                Examples:
```
Birch-san, line 822: it looks like this is just documenting the scalar cases. the bit that I need docs for is the dictionary convention. `{'': device.type}` as the simpest valid input is extremely hard to guess. there really needs to be an explanation of what the key of the dictionary means.

src/diffusers/models/modeling_utils.py
Birch-san: I assume you didn't mean to type `Dict[str, int, torch.device]`

sayakpaul: Fixed.

a-r-r-o-w: Yeah my bad, updated

tests/models/unets/test_models_unet_2d_condition.py
sayakpaul: > Additionally, we should probably run device map tests for all models IMO (can be taken up in future PR)

We have a bunch of `device_map` related tests already in https://github.com/huggingface/diffusers/blob/main/tests/models/test_modeling_common.py.

I can shift the current ones being added through this PR to `test_modeling_common.py` in a separate PR.

> Can we test some more cases like: {"": torch.device("meta"), "decoder": torch.device("cuda")}? For example, if this was a VAE, the intention here is to not load the encoder weights, but directly load the decoder weights to device.

Feel free to add that in a separate PR.

tests/models/unets/test_models_unet_2d_condition.py
sayakpaul: We check if the `device_map` is respected in https://github.com/huggingface/diffusers/blob/main/tests/models/test_modeling_common.py. Then we're testing for invalid `device_map` values for non-dict entries. I added another one in eb913e22e.

Many errors are already handled in `accelerate`. 

So, collectively, we should now be good I think.

src/diffusers/models/modeling_utils.py
sayakpaul: Done.

sayakpaul:
@Birch-san I ran the following with this PR branch and it worked:

```py
from diffusers.models.unets.unet_2d_condition import UNet2DConditionModel
import torch

for device_map in [torch.device('cuda'), {'': torch.device('cuda')}]:
    unet: UNet2DConditionModel = UNet2DConditionModel.from_pretrained(
        'NovelAI/nai-anime-v1-full',
        torch_dtype=torch.float16,
        use_safetensors=True,
        subfolder='unet',
        device_map=device_map,
    ).eval()
```

The `test_passing_non_dict_device_map_works()` tests should ensure `device_map`s that have non-dict entries, including the `torch.device` type are working.

Will add `dict` types in the tests as well.

src/diffusers/models/modeling_utils.py
sayakpaul: Cc: @SunMarc @stevhliu

How is this documented in `transformers`?

SunMarc: You can find this here https://huggingface.co/docs/accelerate/en/concept_guides/big_model_inference#the-devicemap

SunMarc: also can you fix the typing for the `DiffusionPipeline` from_pretrained for device_map since for this specific function, we only allow `balanced` value ?

sayakpaul: Done in 407b67f6f.

sayakpaul: @Birch-san I clarified the docs to include the case of `{"": torch.device("cuda")}` and have added tests for it, too. For other possible and valid `dict` inputs to `device_map`, I would have to defer you to https://huggingface.co/docs/accelerate/en/concept_guides/big_model_inference#the-devicemap as you can notice it's hard to specify that beforehand without doing a bit of investigation. 

So, I would suggest loading your model with "auto" `device_map`, first. And then printing `(model.hf_device_map)` to get a much better handle. This way, you will have a reasonable starting point which you could then use to tweak things around a bit.

SunMarc (approved):
Thanks for making this better !

src/diffusers/models/modeling_utils.py
Birch-san: it's not clear to me from the accelerate docs how the key is used. the fact that `''` works suggests there's some kind of pattern-matching or special-cases, which aren't documented.

sayakpaul: Going to defer to @SunMarc for that (again).

src/diffusers/models/modeling_utils.py
```diff
@@ -957,6 +986,14 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
                 )
             else:
                 device_map = {"": device_map}
+        # {"": device} case.
+        elif isinstance(device_map, dict) and len(device_map) == 1:
```
yiyixuxu: @sayakpaul 
does this not already work for you on `main`?

```python
from diffusers import UNet2DConditionModel
import torch


device =torch.device('cuda')

model = UNet2DConditionModel.from_pretrained(
    "stable-diffusion-v1-5/stable-diffusion-v1-5",
    torch_dtype=torch.float16,
    subfolder="unet",
    device_map={"": device},
)
print("model.device: ", model.device)
```

yiyixuxu: or this 

```python
from diffusers import UNet2DConditionModel
import torch


# device =torch.device('cuda')
device = "cuda"

model = UNet2DConditionModel.from_pretrained(
    "stable-diffusion-v1-5/stable-diffusion-v1-5",
    torch_dtype=torch.float16,
    subfolder="unet",
    device_map={"": device},
)
print("model.device: ", model.device)
```

sayakpaul: My bad. It's already supported. Just removed the code. Thanks for catching.

PR #11682, opened 2025-06-09
make group offloading work with disk/nvme transfers

sayakpaul:
# What does this PR do?

Group offloading is a crucial feature to provide a good speed-memory trade-off for large models on consumer hardware. However, since group offloading relies quite a bit on RAM usage, it can be bottlenecked by its availability. As such, for machines where GPU VRAM > available RAM or machines have limited RAM, group offloading can be far from ideal.

This PR takes a stab at supporting disk/NMVe serialization/deserialization inside group offloading so that users can use the secondary memory to onload/offload model params while also benefiting from the overlapping between compute and data transfer.

Below are some numbers I have gathered with this PR:

| Mode                                    | Time (s) |   RAM (GB) | GPU (GB) |
|-----------------------------------------|---------:|-----------:|---------:|
| base                                    |    6.594 |      1.838 |    33.85 |
| model CPU offload                       |   21.682 |     35.504 |    22.64 |
| sequential CPU offload                  |   68.406 |     33.196 |     2.41 |
| group offload                           |   47.693 |     36.421 |    11.68 |
| group offload with disk / NVMe support  |   55.467 |      2.814 |    11.68 |
| same + compile                          |   55.296 |      3.036 |    11.68 |

<details>
<summary>Code</summary>

```py
import torch
import psutil
import os
import json
import argparse

    t0 = benchmark.Timer(
        stmt="f(*args, **kwargs)",
    )

def run_inference(pipe, pipe_kwargs):
    _ = pipe(**pipe_kwargs)

def initialize_pipeline():
    )
    return pipe

def maybe_apply_offloading(pipe, args):
    if not args.model_cpu_offload and not args.seq_cpu_offload and not args.group_offload:
        pipe = pipe.to("cuda")
    else:
        if args.model_cpu_offload:
        elif args.seq_cpu_offload:
        elif args.group_offload:
            pipe.transformer.enable_group_offload(
                onload_device=torch.device("cuda"), 
                offload_device=torch.device("cpu"), 
                offload_type="block_level",
                num_blocks_per_group=1,
                use_stream=True,
                non_blocking=False,
                offload_to_disk=True if args.offload_to_disk else False,
                offload_path="." if args.offload_to_disk else None,
                record_stream=True
            )
            
            # For the rest of the components, just place on CUDA.
                if name != "transformer" and isinstance(component, torch.nn.Module):
                    component.cuda()

    return pipe


def main(args):
    process = psutil.Process(os.getpid())
    torch.cuda.reset_max_memory_allocated()
    torch.cuda.empty_cache()

    pipe = initialize_pipeline()
    pipe = maybe_apply_offloading(pipe, args)
    pipe_kwargs = {
        "prompt": "A cat holding a sign that says hello world",
        "height": 1024,
        "width": 1024,
        "guidance_scale": 3.5,
        "num_inference_steps": 28,
        "max_sequence_length": 512,
    }
    time = benchmark_fn(run_inference, pipe, pipe_kwargs)
    inference_memory = torch.cuda.max_memory_allocated() / (1024 ** 3)
    inference_memory = float(f"{inference_memory:.2f}")
    ram_bytes = process.memory_info().rss
    ram_gb = ram_bytes / (1024 ** 3)

    # report
    print(f"Peak GPU memory: {inference_memory} GB")
    print(f"Resident CPU memory (RSS): {ram_gb:.2f} GB")

    prefix = "base"
    for key, value in vars(args).items():
        prefix += f"_{key}@{value}"
    
    image = pipe(**pipe_kwargs).images[0]
    image.save(f"{prefix}.png")
    
    artifact_dict = {"time": time, "memory": inference_memory, "ram": ram_gb}
    artifact_dict.update(vars(args))
    with open(f"{prefix}.json", "w") as f:
        json.dump(artifact_dict, f)


if __name__ == "__main__":
    parser.add_argument("--model_cpu_offload", action="store_true")
    parser.add_argument("--seq_cpu_offload", action="store_true")
    parser.add_argument("--group_offload", action="store_true")
    parser.add_argument("--offload_to_disk", action="store_true")
    args = parser.parse_args()

    main(args)
```

</details>

Quality comparison:
preview

The stark background color difference in regular group offloading exists in the `main` branch as well. So, I am not sure what is happening there.

Group offloading with disk serialization/deserialization works with `torch.compile()`, too.

This PR is a PoC and hence, it has some things that can be made better. I'd be fine if the PR is completely dropped or if someone else wants to take it over and see it to completion. Otherwise, I am completely fine working on it.

@asomoza I think you will be quite interested in this one.


src/diffusers/hooks/group_offloading.py
```diff
@@ -59,6 +61,8 @@ def __init__(
         record_stream: Optional[bool] = False,
         low_cpu_mem_usage: bool = False,
         onload_self: bool = True,
+        offload_to_disk: bool = False,
```
a-r-r-o-w: Instead of two different arguments, would maybe converge both to `offload_to_disk_path` which would default to None. If a path is provided, user would mean to use disk offloading

src/diffusers/hooks/group_offloading.py
```diff
@@ -72,7 +76,29 @@ def __init__(
...
+                raise ValueError("`offload_path` must be set when `offload_to_disk=True`.")
+            self.safetensors_file_path = os.path.join(self.offload_path, f"group_{id(self)}.safetensors")
+
+            all_tensors = []
+            for module in self.modules:
+                all_tensors.extend(list(module.parameters()))
+                all_tensors.extend(list(module.buffers()))
+            all_tensors.extend(self.parameters)
+            all_tensors.extend(self.buffers)
+            all_tensors = list(dict.fromkeys(all_tensors))  # Remove duplicates
```
a-r-r-o-w, line 91: (nit): will there be duplicates? I cannot think of a quick example, so maybe we can remove

src/diffusers/hooks/group_offloading.py
```diff
@@ -72,7 +76,29 @@ def __init__(
...
         self.onload_self = onload_self
         self.low_cpu_mem_usage = low_cpu_mem_usage
-        self.cpu_param_dict = self._init_cpu_param_dict()
+
+        self.offload_to_disk = offload_to_disk
+        self.offload_path = offload_path
+        self._is_offloaded_to_disk = False
+
+        if self.offload_to_disk:
+            if self.offload_path is None:
```
a-r-r-o-w: (nit): probably could package this into a separate helper function like what's done with `_init_cpu_param_dict`

src/diffusers/hooks/group_offloading.py
```diff
@@ -124,6 +150,30 @@ def onload_(self):
...
+
+            with context:
+                if self.stream is not None:
+                    # Load to CPU, pin, and async copy to device for overlapping transfer and compute
+                    loaded_cpu_tensors = safetensors.torch.load_file(self.safetensors_file_path, device="cpu")
+                    for key, tensor_obj in self.key_to_tensor.items():
+                        pinned_tensor = loaded_cpu_tensors[key].pin_memory()
+                        tensor_obj.data = pinned_tensor.to(self.onload_device, non_blocking=self.non_blocking)
+                        if self.record_stream:
+                            tensor_obj.data.record_stream(current_stream)
```
a-r-r-o-w, line 162: I think cleaner approach would be to provide a callable to `map_location` (assuming we were using torch.load instead of safetensors), which for each tensor can pin and move to device. Do we know if there is a equivalent to passing a callable with safetensors? If not, this is okay too

src/diffusers/hooks/group_offloading.py
```diff
@@ -169,6 +219,18 @@ def onload_(self):
     @torch.compiler.disable()
     def offload_(self):
         r"""Offloads the group of modules to the offload_device."""
+        if self.offload_to_disk:
```
a-r-r-o-w: (nit): we probably need to refactor this a bit and break into smaller methods so we don't have to branch and do early-returns every time a new feature is added (we can do refactor once we have everything working, so not urgent)

src/diffusers/hooks/group_offloading.py
```diff
@@ -169,6 +219,18 @@ def onload_(self):
...
+            if not self._is_offloaded_to_disk:
+                os.makedirs(os.path.dirname(self.safetensors_file_path), exist_ok=True)
+                tensors_to_save = {
+                    key: tensor.data.to(self.offload_device) for tensor, key in self.tensor_to_key.items()
+                }
+                safetensors.torch.save_file(tensors_to_save, self.safetensors_file_path)
+                self._is_offloaded_to_disk = True
+
+            for tensor_obj in self.tensor_to_key.keys():
+                tensor_obj.data = torch.empty_like(tensor_obj.data, device=self.offload_device)
```
a-r-r-o-w, line 236: What's the reason for this to be different from the non-disk-offload counterpart? That is, is there a reason we're not doing `buffer.data.to(self.offload_device, non_blocking=self.non_blocking)`?

a-r-r-o-w (commented):
Really cool work at getting this started!

- Can we see some results with more compute heavy model like Wan?
- We probably need to look at some profiles to see if there is overlapping happening here when streams are used with disk-offload
- re: stark background color difference; Weird, I'll take a look

sayakpaul:
@a-r-r-o-w thanks for your comments. I will work on them.

> We probably need to look at some profiles to see if there is overlapping happening here when streams are used with disk-offload

I can gather this. Should we gather the CPU and GPU activities through the profiler and export a trace? If you have any references for me to consider, feel free to send over.

a-r-r-o-w:
> I can gather this. Should we gather the CPU and GPU activities through the profiler and export a trace? If you have any references for me to consider, feel free to send over.

I think CPU/GPU activities will measure all the operations in the model, so if you could collect the filtered stream-related operations and related onloading/offloading times, it'll be helpful!

You probably already know but for the readers, this will help with gathering the traces for visualization:
- https://docs.pytorch.org/docs/stable/torch.compiler_profiling_torch_compile.html
- https://docs.pytorch.org/tutorials/recipes/recipes/profiler_recipe.html
- https://ui.perfetto.dev/

sayakpaul:
@a-r-r-o-w for Wan:

|    |    time |   memory |      ram | model_cpu_offload   | seq_cpu_offload   | group_offload   | offload_to_disk   |
|---:|--------:|---------:|---------:|:--------------------|:------------------|:----------------|:------------------|
|  0 | 237.827 |    15.15 |  2.80788 | False               | False             | True            | True              |
|  1 | 151.684 |    41.89 |  1.96027 | False               | False             | False           | False             |
|  2 | 226.929 |    15.15 | 42.4634  | False               | False             | True            | False             |
|  3 | 172.826 |    28.7  | 39.4348  | True                | False             | False           | False             |

a-r-r-o-w:
Awesome, thanks for sharing! The numbers look good.

re: weird color results with group offloading; i looked into it and seems to only happen with block level (I don't know why yet). I think it could be because of some incorrect/missing synchronization somewhere, so will try to fix. If you use leaf_level, it should produce the same result.

src/diffusers/hooks/group_offloading.py
sayakpaul: Since it's a one-liner now, I think it's okay here.

src/diffusers/hooks/group_offloading.py
sayakpaul: There shouldn't really be. But I kept it to prevent edge-cases while reading something similar.

src/diffusers/hooks/group_offloading.py
sayakpaul: Do we know if there would be other alternatives to this code path? If not, I think it's better as is. From skimming through the documentation of `safetensors`, I couldn't find any equivalent of `map_location`.

src/diffusers/hooks/group_offloading.py
sayakpaul: Indeed. Can I do it in an immediate follow-up PR so that it's easier to review?

src/diffusers/hooks/group_offloading.py
sayakpaul: So, we first free up the memory of the accelerator with:

https://github.com/huggingface/diffusers/blob/d8179b10d3fa4bb4e8d1a0d2ff85b36463717f90/src/diffusers/hooks/group_offloading.py#L222

However, since we're also optimizing for RAM usage (can be made clearer through documentation I believe), we need to free up the RAM that is holding the tensor data. After the data has been safely written from RAM to the disk, this step replaces the large data tensor in RAM with a memory-less placeholder. This allows the memory to be released.

sayakpaul:
@a-r-r-o-w could I ask for another review at this point? I have also added a simple test to make sure it's working but I can also add a heavier integration test that checks for VRAM and RAM usage. Apart from that, I think only doc is missing.

src/diffusers/hooks/group_offloading.py
```diff
@@ -481,6 +553,7 @@ def _apply_group_offloading_block_level(
     module: torch.nn.Module,
     num_blocks_per_group: int,
     offload_device: torch.device,
+    offload_to_disk_path: Optional[str],
```
a-r-r-o-w: (nit): let's move this to the bottom of argument list. some of my personal code will break with this change because I pass args instead of by name. Might be inconvenient for other users too

src/diffusers/hooks/group_offloading.py
```diff
@@ -358,6 +423,7 @@ def apply_group_offloading(
     onload_device: torch.device,
     offload_device: torch.device = torch.device("cpu"),
     offload_type: str = "block_level",
+    offload_to_disk_path: Optional[str] = None,
```
a-r-r-o-w: (nit): Let's move this to the bottom arg

src/diffusers/hooks/group_offloading.py
```diff
@@ -205,13 +271,12 @@ class GroupOffloadingHook(ModelHook):
...
-        self,
-        group: ModuleGroup,
-        next_group: Optional[ModuleGroup] = None,
-    ) -> None:
+    def __init__(self, group: ModuleGroup, next_group: Optional[ModuleGroup] = None) -> None:
         self.group = group
         self.next_group = next_group
+        # map param/buffer name -> file path
+        self.param_to_path: Dict[str, str] = {}
+        self.buffer_to_path: Dict[str, str] = {}
```
a-r-r-o-w: Seems like not used?

a-r-r-o-w (commented):
Thanks, the current implementation looks good to me for an initial merge after cleanup/docs! I haven't had the chance to look into memory profile yet, but the Wan benchmarks you shared look correct and as expected to me

sayakpaul:
@a-r-r-o-w thanks! I think your comments are all addressed now. I have added docs, as well. LMK if there are any other comments address. I am also working further optimizing this. Stay tuned 🤗

sayakpaul:
@stevhliu could you review the docs once you have a moment?

sayakpaul:
Failing tests are unrelated as I have locally run them on a CPU and they pass. Going to merge after this round of CI (have checked with @DN6 offline and we're good to merge).

PR #11683, opened 2025-06-10
[WIP] device_map rework and direct weights loading

GrigoryEvko:
This PR improves the diffusers `device_map` functionality, enabling support for all accelerate modes (auto, balanced, balanced_low_0, sequential) and all accelerate device map variants across various devices, including meta and disk. This PR extends `device_map` valid usage to pipelines (e.g., DiffusionPipeline), whereas previously only models were supported.

The PR attempts to implement direct weight loading to the device without first loading on CPU when `device_map` is not None and all components of the model have the same device argument. Known limitation is that safetensors still requires RAM + swap to be larger than the lazy-loaded file (https://github.com/huggingface/safetensors/issues/528). Otherwise, direct pipeline loading works as expected, and no copy of weights is created on CPU.

Another limitation is that for flawless device_map='auto' usage all modules including controlnets, ipadapters, models, etc. shoud have valid `_no_split_modules` entry, currently some controlnet files do not satisfy this requirement.

When user loads a model with from_pretrained method in a process with an existing logger, a helper message is shown to indicate that there's a more efficient and explicit way to load the pipeline:
```
💡 For memory-efficient loading across multiple devices, consider using device mapping:
device_map={'transformer': 'cuda:0', 'text_encoder': 'cuda:0', 'vae': 'cuda:0'}
```
Since the diffusers library heavily relies on the peft and accelerate libraries, both were added to dependencies. Additionally, the torch requirement version was bumped from 1.4 to 1.9.

### DIFF
--- a/examples/community/pipeline_stable_diffusion_xl_ipex.py
+++ b/examples/community/pipeline_stable_diffusion_xl_ipex.py
@@ -73,7 +73,7 @@ EXAMPLE_DOC_STRING = """
 
         >>> # SDXL-Turbo, a distilled version of SDXL 1.0, trained for real-time synthesis
         >>> pipe = StableDiffusionXLPipelineIpex.from_pretrained(
-        ...     "stabilityai/sdxl-turbo", use_safetensors=True
+        ...     "stabilityai/sdxl-turbo", low_cpu_mem_usage=True, use_safetensors=True
         ... )
 
         >>> num_inference_steps = 1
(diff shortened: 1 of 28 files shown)


GrigoryEvko:
@Birch-san @sayakpaul @DN6 @a-r-r-o-w

yiyixuxu:
hey thanks for opening the PR
it is unclear to me what's the purpose of this PR due to the poorly written description. 
Note that the device_map on diffusers (pipeine-level) has completely meanings than these on model_level. we've decided to only support "balanced" and handle that logic inside diffusers 

I think it is ok to support direct loading with a custom map, though

yiyixuxu:
closing this PR due to its low quality

PR #11685, opened 2025-06-10
[WIP] Refactor Attention Modules

DN6:
# What does this PR do?

Draft PR to address some issues with the attention modules. 

1. Moves a number of transformer related blocks/objects in the attention module that would have a better home under models/transformers

2. Moves the definition of the Attention module into `attention.py`,  rather than have it live in `attention_prcessors.py`. 

3. We have a very large number of processors, but with #11368 we should no longer need a good chunk of them and they can be deprecated. I think with these changes we would end up with ~3 processors per model (Attn, IPAdapter, PAG)

4. Make it so that we can bump up our minimum supported Torch version to `>=2.0` and use the `F.sdpa` API for all processors. 

5. There was some discussion around naming of the processors:
https://huggingface.slack.com/archives/C065E480NN9/p1737130514639479
We landed on calling the processors something like AttnProcessorsSDPA, but with #11368 we no longer need to use a dedicated processor per backend, so I think it's okay to just have the class be named AttnProcessor. 

6. Move processor definitions into the model files, so we don't end up with very large files containing all processors.     

7. Introduce AttentionModuleMixin that contains all common methods related to attention operations. New attention modules would inherit from this class. 

8. Introduce an `AttentionMixin` for models so that methods like `set_processor` are not duplicated across models. Although we can probably just add the methods of this Mixin to `ModelMixin`

9. Using Flux as an example here to show how we can define a single Processor to support both fused/unfused qkv projections. 

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/models/transformers/transformer_flux.py
+++ b/src/diffusers/models/transformers/transformer_flux.py
@@ -360,7 +360,6 @@ class FluxTransformerBlock(nn.Module):
         self.norm1 = AdaLayerNormZero(dim)
         self.norm1_context = AdaLayerNormZero(dim)
 
-        # Use specialized FluxAttention instead of generic Attention
         self.attn = FluxAttention(
             query_dim=dim,
             cross_attention_dim=None,


src/diffusers/models/transformers/modeling_common.py
```diff
@@ -0,0 +1,1251 @@
+# Copyright 2024 The HuggingFace Team. All rights reserved.
```
a-r-r-o-w, line 1: I have no strong feelings about the idea of having a common modeling file, but I think if we could minify the implementation a bit and specialize per-model, it will be really clean. Basically, the entire model implementation is top-to-bottom present in a single without requiring imports from `diffusers.*`. I can help complete this if we're okay with the idea

src/diffusers/models/transformers/modeling_common.py
```diff
@@ -0,0 +1,1251 @@
...
+from ..activations import GEGLU, GELU, ApproximateGELU, FP32SiLU, LinearActivation, SwiGLU
+from ..attention_processor import Attention, JointAttnProcessor2_0
+from ..embeddings import SinusoidalPositionalEmbedding
+from ..normalization import AdaLayerNorm, AdaLayerNormContinuous, AdaLayerNormZero, RMSNorm, SD35AdaLayerNormZeroX
+
+
+logger = logging.get_logger(__name__)
+
+
+def _chunked_feed_forward(ff: nn.Module, hidden_states: torch.Tensor, chunk_dim: int, chunk_size: int):
```
a-r-r-o-w, line 31: Ideally, I would like to try and deprecate this method. We should, in theory, be able to chunk across all the dimensions on which a model does not operate on. We discussed a bit in the past about a SplitInferenceHook but I never quite got around to it. I can revive the idea if it sounds like something we would be okay with doing.

src/diffusers/models/transformers/modeling_common.py
```diff
@@ -0,0 +1,1251 @@
...
+            ff_output = gate_mlp * ff_output
+
+        hidden_states = ff_output + hidden_states
+        if hidden_states.ndim == 4:
+            hidden_states = hidden_states.squeeze(1)
+
+        return hidden_states
+
+
+class LuminaFeedForward(nn.Module):
```
a-r-r-o-w, line 592: Let's move to Lumina file?

src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -42,6 +36,270 @@
...
+            if attn.norm_added_k is not None:
+                encoder_key = attn.norm_added_k(encoder_key)
+
+            # Concatenate for joint attention
+            query = torch.cat([encoder_query, query], dim=2)
+            key = torch.cat([encoder_key, key], dim=2)
+            value = torch.cat([encoder_value, value], dim=2)
+
+        if image_rotary_emb is not None:
+            from ..embeddings import apply_rotary_emb
```
a-r-r-o-w, line 120: For the attention refactor, I think having per-model implementation of RoPE preparation would be nice. This aligns with having one-file-per-model and the full implementation being available at once place

src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -42,6 +36,270 @@
...
+            hidden_states = attn.to_out[0](hidden_states)
+            hidden_states = attn.to_out[1](hidden_states)
+            encoder_hidden_states = attn.to_add_out(encoder_hidden_states)
+
+            return hidden_states, encoder_hidden_states
+        else:
+            return hidden_states
+
+
+class FluxIPAdapterAttnProcessor(torch.nn.Module):
```
a-r-r-o-w, line 150: I don't know the context behind making an attention processor a `torch.nn.Module` and why this was done in the past. Is there a way we could decouple the module part from the attention processor by creating a new "AttentionIPAdapter" class and setting its processor to this?

src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -42,6 +36,270 @@
...
+                current_ip_hidden_states = current_ip_hidden_states.to(ip_query.dtype)
+                ip_attn_output += scale * current_ip_hidden_states
+
+            return hidden_states, encoder_hidden_states, ip_attn_output
+        else:
+            return hidden_states
+
+
+@maybe_allow_in_graph
+class FluxAttention(Attention):
```
a-r-r-o-w, line 295: Nice, I like this. Although, my personal preference is to completely move away from centralized classes for model definitions and just implementing the relevant part directly. It's probably going to be a lot of work, but will result in super lean implementations without all the if-statements

src/diffusers/models/attention.py
```diff
@@ -11,36 +11,814 @@
...
 
 
-def _chunked_feed_forward(ff: nn.Module, hidden_states: torch.Tensor, chunk_dim: int, chunk_size: int):
-    # "feed_forward_chunk_size" can be used to save memory
-    if hidden_states.shape[chunk_dim] % chunk_size != 0:
-        raise ValueError(
-            f"`hidden_states` dimension to be chunked: {hidden_states.shape[chunk_dim]} has to be divisible by chunk size: {chunk_size}. Make sure to set an appropriate `chunk_size` when calling `unet.enable_forward_chunking`."
+class AttentionMixin:
+    @property
+    def attn_processors(self) -> Dict[str, AttentionProcessor]:
```
a-r-r-o-w, line 47: Maybe we could simplify this function a bit by rewriting as a loop instead of recursing

src/diffusers/models/attention.py
```diff
@@ -11,36 +11,814 @@
...
+    def set_attn_processor(self, processor: Union[AttentionProcessor, Dict[str, AttentionProcessor]]):
+        r"""
+        Sets the attention processor to use to compute attention.
+
+        Parameters:
+            processor (`dict` of `AttentionProcessor` or only `AttentionProcessor`):
+                The instantiated processor class or a dictionary of processor classes that will be set as the processor
+                for **all** `Attention` layers.
+
+                If `processor` is a dict, the key needs to define the path to the corresponding cross attention
```
a-r-r-o-w, line 79: (nit): This docstring needs an update because it's not just cross attention processors

src/diffusers/models/attention.py
```diff
@@ -11,36 +11,814 @@
...
+                fn_recursive_add_processors(f"{name}.{sub_name}", child, processors)
+
+            return processors
+
+        for name, module in self.named_children():
+            fn_recursive_add_processors(name, module, processors)
+
+        return processors
+
+    def set_attn_processor(self, processor: Union[AttentionProcessor, Dict[str, AttentionProcessor]]):
```
a-r-r-o-w, line 70: I think this implementation could be simplified a bit too to not use recursion

src/diffusers/models/attention.py
```diff
@@ -11,36 +11,814 @@
...
+
+    def fuse_qkv_projections(self):
+        """
+        Enables fused QKV projections. For self-attention modules, all projection matrices (i.e., query, key, value)
+        are fused. For cross-attention modules, key and value projection matrices are fused.
+
+        """
+        for _, attn_processor in self.attn_processors.items():
+            if "Added" in str(attn_processor.__class__.__name__):
+                raise ValueError("`fuse_qkv_projections()` is not supported for models having added KV projections.")
```
a-r-r-o-w, line 112: IMO, we should support fusing the KV projections into single linear. Anything that's commonly used for inference optimization would benefit from having a reference implementation. Not really a priority though, so just making a note

src/diffusers/models/attention.py
```diff
@@ -11,36 +11,814 @@
...
+        from .attention_dispatch import AttentionBackendName
+
+        available_backends = {x.value for x in AttentionBackendName.__members__.values()}
+        if backend not in available_backends:
+            raise ValueError(f"`{backend=}` must be one of the following: " + ", ".join(available_backends))
+
+        backend = AttentionBackendName(backend.lower())
+        self.processor._attention_backend = backend
+
+    def set_use_npu_flash_attention(self, use_npu_flash_attention: bool) -> None:
```
a-r-r-o-w, line 181: Do we want to keep the npu/xla/xformers-specific functions here? I think attention backends should have it covered already, no?

src/diffusers/models/attention.py
```diff
@@ -11,36 +11,814 @@
...
+                            dtype, *_ = op_fw.SUPPORTED_DTYPES
+                        q = torch.randn((1, 2, 40), device="cuda", dtype=dtype)
+                        _ = xformers.ops.memory_efficient_attention(q, q, q)
+                except Exception as e:
+                    raise e
+
+                self.set_attention_backend("xformers")
+
+    @torch.no_grad()
+    def fuse_projections(self):
```
a-r-r-o-w, line 258: Having a fuse/unfuse method per "Attention" class (which may exist at per-model level if we are aligned with my above comments) would be nice and super clean IMO

src/diffusers/models/attention.py
```diff
@@ -11,36 +11,814 @@
...
+
+        if hasattr(self, "to_kv"):
+            delattr(self, "to_kv")
+
+        if hasattr(self, "to_added_qkv"):
+            delattr(self, "to_added_qkv")
+
+        self.fused_projections = False
+
+    def set_attention_slice(self, slice_size: int) -> None:
```
a-r-r-o-w, line 337: Ideally, I would like to completely get rid of the 4-5 functions that follow and do model-specific implementations. For batch/head slicing, re: SplitInferenceHook.

yiyixuxu:
should we close this one now?

PR #11690, opened 2025-06-11
fix "Expected all tensors to be on the same device, but found at least two devices" error

yao-matrix:


yao-matrix:
@SunMarc , could you pls help review? Thx very much.

src/diffusers/models/unets/unet_2d_blocks.py
```diff
@@ -2557,7 +2557,8 @@ def forward(
                     b1=self.b1,
                     b2=self.b2,
                 )
-
+            if hidden_states.device != res_hidden_states.device:
+                res_hidden_states = res_hidden_states.to(hidden_states.device)
```
SunMarc: we shouldn't need that since both hidden_states and res_hidden_states should be on the same device no ? The pre-forward hook added by accelerate should be move all the inputs to the same device.

yao-matrix: @SunMarc , i suppose this is a corner case? `torch.cat` is a weight-less function, so seems cannot covered by the pre-forward hook set by `accelerate`...

SunMarc: I mean since `hidden_states` and `res_hidden_states_tuple` are in the forward definition, they should be moved to the same device by the pre-forward hook added by accelerate

yao-matrix: @SunMarc We run into a corner case here. Since we have 8 cards here, so the determined `device_map`(by https://github.com/huggingface/diffusers/blob/1bc6f3dc0f21779480db70a4928d14282c0198ed/src/diffusers/models/model_loading_utils.py#L64C5-L64C26) is 

> device_map: OrderedDict([('conv_in', 0), ('time_proj', 0), ('time_embedding', 0), ('down_blocks.0', 0), ('down_blocks.1.resnets.0', 1), ('up_blocks.0.resnets.0', 1), ('up_blocks.0.resnets.1', 2), ('up_blocks.0.upsamplers', 2), ('up_blocks.1', 3), ('mid_block.attentions', 3), ('conv_norm_out', 4), ('conv_act', 4), ('conv_out', 4), ('mid_block.resnets', 4)])

We can see `UpBlock` is not the atomic module, its submodules are assigned to different devices(up_blocks.0.resnets.0, up_blocks.0.resnets.1), so pre-hook for `UpBlock` will not help in this case. And since `torch.cat` is not pre-hooked(and cannot since it's a function rather than a module?), so the issue happens.   

If there is no a `torch.cat` btw the sub-blocks in `UpBlock`, things will be all fine.

yao-matrix: @SunMarc, need your inputs in how to proceed for this corner case, thx.

yao-matrix: @SunMarc  We can see a similar case in `transformers` ut `pytest -rA tests/models/chameleon/test_modeling_chameleon.py::ChameleonVision2SeqModelTest::test_model_parallel_beam_search` w/ 2 cards, the error log is "RuntimeError: Expected all tensors to be on the same device, but found at least two devices,src/transformers/models/chameleon/modeling_chameleon.py", the reason is even `residual` is in the same device as `hidden_states` at the beginning, but after they went through some operators as both input and output, they finally placed to different device, but when they come to `+` which is not a nn.Module(so accelerate cannot pre-hook it), error happens.  Do you have some insights on such issues?

yao-matrix: @SunMarc, could you share your insights on the issue i mentioned above? thx very much.

SunMarc: Sorry for the long wait @yao-matrix , if you add `UpBlock2D` in `_no_split_modules` of `UNet2DConditionModel`, the test should pass ! 
`_no_split_modules = ["BasicTransformerBlock", "ResnetBlock2D", "CrossAttnUpBlock2D", "UpBlock2D"]`
I've tested on my end and it works. 

As for ChameleonVision2SeqModelTest, we probably need to also update _no_split_modules.

yao-matrix:
@SunMarc , done thx, pls help review again.

SunMarc (approved):
LGTM !

tests/models/test_modeling_common.py
```diff
@@ -1945,7 +1944,7 @@ def test_push_to_hub_library_name(self):
         delete_repo(self.repo_id, token=TOKEN)
 
 
-@require_torch_gpu
+@require_torch_accelerator
 @require_torch_2
```
sayakpaul, line 1948: This change is unrelated to this PR. Going forward prefer not including unrelated changes in a particular PR.

sayakpaul (approved):
Thanks!

tests/models/test_modeling_common.py
yao-matrix: sure, will follow the rule going forward.

tests/models/unets/test_models_unet_2d_condition.py
```diff
@@ -358,7 +358,7 @@ class UNet2DConditionModelTests(ModelTesterMixin, UNetTesterMixin, unittest.Test
     model_class = UNet2DConditionModel
     main_input_name = "sample"
     # We override the items here because the unet under consideration is small.
-    model_split_percents = [0.5, 0.3, 0.4]
+    model_split_percents = [0.5, 0.34, 0.4]
```
yao-matrix, line 362: since `UpBlock2D` is not split now, need update 0.3->0,34 split percent to make UT pass

yao-matrix:
@sayakpaul , pls help retrigger the ci, thx very much.

PR #11696, opened 2025-06-11
Avoid DtoH sync from access of nonzero() item in scheduler

jbschlosser:
# What does this PR do?

(discussed with @sayakpaul in Slack) I have been optimizing inference performance for the Flux model and saw a DtoH sync point in the performance trace coming from the use of nonzero() followed by item() in the scheduler's `index_for_timestep()` logic: https://github.com/huggingface/diffusers/blob/b0f7036d9af75c5df0f39d2d6353964e4c520534/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L351-L363

This sync causes a minor but non-negligible gap in GPU utilization during the first timestep, especially when torch.compile is utilized (due to cpu-side Dynamo cache lookup overhead after the sync point), as shown here:

trace_with_nonzero2

AFAICT this can be avoided by manually calling `scheduler.set_begin_index(0)`, so this PR does that for `FluxPipeline`. Locally, I was able to see the sync point go away after this change.

Insights welcome regarding:
* A more robust alternative fix
* The best way to test this change (looks like [test_pipeline_flux.py](https://github.com/huggingface/diffusers/blob/main/tests/pipelines/flux/test_pipeline_flux.py) is a good place?)
* Should this fix be generalized beyond `FluxPipeline`?

### DIFF
--- a/src/diffusers/pipelines/flux/pipeline_flux.py
+++ b/src/diffusers/pipelines/flux/pipeline_flux.py
@@ -898,6 +898,7 @@ class FluxPipeline(
             )
 
         # 6. Denoising loop
+        self.scheduler.set_begin_index(0)
         with self.progress_bar(total=num_inference_steps) as progress_bar:
             for i, t in enumerate(timesteps):
                 if self.interrupt:


yiyixuxu:
@a-r-r-o-w 

custom timesteps uses `set_begin_index` too https://github.com/huggingface/diffusers/blob/f3e09114f200f8fbf88b060e218ac5a50464fe8d/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl_img2img.py#L659

yiyixuxu:
yes agree we should apply this change to all text2image pipelines, test-wise, as long as the current test pass we are fine, 
unless we want to add additional torch compile related test

I think this fix is sufficient for all pipelines though, if not, we can look at them case by case

a-r-r-o-w:
@yiyixuxu I think there are many instances where we don't do it the same way as done in SDXL. These will incur the extra cost from `nonzero()`.

- https://github.com/huggingface/diffusers/blob/f3e09114f200f8fbf88b060e218ac5a50464fe8d/src/diffusers/pipelines/wan/pipeline_wan_video2video.py#L446

yiyixuxu:
@a-r-r-o-w yeah, i know, we can fix them though. I think unless you need to look up on every step, it should work by setting a index in the beginning

yiyixuxu:
@jbschlosser @a-r-r-o-w 
I'm merging this now as I understand this is needed pretty urgently 
we can do a follow-up PR for the rest of pipelines :)

sayakpaul:
@yiyixuxu thank you very much for the treatment of this PR. Appreciate that.

PR #11697, opened 2025-06-11
TorchAO compile + offloading tests

a-r-r-o-w:


tests/quantization/test_torch_compile_utils.py
```diff
@@ -64,7 +64,29 @@ def _test_torch_compile_with_cpu_offload(self, quantization_config, torch_dtype=
             # small resolutions to ensure speedy execution.
             pipe("a dog", num_inference_steps=3, max_sequence_length=16, height=256, width=256)
 
-    def _test_torch_compile_with_group_offload(self, quantization_config, torch_dtype=torch.bfloat16):
+    def _test_torch_compile_with_group_offload_leaf(self, quantization_config, torch_dtype=torch.bfloat16):
```
sayakpaul: Maybe we can test with `parameterized` where we test with and without streams?

tests/quantization/torchao/test_torchao.py
```diff
@@ -625,6 +628,48 @@ def test_int_a16w8_cpu(self):
...
+    )
+
+    def test_torch_compile(self):
+        super()._test_torch_compile(quantization_config=self.quantization_config)
+
+    @unittest.skip(
+        "Changing the device of AQT tensor with module._apply (called from doing module.to() in accelerate) does not work "
+        "when compiling."
+    )
+    def test_torch_compile_with_cpu_offload(self):
```
sayakpaul, line 649: Also FYI: https://huggingface.slack.com/archives/C068ZAHJZCZ/p1748618381438319

sayakpaul (approved):
Thank you! Appreciate the detailed comments.

tests/quantization/torchao/test_torchao.py
a-r-r-o-w: Thanks, makes sense!

sayakpaul (approved):
Thanks!

a-r-r-o-w:
@sayakpaul I'm not sure what's causing the tests to fail 🤔 This PR guards the compile test with torchao version/installation requirement but still seemingly causes tests to fail. I'll try to take a look later today if we don't have a quick understanding of what happened here

sayakpaul:
Exactly! Nothing comes to mind as to what could trigger this!

tests/quantization/test_torch_compile_utils.py
```diff
@@ -64,16 +64,17 @@ def _test_torch_compile_with_cpu_offload(self, quantization_config, torch_dtype=
...
+    ):
         torch._dynamo.config.cache_size_limit = 10000
 
         pipe = self._init_pipeline(quantization_config, torch_dtype)
         group_offload_kwargs = {
             "onload_device": torch.device("cuda"),
             "offload_device": torch.device("cpu"),
             "offload_type": "leaf_level",
-            "use_stream": True,
-            "non_blocking": True,
```
sayakpaul, line 76: Should keep the `non_blocking=True` or make it an argument of the function like `use_stream`?

sayakpaul (commented):
Was able to spend some time and the following diff solves the problem:

<details>
<summary>

```diff
diff --git a/tests/quantization/torchao/test_torchao.py b/tests/quantization/torchao/test_torchao.py
index ddf97aca5..28454aae9 100644
--- a/tests/quantization/torchao/test_torchao.py
+++ b/tests/quantization/torchao/test_torchao.py
@@ -631,11 +631,14 @@ class TorchAoSerializationTest(unittest.TestCase):
 
 @require_torchao_version_greater_or_equal("0.7.0")
 class TorchAoCompileTest(QuantCompileTests):
-    quantization_config = PipelineQuantizationConfig(
-        quant_mapping={
-            "transformer": TorchAoConfig(quant_type="int8_weight_only"),
-        },
-    )
+    @property
+    def quantization_config(self):
+        config = PipelineQuantizationConfig(
+            quant_mapping={
+                "transformer": TorchAoConfig(quant_type="int8_weight_only"),
+            },
+        )
+        return config
 
     def test_torch_compile(self):
         super()._test_torch_compile(quantization_config=self.quantization_config)
```

</details>

ChatGPT does a nice job of explaining what is happening:
https://chatgpt.com/share/685951bc-7c88-8013-b317-62683d1a1fa9. What I didn't investigate is that how come they other TorchAO tests are not getting flagged because of torchao installation errors 🤷

tests/quantization/test_torch_compile_utils.py
a-r-r-o-w: `non_blocking=True` does not need to be specified when using streams since it cannot be `False` when `use_stream=True` (see [init](https://github.com/huggingface/diffusers/blob/e8e44a510c152fff17e3f1bba036d635776b5b9f/src/diffusers/hooks/group_offloading.py#L73)). I think there's not much to gain from testing non_blocking because it only provides very limited overlap.

a-r-r-o-w:
@sayakpaul Thanks for looking into it! I've converted all occurrences to properties (include bitsandbytes ones)

tests/quantization/test_torch_compile_utils.py
```diff
@@ -24,7 +24,11 @@
 @require_torch_gpu
 @slow
 class QuantCompileTests(unittest.TestCase):
-    quantization_config = None
+    @property
+    def quantization_config(self):
+        raise NotImplementedError(
+            "This property should be implemented in the subclass to return the appropriate quantization config."
+        )
```
sayakpaul, line 31: 👌

PR #11701, opened 2025-06-12
swap out token for style bot.

sayakpaul:


ydshieh:
You can also consider apply

https://github.com/huggingface/transformers/commit/7c5833694934c289e4297f50417d422a893c9d5d

(to reduce the unnecessary permission for the `GITHUB_TOKEN`)

PR #11705, opened 2025-06-13
[rfc][compile] compile method for DiffusionPipeline

anijain2305:
This PR adds a compile method to the DiffusionPipeline to make it really easy for diffusers users to apply torch.compile.

There are a couple of reasons why we want a deeper integration of compile with diffusion models
* Regional compile (instead of full model compile) is  better fit for diffusion models, and its not easy for the user to apply regional compile precisely.
*  A pipeline has many components, and many of them are not suitable for compile. Since we (or a new model author) know the architecture of the models, we can leverage it to apply compile at the most relevant portions of the pipeline.

-----------------

# Update - Jun 20

We went with a new Option 4 - adding a new method to `ModelMixin` - `compile_repeated_blocks` to avoid any confusion and also get a sense from the community on its adoption. If we see positive feedback, we can move with a deeper integration like Option 1. Currently, there is no data to go for an aggressive approach like Option 1.

-----------------

# Option 1 - Pipeline has a compile method
From the user standpoint, they will have to add this line
```
pipe.compile(**compile_kwargs)
``` 

This will by default find the regions in the `transformer` model (more on how to find regions later) and apply regional compile to those regions.  We can use this to provide more control to the user - if they want to apply compile to pre and pos-transformer model. 

##  Pro
* Easiest for the user

## Cons
* Might be considered a big change, and we might want to phase this out.

# Option 2 - ModelMixin overrides nn.Module compile method (this PR does Option 2)

User does something like this

```
pipe.transformer.compile(use_regional_compile=True/False, **kwargs)
```

This overrides the `nn.Module` compile and provides options for regional compile. This is one level deeper integration, so user will have to know that the pipe must have a `transformer` named attribute. If `use_regional_compile` is True, we will find the regions (more on that later), and apply torch.compile on them.

## Pros
* ModelMixin is the most time consuming part of the model, so the abstraction seems right.

## Cons
* Users might already be using `pipe.transformer.compile` to enable full model compilation. I DO want regional compile as the default (supported by numbers later).  But when users upgrade to new diffusers, they will see regional compile if we make it a default, and might be confused. Turning off regional compile by defaults solves this problem, but it is sub-par UX because of high compile time w/o any real benefit.
* Though minor, its less obvious than `pipe.compile` 

# Option 3 - Combine option 1 and option 2
This could be the best option. We have both option 1 and 2. In option 2, we keep the default regional compile OFF. But option 1 - pipe.compile has regional compile by default, and propagates the regional compile to `pipe.tranformer.compile`. This way, the OOTB solution would be `pipe.compile` and then any more enthusiastic user can play with `pipe.transformer.compile` if they wish.

# How to find regions?
Here, we can make a simple addition to the existing models and provide guidance for the new models. Most of the diffusion models are basically a sequence of Transformer blocks, and these blocks are  obvious from the model `__init__` method. We can add a new class attribute for each model - `_regions_for_compile` - which point to the classes of those Transformer blocks. This is very precise and easy. Relatedly, I also find another field - `_no_split_modules` - which incidentally serves the same purpose. In the absence of the `_regions_for_compile` attribute, we can rely on `_no_split_modules`.
 
# Why regional compile as default?

Regional compilation gives as good speedup as full model compilation for diffusion models with **8x-10x smaller compilation time**. More data to support regional compilation as default is below

image

One experiment that I am missing is `cudagraphs`. I am not sure yet if enabling cudagraphs makes full model better than regional. I will rerun the experiments today with cudagraphs.

cc @sayakpaul


src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -286,6 +286,8 @@ def __init__(
 
         self.gradient_checkpointing = False
 
+        self.compile_region_classes = (FluxTransformerBlock, FluxSingleTransformerBlock)
```
sayakpaul: Should this be a class-level attribute like this?
https://github.com/huggingface/diffusers/blob/62cbde8d41ac39e4b3a1f5bbbbc546cc93f1d84d/src/diffusers/models/modeling_utils.py#L262

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -2027,6 +2027,40 @@ def _maybe_raise_error_if_group_offload_active(
                 return True
         return False
 
+    def compile(
```
sayakpaul: We could make it a part of `ModelMixin` rather than `DiffusionPipeline` I believe:
https://github.com/huggingface/diffusers/blob/62cbde8d41ac39e4b3a1f5bbbbc546cc93f1d84d/src/diffusers/models/modeling_utils.py#L251

For most cases, users want to just do `pipe.transformer.compile()`. So, perhaps easier with this being added to `ModelMixin`?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -2027,6 +2027,40 @@ def _maybe_raise_error_if_group_offload_active(
...
+        self,
+        compile_regions_for_transformer: bool = True,
+        transformer_module_name: str = "transformer",
```
sayakpaul: These two will have to have changed if we proceed to have it inside `ModelMixin`.

src/diffusers/pipelines/pipeline_utils.py
anijain2305: Do you expect any improvements from compile on the pre/post transformer blocks? If not, then yeah, moving it inside makes sense to me.

src/diffusers/models/transformers/transformer_flux.py
anijain2305: Yes, thats better.

src/diffusers/pipelines/pipeline_utils.py
anijain2305: With this change, do we have to do `pipe.transformer.compile()`? Hmm, this could be confusing because torch.nn.Module also has a method named `compile`. If we move in the `ModelMixin`, we might want a different method name.

sayakpaul: Well, to tackle those cases, we could do something like this:
https://github.com/huggingface/diffusers/blob/62cbde8d41ac39e4b3a1f5bbbbc546cc93f1d84d/src/diffusers/models/modeling_utils.py#L1306

So, we include all the acceptable args and kwargs in the `ModelMixin` compile()` method but additionally include the regional/hierarchical compilation related kwargs. Would that work?

anijain2305: Nice! Let me try this out.

StrongerXi: I feel like there are some advantages of `Pipeline.compile`:
1. It captures the user intent better, e.g., "I just want to accelerate this pipeline" (the word "compile" is also great because it signals "expect some initial delay").
2. It's more future proof (diffuser devs get to tweak things to accommodate changes from diffusers/pytorch versions, or even future hardware and pipeline architecture/perf-characteristics).
3. It's more user friendly (small things like `.transformer` does make a noticeable effect on user experience).

(2) and (3) are effectively consequences of (1). What do you think? @sayakpaul

anijain2305: Sayak suggested to phase this out. 

Have `pipeline.transformer.compile` as the first PR

And then `pipeline.compile` in future which can internally call `pipeline.transformer.compile`. Its actually better this way because then the compile region classes is also hidden from the pipe.

sayakpaul: Yes, we will eventually ship `pipe.compile()` :)

yiyixuxu:
thanks for the PR !
How do we determine default `compile_region_classes` for each model?

anijain2305:
> thanks for the PR ! How do we determine default `compile_region_classes` for each model?

This is something that we will have to do manually. I deliberately made that choice. It should be be fairly easy to find those classes, basically the transformer classes. In absence of compile_region_classes, we can fallback to the full model compilation with a warning.

I think this is a fair compromise because it gives control to the model author, and with little guidance it should be very easy.

yiyixuxu:
>It should be be fairly easy to find those classes, basically the transformer classes

is it because that they are repeated? 

can we add a `enable_regional_compile()` on `ModelMixin` since this is the main purpose of that?

sayakpaul:
> can we add a enable_regional_compile() on ModelMixin since this is the main purpose of that?

@yiyixuxu this is a good suggestion. But I wanted to discuss the developer experience a bit. 

Our initial (@anijain2305 feel free to correct me) plan was to override `compile()` at `ModelMixin`. This overriden `compile()` method will have all the `args` and `kwargs` taken by the original `compile()` method of `torch` (note that any `nn.Module` now [comes with](https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.compile) `compile()`). But it will additionally take an argument to enable regional compilation. 

So, it would be something like:

```py
model = FluxTransformer2DModel()
`regional_compilation` is just an arg, name can be changed/decided later.
model.compile(regional_compilation=True, ...)
```

With `enable_regional_compile()`, would the developer experience look like?

```py
model = FluxTransformer2DModel()
model.enable_regional_compile()
model.compile(...)
```

anijain2305:
@sayakpaul @yiyixuxu I updated the PR and also the original description with some discussion points (pros and cons). Let me know what you think.

src/diffusers/models/modeling_utils.py
```diff
@@ -1402,6 +1403,44 @@ def float(self, *args):
         else:
             return super().float(*args)
 
+    @wraps(torch.nn.Module.compile)
+    def compile(self, use_regional_compile: bool = True, *args, **kwargs):
```
anijain2305: If we go with Option 3 - we should turn this to `False` to keep the existing behavior of `pipe.tranformer.compile` same as before.

StrongerXi:
> But when users upgrade to new diffusers, they will see regional compile if we make it a default, and might be confused.

I thought the only thing the users will _see_ here is less compilation time, if they already have `pipeline.transformer.compile(...)`?

yiyixuxu:
i see, maybe enalbe_ is indeed a bit confusing, maybe just `model.reginal_compile()` then?

yiyixuxu:
one thing I want to understand is this this pattern (regional compile) only beneficial to diffusers models?  
should  and would this be supported in pytorch `compile` in the future (or maybe now)?

anijain2305:
> one thing I want to understand is this this pattern (regional compile) only beneficial to diffusers models? should and would this be supported in pytorch `compile` in the future (or maybe now)?

Regional compile is *very* useful for diffusion models because the transformer blocks are big enough that the compile optimization inside the transformer block is enough to give major speedup, and there is very little to be had with compile optimizations across the transformer blocks. 

The regional compile is not a feature per se, its more like saying - "compile responsibly" - there is no specific support required  from the pytorch side to enable this feature. For `torch.compile`, its just another `nn.Module` (or function) that you are compiling. As a user, we are compiling just small repeated regions to keep the compilation cost low.

For LLMs or other non-diffusion models, it is possible that you do want cross-transformer layer optimizations to see compiler benefits (maybe the presence of static kv cache changes the space of optimizations).

For diffusion models, we can do more testing to convince ourselves. One thing that I do want to test is cudagraphs. 

> I thought the only thing the users will _see_ here is less compilation time, if they already have `pipeline.transformer.compile(...)`?

@StrongerXi It will be true in almost all cases, but if a user specifically wants to compile the FULL model for some reason, they will be confused.

anijain2305:
> i see, maybe enalbe_ is indeed a bit confusing, maybe just `model.reginal_compile()` then?

If we go with Option 3 (which is Option 1 + option 2), we might not need `model.regional_compile`.  In this option 3, the ModelMixin will have a `compile` method which takes the additional regional_compile args. These regional compile will all be defaulted to None by default. But the Pipeline `compile` method will use regional by default and setup the regional kwargs whlie calling pipeline.transformer.compile. Let me know if you need to see the code.

yiyixuxu:
>The regional compile is not a feature per se, its more like saying - "compile responsibly" - there is no specific support required from the pytorch side to enable this feature. For torch.compile, its just another nn.Module (or function) that you are compiling. As a user, we are compiling just small repeated regions to keep the compilation cost low.

thanks for the explanation! As I understand, this is something works great in diffusers but not unique to it. i.e. it can potentially benefit many non-diffusers models. I think it would not be a good practice for different libraries to wrap the `compile` API in different ways so I would like to upstream it if it's possible

if it is not possible or does not make sense to add in pytorch, I would prefer to very quickly support it with a separate method instead of wrapping around the `compile` API - it is one less layer and less complex, also easier user to discover this and use

yiyixuxu:
regards to `pipeline.compile()`, I agree with @sayakpaul I think we should focus on model level first and eventually come to `pipeline.compile()` if it is meaningful

anijain2305:
> if it is not possible or does not make sense to add in pytorch, I would prefer to very quickly support it with a separate method instead of wrapping around the `compile` API - it is one less layer and less complex, also easier user to discover this and use

I hear you. Does 'pipe.compile' API offer a good solution? We are not overriding anything because it's a new API and basically decides the best way torch.compile should be used for the diffusion models. Or the 'compile' name seems overloaded.

Cc @sayakpaul as well to hear his thoughts.

anijain2305:
@yiyixuxu I hear you. There  is not really a way to upstream this, as this is completely model dependent. In practice, library/model authors are the best folks to determine where to apply torch.compile (in diffusion models, it turns out to be simple).

I am actually ok with anything that makes it easy to give the best ootb `torch.compile` experience to the end users. With that in mind, I am ok with `regional_compile` API in `ModelMixin`, supplemented with good documentation.

Cc @sayakpaul to hear his thoughts.

sayakpaul:
@anijain2305 let's go with https://github.com/huggingface/diffusers/pull/11705#issuecomment-2978649474? Thanks for being willing to reiterate. I will help with tests and documentation as well to see this through with you.

I think Yiyi's observations are quite aligned with what we see in the community of our users.

I would also be keen on seeing the results with CUDAgraphs now that @a-r-r-o-w made a nice observation by using it (Slack).

DN6:
I'm with @yiyixuxu on the dedicated method for the initial version. If we go with `ModelMixin.compile` with the regional args turned off by default, then we're effectively the same as the torch native compile. If we enable regional compile by default, then it might be confusing for users working with non-DiT models, e.g. Would `pipe.unet.compile` see similar benefits?

We definitely want this feature @anijain2305 (thank you taking the initiative here 💪🏽 ), and since the end user API is still a bit fuzzy, having a dedicated method feeds like the way to go IMO. We can always refactor later based on adoption/feedback from the community.

anijain2305:
I have not forgotten about this. Just been swamped by the impending branch cut for PyTorch 2.8 release.

I agree with all of you, and I will add  a new method in `ModelMixin`, that does regional compile but does not override `compile` method.

anijain2305:
@yiyixuxu @sayakpaul @DN6 This is ready for next round of review.

src/diffusers/models/modeling_utils.py
```diff
@@ -1404,6 +1405,54 @@ def float(self, *args):
...
+        can reduce end-to-end compile time substantially, while preserving the
+        runtime speed-ups you would expect from a full `torch.compile`.
+
+        The set of sub-modules to compile is discovered in one of two ways:
+
+        1. **`_repeated_blocks`** – Preferred.  Define this attribute on your
+        subclass as a list/tuple of class names (strings).  Every module whose
+        class name matches will be compiled.
+
+        2. **`_no_split_modules`** – Fallback.  If the preferred attribute is
```
sayakpaul: `_no_split_modules` is for specifying the modules that have residual connections so that we don't split them when using `device_map` on multiple devices. I don't think it's legacy yet.

src/diffusers/models/modeling_utils.py
```diff
@@ -1404,6 +1405,54 @@ def float(self, *args):
...
+        missing or empty, we fall back to the legacy Diffusers attribute
+        `_no_split_modules`.
+
+        Once discovered, each matching sub-module is compiled by calling
+        ``submodule.compile(*args, **kwargs)``.  Any positional or keyword
```
sayakpaul:
```suggestion
        `submodule.compile(*args, **kwargs)`.  Any positional or keyword
```

src/diffusers/models/modeling_utils.py
```diff
@@ -1404,6 +1405,54 @@ def float(self, *args):
...
+        arguments you supply to :meth:`compile_repeated_blocks` are forwarded
```
sayakpaul:
```suggestion
        arguments you supply to `compile_repeated_blocks` are forwarded
```

src/diffusers/models/modeling_utils.py
```diff
@@ -1404,6 +1405,54 @@ def float(self, *args):
...
+        verbatim to `torch.compile`.
+        """
+        repeated_blocks = getattr(self, "_repeated_blocks", None)
+
+        if not repeated_blocks:
+            logger.warning("_repeated_blocks attribute is empty. Using _no_split_modules to find compile regions.")
```
sayakpaul:
```suggestion
            logger.warning("`_repeated_blocks` attribute is empty. Using `_no_split_modules` to find compile regions.")
```

src/diffusers/models/modeling_utils.py
```diff
@@ -1404,6 +1405,54 @@ def float(self, *args):
...
+
+            repeated_blocks = getattr(self, "_no_split_modules", None)
+
+        if not repeated_blocks:
+            raise ValueError(
+                "Both _repeated_blocks and _no_split_modules attribute are empty. "
+                "Set _repeated_blocks for the model to benefit from faster compilation. "
+            )
```
sayakpaul:
```suggestion
            raise ValueError(
                "Both `_repeated_blocks` and `_no_split_modules` attribute are empty. "
                "Set `_repeated_blocks` for the model to benefit from faster compilation. "
            )
```

src/diffusers/models/modeling_utils.py
```diff
@@ -1404,6 +1405,54 @@ def float(self, *args):
...
+
+        has_compiled_region = False
+        for submod in self.modules():
+            if submod.__class__.__name__ in repeated_blocks:
+                has_compiled_region = True
```
sayakpaul: (nit): should it not be _after_ `submod.compile()`?

sayakpaul (commented):
This is off to a really good start. I don't have any major comments to share here. 

For the TODOs, we have docs to add as well as some tests. For docs, we could add a section in https://huggingface.co/docs/diffusers/en/optimization/torch2.0#torchcompile. For tests, we could add them [here](https://github.com/huggingface/diffusers/blob/ee40088fe5437f8ed65ec96a22250149e4f334cc/tests/models/test_modeling_common.py#L1905).

anijain2305:
> This is off to a really good start. I don't have any major comments to share here.
> 
> For the TODOs, we have docs to add as well as some tests. For docs, we could add a section in [huggingface.co/docs/diffusers/en/optimization/torch2.0#torchcompile](https://huggingface.co/docs/diffusers/en/optimization/torch2.0#torchcompile). For tests, we could add them [here](https://github.com/huggingface/diffusers/blob/ee40088fe5437f8ed65ec96a22250149e4f334cc/tests/models/test_modeling_common.py#L1905).

Will add a few tests tomorrow. I will wait for @yiyixuxu or @DN6 initial review before writing the docs.

tests/models/test_modeling_common.py
```diff
@@ -1931,6 +1931,24 @@ def test_torch_compile_recompilation_and_graph_break(self):
...
 
+    def test_torch_compile_repeated_blocks(self):
+        init_dict, inputs_dict = self.prepare_init_args_and_inputs_for_common()
+
+        model = self.model_class(**init_dict).to(torch_device)
+        model.compile_repeated_blocks(fullgraph=True)
+
+        recompile_limit = 1
+        if self.model_class.__name__ == "UNet2DConditionModel":
+            recompile_limit = 2
```
anijain2305, line 1949: @sayakpaul For UNet, the shape changes between different instances of the transformer model, so we need 2 entries.

anijain2305:
The tests are in.

src/diffusers/models/modeling_utils.py
anijain2305: Removed the word legacy. Please re-read

yiyixuxu: I don't think we need fallback/proxity here
we can make sure every model we have this attribute specified :)

src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -227,6 +227,7 @@ class FluxTransformer2DModel(
     _supports_gradient_checkpointing = True
     _no_split_modules = ["FluxTransformerBlock", "FluxSingleTransformerBlock"]
     _skip_layerwise_casting_patterns = ["pos_embed", "norm"]
+    _repeated_blocks = _no_split_modules
```
yiyixuxu: can we just type the layer names out? when someone implement a new model based on flux, they may assume this two attribute are same here, (but they just happen to be the same in this case)

yiyixuxu (commented):
thanks @anijain2305 !
looking great! let's get this merged this soon:) 

I left some small comments. Mainly I think we don't not need a fallback at all and just throw an error when `_repeated_blocks` is empty 

this simplifies the code and would encourage the user and send a PR our way to set the attribute when it's not set

let's get this merged this soon:)

src/diffusers/models/modeling_utils.py
```diff
@@ -1404,6 +1405,43 @@ def float(self, *args):
...
+    def compile_repeated_blocks(self, *args, **kwargs):
+        """
+        Compiles *only* the frequently repeated sub-modules of a model (e.g. the
+        Transformer layers) instead of compiling the entire model.  This
+        technique—often called **regional compilation** (see the PyTorch recipe
+        https://docs.pytorch.org/tutorials/recipes/regional_compilation.html)
+        can reduce end-to-end compile time substantially, while preserving the
+        runtime speed-ups you would expect from a full `torch.compile`.
+
+        The set of sub-modules to compile is discovered by the presence of
```
sayakpaul: We could also include the doc link that we're gonna add in this PR.

tests/models/test_modeling_common.py
```diff
@@ -1936,6 +1936,24 @@ def test_torch_compile_recompilation_and_graph_break(self):
...
+
+        with (
+            torch._inductor.utils.fresh_inductor_cache(),
+            torch._dynamo.config.patch(recompile_limit=recompile_limit),
```
sayakpaul, line 1953: Neat!

sayakpaul (approved):
Excited to see this shipped!

Let's add the docs as we have great comments from @yiyixuxu as well!

sayakpaul:
@bot /style

docs/source/en/optimization/fp16.md
```diff
@@ -152,9 +152,39 @@ Compilation is slow the first time, but once compiled, it is significantly faste
...
 
-[Accelerate](https://huggingface.co/docs/accelerate/index) provides the [compile_regions](https://github.com/huggingface/accelerate/blob/273799c85d849a1954a4f2e65767216eb37fa089/src/accelerate/utils/other.py#L78) method for automatically compiling the repeated blocks of a `nn.Module` sequentially. The rest of the model is compiled separately.
+[Regional compilation](https://docs.pytorch.org/tutorials/recipes/regional_compilation.html) trims cold-start latency by compiling **only the small, frequently-repeated block(s)** of a model, typically a Transformer layer, enabling reuse of compiled artifacts for every subsequent occurrence.
+For many diffusion architectures this delivers the *same* runtime speed-ups as full-graph compilation yet cuts compile time by **8–10 ×**.
+
+To make this effortless, `ModelMixin` exposes **`compile_repeated_blocks`** API, a helper that wraps `torch.compile` around any sub-modules you designate as repeatable:
```
sayakpaul:
```suggestion
To make this effortless, [`ModelMixin`] exposes [`ModelMixin.compile_repeated_blocks`] API, a helper that wraps `torch.compile` around any sub-modules you designate as repeatable:
```

anijain2305:
@yiyixuxu Just a gentle ping for another review.

yiyixuxu (approved):
thank you for adding this feature!

sayakpaul:
Will merge once this round of CI is through.

sayakpaul:
The failure is probably due to https://github.com/huggingface/diffusers/pull/11759#issuecomment-3006805190. So, will merge accordingly.

sayakpaul:
@anijain2305 huge thanks for this feature!

PR #11707, opened 2025-06-13
[training] show how metadata stuff should be incorporated in training scripts.

sayakpaul:
# What does this PR do?

Since https://github.com/huggingface/diffusers/pull/11324 is merged now, we can safely inject metadata into safetensor files when training LoRAs.


sayakpaul:
Good point. I believe that could be documented in the READMEs.

PR #11708, opened 2025-06-13
[tests] device_map tests for all models.

sayakpaul:
# What does this PR do?

Move out of just UNets.


tests/models/test_modeling_common.py
```diff
@@ -1736,6 +1736,44 @@ def test_auto_model(self, expected_max_diff=5e-5):
...
+        init_dict, _ = self.prepare_init_args_and_inputs_for_common()
+        model = self.model_class(**init_dict)
+        with tempfile.TemporaryDirectory() as tmpdir:
+            model.save_pretrained(tmpdir)
+            with self.assertRaises(ValueError) as err_ctx:
+                _ = self.model_class.from_pretrained(tmpdir, device_map=device_map)
+
+        assert msg_substring in str(err_ctx.exception)
+
+    @parameterized.expand([0, "cuda", torch.device("cuda"), torch.device("cuda:0")])
```
a-r-r-o-w:
```suggestion
    @parameterized.expand([0, "cuda", torch.device("cuda")])
```

Since `torch.device("cuda")` and this are the same, I think nice to reduce 1 extra test run per model

tests/models/test_modeling_common.py
```diff
@@ -1736,6 +1736,44 @@ def test_auto_model(self, expected_max_diff=5e-5):
...
+    @require_torch_gpu
+    def test_passing_dict_device_map_works(self, name, device_map):
+        # There are other valid dict-based `device_map` values too. It's best to refer to
+        # the docs for those: https://huggingface.co/docs/accelerate/en/concept_guides/big_model_inference#the-devicemap.
+        init_dict, inputs_dict = self.prepare_init_args_and_inputs_for_common()
+        model = self.model_class(**init_dict).eval()
+        with tempfile.TemporaryDirectory() as tmpdir:
+            model.save_pretrained(tmpdir)
+            loaded_model = self.model_class.from_pretrained(tmpdir, device_map=device_map)
+            _ = loaded_model(**inputs_dict)
```
a-r-r-o-w: Here, we are not really creating the dict, no?

sayakpaul: Good catch! Should be fixed now.

PR #11711, opened 2025-06-13
[docs] device_map

stevhliu:
Improves `device_map` docs with more explicit example of how to design your own.


stevhliu:
Merging now and we can iterate on it later if we need to :)

Birch-san:
thanks, I still don't understand why the key `''` works though. is there some kind of pattern-matching? I thought there was some convention where it does a prefix match or something…

PR #11712, opened 2025-06-14
[docs] quant_kwargs

stevhliu:
Improves where you can find the `quant_kwargs` in the `PipelineQuantizationConfig` docs

### DIFF
--- a/docs/source/en/quantization/overview.md
+++ b/docs/source/en/quantization/overview.md
@@ -11,7 +11,7 @@ specific language governing permissions and limitations under the License.
 
 -->
 
-# Quantization
+# Getting started
 
 Quantization focuses on representing data with fewer bits while also trying to preserve the precision of the original data. This often means converting a data type to represent the same information with fewer bits. For example, if your model weights are stored as 32-bit floating points and they're quantized to 16-bit floating points, this halves the model size which makes it easier to store and reduces memory usage. Lower precision can also speedup inference because it takes less time to perform calculations with fewer bits.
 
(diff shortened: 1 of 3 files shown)


sayakpaul (approved):
Thanks! Do we have to include an explicit point on how users can derive `qaunt_kwargs` for a given backend?

stevhliu:
I think it's probably best to refer the user to the specific API docs for each backend where they can find all the relevant arguments, and keep this one more high-level.

Not sure why the doc-builder is failing though. bitsandbytes should already be installed in the CI test environment no? 🤔

sayakpaul:
> I think it's probably best to refer the user to the specific API docs for each backend where they can find all the relevant arguments, and keep this one more high-level.

Yeah indeed. I was asking for an explicit note that says just that.

sayakpaul:
> Not sure why the doc-builder is failing though. bitsandbytes should already be installed in the CI test environment no? 🤔

No actually. `bitsandbytes` is not a required dependency in our CI and we install it only when running the quantization tests.

I opened https://github.com/huggingface/diffusers/pull/11728 for this. Let's see.

docs/source/en/quantization/overview.md
```diff
@@ -11,27 +11,33 @@ specific language governing permissions and limitations under the License.
...
-There are two ways you can use [`~quantizers.PipelineQuantizationConfig`] depending on the level of control you want over the quantization specifications of each model in the pipeline.
+There are two ways to use [`~quantizers.PipelineQuantizationConfig`] depending on how much customization you want to apply to the quantization configuration. 
 
-- for more basic and simple use cases, you only need to define the `quant_backend`, `quant_kwargs`, and `components_to_quantize`
-- for more granular quantization control, provide a `quant_mapping` that provides the quantization specifications for the individual model components
+- for basic use cases, define the `quant_backend`, `quant_kwargs`, and `components_to_quantize` arguments
+- for granular quantization control, define a `quant_mapping` that provides the quantization configuration for individual model components
 
-### Simple quantization
+### Basic quantization
```
asomoza, line 27: in line 94 there's a link to 'simple quantization' so this will break it, change the other link then?

asomoza: saw this too late I think

stevhliu: Oops, should be fixed in #12018!

PR #11717, opened 2025-06-15
Fix dimensionalities in `apply_rotary_emb` functions' comments

tolgacangoz:
While working on MAGI-1, I came across these. I thought that understandability is important, so I propose fixing tensor shapes in the comments.

@yiyixuxu @a-r-r-o-w @ylacombe


tolgacangoz:
I thought I could accumulate the things I propose to modify in this PR during MAGI-1, etc.

yiyixuxu (approved):
thanks!

tolgacangoz:
Thanks for merging!

PR #11719, opened 2025-06-16
[LoRA] refactor lora loading at the model-level

sayakpaul:
# What does this PR do?

Currently, there's a significant overlap between these two methods:

https://github.com/huggingface/diffusers/blob/8adc6003ba4dbf5b61bb4f1ce571e9e55e145a99/src/diffusers/loaders/peft.py#L121

https://github.com/huggingface/diffusers/blob/8adc6003ba4dbf5b61bb4f1ce571e9e55e145a99/src/diffusers/loaders/lora_base.py#L322

This makes it challenging to incorporate new changes as the developer has to remember to propagate the changes in both locations. Hence, this PR factors out the common things from both of the places and turns them into utility functions. IMO, it also helps to make both the loading functions a bit cleaner to read.


src/diffusers/loaders/lora_base.py
```diff
@@ -377,89 +373,41 @@ def _load_lora_into_text_encoder(
...
-            for module in ("fc1", "fc2"):
-                rank_key = f"{name}.{module}.lora_B.weight"
-                if rank_key not in state_dict:
-                    continue
-                rank[rank_key] = state_dict[rank_key].shape[1]
+        for name, _ in text_encoder.named_modules():
+            if name.endswith((".q_proj", ".k_proj", ".v_proj", ".out_proj", ".fc1", ".fc2")):
+                rank_key = f"{name}.lora_B.weight"
+                if rank_key in state_dict:
+                    rank[rank_key] = state_dict[rank_key].shape[1]
```
BenjaminBossan, line 380: Not specific to this PR, but would it be better to rename `rank` to `rank_pattern`?

src/diffusers/loaders/lora_base.py
```diff
@@ -377,89 +373,41 @@ def _load_lora_into_text_encoder(
...
-        text_encoder.load_adapter(
-            adapter_name=adapter_name,
-            adapter_state_dict=state_dict,
-            peft_config=lora_config,
-            **peft_kwargs,
-        )
-
-        # scale LoRA layers with `lora_scale`
-        scale_lora_layers(text_encoder, weight=lora_scale)
+        # <Unsafe code
```
BenjaminBossan, line 393: Why is it unsafe?

src/diffusers/loaders/lora_base.py
```diff
@@ -377,89 +373,41 @@ def _load_lora_into_text_encoder(
...
+        with _lora_loading_context(_pipeline):
```
BenjaminBossan: I'm not sure if a context manager is helping here, what was wrong with https://github.com/huggingface/diffusers/pull/11719/files#diff-005343920b6fa6c3f4c1f639ee23f180dfa424d6008e487fdc9962a82f498d15L448-L452?

src/diffusers/loaders/lora_base.py
sayakpaul: The enable/disable offloading stuff.

src/diffusers/loaders/lora_base.py
sayakpaul: Since the argument name is `rank` for `get_peft_kwargs` and since it's a public method, I think it's okay as is.

src/diffusers/loaders/lora_base.py
sayakpaul: Yeah I was thinking about it, too. I thought it could help reduce the duplication. But maybe being explicit about the enabling/disabling of the offloading hooks (like previous) is better?

BenjaminBossan: I think for this specific part of the refactor, I prefer the explicit call, as the context manager doesn't really do much here.

BenjaminBossan (approved):
Thanks, LGTM.

src/diffusers/utils/peft_utils.py
```diff
@@ -288,3 +292,112 @@ def check_peft_version(min_version: str) -> None:
...
+        logger.warning(
+            f"No LoRA keys associated to {model_class_name} found with the {prefix=}. "
+            "This is safe to ignore if LoRA state dict didn't originally have any "
+            f"{model_class_name} related params. You can also try specifying `prefix=None` "
+            "to resolve the warning. Otherwise, open an issue if you think it's unexpected: "
+            "https://github.com/huggingface/diffusers/issues/new"
+        )
+
+
+def _maybe_raise_error_for_ambiguity(config):
```
DN6: Nit. Would prefer to call this something like `_maybe_raise_error_for_ambiguous_keys`. More explicit about we are checking for in the passed config.

src/diffusers/utils/peft_utils.py
```diff
@@ -288,3 +292,112 @@ def check_peft_version(min_version: str) -> None:
...
+    try:
+        yield
+    finally:
+        if is_model_cpu_offload:
+            _pipeline.enable_model_cpu_offload()
+        elif is_sequential_cpu_offload:
+            _pipeline.enable_sequential_cpu_offload()
+
+
+def _maybe_warn_if_no_keys_found(state_dict, prefix, model_class_name):
```
DN6: This is kind of a thin wrapper around a warning message. I don't think it needs to be it's own utility function.

src/diffusers/utils/peft_utils.py
```diff
@@ -288,3 +292,112 @@ def check_peft_version(min_version: str) -> None:
...
+            raise ValueError("lora_bias requires PEFT >= 0.14.0. Please upgrade.")
+
+    try:
+        return LoraConfig(**lora_config_kwargs)
+    except TypeError as e:
+        raise TypeError("`LoraConfig` class could not be instantiated.") from e
+
+
+@contextlib.contextmanager
+def _lora_loading_context(_pipeline):
```
DN6: Makes sense as a context manager, but I don't see it being applied anywhere?

sayakpaul: @BenjaminBossan and I decided to remove its use because we preferred more explicit code to handle offloading.

DN6 (approved):

Issue #11724, opened 2025-06-16
Chroma fails to mask attention in the transformer


AmericanPresidentJimmyCarter:
### Describe the bug

Chroma has attention masks on both the transformer and T5. Right now only T5 computes with the proper mask. The attention masks from T5 should carry forward into the transformer and mask padding tokens except for the last padding token. See the reference implementation.

https://github.com/lodestone-rock/flow/blob/master/src/models/chroma/model.py#L222-L257

### Reproduction

```py

import torch
from diffusers import ChromaTransformer2DModel
from transformers import T5EncoderModel, T5Tokenizer

from optimum.quanto import freeze, qint8, qfloat8, quantize

from chroma_pipeline import ChromaPipeline

bfl_repo = "black-forest-labs/FLUX.1-schnell"
dtype = torch.bfloat16

transformer = ChromaTransformer2DModel.from_single_file(
    "chroma-unlocked-v37-detail-calibrated.safetensors",
    torch_dtype=dtype,
).to('cuda')
quantize(transformer, weights=qint8, exclude=["proj_out", "x_embedder", "norm_out",
    "context_embedder", "distilled_guidance_layer"])
freeze(transformer)

text_encoder = T5EncoderModel.from_pretrained(bfl_repo, subfolder="text_encoder_2", torch_dtype=dtype).to('cuda')
quantize(text_encoder, weights=qint8)
freeze(text_encoder)
tokenizer = T5Tokenizer.from_pretrained(bfl_repo, subfolder="tokenizer_2", torch_dtype=dtype)

pipe = ChromaPipeline.from_pretrained(bfl_repo, transformer=transformer, text_encoder=text_encoder,
    tokenizer=tokenizer, torch_dtype=dtype).to('cuda')
generator = torch.Generator().manual_seed(0)
image = pipe(
    prompt='pikachu playing a violin on mars, sign in the background says, "welcome to mars!!',
    negative_prompt="",
    width=1024,
    height=1024,
    num_inference_steps=35, 
    generator=generator,
    guidance_scale=5,
).images[0]
image.save('chroma/test_chroma0.png')
```

![Image](https://github.com/user-attachments/assets/4d3b503b-c4e7-4652-b2d4-40c1abeac563)

### Logs

```shell

```

### System Info

linux 24.04, 3090, diffusers main

### Who can help?

@DN6 @Ednaordinary

AmericanPresidentJimmyCarter:
[...]

PR #11725, opened 2025-06-16
Chroma Follow Up 

DN6:
# What does this PR do?
Follow up to #11698 

This PR
1. Adds Img2Img pipeline for Chroma
2. Fixes an issue where we neglected passing the modified attention mask to the transformer model, leading to quality issues mentioned in #11724 
3. Clean up Docstrings
 

Fixes # (issue)
#11724

## Before submitting

## Who can review?


DN6:
@AmericanPresidentJimmyCarter do these changes help with the quality issue you're seeing?

cc: @Ednaordinary

nitinmukesh:
Even I was getting poor results and gave up on this. Let me try again.

asian_model_portrait_1

asian_model_portrait_2

AmericanPresidentJimmyCarter:
test_chroma0

This one is looking much better, thank you.

Refer to the issue for another potential issue (maybe needs a new issue?), I think there may also be issues with T5 quantization because of the way the original authors chose to train it.

nitinmukesh:
After installing
pip install git+https://github.com/huggingface/diffusers.git@refs/pull/11725/head

Better now used this config
```python
    quant_kwargs={
        "load_in_4bit": True,
        "bnb_4bit_quant_type": "nf4",
        "bnb_4bit_compute_dtype": dtype,
        "llm_int8_skip_modules": ["distilled_guidance_layer"],
    },
)

```
asian_model_portrait_0

AmericanPresidentJimmyCarter:
test_chroma0

Tested it again with GGUF 8-bit T5 text transformer (see issue). It seems that to get full quality, we're going to have to figure out what the forked transformer the reference is using does to the text embeddings.

AmericanPresidentJimmyCarter:
Another comparison for this branch.

T5 8-bit quanto:

combined_int8

T5 bfloat16:

combined_bf16

T5 8-bit GGUF:

combined_gguf

It's crazy that there's such a world of difference between them, and neither quant seems quite right..  BF16 appears to give the best quality images.

src/diffusers/pipelines/chroma/pipeline_chroma.py
```diff
@@ -256,6 +256,8 @@ def encode_prompt(
         num_images_per_prompt: int = 1,
         prompt_embeds: Optional[torch.FloatTensor] = None,
         negative_prompt_embeds: Optional[torch.FloatTensor] = None,
+        prompt_attention_mask: Optional[torch.Tensor] = None,
+        negative_prompt_attention_mask: Optional[torch.Tensor] = None,
         do_classifier_free_guidance: bool = True,
         max_sequence_length: int = 512,
         lora_scale: Optional[float] = None,
```
Ednaordinary, line 272: We should include `negative_prompt_embeds` (and prompt_attention_mask, negative_prompt_attention mask) in docs. I missed it in the pipeline PR

Ednaordinary:
Also, now that prompt_embeds and negative_prompt embeds consistently stay the same size, we should batch cond and uncond (also simplifying attention_mask + negative_attention_mask into just attention_mask)

Update: made a PR for this in #11729

Ednaordinary:
Hm.. the attention mask lines up with what lodestone shows, so I'm a bit confused why the picture quality is not on par with ~~how it was before batch refactor~~/~~is in ComfyUI~~ (actually after testing, I can't get a good image in either. maybe the prompt is just bad.)
image

image

src/diffusers/pipelines/chroma/pipeline_chroma_img2img.py
```diff
@@ -0,0 +1,1032 @@
...
+        )
+        self.vae_scale_factor = 2 ** (len(self.vae.config.block_out_channels) - 1) if getattr(self, "vae", None) else 8
+        self.latent_channels = self.vae.config.latent_channels if getattr(self, "vae", None) else 16
+
+        # Flux latents are turned into 2x2 patches and packed. This means the latent width and height has to be divisible
+        # by the patch size. So the vae scale factor is multiplied by the patch size to account for this
+        self.image_processor = VaeImageProcessor(vae_scale_factor=self.vae_scale_factor * 2)
+        self.default_sample_size = 128
+
+    def _get_t5_prompt_embeds(
```
yiyixuxu, line 227: Copied from?

src/diffusers/pipelines/chroma/pipeline_chroma_img2img.py
```diff
@@ -0,0 +1,1032 @@
...
+            ]
+            image_latents = torch.cat(image_latents, dim=0)
+        else:
+            image_latents = retrieve_latents(self.vae.encode(image), generator=generator)
+
+        image_latents = (image_latents - self.vae.config.shift_factor) * self.vae.config.scaling_factor
+
+        return image_latents
+
+    def encode_prompt(
```
yiyixuxu, line 295: copied from?

src/diffusers/pipelines/chroma/pipeline_chroma_img2img.py
```diff
@@ -0,0 +1,1032 @@
...
+
+        if not isinstance(image, torch.Tensor):
+            image = self.feature_extractor(image, return_tensors="pt").pixel_values
+
+        image = image.to(device=device, dtype=dtype)
+        image_embeds = self.image_encoder(image).image_embeds
+        image_embeds = image_embeds.repeat_interleave(num_images_per_prompt, dim=0)
+        return image_embeds
+
+    def prepare_ip_adapter_image_embeds(
```
yiyixuxu, line 411: oh we already have ip adapter?

src/diffusers/pipelines/chroma/pipeline_chroma_img2img.py
```diff
@@ -0,0 +1,1032 @@
...
+            )
+        elif prompt is not None and (not isinstance(prompt, str) and not isinstance(prompt, list)):
+            raise ValueError(f"`prompt` has to be of type `str` or `list` but is {type(prompt)}")
+
+        if negative_prompt is not None and negative_prompt_embeds is not None:
+            raise ValueError(
+                f"Cannot forward both `negative_prompt`: {negative_prompt} and `negative_prompt_embeds`:"
+                f" {negative_prompt_embeds}. Please make sure to only forward one of the two."
+            )
+        if prompt_attention_mask is not None and negative_prompt_attention_mask is None:
```
yiyixuxu: I think if negative_prompt is None, negative_prompt_attention_mask can be None too, no?

src/diffusers/pipelines/chroma/pipeline_chroma_img2img.py
```diff
@@ -0,0 +1,1032 @@
...
+            raise ValueError(
+                "Cannot provide `prompt_attention_mask` without also providing `negative_prompt_attention_mask`"
+            )
+
+        if negative_prompt_attention_mask is not None and prompt_attention_mask is None:
+            raise ValueError(
+                "Cannot provide `negative_prompt_attention_mask` without also providing `prompt_attention_mask`"
+            )
+
```
yiyixuxu, line 503: I think we have to pass the mask, if we pass the embeddings directly, no?
```python
if prompt_embeds is not None, prompt_attention_mask is None: 
    raise ValueError(...)
```

src/diffusers/pipelines/chroma/pipeline_chroma_img2img.py
DN6: So the FluxIPAdapter does work with Chroma (you can load the adapter and run inference). It's just that the quality is not very good because the popular ones are trained for Flux Dev, while Chroma is based on Schnell weights. 

I'm cool to remove it, but the consensus [here](https://github.com/huggingface/diffusers/pull/11698#discussion_r2145170178) was to leave it in case an IPAdapter for Chroma was trained.

src/diffusers/pipelines/chroma/pipeline_chroma_img2img.py
DN6: Here we create a custom attention mask as well for Chroma that unmasks a single pad token. So the implementation is different from existing methods.

src/diffusers/pipelines/chroma/pipeline_chroma_img2img.py
DN6: Returns attention masks, so I think it would be different no?

Ednaordinary: Also has `negative_prompt` and `negative_prompt_embeds` which flux doesn't have

Trojaner:
With this PR in combination with https://github.com/huggingface/diffusers/pull/11729 I need to manually pad prompt_embeds and negative_prompt_embeds to the same size now when I pass them manually. Is this intentional and consistent with other pipelines? This was not needed before these changes.

The following error occurs now whereas it does not in the main branch.
```
23:54:30-094459 ERROR    Processing: step=base args={'prompt_embeds': 'cuda:0:torch.bfloat16:torch.Size([1, 81, 4096])', 'negative_prompt_embeds': 'cuda:0:torch.bfloat16:torch.Size([1, 4, 4096])', 'guidance_scale': 4.7, 'generator':   
                         [<torch._C.Generator object at 0x7e096db5d9d0>], 'callback_on_step_end': <function diffusers_callback at 0x7e090f82c040>, 'callback_on_step_end_tensor_inputs': ['latents', 'prompt_embeds', 'noise_pred'],       
                         'num_inference_steps': 26, 'output_type': 'latent', 'width': 1024, 'height': 1024} Sizes of tensors must match except in dimension 0. Expected size 4 but got size 81 for tensor number 1 in the list.            
23:54:30-107754 ERROR    Processing: RuntimeError                                                                                                                                                                                          
╭─────────────────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ───────────────────────────────────────────────────────────────────────────────────────────────────╮
│ /home/ml/sdnext/modules/processing_diffusers.py:105 in process_base                                                                                                                                                                     │
│                                                                                                                                                                                                                                         │
│   104 │   │   else:                                                                                                                                                                                                                     │
│ ❱ 105 │   │   │   output = shared.sd_model(**base_args)                                                                                                                                                                                 │
│   106 │   │   if isinstance(output, dict):                                                                                                                                                                                              │
│                                                                                                                                                                                                                                         │
│ /home/ml/sdnext/venv/lib/python3.12/site-packages/torch/utils/_contextlib.py:116 in decorate_context                                                                                                                                    │
│                                                                                                                                                                                                                                         │
│   115 │   │   with ctx_factory():                                                                                                                                                                                                       │
│ ❱ 116 │   │   │   return func(*args, **kwargs)                                                                                                                                                                                          │
│   117                                                                                                                                                                                                                                   │
│                                                                                                                                                                                                                                         │
│ /home/ml/diffusers/src/diffusers/pipelines/chroma/pipeline_chroma.py:762 in __call__                                                                                                                                                    │
│                                                                                                                                                                                                                                         │
│   761 │   │   if self.do_classifier_free_guidance:                                                                                                                                                                                      │
│ ❱ 762 │   │   │   prompt_embeds = torch.cat([negative_prompt_embeds, prompt_embeds], dim=0)                                                                                                                                             │
│   763                                                                                                                                                                                                                                   │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 4 but got size 81 for tensor number 1 in the list.
```

This is my workaround:
```py
def get_weighted_prompt_embeddings_chroma(
    pipe: ChromaPipeline,
    prompt: str = "",
    negative_prompt: str = "",
    device=None
):
    # ... do some stuff ...
   
    # before the two PRs, this would work:
    # return prompt_embeds, negative_prompt_embeds
    
    # now this is needed:
    return _pad_prompt_embeds_to_same_size(prompt_embeds, negative_prompt_embeds)


def _pad_prompt_embeds_to_same_size(prompt_embeds_a, prompt_embeds_b):
    size_a = prompt_embeds_a.size(1)
    size_b = prompt_embeds_b.size(1)

    if size_a < size_b:
        pad_size = size_b - size_a
        prompt_embeds_a = F.pad(prompt_embeds_a, (0, 0, 0, pad_size))
    elif size_b < size_a:
        pad_size = size_a - size_b
        prompt_embeds_b = F.pad(prompt_embeds_b, (0, 0, 0, pad_size))

    return prompt_embeds_a, prompt_embeds_b
```

Ednaordinary:
#11729 was made yesterday before a bunch of new commits. @DN6 can we keep them padded to the same size or is a different size more preferable to batching?

DN6:
@Trojaner Padding is consistent with other pipelines in Diffusers. We need it if we want to support batch_size > 1. 

#11729 is introducing batched CFG, so prompt_embeds, and negative_prompt_embeds need to be concatenated and hence need to be padded to the same size.

Trojaner:
> @Trojaner Padding is consistent with other pipelines in Diffusers. We need it if we want to support batch_size > 1.

My question was more about the user having to pad the embeds manually instead of the pipeline doing it automatically. I was not questioning the need for padding itself. I am almost sure this is isn't needed to be done by the user when using (at least some) other pipelines.

hameerabbasi:
> My question was more about the user having to pad the embeds manually instead of the pipeline doing it automatically.

If one doesn't pass in the attention mask when going from prompts to embeds, then yes it is automatic. However; I expect SDnext to actually pass the attention mask in due to e.g. weighted parts of prompts.

PR #11727, opened 2025-06-16
[docs] LoRA scale scheduling

stevhliu:
Adds section for how to schedule different LoRA scales during inference


sayakpaul (approved):
Ouf <3

sayakpaul:
Cc: @linoytsaban @apolinario @asomoza too!

stevhliu:
Merging now and we can iterate on it later if we need to! 🙂

PR #11736, opened 2025-06-18
[Quantizers] add `is_compileable` property to quantizers.

sayakpaul:
# What does this PR do?

Like `transformers`.

We have tests for all of these quantizers with `torch.compile()`. So, we should be good.


src/diffusers/quantizers/bitsandbytes/bnb_quantizer.py
```diff
@@ -564,6 +564,10 @@ def is_trainable(self) -> bool:
         # Because we're mandating `bitsandbytes` 0.43.3.
         return True
 
+    @property
+    def is_compileable(self) -> bool:
+        return True
+
```
SunMarc, line 570: add check for bnb version

SunMarc (approved):
Nice ! Thanks for adding this

src/diffusers/quantizers/bitsandbytes/bnb_quantizer.py
sayakpaul: Hmm, not sure about version-guarding since compilation was supported even before the latest release. It's the "no graph-breaks" that is the catch.

DN6 (approved):

PR #11738, opened 2025-06-18
enable deterministic in bnb 4 bit tests

jiqing-feng:
The motivation is that I can pass all tests when running `pytest -rA tests/quantization/bnb/test_4bit.py`, but failed on `pytest -rA tests/quantization/bnb/test_4bit.py::ExtendedSerializationTest`. That's because running all tests will trigger [torch.use_deterministic_algorithms(True)](https://github.com/jiqing-feng/diffusers/blob/c934720629837257b15fd84d27e8eddaa52b76e6/src/diffusers/utils/testing_utils.py#L1090), and it will have impact on all the following tests. So the correct way should be calling this function at the setup stage.


jiqing-feng:
Hi @sayakpaul , could you please review this PR? Thanks!

sayakpaul (commented):
Should it not be propagated to `test_mixed_int8.py` as well?

jiqing-feng:
> Should it not be propagated to `test_mixed_int8.py` as well?

Right, fixed.

jiqing-feng:
Hold on for a while, need to recheck if it works.

jiqing-feng:
Hi @sayakpaul . I have verified this PR on both `test_4bit.py` and `test_mixed_int8.py`, and they work well. Please review the new changes. Thanks!

tests/quantization/bnb/test_4bit.py
```diff
@@ -96,6 +96,10 @@ class Base4bitTests(unittest.TestCase):
     num_inference_steps = 10
     seed = 0
 
+    @classmethod
+    def setUpClass(cls):
```
sayakpaul, line 100: Where is this method called?

jiqing-feng: The `@classmethod` method `setUpClass()` is automatically called once per test class, before any test methods run.

sayakpaul: We have this:
https://github.com/huggingface/diffusers/blob/7fc53b5d66867f9e59398f5a14dd3dd9fbc700dd/tests/quantization/bnb/test_4bit.py#L124

jiqing-feng: setUp will be called before each test module but setUpClass will be called only once before all test modules in a test class. For `torch.use_deterministic_algorithms(True)`, we only need called once so it's better in the setUpClass.

sayakpaul (approved):

PR #11740, opened 2025-06-18
[tests] add compile + offload tests for GGUF.

sayakpaul:


StrongerXi:
Glad to see this!

DN6 (approved):
Changes look good. Can we test with Flux though.

sayakpaul:
Merging it since the WanVace LoRA failures are really turning out to be voodoo!

PR #11743, opened 2025-06-18
[research_projects] add shortened flux training script with quantization 

DerekLiu35:
adds shortened script to be referenced in this blogpost https://github.com/huggingface/blog/pull/2888

### DIFF
--- a/examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
+++ b/examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
@@ -151,7 +151,7 @@ def main(args):
         if args.output_dir is not None:
             os.makedirs(args.output_dir, exist_ok=True)
         accelerator.init_trackers("dreambooth-flux-dev-lora-alphonse-mucha", config=vars(args))
-    
+
     # Load models with quantization
     noise_scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(
         args.pretrained_model_name_or_path, subfolder="scheduler"


DerekLiu35:
@sayakpaul

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
```diff
@@ -0,0 +1,338 @@
...
+        print(f"Pipeline memory usage: {torch.cuda.max_memory_reserved() / 1024**3:.3f} GB")
+    else:
+        print("Training completed. GPU not available for memory tracking.")
+
+    accelerator.end_training()
+
+if __name__ == "__main__":
+    class Args:
+        pretrained_model_name_or_path = "black-forest-labs/FLUX.1-dev"
+        data_df_path = "embeddings_alphonse_mucha.parquet"
```
sayakpaul: Add a note on where this is coming from.

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
```diff
@@ -0,0 +1,338 @@
...
+                )[0]
+
+                vae_scale_factor = 2 ** (len(vae_config.block_out_channels) - 1)
+                model_pred = FluxPipeline._unpack_latents(
+                    model_pred, model_input.shape[2] * vae_scale_factor,
+                    model_input.shape[3] * vae_scale_factor, vae_scale_factor
+                )
+
+                # Compute loss
+                weighting = compute_loss_weighting_for_sd3("none", sigmas)
```
sayakpaul: We have the `weighting_scheme` args. So, let's use it from there.

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
```diff
@@ -0,0 +1,338 @@
...
+                # Prepare inputs
+                latent_image_ids = FluxPipeline._prepare_latent_image_ids(
+                    model_input.shape[0], model_input.shape[2] // 2, model_input.shape[3] // 2,
+                    accelerator.device, torch.float16
+                )
+
+                noise = torch.randn_like(model_input)
+                bsz = model_input.shape[0]
+
+                u = compute_density_for_timestep_sampling("none", bsz, 0.0, 1.0, 1.29)
```
sayakpaul: `args.weighting_scheme`

Let's also make constants for the magic numbers.

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
```diff
@@ -0,0 +1,338 @@
...
+
+    # Prepare for training
+    transformer, optimizer, train_dataloader, lr_scheduler = accelerator.prepare(transformer, optimizer, train_dataloader, lr_scheduler)
+
+    # Register save/load hooks
+    def unwrap_model(model):
+        model = accelerator.unwrap_model(model)
+        return model._orig_mod if is_compiled_module(model) else model
+
+    def save_model_hook(models, weights, output_dir):
```
sayakpaul: If we don't have a load model hook, then I don't think it will make sense to have this either no? Are do we have a utility to resume from a checkpoint in this script?

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
```diff
@@ -0,0 +1,338 @@
...
+                if isinstance(unwrap_model(model), type(unwrap_model(transformer))):
+                    lora_layers = get_peft_model_state_dict(unwrap_model(model))
+                    FluxPipeline.save_lora_weights(output_dir, transformer_lora_layers=lora_layers, text_encoder_lora_layers=None)
+                    weights.pop() if weights else None
+
+    accelerator.register_save_state_pre_hook(save_model_hook)
+    cast_training_params([transformer], dtype=torch.float32) if args.mixed_precision == "fp16" else None
+
+    # Initialize tracking
+    accelerator.init_trackers("dreambooth-flux-dev-lora-alphonse-mucha", config=vars(args)) if accelerator.is_main_process else None
```
sayakpaul:
```suggestion
if accelerator.is_main_proces
    accelerator.init_trackers("dreambooth-flux-dev-lora-alphonse-mucha", config=vars(args))
```

Can we do it while creating the output folder?

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
```diff
@@ -0,0 +1,338 @@
...
+    # Cache latents
+    vae_config = vae.config
+    latents_cache = []
+    for batch in tqdm(train_dataloader, desc="Caching latents"):
+        with torch.no_grad():
+            pixel_values = batch["pixel_values"].to(accelerator.device, dtype=torch.float16)
+            latents_cache.append(vae.encode(pixel_values).latent_dist)
+
+    del vae
+    free_memory()
```
sayakpaul, line 224:
```suggestion
    vae.cpu()
    del vae
    free_memory()
```

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
```diff
@@ -0,0 +1,338 @@
...
+        transformer.enable_gradient_checkpointing()
+
+    # now we will add new LoRA weights to the attention layers
+    transformer_lora_config = LoraConfig(
+        r=args.rank,
+        lora_alpha=args.rank,
+        init_lora_weights="gaussian",
+        target_modules=["to_k", "to_q", "to_v", "to_out.0"],
+    )
+    transformer.add_adapter(transformer_lora_config)
```
sayakpaul, line 186: Should cast the LoRA params to FP32. Do you have a full run with this script that works without FP32 upcasting?

sayakpaul:
@bot /style

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
DerekLiu35: I kept save_model_hook to save intermediate checkpoints, but I probably didn't need to save optimizer states too. Though, yeah I think adding back load model hook to resume from checkpoints is a good idea

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
DerekLiu35: I think I was casting to FP32 below with 
```cast_training_params([transformer], dtype=torch.float32) if args.mixed_precision == "fp16" else None``` below (probably will move it over here and change it to match original training script better.

I do have a full run with this script with reasonable results without FP32 upcasting.
But, I noticed in the [loss curves](https://wandb.ai/derekliu2021/dreambooth-flux-dev-lora-alphonse-mucha/workspace?nw=nwuserderekliu2021&panelDisplayName=loss&panelSectionName=Charts) are slightly different between nano script (rare-voice-24 run) and original script (fanciful-totem-2) so I will need to find where the discrepancy is coming from.

sayakpaul: If it doesn't affect results, probably okay

examples/research_projects/flux_lora_quantization/train_dreambooth_lora_flux_nano.py
sayakpaul: Yeah either we don't support intermediate checkpoints at all or support it. I think okay without to prefer minimalism.

PR #11758, opened 2025-06-19
[wip] enable FastPersist to (potentially) speed up disk saving in group offloading

sayakpaul:
# What does this PR do?

FastPersist and DeepNVMe: https://github.com/deepspeedai/DeepSpeedExamples/tree/master/deepnvme

Blog post: https://pytorch.org/blog/deepnvme-affordable-i-o-scaling-for-deep-learning-applications/

⛔️ Currently not ready for reviews.

PR #11760, opened 2025-06-19
Follow up for Group Offload to Disk 

DN6:
# What does this PR do?
Since we now have multiple offload options available in group offload (standard, low_cpu_mem_usage, offload_to_disk), it would make sense to create separate utility functions for them and try to reuse common patterns e.g. try to use `cpu_param_dict` and the pinned_memory context manager with streams.

This PR is just to organize the module a bit so that it's a bit easier to add further functionality, e.g. #11758 without having a lot of conditional paths.  

Fixes # (issue)

## Before submitting

## Who can review?


DN6:
btw @sayakpaul This line 
```
self.safetensors_file_path = os.path.join(self.offload_to_disk_path, f"group_{id(self)}.safetensors")
```
Uses the memory address of the group object to create the filename, but that can change across sessions. So you could have the offload directory growing quite substantially if running an inference script multiple times. 

Might need to come up with another way to name the files so that new ones aren't created across sessions.

src/diffusers/hooks/group_offloading.py
```diff
@@ -135,9 +135,58 @@ def _pinned_memory_tensors(self):
...
+    def _onload_from_memory(self, current_stream):
+        if self.stream is not None:
+            with self._pinned_memory_tensors() as pinned_memory:
+                self._process_tensors_from_modules(pinned_memory, current_stream)
+        else:
+            self._process_tensors_from_modules(None, current_stream)
+
     @torch.compiler.disable()
     def onload_(self):
-        r"""Onloads the group of modules to the onload_device."""
```
sayakpaul, line 137: Any reason for it to et out? 👁️

src/diffusers/hooks/group_offloading.py
```diff
@@ -175,67 +224,32 @@ def onload_(self):
...
+            safetensors.torch.save_file(tensors_to_save, self.safetensors_file_path)
+
+        # The group is now considered offloaded to disk for the rest of the session.
+        self._is_offloaded_to_disk = True
+
+        # We do this to free up the RAM which is still holding the up tensor data.
+        for tensor_obj in self.tensor_to_key.keys():
+            tensor_obj.data = torch.empty_like(tensor_obj.data, device=self.offload_device)
 
+    @torch.compiler.disable()
```
sayakpaul: I guess we don't need it as it's being called from `offload_()` which is already decorated with it.

a-r-r-o-w (approved):
Thanks, very clean! Nothing specific catches my eye that looks incorrect, so we should be good to merge :)

PR #10660, opened 2025-01-27
Cosmos

a-r-r-o-w:
_The cosmos is within us. We are made of star-stuff. We are a way for the universe to know itself._

## Transformer

<details>
<summary> test attention </summary>

```python
from typing import Optional
from einops import rearrange

import torch
import torch.nn as nn


class RMSNorm(torch.nn.Module):
    def __init__(
        self, dim: int, elementwise_affine: bool = False, eps: float = 1e-6, device=None, dtype=None
    ):
        super().__init__()
        self.eps = eps
        self.learnable_scale = elementwise_affine
        if self.learnable_scale:
            self.weight = nn.Parameter(torch.empty(dim, device=device, dtype=dtype))
        else:
            self.register_parameter("weight", None)

    def forward(self, x):
        r = x * torch.rsqrt(torch.mean(x**2, dim=-1, keepdim=True) + self.eps)
        if self.weight is None:
            return r
        else:
            return r * self.weight.to(dtype=x.dtype, device=x.device)


def get_normalization(name: str, channels: int):
    if name == "I":
        return nn.Identity()
    elif name == "R":
    #     return te.pytorch.RMSNorm(channels, eps=1e-6)
        return RMSNorm(channels, eps=1e-6)
    else:
        raise ValueError(f"Normalization {name} not found")


class Attention(nn.Module):
    def __init__(
        self,
        query_dim: int,
        context_dim=None,
        heads=8,
        dim_head=64,
        dropout=0.0,
        qkv_bias: bool = False,
        out_bias: bool = False,
        qkv_norm: str = "SSI",
        qkv_norm_mode: str = "per_head",
        backend: str = "transformer_engine",
        qkv_format: str = "bshd",
    ) -> None:
        super().__init__()

        self.is_selfattn = context_dim is None  # self attention

        inner_dim = dim_head * heads
        context_dim = query_dim if context_dim is None else context_dim

        self.heads = heads
        self.dim_head = dim_head
        self.qkv_norm_mode = qkv_norm_mode
        self.qkv_format = qkv_format

        if self.qkv_norm_mode == "per_head":
            norm_dim = dim_head
        else:
            raise ValueError(f"Normalization mode {self.qkv_norm_mode} not found, only support 'per_head'")

        self.backend = backend

        self.to_q = nn.Sequential(
            nn.Linear(query_dim, inner_dim, bias=qkv_bias),
            get_normalization(qkv_norm[0], norm_dim),
        )
        self.to_k = nn.Sequential(
[...]

PR #11763, opened 2025-06-21
Avoid creating tensor in CosmosAttnProcessor2_0 (#11761)

### LINKED ISSUE #11761: Avoid creating tensor in CosmosAttnProcessor2_0
https://github.com/huggingface/diffusers/blob/42077e6c734df2fc7bbed373abceab99635500ad/src/diffusers/models/transformers/transformer_cosmos.py#L188-L193
```
        # 4. Prepare for GQA
        query_idx = query.size(3)
        key_idx = key.size(3)
        value_idx = value.size(3)
        key = key.repeat_interleave(query_idx // key_idx, dim=3)
        value = value.repeat_interleave(query_idx // value_idx, dim=3)
```
Speedup ~10% here in Cosmos2TextToImagePipeline and Cosmos2VideoToWorldPipeline.

chenxiao111222:
# What does this PR do?

Fixes #11761 

## Before submitting

## Who can review?


a-r-r-o-w:
Just wanted to make a note that the reason this has to be a tensor is because it seemingly breaks ONNX export. I had it implemented the same way earlier but changed to this after suggestions from the nvidia team. cc @asfiyab-nvidia

yiyixuxu:
ohh thanks for the info @a-r-r-o-w 
10% speed difference is really not small though (if confirmed)

but I think this size ratio is actually determined by config (`inner_dim` vs `inner_kv_dim`) and won't vary at run time, no?

a-r-r-o-w:
@yiyixuxu Yeah it shouldn't vary and we can compute this beforehand. I think the problem stemmed from using an integer (or int-like type) to do the repeat_interleave instead of a tensor. So, it doesn't matter if we compute it with `query.idx(...) / key.idx(...)` or pre-caculate the ratio. I'm not sure about the details but it looks like there were a few similar issues (https://github.com/pytorch/pytorch/issues/100429, for example) which have been marked as resolved. They are very simple examples though, so this "mark dynamic" thing probably does not work with Cosmos (I'm too unfamiliar with ONNX to comment on this).

I don't think we have most of our model definitions compatible with ONNX though (has this been checked before?), so I think it might be okay to make a not and break this compatibility?

yiyixuxu:
ohh thanks @a-r-r-o-w 

do you have the original conversation about onnx export breaking? we could try to look into a solution if there is a reproducible script for the issue 

otherwise, I think the easiest way is we could put an if else in with `if torch.onnx.is_in_onnx_export() else` statement in. Let me know what you think!

a-r-r-o-w:
@yiyixuxu Unfortunately, there's not much to gather from the original conversation. You can find it here: https://github.com/huggingface/diffusers/pull/10660#issuecomment-2752624025

Your suggestion sounds good to me

yiyixuxu:
cc @chenxiao111222 can you add a if torch.onnx.is_in_onnx_export( ) and keep the original code path there?

a-r-r-o-w:
Gentle ping @chenxiao111222. Let me know if you would like me to update with the required changes 🤗

chenxiao111222:
> Gentle ping @chenxiao111222. Let me know if you would like me to update with the required changes 🤗

I'm sorry for the late reply. I mainly use native PyTorch, so I didn't pay attention to the ONNX issue. If you're willing to fix it, please feel free to do so.

PR #11767, opened 2025-06-22
Fix unique memory address when doing group-offloading with disk

sayakpaul:
# What does this PR do?

https://github.com/huggingface/diffusers/pull/11760#issuecomment-2988691717

### DIFF
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -37,7 +37,8 @@ logger = get_logger(__name__)  # pylint: disable=invalid-name
 _GROUP_OFFLOADING = "group_offloading"
 _LAYER_EXECUTION_TRACKER = "layer_execution_tracker"
 _LAZY_PREFETCH_GROUP_OFFLOADING = "lazy_prefetch_group_offloading"
-
+_GROUP_ID_LAZY_LEAF = "lazy_leafs"
+_GROUP_ID_UNMATCHED_GROUP = "top_level_unmatched_modules"
 _SUPPORTED_PYTORCH_LAYERS = (
     torch.nn.Conv1d, torch.nn.Conv2d, torch.nn.Conv3d,
     torch.nn.ConvTranspose1d, torch.nn.ConvTranspose2d, torch.nn.ConvTranspose3d,
@@ -84,7 +85,7 @@ class ModuleGroup:
 
         if self.offload_to_disk_path:
             self._group_id = _group_id
-            short_hash = self._compute_group_hash(self._group_id)
+            short_hash = _compute_group_hash(self._group_id)
             self.safetensors_file_path = os.path.join(self.offload_to_disk_path, f"group_{short_hash}.safetensors")
 
             all_tensors = []
@@ -264,11 +265,6 @@ class ModuleGroup:
                 param.data = param.data.to(self.offload_device, non_blocking=self.non_blocking)
             for buffer in self.buffers:
                 buffer.data = buffer.data.to(self.offload_device, non_blocking=self.non_blocking)
-
-    def _compute_group_hash(self, group_id):
-        hashed_id = hashlib.sha256(group_id.encode("utf-8")).hexdigest()
-        # first 16 characters for a reasonably short but unique name
-        return hashed_id[:16]
 
 
 class GroupOffloadingHook(ModelHook):
@@ -663,7 +659,7 @@ def _apply_group_offloading_block_level(
         stream=None,
         record_stream=False,
         onload_self=True,
-        _group_id="top_level_unmatched_modules",
+        _group_id=_GROUP_ID_UNMATCHED_GROUP,
     )
     if stream is None:
         _apply_group_offloading_hook(module, unmatched_group, None)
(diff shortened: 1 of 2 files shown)


src/diffusers/hooks/group_offloading.py
```diff
@@ -62,6 +65,7 @@ def __init__(
         low_cpu_mem_usage: bool = False,
         onload_self: bool = True,
         offload_to_disk_path: Optional[str] = None,
+        _group_id: Optional[int] = None,
```
DN6:
```suggestion
        group_id: Optional[int] = None,
```

DN6: would set in the following way
```python
self.group_id = group_id or id(self)
```

src/diffusers/hooks/group_offloading.py
```diff
@@ -603,6 +609,9 @@ def _apply_group_offloading_block_level(
 
         for i in range(0, len(submodule), num_blocks_per_group):
             current_modules = submodule[i : i + num_blocks_per_group]
+            start_idx = i
+            end_idx = i + len(current_modules) - 1
+            group_id = f"{name}.{start_idx}_to_{end_idx}"
```
DN6: One liner is good here
```suggestion
            group_id = f"{name}_{i}_{i+len(current_modules)-1}"
```

src/diffusers/hooks/group_offloading.py
```diff
@@ -80,7 +84,9 @@ def __init__(
         self._is_offloaded_to_disk = False
 
         if self.offload_to_disk_path:
-            self.safetensors_file_path = os.path.join(self.offload_to_disk_path, f"group_{id(self)}.safetensors")
+            self._group_id = _group_id
+            short_hash = _compute_group_hash(self._group_id)
+            self.safetensors_file_path = os.path.join(self.offload_to_disk_path, f"group_{short_hash}.safetensors")
```
DN6, line 108: The passed in `group_id` argument should be unique no? I don't think we need to compute a hash.

src/diffusers/hooks/group_offloading.py
```diff
@@ -649,6 +659,7 @@ def _apply_group_offloading_block_level(
         stream=None,
         record_stream=False,
         onload_self=True,
+        _group_id=_GROUP_ID_UNMATCHED_GROUP,
```
DN6: For unmatched group we could also include the module class name e.g. `f"{module.__class__.__name__}_unmatched_group"`

src/diffusers/hooks/group_offloading.py
```diff
@@ -890,3 +904,96 @@ def _get_group_onload_device(module: torch.nn.Module) -> torch.device:
...
     raise ValueError("Group offloading is not enabled for the provided module.")
+
+
+def _compute_group_hash(group_id):
+    hashed_id = hashlib.sha256(group_id.encode("utf-8")).hexdigest()
+    # first 16 characters for a reasonably short but unique name
+    return hashed_id[:16]
+
+
+def _get_expected_safetensors_files(
```
DN6: These appear to be just for tests? Better to have them in the tests folder

src/diffusers/hooks/group_offloading.py
```diff
@@ -890,3 +904,96 @@ def _get_group_onload_device(module: torch.nn.Module) -> torch.device:
         if hasattr(submodule, "_diffusers_hook") and submodule._diffusers_hook.get_hook(_GROUP_OFFLOADING) is not None:
             return submodule._diffusers_hook.get_hook(_GROUP_OFFLOADING).group.onload_device
     raise ValueError("Group offloading is not enabled for the provided module.")
+
+
+def _compute_group_hash(group_id):
```
DN6, line 887: Don't think we need to hash the group id strings, they should be unique already.

src/diffusers/hooks/group_offloading.py
```diff
@@ -783,6 +796,7 @@ def _apply_group_offloading_leaf_level(
             record_stream=False,
             low_cpu_mem_usage=low_cpu_mem_usage,
             onload_self=True,
+            _group_id=_GROUP_ID_LAZY_LEAF,
```
DN6: Don't think this is needed. Lazy hook doesn't do anything with it.

src/diffusers/hooks/group_offloading.py
sayakpaul: I don't see any edge cases, either. But having a hash is a bit more future-proof to me.

src/diffusers/hooks/group_offloading.py
sayakpaul: You mean to exclude `group_id` here? In that case, it would default `id(self)` I think.

src/diffusers/hooks/group_offloading.py
sayakpaul: https://github.com/huggingface/diffusers/pull/11767/files#r2162840565

sayakpaul:
@DN6 a gentle ping

sayakpaul:
@DN6 thanks for your help on this PR :)

PR #11768, opened 2025-06-22
[lora] only remove hooks that we add back

yiyixuxu:


src/diffusers/loaders/lora_base.py
```diff
@@ -442,7 +442,8 @@ def _func_optionally_disable_offloading(_pipeline):
                 logger.info(
                     "Accelerate hooks detected. Since you have called `load_lora_weights()`, the previous hooks will be first removed. Then the LoRA parameters will be loaded and the hooks will be applied again."
                 )
-                remove_hook_from_module(component, recurse=is_sequential_cpu_offload)
+                if is_sequential_cpu_offload or is_model_cpu_offload:
```
yiyixuxu, line 445: cc @sayakpaul 
can you double check here? I'm not sure why we need to remove the hooks, but since we only add back sequential_cpu and model_cpu hooks (https://github.com/huggingface/diffusers/blob/7fc53b5d66867f9e59398f5a14dd3dd9fbc700dd/src/diffusers/loaders/lora_base.py#L409), we should only remove these 2 types

yiyixuxu:
@sayakpaul sounds good

sayakpaul:
Feel free to merge then :)

PR #11771, opened 2025-06-23
[tests] properly skip tests instead of `return`

sayakpaul:
# What does this PR do?

Better accountability IMO.


a-r-r-o-w (approved):
Nice, LGTM! There seems to be a failing test but doesn't look like it comes from this PR

sayakpaul:
Failing tests are unrelated.

PR #11773, opened 2025-06-23
[ci] unify how adhoc installs are done in CI.

sayakpaul:
# What does this PR do?

Haven't touched the ONNX and Flax CI modules as I think they are not used much.


sayakpaul:
@DN6 since it seemed like you're on a reviewing spree today, pinging you.

sayakpaul:
Will continue in https://github.com/huggingface/diffusers/pull/11826.

PR #11775, opened 2025-06-23
[training] add an offload utility that can be used as a context manager.

sayakpaul:
# What does this PR do?

Useful for handling offloading in a simple and easy manner.

@DN6 I am including changes to `examples/**/*.py` to also trigger the fast PR tests on GPU. Hope, that's okay?

### DIFF
--- a/.github/workflows/pr_tests_gpu.yml
+++ b/.github/workflows/pr_tests_gpu.yml
@@ -13,6 +13,7 @@ on:
       - "src/diffusers/loaders/peft.py"
       - "tests/pipelines/test_pipelines_common.py"
       - "tests/models/test_modeling_common.py"
+      - "examples/**/*.py"
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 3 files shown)


linoytsaban:
nice @sayakpaul, thanks! should we also add a test to [test_dreambooth_lora_hidream.py]?(https://github.com/huggingface/diffusers/blob/main/examples/dreambooth/test_dreambooth_lora_hidream.py)

linoytsaban (approved):

sayakpaul:
> should we also add a test to [test_dreambooth_lora_hidream.py]?

I think it should be added as a utility test under `tests/others`. Currently, it's a bit non-trivial to test in terms of how the testing suite is designed for the scripts.

sayakpaul:
@DN6 okay regarding?

> I am including changes to examples/**/*.py to also trigger the fast PR tests on GPU. Hope, that's okay?

sayakpaul:
@linoytsaban two things:

* Would you be able to propagate / introduce the offloading related changes to the other important training scripts?
* In https://github.com/huggingface/diffusers/blob/main/examples/dreambooth/train_dreambooth_lora_flux_kontext.py, we introduced the ability to precompute text embeddings even when each image has a custom prompt. We could propagate that to the rest of the Flux training scripts?

PR #11792, opened 2025-06-24
[chore] raise as early as possible in group offloading

sayakpaul:


src/diffusers/hooks/group_offloading.py
```diff
@@ -63,6 +63,9 @@ def __init__(
         onload_self: bool = True,
         offload_to_disk_path: Optional[str] = None,
     ) -> None:
+        if stream is None and record_stream:
```
a-r-r-o-w: We can remove this completely, no? It's not supposed to be user facing and now apply_group_offloading handles the error case too

sayakpaul: We can but since `ModuleGroup` is a public class I thought of keeping it.

a-r-r-o-w: Not really, it's not a public class - no user is expected to import it or that we document it that way. We don't necessarily need to mark every internal data structure private to imply it shouldn't be used by users

sayakpaul: Removed.

PR #11793, opened 2025-06-24
[tests] skip instead of returning.

sayakpaul:


sayakpaul:
At least we know that the tests were skipped and not passed. Without the changes, we see they were passed.

BenjaminBossan:
> At least we know that the tests were skipped and not passed. Without the changes, we see they were passed.

I guess it's a matter of taste if a check that does not apply should be considered "passing" or "skipped". I can see an argument for both, I don't have a strong opinion on that.

PR #11789, opened 2025-06-24
[tests] Fix HunyuanVideo Framepack device tests

a-r-r-o-w:

PR #11796, opened 2025-06-24
[tests] Fix group offloading and layerwise casting test interaction

a-r-r-o-w:
Still haven't found the core problems for https://github.com/huggingface/diffusers/pull/11789#issuecomment-2999661627 and am investigating it.

Update: cc @DN6 @sayakpaul The cause for the test failures was the inplace modification of the `_WAVELETS` tensors in Cosmos VAE. It was caught by the disk offload tests because it performs an inplace modification of data attributes with an empty tensor


sayakpaul (approved):
🔫

DN6 (approved):

PR #11803, opened 2025-06-25
[CI] disable onnx, mps, flax from the CI

sayakpaul:
# What does this PR do?

In accordance with `transformers`'s [recent move](https://x.com/LysandreJik/status/1933201171130593530) and also something we keep seeing (i.e., none-to-minimal usage of the said frameworks at least from the library), it's wise to disable the CI for these frameworks to keep the load fully dedicated to PyTorch x GPUs, CPUs.


DN6 (commented):
Think we can just remove Flax/ONNX tests completely rather than commenting.

sayakpaul:
@DN6 done.

PR #11804, opened 2025-06-25
Support dynamically loading/unloading loras with group offloading

### LINKED ISSUE #11791: can't load lora weights after apply group offloading
### Describe the bug

i want to dynamically switch the lora with applying group offloading ,but failed

### Reproduction

```python
import torch


model_id = "HunyuanVideo-Diffusers/"
)
pipe.vae.enable_tiling()
pipe.load_lora_weights("a-r-r-o-w/HunyuanVideo-tuxemons/")

list(map(
    lambda x: apply_group_offloading(x, onload_device, offload_device, offload_type="block_level",num_blocks_per_group=1,use_stream=True,non_blocking=True),
    [pipe.text_encoder,pipe.text_encoder_2,pipe.transformer]
))
pipe.vae.to('cuda')


output = pipe(
    prompt="Style of snomexut, a cat-like Tuxemon creature walks in alien-world grass, and observes its surroundings.",
    height=768,
    width=768,
    num_frames=33,
    num_inference_steps=30,
    generator=torch.Generator().manual_seed(73),
).frames[0]
export_to_video(output, "output-tuxemon.mp4", fps=15)



pipe.unload_lora_weights()

output_1 = pipe(
    prompt="Style of snomexut, a cat-like Tuxemon creature walks in alien-world grass, and observes its surroundings.",
    height=768,
    width=768,
    num_frames=33,
    num_inference_steps=30,
    generator=torch.Generator().manual_seed(73),
).frames[0]
export_to_video(output_1, "output-tuxemon-1.mp4", fps=15)


pipe.load_lora_weights("a-r-r-o-w/HunyuanVideo-tuxemons/")
output_2 = pipe(
    prompt="Style of snomexut, a cat-like Tuxemon creature walks in alien-world grass, and observes its surroundings.",
    height=768,
    width=768,
    num_frames=33,
    num_inference_steps=30,
    generator=torch.Generator().manual_seed(73),
).frames[0]
export_to_video(output_2, "output-tuxemon-2.mp4", fps=15)
```

log:

Image

### Logs

```shell

```

### System Info

diffusers 0.34.0.dev0

### Who can help?

@a-r-r-o-w

a-r-r-o-w:
Fixes #11791.

<details>
<summary> reproducer </summary>

```python
import torch

set_verbosity_debug()


)
pipe.vae.enable_tiling()
pipe.load_lora_weights("a-r-r-o-w/HunyuanVideo-tuxemons")
num_inference_steps = 20

list(map(
    lambda x: apply_group_offloading(x, onload_device, offload_device, offload_type="leaf_level", use_stream=True),
    [pipe.transformer]
))
[module.to(onload_device) for module in (pipe.text_encoder, pipe.text_encoder_2, pipe.vae)]

output = pipe(
    prompt="Style of snomexut, a cat-like Tuxemon creature walks in alien-world grass, and observes its surroundings.",
    height=768,
    width=768,
    num_frames=33,
    generator=torch.Generator().manual_seed(73),
).frames[0]
export_to_video(output, "output.mp4", fps=15)

pipe.unload_lora_weights()

output_1 = pipe(
    prompt="Style of snomexut, a cat-like Tuxemon creature walks in alien-world grass, and observes its surroundings.",
    height=768,
    width=768,
    num_frames=33,
    generator=torch.Generator().manual_seed(73),
).frames[0]
export_to_video(output_1, "output2.mp4", fps=15)

pipe.load_lora_weights("a-r-r-o-w/HunyuanVideo-tuxemons")

output_2 = pipe(
    prompt="Style of snomexut, a cat-like Tuxemon creature walks in alien-world grass, and observes its surroundings.",
    height=768,
    width=768,
    num_frames=33,
    generator=torch.Generator().manual_seed(73),
).frames[0]
export_to_video(output_2, "output3.mp4", fps=15)
```
</details>


a-r-r-o-w:
cc @zhangvia I think this should fix the issues you were facing. Could you test? Thanks 🤗

src/diffusers/loaders/lora_base.py
```diff
@@ -440,21 +440,23 @@ def _func_optionally_disable_offloading(_pipeline):
...
-                    )
+            if not isinstance(component, nn.Module) or not hasattr(component, "_hf_hook"):
+                continue
+            is_model_cpu_offload = is_model_cpu_offload or isinstance(component._hf_hook, CpuOffload)
+            is_sequential_cpu_offload = is_sequential_cpu_offload or (
+                isinstance(component._hf_hook, AlignDevicesHook)
+                or hasattr(component._hf_hook, "hooks")
+                and isinstance(component._hf_hook.hooks[0], AlignDevicesHook)
+            )
+            logger.info(
```
sayakpaul, line 464: Shouldn't this be called `if is_model_cpu_offload or is_sequential_cpu_offload`?

tests/lora/utils.py
```diff
@@ -2355,3 +2355,68 @@ def test_inference_load_delete_load_adapters(self):
...
                 output_lora_loaded = pipe(**inputs, generator=torch.manual_seed(0))[0]
                 self.assertTrue(np.allclose(output_adapter_1, output_lora_loaded, atol=1e-3, rtol=1e-3))
+
+    @parameterized.expand([("block_level", True), ("leaf_level", False), ("leaf_level", True)])
+    def test_group_offloading_inference_denoiser(self, offload_type, use_stream):
+        from diffusers.hooks.group_offloading import _GROUP_OFFLOADING
+
+        onload_device = torch_device
+        offload_device = torch.device("cpu")
+        dtype = torch.float32
```
sayakpaul: (nit): by default we should load in fp32. So, maybe not needed.

tests/lora/utils.py
```diff
@@ -2355,3 +2355,68 @@ def test_inference_load_delete_load_adapters(self):
...
+            pipe(**inputs, generator=torch.manual_seed(0))[0]
+
+            # Add the lora again and check if group offloading works
+            pipe.load_lora_weights(os.path.join(tmpdirname, "pytorch_lora_weights.safetensors"))
+            check_if_lora_correctly_set(denoiser)
+            group_offload_hook_3 = denoiser_hook_registry.get_hook(_GROUP_OFFLOADING)
+            self.assertTrue(group_offload_hook_3 is not None)
+            self.assertNotEqual(group_offload_hook, group_offload_hook_3)
+            self.assertNotEqual(group_offload_hook_2, group_offload_hook_3)
+            pipe(**inputs, generator=torch.manual_seed(0))[0]
```
sayakpaul: Should we also assert on the `pipe` call outputs?

sayakpaul (commented):
Thanks! Just some minor comments.

And maybe we could include disabling and enabling group offloading in the existing `_func_optionally_*()` function. But not strong opinions.

src/diffusers/loaders/lora_base.py
a-r-r-o-w: Yes you're right, moved!

tests/lora/utils.py
a-r-r-o-w: Done

src/diffusers/hooks/group_offloading.py
```diff
@@ -268,9 +288,12 @@ class GroupOffloadingHook(ModelHook):
 
     _is_stateful = False
 
-    def __init__(self, group: ModuleGroup, next_group: Optional[ModuleGroup] = None) -> None:
+    def __init__(
+        self, group: ModuleGroup, next_group: Optional[ModuleGroup] = None, *, config: GroupOffloadingConfig
```
sayakpaul, line 312: ❤️

src/diffusers/hooks/group_offloading.py
```diff
@@ -512,84 +537,45 @@ def apply_group_offloading(
...
+
+
+def _apply_group_offloading(module: torch.nn.Module, config: GroupOffloadingConfig) -> None:
+    if config.offload_type == GroupOffloadingType.BLOCK_LEVEL:
+        _apply_group_offloading_block_level(module, config)
+    elif config.offload_type == GroupOffloadingType.LEAF_LEVEL:
+        _apply_group_offloading_leaf_level(module, config)
     else:
-        raise ValueError(f"Unsupported offload_type: {offload_type}")
+        assert False
```
sayakpaul, line 585: Should this be turned into a sensible value error like previous? 👁️

a-r-r-o-w:
> And maybe we could include disabling and enabling group offloading in the existing _func_optionally_*() function. But not strong opinions.

Yep, done!

sayakpaul (approved):
Fantastic!

src/diffusers/hooks/group_offloading.py
a-r-r-o-w: The statement cannot be reached because we instantiate GroupOffloadingType with the value passed by user. Dataclass and enum will raise an error if the value is invalid

assert False is just a placeholder

lllllinux:
> cc @zhangvia I think this should fix the issues you were facing. Could you test? Thanks 🤗

Thanks for the quick fix！ i've Confirmed it.

a-r-r-o-w:
A bit of a problem with the tests that weren't caught until now...

Group offloading with streams is limited in what it can do. If the same layer is invoked twice in the same parent layer's forward, the prefetching logic becomes completely incorrect. This is the case with:
- CogView4 - invokes the same MLP on both hidden_states and encoder_hidden_states in succession
- CogVideoX - invokes the same layernorm on both hidden_states and encoder_hidden_states in succession

One option to make it work is to concatenate inputs across sequence dim and then split again. This will, however, incur some extra perf cost because concatenation/split is not free.

Another option is creating a separate layer for each data stream and sharing the weights. I don't think this should incur memory overhead since the data reference will be same for both layers, but need to test to be sure.

Any other ideas are welcome. If these don't sound good, I propose we skip the tests for now and wait for group offloading logic to become more mature/stable. Other than than PR looks good to merge to me

sayakpaul:
Thanks for investigating these and also for proposing the potential solutions. 

On the surface, I would say we evaluate both approaches and then decide. However, the two models you mentioned probably have limited usage at least with group offloading for now. So, `xfail`ing them is the more reasonable approach. WDYT?

a-r-r-o-w:
I tested the first approach as it's a super simple change. The performance penalty is not noticeable end-to-end but only shows up at a small microsecond scale. I don't think it really matters because, like you mentioned, they probably have very limited usage in the context of group offloading.

For now, I'll specialize the tests for CogView4/CogVideoX by parameterizing with only non-stream tests instead of xfailing them, and add a note. Sounds good?

sayakpaul:
> For now, I'll specialize the tests for CogView4/CogVideoX by parameterizing with only non-stream tests instead of xfailing them, and add a note. Sounds good?

I am good!

a-r-r-o-w:
Updated the tests. There are some gynamistics involved in skipping tests marked with parameterized because it seems like can't be overwritten or specialized in child classes

tests/lora/test_lora_layers_cogvideox.py
```diff
@@ -127,6 +129,13 @@ def test_simple_inference_with_text_denoiser_lora_unfused(self):
     def test_lora_scale_kwargs_match_fusion(self):
         super().test_lora_scale_kwargs_match_fusion(expected_atol=9e-3, expected_rtol=9e-3)
 
+    @parameterized.expand([("block_level", True), ("leaf_level", False)])
+    @require_torch_accelerator
+    def test_group_offloading_inference_denoiser(self, offload_type, use_stream):
+        # TODO: We don't run the (leaf_level, True) test here that is enabled for other models.
+        # The reason for this can be found here: https://github.com/huggingface/diffusers/pull/11804#issuecomment-3013325338
+        super()._test_group_offloading_inference_denoiser(offload_type, use_stream)
```
sayakpaul, line 137: We can do this or we could detect if the test class is either of CogView4 or CogVideoX and use `pytest.skip()`. Upto you.

a-r-r-o-w: Prefer the base test class not having any information about the model test classes that derive it. Current implementation will work for any model that overrides the test, so also much cleaner

PR #11806, opened 2025-06-25
[lora]feat: use exclude modules to loraconfig.

sayakpaul:
# What does this PR do?

If we try to do:

```py
from diffusers import AutoModel, DiffusionPipeline 
import torch

model_id = "Wan-AI/Wan2.1-VACE-14B-diffusers"
vae = AutoModel.from_pretrained(model_id, subfolder="vae", torch_dtype=torch.float32)
pipe = DiffusionPipeline.from_pretrained(model_id, vae=vae, torch_dtype=torch.bfloat16).to("cuda")
pipe.load_lora_weights(
   "vrgamedevgirl84/Wan14BT2VFusioniX", 
   weight_name="FusionX_LoRa/Wan2.1_T2V_14B_FusionX_LoRA.safetensors",
)
```

It prints:

<details>
<summary>Unfold</summary>

```bash
Loading adapter weights from state_dict led to missing keys in the model: vace_blocks.0.proj_out.lora_A.default_0.weight, vace_blocks.0.proj_out.lora_B.default_0.weight, vace_blocks.1.proj_out.lora_A.default_0.weight, vace_blocks.1.proj_out.lora_B.default_0.weight, vace_blocks.2.proj_out.lora_A.default_0.weight, vace_blocks.2.proj_out.lora_B.default_0.weight, vace_blocks.3.proj_out.lora_A.default_0.weight, vace_blocks.3.proj_out.lora_B.default_0.weight, vace_blocks.4.proj_out.lora_A.default_0.weight, vace_blocks.4.proj_out.lora_B.default_0.weight, vace_blocks.5.proj_out.lora_A.default_0.weight, vace_blocks.5.proj_out.lora_B.default_0.weight, vace_blocks.6.proj_out.lora_A.default_0.weight, vace_blocks.6.proj_out.lora_B.default_0.weight, vace_blocks.7.proj_out.lora_A.default_0.weight, vace_blocks.7.proj_out.lora_B.default_0.weight.
```

</details>

It happens because `target_modules` in `LoraConfig` treats the values in `target_modules` as suffixes. In this case, `proj_out` is a part of `target_modules` which is why `vace_blocks_*` get targeted here (and hence the missing warning). 

`LoraConfig` allows us to specify `exclude_modules`, too. This PR introduces support to add it when initializing the `LoraConfig`. 

@apolinario does it work for you?

### DIFF
--- a/tests/lora/test_lora_layers_wan.py
+++ b/tests/lora/test_lora_layers_wan.py
@@ -12,12 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import sys
-import tempfile
 import unittest
 
-import safetensors.torch
 import torch
 from transformers import AutoTokenizer, T5EncoderModel
 
@@ -30,15 +27,13 @@ from diffusers import (
 from diffusers.utils.testing_utils import (
     floats_tensor,
     require_peft_backend,
-    require_peft_version_greater,
     skip_mps,
-    torch_device,
 )
 
 
 sys.path.append(".")
 
-from utils import PeftLoraLoaderMixinTests, check_module_lora_metadata  # noqa: E402
+from utils import PeftLoraLoaderMixinTests  # noqa: E402
 
 
 @require_peft_backend
@@ -146,45 +141,3 @@ class WanLoRATests(unittest.TestCase, PeftLoraLoaderMixinTests):
     @unittest.skip("Text encoder LoRA is not supported in Wan.")
     def test_simple_inference_with_text_lora_save_load(self):
         pass
-
-    @require_peft_version_greater("0.13.2")
-    def test_lora_exclude_modules_for_wan(self):
-        """
-        We test if the modules from `target_modules` and `exclude_modules`, that have
-        overlap in their names, are impacted as expected. Refer to
-        https://github.com/huggingface/diffusers/pull/11806 for more details.
-        """
-        scheduler_cls = self.scheduler_classes[0]
-        components, text_lora_config, denoiser_lora_config = self.get_dummy_components(scheduler_cls)
-        pipe = self.pipeline_class(**components).to(torch_device)
-        _, _, inputs = self.get_dummy_inputs(with_generator=False)
-
-        # Only denoiser for now.
-        denoiser_lora_config.target_modules = ["to_q", "to_k", "to_v", "out"]
-        denoiser_lora_config.exclude_modules = ["proj_out"]
-        pipe, _ = self.add_adapters_to_pipeline(
-            pipe, text_lora_config=text_lora_config, denoiser_lora_config=denoiser_lora_config
-        )
-        # Inference works.
-        _ = pipe(**inputs, generator=torch.manual_seed(0))[0]
-
-        with tempfile.TemporaryDirectory() as tmpdir:
-            modules_to_save = self._get_modules_to_save(pipe, has_denoiser=True)
-            lora_state_dicts = self._get_lora_state_dicts(modules_to_save)
-            lora_metadatas = self._get_lora_adapter_metadata(modules_to_save)
-            self.pipeline_class.save_lora_weights(save_directory=tmpdir, **lora_state_dicts, **lora_metadatas)
-            pipe.unload_lora_weights()
-
-            # Check the state dict. It should not have any `proj_out` related modules.
-            state_dict = safetensors.torch.load_file(os.path.join(tmpdir, "pytorch_lora_weights.safetensors"))
-            # There should not be any `proj_out` modules, but there should still be some modules for `out`.
-            self.assertTrue(not any("proj_out" in k for k in state_dict))
-            self.assertTrue("out" in k for k in state_dict)
-
-            # Check if the metadata matches.
-            out = pipe.lora_state_dict(tmpdir, return_lora_metadata=True)
-            _, parsed_metadata = out
-            check_module_lora_metadata(
-                parsed_metadata=parsed_metadata, lora_metadatas=lora_metadatas, module_key="transformer"
-            )
-            # Inference matching is already tested in `test_lora_exclude_modules`.


src/diffusers/utils/peft_utils.py
```diff
@@ -371,3 +379,20 @@ def _maybe_warn_for_unhandled_keys(incompatible_keys, adapter_name):
 
     if warn_msg:
         logger.warning(warn_msg)
+
+
+def _derive_exclude_modules(model_state_dict, peft_state_dict, adapter_name=None):
```
BenjaminBossan, line 393: A docstring or comment here would be great.

BenjaminBossan (commented):
Generally looks good. As already discussed, for the final PR, we need a version guard on PEFT, as `exclude_modules` was only added in v0.14.0. For the finished PR, there should also be unit tests.

src/diffusers/utils/peft_utils.py
```diff
@@ -371,3 +383,28 @@ def _maybe_warn_for_unhandled_keys(incompatible_keys, adapter_name):
...
+
+    for name in model_state_dict.keys():
+        if string_to_replace:
+            name = name.replace(string_to_replace, "")
+        if "." in name:
+            module_name = name.rsplit(".", 1)[0]
+            all_modules.add(module_name)
+
+    target_modules_set = {name.split(".lora")[0] for name in peft_state_dict.keys()}
+    print(f"{target_modules_set=}")
```
a-r-r-o-w: remove?

tests/lora/utils.py
```diff
@@ -2308,6 +2323,50 @@ def test_lora_adapter_metadata_save_load_inference(self, lora_alpha):
...
+
+        output_no_lora = pipe(**inputs, generator=torch.manual_seed(0))[0]
+        self.assertTrue(output_no_lora.shape == self.output_shape)
+
+        # only supported for `denoiser` now
+        pipe_cp = copy.deepcopy(pipe)
+        pipe_cp, _ = self.check_if_adapters_added_correctly(
+            pipe_cp, text_lora_config=text_lora_config, denoiser_lora_config=denoiser_lora_config
+        )
+        denoiser_exclude_modules = self._get_exclude_modules(pipe_cp)
```
a-r-r-o-w, line 2367: I'm afraid I don't fully understand how this test works but that shouldn't be a blocker. If this tests the exact same condition we're facing in the FusionX lora, then it should be good :)

src/diffusers/utils/peft_utils.py
```diff
@@ -371,3 +383,28 @@ def _maybe_warn_for_unhandled_keys(incompatible_keys, adapter_name):
...
+    `model_state_dict` and `peft_state_dict` and adds a module from `model_state_dict` to the exclusion set if it
+    doesn't exist in `peft_state_dict`.
+    """
+    if model_state_dict is None:
+        return
+    all_modules = set()
+    string_to_replace = f"{adapter_name}." if adapter_name else ""
+
+    for name in model_state_dict.keys():
+        if string_to_replace:
```
a-r-r-o-w, line 405: maybe the if-statement is not needed here because string_to_replace will be an empty string, but no problem keeping as micro optimization

tests/lora/utils.py
sayakpaul: No, not the cleanest test this one. I added a description in the test, see if that helps?

Currently, we don't have:
> If this tests the exact same condition we're facing in the FusionX lora

I will think of a way to include a test for that, too.

src/diffusers/utils/peft_utils.py
```diff
@@ -195,6 +196,16 @@ def get_peft_kwargs(rank_dict, network_alpha_dict, peft_state_dict, is_unet=True
         "use_dora": use_dora,
         "lora_bias": lora_bias,
     }
+
+    # Example: try load FusionX LoRA into Wan VACE
+    exclude_modules = _derive_exclude_modules(model_state_dict, peft_state_dict, adapter_name)
+    if exclude_modules:
+        if not is_peft_version(">=", "0.14.0"):
+            msg = "It seems like there are certain modules that need to be excluded when initializing `LoraConfig`. Your current `peft` version doesn't support passing an `exclude_modules` to `LoraConfig`. Please update it by running `pip install -U peft`."
```
BenjaminBossan: Just wondering: The `exclude_modules` are derived from _all modules - modules of `peft_state_dict`_, right? So in practice, `exlcude_modules` would never be empty. Thus, users with old PEFT versions will always see this warning, even though in the vast majority of cases, there is nothing to worry about. I wonder `exclude_modules` should just be ignored without any warning in that case.

sayakpaul: Or a `debug()` info with this message with an amendment saying "For majority use cases, this should be okay. But if you notice anything unexpected, please file an issue."

BenjaminBossan: Makes sense.

tests/lora/test_lora_layers_wan.py
```diff
@@ -137,3 +146,45 @@ def test_simple_inference_with_text_lora_fused(self):
...
+        https://github.com/huggingface/diffusers/pull/11806 for more details.
+        """
+        scheduler_cls = self.scheduler_classes[0]
+        components, text_lora_config, denoiser_lora_config = self.get_dummy_components(scheduler_cls)
+        pipe = self.pipeline_class(**components).to(torch_device)
+        _, _, inputs = self.get_dummy_inputs(with_generator=False)
+
+        # Only denoiser for now.
+        denoiser_lora_config.target_modules = ["to_q", "to_k", "to_v", "out"]
+        denoiser_lora_config.exclude_modules = ["proj_out"]
```
BenjaminBossan: I wonder: Would this test not pass even without the fix of this PR? When I tested it on main, it passed, even when I commented out the `denoiser_lora_config.exclude_modules = ["proj_out"]` line.

We would need an example with something like:

```python
target_modules = ["proj_out"]
exclude_modules = ["foo.proj_out"]
```

sayakpaul: True. But we will have to then use the Wan VACE model as the current model being tested doesn't meet the module naming criterion we're looking for. I will delete this test case and add it in a future PR then if that sounds good?

BenjaminBossan: SGTM

BenjaminBossan (approved):
Thanks, LGTM.

PR #11814, opened 2025-06-26
Flux: pass joint_attention_kwargs when using gradient_checkpointing

piercus:
Follow up https://github.com/huggingface/diffusers/pull/9517

@yiyixuxu @HorizonWind2004


yiyixuxu (approved):
thanks!

PR #11815, opened 2025-06-26
Kontext fixes

a-r-r-o-w:
Fixes the content filter import to match the latest implementation and a mistake in the pipeline docstring


yiyixuxu (approved):
thanks!

PR #11818, opened 2025-06-27
remove syncs before denoising in Kontext

sayakpaul:
# What does this PR do?

Similar to https://github.com/huggingface/diffusers/pull/11696.

We're working on bringing Kontext to https://github.com/huggingface/flux-fast/.

Cc: @jbschlosser


yiyixuxu (approved):
thanks

PR #11820, opened 2025-06-27
update: FluxKontextInpaintPipeline support

vuongminh1907:
# 🚀 What does this PR do?

Hello! 👋 I'm truly impressed with Flux Kontext, but I noticed that inpainting functionality hasn’t been fully integrated yet. This PR adds support for inpainting using the 🤗 Diffusers library.

This contribution introduces:

## 🎯 Inpainting with text only

1. Example using **FluxKontextInpaintPipeline** with just a prompt:

```
import torch 
from diffusers import FluxKontextInpaintPipeline

prompt = "Change the yellow dinosaur to green one"

img_url = "https://github.com/ZenAI-Vietnam/Flux-Kontext-pipelines/blob/main/assets/dinosaur_input.jpeg?raw=true"
mask_url = "https://github.com/ZenAI-Vietnam/Flux-Kontext-pipelines/blob/main/assets/dinosaur_mask.png?raw=true"

source = load_image(img_url)
mask = load_image(mask_url)

image = pipe(prompt=prompt, image=source, mask_image=mask,strength=1.0).images[0]
image.save("kontext_inpainting_normal.png")
```

2. 🖼️ Original image and mask:

ComfyUI_temp_fnmxi_00001_
download_21

3. ✅ Result using FluxKontextInpaintPipeline:
flux_inpainting

4. ⚠️ When using the regular **FluxKontext** editing pipeline, the color change was not correctly applied to the target object:
generation-b6483443-2cd3-47da-9d97-8966b9d586bb

## 🧩 Inpainting with image conditioning

1. In addition to text prompts, **FluxKontextInpaintPipeline** also supports conditioning on a reference image via the _image_reference_ parameter:

```
import torch 
from diffusers import FluxKontextInpaintPipeline

pipe = FluxKontextInpaintPipeline.from_pretrained("black-forest-labs/FLUX.1-Kontext-dev", torch_dtype=torch.bfloat16)
pipe.to("cuda")
prompt = "Replace this ball"

img_url = "https://images.pexels.com/photos/39362/the-ball-stadion-football-the-pitch-39362.jpeg?auto=compress&cs=tinysrgb&dpr=1&w=500"
mask_url = "https://github.com/ZenAI-Vietnam/Flux-Kontext-pipelines/blob/main/assets/ball_mask.png?raw=true"
image_reference_url = "https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcTah3x6OL_ECMBaZ5ZlJJhNsyC-OSMLWAI-xw&s"

source = load_image(img_url)
mask = load_image(mask_url)
image_reference = load_image(image_reference_url)

mask = pipe.mask_processor.blur(mask, blur_factor=12)
image = pipe(prompt=prompt, image=source, mask_image=mask,image_reference=image_reference,strength=1.0).images[0]
image.save("kontext_inpainting_ref.png")
```

2. 📥 Input image, mask, and reference image:
ball_ref
ball_mask
ball_input

3. 🎉 Output using FluxKontextInpaintPipeline:
flux_inpainting_ball

I hope this PR will be helpful for the community and contribute positively to the Diffusers ecosystem! 🌱

Core library:

- Schedulers: @yiyixuxu
- Training examples: @sayakpaul 
- JAX and MPS: @pcuenca
- Audio: @sanchit-gandhi


nitinmukesh:
Awesome. 👍

apolinario:
Fantastic! Would be cool to have a demo for it on Hugging Face Spaces while the PR gets reviewed :-)

src/diffusers/pipelines/flux/pipeline_flux_kontext_inpaint.py
```diff
@@ -0,0 +1,1398 @@
...
+        max_sequence_length: int = 512,
+        max_area: int = 1024**2,
+        _auto_resize: bool = True,
+    ):
+        r"""
+        Function invoked when calling the pipeline for generation.
+
+        Args:
+            image (`torch.Tensor`, `PIL.Image.Image`, `np.ndarray`, `List[torch.Tensor]`, `List[PIL.Image.Image]`, or `List[np.ndarray]`):
+                `Image`, numpy array or tensor representing an image batch to be used as the starting point. For both
```
yiyixuxu:
```suggestion
                `Image`, numpy array or tensor representing an image batch to to be inpainted (which parts of the image to be masked out with `mask_image` and repainted according to `prompt` and `image_reference`)
```

src/diffusers/pipelines/flux/pipeline_flux_kontext_inpaint.py
```diff
@@ -0,0 +1,1398 @@
...
+                numpy array and pytorch tensor, the expected value range is between `[0, 1]` If it's a tensor or a list
+                or tensors, the expected shape should be `(B, C, H, W)` or `(C, H, W)`. If it is a numpy array or a
+                list of arrays, the expected shape should be `(B, H, W, C)` or `(H, W, C)` It can also accept image
+                latents as `image`, but if passing latents directly it is not encoded again.
+            image_reference (`torch.Tensor`, `PIL.Image.Image`, `np.ndarray`, `List[torch.Tensor]`, `List[PIL.Image.Image]`, or `List[np.ndarray]`):
+                `Image`, numpy array or tensor representing an image batch to be used as the starting point. For both
```
yiyixuxu:
```suggestion
                `Image`, numpy array or tensor representing an image batch to be used as the starting point for the masked area. For both
```

src/diffusers/pipelines/flux/pipeline_flux_kontext_inpaint.py
```diff
@@ -0,0 +1,1398 @@
...
+        return self._interrupt
+
+    @torch.no_grad()
+    @replace_example_docstring(EXAMPLE_DOC_STRING)
+    def __call__(
+        self,
+        image: Optional[PipelineImageInput] = None,
+        image_reference: Optional[PipelineImageInput] = None,
+        mask_image: PipelineImageInput = None,
+        masked_image_latents: PipelineImageInput = None,
```
yiyixuxu: let's not support `masked_image_latents` for now so that we can simplify the logic a bit here - it's not used here

src/diffusers/pipelines/flux/pipeline_flux_kontext_inpaint.py
```diff
@@ -0,0 +1,1398 @@
...
+        mask_condition = self.mask_processor.preprocess(
+            mask_image, height=height, width=width, resize_mode=resize_mode, crops_coords=crops_coords
+        )
+
+        if masked_image_latents is None:
+            masked_image = init_image * (mask_condition < 0.5)
+        else:
+            masked_image = masked_image_latents
+
+        mask, masked_image_latents = self.prepare_mask_latents(
```
yiyixuxu: I don't see masked_image_latents being used in this pipeline, let's remove logics that we don't need here

src/diffusers/pipelines/flux/pipeline_flux_kontext_inpaint.py
```diff
@@ -0,0 +1,1398 @@
...
+            max_sequence_length=max_sequence_length,
+        )
+
+        self._guidance_scale = guidance_scale
+        self._joint_attention_kwargs = joint_attention_kwargs
+        self._current_timestep = None
+        self._interrupt = False
+
+        # 2. Preprocess image
+        if image is not None and not (isinstance(image, torch.Tensor) and image.size(1) == self.latent_channels):
```
yiyixuxu, line 1134: I think `image` is not optional for this pipeline, no?

yiyixuxu (commented):
thanks for the super cool PR! I left some comments 
would you be able to provide more results for different strength values for both with and without the image_reference?

vuongminh1907:
Hi @yiyixuxu, I just updated the code based on your comments — thank you so much!

About updating the results with more `strength` values, I did a quick test and here are the outputs:

## Inpainting with text only
5V (1)
## Inpainting with image conditioning
5V (2)

It seems that `KontextInpaint` works quite well with `strength=1.0`.

vuongminh1907:
> Fantastic! Would be cool to have a demo for it on Hugging Face Spaces while the PR gets reviewed :-)

While waiting for the PR to be reviewed, feel free to test it using: 
`pip install git+https://github.com/vuongminh1907/diffusers`

nitinmukesh:
Thanks again @vuongminh1907 

Tested using 
pip install git+https://github.com/huggingface/diffusers.git@refs/pull/11820/head

Working good with nunchaku optimization. Will do more tests.

Kindly review and merge.

**Nunchaku output**


kontext_inpainting_normal

vuongminh1907:
Just to confirm @yiyixuxu , should I add another test case, or will you take care of it?

nne998:
Hi! Nice work! Can you add a custom node to  ComfyUI to support this ? @vuongminh1907

vuongminh1907:
> Hi! Nice work! Can you add a custom node to ComfyUI to support this ? @vuongminh1907 
okey @nne998, I will take a look soon

vuongminh1907:
@nne998, I think ComfyUI can also work, these are 2 examples:

1. Prompt: `Make the girl take a phone instead of bread`
phone

2. Prompt: `Make the girl take eating apple instead of bread`
aplle

strawberrymelonpanda:
> I think ComfyUI can also work, these are 2 examples:

@vuongminh1907 How about the image_reference parameter?  Will that work with existing nodes?
Great work, and thanks!

_Edit: Apologies for being somewhat off-topic with a ComfyUI question. I was linked here from a ComfyUI issue and didn't realize I'd switched repos at first._

vuongminh1907:
@strawberrymelonpanda, I’ll take a look at the `image_reference` later. In the meantime, I’ve just released the `Kontext Inpainting Node` here: [ComfyUI-Kontext-Inpainting](https://github.com/ZenAI-Vietnam/ComfyUI-Kontext-Inpainting)

yiyixuxu:
would you be able to add a test case? can reference the current tests for flux pipelines https://github.com/huggingface/diffusers/tree/main/tests/pipelines/flux

vuongminh1907:
thanks @yiyixuxu , I added test case and it passed all checks.

yiyixuxu:
@bot /style

HaydenCupcake:
> Hi @yiyixuxu, I just updated the code based on your comments — thank you so much!
> 
> About updating the results with more `strength` values, I did a quick test and here are the outputs:
> 
> ## Inpainting with text only
> 5V (1)
> ## Inpainting with image conditioning
> 5V (2)
> 
> It seems that `KontextInpaint` works quite well with `strength=1.0`.
> 
> 

Hi, is there a ComfyUI workflow that you can share that already includes "Inpainting with image conditioning"?

nne998:
Hi there, I’ve tested the 'Kontext Inpainting Conditioning' node in ComfyUI, and it works great! Could you add support for reference image conditioning to this node?  @vuongminh1907

kkumar61:
> > Hi @yiyixuxu, I just updated the code based on your comments — thank you so much!
> > About updating the results with more `strength` values, I did a quick test and here are the outputs:
> > ## Inpainting with text only
> > 5V (1)
> > ## Inpainting with image conditioning
> > 5V (2)
> > It seems that `KontextInpaint` works quite well with `strength=1.0`.
> 
> Hi, is there a ComfyUI workflow that you can share that already includes "Inpainting with image conditioning"?

@HaydenCupcake I've been trying your pipeline and it seems to working only on your given example images when reference image is also given. It also doesn't work if we modify the image size ratios of your source image and reference image. Any idea what might be the issue?

HaydenCupcake:
> @HaydenCupcake I've been trying your pipeline and it seems to working only on your given example images when reference image is also given. It also doesn't work if we modify the image size ratios of your source image and reference image. Any idea what might be the issue?

+ +@yiyixuxu

Samuel-Recovery-123:
@yiyixuxu I have encountered the same problem with @HaydenCupcake

nitinmukesh:
@Samuel-Recovery-123  @kkumar61 

Please could you share the input images and prompt for me to check quickly.

yiyixuxu:
hi @kkumar61 
do you know that it does not work functionally or just the result is not as expected? can you share a code example?

> it seems to working only on your given example images when reference image is also given

kkumar61:
> hi @kkumar61 do you know that it does not work functionally or just the result is not as expected? can you share a code example?
> 
> > it seems to working only on your given example images when reference image is also given

@yiyixuxu after enough tests, it seems to be an issue with the understanding of Kontext itself. May be it's not related to the pipeline. Thanks for your attention into this!

kkumar61:
@yiyixuxu @nitinmukesh I've an interesting finding, you might want to take a look at it and let me know what are your thoughts on this!

Source Image-

old

Target product to be placed-

new

Output--

inpainted_image_g4 0_s1 0



now with another shot of the same target product--

big_new


output--

inpainted_image_g4 0_s1 0

asomoza:
@kkumar61  Kontext wasn't trained for using masks, the model was trained to use the whole image so your result makes sense, if you want a product placement you should not use inpainting but the regular pipeline or else you're going to get similar results until you win the lottery and the model places the sofa in the exact same spot that your mask allows it.

To me, this pipeline will only work for some very specific scenarios like the dinosaur one, if you have multiple objects in your image and you just want to change a specific one of them, then you can mask it to so the model works on that one, but again, you can just draw a rectangle over it an tell the model to change the "object" inside the rectangle.

vuongminh1907:
@kkumar61 @asomoza , Kontext Inpainting works well only with text conditioning. When using a reference image, it doesn't perform as well because the model wasn't trained for that task. Naturally, it works only with specific cases. If you want to try inpainting with references, let's train a LoRA for your own task( Like ACE did) using this `KontextInpaintPipeline.` I think that would work better.

PR #11825, opened 2025-06-28
[tests] add test for hotswapping + compilation on resolution changes

sayakpaul:
# What does this PR do?

As promised :)

### DIFF
--- a/src/diffusers/loaders/unet_loader_utils.py
+++ b/src/diffusers/loaders/unet_loader_utils.py
@@ -14,6 +14,8 @@
 import copy
 from typing import TYPE_CHECKING, Dict, List, Union
 
+from torch import nn
+
 from ..utils import logging
 
 
@@ -52,7 +54,7 @@ def _maybe_expand_lora_scales(
             weight_for_adapter,
             blocks_with_transformer,
             transformer_per_block,
-            unet.state_dict(),
+            model=unet,
             default_scale=default_scale,
         )
         for weight_for_adapter in weight_scales
@@ -65,7 +67,7 @@ def _maybe_expand_lora_scales_for_one_adapter(
     scales: Union[float, Dict],
     blocks_with_transformer: Dict[str, int],
     transformer_per_block: Dict[str, int],
-    state_dict: None,
+    model: nn.Module,
     default_scale: float = 1.0,
 ):
     """
@@ -154,6 +156,7 @@ def _maybe_expand_lora_scales_for_one_adapter(
 
         del scales[updown]
 
+    state_dict = model.state_dict()
     for layer in scales.keys():
         if not any(_translate_into_actual_layer_name(layer) in module for module in state_dict.keys()):
             raise ValueError(
--- a/src/diffusers/schedulers/scheduling_scm.py
+++ b/src/diffusers/schedulers/scheduling_scm.py
@@ -168,7 +168,6 @@ class SCMScheduler(SchedulerMixin, ConfigMixin):
         else:
             # max_timesteps=arctan(80/0.5)=1.56454 is the default from sCM paper, we choose a different value here
             self.timesteps = torch.linspace(max_timesteps, 0, num_inference_steps + 1, device=device).float()
-        print(f"Set timesteps: {self.timesteps}")
 
         self._step_index = None
         self._begin_index = None


sayakpaul:
Cc: @strongerxi and @anijain2305 as well.

tests/models/test_modeling_common.py
```diff
@@ -1350,7 +1350,6 @@ def test_model_parallelism(self):
                 new_model = self.model_class.from_pretrained(tmp_dir, device_map="auto", max_memory=max_memory)
                 # Making sure part of the model will actually end up offloaded
                 self.assertSetEqual(set(new_model.hf_device_map.values()), {0, 1})
-                print(f" new_model.hf_device_map:{new_model.hf_device_map}")
```
sayakpaul, line 1353: Unrelated but hopefully okay :-)

tests/models/test_modeling_common.py
```diff
@@ -2110,19 +2114,30 @@ def check_model_hotswap(self, do_compile, rank0, rank1, target_modules0, target_
...
-                output0_after = model(**inputs_dict)["sample"]
-            assert torch.allclose(output0_before, output0_after, atol=tol, rtol=tol)
+                # additionally check if dynamic compilation works.
+                if different_resolutions is not None:
+                    for height, width in self.different_shapes_for_compilation:
+                        new_inputs_dict = self.prepare_dummy_input(height=height, width=width)
+                        _ = model(**new_inputs_dict)
+                else:
+                    output0_after = model(**inputs_dict)["sample"]
+                    assert torch.allclose(output0_before, output0_after, atol=tol, rtol=tol)
```
sayakpaul: We don't need assertions when testing for different shapes as that would make the test unnecessarily complicated. We just check for recompilations.

tests/models/test_modeling_common.py
```diff
@@ -2240,3 +2255,21 @@ def test_hotswap_second_adapter_targets_more_layers_raises(self):
...
                 )
                 assert any("Hotswapping adapter0 was unsuccessful" in log for log in cm.output)
+
+    @parameterized.expand([(11, 11), (7, 13), (13, 7)])
+    @require_torch_version_greater("2.7.1")
+    def test_hotswapping_compile_on_different_shapes(self, rank0, rank1):
+        different_shapes_for_compilation = self.different_shapes_for_compilation
+        if different_shapes_for_compilation is None:
+            pytest.skip(f"Skipping as `different_shapes_for_compilation` is not set for {self.__class__.__name__}.")
+        torch.fx.experimental._config.use_duck_shape = False
```
sayakpaul, line 2267: Very important!

tests/models/test_modeling_common.py
```diff
@@ -2019,6 +2018,8 @@ class LoraHotSwappingForModelTesterMixin:
 
     """
 
+    different_shapes_for_compilation = None
```
BenjaminBossan, line 2021: Instead of handling this as an attribute, would it not be simpler to handle the shapes via the `different_resolutions` argument that was added to `check_model_hotswap`? IIUC, it's the same value anyway, since you pass:

```python
different_shapes_for_compilation = self.different_shapes_for_compilation
...
self.check_model_hotswap(
    ...
    different_resolutions=different_shapes_for_compilation
)
```

So I would just remove `different_shapes_for_compilation` completely.

tests/models/test_modeling_common.py
BenjaminBossan: It would help to add a comment on why it's very important.

tests/models/test_modeling_common.py
sayakpaul: We rely on the attribute to determine if we should run or skip the underlying test. So, I am not sure how your suggestion stands.

tests/models/test_modeling_common.py
sayakpaul: Done.

tests/models/test_modeling_common.py
BenjaminBossan: `check_model_hotswap` works even with `different_resolutions=None`, so it's not really necessary to skip it. I guess it could be done as an optimization, but then I would remove the `different_resolutions` argument instead and always rely on `self.different_shapes_for_compilation`. Having two arguments with different names but always the same value is confusing IMO.

sayakpaul: Fixed.

docs/source/en/tutorials/using_peft_for_inference.md
```diff
@@ -315,6 +315,8 @@ pipeline.load_lora_weights(
 > [!TIP]
 > Move your code inside the `with torch._dynamo.config.patch(error_on_recompile=True)` context manager to detect if a model was recompiled. If a model is recompiled despite following all the steps above, please open an [issue](https://github.com/huggingface/diffusers/issues) with a reproducible example.
 
+If you expect to varied resolutions during inference with this feature, then make sure set `dynamic=True` during compilation. Refer to [this document](../optimization/fp16#dynamic-shape-compilation) for more details.
```
StrongerXi, line 318: I wonder if we should publicize `use_duck_shape = False` as well...

PR #11826, opened 2025-06-28
[ci] enable hotswapping tests on our nightly CI.

sayakpaul:
# What does this PR do?

LoRA hotswapping with compilation is a crucial feature which deserves (at least nightly) testing. This PR enables that. 

This PR also sets the `RUN_SLOW` flag in `push_tests.yml` for the compilation test job as without it, those tests won't run. Should we enable hotswapping tests in that workflow, too? I am okay if not.

Additionally, a few questions:

* We have both `push_tests_fast.yml` and `push_tests.yml`. I am assuming the tests under `push_tests_fast.yml` would have already run before the changes made their way to `main`. So, I wonder if we should remove `push_tests_fast.yml`?
* Should we allow slow tests to run during `push_tests`? We should most definitely look into how we segregate slow and nightly our tests for this.

I believe this deserves attention as it would help us reduce our CI time.

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -193,24 +193,28 @@ jobs:
       run: |
         python -m venv /opt/venv && export PATH="/opt/venv/bin:$PATH"
         python -m uv pip install -e [quality,test,training]
+        python -m uv pip install peft@git+https://github.com/huggingface/peft.git
     - name: Environment
       run: |
         python utils/print_env.py
-    - name: Run torch compile tests on GPU
+    - name: Run torch hotswap + compile tests on GPU
       env:
         HF_TOKEN: ${{ secrets.DIFFUSERS_HF_HUB_READ_TOKEN }}
         RUN_COMPILE: yes
       run: |
         python -m pytest -n 1 --max-worker-restart=0 --dist=loadfile -s -v -k "compile" --make-reports=tests_torch_compile_cuda tests/
+        python -m pytest -n 1 --max-worker-restart=0 --dist=loadfile -s -v -k "hotswap" --make-reports=tests_torch_hotswap_cuda tests/
     - name: Failure short reports
       if: ${{ failure() }}
-      run: cat reports/tests_torch_compile_cuda_failures_short.txt
+      run: |
+        cat reports/tests_torch_compile_cuda_failures_short.txt
+        cat reports/tests_torch_hotswap_cuda_failures_short.txt
 
     - name: Test suite reports artifacts
       if: ${{ always() }}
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_tests.yml
```diff
@@ -193,24 +193,28 @@ jobs:
...
       run: |
         python utils/print_env.py
-    - name: Run torch compile tests on GPU
+    - name: Run torch hotswap + compile tests on GPU
       env:
         HF_TOKEN: ${{ secrets.DIFFUSERS_HF_HUB_READ_TOKEN }}
         RUN_COMPILE: yes
       run: |
         python -m pytest -n 1 --max-worker-restart=0 --dist=loadfile -s -v -k "compile" --make-reports=tests_torch_compile_cuda tests/
+        python -m pytest -n 1 --max-worker-restart=0 --dist=loadfile -s -v -k "hotswap" --make-reports=tests_torch_hotswap_cuda tests/
```
DN6, line 206: I think this would only run tests with the name hotswap in them and not all compile tests. 

I think the `is_torch_compile` decorator and `RUN_COMPILE` env variable can only used to skip/include compile tests in the larger test suite. They can't be used to isolate compile tests only. 

We should make `is_torch_compile` a marker similar to https://github.com/huggingface/diffusers/pull/11786. And then add it to hotswap tests. That way you have more control over how compile tests run (you can include them as part of larger test suite, isolate them or skip them)

.github/workflows/nightly_tests.yml
```diff
@@ -193,24 +193,28 @@ jobs:
...
     - name: Failure short reports
       if: ${{ failure() }}
-      run: cat reports/tests_torch_compile_cuda_failures_short.txt
+      run: |
+        cat reports/tests_torch_compile_cuda_failures_short.txt
+        cat reports/tests_torch_hotswap_cuda_failures_short.txt
```
DN6, line 211: Can just group hotswap and compile together I think.

DN6:
We don't run slow tests on merge to main. It takes to long to run them on every merge (> 1 hr)  
https://github.com/huggingface/diffusers/pull/8870

If we want to run compile tests on merge, we can just make fast test for them?

DN6:
And we can remove push_tests_fast. It's duplicate work now. Think I forgot to remove it.

sayakpaul:
Discussed it with @dn6 offline. Will change the scope of the PR a bit to maintain consistency in terms of the types of tests being run during each stage (merge, nightly, etc.).

sayakpaul:
Will club https://github.com/huggingface/diffusers/pull/11773 with it too.

PR #11837, opened 2025-06-30
[docs] torch.compile blog post

stevhliu:
Light refactor of the regional compilation docs and adds the https://pytorch.org/blog/presenting-flux-fast-making-flux-go-brrr-on-h100s/ blog post.

### DIFF
--- a/docs/source/en/optimization/fp16.md
+++ b/docs/source/en/optimization/fp16.md
@@ -174,11 +174,18 @@ Feel free to open an issue if dynamic compilation doesn't work as expected for a
 
 ### Regional compilation
 
+[Regional compilation](https://docs.pytorch.org/tutorials/recipes/regional_compilation.html) trims cold-start latency by only compiling the *small and frequently-repeated block(s)* of a model - typically a transformer layer - and enables reusing compiled artifacts for every subsequent occurrence.
+For many diffusion architectures, this delivers the same runtime speed-ups as full-graph compilation and reduces compile time by 8–10x.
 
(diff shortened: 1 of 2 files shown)


docs/source/en/optimization/fp16.md
```diff
@@ -174,11 +174,18 @@ Feel free to open an issue if dynamic compilation doesn't work as expected for a
 
 ### Regional compilation
 
+[Regional compilation](https://docs.pytorch.org/tutorials/recipes/regional_compilation.html) trims cold-start latency by only compiling the *small and frequently-repeated block(s)* of a model - typically a transformer layer - and enables reusing compiled artifacts for every subsequent occurrence.
```
sayakpaul, line 177: Actually regional compilation is preferred this way (with `compile_repeated_blocks()`). So, I would keep it as is.

docs/source/en/optimization/fp16.md
```diff
@@ -296,3 +307,7 @@ An input is projected into three subspaces, represented by the projection matric
 ```py
 pipeline.fuse_qkv_projections()
 ```
+
+## Resources
```
sayakpaul, line 297: Will support https://github.com/huggingface/flux-fast/pull/10 and Kontext is already supported.

sayakpaul (approved):
Perfect. I think we should also try to add a "Tip" for the readers in https://huggingface.co/docs/diffusers/main/en/optimization/speed-memory-optims as well as regional compilation and other compilation related things apply there too.

PR #11841, opened 2025-07-01
[docs] Batch generation

stevhliu:
Improves docs on batch generation following discussion [here](https://huggingface.slack.com/archives/C041M7M98JV/p1750471112003339)


asomoza:
maybe we can add also the batch option of using `num_images_per_prompt` or both (in conjunction with multiple prompts) at the same time?

docs/source/en/using-diffusers/batched_inference.md
```diff
@@ -0,0 +1,264 @@
...
+    axes[i].axis('off')
+
+plt.tight_layout()
+plt.show()
+```
+
+</hfoption>
+<hfoption id="image-to-image">
+
+To generate a batch of images, pass a list of images to the pipeline.
```
asomoza: maybe point out that this is for img2img?

asomoza (approved):
looks good, just one comment

PR #11856, opened 2025-07-03
Fix Wan AccVideo/CausVid fuse_lora

a-r-r-o-w:
Both AccVideo and CausVid work on my end with fuse_lora.

@apolinario Could you give this a try and see if it resolves the issue you faced?


apolinario:
Both commits fix the issue of not allowing `fuse_lora` on CausVid and AccVid

src/diffusers/loaders/lora_conversion_utils.py
```diff
@@ -1825,24 +1825,22 @@ def _convert_non_diffusers_wan_lora_to_diffusers(state_dict):
...
-            if all_zero or all_absdiff_lower_than_threshold:
-                logger.debug(
-                    f"Removed {diff_k} key from the state dict as it's all zeros, or values lower than hardcoded threshold."
-                )
-                original_state_dict.pop(diff_k)
+    for key in list(original_state_dict.keys()):
+        if key.endswith((".diff", ".diff_b")) and "norm" in key:
+            # NOTE: we don't support this because norm layer diff keys are just zeroed values. We can support it
+            # in future if needed and they are not zeroed.
+            original_state_dict.pop(key)
```
sayakpaul, line 1836: Not a blocker but might just quickly zero if it's all zeros:
https://github.com/huggingface/diffusers/blob/8c938fb410e79a0d04d727b68edf28e4036c0ca5/src/diffusers/loaders/lora_conversion_utils.py#L835

src/diffusers/loaders/lora_conversion_utils.py
```diff
@@ -1825,24 +1825,22 @@ def _convert_non_diffusers_wan_lora_to_diffusers(state_dict):
...
+            logger.debug(f"Removing {key} key from the state dict as it is a norm diff key. This is unsupported.")
+
+        if "time_projection" in key and not has_time_projection_weight:
```
sayakpaul, line 1839: 🧠

sayakpaul (approved):
Thanks, Aryan!

src/diffusers/loaders/lora_conversion_utils.py
a-r-r-o-w: From the original codebases where this comes from, I don't think these keys can be anything but zero and yesterday's investigation revealed the same. Probably best to not do anything else here IMO

a-r-r-o-w:
For some reason the tests fail with the latest main merged. It's again one of those errors that when I try to reproduce locally, just passes. If I run the test alone, it passes. If I run all the tests, the test fails. 😢

sayakpaul:
@a-r-r-o-w yeah the VACE tests seem to be a bit `flaky` and I currently don't have an idea why. So, feel free to merge. We can also mark the tests with `is_flaky` for the moment.

PR #11863, opened 2025-07-04
[fix] VaeImageProcessor.resize method can take list as input

Net-Mist:
# What does this PR do?

Hello !

When working on Flux-Kontext, I discovered a silent bug regarding the resizing of the image, when calling the pipeline with several images.

- The `__call__` method of the `FluxKontextPipeline` class takes an `image` of type `Optional[PipelineImageInput]`. This variable can be a list in case of batch size > 1
-  then the `resize` method of the `VaeImageProcessor` class is called (`image = self.image_processor.resize(image, image_height, image_width)`). But this method does not support lists as input. Due to the way this method is written (overriding the `image` variable and then returning it), it doesn't resize the image and return it as-is.

In this fix, I corrected this by changing a minimal number of lines of code to handle lists. However, please let me know if you'd prefer that I also add additional input checks or safeguards.

Fixes # (issue)

## Before submitting

## Who can review?


Net-Mist:
@yiyixuxu @asomoza 
If I'm not mistaken, you are the 2 most relevant people for this PR. Else, sorry for the noise, and please redirect me to the right people

Net-Mist:
Hi @asomoza !
Just following up on this PR to get your feedback. Apologies if this isn’t the right process—if that’s the case, could you please let me know the proper steps?

asomoza:
thanks @Net-Mist the `resize` function is more of an internal method that's called from the `preprocess` function, in that case,  it resizes the image [correctly even if it is a list](https://github.com/huggingface/diffusers/blob/9db9be65f3b9f0902b01428b281c60cffbff09d2/src/diffusers/image_processor.py#L694).

I think the best option here is to refactor the pipeline which is the only case that this happens instead of the `resize` function that all the pipelines use which will add one more check they don't need.

Also I don't understand fully why the kontext pipeline uses the `resize` and the `preprocess`  separately, so ccing @a-r-r-o-w who is better suited to do a review of this PR.

src/diffusers/image_processor.py
```diff
@@ -523,6 +523,16 @@ def resize(
                 size=(height, width),
             )
             image = self.pt_to_numpy(image)
+        elif isinstance(image, list):
```
yiyixuxu, line 526: I think `resize` is intended to only accept single image 
@a-r-r-o-w any reason we run resize sepereately here? (instead of being part of the `preprocess`)? https://github.com/huggingface/diffusers/blob/638cc035e5ecf5c05331c449745f327dbb15e4de/src/diffusers/pipelines/flux/pipeline_flux_kontext.py#L967

a-r-r-o-w: I think preprocess should be okay. I don't remember why I used resize (maybe I just forgot that preprocess existed when I wrote this code. either way I don't remember the context now, but I'll check original repo again)

Issue #10722, opened 2025-02-05
RuntimeError: The size of tensor a (4608) must match the size of tensor b (5120) at non-singleton dimension 2 during DreamBooth Training with Prior Preservation


yinguoweiOvO:
### Describe the bug

I am trying to run "train_dreambooth_lora_flux.py" on my dataset, but the error will happen if --with_prior_preservation is used. 

**Who can help me? Thanks!**

### Reproduction

python ./examples/dreambooth/train_dreambooth_lora_flux.py \
  --pretrained_model_name_or_path=$MODEL_NAME  \
  --instance_data_dir=$INSTANCE_DIR \
  --output_dir=$OUTPUT_DIR \
  --with_prior_preservation \
  --class_data_dir="my_file" \
  --class_prompt="A photo" \
  --instance_prompt="A sks photo" \
  --resolution=1024 \
  --rank=32 \
  --max_train_steps=5000 \
  --checkpointing_steps=100 \
  --seed="0" \
  --mixed_precision="bf16" \
  --train_batch_size=1 \
  --guidance_scale=1 \
  --gradient_accumulation_steps=4 \
  --optimizer="prodigy" \
  --learning_rate=1. \
  --report_to="tensorboard" \
  --lr_scheduler="constant" \
  --lr_warmup_steps=0

### Logs

```shell
Traceback (most recent call last):
  File "/data4/work/yinguowei/code/diffusers/./examples/dreambooth/train_dreambooth_lora_flux.py", line 1926, in <module>
    main(args)
  File "/data4/work/yinguowei/code/diffusers/./examples/dreambooth/train_dreambooth_lora_flux.py", line 1720, in main
    model_pred = transformer(
  File "/data/miniconda3/envs/diffusers_ygw/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/data/miniconda3/envs/diffusers_ygw/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
  File "/data/miniconda3/envs/diffusers_ygw/lib/python3.10/site-packages/accelerate/utils/operations.py", line 819, in forward
    return model_forward(*args, **kwargs)
  File "/data/miniconda3/envs/diffusers_ygw/lib/python3.10/site-packages/accelerate/utils/operations.py", line 807, in __call__
[...]

PR #11873, opened 2025-07-06
[fix] fix for prior preservation and mixed precision sampling

Brvcket:
There were 4 issues fixed:

1. When sampling with prior preservation, the prompt was not passed to the correct argument position, causing a generation error during sampling.
2. When sampling with prior preservation, the transformer passed to the pipeline did not have the torch_dtype argument, causing issues with mixed precision.
3. When sampling with prior preservation, the prompt was passed in fp32 without proper casting, leading to mixed precision mismatches.
4. When not args.train_text_encoder, prompt_embeds and pooled_prompt_embeds were None, leading to the same error as reported in https://github.com/huggingface/diffusers/issues/10722.

### DIFF
--- a/examples/dreambooth/train_dreambooth_lora_flux_kontext.py
+++ b/examples/dreambooth/train_dreambooth_lora_flux_kontext.py
@@ -1193,6 +1193,7 @@ def main(args):
                 subfolder="transformer",
                 revision=args.revision,
                 variant=args.variant,
+                torch_dtype=torch_dtype,
             )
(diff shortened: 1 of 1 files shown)


Brvcket:
@linoytsaban Am I correct in assuming that your review is required? I'm sorry if I tagged the wrong person

linoytsaban:
thanks @Brvcket!

sayakpaul (approved):
Thanks a lot!

PR #11877, opened 2025-07-07
[chore] deprecate blip controlnet pipeline.

sayakpaul:
# What does this PR do?

`transformers` has likely introduced a breaking change which is what is causing [this](https://github.com/huggingface/diffusers/actions/runs/16109620596/job/45450689391?pr=11876#step:8:3354):

```
FAILED tests/pipelines/controlnet/test_controlnet_blip_diffusion.py::BlipDiffusionControlNetPipelineFastTests::test_attention_slicing_forward_pass - TypeError: CLIPEncoder.forward() got an unexpected keyword argument 'return_dict'
```

But in retrospect, we should deprecate the BLIP ControlNet pipeline anyway. This PR does that.

### DIFF
--- a/src/diffusers/pipelines/controlnet/pipeline_controlnet_blip_diffusion.py
+++ b/src/diffusers/pipelines/controlnet/pipeline_controlnet_blip_diffusion.py
@@ -116,6 +116,7 @@ class BlipDiffusionControlNetPipeline(DeprecatedPipelineMixin, DiffusionPipeline
             Position of the context token in the text encoder.
     """
 
+    _last_supported_version = "0.33.1"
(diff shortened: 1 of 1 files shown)


DN6 (approved):

DN6:
Just need to add `last_supported_version` to the class
https://github.com/huggingface/diffusers/blob/425a715e35479338c06b2a68eb3a95790c1db3c5/src/diffusers/pipelines/pia/pipeline_pia.py#L144

sayakpaul:
Thanks. In general, do you think we should get rid of the tests for the deprecated pipelines? Will reduce LoC. @DN6

DN6:
We can remove the pipeline tests, yes

PR #11881, opened 2025-07-08
[refactor] Flux single file implementation

a-r-r-o-w:
Question: should we remove all intermediate imports of norm/embed layers, implement within the file, and make the entire implementation truly single file?

The minimal flux inference code does not run yet. The code from upstream branch is not fully completed and type hints are done with classes that don't exist.


src/diffusers/models/embeddings.py
```diff
@@ -1244,31 +1244,21 @@ class FluxPosEmbed(nn.Module):
     # modified from https://github.com/black-forest-labs/flux/blob/c00d7c60b085fce8058b9df845e036090873f2ce/src/flux/modules/layers.py#L11
     def __init__(self, theta: int, axes_dim: List[int]):
         super().__init__()
+
```
DN6, line 1247: Think here you can patch `__new__` with the correct import and return the new object. Similar to 
https://github.com/huggingface/diffusers/blob/c87575dde6b5567b7914c39280bb9a2587b33f66/src/diffusers/models/attention_processor.py#L3829

Issue #11828, opened 2025-06-29
StableDiffusionXLControlNetUnionInpaintPipeline requires a float for controlnet_conditioning_scale & control_guidance_end


ewwwgiddings:
### Summary

StableDiffusionXLControlNetUnionInpaintPipeline requires a float for `controlnet_conditioning_scale` & `control_guidance_end` when they should also accept a `List[float]`. 

After PR #10723 the txt-to-img Union pipeline `pipeline_controlnet_union_sd_xl.py` accepts a list/tuple so each active ControlNet branch can have its own conditioning scale. The in-paint counterpart `pipeline_controlnet_union_inpaint_sd_xl.py` still contains the old check:

```
elif isinstance(self.controlnet, ControlNetUnionModel):
    if not isinstance(controlnet_conditioning_scale, float):
        raise TypeError(
            "For single controlnet: `controlnet_conditioning_scale` must be type `float`."
        )
```

Passing a list raises `TypeError: For single controlnet: `controlnet_conditioning_scale` must be type `float`.`

It was said that it would be added in that PR: https://github.com/huggingface/diffusers/pull/10723#issuecomment-2655309539

I don't see any mention of the `control_guidance_end` behavior in any PR's so maybe it was missed.

### Expected Behaviour

- Accept List[float] exactly like the txt-to-img Union pipeline. I'm guessing it could be added almost exactly like this https://github.com/huggingface/diffusers/pull/10723/files
- `control_guidance_end` should also be able to accept a `List[float]`

### Reproduction

```
from diffusers import StableDiffusionXLControlNetUnionInpaintPipeline, ControlNetUnionModel, AutoencoderKL
from diffusers.utils import load_image
import torch
import numpy as np
from PIL import Image

prompt = "A cat"
# download an image
image = load_image(
    "https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/in_paint/overture-creations-5sI6fQgYIuo.png"
).resize((1024, 1024))
mask = load_image(
    "https://huggingface.co/datasets/hf-internal-testing/diffusers-images/resolve/main/in_paint/overture-creations-5sI6fQgYIuo_mask.png"
).resize((1024, 1024))
# initialize the models and pipeline
controlnet = ControlNetUnionModel.from_pretrained(
    "brad-twinkl/controlnet-union-sdxl-1.0-promax", torch_dtype=torch.float16
)
[...]

PR #11888, opened 2025-07-08
[ControlnetUnion] Multiple Fixes

### LINKED ISSUE #11861: ControlnetUnion not working correctly
### Describe the bug

After https://github.com/huggingface/diffusers/pull/10723 the way controlnet union works with the control images is not correct.

The more clear and concise way of showing this is using a depth map with the tile condition, the tile condition should return almost the same image but in this case it return an image as if it was using a depth map.

I noticed that with this issue, the inpainting also suffers a lot and mostly generates subpar inpainted images compared to the original implementation or how it was before that PR.

### Reproduction

Using this code before and after the relevant PR.

```python
import torch

from diffusers import ControlNetUnionModel, StableDiffusionXLControlNetUnionPipeline


torch_dtype = torch.float16
prompt = "A cat"

control_image = load_image(
    "https://huggingface.co/datasets/OzzyGT/testing-resources/resolve/main/resources/cat_depthmap.png"
)

controlnet_model = ControlNetUnionModel.from_pretrained(
    "OzzyGT/controlnet-union-promax-sdxl-1.0", variant="fp16", torch_dtype=torch_dtype
)

pipe = StableDiffusionXLControlNetUnionPipeline.from_pretrained(
    variant="fp16",
    controlnet=controlnet_model,
    torch_dtype=torch_dtype,
)

pipe.to("cuda")

image = pipe(
    prompt=prompt,
    guidance_scale=5.0,
    num_inference_steps=20,
    control_image=control_image,
    controlnet_conditioning_scale=0.5,
    control_mode=6,
).images[0]

image.save("controlnet_union_test.png")
```

|Before|After|
|---|---|
|Image|Image|

### Who can help?

I'm opening this issue to the community since I don't have time right now, if no one wants to tackle it, I'll do it when I have time.

asomoza:
# What does this PR do?

This PR fixes the problems in the related issues, also it surprises me that no one have ever used this controlnet without cfg because it doesn't work, this PR also fixes that.

fixes #11861 #11828

## Who can review?

@yiyixuxu


src/diffusers/pipelines/controlnet/pipeline_controlnet_union_inpaint_sd_xl.py
```diff
@@ -240,8 +247,8 @@ def __init__(
     ):
         super().__init__()
 
-        if not isinstance(controlnet, ControlNetUnionModel):
-            raise ValueError("Expected `controlnet` to be of type `ControlNetUnionModel`.")
+        if isinstance(controlnet, (list, tuple)):
```
yiyixuxu, line 250: ohh, so you would need to use multiple models too for controlnet union?

asomoza: yeah this was added in this [PR](https://github.com/huggingface/diffusers/pull/10747) because of this [comment](https://github.com/huggingface/diffusers/issues/10656#issuecomment-2619019105) 

Basically is for people that want to control the start and the end for each condition, if you don't need that, you can use one controlnet union with multiple conditions.

Also is to make it the same as other controlnets so people don't have to code a special case for just this one in the UIs.

yiyixuxu (approved):
thanks!

PR #11904, opened 2025-07-10
Speedup model loading by 4-5x ⚡

a-r-r-o-w:
All thanks to @Cyrilvallez's PR: https://github.com/huggingface/transformers/pull/36380

The accelerate PR is required because we end up calling `clear_device_cache` in a loop (over the sharded files). This is bad. Without this, you'll see no speedup.

Another small optimization is using non_blocking everywhere and syncing just before returning control to the user. This is slightly faster.

```python
import time
t_ini = time.time()

import torch
from diffusers import FluxPipeline, FluxTransformer2DModel
print(f"import time: {time.time() - t_ini:.3f}s")

t0 = time.time()
torch.cuda.synchronize()
print(f"CUDA sync time: {time.time() - t0:.3f}s")

print("starting model load")
t1 = time.time()
transformer = FluxTransformer2DModel.from_pretrained(model_id, subfolder="transformer", torch_dtype=torch.bfloat16, device_map="cuda")
torch.cuda.synchronize()
t2 = time.time()

diff = t2 - t1
print(f"time: {diff:.3f}s")

pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", transformer=transformer, torch_dtype=torch.bfloat16)
pipe.text_encoder.to("cuda")
pipe.text_encoder_2.to("cuda")
pipe.vae.to("cuda")
image = pipe(prompt, num_inference_steps=28, guidance_scale=4.0).images[0]
image.save("flux.png")
```

Sister PR in accelerate required to obtain speedup: https://github.com/huggingface/accelerate/pull/3674

- On main: `16.765s`
- On this branch: `4.521s`


src/diffusers/loaders/single_file_model.py
```diff
@@ -430,6 +431,8 @@ def from_single_file(cls, pretrained_model_link_or_path_or_dict: Optional[str] =
                 keep_in_fp32_modules=keep_in_fp32_modules,
                 unexpected_keys=unexpected_keys,
             )
+            empty_device_cache()
+            device_synchronize()
```
sayakpaul, line 437: Maybe having a comment saying why we call before the rest of the code would be helpful.

src/diffusers/loaders/single_file_utils.py
```diff
@@ -1689,6 +1690,8 @@ def create_diffusers_clip_model_from_ldm(
 
     if is_accelerate_available():
         load_model_dict_into_meta(model, diffusers_format_checkpoint, dtype=torch_dtype)
+        empty_device_cache()
+        device_synchronize()
```
sayakpaul, line 1696: Same.

src/diffusers/models/model_loading_utils.py
```diff
@@ -520,3 +526,64 @@ def load_gguf_checkpoint(gguf_checkpoint_path, return_tensors=False):
         parsed_parameters[name] = GGUFParameter(weights, quant_type=quant_type) if is_gguf_quant else weights
 
     return parsed_parameters
+
+
+def _find_mismatched_keys(
```
sayakpaul: Taken out of here:
https://github.com/huggingface/diffusers/blob/9f4d997d8ffabe455c4f743139d3b5d5039339da/src/diffusers/models/modeling_utils.py#L1509

src/diffusers/models/model_loading_utils.py
```diff
@@ -520,3 +526,64 @@ def load_gguf_checkpoint(gguf_checkpoint_path, return_tensors=False):
...
+    new_device_map = {}
+    for module, device in device_map.items():
+        new_device_map.update(
+            {p: device for p in param_names if p == module or p.startswith(f"{module}.") or module == ""}
+        )
+    return new_device_map
+
+
+# Adapted from: https://github.com/huggingface/transformers/blob/0687d481e2c71544501ef9cb3eef795a6e79b1de/src/transformers/modeling_utils.py#L5859
+def _caching_allocator_warmup(model, expanded_device_map: Dict[str, torch.device], dtype: torch.dtype) -> None:
```
sayakpaul, line 562: Should we account for `hf_quantizer` too?
https://github.com/huggingface/transformers/blob/c6ee0b1da8ff57102548430e18480fa78a106022/src/transformers/modeling_utils.py#L5862

src/diffusers/models/modeling_utils.py
```diff
@@ -1482,18 +1481,21 @@ def _load_pretrained_model(
                     " for them. Alternatively, make sure you have `safetensors` installed if the model you are using"
                     " offers the weights in this format."
                 )
-            if offload_folder is not None:
+            else:
                 os.makedirs(offload_folder, exist_ok=True)
             if offload_state_dict is None:
                 offload_state_dict = True
 
+        # Caching allocator warmup
```
sayakpaul: Expanding the comment to discuss why we only warm up `if device_map is not None` would be helpful.

src/diffusers/utils/torch_utils.py
```diff
@@ -182,5 +182,14 @@ def get_device():
...
+        return
     device_mod = getattr(torch, device_type, torch.cuda)
     device_mod.empty_cache()
+
+
+def device_synchronize(device_type: Optional[str] = None):
+    if device_type is None:
+        device_type = get_device()
+    device_mod = getattr(torch, device_type, torch.cuda)
+    device_mod.synchronize()
```
sayakpaul, line 197: I guess all different backends ought to have this method. Just flagging.

sayakpaul (commented):
Thanks a lot for quickly getting this up 🔥

My comments are mostly minor, the major one being adding `hf_quantizer` to the allocator function.

Additionally, for a potentially better user-experience, if we could try to rethink the `to()` method of `DiffusionPipeline`, it would be helpful. I mean the following.

Currently, from what I understand, we have to first initialize the denoiser using `device_map` and then the rest of the components. If a user is calling `.to()` on a `DiffusionPipeline`, we could consider using `device_map="cuda"` for dispatching the model-level components to CUDA. I don't immediately see a downside to it.

src/diffusers/utils/torch_utils.py
a-r-r-o-w: afaik, synchronize should be available on all devices. Just the empty_cache function required a special check because it would fail if device was cpu

src/diffusers/models/model_loading_utils.py
a-r-r-o-w: I will add some code to ignore caching allocator warmup if quantizer is present for the time being. I started looking into this because I was facing some terrible load times on some compute providers I was testing on, and so would prefer to have just the bf16 load fast for now. Quantizers can come in a follow-up if you're interested in taking that up, but atm I have limited bandwidth to test that case and add support

sayakpaul: Sure, I will do it.

a-r-r-o-w: Updated. If quantizer is not None, allocator warmup will be ignored

src/diffusers/models/modeling_utils.py
a-r-r-o-w: added comment

src/diffusers/loaders/single_file_model.py
a-r-r-o-w: added comment

sayakpaul (approved):
Ship

src/diffusers/models/model_loading_utils.py
```diff
@@ -252,6 +254,10 @@ def load_model_dict_into_meta(
                 param = param.to(dtype)
                 set_module_kwargs["dtype"] = dtype
 
+        if is_accelerate_version(">=", "1.9.0.dev0"):
```
SunMarc:
```suggestion
        if is_accelerate_version(">", "1.8.0"):
```
This way we don't need to put dev. I'll also do a accelerate release next week !

SunMarc (approved):
Thanks for this ! Just a nit

src/diffusers/models/model_loading_utils.py
```diff
@@ -252,6 +254,10 @@ def load_model_dict_into_meta(
                 param = param.to(dtype)
                 set_module_kwargs["dtype"] = dtype
 
+        if is_accelerate_version(">", "1.8.0"):
```
a-r-r-o-w:
```suggestion
        if is_accelerate_version(">", "1.8.1"):
```

PR #11905, opened 2025-07-10
[utils] account for MPS when available in get_device().

sayakpaul:


src/diffusers/utils/torch_utils.py
```diff
@@ -175,6 +175,8 @@ def get_device():
         return "npu"
     elif hasattr(torch, "xpu") and torch.xpu.is_available():
         return "xpu"
+    elif torch.backends.mps.is_available() and hasattr(torch, "mps"):
```
a-r-r-o-w: not sure why we need to check hasattr part here if the first part is true?

PR #11920, opened 2025-07-14
[tests] Add test slices for Wan

a-r-r-o-w:


a-r-r-o-w:
Gentle ping @DN6

tests/pipelines/wan/test_wan_video_to_video.py
```diff
@@ -123,11 +122,15 @@ def test_inference(self):
...
-        expected_video = torch.randn(17, 3, 16, 16)
-        max_diff = np.abs(generated_video - expected_video).max()
-        self.assertLessEqual(max_diff, 1e10)
+
+        # fmt: off
+        expected_slice = torch.tensor([0.4522, 0.4534, 0.4532, 0.4553, 0.4526, 0.4538, 0.4533, 0.4547, 0.513, 0.5176, 0.5286, 0.4958, 0.4955, 0.5381, 0.5154, 0.5195])
+        # fmt:on
+
+        generated_slice = generated_video.flatten()
+        generated_slice = torch.cat([generated_slice[:8], generated_slice[-8:]])
```
sayakpaul, line 132: Curious: why do we have to slice and cat?

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -247,3 +268,32 @@ def get_dummy_inputs(self, device, seed=0):
             "output_type": "pt",
         }
         return inputs
+
+    def test_inference(self):
+        device = "cpu"
+
+        components = self.get_dummy_components()
+        pipe = self.pipeline_class(**components)
+        pipe.to(device)
```
sayakpaul, line 277: Not important for this PR. But at some point, I think we should switch to `torch_device` instead of fixing devices like this. I see downsides and upsides:

* Upside: same code can be used for quickly checking if the `pipe` can be run on different accelerators (depending on what we get for `torch_device`.
* Downside: The `expected_slice` will vary but I guess we can leverage a combination of cosine similarity based checks, enforcement of determinism (like [this](https://github.com/huggingface/diffusers/blob/ef1e6287291a945665b0c43f4863228b8e9131df/src/diffusers/utils/testing_utils.py#L1129)).

No need to fret for this PR. We will get to it later.

sayakpaul (approved):

tests/pipelines/wan/test_wan_video_to_video.py
a-r-r-o-w: Just a convention I've been following in most of my test PRs. It checks some start and end values of the tensor for correctness, and I think that's better than just checking at one side (for example, sometimes when writing triton, part of the logic is correct and it will work correctly for most of the tensor, but if the load masks and default value are not correct, the values at the end of tensor may be computed incorrectly. By just checking the start or end of the tensor, you may get a false impression that the algorithm is correct. I think I picked this way of checking slices from triton unit tests or somewhere similar, i don't remember anymore)

a-r-r-o-w:
Failing tests are unrelated

tests/pipelines/wan/test_wan_video_to_video.py
sayakpaul: Wow, that is indeed better. Thanks for the note! I will adapt this.

PR #11936, opened 2025-07-15
[docs] Update toctree

stevhliu:
* collapses some sections to make the docs feel less overwhelming
* reorganizes some of the sections according to topics (for example, there is no longer an explicit "tutorial" section)


docs/source/en/_toctree.yml
```diff
@@ -1,36 +1,39 @@
-- sections:
+- title: Get started
```
sayakpaul, line 1: So, in getting started, we're including installation, effective and efficient diffusion, and others. What are we discarding?

docs/source/en/_toctree.yml
```diff
@@ -1,36 +1,39 @@
...
-  - local: using-diffusers/write_own_pipeline
-    title: Understanding pipelines, models and schedulers
-  - local: tutorials/autopipeline
-    title: AutoPipeline
-  - local: tutorials/basic_training
-    title: Train a diffusion model
-  title: Tutorials
-- sections:
+
+- title: Pipeline
```
sayakpaul: Should we call it `DiffusionPipeline`?

I think we should really take the step to make people understand the beauty behind abstracting the core logic of how different components are connected. Could be a good opportunity in this PR to do so, if that's not already evident enough. WDYT? I created the following GIF for a presentation of mine which could be reused.

generation_pipeline

docs/source/en/_toctree.yml
```diff
@@ -43,25 +46,12 @@
...
-    title: Text-to-image
-  - local: using-diffusers/img2img
-    title: Image-to-image
-  - local: using-diffusers/inpaint
-    title: Inpainting
-  - local: using-diffusers/text-img2vid
-    title: Video generation
-  - local: using-diffusers/depth2img
-    title: Depth-to-image
-  title: Generative tasks
```
sayakpaul, line 61: Where are these going?

docs/source/en/_toctree.yml
```diff
@@ -76,14 +66,34 @@
...
+  - local: optimization/tome
+    title: Token merging
+  - local: optimization/deepcache
+    title: DeepCache
+  - local: optimization/tgate
+    title: TGATE
+  - local: optimization/xdit
+    title: xDiT
+  - local: optimization/para_attn
+    title: ParaAttention
```
sayakpaul: WDYT about clubbing these pages under "third party"?

sayakpaul (commented):
Looking really nice!

docs/source/en/_toctree.yml
stevhliu: Nothing is being discarded here, it should all be the same :)

I think in a different PR though, we could update the "effective and efficient diffusion" though though

docs/source/en/_toctree.yml
stevhliu: Sounds good to me!

docs/source/en/_toctree.yml
stevhliu: These are included in a "Task recipes" [section](https://moon-ci-docs.huggingface.co/docs/diffusers/pr_11936/en/using-diffusers/unconditional_image_generation) within "Resources"

docs/source/en/_toctree.yml
stevhliu: Nested them in a sub-section under "Inference Optimization"

docs/source/en/_toctree.yml
sayakpaul: Sure.

sayakpaul (approved):
Lovely!

PR #11947, opened 2025-07-17
[docs] clarify the mapping between `Transformer2DModel` and finegrained variants.

sayakpaul:
# What does this PR do?

Our docstrings for PixArt Alpha say that the `transformer` is a `PixArtTransformer2DModel` type but the [configuration of the pre-trained checkpoint](https://huggingface.co/PixArt-alpha/PixArt-XL-2-1024-MS/blob/main/transformer/config.json#L2) says it's `Transformer2DModel`. This PR clarifies this confusion.

Cc: @dasayan05

### DIFF
--- a/src/diffusers/pipelines/dit/pipeline_dit.py
+++ b/src/diffusers/pipelines/dit/pipeline_dit.py
@@ -46,8 +46,9 @@ class DiTPipeline(DiffusionPipeline):
 
     Parameters:
         transformer ([`DiTTransformer2DModel`]):
-            A class conditioned `DiTTransformer2DModel` to denoise the encoded image latents. Initially
-            published as [Transformer2DModel](https://huggingface.co/facebook/DiT-XL-2-256/blob/main/transformer/config.json#L2) in the config, but the mismatch can be ignored.
+            A class conditioned `DiTTransformer2DModel` to denoise the encoded image latents. Initially published as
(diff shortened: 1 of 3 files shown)


src/diffusers/pipelines/dit/pipeline_dit.py
```diff
@@ -46,7 +46,10 @@ class DiTPipeline(DiffusionPipeline):
 
     Parameters:
         transformer ([`DiTTransformer2DModel`]):
-            A class conditioned `DiTTransformer2DModel` to denoise the encoded image latents.
+            A class conditioned `DiTTransformer2DModel` to denoise the encoded image latents. It was initially
+            published as `Transformer2DModel` which is why [the
+            configuration](https://huggingface.co/facebook/DiT-XL-2-256/blob/main/transformer/config.json#L2) still
+            shows the class name as `Transformer2DModel`. This mismatch can be safely ignored.
```
stevhliu: I would update with this shorter version

```suggestion
            A class conditioned `DiTTransformer2DModel` to denoise the encoded image latents. Initially
            published as [Transformer2DModel](https://huggingface.co/facebook/DiT-XL-2-256/blob/main/transformer/config.json#L2) in the config, but the mismatch can be ignored.
```

stevhliu (approved):
Thanks for clarifying!

sayakpaul:
Failing tests are unrelated.

PR #11963, opened 2025-07-21
[Modular] update the collection behavior

yiyixuxu:
We guarantee unique component names under each collection, i.e. for each collection, there is only one "unet", one "vae" .... 

Currently, when a new component enters the collection,  we remove the existing one
e.g. node1 has a "unet" that's sdxl default, when we add a new unet from node1, say "CyberRealisticXL", the sdxl default unet will be removed from the components manager

This is wrong, and this PR changes the behavior so that we simply dissociate the existing from the collection, and won't delete it if the component is still associated with a different collection


asomoza:
LGTM, thanks!

asomoza (approved):

PR #11941, opened 2025-07-16
Context Parallel w/ Ring & Ulysses & Unified Attention

a-r-r-o-w:
Adds support for ring, ulysses and unified attention natively. For a minimal PoC, I've limited changes to Flux.

Supported attention backends with CP: cuDNN, FA2, Sage.

Requires #11916 to be merged first.

### Minimal example

```python
import torch
from diffusers import FluxPipeline
from diffusers import ParallelConfig, enable_parallelism

try:
    torch.distributed.init_process_group("nccl")
    rank = torch.distributed.get_rank()
    device = torch.device("cuda", rank % torch.cuda.device_count())
    torch.cuda.set_device(device)

    pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", torch_dtype=torch.bfloat16)
    pipe.to(device)
    pipe.transformer.parallelize(config=ParallelConfig(ulysses_degree=2))
    # pipe.transformer.set_attention_backend("_native_cudnn")
    pipe.transformer.set_attention_backend("flash")
    # pipe.transformer.compile(fullgraph=True)
    # pipe.transformer.compile(fullgraph=True, mode="max-autotune")

    prompt = "A cat holding a sign that says 'hello world'"
    
    # Must specify generator so all ranks start with same latents (or pass your own)
    generator = torch.Generator().manual_seed(42)
    with enable_parallelism(pipe):
        image = pipe(prompt, num_inference_steps=2, guidance_scale=4.0, generator=generator).images[0]
        image = pipe(prompt, num_inference_steps=28, guidance_scale=4.0, generator=generator).images[0]
    
    if rank == 0:
        image.save("output.png")

except Exception as e:
    print(f"An error occurred: {e}")
    torch.distributed.breakpoint()
    raise

finally:
    if torch.distributed.is_initialized():
        torch.distributed.destroy_process_group()
```

<details>
<summary> Wan </summary>

```python
import torch
from diffusers import AutoencoderKLWan, WanPipeline, ParallelConfig, enable_parallelism
from diffusers.utils import export_to_video
from diffusers.schedulers.scheduling_unipc_multistep import UniPCMultistepScheduler

try:
    torch.distributed.init_process_group("nccl")
    rank = torch.distributed.get_rank()
    device = torch.device("cuda", rank % torch.cuda.device_count())
    torch.cuda.set_device(device)
[...]

PR #11967, opened 2025-07-21
Remove logger warnings for attention backends and hard error during runtime instead

### LINKED ISSUE #11959: Passing enable_gqa flag in attention dispatcher should be version guarded
Just realized this is potentially breaking: https://github.com/huggingface/diffusers/blob/cde02b061b6f13012dfefe76bc8abf5e6ec6d3f3/src/diffusers/models/attention_dispatch.py#L229

I believe enable_gqa was first released in PT 2.5.0. For versions before that, this flag should not be passed. Additionally, some user code may have wrappers around SDPA and do not expect the enable_gqa flag (think of TorchFunctionMode being used to intercept call).

a-r-r-o-w:
- Fixes #11959
- Instead of raising warning with logger and erroring out with a NoneType exception, we now check the imports during runtime and error out instead

```python
import torch

).to("cuda")

image = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/flux_ip_adapter_input.jpg").resize((1024, 1024))

pipe.load_ip_adapter(
    "XLabs-AI/flux-ip-adapter",
    weight_name="ip_adapter.safetensors",
    image_encoder_pretrained_model_name_or_path="openai/clip-vit-large-patch14"
)
pipe.set_ip_adapter_scale(1.0)
pipe.transformer.set_attention_backend("_flash_3") # errors out if you don't have FA3
pipe.transformer.set_attention_backend("sage") # errors out if you don't have sage 

image = pipe(
    width=1024,
    height=1024,
    prompt="wearing sunglasses",
    negative_prompt="",
    true_cfg_scale=4.0,
    generator=torch.Generator().manual_seed(4444),
    ip_adapter_image=image,
).images[0]

image.save('output.png')
```

### DIFF
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -187,12 +187,15 @@ class _AttentionBackendRegistry:
 
 
 @contextlib.contextmanager
-def attention_backend(backend: AttentionBackendName = AttentionBackendName.NATIVE):
+def attention_backend(backend: Union[str, AttentionBackendName] = AttentionBackendName.NATIVE):
     """
     Context manager to set the active attention backend.
     """
     if backend not in _AttentionBackendRegistry._backends:
         raise ValueError(f"Backend {backend} is not registered.")
+
+    backend = AttentionBackendName(backend)
+    _check_attention_backend_requirements(backend)
 
     old_backend = _AttentionBackendRegistry._active_backend
     _AttentionBackendRegistry._active_backend = backend
@@ -225,8 +228,6 @@ def dispatch_attention_fn(
     else:
         backend_name = AttentionBackendName(backend)
         backend_fn = _AttentionBackendRegistry._backends.get(backend_name)
-
-    _check_backend_requirements(backend_name)
 
     kwargs = {
         "query": query,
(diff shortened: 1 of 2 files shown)


src/diffusers/models/modeling_utils.py
```diff
@@ -627,6 +627,8 @@ def set_attention_backend(self, backend: str) -> None:
         # TODO: the following will not be required when everything is refactored to AttentionModuleMixin
         from .attention_processor import Attention, MochiAttention
 
+        logger.warning("Attention backends are an experimental feature and the API may be subject to change.")
```
a-r-r-o-w, line 630: cc @DN6 Added this here because things are probably going to moving a little too fast for now. #11941 introduces several changes to simplify the interface for operating with the templated implementations. For custom parameters that the backends can accept, it is very easy for a user to register a custom dispatch function (i.e. we won't support all parameter customizability on our end and just use the defaults)

src/diffusers/models/attention_dispatch.py
```diff
@@ -218,6 +226,8 @@ def dispatch_attention_fn(
         backend_name = AttentionBackendName(backend)
         backend_fn = _AttentionBackendRegistry._backends.get(backend_name)
 
+    _check_backend_requirements(backend_name)
```
DN6: Can't we add this check to `set_attention_backend` in ModelMixin? Then we wouldn't need to add the cache?

src/diffusers/models/attention_dispatch.py
```diff
@@ -305,6 +316,60 @@ def _check_shape(
 # ===== Helper functions =====
 
 
+# LRU cache is hack to avoid checking the backend requirements multiple times. Maybe not needed
+# because CPU is running much farther ahead of the accelerator and this will not be blocking anyway.
+@functools.lru_cache(maxsize=16)
+def _check_backend_requirements(backend: AttentionBackendName) -> None:
```
DN6:
```suggestion
def _check_attention_backend_requirements(backend: AttentionBackendName) -> None:
```

src/diffusers/models/attention_dispatch.py
yiyixuxu: +1 on this

a-r-r-o-w: nice yes, done!

DN6 (approved):
Thank you 🙏🏽

a-r-r-o-w:
Failing test is unrelated

PR #11979, opened 2025-07-23
[tests] enforce torch version in the compilation tests.

sayakpaul:
# What does this PR do?

Should enforce this for 
https://github.com/huggingface/diffusers/blob/7ae6347e330a2b12da0b563370b6605a18bf0ed2/tests/quantization/test_torch_compile_utils.py#L26

But I will open a separate PR for that.


sayakpaul:
Failing tests are unrelated.

PR #11987, opened 2025-07-24
[compile] logger statements create unnecessary guards during dynamo tracing

a-r-r-o-w:
If compiling with `fullgraph=True`, we first need to remove the `torch.compiler.disable` calls we have. Once this is done, Dynamo will error out on the logger statements because they can't be traced. We can conditionally log them to fix this error (what this PR does). We will then hit an error because of using context manager for pinning tensors. We can move the body of that function into the functions directly to avoid using context managers. We will then error out on the call to `tensor.is_pinned()` because Dynamo cannot trace it, which seems like a problem that needs to be fixed on pytorch-side.

If compiling with `fullgraph=False`, we add a lot of unnecessary guards and have some more graph breaks. By only conditionally doing the logging, we can avoid some of them.

See this comment: https://github.com/huggingface/diffusers/pull/11960#issuecomment-3114761867

### DIFF
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -367,7 +367,8 @@ class LazyPrefetchGroupOffloadingHook(ModelHook):
     def initialize_hook(self, module):
         def make_execution_order_update_callback(current_name, current_submodule):
             def callback():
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):
Thanks!

Issue #11981, opened 2025-07-23
Groupoffloading introduce bad results


seed93:
### Describe the bug

I am using groupoffloading for saving gpu memory. I got worse results with a cosine similarity aboud 0.934 on A800, which is unexpected. And I got results with a cosine similarity about 0.78 on 4090, which is worse. 
Could anyone give me any suggestions to fix the precision?

### Reproduction

```
apply_group_offloading(
    transformer,
    onload_device=torch.device(f"cuda:{self.local_rank}"),
    offload_device=torch.device("cpu"),
    offload_type="block_level",
    num_blocks_per_group=1,
    non_blocking=True,
    use_stream=True,
)

### Logs

```shell

```

### System Info

I tried diffusers 0.33.1 and 0.34.

a-r-r-o-w:
Hi @seed93 , we've come across different/worse results in atleast one other case (reported by Sayak from the team) with block level offloading. It is most likely related to faulty synchronisation of streams, but we haven't been able to identify the issue yet. Ideally, group offloading will be lossless once the potential bug is fixed 

Do you notice any problems with leaf level offloading?

seed93:
> Hi [@seed93](https://github.com/seed93) , we've come across different/worse results in atleast one other case (reported by Sayak from the team) with block level offloading. It is most likely related to faulty synchronisation of streams, but we haven't been able to identify the issue yet. Ideally, group offloading will be lossless once the potential bug is fixed
> 
> Do you notice any problems with leaf level offloading?

Thank you for your quick reply. I tried leaf_level on a800 just minites ago and it worked good.

seed93:
One more thing, I failed to run leaf_level on 4090 with an error like this:
```
Failed running call_function <method 'set_' of 'torch._C.TensorBase' objects>(*(Parameter(FakeTensor(..., device='cuda:0', size=(4096, 4096), dtype=torch.bfloat16,
           requires_grad=True)), FakeTensor(..., size=(4096, 4096), dtype=torch.bfloat16)), **{}):
Unhandled FakeTensor Device Propagation for aten.set_.source_Tensor, found two different devices cuda:0, cpu

from user code:
   File "/usr/local/lib/python3.12/dist-packages/diffusers/hooks/hooks.py", line 149, in torch_dynamo_resume_in_new_forward_at_147
    return function_reference.post_forward(module, output)
[...]

PR #11990, opened 2025-07-25
[refactor] condense group offloading 

a-r-r-o-w:
The current implementation after many of the recent updates is hard to understand or debug/reason through. There is also one code path that seems completely unused (see the removed `self.offload_to_disk_path` related changes and the call into `self._onload_from_disk`).

This PR tries to refactor and clean up some of the implementation so that implementing new changes is easier in the future.

Related to [comment](https://github.com/huggingface/diffusers/issues/11981#issuecomment-3113679476) which I'm trying to debug through

### DIFF
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -101,7 +101,7 @@ class ModuleGroup:
         self.offload_to_disk_path = offload_to_disk_path
         self._is_offloaded_to_disk = False
 
-        if self.offload_to_disk_path:
+        if self.offload_to_disk_path is not None:
             # Instead of `group_id or str(id(self))` we do this because `group_id` can be "" as well.
             self.group_id = group_id if group_id is not None else str(id(self))
             short_hash = _compute_group_hash(self.group_id)
@@ -120,6 +120,12 @@ class ModuleGroup:
             self.cpu_param_dict = {}
         else:
             self.cpu_param_dict = self._init_cpu_param_dict()
+
+        self._torch_accelerator_module = (
+            getattr(torch, torch.accelerator.current_accelerator().type)
+            if hasattr(torch, "accelerator")
+            else torch.cuda
+        )
 
     def _init_cpu_param_dict(self):
         cpu_param_dict = {}
@@ -144,16 +150,12 @@ class ModuleGroup:
 
     @contextmanager
     def _pinned_memory_tensors(self):
-        pinned_dict = {}
         try:
-            for param, tensor in self.cpu_param_dict.items():
-                if not tensor.is_pinned():
-                    pinned_dict[param] = tensor.pin_memory()
-                else:
-                    pinned_dict[param] = tensor
-
+            pinned_dict = {
+                param: tensor.pin_memory() if not tensor.is_pinned() else tensor
+                for param, tensor in self.cpu_param_dict.items()
+            }
             yield pinned_dict
-
         finally:
             pinned_dict = None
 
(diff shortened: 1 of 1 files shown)


src/diffusers/hooks/group_offloading.py
```diff
@@ -138,112 +144,76 @@ def _init_cpu_param_dict(self):
...
+            pinned_dict = {
+                param: tensor.pin_memory() if not tensor.is_pinned() else tensor
+                for param, tensor in self.cpu_param_dict.items()
+            }
             yield pinned_dict
-
         finally:
             pinned_dict = None
 
-    def _transfer_tensor_to_device(self, tensor, source_tensor, current_stream=None):
```
sayakpaul, line 154: Very clean!

src/diffusers/hooks/group_offloading.py
```diff
@@ -138,112 +144,76 @@ def _init_cpu_param_dict(self):
...
-            else torch.cuda
-        )
-        context = nullcontext() if self.stream is None else torch_accelerator_module.stream(self.stream)
-        current_stream = torch_accelerator_module.current_stream() if self.record_stream else None
+        with context:
+            # Load to CPU (if using streams) or directly to target device, pin, and async copy to device
+            device = self.onload_device if self.stream is None else "cpu"
+            loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device=device)
 
-        if self.offload_to_disk_path:
```
sayakpaul, line 214: Perfect!

src/diffusers/hooks/group_offloading.py
```diff
@@ -138,112 +144,76 @@ def _init_cpu_param_dict(self):
...
-            getattr(torch, torch.accelerator.current_accelerator().type)
-            if hasattr(torch, "accelerator")
-            else torch.cuda
-        )
-        context = nullcontext() if self.stream is None else torch_accelerator_module.stream(self.stream)
-        current_stream = torch_accelerator_module.current_stream() if self.record_stream else None
+        with context:
+            # Load to CPU (if using streams) or directly to target device, pin, and async copy to device
+            device = self.onload_device if self.stream is None else "cpu"
+            loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device=device)
```
sayakpaul, line 189: `device` when supplied as `torch.device("cuda")` to the `onload_device` argument, it would fail here from `safetensors` complaining `invalid device cuda`. Simply wrapping up it within `str()` solves the issue.

sayakpaul (approved):
Looks super clean! 

I just ran some benchmarks to confirm if the changes don't have any detrimental effect on the speed-memory trade-off ([code](https://gist.github.com/sayakpaul/c18008cf3ccecbb7c99ce17350a18f1c)) and they look alright to me.

Would maybe also run all the group offloading tests on the GPU, too.

DN6 (approved):
Nice 👍🏽

src/diffusers/hooks/group_offloading.py
a-r-r-o-w: updated!

a-r-r-o-w:
@sayakpaul Could you take another look? added a test as well

tests/hooks/test_group_offloading.py
```diff
@@ -274,3 +310,54 @@ def test_block_level_stream_with_invocation_order_different_from_initialization_
 
         with context:
             model(self.input)
+
+    @parameterized.expand([("block_level",), ("leaf_level",)])
+    def test_block_level_offloading_with_parameter_only_module_group(self, offload_type: str):
+        if torch.device(torch_device).type not in ["cuda", "xpu"]:
+            return
```
sayakpaul, line 317: We could use `require_torch_accelerator()` here.

sayakpaul (approved):
🔥

seed93:
This PR will cause wrong result with Flux.1-dev/kontext. Will anyone have a look? @a-r-r-o-w

PR #11997, opened 2025-07-27
[modular diffusers] Wan I2V/FLF2V

a-r-r-o-w:


src/diffusers/modular_pipelines/wan/encoders.py
```diff
@@ -240,3 +257,238 @@ def __call__(self, components: WanModularPipeline, state: PipelineState) -> Pipe
...
+                required=False,
+                description="The last image to condition the generation on for last-frame conditioned video generation.",
+            ),
+        ]
+
+    @property
+    def intermediate_outputs(self) -> List[OutputParam]:
+        return [
+            OutputParam(
+                "encoder_hidden_states_image",
```
yiyixuxu: this should be called `image_embeds`, no?

src/diffusers/modular_pipelines/wan/encoders.py
```diff
@@ -240,3 +257,238 @@ def __call__(self, components: WanModularPipeline, state: PipelineState) -> Pipe
...
+            image = [block_state.image, block_state.last_image]
+
+        block_state.encoder_hidden_states_image = self.encode_image(components, image, block_state.device)
+
+        # Add outputs
+        self.set_block_state(state, block_state)
+        return components, state
+
+
+class WanVaeEncoderStep(PipelineBlock):
```
yiyixuxu, line 345: I think the vae enoder step should just encode the image and return the `image_latents` 

the rest of logic should go into `prepare_latents`

This way it's more "modular", both for developing and using
e.g. at runtime, if you only want to change first or last frame you only need to encode one of them and use the image_latents directly; same if you want to change num_frames

src/diffusers/modular_pipelines/wan/denoise.py
```diff
@@ -34,6 +34,56 @@
...
+                type_hint=torch.Tensor,
+                description="The latent condition to use for the denoising process.",
+            ),
+        ]
+
+    @property
+    def intermediate_outputs(self) -> List[OutputParam]:
+        return [
+            OutputParam(
+                "concatenated_latents",
```
yiyixuxu: so I think we don't want to have a different name for different pipelines/models, in SDXL I used `scaled_latent`for this variable but I think it's better we can just do `latent_model_inputs` like in regular pipeline, it's a pretty good name

yiyixuxu (commented):
looking great! thanks @a-r-r-o-w 
I left some comments, let me know what you think! from here I think we it's very easy to support wan 2.2!

yiyixuxu:
another comment w.r.t auto_blocks is I think we only need to pack workflows that can use the same checkpoint into the same package, e.g. for 5B the i2v and t2v can be packaged into an autoblocks; but not for 14B t2v and i2v we do not need to combine them because you need to load diffeerent checkpoint anyway and we should be able map the checkpoint to corresponding blocks

src/diffusers/modular_pipelines/wan/encoders.py
a-r-r-o-w: Ah right, will update

src/diffusers/modular_pipelines/wan/denoise.py
a-r-r-o-w: Sounds good!

src/diffusers/modular_pipelines/wan/encoders.py
a-r-r-o-w: Ohh, I understand now. I missed the PrepareLatents nodes specific to img2img example in SDXL. Will implement it correctly soon

a-r-r-o-w: @yiyixuxu I was taking a look at this today. A little confused by what is entailed by "rest of the logic" here. Is it just the following part, or more?

```python
        mask_lat_size = torch.ones(batch_size, 1, num_frames, latent_height, latent_width)
        if last_image is None:
            mask_lat_size[:, :, list(range(1, num_frames))] = 0
        else:
            mask_lat_size[:, :, list(range(1, num_frames - 1))] = 0
        first_frame_mask = mask_lat_size[:, :, 0:1]
        first_frame_mask = torch.repeat_interleave(
            first_frame_mask, dim=2, repeats=components.vae_scale_factor_temporal
        )
        mask_lat_size = torch.concat([first_frame_mask, mask_lat_size[:, :, 1:, :]], dim=2)
        mask_lat_size = mask_lat_size.view(
            batch_size, -1, components.vae_scale_factor_temporal, latent_height, latent_width
        )
        mask_lat_size = mask_lat_size.transpose(1, 2)
        mask_lat_size = mask_lat_size.to(latent_condition.device)
        latent_condition = torch.concat([mask_lat_size, latent_condition], dim=1)
```

yiyixuxu: I see, it was indeed confusing, I think I didn't not think it through before

How about 
1. have a `encode_vae_video` method that just encodes videos (after the image condition is pre-processeed and converted into 5D tensor) that can be imported from diffusers and used by different blocks, including the custom one( I started to refactor SDXL and I made similar one here for image https://github.com/huggingface/diffusers/blob/dc6a4d4cb4e7d46e3ac36e1c18be14ede174e104/src/diffusers/modular_pipelines/stable_diffusion_xl/encoders.py#L105)
2. in `WanVaeEncoderStep`, it should include logic to create the `viceo_condition` and put it through `encode_vae_video` to encode into `latent_condition`; The logic to create mask should stay in here too since it is closely related in how the `video_condition` is created; we should make a different `Wan*VaeEncoderStep` for 5B IT2V 
3. In a separate `prepare_latents` should only include:
      1.  generate `randn_tensor`, 
      2.  adjust the `latent_condition` and `mask` based on `batch_size` (or if you just want to keep the `prepare_latents` for 1, we can handle this logic somewhere else)

basically 
* if user want to increase `num_videos_per_prompt` , they should not need to encode images again
* if you want to use a different initial noise, you should not need to encode again

let me know what you think

tin2tin:
Uh, has this one gone out of sight - out of mind?

yiyixuxu:
already supported so closing this one :)

PR #12002, opened 2025-07-28
[tests] tighten compilation tests for quantization

sayakpaul:
# What does this PR do?

1. When not using any kind of offloading but just `torch.compile` with `fullgraph=True` on quantized models, we don't want any recompilations to get in the way of performance. This PR ensures that.
2. When using offloading, regional compilation with `fullgraph=True` is better in terms of cold-start and also overall execution time. When `_repeated_blocks` is available for a model class, we make use of `compile_repeated_blocks()` instead of `compile()`.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -25,7 +25,7 @@ jobs:
       group: aws-g6e-4xlarge
     container:
       image: diffusers/diffusers-pytorch-cuda
-      options: --shm-size "16gb" --ipc host --gpus 0
+      options: --shm-size "16gb" --ipc host --gpus all
     steps:
       - name: Checkout diffusers
         uses: actions/checkout@v3
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -61,7 +61,7 @@ jobs:
       group: aws-g4dn-2xlarge
     container:
       image: diffusers/diffusers-pytorch-cuda
-      options: --shm-size "16gb" --ipc host --gpus 0
+      options: --shm-size "16gb" --ipc host --gpus all
     steps:
       - name: Checkout diffusers
         uses: actions/checkout@v3
@@ -107,7 +107,7 @@ jobs:
       group: aws-g4dn-2xlarge
     container:
       image: diffusers/diffusers-pytorch-cuda
-      options: --shm-size "16gb" --ipc host --gpus 0
+      options: --shm-size "16gb" --ipc host --gpus all
     defaults:
       run:
         shell: bash
@@ -178,7 +178,7 @@ jobs:
 
     container:
       image: diffusers/diffusers-pytorch-cuda
-      options: --gpus 0 --shm-size "16gb" --ipc host
+      options: --gpus all --shm-size "16gb" --ipc host
 
     steps:
     - name: Checkout diffusers
@@ -222,7 +222,7 @@ jobs:
       group: aws-g6e-xlarge-plus
     container:
       image: diffusers/diffusers-pytorch-cuda
-      options: --shm-size "16gb" --ipc host --gpus 0
+      options: --shm-size "16gb" --ipc host --gpus all
     steps:
       - name: Checkout diffusers
         uses: actions/checkout@v3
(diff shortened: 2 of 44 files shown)


tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -847,6 +847,10 @@ def quantization_config(self):
             components_to_quantize=["transformer", "text_encoder_2"],
         )
 
+    @pytest.mark.xfail(
```
sayakpaul, line 850: @matthewdouglas I get:

```
- 0/0: expected type of 'module._modules['norm_out']._modules['linear']._parameters['weight'].CB' to be a tensor type, ' but found <class 'NoneType'>
```

sayakpaul:
@anijain2305 I am planning to switch to `compile_repeated_blocks(fullgraph=True)` for:

https://github.com/huggingface/diffusers/blob/284150449d805ab01aeed18fc3bae930e14fa70f/tests/quantization/test_torch_compile_utils.py#L95

This is to get rid of:

https://github.com/huggingface/diffusers/blob/284150449d805ab01aeed18fc3bae930e14fa70f/tests/quantization/test_torch_compile_utils.py#L70

However, doing so with `pytest tests/quantization/gguf/test_gguf.py::GGUFCompileTests::test_torch_compile_with_group_offload_leaf` results into:

```bash
E               torch._dynamo.exc.Unsupported: Attempted to call function marked as skipped
E                 Explanation: Dynamo developers have intentionally marked that the function `current_accelerator` in file `/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/accelerator/__init__.py` should not be traced.
E                 Hint: Avoid calling the function `current_accelerator`.
E                 Hint: Apply `@torch._dynamo.dont_skip_tracing` to the function `current_accelerator` to force tracing into the function. More graph breaks may occur as a result of attempting to trace into the function.
E                 Hint: Please file an issue to PyTorch.
E               
E                 Developer debug context: module: torch.accelerator, qualname: current_accelerator, skip reason: <missing reason>
E               
E               
E               from user code:
E                  File "/fsx/sayak/diffusers/src/diffusers/models/transformers/transformer_flux.py", line 448, in forward
E                   norm_hidden_states, gate_msa, shift_mlp, scale_mlp, gate_mlp = self.norm1(hidden_states, emb=temb)
E                 File "/fsx/sayak/diffusers/src/diffusers/models/normalization.py", line 168, in forward
E                   emb = self.linear(self.silu(emb))
E                 File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1778, in _call_impl
E                   return forward_call(*args, **kwargs)
E                 File "/fsx/sayak/diffusers/src/diffusers/hooks/hooks.py", line 188, in new_forward
E                   args, kwargs = function_reference.pre_forward(module, *args, **kwargs)
E                 File "/fsx/sayak/diffusers/src/diffusers/hooks/group_offloading.py", line 339, in pre_forward
E                   self.group.onload_()
E                 File "/fsx/sayak/diffusers/src/diffusers/hooks/group_offloading.py", line 213, in onload_
E                   getattr(torch, torch.accelerator.current_accelerator().type)
E               
E               Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"
```

I also had to comment out:

https://github.com/huggingface/diffusers/blob/284150449d805ab01aeed18fc3bae930e14fa70f/src/diffusers/hooks/group_offloading.py#L210

Do you have any recommendations?

tests/quantization/bnb/test_mixed_int8.py
matthewdouglas: For the time being I'm not sure that we can do a whole lot to avoid this for bnb int8. At the very least it is not a high priority for us. Not 100% sure but it's possible you could get around this by making a forward pass through the model prior to compiling it.

sayakpaul: I will note this down then in the `xfail` reason.

sayakpaul:
@matthewdouglas a gentle ping for review.

matthewdouglas (approved):

sayakpaul:
@DN6 can you give this a look as well?

Issue #11948, opened 2025-07-17
Impossible to load WanTransformer3DModel when offline using the 'from_pretrained' function.


guillaumejs2403:
### Describe the bug

Hi!
I am using the `Wan2.1-1.3B-Diffusers` transformer for a project. When loading the transformer (not the complete pipeline), I get a `ConnectionError` error even passing the `local_files_only=True` input. However, when loading the complete pipeline, I do not get any error. Please note that the server I use does not have access to internet. Even when I use the `HF_HUB_OFFLINE=1` env variable, I get an exception about offline mode.

### Reproduction

```python
import torch
from diffusers import WanTransformer3DModel
transformer = WanTransformer3DModel.from_pretrained("Wan-AI/Wan2.1-T2V-1.3B-Diffusers", subfolder="transformer", torch_dtype=torch.bfloat16, local_files_only=True)
```

### Logs

```shell
Error log without HF_HUB_OFFLINE=1


---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
File ~/.conda/envs/video/lib/python3.11/site-packages/urllib3/connection.py:198, in HTTPConnection._new_conn(self)
    197 try:
--> 198     sock = connection.create_connection(
    199         (self._dns_host, self.port),
    200         self.timeout,
    201         source_address=self.source_address,
    202         socket_options=self.socket_options,
    203     )
    204 except socket.gaierror as e:

File ~/.conda/envs/video/lib/python3.11/site-packages/urllib3/util/connection.py:85, in create_connection(address, timeout, source_address, socket_options)
     84 try:
---> 85     raise err
     86 finally:
     87     # Break explicitly a reference cycle

File ~/.conda/envs/video/lib/python3.11/site-packages/urllib3/util/connection.py:73, in create_connection(address, timeout, source_address, socket_options)
     72     sock.bind(source_address)
---> 73 sock.connect(sa)
     74 # Break explicitly a reference cycle

OSError: [Errno 101] Network is unreachable

The above exception was the direct cause of the following exception:

NewConnectionError                        Traceback (most recent call last)
File ~/.conda/envs/video/lib/python3.11/site-packages/urllib3/connectionpool.py:787, in HTTPConnectionPool.urlopen(self, method, url, body, headers, retries, redirect, assert_same_host, timeout, pool_timeout, release_conn, chunked, body_pos, preload_content, decode_content, **response_kw)
    786 # Make the request on the HTTPConnection object
[...]

PR #12005, opened 2025-07-28
[core] respect `local_files_only=True` when using sharded checkpoints

sayakpaul:
# What does this PR do?

See: https://github.com/huggingface/diffusers/issues/11948

### DIFF
--- a/src/diffusers/utils/hub_utils.py
+++ b/src/diffusers/utils/hub_utils.py
@@ -405,20 +405,14 @@ def _get_checkpoint_shard_files(
 
     # If the repo doesn't have the required shards, error out early even before downloading anything.
     if not local_files_only:
-        try:
-            model_files_info = model_info(pretrained_model_name_or_path, revision=revision, token=token)
-            for shard_file in original_shard_filenames:
-                shard_file_present = any(shard_file in k.rfilename for k in model_files_info.siblings)
-                if not shard_file_present:
-                    raise EnvironmentError(
-                        f"{shards_path} does not appear to have a file named {shard_file} which is "
-                        "required according to the checkpoint index."
-                    )
-        except ConnectionError as e:
-            raise EnvironmentError(
-                f"We couldn't connect to '{HUGGINGFACE_CO_RESOLVE_ENDPOINT}' to load {pretrained_model_name_or_path}. You should try"
-                " again after checking your internet connection."
-            ) from e
+        model_files_info = model_info(pretrained_model_name_or_path, revision=revision, token=token)
+        for shard_file in original_shard_filenames:
+            shard_file_present = any(shard_file in k.rfilename for k in model_files_info.siblings)
+            if not shard_file_present:
+                raise EnvironmentError(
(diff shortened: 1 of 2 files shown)


src/diffusers/utils/hub_utils.py
```diff
@@ -403,9 +404,26 @@ def _get_checkpoint_shard_files(
 
     ignore_patterns = ["*.json", "*.md"]
     # `model_info` call must guarded with the above condition.
-    model_files_info = model_info(pretrained_model_name_or_path, revision=revision, token=token)
```
DN6, line 406: So the purpose of this check is to verify if the necessary sharded files are present in the model repo before attempting a download, presumably to avoid a large download if all files aren't present. If we cannot connect to the hub, we just have to assume the necessary shard files are already present locally.  

I think we can just skip this check if `local_files_only=True`  and then check if all the shard filenames are present in the `cached_folder`

sayakpaul: How about now?

DN6: I think just this is sufficient 
```
if not local_files_only:
    # run model_info check
```
Run snapshot download 

Then after the cached_filenames is created, iterate over the files to verify they exist

```
for filename in cached_filename:
      if not if not os.path.exists(filename):
           raise EnvironmentError("expected file not present in {cached_folder}")
```

sayakpaul: 1. We don't have to run `snapshot_download()` when `local_files_only=False`, that might be unnecessary. 
2. Why run `snapshot_download()` after also running `model_info()`?
3. Even if we run `snapshot_download()` regardless of `local_files_only` var, I think we should have it inside `try-except` in case the endpoint cannot be pinged for some reason and raise the `ConnectionError` as before.

sayakpaul: See if [b7af511](https://github.com/huggingface/diffusers/pull/12005/commits/b7af5111c483440f11a67f512697a84186538278) resolves this.

sayakpaul: Ah I see what you mean. Let me update. Sorry about the back and forth.

sayakpaul:
@DN6 see if the latest changes work for you.

src/diffusers/utils/hub_utils.py
```diff
@@ -428,6 +436,16 @@ def _get_checkpoint_shard_files(
         if subfolder is not None:
             cached_folder = os.path.join(cached_folder, subfolder)
 
+        # Check again after downloading/loading from the cache.
+        model_files_info = _get_filepaths_for_folder(cached_folder)
```
DN6: We don't need a new function to walk over the cached folder. If you just iterate over `cached_filenames` and check if the file exits. Avoid looping over files multiple times this way. 

```python
for cached_file in cached_filenames:
      if not os.path.exists(cached_file):
                raise EnvironmentError(f"{cached_file} not found in {cached_folder which is required..." 
```

DN6 (approved):

PR #12020, opened 2025-07-30
[docs] Update index

stevhliu:
Updates `index.md` since we've moved a bit away from the older layout of tutorials/guides/conceptual guides/reference.

### DIFF
--- a/docs/source/en/index.md
+++ b/docs/source/en/index.md
@@ -12,37 +12,24 @@ specific language governing permissions and limitations under the License.
 
 <p align="center">
     <br>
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):
Nice but curious to know why moved away from the previous layout as it looked quite complete to me.

stevhliu:
We don't have a dedicated sections for tutorials/guides/conceptual guides/reference in the `toctree`, so I think it makes more sense to remove those buttons to be more aligned. I want users to think about the docs more in terms of topics they're interested in rather than whether they're looking for a tutorial, guide, etc.

Hope that makes sense! 🤗

PR #12004, opened 2025-07-28
[WIP] Wan2.2

yiyixuxu:
install from PR 

```
pip install git+https://github.com/huggingface/diffusers.git@wan2.2
```

TI2V (only Text-to-image is supported for now, adding I2V soon)

```python
import torch
import numpy as np
from diffusers import WanPipeline, AutoencoderKLWan, WanTransformer3DModel, UniPCMultistepScheduler
from diffusers.utils import export_to_video, load_image

dtype = torch.bfloat16
device = "cuda"

model_id = "Wan-AI/Wan2.2-TI2V-5B-Diffusers"
vae = AutoencoderKLWan.from_pretrained(model_id, subfolder="vae", torch_dtype=torch.float32)
pipe = WanPipeline.from_pretrained(model_id, vae=vae, torch_dtype=dtype)
pipe.to(device)

height = 704
width = 1280
num_frames = 121
num_inference_steps = 50
guidance_scale = 5.0


prompt = "Two anthropomorphic cats in comfy boxing gear and bright gloves fight intensely on a spotlighted stage."
negative_prompt = "色调艳丽，过曝，静态，细节模糊不清，字幕，风格，作品，画作，画面，静止，整体发灰，最差质量，低质量，JPEG压缩残留，丑陋的，残缺的，多余的手指，画得不好的手部，画得不好的脸部，畸形的，毁容的，形态畸形的肢体，手指融合，静止不动的画面，杂乱的背景，三条腿，背景人很多，倒着走"

output = pipe(
    prompt=prompt,
    negative_prompt=negative_prompt,
    height=height,
    width=width,
    num_frames=num_frames,
    guidance_scale=guidance_scale,
    num_inference_steps=num_inference_steps,
).frames[0]
export_to_video(output, "5bit2v_output.mp4", fps=24)
```

14B T2V

```py
import torch
import numpy as np
from diffusers import WanPipeline, AutoencoderKLWan
from diffusers.utils import export_to_video, load_image

dtype = torch.bfloat16
device = "cuda:2"
vae = AutoencoderKLWan.from_pretrained("Wan-AI/Wan2.2-T2V-A14B-Diffusers", subfolder="vae", torch_dtype=torch.float32)
pipe = WanPipeline.from_pretrained("Wan-AI/Wan2.2-T2V-A14B-Diffusers", vae=vae, torch_dtype=dtype)
pipe.to(device)

height = 720
width = 1280


[...]

PR #12024, opened 2025-07-30
[wan2.2] follow-up

yiyixuxu:
```py
# wan 2.2 5b i2v slow test (using only low-noise stage transformer)
import torch
import numpy as np
from diffusers import WanImageToVideoPipeline
from diffusers.utils import export_to_video, load_image
import time


model_id = "Wan-AI/Wan2.2-I2V-A14B-Diffusers"
dtype = torch.bfloat16
device = "cuda:1"

pipe = WanImageToVideoPipeline.from_pretrained(model_id, torch_dtype=dtype, transformer=None, boundary_ratio=1.0)
pipe.to(device)


image = load_image(
    "https://huggingface.co/datasets/YiYiXu/testing-images/resolve/main/wan_i2v_input.JPG"
)

max_area = 480 * 832
aspect_ratio = image.height / image.width
mod_value = pipe.vae_scale_factor_spatial * pipe.transformer_2.config.patch_size[1]
height = round(np.sqrt(max_area * aspect_ratio)) // mod_value * mod_value
width = round(np.sqrt(max_area / aspect_ratio)) // mod_value * mod_value
image = image.resize((width, height))
prompt = "Summer beach vacation style, a white cat wearing sunglasses sits on a surfboard. The fluffy-furred feline gazes directly at the camera with a relaxed expression. Blurred beach scenery forms the background featuring crystal-clear waters, distant green hills, and a blue sky dotted with white clouds. The cat assumes a naturally relaxed posture, as if savoring the sea breeze and warm sunlight. A close-up shot highlights the feline's intricate details and the refreshing atmosphere of the seaside."

negative_prompt = "色调艳丽，过曝，静态，细节模糊不清，字幕，风格，作品，画作，画面，静止，整体发灰，最差质量，低质量，JPEG压缩残留，丑陋的，残缺的，多余的手指，画得不好的手部，画得不好的脸部，畸形的，毁容的，形态畸形的肢体，手指融合，静止不动的画面，杂乱的背景，三条腿，背景人很多，倒着走"

output = pipe(
    image=image,
    prompt=prompt,
    height=height,
    width=width,
    num_frames=81,
    guidance_scale=3.5,
    num_inference_steps=40,
    generator=generator,
).frames[0]


export_to_video(output, "output.mp4", fps=16)

```

### DIFF
--- a/src/diffusers/pipelines/wan/pipeline_wan.py
+++ b/src/diffusers/pipelines/wan/pipeline_wan.py
@@ -131,9 +131,9 @@ class WanPipeline(DiffusionPipeline, WanLoraLoaderMixin):
         self,
         tokenizer: AutoTokenizer,
         text_encoder: UMT5EncoderModel,
-        transformer: WanTransformer3DModel,
         vae: AutoencoderKLWan,
         scheduler: FlowMatchEulerDiscreteScheduler,
+        transformer: Optional[WanTransformer3DModel] = None,
         transformer_2: Optional[WanTransformer3DModel] = None,
         boundary_ratio: Optional[float] = None,
         expand_timesteps: bool = False,  # Wan2.2 ti2v
@@ -526,7 +526,7 @@ class WanPipeline(DiffusionPipeline, WanLoraLoaderMixin):
             device=device,
         )
 
-        transformer_dtype = self.transformer.dtype
+        transformer_dtype = self.transformer.dtype if self.transformer is not None else self.transformer_2.dtype
         prompt_embeds = prompt_embeds.to(transformer_dtype)
         if negative_prompt_embeds is not None:
             negative_prompt_embeds = negative_prompt_embeds.to(transformer_dtype)
@@ -536,7 +536,7 @@ class WanPipeline(DiffusionPipeline, WanLoraLoaderMixin):
         timesteps = self.scheduler.timesteps
 
         # 5. Prepare latent variables
-        num_channels_latents = self.transformer.config.in_channels
+        num_channels_latents = self.transformer.config.in_channels if self.transformer is not None else self.transformer_2.config.in_channels
         latents = self.prepare_latents(
             batch_size * num_videos_per_prompt,
             num_channels_latents,
--- a/src/diffusers/pipelines/wan/pipeline_wan_i2v.py
+++ b/src/diffusers/pipelines/wan/pipeline_wan_i2v.py
@@ -162,17 +162,17 @@ class WanImageToVideoPipeline(DiffusionPipeline, WanLoraLoaderMixin):
 
     model_cpu_offload_seq = "text_encoder->image_encoder->transformer->transformer_2->vae"
     _callback_tensor_inputs = ["latents", "prompt_embeds", "negative_prompt_embeds"]
-    _optional_components = ["transformer_2", "image_encoder", "image_processor"]
+    _optional_components = ["transformer", "transformer_2", "image_encoder", "image_processor"]
 
     def __init__(
         self,
         tokenizer: AutoTokenizer,
         text_encoder: UMT5EncoderModel,
-        transformer: WanTransformer3DModel,
         vae: AutoencoderKLWan,
         scheduler: FlowMatchEulerDiscreteScheduler,
         image_processor: CLIPImageProcessor = None,
         image_encoder: CLIPVisionModel = None,
+        transformer: WanTransformer3DModel = None,
         transformer_2: WanTransformer3DModel = None,
         boundary_ratio: Optional[float] = None,
         expand_timesteps: bool = False,
@@ -669,12 +669,13 @@ class WanImageToVideoPipeline(DiffusionPipeline, WanLoraLoaderMixin):
         )
 
         # Encode image embedding
-        transformer_dtype = self.transformer.dtype
+        transformer_dtype = self.transformer.dtype if self.transformer is not None else self.transformer_2.dtype
         prompt_embeds = prompt_embeds.to(transformer_dtype)
         if negative_prompt_embeds is not None:
             negative_prompt_embeds = negative_prompt_embeds.to(transformer_dtype)
 
-        if self.config.boundary_ratio is None and not self.config.expand_timesteps:
+        # only wan 2.1 i2v transformer accepts image_embeds
+        if self.transformer is not None and self.transformer.config.added_kv_proj_dim is not None:
             if image_embeds is None:
                 if last_image is None:
                     image_embeds = self.encode_image(image, device)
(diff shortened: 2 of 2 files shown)


JoeGaffney:
Hey, 

With them both being accessed in one loop would **enable_model_cpu_offload** still work. Like when it hits the transformer_2 it could offload transformer?

Cheers,
Joe

okaris:
@JoeGaffney interesting question, i was trying to debug a similar issue with `enable_model_cpu_offload` yesterday and i think i will need to open an issue.

for wan2.2 i2v the execution path is 
`text_encoder->vae(encode)->transformer->transformer_2->vae(decode)`

while the offload sequence is defined as

`text_encoder->transformer->transformer_2->vae`

for me it causes the text_encoder to stay on gpu after encoding the initial image, but also changing the sequence is problematic and currently it supports a single position. i know this deserves it's own issue, still collecting examples

coming back to your comment i think it's a valid argument that if the sequence is defined as `transformer->transformer_2` and  2 never runs it might get stuck on the gpu.

JoeGaffney:
Hey @okaris
I'm wondering if the behaviours have diverged enough between 2.1 and 2.2 that these should maybe be separate pipelines. I'm not an expert in authoring diffusers pipelines, but there seem to be significant changes in expectations and data flow. Maybe there are common components that could be lifted out to avoid duplication, while still reflecting the divergence.

Also, I think referencing something like this in the example:

```python
 image_processor = ModularPipeline.from_pretrained("YiYiXu/WanImageProcessor14B", trust_remote_code=True)
 image = image_processor(
     image="https://huggingface.co/datasets/YiYiXu/testing-images/resolve/main/wan_i2v_input.JPG",
     output="processed_image"
)
```
 ...feels a bit fragile. It relies on remote, opaque behavior. I get that it's convenient, but it’s not ideal for production integration or debugging. It would be great to also provide a minimal example that shows how to prepare inputs manually. 

Cheers,
Joe

ukaprch:
Some have mentioned that during their testing the 1st stage is relatively useless. Has anyone on the team actually done sufficient testing on WAN 2.2 to make a determination as to what approach works the best?

yiyixuxu:
>for me it causes the text_encoder to stay on gpu after encoding the initial image, but also changing the sequence is problematic and currently it supports a single position. i know this deserves it's own issue, still collecting examples

so, yes text_encoder will stay on gpu after encoding the initial image, but it will get offloaded when `transformer` is used - so it does not matter that it stayed on GPU since the memory bottleneck is transformer (i.e. it does not increasee overall memory requirement)

vae would stay on GPU until the it's used again, including the time when transformers are loaded - but it's really small relatively so does not make that much diffeerence.  we could force offload vae though if it's needed

mayankagrawal10198:
@yiyixuxu How much boundary_ratio I should keep for lightx2v lora where inference step is just 4
please refer this https://github.com/huggingface/diffusers/commit/9a2eaed002af7e86580cf2df96272c36176feda6

src/diffusers/models/transformers/transformer_wan.py
```diff
@@ -324,7 +324,7 @@ def forward(
     ):
         timestep = self.timesteps_proj(timestep)
         if timestep_seq_len is not None:
-            timestep = timestep.unflatten(0, (1, timestep_seq_len))
+            timestep = timestep.unflatten(0, (-1, timestep_seq_len))
```
yiyixuxu, line 327: so that it works with batch_size > 1

tests/pipelines/wan/test_wan.py
```diff
@@ -85,29 +87,13 @@ def get_dummy_components(self):
             rope_max_seq_len=32,
         )
 
-        torch.manual_seed(0)
```
yiyixuxu, line 88: revert the change in #12004

tests/pipelines/wan/test_wan_22.py
```diff
@@ -0,0 +1,368 @@
...
+            "generator",
+            "latents",
+            "return_dict",
+            "callback_on_step_end",
+            "callback_on_step_end_tensor_inputs",
+        ]
+    )
+    test_xformers_attention = False
+    supports_dduf = False
+    test_optional_components = ["transformer_2", "transformer"]
```
yiyixuxu:
```suggestion
```

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -86,23 +88,6 @@ def get_dummy_components(self):
             image_dim=4,
         )
 
-        torch.manual_seed(0)
```
yiyixuxu, line 89: reevert changes in #12004

tests/pipelines/wan/test_wan_22.py
```diff
@@ -0,0 +1,367 @@
...
+        return components
+
+    def get_dummy_inputs(self, device, seed=0):
+        if str(device).startswith("mps"):
+            generator = torch.manual_seed(seed)
+        else:
+            generator = torch.Generator(device=device).manual_seed(seed)
+        inputs = {
+            "prompt": "dance monkey",
+            "negative_prompt": "negative",  # TODO
```
sayakpaul: 👀 why is it a TODO?

tests/pipelines/wan/test_wan_22.py
```diff
@@ -0,0 +1,367 @@
...
+
+        inputs = self.get_dummy_inputs(generator_device)
+        torch.manual_seed(0)
+        output_loaded = pipe_loaded(**inputs)[0]
+
+        max_diff = np.abs(output.detach().cpu().numpy() - output_loaded.detach().cpu().numpy()).max()
+        self.assertLess(max_diff, expected_max_difference)
+
+
+class Wan225BPipelineFastTests(PipelineTesterMixin, unittest.TestCase):
```
sayakpaul, line 199: (nit): adding a line to tell how it's different from the above might be helpful.

tests/pipelines/wan/test_wan_22.py
```diff
@@ -0,0 +1,367 @@
...
+        self.assertTrue(
+            torch.allclose(generated_slice, expected_slice, atol=1e-3),
+            f"generated_slice: {generated_slice}, expected_slice: {expected_slice}",
+        )
+
+    @unittest.skip("Test not supported")
+    def test_attention_slicing_forward_pass(self):
+        pass
+
+    def test_components_function(self):
```
sayakpaul, line 319: Cool test!

sayakpaul (approved):

tests/pipelines/wan/test_wan_22.py
yiyixuxu:
```suggestion
            "negative_prompt": "negative", 
```

yiyixuxu:
@bot /style

PR #12027, opened 2025-07-31
[lora] factor out the overlaps in `save_lora_weights()`.

sayakpaul:
# What does this PR do?

Factors out the common bits shared by different `save_lora_weights()` functions while preserving the explicitness.


sayakpaul:
@DN6 a gentle ping.

DN6 (approved):
Thanks 👍🏽

PR #12000, opened 2025-07-28
[training-scripts] Make pytorch examples UV-compatible

sayakpaul:
# What does this PR do?

Similar to: https://github.com/huggingface/transformers/pull/39635.

Used the following scripts (thanks to o4 for pairing) to automate:
https://gist.github.com/sayakpaul/7f1f171d55c1122379ab0f5bdd4bd3dc

PR #12036, opened 2025-07-31
Add UV script comment block

apolinario:
This PR makes our advanced diffusion training compatible with uv, which therefore also makes them compatible with Hugging Face jobs: https://huggingface.co/docs/huggingface_hub/en/guides/jobs

The idea is to test it out with this PR, working well we could expand the uv comment for all diffusers scripts to add both uv and hf jobs compatibility

cc @sayakpaul @linoytsaban

### DIFF
--- a/examples/advanced_diffusion_training/train_dreambooth_lora_flux_advanced.py
+++ b/examples/advanced_diffusion_training/train_dreambooth_lora_flux_advanced.py
@@ -34,6 +34,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 
+# /// script
+# dependencies = [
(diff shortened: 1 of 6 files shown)


sayakpaul:
https://github.com/huggingface/diffusers/pull/12000/files

linoytsaban:
@apolinario @sayakpaul since sayak's open PR covers this script as well, lets maybe just add the additional dependencies in this PR to the advanced block in #12000 ?

apolinario:
Thanks, sorry for missing it. Yeah, I think we need a few more deps for the advanced scripts. Let's chat on @sayakpaul's PR then!

sayakpaul:
I used `requirements.txt` to fetch those. I will merge mine in sometime. Feel free to merge those updates in this PR and iterate.

apolinario:
Gotcha. I'll keep this PR then and merge yours to it and add the deps for the advanced script on top of it

PR #12055, opened 2025-08-03
Qwen-Image

naykun:
# Adding Qwen-Image

This PR introduces Qwen-Image into the diffusers library. For more information about Qwen-Image and advanced usage examples, see the official repository:  [Qwen-Image](https://github.com/QwenLM/Qwen-Image)

Thanks the support of @yiyixuxu @a-r-r-o-w


tests/pipelines/qwenimage/test_qwen_image.py
```diff
@@ -0,0 +1,35 @@
+import torch
```
yiyixuxu: can we remove this file?

yiyixuxu (approved):
thanks!

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -0,0 +1,634 @@
...
+        timesteps_proj = self.time_proj(timestep)
+        timesteps_emb = self.timestep_embedder(timesteps_proj.to(dtype=hidden_states.dtype))  # (N, D)
+
+        conditioning = timesteps_emb
+
+        return conditioning
+
+
+class QwenEmbedRope(nn.Module):
+    def __init__(self, theta: int, axes_dim: list[int], scale_rope=False):
```
yiyixuxu:
```suggestion
    def __init__(self, theta: int, axes_dim: List[int], scale_rope=False):
```

PR #12057, opened 2025-08-04
tests + minor refactor for QwenImage

a-r-r-o-w:
cc @naykun

### DIFF
--- a/src/diffusers/models/transformers/transformer_qwenimage.py
+++ b/src/diffusers/models/transformers/transformer_qwenimage.py
@@ -140,7 +140,7 @@ def apply_rotary_emb_qwen(
 
 
 class QwenTimestepProjEmbeddings(nn.Module):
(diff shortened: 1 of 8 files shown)


src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -495,8 +493,7 @@ def __init__(
         attention_head_dim: int = 128,
         num_attention_heads: int = 24,
         joint_attention_dim: int = 3584,
-        pooled_projection_dim: int = 768,
-        guidance_embeds: bool = False,
+        guidance_embeds: bool = False,  # TODO: this should probably be removed
```
yiyixuxu, line 496: cc @naykun can you confirm if we need this config?

yiyixuxu (approved):
thanks @a-r-r-o-w

PR #12061, opened 2025-08-04
enable compilation in qwen image.

sayakpaul:
# What does this PR do?

* Adds tests for the Qwen transformer model tests.
* Enable compilation with triggering recompilations.

Timing (compilation):

* PR branch: `timings.mean()=tensor(42.8954)`
* `tests/qwen-image`: `timings.mean()=tensor(75.9385)`

<details>
<summary>Code</summary>

```py
import torch
import time

pipe = DiffusionPipeline.from_pretrained("Qwen/Qwen-Image", torch_dtype=torch.bfloat16).to("cuda")
# pipe.load_lora_weights("trained-qwen-image-lora")

timings = []
for _ in range(3):
    start = time.time()
    image = pipe(
        "a 3dicon, a llama with a signboard saying 'Qwen is awesome'", 
        guidance_scale=1.0, 
        num_inference_steps=50,

    ).images[0]
    end = time.time()
    timings.append(end - start)

timings = torch.tensor(timings)
print(f"{timings.mean()=}")
image.save("llama_pretrained.png")
```

</details>

### DIFF
--- a/src/diffusers/models/transformers/transformer_qwenimage.py
+++ b/src/diffusers/models/transformers/transformer_qwenimage.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import functools
 import math
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -162,7 +163,7 @@ class QwenEmbedRope(nn.Module):
         self.axes_dim = axes_dim
         pos_index = torch.arange(1024)
         neg_index = torch.arange(1024).flip(0) * -1 - 1
-        self.pos_freqs = torch.cat(
+        pos_freqs = torch.cat(
             [
                 self.rope_params(pos_index, self.axes_dim[0], self.theta),
                 self.rope_params(pos_index, self.axes_dim[1], self.theta),
@@ -170,7 +171,7 @@ class QwenEmbedRope(nn.Module):
             ],
             dim=1,
         )
-        self.neg_freqs = torch.cat(
+        neg_freqs = torch.cat(
             [
                 self.rope_params(neg_index, self.axes_dim[0], self.theta),
                 self.rope_params(neg_index, self.axes_dim[1], self.theta),
@@ -179,6 +180,8 @@ class QwenEmbedRope(nn.Module):
             dim=1,
         )
         self.rope_cache = {}
+        self.register_buffer("pos_freqs", pos_freqs, persistent=False)
+        self.register_buffer("neg_freqs", neg_freqs, persistent=False)
 
         # 是否使用 scale rope
         self.scale_rope = scale_rope
@@ -198,33 +201,17 @@ class QwenEmbedRope(nn.Module):
         Args: video_fhw: [frame, height, width] a list of 3 integers representing the shape of the video Args:
         txt_length: [bs] a list of 1 integers representing the length of the text
         """
-        if self.pos_freqs.device != device:
-            self.pos_freqs = self.pos_freqs.to(device)
-            self.neg_freqs = self.neg_freqs.to(device)
-
         if isinstance(video_fhw, list):
             video_fhw = video_fhw[0]
         frame, height, width = video_fhw
         rope_key = f"{frame}_{height}_{width}"
 
-        if rope_key not in self.rope_cache:
-            seq_lens = frame * height * width
-            freqs_pos = self.pos_freqs.split([x // 2 for x in self.axes_dim], dim=1)
-            freqs_neg = self.neg_freqs.split([x // 2 for x in self.axes_dim], dim=1)
-            freqs_frame = freqs_pos[0][:frame].view(frame, 1, 1, -1).expand(frame, height, width, -1)
-            if self.scale_rope:
-                freqs_height = torch.cat([freqs_neg[1][-(height - height // 2) :], freqs_pos[1][: height // 2]], dim=0)
-                freqs_height = freqs_height.view(1, height, 1, -1).expand(frame, height, width, -1)
-                freqs_width = torch.cat([freqs_neg[2][-(width - width // 2) :], freqs_pos[2][: width // 2]], dim=0)
-                freqs_width = freqs_width.view(1, 1, width, -1).expand(frame, height, width, -1)
-
-            else:
-                freqs_height = freqs_pos[1][:height].view(1, height, 1, -1).expand(frame, height, width, -1)
-                freqs_width = freqs_pos[2][:width].view(1, 1, width, -1).expand(frame, height, width, -1)
-
-            freqs = torch.cat([freqs_frame, freqs_height, freqs_width], dim=-1).reshape(seq_lens, -1)
-            self.rope_cache[rope_key] = freqs.clone().contiguous()
-        vid_freqs = self.rope_cache[rope_key]
+        if not torch.compiler.is_compiling():
+            if rope_key not in self.rope_cache:
+                self.rope_cache[rope_key] = self._compute_video_freqs(frame, height, width)
+            vid_freqs = self.rope_cache[rope_key]
+        else:
+            vid_freqs = self._compute_video_freqs(frame, height, width)
 
         if self.scale_rope:
             max_vid_index = max(height // 2, width // 2)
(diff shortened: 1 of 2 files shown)


src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -198,33 +201,17 @@ def forward(self, video_fhw, txt_seq_lens, device):
         Args: video_fhw: [frame, height, width] a list of 3 integers representing the shape of the video Args:
         txt_length: [bs] a list of 1 integers representing the length of the text
         """
-        if self.pos_freqs.device != device:
-            self.pos_freqs = self.pos_freqs.to(device)
-            self.neg_freqs = self.neg_freqs.to(device)
-
```
sayakpaul, line 204: Recompilation trigger one.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -198,33 +201,17 @@ def forward(self, video_fhw, txt_seq_lens, device):
...
         if isinstance(video_fhw, list):
             video_fhw = video_fhw[0]
         frame, height, width = video_fhw
         rope_key = f"{frame}_{height}_{width}"
 
-        if rope_key not in self.rope_cache:
```
sayakpaul, line 210: Recompilation trigger two.

tests/models/test_modeling_common.py
```diff
@@ -1711,6 +1711,11 @@ def test_group_offloading_with_disk(self, offload_type, record_stream, atol=1e-5
         if not self.model_class._supports_group_offloading:
             pytest.skip("Model does not support group offloading.")
 
+        if self.model_class.__name__ == "QwenImageTransformer2DModel":
+            pytest.skip(
+                "QwenImageTransformer2DModel doesn't support group offloading with disk. Needs to be investigated."
+            )
+
```
sayakpaul, line 1718: Will investigate in a follow-up.

a-r-r-o-w:
@sayakpaul Could you rebase with main? Sorry I didn't see this before stacked over tests/qwen-image

sayakpaul:
Done!

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -236,6 +223,25 @@ def forward(self, video_fhw, txt_seq_lens, device):
 
         return vid_freqs, txt_freqs
 
+    @functools.lru_cache(maxsize=None)
+    def _compute_video_freqs(self, frame, height, width):
```
a-r-r-o-w, line 227: TODO: we need to remove frame (can be done in future PR)

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -236,6 +223,25 @@ def forward(self, video_fhw, txt_seq_lens, device):
 
         return vid_freqs, txt_freqs
 
+    @functools.lru_cache(maxsize=None)
```
a-r-r-o-w, line 226: Let's remove the `self.rope_cache` and just use the lru_cache implementation? WDYT @yiyixuxu?

WDYT about maybe putting maxsize=128 or something here so that long running services that use diffusers don't accidentally die with OOM (probably very unlikely though) @sayakpaul?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -179,6 +180,8 @@ def __init__(self, theta: int, axes_dim: List[int], scale_rope=False):
             dim=1,
         )
         self.rope_cache = {}
+        self.register_buffer("pos_freqs", pos_freqs, persistent=False)
```
a-r-r-o-w, line 183: This is most likely not equivalent. When registered as buffer, if the model is loaded in bf16, the precision of these will bf16 instead of fp32. Doing RoPE in bf16 may harm image quality, so we need to be careful here. Not sure what's best to do here -- maybe for now we can put the rope layer in `_keep_modules_in_fp32`?

This recompilation related problem seems to have become too common with RoPE. Maybe we need to rethink the design a bit.

sayakpaul: Just for the record, sharing the recompilation error we get without the buffer implementation:

```bash

>               raise exc.RecompileError(message)

E               torch._dynamo.exc.RecompileError: Recompiling function forward in /fsx/sayak/diffusers/src/diffusers/models/transformers/transformer_qwenimage.py:529

E                   triggered by the following guard failure(s):

E                   - 0/0: tensor 'self._modules['pos_embed'].neg_freqs' dispatch key set mismatch. expected DispatchKeySet(CPU, BackendSelect, ADInplaceOrView, AutogradCPU), actual DispatchKeySet(CUDA, BackendSelect, ADInplaceOrView, AutogradCUDA)


../miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/_dynamo/guards.py:3822: RecompileError
```

But I agree with your first point.

src/diffusers/models/transformers/transformer_qwenimage.py
sayakpaul: `maxsize=128` sounds reasonable to me.

sayakpaul:
@a-r-r-o-w I dug deeper into the RoPE embed stuff you brought up in https://github.com/huggingface/diffusers/pull/12061#discussion_r2251239258. Summary below.

I am using the code from the OP for all investigations with the change of `generator=torch.manual_seed(0)` to the pipeline call.

First, I printed `print(pipe.transformer.pos_embed.pos_freqs.dtype)` from both `main` and this branch and got `torch.complex64` in both cases. I also printed the `freqs.dtype` after [here](https://github.com/huggingface/diffusers/blob/9c1d4e3be1580b3174cb0eb099a135aeb55a807c/src/diffusers/models/transformers/transformer_qwenimage.py#L192) and [here](https://github.com/huggingface/diffusers/blob/9c1d4e3be1580b3174cb0eb099a135aeb55a807c/src/diffusers/models/transformers/transformer_qwenimage.py#L193). In both cases, I got `torch.float32` and `torch.complex64`, respectively.

Then, I moved on to qualitative comparisons. Result is below:

<details>
<summary>Unfold</summary>

| Alt text 1 | Alt text 2 |
|:-------------------------:|:-------------------------:|
|       `main`           |       PR branch           |

</details>

I think since the buffers we're registering in this PR aren't persistent, we should be okay because `rope_params` already returns the freqs in `torch.complex64`.

LMK your thoughts.

src/diffusers/models/transformers/transformer_qwenimage.py
anijain2305: The recompilation message here says that on the first compilation `'self._modules['pos_embed'].neg_freqs'` was a CPU tensor, and on second it became a CUDA tensor. Does that match your expectation? If yes, is it possible to change that somehow. If there is something special happening on the first invocation, you can put compile on the second invocation onwards.

src/diffusers/models/transformers/transformer_qwenimage.py
anijain2305: Is this again because there is something special happening on the first run?

anijain2305 (commented):
Trying to understand if there is something special happening on the first run, in which case, perhaps enabling compile from the second invocation onwards might be a better way.

src/diffusers/models/transformers/transformer_qwenimage.py
sayakpaul: > If there is something special happening on the first invocation, you can put compile on the second invocation onwards.

Is, IMO, an involved user-experience we should probably avoid.

src/diffusers/models/transformers/transformer_qwenimage.py
sayakpaul: The `forward` method has a side effect with the dict caching method. It modifies the `self.rope_cache` dictionary.

sayakpaul:
Posting here a summary of what @anijain2305 and I discussed offline. 

## Regional compilation is sweet!

FWIW, `compile_repeated_blocks()` (regional compilation) performs about the same as full compilation but cuts the cold start drastically :) 

So, to test this, I just added `_repeated_blocks = ["QwenImageTransformerBlock"]` to the modeling implementation in `main` and ran the OP timing code. Obtained the results as same as full compilation. So, we have that option. 

However, full compilation is needed when catering to LoRA hotswapping use cases as LoRA target modules tend to also target non-transformer blocks (the ones not in `_repeated_blocks`).

## Full compilation modifications

There are multiple ways to tackle this. The changes introduced in this PR are just one way to resolve the issues. One can enforce eager execution on the first invocation of the transformer and then fall back to default compilation. However, this is a bit of a shame in terms of user-experience. 

Perhaps we can document all of this in the Qwen pipeline doc page? WDYT?

Cc: @a-r-r-o-w @yiyixuxu

a-r-r-o-w:
Ah I see now that the result returned from `rope_params` is complex64, thanks for looking into it! I was previously under the impression this was float32 (missing the fact that `torch.polar` creates complex64). The dtype would've been as issue if it was fp32 when model is loaded in bf16, but indeed complex64 is expected to remain unchanged. So, the changes look good to me.

I remember that it was significantly slower to run RoPE computation with complex numbers, and that inductor does not support optimizations when complex numbers are involved (I don't remember the exact problem, but there is definitely some speedup to be gained by removing complex number use). Not for this PR, but this should be refactored like what was done for Wan

a-r-r-o-w (approved):
Thanks, LGTM!

sayakpaul:
I have run the compilation tests, they are passing. Will merge after the CI.

PR #12089, opened 2025-08-07
fix flux type hint

DefTruth:
FluxSingleTransformerBlock:

The output type should be Tuple[torch.Tensor, torch.Tensor], not torch.Tensor.


a-r-r-o-w (approved):
Thanks, nice catch!

PR #12093, opened 2025-08-07
[wip] Text to Image Pipeline for wan 

luke14free:
# What does this PR do?

Adds a new pipeline to support Wan low-noise transformer only image generation, which is getting a lot of traction (esp with new loras on civit).

Since this new pipeline is just a reshuffle of the components of t2v one, I am quite unsure if what I did by overwriting from_pretrained makes sense/is acceptable. I also had to add a new return type which again might not follow your style, feel free to ask for changes.

The general idea of the pipeline is that I am only using the transformer_2 (the low noise) which is being loaded as the main transformer. Then I perform the prediction of just one frame and return that. It works with ggufs as expected by passing a gguf transformer in the pipeline.

## Who can review?

maybe @yiyixuxu ? not sure who is the right person

### DIFF
--- a/src/diffusers/pipelines/wan/pipeline_wan_t2i.py
+++ b/src/diffusers/pipelines/wan/pipeline_wan_t2i.py
@@ -627,7 +627,9 @@ class WanTextToImagePipeline(DiffusionPipeline, WanLoraLoaderMixin):
             images = self.vae.decode(latents, return_dict=False)[0]
             images = self.video_processor.postprocess_video(images, output_type=output_type)
             # Squeeze temporal dimension for single-frame output (num_frames=1)
-            images = images.squeeze(2)  # Remove temporal dimension: (B, C, 1, H, W) -> (B, C, H, W)
+            # Only squeeze if temporal dimension has size 1
+            if images.dim() == 5 and images.shape[2] == 1:
+                images = images.squeeze(2)  # Remove temporal dimension: (B, C, 1, H, W) -> (B, C, H, W)
             if isinstance(images, list):
                 images = [img[0] if len(img) > 0 else img for img in images]
         else:


luke14free:
[wip] I need to figure out why images are being generated as oversaturated/very red, think I am missing a video processing step after generation

asomoza:
Hi @luke14free , what would be the difference between using the normal pipeline with a `boundary_ratio = 1.0`  and `num_frames=1` and this?

luke14free:
hey @asomoza it is pretty much the same, except you don't also load transformer (the high noise one). you can 100% make it work with existing pipelines. just thought it would have been cleaner to have a separate pipeline, but in case it's not i'll close

asomoza:
@luke14free not completely sure (haven't tested it yet) but I think you can pass a `transformer=None` so it doesn't load. You can keep it open if you want and see if there's some interest in having a specific pipeline for this but mostly we don't create additional pipelines when you can do the same with the base one.

luke14free:
indeed you can, I tested it myself and it's what I am doing here. I think you are right it makes little sense to have this, I'll close it anyways

luke14free:
```python

vae = AutoencoderKLWan.from_pretrained(
    "Wan-AI/Wan2.2-T2V-A14B-Diffusers", 
    subfolder="vae", 
    torch_dtype=torch.float32
)

transformer_low_noise = WanTransformer3DModel.from_pretrained(
    "Wan-AI/Wan2.2-T2V-A14B-Diffusers",
    subfolder="transformer_2",
    torch_dtype=torch.bfloat16,
)

self.pipe = WanPipeline.from_pretrained(
    "Wan-AI/Wan2.2-T2V-A14B-Diffusers",
    vae=vae,
    transformer=transformer_low_noise,
    boundary_ratio=None,
    transformer_2=None,
    torch_dtype=torch.bfloat16
)

output = pipe(
    prompt=input_data.prompt,
    negative_prompt=input_data.negative_prompt,
    height=height,
    width=width,
    num_frames=1,  # Required for text-to-image to create proper temporal dimension
    guidance_scale=input_data.guidance_scale,
    num_inference_steps=input_data.num_inference_steps,
).frames[0]
```

for anyone interested, this is how you do it with the existing pipelines :)

PR #12106, opened 2025-08-09
[tests] Add inference test slices for SD3 and remove unnecessary tests

a-r-r-o-w:
In preparation for attention backend refactor, we should ensure that the inference slices match.

Also, there were some tests that don't make sense to test, so I removed those. Let me know if I should bring it back


tests/pipelines/stable_diffusion_3/test_pipeline_stable_diffusion_3.py
```diff
@@ -124,37 +124,22 @@ def get_dummy_inputs(self, device, seed=0):
...
-
-        inputs = self.get_dummy_inputs(torch_device)
-        inputs["prompt_2"] = "a different prompt"
-        inputs["prompt_3"] = "another different prompt"
-        output_different_prompts = pipe(**inputs).images[0]
-
-        max_diff = np.abs(output_same_prompt - output_different_prompts).max()
-
-        # Outputs should be different here
-        assert max_diff > 1e-2
```
sayakpaul, line 141: Makes sense to remove because we are not checking for any assertion slice. We don't get much issues for it, either.

PR #12148, opened 2025-08-14
[docs] Diffusion pipeline

stevhliu:
Refreshes the [Load pipelines](https://huggingface.co/docs/diffusers/main/en/using-diffusers/loading#parallel-loading) doc with more focus on the `DiffusionPipeline` class and loading it.

Main changes:

➕ new device placement section (focusing review on this section would be great!)
➖ `## Checkpoint variants` section will be discussed in the [Model files and layout](https://huggingface.co/docs/diffusers/main/en/using-diffusers/other-formats) doc
➖ `## DiffusionPipeline explained` section will be moved to the Resources section

### DIFF
--- a/docs/source/en/using-diffusers/loading.md
+++ b/docs/source/en/using-diffusers/loading.md
@@ -29,9 +29,8 @@ import torch
 from diffusers import DiffusionPipeline
 
 pipeline = DiffusionPipeline.from_pretrained(
-  "Qwen/Qwen-Image",
-  torch_dtype=torch.bfloat16
-).to("cuda")
+  "Qwen/Qwen-Image", torch_dtype=torch.bfloat16, device_map="cuda"
+)
 ```
 
 Every model has a specific pipeline subclass that inherits from [`DiffusionPipeline`]. A subclass usually has a narrow focus and are task-specific. See the table below for an example.
@@ -49,9 +48,8 @@ import torch
 from diffusers import QwenImagePipeline
 
 pipeline = QwenImagePipeline.from_pretrained(
-  "Qwen/Qwen-Image",
-  torch_dtype=torch.bfloat16
-).to("cuda")
+  "Qwen/Qwen-Image", torch_dtype=torch.bfloat16, device_map="cuda"
+)
 ```
 
 ### Local pipelines
@@ -70,9 +68,8 @@ import torch
 from diffusers import QwenImagePipeline
 
 pipeline = QwenImagePipeline.from_pretrained(
-  "path/to/local/Qwen/Qwen-Image",
-  torch_dtype=torch.bfloat16
-).to("cuda")
+  "path/to/local/Qwen/Qwen-Image", torch_dtype=torch.bfloat16, device_map="cuda"
+)
 ```
 
 The [`~QwenImagePipeline.from_pretrained`] method won't download files from the Hub when it detects a local path. But this also means it won't download and cache any updates that have been made to the model.
@@ -88,8 +85,8 @@ import torch
 from diffusers import HunyuanVideoPipeline
 
 pipeline = HunyuanVideoPipeline.from_pretrained(
-    "hunyuanvideo-community/HunyuanVideo",
-    torch_dtype={"transformer": torch.bfloat16, "default": torch.float16},
+  "hunyuanvideo-community/HunyuanVideo",
+  torch_dtype={"transformer": torch.bfloat16, "default": torch.float16},
 )
 print(pipeline.transformer.dtype, pipeline.vae.dtype)
 ```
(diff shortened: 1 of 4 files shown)


src/diffusers/__init__.py
```diff
@@ -36,6 +36,15 @@
     "configuration_utils": ["ConfigMixin"],
     "guiders": [],
     "hooks": [],
+    "image_processor": [
```
stevhliu: Is it ok to add `image_processor` and `video_processor` to `__init__.py`? Otherwise, my local preview of the docs won't build. No big deal either way though :)

docs/source/en/using-diffusers/loading.md
```diff
@@ -10,598 +10,269 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
+pipeline = QwenImagePipeline.from_pretrained(
+  "Qwen/Qwen-Image", torch_dtype=torch.bfloat16, device_map="cuda"
 )
 ```
 
-### Local pipeline
+### Local pipelines
 
-To load a pipeline locally, use [git-lfs](https://git-lfs.github.com/) to manually download a checkpoint to your local disk.
+Pipelines can also be run locally. Download a checkpoint to your setup with [git-lfs](https://git-lfs.github.com/).
```
sayakpaul: Hmm, maybe we should suggest the use of `snapshot_download()`?

docs/source/en/using-diffusers/loading.md
```diff
@@ -10,598 +10,269 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
-> Some pipeline methods may not function properly on new pipelines created with [`~DiffusionPipeline.from_pipe`]. For instance, the [`~DiffusionPipeline.enable_model_cpu_offload`] method installs hooks on the model components based on a unique offloading sequence for each pipeline. If the models are executed in a different order in the new pipeline, the CPU offloading may not work correctly.
->
-> To ensure everything works as expected, we recommend re-applying a pipeline method on a new pipeline created with [`~DiffusionPipeline.from_pipe`].
+The `device_map` argument determines individual model or pipeline placement on an accelerator like a GPU. It is especially helpful when there are multiple GPUs.
 
-```python
-pipe_sag = StableDiffusionSAGPipeline.from_pipe(
-    pipe_sd
-)
+Diffusers currently provides three options to `device_map`, `"cuda"`, `"balanced"` and `"auto"`. Refer to the table below to compare the three placement strategies.
```
sayakpaul, line 111: `auto` isn't supported for pipelines.

docs/source/en/using-diffusers/loading.md
```diff
@@ -10,598 +10,269 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
+<hfoption id="pipeline">
 
 ```py
-print(f"Max memory allocated: {bytes_to_giga_bytes(torch.cuda.max_memory_allocated())} GB")
-"Max memory allocated: 4.406213283538818 GB"
+import torch
+
+pipeline = DiffusionPipeline.from_pretrained(
+  "black-forest-labs/FLUX.1-dev",
```
sayakpaul: Do we wanna mix Qwen and Flux?

docs/source/en/using-diffusers/loading.md
```diff
@@ -10,598 +10,269 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
+  device_map="cuda",
+)
 ```
 
-Let's animate the image with the [`AnimateDiffPipeline`] and also add a [`MotionAdapter`] module to the pipeline. For the [`AnimateDiffPipeline`], you need to unload the IP-Adapter first and reload it *after* you've created your new pipeline (this only applies to the [`AnimateDiffPipeline`]).
+</hfoption>
+<hfoption id="individual model">
 
 ```py
-from diffusers import AnimateDiffPipeline, MotionAdapter, DDIMScheduler
```
sayakpaul, line 282: Are we deleting these?

docs/source/en/using-diffusers/loading.md
```diff
@@ -10,598 +10,269 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
-| StableDiffusionSAGPipeline | 4.400 |
-| AnimateDiffPipeline | 15.178 |
+Large models are often [sharded](../training/distributed_inference#model-sharding) into smaller files so that they are easier to load. Diffusers supports loading shards in parallel to speed up the loading process.
 
-The [`AnimateDiffPipeline`] has the highest memory requirement, so the *total memory-usage* is based only on the [`AnimateDiffPipeline`]. Your memory-usage will not increase if you create additional pipelines as long as their memory requirements doesn't exceed that of the [`AnimateDiffPipeline`]. Each pipeline can be used interchangeably without any additional memory overhead.
+Set the environment variables below to enable parallel loading.
 
-## Safety checker
+- Set `HF_ENABLE_PARALLEL_LOADING` to `"YES"` to enable parallel loading of shards.
+- Set `HF_PARALLEL_LOADING_WORKERS` to configure the number of parallel threads to use when loading shards. More workers loads a model faster but uses more memory.
```
sayakpaul: We can remove this. We don't need it.
https://github.com/huggingface/diffusers/blob/d03240801f2ac2b4d1f49584c1c5628b98583f6a/src/diffusers/models/model_loading_utils.py#L406

src/diffusers/__init__.py
```diff
@@ -160,6 +160,15 @@
...
     )
+    _import_structure["image_processor"] = [
+        "IPAdapterMaskProcessor",
+        "PixArtImageProcessor",
+        "VaeImageProcessor",
+        "VaeImageProcessorLDM3D",
+    ]
+    _import_structure["video_processor"] = [
+        "VideoProcessor",
+    ]
```
sayakpaul: These changes seem unrelated?

docs/source/en/using-diffusers/loading.md
stevhliu: Good point, I'll switch to Qwen and reduce the overall number of different models we use as examples to avoid confusion!

docs/source/en/using-diffusers/loading.md
stevhliu: Yeah I don't think we need to also show animating the image when we already have an example of using `from_pipe` for PAG. It seems redundant and adds bloat

src/diffusers/__init__.py
stevhliu: It's unrelated, but if it doesn't hurt then it might be more convenient to include it so we can preview the docs locally. Let me know what you think! :)

sayakpaul: Let's separate them out in this PR and open another one?

PR #12151, opened 2025-08-15
Deprecate Flax support

sayakpaul:
Similar to https://github.com/huggingface/transformers/pull/38758/


src/diffusers/models/attention_flax.py
```diff
@@ -151,6 +156,11 @@ class FlaxAttention(nn.Module):
     dtype: jnp.dtype = jnp.float32
 
     def setup(self):
+        logger.warning(
```
sayakpaul, line 159: Thought of adding it to all the public Flax classes, can remove.

src/diffusers/models/attention_flax.py
```diff
@@ -151,6 +156,11 @@ class FlaxAttention(nn.Module):
...
+            "Flax classes are deprecated and will be removed in Diffusers v1. We "
```
DN6: Small nit
```suggestion
            "Flax classes are deprecated and will be removed in Diffusers v1.0.0 We "
```

DN6 (approved):
LGTM 👍🏽

sayakpaul:
@DN6 should we consider removing JAX and Flax from `setup.py` as well?

sayakpaul:
Over to you, @yiyixuxu :)

sayakpaul:
Failing test is unrelated.

DN6:
@sayakpaul Can do in a follow up, but we also need to make torch a required dep.

PR #12161, opened 2025-08-15
make models and pipeline docs device agnostic, to cover accelerators other than CUDA

yao-matrix:
As PyTorch are supporting more accelerators like XPU other than CUDA, and diffusers already device-agnostic, update model and pipeline docs to device agnostic.

@stevhliu , @sayakpaul pls help review, thx very much.


sayakpaul:
I will defer to @stevhliu for taking the final call, but my 2 cents is that we shouldn't be updating with this level of granularity for devices we don't ourselves test. 

So, it might be easier to mention that the same code might run on other accelerators such as NPU without any changes.

yao-matrix:
@stevhliu , pls share you insights. My thought is to make data scientist can work on any accelerator which is compatible w/ HF ecosystem(means added device support and passed ci test cases) using the example codes in docs w/o changes, i suppose it's a smooth first-touch experience.

stevhliu:
Thanks for your PR @yao-matrix! I understand where you're coming from to try and make the experience easier for users who are working with other accelerators besides CUDA and thats awesome 😄 

In this case, I agree with @sayakpaul that it might be a bit aggressive to change at the moment without having tested on these devices ourselves. Instead of making it the default, I would add a note similar to what Sayak suggested.

yao-matrix:
> Thanks for your PR @yao-matrix! I understand where you're coming from to try and make the experience easier for users who are working with other accelerators besides CUDA and thats awesome 😄
> 
> In this case, I agree with @sayakpaul that it might be a bit aggressive to change at the moment without having tested on these devices ourselves. Instead of making it the default, I would add a note similar to what Sayak suggested.

@stevhliu, understand your concern on ci success rate validated by HF. I'll pending it to a more mature time.

As a side note, for Intel XPU, diffusers UT pass rate collected by us is 95.4% (9955 pass over 10435 cases)
 w/ PT 2.7 now(same level of data compared w/ the data we collected on A100 machine).

PR #12169, opened 2025-08-18
Minor modification to support DC-AE-turbo

chenjy2003:
# What does this PR do?

This PR will add minor modification to support DC-AE-turbo into the diffusers lib. `dc-ae-turbo-f32c32-sana-1.1` has the same encoder as `dc-ae-f32c32-sana-1.0` and a lightweight decoder to accelerate decoding.

Could you please have a look? Thanks! @sayakpaul @yiyixuxu
Cc: @lawrence-cj

### DIFF
--- a/src/diffusers/models/transformers/transformer_cogview4.py
+++ b/src/diffusers/models/transformers/transformer_cogview4.py
@@ -28,7 +28,7 @@ from ..cache_utils import CacheMixin
 from ..embeddings import CogView3CombinedTimestepSizeEmbeddings
 from ..modeling_outputs import Transformer2DModelOutput
 from ..modeling_utils import ModelMixin
-from ..normalization import AdaLayerNormContinuous
+from ..normalization import LayerNorm, RMSNorm
 
 
 logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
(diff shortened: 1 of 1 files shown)


src/diffusers/models/autoencoders/autoencoder_dc.py
```diff
@@ -441,6 +448,9 @@ def __init__(
         downsample_block_type: str = "pixel_unshuffle",
         decoder_norm_types: Union[str, Tuple[str]] = "rms_norm",
         decoder_act_fns: Union[str, Tuple[str]] = "silu",
+        encoder_out_shortcut: bool = True,
```
a-r-r-o-w, line 451: Just noting for other reviewers - this change looks safe because previous out_shortcut default was True as well in the encoder

src/diffusers/models/autoencoders/autoencoder_dc.py
```diff
@@ -441,6 +448,9 @@ def __init__(
...
+        decoder_in_shortcut: bool = True,
```
a-r-r-o-w, line 452: This change is safe because previous in_shortcut default was True in decoder

a-r-r-o-w:
Failing tests are unrelated

chenjy2003:
@a-r-r-o-w Thanks so much for your support!

PR #12043, opened 2025-08-01
support `hf_quantizer` in cache warmup.

sayakpaul:
# What does this PR do?

Takes the warmup function close to https://github.com/huggingface/transformers/blob/d3b8627b56caa7ca8fac113c9f28d0256db0194d/src/transformers/modeling_utils.py#L5969

I have gone ahead and also run a snippet from https://github.com/huggingface/diffusers/pull/11904#issue-3217702392 and noticed similar timings. So, running the snippet `main` and this PR branch should yield similar results (not identical because we cannot control it but the difference should be negligible).

PR #12172, opened 2025-08-18
fix: caching allocator behaviour for quantization.

sayakpaul:
# What does this PR do?

Thanks @JoeGaffney for bringing up this issue in https://github.com/huggingface/diffusers/pull/12043#issuecomment-3194404563. Much appreciated. 

This PR fixes it.

## What happened?

<details>
<summary>Expand</summary>

We had a unit mismatch issue in the previous implementation from https://github.com/huggingface/diffusers/pull/12043. We accumulated bytes per device:
`param_byte_count = param.numel() * param.element_size()` -- this is bytes.

But we then passed that byte count directly to `torch.empty` as if it were a number of elements: `torch.empty(byte_count // factor, dtype=dtype, device=device)`. `torch.empty(n, dtype=...)` interprets n as an element count, NOT bytes. That’s too big by a factor of `element_size(target_dtype)` (and often also by `element_size(param)` if they differ).

The previous implementation from https://github.com/huggingface/diffusers/pull/11904 actually does it correctly. It summed elements (`math.prod(param.shape)`) and then allocated exactly that many elements with the chosen dtype. Sorry @a-r-r-o-w! 

I haven't verified yet, but this might be broken in `transformers`, too, from which this is adapted.

</details>

## Results

I have verified that with this PR, the non-quantized checkpoints load with similar speeds and yield similar reserved memory. But it's the quantized checkpoints where things get messy. 

```bash
Allocated: 6704.4 MB, Reserved: 23995.6 MB (`main`)
Allocated: 6701.2 MB, Reserved: 6905.9 MB (https://github.com/huggingface/diffusers/pull/11904)
Allocated: 6704.0 MB, Reserved: 6878.7 MB (this PR)
```

<details>
<summary>Script</summary>

```py
import pytest
import torch
from diffusers import BitsAndBytesConfig, FluxTransformer2DModel


@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA required")
def test_bnb_quantized_model_warmup():
    torch_dtype = torch.bfloat16

    # Quantization config for 4-bit BNB
    quant_config = BitsAndBytesConfig(load_in_4bit=True, bnb_4bit_quant_type="nf4", bnb_4bit_compute_dtype=torch_dtype)

    # Load model (actual warmup path triggered internally)
    model = FluxTransformer2DModel.from_pretrained(
        model_id, subfolder="transformer", quantization_config=quant_config, torch_dtype=torch_dtype
    )
    # model = FluxTransformer2DModel.from_pretrained(
    #     model_id, subfolder="transformer", device_map="cuda", torch_dtype=torch_dtype
    # )

    # Check memory stats
    mem_alloc = torch.cuda.memory_allocated()
    mem_reserved = torch.cuda.memory_reserved()
    print(f"Allocated: {mem_alloc/1e6:.1f} MB, Reserved: {mem_reserved/1e6:.1f} MB")

    # Assert some reasonable range 
    assert mem_alloc > 0, "Model should allocate some GPU memory"
    assert mem_reserved > 0, "Warmup should reserve some GPU memory"
```

Run with `pytest test_memory_allocation.py -vs`.

</details>

Thanks to @JoeGaffney for providing the snippet and for investigating this issue. In a separate PR, I would love to brainstorm a unit test for this with you. Even better -- if you're open to contributing it :) 

@asomoza check it when you have a moment.

### DIFF
--- a/examples/dreambooth/README_qwen.md
+++ b/examples/dreambooth/README_qwen.md
@@ -75,9 +75,9 @@ Now, we can launch training using:
 ```bash
 export MODEL_NAME="Qwen/Qwen-Image"
 export INSTANCE_DIR="dog"
-export OUTPUT_DIR="trained-sana-lora"
+export OUTPUT_DIR="trained-qwenimage-lora"
 
-accelerate launch train_dreambooth_lora_sana.py \
+accelerate launch train_dreambooth_lora_qwenimage.py \
   --pretrained_model_name_or_path=$MODEL_NAME  \
   --instance_data_dir=$INSTANCE_DIR \
   --output_dir=$OUTPUT_DIR \
--- a/src/diffusers/models/autoencoders/autoencoder_dc.py
+++ b/src/diffusers/models/autoencoders/autoencoder_dc.py
@@ -299,6 +299,7 @@ class Decoder(nn.Module):
         act_fn: Union[str, Tuple[str]] = "silu",
         upsample_block_type: str = "pixel_shuffle",
         in_shortcut: bool = True,
+        conv_act_fn: str = "relu",
     ):
         super().__init__()
 
@@ -349,7 +350,7 @@ class Decoder(nn.Module):
         channels = block_out_channels[0] if layers_per_block[0] > 0 else block_out_channels[1]
 
         self.norm_out = RMSNorm(channels, 1e-5, elementwise_affine=True, bias=True)
-        self.conv_act = nn.ReLU()
+        self.conv_act = get_activation(conv_act_fn)
         self.conv_out = None
 
         if layers_per_block[0] > 0:
@@ -414,6 +415,12 @@ class AutoencoderDC(ModelMixin, ConfigMixin, FromOriginalModelMixin):
             The normalization type(s) to use in the decoder.
         decoder_act_fns (`Union[str, Tuple[str]]`, defaults to `"silu"`):
             The activation function(s) to use in the decoder.
+        encoder_out_shortcut  (`bool`, defaults to `True`):
+            Whether to use shortcut at the end of the encoder.
+        decoder_in_shortcut (`bool`, defaults to `True`):
+            Whether to use shortcut at the beginning of the decoder.
+        decoder_conv_act_fn (`str`, defaults to `"relu"`):
+            The activation function to use at the end of the decoder.
         scaling_factor (`float`, defaults to `1.0`):
             The multiplicative inverse of the root mean square of the latent features. This is used to scale the latent
             space to have unit variance when training the diffusion model. The latents are scaled with the formula `z =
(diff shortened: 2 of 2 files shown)


asomoza:
this fixes it for me, no more OOMs

src/diffusers/models/model_loading_utils.py
```diff
@@ -726,23 +726,29 @@ def _caching_allocator_warmup(
...
-            param = model.get_parameter(param_name)
+            p = model.get_parameter(param_name)
         except AttributeError:
-            param = model.get_buffer(param_name)
-        # The dtype of different parameters may be different with composite models or `keep_in_fp32_modules`
-        param_byte_count = param.numel() * param.element_size()
+            try:
+                p = model.get_buffer(param_name)
+            except AttributeError:
+                raise AttributeError(f"Parameter {param_name} not found in model")
```
a-r-r-o-w:
```suggestion
                raise AttributeError(f"Parameter or buffer with name={param_name} not found in model")
```

or something similar maybe since we check both params and buffers

JoeGaffney:
Hey thanks @sayakpaul 


Seems to work on main.
```
BitsAndBytesConfig {
  "_load_in_4bit": true,
  "_load_in_8bit": false,
  "bnb_4bit_compute_dtype": "bfloat16",
  "bnb_4bit_quant_storage": "uint8",
  "bnb_4bit_quant_type": "nf4",
  "bnb_4bit_use_double_quant": false,
  "llm_int8_enable_fp32_cpu_offload": false,
  "llm_int8_has_fp16_weight": false,
  "llm_int8_skip_modules": null,
  "llm_int8_threshold": 6.0,
  "load_in_4bit": true,
  "load_in_8bit": false,
  "quant_method": "bitsandbytes"
}

Before moving to GPU Allocated: 6704.6 MB, Reserved: 6830.4 MB
After moving to GPU Allocated: 6704.6 MB, Reserved: 6830.4 MB
After moving to CPU Allocated: 0.0 MB, Reserved: 0.0 MB
PASSED

TorchAoConfig {
  "modules_to_not_convert": null,
  "quant_method": "torchao",
  "quant_type": "int8_weight_only",
  "quant_type_kwargs": {}
}

Before moving to GPU Allocated: 0.0 MB, Reserved: 0.0 MB
After moving to GPU Allocated: 12014.9 MB, Reserved: 12306.1 MB
After moving to CPU Allocated: 0.0 MB, Reserved: 0.0 MB
PASSED
```

Happy to try and add a unit test. Maybe with some smaller model with known sizes we could do some asserts to validate some values of popular quant configs.

One behaviour thing here though is that BnB seems to Reserve and allocate on load. But TorchAO seems to be only when moving to cuda. Maybe this is expected though.

sayakpaul:
> One behaviour thing here though is that BnB seems to Reserve and allocate on load. But TorchAO seems to be only when moving to cuda. Maybe this is expected though.

This is correct. You NEED a supported accelerator for the weights to initially load for bnb. However, for 4bit quantized models in bnb, you can move them to CPU. Test:
https://github.com/huggingface/diffusers/blob/9918d13eba295d878aa68f5b1ae10aca1a2fc2f6/tests/quantization/bnb/test_4bit.py#L288

TorchAO doesn't have this nature. 

> Happy to try and add a unit test. Maybe with some smaller model with known sizes we could do some asserts to validate some values of popular quant configs.

Sure, let's maybe try adding one for each of bnb and torchao? Thanks a lot in advance.

PR #12176, opened 2025-08-18
[chore] remove extra validation check in determine_device_map

sayakpaul:


src/diffusers/models/model_loading_utils.py
```diff
@@ -112,9 +112,6 @@ def _determine_device_map(
         device_map_kwargs["max_memory"] = max_memory
         device_map = infer_auto_device_map(model, dtype=target_dtype, **device_map_kwargs)
 
-        if hf_quantizer is not None:
-            hf_quantizer.validate_environment(device_map=device_map)
-
```
SunMarc, line 117: why remove that ?

sayakpaul: Because:
https://github.com/huggingface/diffusers/blob/7a2b78bf0f788d311cc96b61e660a8e13e3b1e63/src/diffusers/models/modeling_utils.py#L1278C9-L1280C1

SunMarc: oh indeed ;)

SunMarc (approved):
LGTM !

PR #12177, opened 2025-08-18
[Modular] Provide option to disable custom code loading globally via env variable

DN6:
# What does this PR do?
Based on discussion here:
https://github.com/huggingface/diffusers/discussions/12033

Add an option to globally disable any downloading of remote code via an env variable. This is a bit of a guard against wrapper code around diffusers that might automatically download remote code without the user knowing. 

Note. It's not 100% foolproof since wrapper code could also override the env variable before doing any diffusers imports. I don't have a great solution for this at the moment. For now, I've added a warning for every time remote code is downloaded so that the user is at least alerted when this is happening.   

cc: @JoeGaffney

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/modular_pipelines/modular_pipeline.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline.py
@@ -290,7 +290,7 @@ class ModularPipelineBlocks(ConfigMixin, PushToHubMixin):
     def from_pretrained(
         cls,
         pretrained_model_name_or_path: str,
-        trust_remote_code: Optional[bool] = None,
+        trust_remote_code: bool = False,
         **kwargs,
     ):
         hub_kwargs_names = [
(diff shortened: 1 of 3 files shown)


JoeGaffney:
Hey @DN6,

Thanks for this looks like a solid mechanism.

You’re right that someone could technically do:

```python
os.environ["DIFFUSERS_DISABLE_REMOTE_CODE"] = "0"
import diffusers
```

But in practice, this is mostly fine. The env var is read at import time, so once diffusers is imported, the value is cached and later changes won’t affect behavior.

Making it opt-in is probably a good call to avoid breaking existing workflows. In production or more secure environments, users can enable this to enforce stricter control over remote code execution.

yiyixuxu (approved):
thanks1

PR #12196, opened 2025-08-20
[Modular] implement requirements validation for custom blocks

sayakpaul:
# What does this PR do?

As discussed with Dhruv.

When implementing custom blocks (which we all believe will grow big :)), it can be useful for the block author and the users to have some kind of dependencies and their versions specified. This PR implements that feature. 

Sample output: https://huggingface.co/diffusers-internal-dev/gemini-prompt-expander/blob/main/modular_config.json.

### DIFF
--- a/src/diffusers/modular_pipelines/modular_pipeline.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline.py
@@ -232,7 +232,7 @@ class ModularPipelineBlocks(ConfigMixin, PushToHubMixin):
 
     config_name = "modular_config.json"
     model_name = None
-    _requirements: Union[List[Tuple[str, str], Tuple[str, str]]] = None
+    _requirements: Union[List[Tuple[str, str]], Tuple[str, str]] = None
 
     @classmethod
     def _get_signature_keys(cls, obj):
@@ -275,7 +275,7 @@ class ModularPipelineBlocks(ConfigMixin, PushToHubMixin):
     def _get_requirements(self):
         if getattr(self, "_requirements", None) is not None:
             defined_reqs = self._requirements
-            if not isinstance(defined_reqs):
+            if not isinstance(defined_reqs, list):
                 defined_reqs = [defined_reqs]
 
             final_reqs = []
@@ -290,6 +290,7 @@ class ModularPipelineBlocks(ConfigMixin, PushToHubMixin):
                         f"Version for {pkg} was specified to be {specified_ver} whereas the actual version found is {pkg_actual_ver}. Ignore if this is not concerning."
                     )
                 final_reqs.append((pkg, specified_ver))
+            return final_reqs
 
         else:
             return None
--- a/src/diffusers/modular_pipelines/modular_pipeline.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline.py
@@ -272,29 +272,6 @@ class ModularPipelineBlocks(ConfigMixin, PushToHubMixin):
 
         return input_names
 
-    def _get_requirements(self):
-        if getattr(self, "_requirements", None) is not None:
-            defined_reqs = self._requirements
-            if not isinstance(defined_reqs, list):
-                defined_reqs = [defined_reqs]
-
-            final_reqs = []
-            for pkg, specified_ver in defined_reqs:
-                pkg_available, pkg_actual_ver = _is_package_available(pkg)
-                if not pkg_available:
-                    raise ValueError(
-                        f"{pkg} was specified in the requirements but wasn't found. Please check your environment."
-                    )
-                if specified_ver != pkg_actual_ver:
-                    logger.warning(
-                        f"Version for {pkg} was specified to be {specified_ver} whereas the actual version found is {pkg_actual_ver}. Ignore if this is not concerning."
-                    )
-                final_reqs.append((pkg, specified_ver))
-            return final_reqs
-
-        else:
-            return None
-
     @property
     def required_inputs(self) -> List[InputParam]:
         return self._get_required_inputs()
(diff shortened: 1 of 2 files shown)


sayakpaul:
@DN6 a gentle ping :)

sayakpaul:
Gentle ping @DN6

yiyixuxu:
ummm, should requirements be defined at the pipeline level instead of the block level?
Blocks are meant to be reusable components that can be assembled in different ways. While individual blocks can define default loading behavior (which may have different dependency rquirement), these should be overridable at the pipeline level when blocks are assembled together in modular pipeline

the person who put together the pipeline should be responsible and have the final say over the final components, e.g requirements and modular_model_index.json etc, no?

sayakpaul:
I think implementing it at both the block and pipeline levels could be nice. What I am thinking:

* If a pipeline doesn't define any custom requirements, then we could check if any of its blocks define them and perform validation.
* If a pipeline defines it, then the validation logic, at least, still remains the same. 

WDYT about this approach? 👀

yiyixuxu:
for blocks, maybe define requirements within the block definition similar to inputs/outputs?
```py
class GeminiPromptExpander:
    _requirements = {
        "google-generativeai": ">=0.8.0",
        "pillow": ">=10.0.0"
    }

class FluxTransformer:
    _requirements = {
        "transformers": ">=4.44.0",
        "sentencepiece": ">=0.2.0"
    }
```
When we combine them:
```py
pipe = SequentialPipelineBlocks.from_blocks_dict({
    "prompt_expander": GeminiPromptExpander, 
    "transformer": FluxTransformer
})

print(pipe._requirements)
# Output:
# {
#     "prompt_expander": {"google-generativeai": ">=0.8.0", "pillow": ">=10.0.0"},
#     "transformer": {"transformers": ">=4.44.0", "sentencepiece": ">=0.2.0"}
# }
```
This should be just meta data thought, it just helps generate the final requirements.txt, but the pipeline author still need to manually write the requirement file and test it

sayakpaul:
@yiyixuxu yes that makes sense. I will apply those changes in the PR.

sayakpaul:
@yiyixuxu I have implemented what we discussed above. Individual `ModularPipelineBlocks` can implement a `_requirements` class-level attribute. And then `SequentialPipelineBlocks` now has `_requirements` property, which basically collates all the requirements from its blocks, if available.

```py
from diffusers.modular_pipelines import SequentialPipelineBlocks

class GeminiPromptExpander:
    _requirements = {
        "google-generativeai": ">=0.8.0",
        "pillow": ">=10.0.0"
    }

class FluxTransformer:
    _requirements = {
        "transformers": ">=4.44.0",
        "sentencepiece": ">=0.2.0"
    }

pipe = SequentialPipelineBlocks.from_blocks_dict({
    "prompt_expander": GeminiPromptExpander, 
    "transformer": FluxTransformer
})

print(pipe._requirements)
pipe.save_pretrained(".")
```

LMK.

src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -301,6 +303,19 @@ def from_pretrained(
...
     ):
+        config = cls.load_config(pretrained_model_name_or_path)
+        has_remote_code = "auto_map" in config and cls.__name__ in config["auto_map"]
+        trust_remote_code = resolve_trust_remote_code(
+            trust_remote_code, pretrained_model_name_or_path, has_remote_code
+        )
+        if not (has_remote_code and trust_remote_code):
+            raise ValueError(
+                "Selected model repository does not happear to have any custom code or does not have a valid `config.json` file."
+            )
```
sayakpaul: Same code as previous but just lifted a bit above.

sayakpaul:
Also cc: @stevhliu. Could you help with the docs for this?

src/diffusers/modular_pipelines/modular_pipeline.py
DN6: We  need to be able to pass `hub_kwargs` when loading the config.

src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -382,6 +384,19 @@ def from_pretrained(
         trust_remote_code: bool = False,
         **kwargs,
     ):
+        config = cls.load_config(pretrained_model_name_or_path)
+        has_remote_code = "auto_map" in config and cls.__name__ in config["auto_map"]
+        trust_remote_code = resolve_trust_remote_code(
+            trust_remote_code, pretrained_model_name_or_path, has_remote_code
+        )
```
DN6: We need to pass `hub_kwargs` to `load_config` in order to load revisions, set cache_dir, local_dir etc.

src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -382,6 +384,19 @@ def from_pretrained(
...
+        if not (has_remote_code and trust_remote_code):
```
DN6:
```suggestion
        if not has_remote_code and trust_remote_code:
```

src/diffusers/modular_pipelines/modular_pipeline_utils.py
```diff
@@ -972,6 +974,89 @@ def make_doc_string(
...
+        normalized_reqs = _normalize_requirements(reqs)
+
+    if not normalized_reqs:
+        return {}
+
+    final: dict[str, str] = {}
+    for req, specified_ver in normalized_reqs.items():
+        req_available, req_actual_ver = _is_package_available(req)
+        if not req_available:
+            logger.warning(f"{req} was specified in the requirements but wasn't found in the current environment.")
```
DN6:
```suggestion
            logger.warning(f"{req} was specified but wasn't found in the current environment.")
```

sayakpaul:
@DN6 I have updated the `hub_kwargs` related comments. I decided to keep the other logic as is because I don't think it would hurt to have some standardization, given that it has more benefits. For example, if a developer is including `requirements` then it's better if they do it in a standard way so it's better for the community to prevent reproducibility issues.

src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -1247,6 +1257,14 @@ def doc(self):
             expected_configs=self.expected_configs,
         )
 
+    @property
+    def _requirements(self) -> dict[str, str]:
```
yiyixuxu, line 1270: should we add the same `_requirements` property to `ConditionalPipelineBlocks` & others?

right now, unless it's all `SequentialPipelineBlocks`, the nested requirements get lost, no?

```
block_a (conditional)  ---------------------> block_b 
 -block_a.1  
 -block_a.2
```
block_a's subblock requirement will be lost

the algo might be a bit aggresive for conditional blocks (we'd collect from all branches even though the user may only want to run one), but since this is warn-only I think that's fine

In practice, I think we'd recommend people override `_requirements` directly on their final block class rather than relying on the auto-accumulation, but good to have a default and have this meta data

PR #12203, opened 2025-08-21
[tests] feat: add AoT compilation tests

sayakpaul:
# What does this PR do?

AoT compilation is exciting because it helps cut the framework overhead. It also helps realize similar benefits as JiT compilation for environments where JiT might not be feasible (ZeroGPU Spaces, for example). For example, with AoT-compilation, we were able to obtain 1.7x speedups in latency on a ZeroGPU Space. Wouldn't have been feasible, otherwise.

To run, `RUN_SLOW=1 RUN_COMPILE=yes pytest tests/models/ -k "aot"`.

### DIFF
--- a/tests/models/test_modeling_common.py
+++ b/tests/models/test_modeling_common.py
@@ -2142,17 +2142,9 @@ class TorchCompileTesterMixin:
 
         with tempfile.TemporaryDirectory() as tmpdir:
             package_path = os.path.join(tmpdir, f"{self.model_class.__name__}.pt2")
-            _ = torch._inductor.aoti_compile_and_package(
-                exported_model,
-                package_path=package_path,
-                inductor_configs={
-                    "aot_inductor.package_constants_in_so": False,
-                    "aot_inductor.package_constants_on_disk": True,
-                    "aot_inductor.package": True,
-                },
-            )
+            _ = torch._inductor.aoti_compile_and_package(exported_model, package_path=package_path)
             assert os.path.exists(package_path)
-            loaded_binary = load_package(package_path)
+            loaded_binary = load_package(package_path, run_single_threaded=True)
 
         model.forward = loaded_binary
 
--- a/tests/models/test_modeling_common.py
+++ b/tests/models/test_modeling_common.py
@@ -2059,6 +2059,7 @@ class TorchCompileTesterMixin:
         init_dict, inputs_dict = self.prepare_init_args_and_inputs_for_common()
 
         model = self.model_class(**init_dict).to(torch_device)
+        model.eval()
         model = torch.compile(model, fullgraph=True)
 
         with (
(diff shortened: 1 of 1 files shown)


tests/models/test_modeling_common.py
```diff
@@ -2111,25 +2112,46 @@ def test_compile_with_group_offloading(self):
...
+        from torch._inductor.package import load_package
+
+        init_dict, inputs_dict = self.prepare_init_args_and_inputs_for_common()
+
+        model = self.model_class(**init_dict).to(torch_device)
+        exported_model = torch.export.export(model, args=(), kwargs=inputs_dict)
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            package_path = os.path.join(tmpdir, f"{self.model_class.__name__}.pt2")
+            _ = torch._inductor.aoti_compile_and_package(exported_model, package_path=package_path)
```
angelayi, line 2145: Not passing in a path should also automatically give you a path in the tmp dir!

tests/models/test_modeling_common.py
```diff
@@ -2111,25 +2112,46 @@ def test_compile_with_group_offloading(self):
...
+            assert os.path.exists(package_path)
+            loaded_binary = load_package(package_path, run_single_threaded=True)
+
+        model.forward = loaded_binary
+
+        with torch.no_grad():
+            _ = model(**inputs_dict)
+            _ = model(**inputs_dict)
```
angelayi, line 2153: Is there a particular reason why you're running it twice?

angelayi (commented):
Thanks for adding a test!

tests/models/test_modeling_common.py
sayakpaul: To emulate the real scenario as the model is typically invoked more than once during the actual generation process.

cbensimon:
Very nice! Ideally we should pass multiple inputs to the exported (and/or compiled) model in order to test proper dynamism (for instance in the case of `QwenImageTransformer2DModel`, export test might pass but we'll miss the fact that it will only work at a given image shape). Of course this probably means quite a (too) big refactor of `prepare_init_args_and_inputs_for_common`

sayakpaul:
> Very nice! Ideally we should pass multiple inputs to the exported (and/or compiled) model in order to test proper dynamism (for instance in the case of QwenImageTransformer2DModel, export test might pass but we'll miss the fact that it will only work at a given image shape). Of course this probably means quite a (too) big refactor of prepare_init_args_and_inputs_for_common

We already have this kind of test:
https://github.com/huggingface/diffusers/blob/91a151b5c698836ce3bac85fd9b5a2b10a726c99/tests/models/test_modeling_common.py#L2119

I think it's fine for now to just test for static shapes. If dynamic AoT becomes more popular, we can revisit it.

angelayi (approved):

cbensimon:
> We already have this kind of test:

Very nice! I think that `prepare_dummy_inputs` could actually be useful outside of tests too (for `torch.export.export` input examples)

PR #12204, opened 2025-08-21
Fix typos and inconsistencies

TianqiTang1117:
Summary
- Corrected a typo in the text-to-video guide so it now refers to “memory available on consumer hardware,” improving clarity in the documentation
- Clarified package name inference in the import utilities and fixed a misleading Safetensors log message to point to the correct environment variable
- Strengthened the hook inference test by adding an extra assertion to ensure all computed outputs remain consistent after removing hooks

Notes
The peft dependency could not be installed due to network restrictions, so the test could not be executed.


sayakpaul (approved):
LGTM! thanks!

PR #12206, opened 2025-08-21
Fix PyTorch 2.3.1 compatibility: add version guard for torch.library.…

### LINKED ISSUE #12195: Version 0.35.0 AutoencoderKL failed with PyTorch 2.3.1: AutoAttributeError: module 'torch.library' has no attribute 'custom_op'
### Describe the bug

This should be an old issue, but it appears again in version 0.35.0, maybe also in 0.35.1.
Maybe need to protect the import again if PyTorch 2.3.1 is still in the supporting list?

### Reproduction

from diffusers import AutoencoderKL

import AutoencoderKL will lead to : AutoAttributeError: module 'torch.library' has no attribute 'custom_op'

### Logs

```shell

```

### System Info

Diffusers: 0.35.0
PyTorch: 2.3.1

### Who can help?

_No response_

Aishwarya0811:
## What does this PR do?

Fixes #12195 by adding version guards for `torch.library.custom_op` and `torch.library.register_fake` which are not available in PyTorch 2.3.1.

## Problem
- `torch.library.custom_op` and `torch.library.register_fake` were introduced in PyTorch 2.4
- Users with PyTorch 2.3.1 get `AttributeError: module 'torch.library' has no attribute 'custom_op'` 
- This breaks `from diffusers import AutoencoderKL` and other imports

## Solution  
- Added `hasattr()` checks before using these torch.library functions
- Functions are only registered when available in the PyTorch version
- Maintains backward compatibility with PyTorch 2.3.1 without breaking newer versions

## Testing
- ✅ Tested with PyTorch 2.3.1 - import now works without errors
- ✅ Verified AutoencoderKL imports and functions correctly  
- ✅ No regressions expected with newer PyTorch versions

Fixes #12195

## Who can review?

@sayakpaul @yiyixuxu - This is a PyTorch compatibility fix for core library functionality.

### DIFF
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -109,24 +109,27 @@ if _CAN_USE_XFORMERS_ATTN:
     import xformers.ops as xops
 else:
     xops = None
-    
+
 # Version guard for PyTorch compatibility - custom_op was added in PyTorch 2.4
 if torch.__version__ >= "2.4.0":
     _custom_op = torch.library.custom_op
     _register_fake = torch.library.register_fake
 else:
+
     def custom_op_no_op(name, fn=None, /, *, mutates_args, device_types=None, schema=None):
         def wrap(func):
             return func
+
         return wrap if fn is None else fn
-    
+
     def register_fake_no_op(op, fn=None, /, *, lib=None, _stacklevel=1):
         def wrap(func):
             return func
+
         return wrap if fn is None else fn
-    
+
     _custom_op = custom_op_no_op
-    _register_fake = register_fake_no_op    
+    _register_fake = register_fake_no_op
 
 
 logger = get_logger(__name__)  # pylint: disable=invalid-name
@@ -493,6 +496,7 @@ def _flex_attention_causal_mask_mod(batch_idx, head_idx, q_idx, kv_idx):
 # Registrations are required for fullgraph tracing compatibility
 # TODO: this is only required because the beta release FA3 does not have it. There is a PR adding
 # this but it was never merged: https://github.com/Dao-AILab/flash-attention/pull/1590
+
 
 @_custom_op("flash_attn_3::_flash_attn_forward", mutates_args=(), device_types="cuda")
 def _wrapped_flash_attn_3_original(


a-r-r-o-w (commented):
Hi, thanks for the fix! I think we can improve a bit here by using the approach followed in #11941. Specifically, the following lines:

https://github.com/huggingface/diffusers/blob/0018b62e58c710bca516ad5755762a7c8b7b3127/src/diffusers/models/attention_dispatch.py#L121-L139

Aishwarya0811:
Hi @a-r-r-o-w  
Thanks for the feedback! I've updated the PR to use the dummy decorator approach as suggested, following the pattern from #11941. The code is much cleaner now without the indentation changes. 

Tested with PyTorch 2.3.1 and the import works correctly.

src/diffusers/models/attention_dispatch.py
```diff
@@ -473,12 +473,29 @@ def _flex_attention_causal_mask_mod(batch_idx, head_idx, q_idx, kv_idx):
...
 # Registrations are required for fullgraph tracing compatibility
-
-
-# TODO: library.custom_op and register_fake probably need version guards?
+# Version guard for PyTorch compatibility - custom_op was added in PyTorch 2.4
 # TODO: this is only required because the beta release FA3 does not have it. There is a PR adding
 # this but it was never merged: https://github.com/Dao-AILab/flash-attention/pull/1590
-@torch.library.custom_op("flash_attn_3::_flash_attn_forward", mutates_args=(), device_types="cuda")
+
+if torch.__version__ >= "2.4.0":
```
a-r-r-o-w: Let's move this to the top of the file similar to how it's done in the linked PR, and prefix `custom_op`/`register_fake` with an underscore to mark as private. This allows us to make breaking changes without needing to give out warnings a few releases ahead

src/diffusers/models/attention_dispatch.py
```diff
@@ -473,12 +473,29 @@ def _flex_attention_causal_mask_mod(batch_idx, head_idx, q_idx, kv_idx):
...
+    custom_op = torch.library.custom_op
+    register_fake = torch.library.register_fake
```
a-r-r-o-w:
```suggestion
    _custom_op = torch.library.custom_op
    _register_fake = torch.library.register_fake
```

src/diffusers/models/attention_dispatch.py
```diff
@@ -473,12 +473,29 @@ def _flex_attention_causal_mask_mod(batch_idx, head_idx, q_idx, kv_idx):
...
+            return func
+        return wrap if fn is None else fn
+    
+    def register_fake_no_op(op, fn=None, /, *, lib=None, _stacklevel=1):
+        def wrap(func):
+            return func
+        return wrap if fn is None else fn
+    
+    custom_op = custom_op_no_op
+    register_fake = register_fake_no_op
```
a-r-r-o-w:
```suggestion
    _custom_op = custom_op_no_op
    _register_fake = register_fake_no_op
```

src/diffusers/models/attention_dispatch.py
```diff
@@ -473,12 +473,29 @@ def _flex_attention_causal_mask_mod(batch_idx, head_idx, q_idx, kv_idx):
...
+
+
+@custom_op("flash_attn_3::_flash_attn_forward", mutates_args=(), device_types="cuda")
```
a-r-r-o-w:
```suggestion
@_custom_op("flash_attn_3::_flash_attn_forward", mutates_args=(), device_types="cuda")
```

src/diffusers/models/attention_dispatch.py
```diff
@@ -487,7 +504,7 @@ def _wrapped_flash_attn_3_original(
     return out, lse
 
 
-@torch.library.register_fake("flash_attn_3::_flash_attn_forward")
+@register_fake("flash_attn_3::_flash_attn_forward")
```
a-r-r-o-w:
```suggestion
@_register_fake("flash_attn_3::_flash_attn_forward")
```

a-r-r-o-w (commented):
Thanks for the updates! Just one more comment

src/diffusers/models/attention_dispatch.py
Aishwarya0811: Thanks! I've applied all the suggested changes using the commit suggestions. The version check is now moved to the top and uses private naming with underscores.

Aishwarya0811:
Hi @a-r-r-o-w I've addressed all your feedback - moved the version check to the top of the file and used private naming with underscores.

a-r-r-o-w:
@bot /style

PR #12212, opened 2025-08-21
[docs] Pipeline callbacks

stevhliu:
Refreshes the callback docs and adds API references for the officially supported callbacks


asomoza (approved):
thanks, looks good to me, much simpler and cleaner!

PR #12213, opened 2025-08-22
[refactor] Make guiders return their inputs

a-r-r-o-w:


src/diffusers/guiders/tangential_classifier_free_guidance.py
```diff
@@ -89,7 +92,7 @@ def forward(self, pred_cond: torch.Tensor, pred_uncond: Optional[torch.Tensor] =
         if self.guidance_rescale > 0.0:
             pred = rescale_noise_cfg(pred, pred_cond, self.guidance_rescale)
 
-        return pred, {}
+        return pred, guider_inputs
```
yiyixuxu: I would just define a `GuiderOutput`

and do 

```py
return GuiderOutput(pred=pred, pred_cond=pred_cond)
```


and in the denoiser step, we can do this for now since we don't need other variables for SDXL/wan/flux

```py
block_state.noise_pred = components.guider(guider_state)[0]
```

PR #12236, opened 2025-08-25
[core] use `kernels` to support `_flash_3_hub` attention backend

sayakpaul:
# What does this PR do?

Building FA3 from the source is a time-consuming process, and we could use the [`kernels` lib](https://github.com/huggingface/kernels) for this. 

This PR attempts a dirty implementation of using `kernels` to set the FA3 backend when requested by the user. It's an extremely early implementation, so apologies in advance.

We could let users specify something like `set_attention_backend("kernels-community/vllm-flash-attn3", interface="flash_attn_func")` when they don't have the FA3 build locally available. But that's a matter for our discussion.

Additionally, this also helps keep `diffusers` Hub-first as `kernels` provides a great way to leverage the platform, IMO.

Minimal code to test:

```py
import torch

    model_id, torch_dtype=torch.bfloat16
).to("cuda")

pipe.transformer.set_attention_backend("_flash_3_hub")

prompt = "A cat holding a sign that says 'hello world'"
image = pipe(prompt, num_inference_steps=28, guidance_scale=4.0).images[0]

image.save("output.png")
```

Some comments are inline.


src/diffusers/models/attention_dispatch.py
```diff
@@ -346,7 +355,7 @@ def _check_attention_backend_requirements(backend: AttentionBackendName) -> None
             )
 
     elif backend in [AttentionBackendName._FLASH_3, AttentionBackendName._FLASH_VARLEN_3]:
-        if not _CAN_USE_FLASH_ATTN_3:
+        if not _CAN_USE_FLASH_ATTN_3 and (flash_attn_3_func is None and flash_attn_3_varlen_func is None):
```
sayakpaul: Can modify the logic of "_CAN_USE_FLASH_ATTN_3", too.

src/diffusers/models/attention_dispatch.py
```diff
@@ -636,24 +645,34 @@ def _flash_attention_3(
...
+        "k_descale": None,
+        "v_descale": None,
+        "window_size": window_size,
+        "attention_chunk": 0,
+        "softcap": softcap,
+        "num_splits": 1,
+        "pack_gqa": None,
+        "deterministic": deterministic,
+        "sm_margin": 0,
+    }
```
sayakpaul: `attention_chunk` isn't an acceptable argument in the VLLM implementation. We set it to `0`, anyway.

a-r-r-o-w: I think we can just remove attention_chunk. It defaults to 0 in the original FA3 implementation so there won't be a change in behavior if we remove it. Looking up signature of function and popping out unnecessary kwargs at each block on each forward seems like it'll have some CPU overhead

a-r-r-o-w (commented):
Let's not do it this way i.e. changing the behavior when FA3 is not installed. Instead, let's add new backend `_flash_3_hf` (or something similar) so that user has to explicitly set it to download the kernel. As an end user, running any remote code should require some form of explicit consent imo, and this approach is better than defaulting to downloading from Hub.

src/diffusers/models/attention_dispatch.py
```diff
@@ -514,6 +540,22 @@ def _(query: torch.Tensor, key: torch.Tensor, value: torch.Tensor) -> Tuple[torc
     return torch.empty_like(query), query.new_empty(lse_shape)
 
 
+@_custom_op("vllm_flash_attn3::_flash_attn_forward", mutates_args=(), device_types="cuda")
```
sayakpaul: Currently, fullgraph tracing for compile is failing with:

```bash
torch._dynamo.exc.Unsupported: Operator does not support running with fake tensors
  Explanation: 
  Hint: see https://docs.google.com/document/d/1GgvOe7C8_NVOMLOCwDaYV1mXXyHMXY7ExoewHqooxrs/edit#heading=h.64r4npvq0w0 for how to fix

  Developer debug context: unsupported operator: _vllm_flash_attn3_28fbd26_dirty.fwd.default
```

I am referred to https://docs.pytorch.org/tutorials/advanced/python_custom_ops.html but couldn't figure how to craft this input. 

Any advice? @anijain2305 

Code:

```py
import torch

    model_id, torch_dtype=torch.bfloat16
).to("cuda")

pipe.transformer.set_attention_backend("_flash_3_hub")

prompt = "A cat holding a sign that says 'hello world'"

    image = pipe(
        prompt, num_inference_steps=28, guidance_scale=4.0, generator=torch.manual_seed(0)
    ).images[0]
    image.save("output.png")
```

src/diffusers/models/attention_dispatch.py
```diff
@@ -67,6 +70,20 @@
     flash_attn_3_func = None
     flash_attn_3_varlen_func = None
 
+if is_kernels_available():
+    from ..utils.kernels_utils import _get_fa3_from_hub
+
+    flash_attn_interface_hub = _get_fa3_from_hub()
```
a-r-r-o-w, line 78: This still has the same problem as before, no? Unless I explicitly want to use FA3 from the Hub, it shouldn't be downloaded imo. The behavior looks like it will always download the kernel as soon as this file ends up being imported.

Maybe let's trigger this download only when user calls `ModelMixin::set_attention_backend` with `*_hub` backend. If `*_hub` is specified, a function in `attention_dispatch.py`, that has access to the required local hub function names (`_flash_attn_interface_hub`, etc.), can be called to update them with the imported hub kernel modules

src/diffusers/models/attention_dispatch.py
```diff
@@ -132,8 +149,6 @@ def wrap(func):
     _register_fake = register_fake_no_op
 
 
-logger = get_logger(__name__)  # pylint: disable=invalid-name
-
```
a-r-r-o-w: Let's undo this change?

src/diffusers/models/attention_dispatch.py
```diff
@@ -153,6 +168,8 @@ class AttentionBackendName(str, Enum):
     FLASH_VARLEN = "flash_varlen"
     _FLASH_3 = "_flash_3"
     _FLASH_VARLEN_3 = "_flash_varlen_3"
+    _FLASH_3_HUB = "_flash_3_hub"
+    _FLASH_VARLEN_3_HUB = "_flash_varlen_3_hub"  # not supported yet.
```
a-r-r-o-w: We can comment it out for now

src/diffusers/models/attention_dispatch.py
```diff
@@ -514,6 +544,22 @@ def _(query: torch.Tensor, key: torch.Tensor, value: torch.Tensor) -> Tuple[torc
     return torch.empty_like(query), query.new_empty(lse_shape)
 
 
+@_custom_op("vllm_flash_attn3::_flash_attn_forward", mutates_args=(), device_types="cuda")
```
a-r-r-o-w: (nit)

```suggestion
@_custom_op("vllm_flash_attn3::flash_attn", mutates_args=(), device_types="cuda")
```

because this calls an autograd function that has both forward and backward: https://huggingface.co/kernels-community/vllm-flash-attn3/blob/main/torch-ext/vllm_flash_attn3/flash_attn_interface.py#L546

src/diffusers/models/attention_dispatch.py
```diff
@@ -514,6 +544,22 @@ def _(query: torch.Tensor, key: torch.Tensor, value: torch.Tensor) -> Tuple[torc
...
+def _wrapped_flash_attn_3_hub(
+    query: torch.Tensor, key: torch.Tensor, value: torch.Tensor
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    out, lse = flash_attn_3_hub_func(query, key, value)
+    lse = lse.permute(0, 2, 1)
+    return out, lse
+
+
+@_register_fake("vllm_flash_attn3::_flash_attn_forward")
```
a-r-r-o-w:
```suggestion
@_register_fake("vllm_flash_attn3::flash_attn")
```

same nit as above

src/diffusers/utils/kernels_utils.py
```diff
@@ -0,0 +1,17 @@
...
+    if not is_kernels_available():
+        return None
+    else:
+        from kernels import get_kernel
+
+        try:
+            flash_attn_3_hub = get_kernel(_DEFAULT_HUB_ID_FA3)
+            return flash_attn_3_hub
+        except Exception:
+            return None
```
a-r-r-o-w: Doing this kind of blocks the actual error from propagating to the user, so it will be hard to debug what the problem is. As this is the function trying to fetch from a remote server, handling any errors here makes more sense to me instead of returning a None. Maybe something like:

```python
def _get_fa3_from_hub():
    if not is_kernels_available():
        raise EnvironmentError("message about kernels package not being installed")
    else:
        from kernels import get_kernel

        try:
            flash_attn_3_hub = get_kernel(_DEFAULT_HUB_ID_FA3)
            return flash_attn_3_hub
        except Exception as e:
            logger.error(f"An error occurred while fetching kernel \"{_DEFAULT_HUB_ID_FA3}\" from the Hub: {e}")
            raise // re-raise so that user has full traceback 
```

IMO, this function should only be called if the user intends to use FA3 from the Hub, instead of being called always during the first import of attention dispatcher

a-r-r-o-w (commented):
thanks, looking much better!

sayakpaul:
Getting the following error when trying to compile:

<details>
<summary>Expand</summary>

```bash
File "/fsx/sayak/diffusers/check_fa3_backend.py", line 18, in <module>
    image = pipe(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
  File "/fsx/sayak/diffusers/src/diffusers/pipelines/flux/pipeline_flux.py", line 919, in __call__
    noise_pred = self.transformer(
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1771, in _wrapped_call_impl
    return self._compiled_call_impl(*args, **kwargs)  # type: ignore[misc]
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/_dynamo/eval_frame.py", line 745, in compile_wrapper
    raise e.with_traceback(None) from e.__cause__  # User compiler error
torch._dynamo.exc.Unsupported: torch.* op returned non-Tensor
  Explanation: torch.* ops that return a non-Tensor cannot be traced into the Dynamo FX graph output


  Developer debug context: example_value type: bool; op: call_function; target: <function compiled_with_cxx11_abi at 0x7f67768c0160>


from user code:
   File "/fsx/sayak/diffusers/src/diffusers/models/transformers/transformer_flux.py", line 720, in forward
    encoder_hidden_states, hidden_states = block(
  File "/fsx/sayak/diffusers/src/diffusers/models/transformers/transformer_flux.py", line 443, in forward
    attention_outputs = self.attn(
  File "/fsx/sayak/diffusers/src/diffusers/models/transformers/transformer_flux.py", line 342, in forward
    return self.processor(self, hidden_states, encoder_hidden_states, attention_mask, image_rotary_emb, **kwargs)
  File "/fsx/sayak/diffusers/src/diffusers/models/transformers/transformer_flux.py", line 116, in __call__
    hidden_states = dispatch_attention_fn(
  File "/fsx/sayak/diffusers/src/diffusers/models/attention_dispatch.py", line 293, in dispatch_attention_fn
    return backend_fn(**kwargs)
  File "/fsx/sayak/diffusers/src/diffusers/models/attention_dispatch.py", line 717, in _flash_attention_3_hub
    out, lse, *_ = flash_attn_3_hub_func(
  File "/fsx/sayak/diffusers/src/diffusers/models/attention_dispatch.py", line 226, in flash_attn_3_hub_func
    return _load_fa3_hub().flash_attn_func(*args, **kwargs)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torch/_dynamo/polyfills/__init__.py", line 193, in getattr_and_trace
    return fn(*args[2:], **kwargs)
  File "/fsx/sayak/diffusers/src/diffusers/models/attention_dispatch.py", line 217, in _load_fa3_hub
    fa3_hub = _get_fa3_from_hub()  # won't re-download if already present
  File "/fsx/sayak/diffusers/src/diffusers/utils/kernels_utils.py", line 18, in _get_fa3_from_hub
    flash_attn_3_hub = get_kernel(_DEFAULT_HUB_ID_FA3)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/kernels/utils.py", line 234, in get_kernel
    package_name, package_path = install_kernel(repo_id, revision=revision)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/kernels/utils.py", line 117, in install_kernel
    variant = build_variant()
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/kernels/utils.py", line 64, in build_variant
    cxxabi = "cxx11" if torch.compiled_with_cxx11_abi() else "cxx98"

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"
```

</details>

Maybe we can investigate this in a future PR.

Code:

```
import torch

    model_id, torch_dtype=torch.bfloat16
).to("cuda")

pipe.transformer.set_attention_backend("_flash_3_hub")

prompt = "A cat holding a sign that says 'hello world'"

    image = pipe(
        prompt, num_inference_steps=28, guidance_scale=4.0, generator=torch.manual_seed(0)
    ).images[0]
    image.save("output.png")
```

Cc: @anijain2305

src/diffusers/models/attention_dispatch.py
anijain2305: This means that the custom op is not compatible with torch.compile. There are 2 ways to go

1) Short term - Allow graph breaks (fullgraph=False) - This will compile around the custom op.
2) Long term - Make the custom op torch.compile compatible - https://docs.pytorch.org/tutorials/advanced/custom_ops_landing_page.html - This requires some serious work.

sayakpaul:
The recompilation issues are gone thanks to a recent fix from @danieldk. I will button up this PR and let you know once it’s ready for another review. @a-r-r-o-w

src/diffusers/models/attention_dispatch.py
```diff
@@ -207,6 +210,22 @@ def list_backends(cls):
...
+    fa3_hub = _get_fa3_from_hub()  # won't re-download if already present
+    if fa3_hub is None:
+        raise RuntimeError(
+            "Failed to load FlashAttention-3 kernels from the Hub. Please ensure the wheel is available for your platform."
+        )
+    return fa3_hub
+
+
+def flash_attn_3_hub_func(*args, **kwargs):
+    return _load_fa3_hub().flash_attn_func(*args, **kwargs)
```
DN6: Rather than loading the kernel every time we invoke flash attention, it would be better to import the function at the top of the file, similar to the other FA backends. 

https://github.com/huggingface/diffusers/blob/e58711e73cba70c1c02aaa67f80945a1458901b8/src/diffusers/models/attention_dispatch.py#L63-L68

sayakpaul: Sure but see
https://github.com/huggingface/diffusers/pull/12236#discussion_r2300940273

We shouldn’t make remote calls to the Hub unless requested.

DN6: We can enable user permission via env variable similar to GGUF kernels
https://github.com/huggingface/diffusers/blob/87b800e1546ecd1819a1eed4bfdf22e22f126588/src/diffusers/quantizers/gguf/utils.py#L33

sayakpaul: God that simplifies a lot of stuff. Thanks, Dhruv!

DN6: Actually wondering if we just add a `DIFFUSERS_ENABLE_HUB_KERNELS` constant that is used for all kernel cases.

sayakpaul:
@DN6 I have followed the ENV var approach to allow users to load from the Hub for using pre-built FA3 kernels. Thanks to @danieldk for pushing the changes for op registrations -- `torch.compile` with fullgraph traceability now works perfectly fine.

Could you please review? @a-r-r-o-w too.

src/diffusers/models/attention_dispatch.py
```diff
@@ -67,6 +67,17 @@
     flash_attn_3_func = None
     flash_attn_3_varlen_func = None
 
+if DIFFUSERS_ENABLE_HUB_KERNELS:
```
sayakpaul, line 71: Made it a constant in `constants.py` as I think it will be shared across modules.

src/diffusers/models/attention_dispatch.py
```diff
@@ -657,6 +674,42 @@ def _flash_attention_3(
...
+    key: torch.Tensor,
+    value: torch.Tensor,
+    scale: Optional[float] = None,
+    is_causal: bool = False,
+    window_size: Tuple[int, int] = (-1, -1),
+    softcap: float = 0.0,
+    deterministic: bool = False,
+    return_attn_probs: bool = False,
+) -> torch.Tensor:
+    out = flash_attn_3_func_hub(
```
sayakpaul, line 700: Follow (internal) [this link](https://huggingface.slack.com/archives/C083BS3K2F9/p1756395601227149?thread_ts=1756282399.872609&cid=C083BS3K2F9)

src/diffusers/models/attention_dispatch.py
```diff
@@ -129,7 +141,6 @@ def wrap(func):
         return wrap if fn is None else fn
 
     _custom_op = custom_op_no_op
-    _register_fake = register_fake_no_op
```
a-r-r-o-w: Removed by mistake?

sayakpaul: Good catch. Sorry. Fixed.

DN6 (approved):
Thanks 👍🏽

PR #12237, opened 2025-08-25
[docs] Reproducibility

stevhliu:
Refreshes the reproducibility docs with a better explanation of whats happening when calling `enable_full_determinism()` and also removes a duplicate entry in the `toctree`.

### DIFF
--- a/docs/source/en/using-diffusers/reusing_seeds.md
+++ b/docs/source/en/using-diffusers/reusing_seeds.md
@@ -23,8 +23,6 @@ Pipelines rely on [torch.randn](https://pytorch.org/docs/stable/generated/torch.
 > [!TIP]
 > If reproducibility is important to your use case, we recommend always using a CPU `Generator`. The performance loss is often negligible and you'll generate more similar values.
 
-The `Generator` object should be passed to the pipeline instead of an integer seed. `Generator` maintains a *random state* that is consumed and modified when used. Once consumed, the same `Generator` object produces different results in subsequent calls, even across different pipelines, because it's *state* has changed.
-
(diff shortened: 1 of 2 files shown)


docs/source/en/using-diffusers/reusing_seeds.md
```diff
@@ -10,129 +10,78 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
-```
-
-Running the code above prints one value, but if you run it again you get a different value.
-
-Each time the pipeline is run, [torch.randn](https://pytorch.org/docs/stable/generated/torch.randn.html) uses a different random seed to create the Gaussian noise tensors. This leads to a different result each time it is run and enables the diffusion pipeline to generate a different random image each time.
+> [!TIP]
+> If reproducibility is important to your use case, we recommend always using a CPU `Generator`. The performance loss is often negligible and you'll generate more similar values.
 
-But if you need to reliably generate the same image, that depends on whether you're running the pipeline on a CPU or GPU.
+The `Generator` object should be passed to the pipeline instead of an integer seed. `Generator` maintains a *random state* that is consumed and modified when used. Once consumed, the same `Generator` object produces different results in subsequent calls, even across different pipelines, because it's *state* has changed.
```
sayakpaul: Very important point. Maybe we could even supplement an example here:

```diff
generator = torch.manual_seed(0)

for _ in range(5):
-    image = pipe(prompt, generator=generator)
+    image = pipe(prompt, generator=torch.manual_seed(0))
```

PR #12242, opened 2025-08-26
[packaging] make `torch` a required dependency

sayakpaul:
# What does this PR do?

As discussed, this PR makes `torch` a required dependency and ~removes flax, jax, and jaxlib from the setup.py to keep things simpler (followup from https://github.com/huggingface/diffusers/pull/12151)~.

We cannot remove Flax yet because we have some backends registered for handling the pipeline imports.


sayakpaul:
@DN6 a gentle ping.

DN6:
Think it's best to merge this once we have removed all the Flax code actually. Since you can technically still use Diffusers without torch.

sayakpaul:
Converted to a draft!

PR #12244, opened 2025-08-26
[docs] Remove Flax

stevhliu:
Follows up on #12151 and removes Flax from the docs 🧼


sayakpaul (approved):
🧹

Let's also try to button up the README a bit?

PR #12245, opened 2025-08-27
Fix Table Rendering in ReadME

ParagEkbote:
# What does this PR do?
This PR is a nit fix to fix the rendering of the readme table for community pipelines.

The text : `To load a custom pipeline you just need to pass the custom_pipeline argument to DiffusionPipeline, as one of the files in diffusers/examples/community. Feel free to send a PR with your own pipelines, we will merge them quickly.` 
was being added in the table which is not required.

Could you please review?

## Before submitting

## Who can review?
@stevhliu


stevhliu (approved):
Thanks for the fix!

PR #12246, opened 2025-08-27
[docs] Schedulers

stevhliu:
Refactors the schedulers docs:

- combine `Load schedulers and models` with the `Scheduler` guide so all the info is in one place
- clarify a bit the difference between schedules, spacing, and sigmas (found it to be a bit confusing before)


sayakpaul (approved):
Very very nice!

asomoza:
@stevhliu very nice! I have a couple of comments:

- for FlowMatch models we don't have that much choices, so maybe we can list them? All in the spirit to give the user clear understanding of the schedulers, we can use all the schedulers with SD1.5 and SDXL but for SD3, Flux and all the newer models, we're pretty much stuck with Euler and a couple more.
- For realistic images the best one IMO is DMD2 with the LCM scheduler for SDXL, maybe we can show that combination so people are more aware of it.
- For anime, most people use Euler or Euler Ancestral since it gives a more flat and softer look.
- `TCD` is also a good scheduler for distilled models.
- The all rounder one IMO is `DPMSolverMultistepScheduler` with `sde-dpmsolver++` and `karras sigmas`

This is just my experience and suggestions, feel free to ignore them or add them, but mostly I recommend that we add some schedulers recommendations for the low step inference loras/models.

PR #12250, opened 2025-08-28
Add WanLowNoiseUpscalePipeline and upscale_image to WAN pipeline module

### LINKED ISSUE #12173: Wan i2i with latent upscale
### Model/Pipeline/Scheduler description

I made a wan i2i pipeline that can also somewhat handle upscaling (not really a true upscaler since it's very creative). It only uses the low noise transformer and can generate good results in 20 steps. It's useful to transform low quality pictures and to change the overall vibe of the composition leveraging the low noise transformer in Wan to add details.

A few examples:

* small upscale, simple mood change: https://1nf.sh/tasks/30dt0n0npdx5fa434cpryfs7je
* 5x upscale from low res (200px) : https://1nf.sh/tasks/19t5ah8bf27y8fqwsdyh058hz4

### Open source status

- [x] The model implementation is available.
- [x] The model weights are available (Only relevant if addition is not a scheduler).

### Provide useful links for the implementation

Not really sure there is appetite to have it in diffusers but in case code is here if you want to play with it https://gist.github.com/luke14free/33365d5fc4ffc1b83e18a5d0c4094b1a - its not perfect especially at low noise strengths & high scales & low cfg (it's not a good upscaler unless you allow it to be creative). If you happen to try it, feedback on the implementation is also appreciated, especially around those low strengths.

MengAiDev:
Add WanLowNoiseUpscalePipeline and upscale_image to WAN pipeline module
Fixes #12173
Thanks for this: https://gist.github.com/luke14free/33365d5fc4ffc1b83e18a5d0c4094b1a

@yiyixuxu @asomoza

### DIFF
--- a/src/diffusers/__init__.py
+++ b/src/diffusers/__init__.py
@@ -592,12 +592,14 @@ else:
             "VisualClozePipeline",
             "VQDiffusionPipeline",
             "WanImageToVideoPipeline",
+            "WanLowNoiseUpscalePipeline",
             "WanPipeline",
             "WanVACEPipeline",
             "WanVideoToVideoPipeline",
             "WuerstchenCombinedPipeline",
             "WuerstchenDecoderPipeline",
             "WuerstchenPriorPipeline",
+            "upscale_image",
         ]
     )
 
--- a/src/diffusers/pipelines/__init__.py
+++ b/src/diffusers/pipelines/__init__.py
@@ -380,7 +380,7 @@ else:
         "WuerstchenDecoderPipeline",
         "WuerstchenPriorPipeline",
     ]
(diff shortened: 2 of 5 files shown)


luke14free:
I have made a few changes to the pipeline since I opened the issue, again my point here was to gather feedback on wether it would have been useful before opening a PR myself. that version performs poorly with high scale and low guidance/strength as per my original issue, I don't think you fixed that.

MengAiDev:
@luke14free I see you have made a PR that has be closed before, solving this problem is not easy, and I just use the code from the gist. I will close this PR now before I find the solution.

PR #12264, opened 2025-09-01
[Flux LoRA] fix for prior preservation and mixed precision sampling, follow up on #11873

linoytsaban:
propagates changes from #11873 to other flux lora scripts


linoytsaban:
@bot /style

sayakpaul (approved):
Thanks!

PR #12267, opened 2025-09-02
fix many type hint errors

DefTruth:
I'm the maintainer of cache-dit. cache-dit relies on type annotations to match the forward pattern of blocks more precisely. However, I found that there are a large number of type annotation errors in diffusers. For instance, HiDream, HunyuanVideo, cogvideox, auraflow, cogview, lumina, etc. This PR has fixed all related errors.

PTAL ~ @a-r-r-o-w @sayakpaul 

for example, fix hidream type hint:

```python
torch.Tensor -> Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]
```


sayakpaul:
Why close it?

DefTruth:
> Why close it?

i have open a new pr: #12289

PR #12271, opened 2025-09-03
Deprecate slicing and tiling methods from `DiffusionPipeline`

sayakpaul:
# What does this PR do?

Calling `pipe.enable_vae_tiling()` is same calling `pipe.vae.enable_tiling()`. Same applies for slicing, too.

We can remove these methods to encourage users to be more explicit (in this case even the line of code doesn't change).

This PR starts a deprecation cycle for these methods.

### DIFF
--- a/examples/community/pipeline_faithdiff_stable_diffusion_xl.py
+++ b/examples/community/pipeline_faithdiff_stable_diffusion_xl.py
@@ -1705,6 +1705,12 @@ class FaithDiffStableDiffusionXLPipeline(
         compute decoding and encoding in several steps. This is useful for saving a large amount of memory and to allow
         processing larger images.
         """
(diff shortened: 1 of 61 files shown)


examples/community/pipeline_faithdiff_stable_diffusion_xl.py
```diff
@@ -1705,6 +1705,12 @@ def enable_vae_tiling(self):
         compute decoding and encoding in several steps. This is useful for saving a large amount of memory and to allow
         processing larger images.
         """
+        depr_message = "Calling `enable_vae_tiling()` on a `DiffusionPipeline` is deprecated and this method will be removed in a future version. Please use `pipe.vae.enable_tiling()`."
```
DN6: Perhaps use "Calling `enable_vae_tiling()` on a `{cls.__name__}` is deprecated`? So the specific pipeline name shows up in the message?

PR #12277, opened 2025-09-03
[tests] disable xformer tests for pipelines it isn't popular.

sayakpaul:
# What does this PR do?

https://github.com/huggingface/diffusers/actions/runs/17422018837/job/49462106080#step:7:1297

We can look into these if there's enough community interest. Otherwise, SDPA, Flash should be enough.


sayakpaul:
Gently pinging @DN6

DN6 (approved):

PR #12281, opened 2025-09-03
[docs] Inference section cleanup

stevhliu:
Removes some duplicate entries in the `toctree`. Also moves the `Controlling image quality` guide to the Community optimizations section as it is entirely focused on FreeU.

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -58,12 +58,6 @@
     title: Batch inference
   - local: training/distributed_inference
     title: Distributed inference
-  - local: using-diffusers/scheduler_features
-    title: Scheduler features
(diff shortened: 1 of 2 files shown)


sayakpaul (approved):
I think we can slowly remove these from our documentation as this has grown quite old.

stevhliu:
Sounds good to me! I can add a label at the top warning that it will be deprecated in the near future. Which of these community optimizations should we remove? FreeU, T-GATE, ToME, and maybe DeepCache?

sayakpaul:
I think ToME, DeepCache, etc. are still relevant in the sense that they help accelerate speed. But I guess we don't have to cater to the deprecation notice immediately.

PR #12283, opened 2025-09-04
[core] feat: support group offloading at the pipeline level

sayakpaul:
# What does this PR do?

```py
apply_group_offloading(
    text_encoder,
    onload_device=onload_device,
    offload_device=offload_device,
    offload_type="leaf_level",
    use_stream=True
)
transformer.enable_group_offload(
    onload_device=onload_device,
    offload_device=offload_device,
    offload_type="leaf_level",
    use_stream=True
)
apply_group_offloading(
    vae,
    onload_device=onload_device,
    offload_device=offload_device,
    offload_type="block_level",
    offload_type="leaf_level",
    use_stream=True
)
```

to

```py
pipe.enable_group_offload(
    onload_device=onload_device,
    offload_device=offload_device,
    offload_type="leaf_level",
    use_stream=True
)
```

Of course, if users still want to apply different offloading techniques to different model-level components, they can easily choose to do so. But IMO, `enable_group_offload()` is an easier entrypoint. 

We can allow users to pass mappings like we do for `quant_mapping` in `PipelineQuantizationConfig` in the future.

Will request for reviews after CI.

## TODOs

- [ ] Docs (cc: @stevhliu)


src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1334,6 +1334,143 @@ def enable_sequential_cpu_offload(self, gpu_id: Optional[int] = None, device: Un
...
+        onload_device: torch.device,
+        offload_device: torch.device = torch.device("cpu"),
+        offload_type: str = "block_level",
+        num_blocks_per_group: Optional[int] = None,
+        non_blocking: bool = False,
+        use_stream: bool = False,
+        record_stream: bool = False,
+        low_cpu_mem_usage=False,
+        offload_to_disk_path: Optional[str] = None,
+        exclude_modules: Optional[Union[str, List[str]]] = None,
```
sayakpaul, line 1348: I think it's okay to expose this as an argument as opposed to how we do `model_cpu_offload_seq`, for example:

https://github.com/huggingface/diffusers/blob/764b62473ac10afd9c52e6b3f3f528f719bc7a34/src/diffusers/pipelines/flux/pipeline_flux.py#L179

This is because model CPU offloading relies on a sequence for device management. I don't think we have that constraint in the case of group offloading.

stevhliu:
Really neat how simplified this has become! 🎉

a-r-r-o-w (approved):
Nice! We could reduce LoC by creating a kwargs dict and passing into both branches, but definitely not a blocker

Tests look good but maybe the two could be combined into one to reduce total run count

PR #12285, opened 2025-09-04
[docs] Distributed inference

stevhliu:
Reduces bloat by removing the `## Model sharding` section because it isn't really about distributed inference (its more selectively loading and deleting models) and doesn't show processing multiple prompts in parallel.


docs/source/en/training/distributed_inference.md
```diff
@@ -12,51 +12,55 @@ specific language governing permissions and limitations under the License.
...
-<Tip>
-
-Refer to this minimal example [script](https://gist.github.com/sayakpaul/cfaebd221820d7b43fae638b4dfa01ba) for running inference across multiple GPUs. To learn more, take a look at the [Distributed Inference with 🤗 Accelerate](https://huggingface.co/docs/accelerate/en/usage_guides/distributed_inference#distributed-inference-with-accelerate) guide.
-
-</Tip>
-
 ## PyTorch Distributed
 
-PyTorch supports [`DistributedDataParallel`](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html) which enables data parallelism.
+PyTorch [DistributedDataParallel](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html) enables [data parallelism](https://huggingface.co/spaces/nanotron/ultrascale-playbook?section=data_parallelism), which replicates the same model on each device, to process different batches of data in parallel.
```
sayakpaul, line 61: Gg!

docs/source/en/training/distributed_inference.md
```diff
@@ -101,139 +105,14 @@ if __name__ == "__main__":
...
+Call `torchrun` to run the inference script and use the `--nproc_per_node` argument to set the number of GPUs to use.
 
 ```bash
 torchrun run_distributed.py --nproc_per_node=2
 ```
 
-> [!TIP]
-> You can use `device_map` within a [`DiffusionPipeline`] to distribute its model-level components on multiple devices. Refer to the [Device placement](../tutorials/inference_with_big_models#device-placement) guide to learn more.
-
-## Model sharding
```
sayakpaul, line 113: What happened here?

stevhliu: I removed it because it seems more like a "recipe" for progressively and strategically fitting models on a GPU by loading and removing them. I don't think a user is really learning anything new/useful about `device_map` here compared to the [device placement](https://huggingface.co/docs/diffusers/main/en/using-diffusers/loading#device-placement) docs to which there is a link at the bottom.

I would suggest removing it or at least moving it to Resources > Task Recipes.

sayakpaul: Removal doesn't sound right as the content is useful IMO. Including it in "Task Recipes" might also hamper its discoverability.

stevhliu: Ok, I added it back :)

sayakpaul: Looks like it's still discarded?

stevhliu: Should be [here](https://github.com/huggingface/diffusers/pull/12285/files#diff-26cf01819f3da268ac83ef8d3b526ef62b16b59cb801ceb5684005be867a62d2R114) now!

sayakpaul (approved):

PR #12234, opened 2025-08-25
[quant] allow `components_to_quantize` to be a non-list for single components

sayakpaul:
# What does this PR do?

Just a small QoL improvement for pipeline-level quantization.

```py
PipelineQuantizationConfig(
    quant_backend="bitsandbytes_4bit",
    quant_kwargs={"load_in_4bit": True, "bnb_4bit_quant_type": "nf4", "bnb_4bit_compute_dtype": torch.bfloat16},
    components_to_quantize="transformer", # instead of ["transformer"]
)
```

PR #12287, opened 2025-09-04
[docs] components_to_quantize

stevhliu:
Companion docs to #12234 to update examples with single model on pipeline-level quantization


sayakpaul (approved):
Thanks!

I changed the base from `main` to my PR branch.

PR #12293, opened 2025-09-05
refactor: improve handling when only masked_image_latents are provided in Flux.1 Fill dev

Men1scus:


src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -681,7 +681,6 @@ def prepare_latents(
         # latent height and width to be divisible by 2.
         height = 2 * (int(height) // (self.vae_scale_factor * 2))
         width = 2 * (int(width) // (self.vae_scale_factor * 2))
-        shape = (batch_size, num_channels_latents, height, width)
```
yiyixuxu, line 684: do you have an script on how to run this with `masked_image_latents`? it seems if `image` is None it won't work here

Men1scus: https://gist.github.com/Men1scus/9f4a9e139e0dfd8e662147760de2d7b1

src/diffusers/pipelines/flux/pipeline_flux_fill.py
```diff
@@ -924,7 +926,7 @@ def __call__(
         latent_timestep = timesteps[:1].repeat(batch_size * num_images_per_prompt)
 
         # 5. Prepare latent variables
-        num_channels_latents = self.vae.config.latent_channels
+        num_channels_latents = self.vae.config.latent_channels if init_image is not None else None
```
yiyixuxu, line 929: why setting it to None here?

src/diffusers/pipelines/flux/pipeline_flux_fill.py
yiyixuxu: why move this code?

Men1scus: My code changes are based on [Sayak Paul’s gist](https://gist.github.com/sayakpaul/23862a2e7f5ab73dfdcc513751289bea).
Because of the RTX 3090’s 24GB VRAM limitation, I preprocess the input image and mask into masked_image_latents before feeding them into the transformer.

src/diffusers/pipelines/flux/pipeline_flux_fill.py
Men1scus: When the `latents` parameter in `self.prepare_latents` is not None:

```python
if latents is not None:
    return latents.to(device=device, dtype=dtype), latent_image_ids
```

it will return early, and `num_channels_latents` will not be used,
because during denoising my VAE has already been deleted, so I cannot get it through `self.vae.config.latent_channels`.

PR #12298, opened 2025-09-08
[tests] cache non lora pipeline outputs.

sayakpaul:
# What does this PR do?

As discussed.


tests/lora/utils.py
```diff
@@ -129,6 +129,34 @@ class PeftLoraLoaderMixinTests:
...
 
+    cached_non_lora_outputs = {}
+
+    @pytest.fixture(scope="class", autouse=True)
+    def cache_non_lora_outputs(self):
+        """
+        This fixture will be executed once per test class and will populate
+        the cached_non_lora_outputs dictionary.
+        """
+        for scheduler_cls in self.scheduler_classes:
```
DN6: It seems like we mostly use a single scheduler for testing with a few exceptions (CogVideoX) 

The LoRA test where scheduler type matters would be LCM LoRAs, and in those tests it looks like LCM scheduler is set directly?
https://github.com/huggingface/diffusers/blob/4067d6c4b64f2b606f9806d4a8b15d5fd5cbea1e/tests/lora/test_lora_layers_sdxl.py#L227

So why not refactor this entirely to just work with a single scheduler, and we can use the pipeline class name as the cache key?

tests/lora/utils.py
```diff
@@ -129,6 +129,34 @@ class PeftLoraLoaderMixinTests:
     text_encoder_target_modules = ["q_proj", "k_proj", "v_proj", "out_proj"]
     denoiser_target_modules = ["to_q", "to_k", "to_v", "to_out.0"]
 
+    cached_non_lora_outputs = {}
+
+    @pytest.fixture(scope="class", autouse=True)
```
DN6: We use setUp and tearDown convention instead of fixtures. I would stick with that for consistency.

DN6: To handle multiple test classes in a single file we can use a dict with keys based on the testing pipeline class name.

tests/lora/utils.py
```diff
@@ -129,6 +129,34 @@ class PeftLoraLoaderMixinTests:
...
+    def cache_non_lora_outputs(self):
```
DN6: Nit. Would name it `get_base_pipeline_output`

tests/lora/utils.py
sayakpaul: I would prefer to keep the scope of this PR fixed to caching only.

tests/lora/utils.py
sayakpaul: > To handle multiple test classes in a single file we can use a dict with keys based on the testing pipeline class name.

Didn't get it.

> We use setUp and tearDown convention instead of fixtures. I would stick with that for consistency.

The issue is in how pytest handles test discovery and execution, especially when mixing it with unittest-style setups. When you run a single test method directly, pytest doesn't automatically execute the setUpClass method. We often run individual tests (or even a group of tests) for better debugging. So, not sure how setup and teardown would work in this setup.

tests/lora/utils.py
sayakpaul: For SD and SDXL, we need to test for both schedulers which is why we default to:
https://github.com/huggingface/diffusers/blob/43459079ab06d4f94f2d93ba7153e2c5310928d3/tests/lora/utils.py#L112

When the utils.py was added, we didn't have many models with LoRA support, hence this was the default to test with. But that has changed. I don't think we have to test for both schedulers even for SD and SDXL since we have:
https://github.com/huggingface/diffusers/blob/43459079ab06d4f94f2d93ba7153e2c5310928d3/tests/lora/test_lora_layers_sdxl.py#L223

But two schedulers are also used for CogVideoX as you mentioned. Do you propose to have a separate test class for that? Something like `CogVideoXDPMSchedulerLoRATests`? 

I agree to this suggestion, generally. But would really prefer to do that in a separate PR.

tests/lora/utils.py
sayakpaul: Tried this:

<details>
<summary>Unfold</summary>

```diff
diff --git a/tests/lora/utils.py b/tests/lora/utils.py
index c9ea203b5..bae588690 100644
--- a/tests/lora/utils.py
+++ b/tests/lora/utils.py
@@ -129,17 +129,21 @@ class PeftLoraLoaderMixinTests:
     text_encoder_target_modules = ["q_proj", "k_proj", "v_proj", "out_proj"]
     denoiser_target_modules = ["to_q", "to_k", "to_v", "to_out.0"]
 
-    cached_non_lora_outputs = {}
+    cached_base_pipe_outs = {}
 
-    @pytest.fixture(scope="class", autouse=True)
-    def cache_non_lora_outputs(self):
-        """
-        This fixture will be executed once per test class and will populate
-        the cached_non_lora_outputs dictionary.
-        """
+    def setUp(self):
+        self.get_base_pipe_outs()
+        super().setUp()
+
+    def get_base_pipe_outs(self):
+        cached_base_pipe_outs = getattr(type(self), "cached_base_pipe_outs", {})
+        all_scheduler_names = [scheduler_cls.__name__ for scheduler_cls in self.scheduler_classes]
+        if cached_base_pipe_outs is not None and all(k in cached_base_pipe_outs for k in all_scheduler_names):
+            return
+
+        cached_base_pipe_outs = cached_base_pipe_outs or {}
         for scheduler_cls in self.scheduler_classes:
-            # Check if the output for this scheduler is already cached to avoid re-running
-            if scheduler_cls.__name__ in self.cached_non_lora_outputs:
+            if scheduler_cls.__name__ in cached_base_pipe_outs:
                 continue
 
             components, _, _ = self.get_dummy_components(scheduler_cls)
@@ -151,11 +155,9 @@ class PeftLoraLoaderMixinTests:
             # explicitly.
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
             output_no_lora = pipe(**inputs, generator=torch.manual_seed(0))[0]
-            self.cached_non_lora_outputs[scheduler_cls.__name__] = output_no_lora
+            cached_base_pipe_outs.update({scheduler_cls.__name__: output_no_lora})
 
-        # Ensures that there's no inconsistency when reusing the cache.
-        yield
-        self.cached_non_lora_outputs.clear()
+        setattr(type(self), "cached_base_pipe_outs", cached_base_pipe_outs)
 
     def get_dummy_components(self, scheduler_cls=None, use_dora=False, lora_alpha=None):
         if self.unet_kwargs and self.transformer_kwargs:
@@ -348,7 +350,7 @@ class PeftLoraLoaderMixinTests:
         Tests a simple inference and makes sure it works as expected
         """
         for scheduler_cls in self.scheduler_classes:
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
     def test_simple_inference_with_text_lora(self):
@@ -363,7 +365,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config=None)
@@ -446,7 +448,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config)
@@ -502,7 +504,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config=None)
@@ -540,7 +542,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config=None)
@@ -572,7 +574,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config=None)
@@ -607,7 +609,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config=None)
@@ -658,7 +660,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config=None)
@@ -709,7 +711,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config=None)
@@ -752,7 +754,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config)
@@ -793,7 +795,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config)
@@ -837,7 +839,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, denoiser = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config)
@@ -875,7 +877,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, denoiser = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config)
@@ -954,7 +956,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
 
             if "text_encoder" in self.pipeline_class._lora_loadable_modules:
                 pipe.text_encoder.add_adapter(text_lora_config, "adapter-1")
@@ -1083,7 +1085,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
 
             pipe.text_encoder.add_adapter(text_lora_config, "adapter-1")
             self.assertTrue(check_if_lora_correctly_set(pipe.text_encoder), "Lora not correctly set in text encoder")
@@ -1140,7 +1142,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
 
             if "text_encoder" in self.pipeline_class._lora_loadable_modules:
                 pipe.text_encoder.add_adapter(text_lora_config, "adapter-1")
@@ -1303,7 +1305,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
 
             if "text_encoder" in self.pipeline_class._lora_loadable_modules:
                 pipe.text_encoder.add_adapter(text_lora_config, "adapter-1")
@@ -1397,7 +1399,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
 
             if "text_encoder" in self.pipeline_class._lora_loadable_modules:
                 pipe.text_encoder.add_adapter(text_lora_config, "adapter-1")
@@ -1641,7 +1643,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             if "text_encoder" in self.pipeline_class._lora_loadable_modules:
@@ -1722,7 +1724,7 @@ class PeftLoraLoaderMixinTests:
                 pipe.set_progress_bar_config(disable=None)
                 _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-                output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+                output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
                 self.assertTrue(output_no_lora.shape == self.output_shape)
 
                 if "text_encoder" in self.pipeline_class._lora_loadable_modules:
@@ -1777,7 +1779,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_dora_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_dora_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_dora_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config)
@@ -1909,7 +1911,7 @@ class PeftLoraLoaderMixinTests:
         pipe.set_progress_bar_config(disable=None)
 
         _, _, inputs = self.get_dummy_inputs(with_generator=False)
-        original_out = self.cached_non_lora_outputs[scheduler_cls.__name__]
+        original_out = self.cached_base_pipe_outs[scheduler_cls.__name__]
 
         no_op_state_dict = {"lora_foo": torch.tensor(2.0), "lora_bar": torch.tensor(3.0)}
         logger = logging.get_logger("diffusers.loaders.peft")
@@ -1955,7 +1957,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
             self.assertTrue(output_no_lora.shape == self.output_shape)
 
             pipe, _ = self.add_adapters_to_pipeline(pipe, text_lora_config, denoiser_lora_config)
@@ -2309,7 +2311,7 @@ class PeftLoraLoaderMixinTests:
         pipe = self.pipeline_class(**components).to(torch_device)
         _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-        output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+        output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
         self.assertTrue(output_no_lora.shape == self.output_shape)
 
         pipe, _ = self.add_adapters_to_pipeline(
@@ -2359,7 +2361,7 @@ class PeftLoraLoaderMixinTests:
             pipe.set_progress_bar_config(disable=None)
             _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-            output_no_lora = self.cached_non_lora_outputs[scheduler_cls.__name__]
+            output_no_lora = self.cached_base_pipe_outs[scheduler_cls.__name__]
 
             if "text_encoder" in self.pipeline_class._lora_loadable_modules:
                 pipe.text_encoder.add_adapter(text_lora_config)

```

</details>

Doesn't work when `pytest -s tests/lora/test_lora_layers_sana.py::SanaLoRATests::test_simple_inference`, complaining:

```
FAILED tests/lora/test_lora_layers_sana.py::SanaLoRATests::test_simple_inference - KeyError: 'FlowMatchEulerDiscreteScheduler'
```

tests/lora/utils.py
DN6: For CogVideo, does the choice of scheduler affect the LoRA features (loading, fusing etc) in any way? We can do it in a separate PR that's fine.

tests/lora/utils.py
DN6: Hmm it might be the order in which the setUp method is invoked. It's probably calling the setUp of the Unitest class instead of the Mixin class because of the ordering here. 
https://github.com/huggingface/diffusers/blob/55f0b3d758602226af01a48059145edd55289cfa/tests/lora/test_lora_layers_sana.py#L32

Changing to this `class SanaLoRATests(PeftLoraLoaderMixinTests, unittest.TestCase)` should work. Although it would also be good to use `type(self).cached_base_pipe_outs[scheduler_cls.__name__]` to show we're trying to access the cache via class attribute and not the instance attribute.

Example Mixin

```py
class PeftLoraLoaderMixinTests:
    def setUp(self):
        self._cache_base_pipeline_output()
        super().setUp()

    def _cache_base_pipeline_output(self):
          # cache the output

    def get_base_pipeline_output(self, scheduler_cls):
        return type(self).cached_base_pipe_outputs[scheduler_cls.__name__]
```

tests/lora/utils.py
sayakpaul: It shouldn't. As I said, I generally agree to your suggestion.

sayakpaul:
@DN6 let's try to merge this soon otherwise, it might end up gathering a bunch of brutal conflicts.

tests/lora/test_lora_layers_flux.py
```diff
@@ -170,8 +167,7 @@ def test_lora_expansion_works_for_absent_keys(self):
         pipe.set_progress_bar_config(disable=None)
         _, _, inputs = self.get_dummy_inputs(with_generator=False)
 
-        output_no_lora = pipe(**inputs, generator=torch.manual_seed(0)).images
-        self.assertTrue(output_no_lora.shape == self.output_shape)
+        output_no_lora = self.cached_non_lora_outputs[self.scheduler_cls.__name__]
```
DN6: Do we need to fetch based on scheduler class name now that we only use a single scheduler?

tests/lora/test_lora_layers_flux.py
```diff
@@ -218,9 +214,7 @@ def test_lora_expansion_works_for_extra_keys(self):
         pipe = pipe.to(torch_device)
         pipe.set_progress_bar_config(disable=None)
         _, _, inputs = self.get_dummy_inputs(with_generator=False)
-
-        output_no_lora = pipe(**inputs, generator=torch.manual_seed(0)).images
-        self.assertTrue(output_no_lora.shape == self.output_shape)
+        output_no_lora = self.cached_non_lora_outputs[self.scheduler_cls.__name__]
```
DN6: More of a nit, but I would create a method in the Mixin called `get_base_pipeline_output` and return from there, rather than directly accessing the cache.

sayakpaul:
@DN6 the final two comments should have been addressed.

tests/lora/utils.py
```diff
@@ -126,13 +126,26 @@ class PeftLoraLoaderMixinTests:
     text_encoder_target_modules = ["q_proj", "k_proj", "v_proj", "out_proj"]
     denoiser_target_modules = ["to_q", "to_k", "to_v", "to_out.0"]
 
-    def get_dummy_components(self, use_dora=False, lora_alpha=None):
+    cached_non_lora_output = None
+
+    @pytest.fixture(scope="class", autouse=True)
```
DN6: Think this might be a leftover from previous refactoring. It's not used anywhere. Think it can be removed.

DN6 (approved):
Nice 👍🏽

PR #12306, opened 2025-09-09
Fix Stable Diffusion 3.x pooled prompt embedding with multiple images

friedrich:
When having more than one prompts and and also `num_images_per_prompt > 1`, pooled prompt embeddings are repeated incorrectly. This leads to the pooled prompt embeddings getting assigned to the wrong images.

While CLIP and T5 prompt embeddings are repeated along the sequence dimension, pooled CLIP embeddings are currently repeated along the batch dimension. This PR fixes this by repeating over the embedding dimension. 

Corresponding fix for Flux.1: https://github.com/huggingface/diffusers/commit/05b38c3c0da266cc1e5c5cbd768d972fcf25ecf0

@yiyixuxu @sayakpaul


yiyixuxu (approved):
thanks!

PR #12312, opened 2025-09-10
[docs] Prompting

stevhliu:
Reworks the "Prompt techniques" guide to include better and clearer prompt writing instructions and cleans up the prompt weighting section a bit.

Also cleans up the "Batched inference" docs and includes example outputs.


docs/source/en/using-diffusers/weighted_prompts.md
```diff
@@ -10,36 +10,92 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
+pipeline(prompt_embeds=prompt_embeds, pooled_prompt_embeds=pooled_prompt_embeds).images[0]
+```
+
+<div class="flex justify-center">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/prompt-embed-sdxl.png"/>
+</div>
+
+Prompt weighting is also supported for adapters like [Textual inversion](./textual_inversion_inference) and [DreamBooth](./dreambooth).
+
 ## Prompt enhancing with GPT2
```
stevhliu: @asomoza, is this still a useful or popular technique? Seems like quite a bit of work to code up the prompt enhancer (coming up with the words and styles) when you can just write a prompt with those same words in the first place.

No strong opinions either way, just looking for opportunities to streamline the docs a bit more :)

sayakpaul:
@asomoza could you give this a review?

docs/source/en/using-diffusers/weighted_prompts.md
asomoza: this one in particular is not popular now, this was initially a technique used in Fooocus which made it really impressive back in the days, but today you can use very small llms locally or online to do the same and better, we have come a long way, we should probably remove this.

Instead I suggest to link some good models or we can even link to a search like this one: https://huggingface.co/models?sort=downloads&search=prompt+enhancer

There are also some models that have their [own prompt enhance](https://hunyuan-promptenhancer.github.io/)r like [HunyuanImage-2.1](https://huggingface.co/tencent/HunyuanImage-2.1), so this technique is still relevant but now it's being done by the researchers and model owners right out of the box instead of the users.

docs/source/en/using-diffusers/weighted_prompts.md
```diff
@@ -10,36 +10,92 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
...
   </div>
 </div>
 
+Try to be as specific and precise as possible and provide as much context as you can to help a model understand what to generate. For example, consider using specific photography or cinematographic terms such as lens type, focal lengths, camera angles and scene depths.
+
+## Prompt weighting
+
+Prompt weighting emphasizes and de-emphasizes certain words or concepts in a prompt to help a model focus. It works by scaling the attention score for a token by some value, allowing you to fine-tune how much influence a word or concept has.
+
+Diffusers supports prompt weighting through the `prompt_embeds` and `pooled_prompt_embeds` arguments. These arguments accept the scaled text embedding vector. To get these embeddings, we recommend using the [sd_embed](https://github.com/xhinker/sd_embed) library, which also supports longer prompts. Run the command below to install the sd_embed library.
```
asomoza: we should probably do a note here that the only models supported are SD, SDXL, SD3, Cascade and Flux. Flux is kind of a plus but not really need it, all the newer models have a really good prompt adherence that we don't need prompt weighting with them

stevhliu:
Thanks @asomoza! Feedback has been addressed, let me know if there is anything else, otherwise I think we can merge 🙂

PR #12314, opened 2025-09-10
Fix ftfy name error in Wan pipeline

dsocek:
# What does this PR do?

Small fix for `ftfy` name error in Wan pipeline.  

**Before fix:**
If we don't have `ftfy` package, pipe will error out:
`NameError: name 'ftfy' is not defined`

**After fix:**
With or without `ftfy`, pipe runs OK


yiyixuxu (approved):
thanks!

PR #12320, opened 2025-09-11
[docs] Attention backends

stevhliu:
Docs for enabling different attention backends.

### DIFF
--- a/docs/source/en/optimization/attention_backends.md
+++ b/docs/source/en/optimization/attention_backends.md
@@ -16,7 +16,7 @@ specific language governing permissions and limitations under the License. -->
 
 Diffusers provides several optimized attention algorithms that are more memory and computationally efficient through it's *attention dispatcher*. The dispatcher acts as a router for managing and switching between different attention implementations and provides a unified interface for interacting with them.
 
-Available attention implementations include the following.
+Refer to the table below for an overview of the available attention families and to the [Available backends](#available-backends) section for a more complete list.
 
 | attention family | main feature |
 |---|---|
@@ -34,7 +34,7 @@ The [`~ModelMixin.set_attention_backend`] method iterates through all the module
 The example below demonstrates how to enable the `_flash_3_hub` implementation for FlashAttention-3 from the [kernel](https://github.com/huggingface/kernels) library, which allows you to instantly use optimized compute kernels from the Hub without requiring any setup.
 
 > [!TIP]
-> FlashAttention-3 is not supported for non-Hopper architectures, in which case, use FlashAttention (set_attention_backend("flash")).
+> FlashAttention-3 is not supported for non-Hopper architectures, in which case, use FlashAttention with `set_attention_backend("flash")`.
 
 ```py
 import torch


docs/source/en/optimization/attention_backends.md
```diff
@@ -0,0 +1,191 @@
...
+
+This guide will show you how to use the dispatcher to set and use the different attention backends.
+
+## FlashAttention
+
+[FlashAttention](https://github.com/Dao-AILab/flash-attention) reduces memory traffic by making better use of on-chip shared memory (SRAM) instead of global GPU memory so the data doesn't have to travel far. The latest variant, FlashAttention-3, is further optimized for modern GPUs (Hopper/Blackwell) and also overlaps computations and handles FP8 attention better.
+
+There are several available FlashAttention variants, including variable length and the original FlashAttention. For a full list of supported implementations, check the list [here](https://github.com/huggingface/diffusers/blob/5e181eddfe7e44c1444a2511b0d8e21d177850a0/src/diffusers/models/attention_dispatch.py#L163).
+
+The example below demonstrates how to enable the `_flash_3_hub` implementation. The [kernel](https://github.com/huggingface/kernels) library allows you to instantly use optimized compute kernels from the Hub without requiring any setup.
```
sayakpaul:
```suggestion
The example below demonstrates how to enable the `_flash_3_hub` implementation for Flash Attention 3. The [kernels](https://github.com/huggingface/kernels) library allows you to instantly use optimized compute kernels from the Hub without requiring any setup.
```

But we should aso note the hardware restrictions in using FA3 as it's not supported in non-hopper architectures. In that case, regular Flash should be used through `set_attention_backend("flash")`.

docs/source/en/optimization/attention_backends.md
```diff
@@ -0,0 +1,191 @@
...
+
+To restore the default attention backend, call [`~ModelMixin.reset_attention_backend`].
+
+```py
+pipeline.transformer.reset_attention_backend()
+```
+
+## PyTorch native
+
+PyTorch includes a [native implementation](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) of several optimized attention implementations including [FlexAttention](https://pytorch.org/blog/flexattention/), FlashAttention, memory-efficient attention, and a C++ version.
```
sayakpaul: 1. Flex uses a different path in `torch`:
https://github.com/huggingface/diffusers/blob/5e181eddfe7e44c1444a2511b0d8e21d177850a0/src/diffusers/models/attention_dispatch.py#L847

2. The backends that leverage `nn.functional.scaled_dot_product_attention()` can be found in https://github.com/huggingface/diffusers/blob/5e181eddfe7e44c1444a2511b0d8e21d177850a0/src/diffusers/models/attention_dispatch.py (search with `scaled_dot_product_attention(`).

stevhliu:
Sounds good, feel free to contribute to the table with any notes you may have!

I skipped making an additional example at the end to avoid being repetitive since I've already included an example when introducing `set_attention_backend` and with `attention_backend`. I think once users learn that, its easy for them to just plug in the attention backend names.

docs/source/en/optimization/attention_backends.md
```diff
@@ -0,0 +1,97 @@
...
+Diffusers provides several optimized attention algorithms that are more memory and computationally efficient through it's *attention dispatcher*. The dispatcher acts as a router for managing and switching between different attention implementations and provides a unified interface for interacting with them.
+
+Available attention implementations include the following.
+
+| attention family | main feature |
+|---|---|
+| FlashAttention | minimizes memory reads/writes through tiling and recomputation |
+| SageAttention | quantizes attention to int8 |
+| PyTorch native | built-in PyTorch implementation using [scaled_dot_product_attention](./fp16#scaled-dot-product-attention) |
+| xFormers | memory-efficient attention with support for various attention kernels |
```
sayakpaul, line 26: Should we include a column for their original repo links?

docs/source/en/optimization/attention_backends.md
```diff
@@ -0,0 +1,97 @@
...
+> FlashAttention-3 is not supported for non-Hopper architectures, in which case, use FlashAttention (set_attention_backend("flash")).
+
+```py
+import torch
+from diffusers import QwenImagePipeline
+
+pipeline = QwenImagePipeline.from_pretrained(
+    "Qwen/Qwen-Image", torch_dtype=torch.bfloat16, device_map="cuda"
+)
+pipeline.transformer.set_attention_backend("_flash_3_hub")
```
sayakpaul, line 46: Would also add the `pipeline(...)` at the end to make it end-to-end.

docs/source/en/optimization/attention_backends.md
sayakpaul: Ah didn't realize "## Available backends". It's perfect!

PR #12322, opened 2025-09-12
[wip] [core] introduce pipeline-specific mixins to hold common methods

sayakpaul:
# What does this PR do?

> [!IMPORTANT]
> In the interest of a better branch name: https://github.com/huggingface/diffusers/pull/12820

As discussed internally, this PR introduces pipeline-specific Mixin classes that hold common methods shared across different task-specific pipelines. The actual pipelines (such as `StableDiffusionPipeline`, `QwenImagePipeline`, `QwenImageImg2ImgPipeline`, etc.) would then subclass from these mixins.

Examples of such common methods vary from pipeline to pipeline, but some examples include `encode_prompt()`, properties such as `guidance_scale`, etc.

Additionally, we have a couple of methods like `retrieve_latents()` and `retrieve_timesteps()` originated in `stable_diffusion` and we copied them over to other pipelines that use these methods. Same for `calculate_shift` that originated in `flux`. This PR also considers this situation and follows a reasonable approach to treat it (more below).

I know we also discussed moving the specific LoraLoaderMixins to their respective pipeline modules. The PR only does that for Stable Diffusion to gather feedback. Please keep on reading.

> [!NOTE]
> This PR introduces Mixin classes for four popular pipelines: Flux, Qwen, SDXL, and SD. Once the PR is merged, we could work with the community to do this for other influential pipelines.

---

## Guiding principles

A sample structure for `src/diffusers/pipelines/flux`:

```bash
.
├── __init__.py
├── modeling_flux.py
├── pipeline_flux_control_img2img.py
├── pipeline_flux_control_inpaint.py
├── pipeline_flux_control.py
├── pipeline_flux_controlnet_image_to_image.py
├── pipeline_flux_controlnet_inpainting.py
├── pipeline_flux_controlnet.py
├── pipeline_flux_fill.py
├── pipeline_flux_img2img.py
├── pipeline_flux_inpaint.py
├── pipeline_flux_kontext_inpaint.py
├── pipeline_flux_kontext.py
├── pipeline_flux_prior_redux.py
├── pipeline_flux_utils.py <------ holds the Mixin
├── pipeline_flux.py
└── pipeline_output.py
```

`pipeline_flux_utils.py` here has the Mixin class, holding the common methods. It additionally has the `calculate_shift()`  method, which is imported by the other pipelines under `src/diffusers/pipelines/flux`. Those pipelines won't have to maintain `# Copied from ...` versions of the `calculate_shift()` method anymore. They can just import like `from .pipeline_flux_utils import calculate_shift`. This is how we reduce LOCs, further on an intra-pipeline basis.

However, if a new pipeline (other than Flux), wants to leverage `calculate_shift()`, it should still follow `# Copied from ...` versions of the `calculate_shift()` instead of a direct import. This way, we can let individual pipeline-level methods evolve independently without introducing complex dependency patterns. For example, `pipeline_flux_utils.py` uses the `# Copied from ...` versions of methods like `retrieve_latents()`.

### LoRA

I experimented with moving `StableDiffusionLoraLoaderMixin` to `src/diffusers/pipelines/stable_diffusion/lora_utils.py`. However, there are a couple of derivative SD pipelines that subclass from `StableDiffusionLoraLoaderMixin` (imported from `lora_pipeline.py`). 

So, that will deprecation warnings for those. Two approaches come to mind:

* Do `from ..pipelines.stable_diffusion.lora_utils import StableDiffusionLoraLoaderMixin`. I don't like it at all because the dependency pattern is bad.
* Do a `"# Copied from ...` version of `StableDiffusionLoraLoaderMixin` inside a `lora_utils.py` the respective pipeline module. `src/diffusers/pipelines/animatediff`, for example. I prefer this one.

WDYT?

Also, if we decided to deprecate `StableDiffusionLoraLoaderMixin` from `diffusers.loaders.lora_pipeline`, it would lead to a circular import problem:
https://github.com/huggingface/diffusers/actions/runs/20021061502/job/57407882632?pr=12322#step:16:80

So, if we decided to go this route of separating the LoRA loaders, this might have to be a breaking change.

---

## Notes

* In the Qwen mixin, kept `encode_prompt()` varying while unifying the `_get_qwen_prompt_embeds()` method in `QwenImageMixin`. Comments are in line.
* There are a few methods that are shared by 2/3 pipelines such as `prepare_image()` is shared by `QwenImageControlNetInpaintPipeline` and `QwenImageControlNetPipeline` but not others. We can choose to move them to `pipeline_qwen_utils.py` but I am not strongly opinionated on this. The same applies to methods like `prepare_mask_latents()`. My preference here is to keep them as is for now.

We can take similar principles for Flux, too. Flux Control and ControlNet pipelines have a `prepare_image()` method. So, we could have something like `FluxControlMixin(FluxMixin)` and include `prepare_image()` there. But I felt like it was getting complex so, decided to skip.

---

## Additional questions

We have 
https://github.com/huggingface/diffusers/blob/6bf668c4d217ebc96065e673d8a257fd79950d34/src/diffusers/pipelines/pipeline_utils.py#L2175
This PR also has a similar Mixin class but without the methods from the above Mixin. To avoid confusion, I have named it `SDMixin`. 

WDYT about collating the two classes (`StableDiffusionMixin` and `SDMixin` (introduced in this PR)) into one?

### DIFF
--- a/src/diffusers/modular_pipelines/flux2/before_denoise.py
+++ b/src/diffusers/modular_pipelines/flux2/before_denoise.py
@@ -49,7 +49,7 @@ def compute_empirical_mu(image_seq_len: int, num_steps: int) -> float:
     return float(mu)
 
 
-# Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion.retrieve_timesteps
+# Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion_utils.retrieve_timesteps
 def retrieve_timesteps(
     scheduler,
     num_inference_steps: Optional[int] = None,
--- a/src/diffusers/modular_pipelines/z_image/before_denoise.py
+++ b/src/diffusers/modular_pipelines/z_image/before_denoise.py
@@ -112,7 +112,7 @@ def calculate_dimension_from_latents(latents: torch.Tensor, vae_scale_factor_spa
     return height, width
 
 
-# Copied from diffusers.pipelines.flux.pipeline_flux.calculate_shift
+# Copied from diffusers.pipelines.flux.pipeline_flux_utils.calculate_shift
 def calculate_shift(
     image_seq_len,
     base_seq_len: int = 256,
@@ -126,7 +126,7 @@ def calculate_shift(
     return mu
 
 
-# Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion.retrieve_timesteps
+# Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion_utils.retrieve_timesteps
 def retrieve_timesteps(
     scheduler,
     num_inference_steps: Optional[int] = None,


src/diffusers/pipelines/qwenimage/pipeline_qwen_utils.py
```diff
@@ -0,0 +1,310 @@
...
+
+        latents = latents.view(batch_size, height // 2, width // 2, channels // 4, 2, 2)
+        latents = latents.permute(0, 3, 1, 4, 2, 5)
+
+        latents = latents.reshape(batch_size, channels // (2 * 2), 1, height, width)
+
+        return latents
+
+
+class QwenImagePipelineMixin(QwenImageMixin):
```
DN6, line 297: You can also add additional Mixins here such as LoRA

sayakpaul:
> Looking good 👍🏽 Still on the fence on whether we use multiple levels of inheritance with QwenImageMixin. But other than that this looks good to me.

With you. I am also on the fence. Since the mixins for QwenImage and QwenImageEdit primarily differ in `encode_prompt()`, I decided to design it like that. We can still rewrite the `encode_prompt()` method in a single mixin so that it can deal with the images, but it could complicate things. It will also end up changing the signature of the method, leaving room for breaking changes.

But we can definitely uniformize the `_get_qwen_prompt_embeds()` method as it's private anyway. WDYT?

> Perhaps we try with to do the same with Flux and SDXL to see if some common patterns show up for these Mixins.

Yes, I will work on that.

src/diffusers/pipelines/qwenimage/pipeline_qwen_utils.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def _extract_masked_hidden(self, hidden_states: torch.Tensor, mask: torch.Tensor):
+        bool_mask = mask.bool()
+        valid_lengths = bool_mask.sum(dim=1)
+        selected = hidden_states[bool_mask]
+        split_result = torch.split(selected, valid_lengths.tolist(), dim=0)
+
+        return split_result
+
+    def _get_qwen_prompt_embeds(
```
sayakpaul, line 204: Decided to unify it to further reduce LOC. It's a private method anyway, so, we should be safe.

src/diffusers/pipelines/qwenimage/pipeline_qwen_utils.py
```diff
@@ -0,0 +1,293 @@
...
+        latents = latents.view(batch_size, height // 2, width // 2, channels // 4, 2, 2)
+        latents = latents.permute(0, 3, 1, 4, 2, 5)
+
+        latents = latents.reshape(batch_size, channels // (2 * 2), 1, height, width)
+
+        return latents
+
+
+class QwenImagePipelineMixin(QwenImageMixin):
+    def encode_prompt(
```
sayakpaul, line 298: Only differing `encode_prompt()` to not introduce any breaking change.

sayakpaul:
@DN6 pushed updates for Flux. LMK what you think.

sayakpaul:
In the interest of a better branch name: https://github.com/huggingface/diffusers/pull/12820

PR #12330, opened 2025-09-15
[tests] clean the pipeline level group offloading tests

sayakpaul:
# What does this PR do?

https://github.com/huggingface/diffusers/pull/12283 should not have introduced a separate test for testing pipeline-level group offloading inference, as we already have one:

https://github.com/huggingface/diffusers/blob/f5c113e4395bc373ab540fc5a1f7490b7120c40f/tests/pipelines/test_pipelines_common.py#L2248

This test modifies the existing `test_group_offloading_inference()` test to make use of the pipeline-level `enable_group_offload()`.

### DIFF
--- a/tests/pipelines/test_pipelines_common.py
+++ b/tests/pipelines/test_pipelines_common.py
@@ -32,7 +32,6 @@ from diffusers import (
     UNet2DConditionModel,
     apply_faster_cache,
 )
-from diffusers.hooks import apply_group_offloading
 from diffusers.hooks.faster_cache import FasterCacheBlockHook, FasterCacheDenoiserHook
 from diffusers.hooks.first_block_cache import FirstBlockCacheConfig
 from diffusers.hooks.pyramid_attention_broadcast import PyramidAttentionBroadcastHook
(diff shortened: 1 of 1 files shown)


tests/pipelines/test_pipelines_common.py
```diff
@@ -2395,40 +2320,62 @@ def test_pipeline_level_group_offloading_sanity_checks(self):
...
-        torch.manual_seed(0)
-        inputs = self.get_dummy_inputs(torch_device)
-        inputs["generator"] = torch.manual_seed(0)
-        out = pipe(**inputs)[0]
+        def enable_group_offload_on_component(pipe, group_offloading_kwargs):
+            # We intentionally don't test VAE's here. This is because some tests enable tiling on the VAE. If
+            # tiling is enabled and a forward pass is run, when accelerator streams are used, the execution order of
+            # the layers is not traced correctly. This causes errors. For apply group offloading to VAE, a
+            # warmup forward pass (even with dummy small inputs) is recommended.
+            exclude_modules = []
```
DN6: Can't we just set directly? Why nested loops? Or even set them via  `group_offloading_kwargs`?
```py
exclude_modules = ["vae", "vqvae", "image_encoder"]
```

tests/pipelines/test_pipelines_common.py
```diff
@@ -2395,40 +2320,62 @@ def test_pipeline_level_group_offloading_sanity_checks(self):
...
+            )
+            for component_name, component in pipe.components.items():
+                if component_name not in exclude_modules and isinstance(component, torch.nn.Module):
+                    self.assertTrue(
+                        all(
+                            module._diffusers_hook.get_hook("group_offloading") is not None
+                            for module in component.modules()
+                            if hasattr(module, "_diffusers_hook")
+                        )
+                    )
```
DN6, line 2354: IMO easier to read 

```py
            for component_name, component in pipe.components.items():
                if component_name in exclude_modules:
                      continue
                # test assertions      
```

tests/pipelines/test_pipelines_common.py
```diff
@@ -2395,40 +2320,62 @@ def test_pipeline_level_group_offloading_sanity_checks(self):
...
+        enable_group_offload_on_component(pipe, {"offload_type": "leaf_level"})
+        output_with_group_offloading2 = run_forward(pipe)
+
+        if torch.is_tensor(output_without_group_offloading):
+            output_without_group_offloading = output_without_group_offloading.detach().cpu().numpy()
+            output_with_group_offloading1 = output_with_group_offloading1.detach().cpu().numpy()
+            output_with_group_offloading2 = output_with_group_offloading2.detach().cpu().numpy()
+
+        self.assertTrue(np.allclose(output_without_group_offloading, output_with_group_offloading1, atol=1e-4))
+        self.assertTrue(np.allclose(output_without_group_offloading, output_with_group_offloading2, atol=1e-4))
```
DN6: Since the plan is to remove unittest as a dep, let's use standard asserts when adding new tests.

tests/pipelines/test_pipelines_common.py
```diff
@@ -2395,40 +2320,62 @@ def test_pipeline_level_group_offloading_sanity_checks(self):
...
 
-        max_diff = np.abs(to_np(out) - to_np(out_offload)).max()
-        self.assertLess(max_diff, expected_max_difference)
+        pipe = create_pipe()
+        enable_group_offload_on_component(pipe, {"offload_type": "block_level", "num_blocks_per_group": 1})
+        output_with_group_offloading1 = run_forward(pipe)
+
+        pipe = create_pipe()
+        enable_group_offload_on_component(pipe, {"offload_type": "leaf_level"})
+        output_with_group_offloading2 = run_forward(pipe)
```
DN6: I would actually turn these into separate tests, so it's easier to isolate issues.

tests/pipelines/test_pipelines_common.py
sayakpaul: Just readability convenience. 

Passing `exclude_modules = ["vae", "vqvae", "image_encoder"]` is a bit misleading for the test, IMO. We would still log an info if non-member components are passed, though:

https://github.com/huggingface/diffusers/blob/b50014067d7d31ef09d6dad62175bc19c2e258bf/src/diffusers/pipelines/pipeline_utils.py#L1443

However, I have simplified this in a commit:

```py
exclude_modules = {"vae", "vqvae", "image_encoder"}
exclude_modules = list(exclude_modules & set(pipe.components.keys()))
```

PR #12357, opened 2025-09-20
feat: Add QwenImageEditPlus to support future feature upgrades

naykun:


src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,936 @@
...
+
+
+logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
+
+EXAMPLE_DOC_STRING = """
+    Examples:
+        ```py
+        >>> import torch
+        >>> from PIL import Image
+        >>> from diffusers import QwenImageEditPipeline
```
sayakpaul:
```suggestion
        >>> from diffusers import QwenImageEditPlusPipeline
```

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,936 @@
...
+        >>> from diffusers.utils import load_image
+
+        >>> pipe = QwenImageEditPipeline.from_pretrained("Qwen/Qwen-Image-Edit", torch_dtype=torch.bfloat16)
```
sayakpaul: May have to also update the checkpoint? 👀 

```suggestion
        >>> pipe = QwenImageEditPlusPipeline.from_pretrained("Qwen/Qwen-Image-Edit", torch_dtype=torch.bfloat16)
```

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,936 @@
...
+        )
+        encoder_attention_mask = torch.stack(
+            [torch.cat([u, u.new_zeros(max_seq_len - u.size(0))]) for u in attn_mask_list]
+        )
+
+        prompt_embeds = prompt_embeds.to(dtype=dtype, device=device)
+
+        return prompt_embeds, encoder_attention_mask
+
+    def encode_prompt(
```
sayakpaul, line 287: Just double-checking if this is the same as the one we use in `QwenImageEditPipeline`? If so, let's use the `"# Copied from ..."` statement?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,936 @@
...
+    # Copied from diffusers.pipelines.qwenimage.pipeline_qwenimage.QwenImagePipeline._extract_masked_hidden
+    def _extract_masked_hidden(self, hidden_states: torch.Tensor, mask: torch.Tensor):
+        bool_mask = mask.bool()
+        valid_lengths = bool_mask.sum(dim=1)
+        selected = hidden_states[bool_mask]
+        split_result = torch.split(selected, valid_lengths.tolist(), dim=0)
+
+        return split_result
+
+    def _get_qwen_prompt_embeds(
```
sayakpaul, line 229: Is there a difference between this and the one used in `QwenImageEditPipeline`? If not, let's use the `"# Copied from ..."` statement?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,936 @@
...
+
+        _, seq_len, _ = prompt_embeds.shape
+        prompt_embeds = prompt_embeds.repeat(1, num_images_per_prompt, 1)
+        prompt_embeds = prompt_embeds.view(batch_size * num_images_per_prompt, seq_len, -1)
+        prompt_embeds_mask = prompt_embeds_mask.repeat(1, num_images_per_prompt, 1)
+        prompt_embeds_mask = prompt_embeds_mask.view(batch_size * num_images_per_prompt, seq_len)
+
+        return prompt_embeds, prompt_embeds_mask
+
+    def check_inputs(
```
sayakpaul, line 329: Can we use the one from `QwenImageEditPipeline`? If so, let's use the `"# Copied from ..."` statement?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,936 @@
...
+        width = 2 * (int(width) // (vae_scale_factor * 2))
+
+        latents = latents.view(batch_size, height // 2, width // 2, channels // 4, 2, 2)
+        latents = latents.permute(0, 3, 1, 4, 2, 5)
+
+        latents = latents.reshape(batch_size, channels // (2 * 2), 1, height, width)
+
+        return latents
+
+    def _encode_vae_image(self, image: torch.Tensor, generator: torch.Generator):
```
sayakpaul, line 411: Can we reuse the one from `QwenImageEditPipeline`?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,936 @@
...
+        origin_dtype = image.dtype
+        with torch.autocast(image.device.type, torch.float32):
```
sayakpaul: Hmm this is a bit of an anti-pattern. We avoid using `autocast` within pipeline code as much as possible. If the goal is to always ensure that the VAE encoding takes place in `torch.float32`, then we could instruct the user in doing so, like we do in 
https://huggingface.co/docs/diffusers/main/en/api/pipelines/wan

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,936 @@
...
+        """
+        depr_message = f"Calling `disable_vae_tiling()` on a `{self.__class__.__name__}` is deprecated and this method will be removed in a future version. Please use `pipe.vae.disable_tiling()`."
+        deprecate(
+            "disable_vae_tiling",
+            "0.40.0",
+            depr_message,
+        )
+        self.vae.disable_tiling()
+
+    def prepare_latents(
```
sayakpaul, line 434: Just double-checking if it's possible to use one from `QwenImageEditPipeline`? 👁️

sayakpaul (commented):
Thanks a lot for working on this! My comments are minor in nature except for the comment on using `autocast` from VAE encoding. 

We can actually also pass a dict to `torch_dtype` when initializing the pipeline like so:

https://github.com/huggingface/diffusers/blob/edd614ea38a7136cf66e1cb4dad696233c17358c/tests/pipelines/test_pipelines_common.py#L233-L2332

LMK if anything is unclear.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
naykun: I was wondering if we might consider making FP32 VAE encoding the default — just to help users who may not catch the note in the documentation.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
naykun: This function is a bit different~

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
naykun: This function operates slightly differently, as it supports encoding multiple condition images.

naykun:
Hi @sayakpaul,

Thank you for your review! I’ve updated the code to include the reuse statement as suggested — with the exception of two functions that behave differently by design.

Regarding the autocast issue: if we use torch_dtype_dict during checkpoint saving, would that allow users to encode images in FP32 transparently?

sayakpaul:
> Regarding the autocast issue: if we use torch_dtype_dict during checkpoint saving, would that allow users to encode images in FP32 transparently?

If we save using `save_pretrained()` with `torch_dtype` as a dict, it should work. However, to enforce it to the users in a transparent way, we could first check if the `vae.dtype` is in `float32` inside the `encode_vae_image()` function and then raise an error/warning educating the user about `torch_dtype` dict thing. WDYT?

yiyixuxu:
@naykun we can enforce upcast like in SDXL, but only when dtype isn't already torch.float32 
https://github.com/huggingface/diffusers/blob/edd614ea38a7136cf66e1cb4dad696233c17358c/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L1267

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,937 @@
...
+        Disable tiled VAE decoding. If `enable_vae_tiling` was previously enabled, this method will go back to
+        computing decoding in one step.
+        """
+        depr_message = f"Calling `disable_vae_tiling()` on a `{self.__class__.__name__}` is deprecated and this method will be removed in a future version. Please use `pipe.vae.disable_tiling()`."
+        deprecate(
+            "disable_vae_tiling",
+            "0.40.0",
+            depr_message,
+        )
+        self.vae.disable_tiling()
```
yiyixuxu:
```suggestion
```
we don't need to add these methods anymore, we are deprecating them

naykun:
@sayakpaul @yiyixuxu 
Thank you both for your thoughtful feedback. After re-evaluating the trade-offs — particularly the benefits of autocast versus potential impacts on efficiency and compatibility with the existing ecosystem — we’ve decided not to enforce this change at this time. By default, we’ll continue using bf16 as the encoding method.

All other concerns have been addressed. Please let me know if you have any further suggestions or changes you’d like us to consider.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
```diff
@@ -0,0 +1,884 @@
...
+            processor=processor,
+            transformer=transformer,
+            scheduler=scheduler,
+        )
+        self.vae_scale_factor = 2 ** len(self.vae.temperal_downsample) if getattr(self, "vae", None) else 8
+        self.latent_channels = self.vae.config.z_dim if getattr(self, "vae", None) else 16
+        # QwenImage latents are turned into 2x2 patches and packed. This means the latent width and height has to be divisible
+        # by the patch size. So the vae scale factor is multiplied by the patch size to account for this
+        self.image_processor = VaeImageProcessor(vae_scale_factor=self.vae_scale_factor * 2)
+        self.vl_processor = processor
```
sayakpaul: (nit): we're not using `vl_processor` anywhere in the code.

sayakpaul (approved):
All good from my end. I think we can also add an entry about this in the [docs](https://huggingface.co/docs/diffusers/main/en/api/pipelines/qwenimage)?

I can also help add a test in a separate PR if you want.

Off to @yiyixuxu for the final approval.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
naykun: nice catch 🤗

naykun:
> All good from my end. I think we can also add an entry about this in the [docs](https://huggingface.co/docs/diffusers/main/en/api/pipelines/qwenimage)?
> 
> I can also help add a test in a separate PR if you want.
> 
Thank you! That would be so great 😊

PR #12372, opened 2025-09-22
[docs] Remote inference

stevhliu:
- Renamed to "remote" inference since that is more descriptive of what it actually is while "hybrid" is a bit vague
- Consolidates encoding and decoding to a single doc with some minor edits and clarifications
- Moves the hybrid inference API reference to the API section
- Moves the guide itself to the Inference section


stevhliu:
@DN6, gentle reminder this is ready :)

PR #12374, opened 2025-09-23
[tests] introduce `VAETesterMixin` to consolidate tests for slicing and tiling

sayakpaul:
# What does this PR do?

More reductions in LoC.

### DIFF
--- a/tests/models/autoencoders/test_models_autoencoder_magvit.py
+++ b/tests/models/autoencoders/test_models_autoencoder_magvit.py
@@ -89,3 +89,9 @@ class AutoencoderKLMagvitTests(ModelTesterMixin, AutoencoderTesterMixin, unittes
     @unittest.skip("Unsupported test.")
     def test_forward_with_norm_groups(self):
         pass
+
+    @unittest.skip(
+        "Unsupported test. Error: RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 9 but got size 12 for tensor number 1 in the list."
+    )
+    def test_enable_disable_slicing(self):
+        pass
--- a/tests/models/autoencoders/testing_utils.py
+++ b/tests/models/autoencoders/testing_utils.py
@@ -4,6 +4,7 @@ import numpy as np
 import pytest
 import torch
 
+from diffusers.models.autoencoders.vae import DecoderOutput
 from diffusers.utils.torch_utils import torch_device
 
 
@@ -64,12 +65,17 @@ class AutoencoderTesterMixin:
         if accepts_generator:
             inputs_dict["generator"] = torch.manual_seed(0)
         output_without_tiling = model(**inputs_dict)[0]
+        # Mochi-1
+        if isinstance(output_without_tiling, DecoderOutput):
+            output_without_tiling = output_without_tiling.sample
 
         torch.manual_seed(0)
         model.enable_tiling()
         if accepts_generator:
             inputs_dict["generator"] = torch.manual_seed(0)
         output_with_tiling = model(**inputs_dict)[0]
+        if isinstance(output_with_tiling, DecoderOutput):
+            output_with_tiling = output_with_tiling.sample
 
         assert (
             output_without_tiling.detach().cpu().numpy() - output_with_tiling.detach().cpu().numpy()
@@ -80,6 +86,8 @@ class AutoencoderTesterMixin:
         if accepts_generator:
             inputs_dict["generator"] = torch.manual_seed(0)
         output_without_tiling_2 = model(**inputs_dict)[0]
+        if isinstance(output_without_tiling_2, DecoderOutput):
+            output_without_tiling_2 = output_without_tiling_2.sample
 
         assert np.allclose(
             output_without_tiling.detach().cpu().numpy().all(),
@@ -104,12 +112,17 @@ class AutoencoderTesterMixin:
 
         torch.manual_seed(0)
         output_without_slicing = model(**inputs_dict)[0]
+        # Mochi-1
+        if isinstance(output_without_slicing, DecoderOutput):
+            output_without_slicing = output_without_slicing.sample
 
         torch.manual_seed(0)
         model.enable_slicing()
         if accepts_generator:
             inputs_dict["generator"] = torch.manual_seed(0)
         output_with_slicing = model(**inputs_dict)[0]
+        if isinstance(output_with_slicing, DecoderOutput):
+            output_with_slicing = output_with_slicing.sample
 
         assert (
             output_without_slicing.detach().cpu().numpy() - output_with_slicing.detach().cpu().numpy()
@@ -120,6 +133,8 @@ class AutoencoderTesterMixin:
         if accepts_generator:
             inputs_dict["generator"] = torch.manual_seed(0)
         output_without_slicing_2 = model(**inputs_dict)[0]
+        if isinstance(output_without_slicing_2, DecoderOutput):
+            output_without_slicing_2 = output_without_slicing_2.sample
 
         assert np.allclose(
             output_without_slicing.detach().cpu().numpy().all(),


tests/models/autoencoders/test_models_autoencoder_hunyuan_video.py
```diff
@@ -25,13 +25,13 @@
...
 )
-from ..test_modeling_common import ModelTesterMixin, UNetTesterMixin
+from ..test_modeling_common import ModelTesterMixin, UNetTesterMixin, VAETestMixin
 
 
 enable_full_determinism()
 
 
-class AutoencoderKLHunyuanVideoTests(ModelTesterMixin, UNetTesterMixin, unittest.TestCase):
+class AutoencoderKLHunyuanVideoTests(ModelTesterMixin, VAETestMixin, UNetTesterMixin, unittest.TestCase):
```
DN6: Nit. Let's call it `AutoencoderTesterMixin` to be consistent. Also it looks like UNetTesterMixin is used in all the autoencoder tests? It only has a single test method. Let's add it to the  `AutoencoderTesterMixin` and just have these classes inherit from their specific Mixin

sayakpaul: > Also it looks like UNetTesterMixin is used in all the autoencoder tests? It only has a single test method. Let's add it to the AutoencoderTesterMixin and just have these classes inherit from their specific Mixin

`UNetTesterMixin` is also used in testing of some of the UNets.

sayakpaul: Done!

tests/models/test_modeling_common.py
```diff
@@ -1941,7 +1941,7 @@ def test_passing_dict_device_map_works(self, name, device):
             _ = loaded_model(**inputs_dict)
 
 
-class VAETestMixin:
+class AutoencoderTesterMixin:
```
DN6: Since we're probably going to add quite a few Mixins during refactors for tests, let's move this model specific ones into the the model test modules themselves. So we would add a `testing_utils.py` under `autoencoders` and this Mixin would go in there.

DN6:
I think we can remove the unittest dependency for these tests too?

sayakpaul:
Oh I think for that, we need to make `test_modeling_common` get rid of `unittest`, first.

tests/models/autoencoders/test_models_asymmetric_autoencoder_kl.py
```diff
@@ -35,13 +35,13 @@
...
 )
-from ..test_modeling_common import ModelTesterMixin, UNetTesterMixin
+from ..test_modeling_common import ModelTesterMixin
 
 
 enable_full_determinism()
 
 
-class AutoencoderKLTests(ModelTesterMixin, UNetTesterMixin, unittest.TestCase):
+class AsymmetricAutoencoderKLTests(ModelTesterMixin, unittest.TestCase):
```
dg845: The following autoencoder modules currently don't inherit from `AutoencoderTesterMixin`:

- `AsymmetricAutoencoderKL`
- `AutoencoderKLCosmos`
- `AutoencoderDC`
- `AutoencoderKLTemporalDecoder`
- `AutoencoderKLMagvit`
- `AutoencoderKLMochi`
- `AutoencoderOobleck`
- `VQModel`

Is this because they don't have `enable_tiling` or `enable_slicing` methods? Since `test_enable_disable_tiling` and `test_enable_disable_slicing` check for these methods, it seems like it should be safe for them to inherit from `AutoencoderTesterMixin`. It would also have the added benefit of allowing them to have the one test that was moved from `UNetTesterMixin` into `AutoencoderTesterMixin`.

sayakpaul: No one has used it :/

tests/models/autoencoders/test_models_autoencoder_cosmos.py
```diff
@@ -17,13 +17,13 @@
...
 from ...testing_utils import enable_full_determinism, floats_tensor, torch_device
-from ..test_modeling_common import ModelTesterMixin, UNetTesterMixin
+from ..test_modeling_common import ModelTesterMixin
 
 
 enable_full_determinism()
 
 
-class AutoencoderKLCosmosTests(ModelTesterMixin, UNetTesterMixin, unittest.TestCase):
+class AutoencoderKLCosmosTests(ModelTesterMixin, unittest.TestCase):
```
dg845: Similarly, should `AutoencoderKLCosmosTests` inherit from `AutoencoderTesterMixin`?

sayakpaul: We don't know if thet support slicing and tiling.

tests/models/autoencoders/test_models_autoencoder_dc.py
```diff
@@ -22,13 +22,13 @@
...
 )
-from ..test_modeling_common import ModelTesterMixin, UNetTesterMixin
+from ..test_modeling_common import ModelTesterMixin
 
 
 enable_full_determinism()
 
 
-class AutoencoderDCTests(ModelTesterMixin, UNetTesterMixin, unittest.TestCase):
+class AutoencoderDCTests(ModelTesterMixin, unittest.TestCase):
```
dg845: Similarly, should `AutoencoderDCTests` inherit from `AutoencoderTesterMixin`?

dg845: Closing as I have consolidated this comment into https://github.com/huggingface/diffusers/pull/12374#discussion_r2438358706. Feel free to reopen :).

tests/models/autoencoders/test_models_asymmetric_autoencoder_kl.py
sayakpaul: >  It would also have the added benefit of allowing them to have the one test that was moved from UNetTesterMixin into AutoencoderTesterMixin.

Do you mean `test_forward_with_norm_groups`?

dg845: Sorry, I edited my comment above to try to consolidate all of the models that don't inherit from `AutoencoderTesterMixin`. Do you think it would make sense for all autoencoder test classes to inherit from `AutoencoderTesterMixin`, as the slicing/tiling tests will be skipped if `enable_slicing`/`enable_tiling` methods are not available?

dg845: > Do you mean `test_forward_with_norm_groups`?

Yeah I think that's the one

sayakpaul:
@dg845 thanks! Nice suggestion: https://github.com/huggingface/diffusers/pull/12374#discussion_r2438358706. 

I applied it accordingly. It also helped us reduce a bit more LoC. PTAL.

dg845:
I saw that in the https://github.com/huggingface/diffusers/actions/runs/18580604097/job/52974552004?pr=12374 CI run the following tests failed:

```
FAILED tests/models/autoencoders/test_models_autoencoder_dc.py::AutoencoderDCTests::test_layerwise_casting_inference - RuntimeError: GET was unable to find an engine to execute this computation
FAILED tests/models/autoencoders/test_models_autoencoder_dc.py::AutoencoderDCTests::test_layerwise_casting_memory - RuntimeError: GET was unable to find an engine to execute this computation
```

I tried both of them locally on DGX and they passed (although my environment is not exactly synced up to the CI environment). Are these tests just flaky? It looks like the error might have to do with GPU availability.

EDIT: It looks like these same tests failed again on the next CI run with the same error: https://github.com/huggingface/diffusers/actions/runs/18583169787/job/52981958899?pr=12374.

sayakpaul:
Yeah those are unrelated to the PR, I think.

sayakpaul:
Fixed the Mochi related tests:
https://github.com/huggingface/diffusers/actions/runs/18583169808/job/52981961284?pr=12374#step:7:11793

For `AutoencoderKLMagvitTests`, failing, I have skipped them for now as they need deeper inspection (doesn't have to be prioritized as it's not used).

LMK what you think. 

@dg845

sayakpaul:
Also, regarding https://github.com/huggingface/diffusers/actions/runs/18583169787/job/52981958899?pr=12374, I think those are failing because of the CI hardware (uses T4 instances having compute capability of 7.5). Taking a look. The FP8 stuff there needs some hardware compatibility to work.

dg845 (approved):
LGTM! (assuming nothing weird comes up in the CI 😅)

PR #12387, opened 2025-09-25
[core] support flash attention through `kernels`

sayakpaul:
# What does this PR do?

Follow-up of https://github.com/huggingface/diffusers/pull/12236. 

Testing code:

```py
import torch

    model_id, torch_dtype=torch.bfloat16
).to("cuda")

pipe.transformer.set_attention_backend("flash_hub")

prompt = "A cat holding a sign that says 'hello world'"

    image = pipe(
        prompt, num_inference_steps=28, guidance_scale=4.0, generator=torch.manual_seed(0)
    ).images[0]
    image.save("output.png")
```

> [!TIP]
> Works with `torch.compile` fullgraph compatibility. 

I have tested the code on H100 and A100, and it works.


src/diffusers/models/attention_dispatch.py
```diff
@@ -170,6 +173,8 @@ class AttentionBackendName(str, Enum):
     # `flash-attn`
     FLASH = "flash"
     FLASH_VARLEN = "flash_varlen"
+    FLASH_HUB = "flash_hub"
```
sayakpaul, line 179: Flash Attention is stable. So, we don't have to mark it private like FA3.

src/diffusers/models/attention_dispatch.py
```diff
@@ -80,12 +80,15 @@
...
         )
-    from ..utils.kernels_utils import _get_fa3_from_hub
+    from ..utils.kernels_utils import _get_fa3_from_hub, _get_fa_from_hub
 
-    flash_attn_interface_hub = _get_fa3_from_hub()
-    flash_attn_3_func_hub = flash_attn_interface_hub.flash_attn_func
+    fa3_interface_hub = _get_fa3_from_hub()
+    flash_attn_3_func_hub = fa3_interface_hub.flash_attn_func
+    fa_interface_hub = _get_fa_from_hub()
+    flash_attn_func_hub = fa_interface_hub.flash_attn_func
```
MekkCyber, line 91: Why are we fetching both kernels here ?

src/diffusers/models/attention_dispatch.py
```diff
@@ -170,6 +173,8 @@ class AttentionBackendName(str, Enum):
     # `flash-attn`
     FLASH = "flash"
     FLASH_VARLEN = "flash_varlen"
+    FLASH_HUB = "flash_hub"
+    # FLASH_VARLEN_HUB = "flash_varlen_hub" # not supported yet.
```
MekkCyber, line 180: is this related to the kernel or it just needs more time to be integrated ?

sayakpaul: We don't have models that use varlen.

src/diffusers/models/attention_dispatch.py
sayakpaul: Because of the way APIs for attention backends are designed and also to support `torch.compile` with fullgraph traceability (when possible).  

We will let it grow a bit and upon feedback, we can revisit how to better deal with this.

src/diffusers/models/attention_dispatch.py
bghira: @sayakpaul qwen image uses varlen. also, native fused qkv+mlp attn requires varlen function.

sayakpaul:
@DN6 a gentle ping on this one.

src/diffusers/utils/kernels_utils.py
```diff
@@ -2,22 +2,32 @@
...
+
+    hub_id, kwargs = _DEFAULT_HUB_IDS[key]
+    try:
+        return get_kernel(hub_id, **kwargs)
+    except Exception as e:
+        logger.error(f"An error occurred while fetching kernel '{hub_id}' from the Hub: {e}")
+        raise
+
+
+def _get_fa3_from_hub():
```
DN6, line 28: This is a very thin wrapper. I would just call _get_from_hub("fa3") directly in attention_dispatch.

src/diffusers/utils/kernels_utils.py
```diff
@@ -2,22 +2,32 @@
...
 
-_DEFAULT_HUB_ID_FA3 = "kernels-community/flash-attn3"
+_DEFAULT_HUB_IDS = {
+    "fa3": ("kernels-community/flash-attn3", {"revision": "fake-ops-return-probs"}),
+    "fa": ("kernels-community/flash-attn", {}),
+}
 
 
-def _get_fa3_from_hub():
+def _get_from_hub(key: str):
```
DN6, line 16:
```suggestion
def _get_kernel_from_hub(key: str):
```

sayakpaul:
Closing for https://github.com/huggingface/diffusers/pull/12439.

PR #12394, opened 2025-09-26
[docs] slight edits to the attention backends docs.

sayakpaul:


docs/source/en/optimization/attention_backends.md
```diff
@@ -78,10 +78,16 @@ with attention_backend("_flash_3_hub"):
     image = pipeline(prompt).images[0]
 ```
 
+> [!TIP]
+> Most of these attention backends come with `torch.compile` compatibility without any graph breaks. Consider using it for maximum speedups.
```
stevhliu:
```suggestion
> Most attention backends support `torch.compile` without graph breaks and can be used to further speed up inference.
```

docs/source/en/optimization/attention_backends.md
```diff
@@ -78,10 +78,16 @@ with attention_backend("_flash_3_hub"):
...
+
 ## Available backends
 
 Refer to the table below for a complete list of available attention backends and their variants.
 
+<details>
```
stevhliu, line 88: May be simpler to replace the table at the beginning of the docs with this more complete one, wdyt?

stevhliu (approved):
Thanks!

docs/source/en/optimization/attention_backends.md
sayakpaul: Actually the first table is nice in the sense that it provides a consolidated overview of what's supported. So, I'd prefer to keep it.

PR #12407, opened 2025-09-29
[training-scripts] Make more examples UV-compatible (follow up on #12000)

linoytsaban:
make kontext and qwenimage also uv compatible, follow up on #12000


linoytsaban:
@sayakpaul I wonder if qwen might need newer versions for transformers and/or peft, we don't have a specific requirements.txt for it

sayakpaul:
> I wonder if qwen might need newer versions for transformers and/or peft, we don't have a specific requirements.txt for it

We can use the latest stable versions of both after running dummy runs? 👀

linoytsaban:
btw @sayakpaul just realized from testing we need to add `torchvision` as well (to scripts from the original PR as well).
Also, to properly support features like loading a dataset from the hub, training with 8-bit Adam or training with prodigy, we need to add `datasets`, `bitsandbytes`, `prodigyopt` - though this goes somewhat against are policy of keeping the dependencies to the required minimum (and differs from the deps in requirements.txt). WDYT?

sayakpaul (approved):

linoytsaban:
failing test unrelated

Issue #12409, opened 2025-09-30
Diffusers crash due to hard-dependency on `torch._C._distributed_c10d`


vladmandic:
### Describe the bug

`BriaTransformer2DModel` introduces a hard dependency on `torch._C._distributed_c10d` which is NOT present in all **torch** variants/flavors. 

Any `torch._C` import should be guarded with *if-available*, not a hard dependency on the entire library.
Without that, diffusers stop being a cross-platform solution.

### Reproduction

Attempt to import diffusers, it will immediately fail
Downstream issue: <https://github.com/vladmandic/sdnext/issues/4240>

### Logs

```shell
│E:\Work\LLM\sdnext-amd\venv\Lib\site-packages\diffusers\models\transformers\__init__.py:20 in <module>                │
│                                                                                                                      │
│  19     from .transformer_allegro import AllegroTransformer3DModel                                                   │
│❱ 20     from .transformer_bria import BriaTransformer2DModel                                                         │
│  21     from .transformer_chroma import ChromaTransformer2DModel                                                     │
│                                                                                                                      │
│E:\Work\LLM\sdnext-amd\venv\Lib\site-packages\diffusers\models\transformers\transformer_bria.py:14 in <module>        │
│                                                                                                                      │
│   13 from ..attention import AttentionModuleMixin, FeedForward                                                       │
│❱  14 from ..attention_dispatch import dispatch_attention_fn                                                          │
│   15 from ..cache_utils import CacheMixin                                                                            │
│                                                                                                                      │
│E:\Work\LLM\sdnext-amd\venv\Lib\site-packages\diffusers\models\attention_dispatch.py:23 in <module>                   │
│                                                                                                                      │
│    22 import torch                                                                                                   │
│❱   23 import torch.distributed._functional_collectives as funcol                                                     │
│    24                                                                                                                │
│                                                                                                                      │
│E:\Work\LLM\sdnext-amd\venv\Lib\site-packages\torch\distributed\_functional_collectives.py:9 in <module>              │
│                                                                                                                      │
│     8 import torch.distributed as dist                                                                               │
│❱    9 import torch.distributed.distributed_c10d as c10d                                                              │
│    10 from torch.distributed.device_mesh import DeviceMesh                                                           │
│                                                                                                                      │
│E:\Work\LLM\sdnext-amd\venv\Lib\site-packages\torch\distributed\distributed_c10d.py:23 in <module>                    │
│                                                                                                                      │
[...]

PR #12420, opened 2025-10-02
[core] conditionally import torch distributed stuff.

sayakpaul:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/issues/12409


vladmandic:
issue is only partially resolved, it just moved the problem to a slightly different code-path
see https://github.com/huggingface/diffusers/issues/12409#issuecomment-3362829709 for details.

PR #12424, opened 2025-10-03
fix dockerfile definitions.

sayakpaul:
# What does this PR do?

Docker build is broken from the past two days:
https://github.com/huggingface/diffusers/actions/runs/18208988147/job/51845669366#step:5:1173

Only changes the Dockerfiles that we use in `.github/workflows/build_docker_images.yml`.

### DIFF
--- a/src/diffusers/modular_pipelines/flux/before_denoise.py
+++ b/src/diffusers/modular_pipelines/flux/before_denoise.py
@@ -13,12 +13,12 @@
 # limitations under the License.
 
 import inspect
-from typing import Any, List, Optional, Tuple, Union
+from typing import List, Optional, Union
 
 import numpy as np
 import torch
 
-from ...models import AutoencoderKL
+from ...pipelines import FluxPipeline
 from ...schedulers import FlowMatchEulerDiscreteScheduler
 from ...utils import logging
 from ...utils.torch_utils import randn_tensor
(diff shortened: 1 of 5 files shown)


sayakpaul:
@DN6 could you give this a quick check if the changes are alright? I have also gone ahead and included the "test" related installations in the Dockerfiles themselves so that we don't have install them during CI (those changes will come in a separate PR).

docker/diffusers-pytorch-cpu/Dockerfile
```diff
@@ -1,50 +1,37 @@
...
-        transformers matplotlib  \
-        hf_transfer
+RUN pip install uv
+RUN uv pip install --no-cache-dir \
+    torch \
+    torchvision \
+    torchaudio \
+    --extra-index-url https://download.pytorch.org/whl/cpu
+
+RUN uv pip install --no-cache-dir "git+https://github.com/huggingface/diffusers.git@main#egg=diffusers[test]"
```
DN6, line 27: Is this needed? We already install test deps in the workflow file?

DN6 (approved):

docker/diffusers-pytorch-cpu/Dockerfile
sayakpaul: Yeah so in a future PR (immediately after this one), I will remove that step so that we can save some time skipping those installs.

DN6 (approved):

PR #12425, opened 2025-10-03
fix more torch.distributed imports

sayakpaul:
# What does this PR do?

https://github.com/huggingface/diffusers/pull/12420 partially fixed it.


src/diffusers/hooks/context_parallel.py
```diff
@@ -17,14 +17,17 @@
...
-    ContextParallelInput,
-    ContextParallelModelPlan,
-    ContextParallelOutput,
-)
+
+
+if torch.distributed.is_available():
+    import torch.distributed._functional_collectives as funcol
+
+    from ..models._modeling_parallel import (
```
DN6: Don't think you need to guard this. I think the checks inside `torch.distributed` should output stubs for the type annotations used in the ContextParallelConfig object.

Placing them under a guard would cause issues with slow imports because ContextParallelConfig is in the main diffusers init

sayakpaul: Resolved.

PR #12439, opened 2025-10-06
[core] support sage attention through `kernels`

sayakpaul:
# What does this PR do?

Code to test:

```py
import torch 

repo_id = "black-forest-labs/FLUX.1-dev"
pipe = DiffusionPipeline.from_pretrained(repo_id, torch_dtype=torch.bfloat16).to("cuda")
pipe.transformer.set_attention_backend("sage_hub")

image = pipe(
    prompt="a dog sitting by the sea, waiting for its companion to come",
    guidance_scale=3.5,
    num_inference_steps=30,
    max_sequence_length=512,
    generator=torch.manual_seed(0)
).images[0]
image.save("sage_flux.png")
```

Result: 
image

---

## Notes

1. It would be nice to get `torch.compile` support when using sage attention like we have for flash and flash 3.
2. We have other `sageattn` variants (see [here](https://github.com/huggingface/diffusers/blob/ce90f9b2db9f459f463a3239dc4c24a04072fd43/src/diffusers/models/attention_dispatch.py#L103C5-L108C27)), which would be cool to expose from the [Hub kernel](https://huggingface.co/kernels-community/sage_attention). 

Cc: @ MekkCyber

### DIFF
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -83,12 +83,16 @@ if DIFFUSERS_ENABLE_HUB_KERNELS:
         raise ImportError(
             "To use FA3 kernel for your hardware from the Hub, the `kernels` library must be installed. Install with `pip install kernels`."
         )
-    from ..utils.kernels_utils import _get_fa3_from_hub
+    from ..utils.kernels_utils import _DEFAULT_HUB_ID_FA3, _DEFAULT_HUB_ID_SAGE, _get_kernel_from_hub
 
-    flash_attn_interface_hub = _get_fa3_from_hub()
+    flash_attn_interface_hub = _get_kernel_from_hub(_DEFAULT_HUB_ID_FA3)
     flash_attn_3_func_hub = flash_attn_interface_hub.flash_attn_func
+
+    sage_interface_hub = _get_kernel_from_hub(_DEFAULT_HUB_ID_SAGE)
+    sage_attn_func_hub = sage_interface_hub.sageattn
 else:
     flash_attn_3_func_hub = None
+    sage_attn_func_hub = None
 
 if _CAN_USE_SAGE_ATTN:
     from sageattention import (
@@ -190,6 +194,7 @@ class AttentionBackendName(str, Enum):
 
     # `sageattention`
     SAGE = "sage"
+    SAGE_HUB = "sage_hub"
     SAGE_VARLEN = "sage_varlen"
     _SAGE_QK_INT8_PV_FP8_CUDA = "_sage_qk_int8_pv_fp8_cuda"
     _SAGE_QK_INT8_PV_FP8_CUDA_SM90 = "_sage_qk_int8_pv_fp8_cuda_sm90"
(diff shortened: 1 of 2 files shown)


sayakpaul:
> they are the same as sageattn, what i mean is sageattn is just a wrapper that dispatches to the correct kernel depending on the hardware used :

So, you mean we shouldn't have to have different dispatched functions like this? 
https://github.com/huggingface/diffusers/blob/ce90f9b2db9f459f463a3239dc4c24a04072fd43/src/diffusers/models/attention_dispatch.py#L194

MekkCyber:
Yes I think we don't need that because it depends on the hardware. For example if a user chooses : `_sage_qk_int8_pv_fp8_cuda` on A100 (`8.0`) it will fail, because this function is only supported and compiled for `8.9` gpus

src/diffusers/models/attention_dispatch.py
```diff
@@ -162,10 +171,6 @@ def wrap(func):
 # - CP with sage attention, flex, xformers, other missing backends
 # - Add support for normal and CP training with backends that don't support it yet
 
-_SAGE_ATTENTION_PV_ACCUM_DTYPE = Literal["fp32", "fp32+fp32"]
-_SAGE_ATTENTION_QK_QUANT_GRAN = Literal["per_thread", "per_warp"]
-_SAGE_ATTENTION_QUANTIZATION_BACKEND = Literal["cuda", "triton"]
```
sayakpaul, line 165: I don't see their usage, hence removed.

woct0rdho:
FYI, I've ported SageAttention to Python stable ABI (ABI3) and libtorch stable ABI, which should simplify building for HF Kernels:
https://github.com/woct0rdho/SageAttention/compare/main...abi3_stable

There are also some refactors in my main branch to simplify building. If someone can maintain the build system, then I no longer need to maintain my repo :)

sayakpaul:
This PR is ready to be reviewed now. As discussed with @MekkCyber over DMs, we're disabling `torch.compile` support now as the [`compile` branch](https://huggingface.co/kernels-community/sage_attention/tree/compile) leads to garbage outputs. 

In order for us to support it with `torch.compile`, some kind of lightweight dispatcher might be needed. https://github.com/huggingface/diffusers/pull/12439/commits/d3441340b9bb88d06a1eca5b60cd3ab7ef0f4683 added support for that but I have removed it for now for the above-mentioned purpos. Those changes are still safe in `sage-kernels-dispatch` branch.

I think we should be good with the PR.

Cc: @MekkCyber @DN6

src/diffusers/models/attention_dispatch.py
```diff
@@ -83,12 +83,17 @@
...
         )
-    from ..utils.kernels_utils import _get_fa3_from_hub
+    from ..utils.kernels_utils import _DEFAULT_HUB_ID_FA3, _DEFAULT_HUB_ID_SAGE, _get_kernel_from_hub
 
-    flash_attn_interface_hub = _get_fa3_from_hub()
+    flash_attn_interface_hub = _get_kernel_from_hub(_DEFAULT_HUB_ID_FA3)
     flash_attn_3_func_hub = flash_attn_interface_hub.flash_attn_func
+
+    sage_interface_hub = _get_kernel_from_hub(_DEFAULT_HUB_ID_SAGE)
+    sage_attn_func_hub = sage_interface_hub.sageattn
```
DN6: Not a huge fan of downloading all kernels if the env variable is set, since it's downloading stuff without explicit user consent. I think we need to rethink this part a bit.

sayakpaul: Exactly. This is why https://github.com/huggingface/diffusers/pull/12475/. Let's get that reviewed and merged first as it will unblock this PR and also #12387

sayakpaul:
@DN6 it should be up for another review. I have updated the test suite and ensured that they pass successfully as well. PTAL.

src/diffusers/models/attention_dispatch.py
```diff
@@ -1350,6 +1354,40 @@ def _flash_attention(
...
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
+    dropout_p: float = 0.0,
+    is_causal: bool = False,
+    scale: Optional[float] = None,
+    return_lse: bool = False,
+    _parallel_config: Optional["ParallelConfig"] = None,
+) -> torch.Tensor:
+    if _parallel_config:
```
DN6: Use:
```python
@_AttentionBackendRegistry.register(
    AttentionBackendName.FLASH_HUB,
    constraints=[_check_device, _check_qkv_dtype_bf16_or_fp16, _check_shape],
    supports_context_parallel=False,
)
```
https://github.com/huggingface/diffusers/blob/544ba677dd97a49c8124208837025aa8b5ab639e/src/diffusers/models/attention_dispatch.py#L1307

It will raise an error when trying to enable parallelism with this backend. This check isn't needed.

DN6 (approved):
Good to merge. But we need to remove the parallel config check and use `supports_context_parallel=False`

PR #12440, opened 2025-10-06
[Qwen LoRA training] fix bug when offloading

linoytsaban:


examples/dreambooth/train_dreambooth_lora_qwen_image.py
```diff
@@ -1338,7 +1338,7 @@ def compute_text_embeddings(prompt, text_encoding_pipeline):
                         batch["pixel_values"] = batch["pixel_values"].to(
                             accelerator.device, non_blocking=True, dtype=vae.dtype
                         )
-                    latents_cache.append(vae.encode(batch["pixel_values"]).latent_dist)
+                        latents_cache.append(vae.encode(batch["pixel_values"]).latent_dist)
```
linoytsaban, line 1341: so that `vae` & `batch["pixel_values"]` are both on the same device. w/o this change the `vae` is back on cpu but `batch["pixel_values"]` are on `accelerator.device`

sayakpaul (approved):
Okay to merge after "smol" is updated to "small". :v

PR #12443, opened 2025-10-06
Fix Flux Context Parallel Bug (Incoherent Image Generation)

mali-afridi:
# What does this PR do?
# Fix Context Parallelism: Implement Ring Attention Pattern for Coherent Multi-GPU Generation

## 🐛 Problem
I did some testings of the https://huggingface.co/docs/diffusers/main/training/distributed_inference on main branch. 
Context parallelism in diffusers was producing **fragmented/split images** when using multiple GPUs. Instead of generating a single coherent image, each GPU was independently generating its own portion, resulting in visible seams or completely different content in each image segment.

**Example**: Running with `torchrun --nproc-per-node=2` would produce an image that looked like two different images side-by-side rather than one unified image.

## 🔍 Root Cause Analysis

The issue stems from how attention was computed in context parallel mode:

### Before (Broken):
```python
# Each GPU only sees its local chunk
GPU 0: Q=[B, S/2, H, D], K=[B, S/2, H, D], V=[B, S/2, H, D]  # First half only
GPU 1: Q=[B, S/2, H, D], K=[B, S/2, H, D], V=[B, S/2, H, D]  # Second half only

# Result: Each GPU's attention can only see S/2 context, missing the other half
```

Each GPU was computing attention using only its **local sequence chunk** for Q, K, and V. This meant:
- GPU 0's queries could only attend to the first half of the sequence
- GPU 1's queries could only attend to the second half of the sequence
- No cross-GPU attention was happening → broken context → independent generations

## ✅ Solution: Ring Attention Pattern

This PR implements the **Ring Attention** pattern where:
- **Query (Q)** remains local to each GPU (for computational efficiency)
- **Key (K) and Value (V)** are gathered from all GPUs (for full context)

### After (Fixed):
```python
# Each GPU's Q is local, but K and V are global
GPU 0: Q=[B, S/2, H, D], K=[B, S_full, H, D], V=[B, S_full, H, D]
GPU 1: Q=[B, S/2, H, D], K=[B, S_full, H, D], V=[B, S_full, H, D]

# Result: Each GPU's queries can attend to the FULL sequence context!
```

## 📝 Implementation Details

The fix is applied directly in the attention processors after rotary embeddings but before attention computation:

**FluxAttnProcessor** (`transformer_flux.py`):
   ```python
   # After rotary embeddings
       # Gather K and V from all GPUs
       key_list = [torch.empty_like(key) for _ in range(world_size)]
       value_list = [torch.empty_like(value) for _ in range(world_size)]
       
[...]

PR #12446, opened 2025-10-07
Fix Context Parallel validation checks

DN6:
# What does this PR do?
Currently CP inference will run with split hooks even if the attention backend doesn't support it. This can lead to weird results #12443 

This PR

1. Fixes the validation check in the AttentionBackendRegistry to return False for invalid backends
2. Removes redundant checks in the ContextParallelConfig and adds properties to access device mesh config arguments in a more general way.   
3. Move CP config related validation checks into the CP Config itself, rather than in `enable_parallelism`   

Fixes # (issue)

## Before submitting

## Who can review?


src/diffusers/models/_modeling_parallel.py
```diff
@@ -79,29 +79,47 @@ def __post_init__(self):
...
         self._mesh = mesh
-        if self.ring_degree is None:
-            self.ring_degree = 1
-        if self.ulysses_degree is None:
-            self.ulysses_degree = 1
-        if self.rotate_method != "allgather":
-            raise NotImplementedError(
-                f"Only rotate_method='allgather' is supported for now, but got {self.rotate_method}."
+
+        if self.ulysses_degree * self.ring_degree > world_size:
```
sayakpaul, line 117: Should we hit line as both cannot be set, right?

src/diffusers/models/_modeling_parallel.py
```diff
@@ -79,29 +79,47 @@ def __post_init__(self):
...
-        if self._ring_local_rank is None:
-            self._ring_local_rank = self._ring_mesh.get_local_rank()
-        if self._ulysses_local_rank is None:
-            self._ulysses_local_rank = self._ulysses_mesh.get_local_rank()
+
+        self._flattened_mesh = self._mesh._flatten()
+        self._ring_mesh = self._mesh["ring"]
+        self._ulysses_mesh = self._mesh["ulysses"]
+        self._ring_local_rank = self._ring_mesh.get_local_rank()
+        self._ulysses_local_rank = self._ulysses_mesh.get_local_rank()
```
sayakpaul, line 126: Can't they be `None`? Why unguard?

src/diffusers/models/modeling_utils.py
```diff
@@ -1483,59 +1483,73 @@ def enable_parallelism(
...
-            if cp_config.ring_degree * cp_config.ulysses_degree > world_size:
-                raise ValueError(
-                    f"The product of `ring_degree` ({cp_config.ring_degree}) and `ulysses_degree` ({cp_config.ulysses_degree}) must not exceed the world size ({world_size})."
-                )
-            cp_mesh = torch.distributed.device_mesh.init_device_mesh(
-                device_type=device_type,
-                mesh_shape=(cp_config.ring_degree, cp_config.ulysses_degree),
-                mesh_dim_names=("ring", "ulysses"),
-            )
+            from .attention_dispatch import AttentionBackendName, _AttentionBackendRegistry
```
sayakpaul: (nit): Would prefer to keep the import at the beginning of the method implementation, if possible (after `from .attention_processor import Attention, MochiAttention`).

src/diffusers/models/_modeling_parallel.py
DN6: Both can be set techinically, but currently both can't be  > 1. Also this is for cases where you have 3 GPUs available and you set something like ulysses_degree=1 and ring_degree==4 (more GPUs being requested is greater than world_size)

src/diffusers/models/_modeling_parallel.py
DN6: They are internal attributes that are derived from mesh which is set through the `setup` method. The device mesh object is also only dynamically created when `enabled_parallelism` is called.  

The guards are redundant, they would always be `None` unless set explicitly for some custom debugging.

src/diffusers/models/_modeling_parallel.py
sayakpaul: Feels slightly confusing to me but since we're erroring out early for unsupported `ulysses_degree` and `ring_degree` value combos, I think it's okay.

src/diffusers/models/_modeling_parallel.py
```diff
@@ -79,29 +79,47 @@ def __post_init__(self):
...
+            )
+        if self.rotate_method != "allgather":
+            raise NotImplementedError(
+                f"Only rotate_method='allgather' is supported for now, but got {self.rotate_method}."
+            )
+
+    @property
+    def mesh_shape(self) -> Tuple[int, int]:
+        """Shape of the device mesh (ring_degree, ulysses_degree)."""
+        return (self.ring_degree, self.ulysses_degree)
```
sayakpaul, line 104: Would it be possible to add a small explainer about what it would mean for different values, for example - "(3, 1), (1, 3)", etc.? When both are being set, both cannot be > 1.

src/diffusers/models/attention_dispatch.py
```diff
@@ -251,15 +253,12 @@ def list_backends(cls):
...
+        backend: AttentionBackendName,
     ) -> bool:
-        supports_context_parallel = backend in cls._supports_context_parallel
-        is_degree_greater_than_1 = parallel_config is not None and (
-            parallel_config.context_parallel_config.ring_degree > 1
-            or parallel_config.context_parallel_config.ulysses_degree > 1
-        )
-        return supports_context_parallel and is_degree_greater_than_1
+        supports_context_parallel = backend.value in cls._supports_context_parallel
+        return supports_context_parallel
```
sayakpaul, line 261: Very nice cleanup here!

src/diffusers/models/modeling_utils.py
```diff
@@ -1484,59 +1484,72 @@ def enable_parallelism(
...
         rank = torch.distributed.get_rank()
         world_size = torch.distributed.get_world_size()
         device_type = torch._C._get_accelerator().type
         device_module = torch.get_device_module(device_type)
         device = torch.device(device_type, rank % device_module.device_count())
 
-        cp_mesh = None
+        attention_classes = (Attention, MochiAttention, AttentionModuleMixin)
+
+        # Step 1: Validate attention backend supports context parallelism if enabled
```
sayakpaul: (nit): Seems like we are not documenting other steps. So, maybe we can remove this comment or add other steps as comments.

sayakpaul (approved):
Just some nits. But not merge-blocker.

DefTruth:
@sayakpaul It worked for me. When I set the attn backend to native, it threw an error correctly. but i think the `native` attn backend should be enable for context parallelism as #12563 

```bash
INFO 10-31 07:05:23 [__init__.py:64] Found attention_backend from config, set attention backend to: native
[rank1]:     transformer.enable_parallelism(
[rank1]:   File "/workspace/dev/vipshop/diffusers/src/diffusers/models/modeling_utils.py", line 1529, in enable_parallelism
[rank1]:     raise ValueError(
[rank1]: ValueError: Context parallelism is enabled but the attention processor 'WanAttnProcessor' is using backend 'native' which does not support context parallelism. Please set a compatible attention backend: ['_native_cudnn', 'flash', 'sage'] using `model.set_attention_backend()` before calling `enable_parallelism()`.
```

sayakpaul:
@DN6 I think we might have to consider https://github.com/huggingface/diffusers/pull/12563 now that native is supported for Ulysses.

PR #12448, opened 2025-10-07
Fix checkpoint overwrite bug in train_sana_sprint_diffusers

### LINKED ISSUE #12444: [train_sana_sprint] Custom save_hook overwrites trained transformer with frozen model during checkpointing
### Describe the bug

In the train_sana_sprint_diffusers.py example script, the custom save_model_hook incorrectly saves the wrong model. The isinstance() check is too broad because both the trained transformer and the frozen pretrained_model are instances of the same class. This causes the hook to save the trained model and then immediately overwrite it with the frozen, untrained reference model in the same checkpoint directory.

### Reproduction

```
# In train_sana_sprint_diffusers.py, this save hook causes the issue:
def save_model_hook(models, weights, output_dir):
    if accelerator.is_main_process:
        for model in models:
            unwrapped_model = unwrap_model(model)
            # This check is too broad and matches both the trained and frozen models
            if isinstance(unwrapped_model, type(unwrap_model(transformer))):
                model = unwrapped_model
                model.save_pretrained(os.path.join(output_dir, "transformer"))
            # ... rest of the function ...
```

### Logs

```shell

```

### System Info

- **OS**: Windows 11
- **Python**: 3.10.18
- **`diffusers` version**: 0.35.1
- **`transformers` version**: 4.57.0
- **`torch` version**: 2.8.0
- **`accelerate` version**: 1.10.1
- **`huggingface-hub` version**: 0.35.3
- **`safetensors` version**: 0.6.2

### Who can help?

# Suggested Fix
```
def save_model_hook(models, weights, output_dir):
    if accelerator.is_main_process:
        for model in models:
            unwrapped_model = unwrap_model(model)

            # Handle transformer model
            if isinstance(unwrapped_model, type(unwrap_model(transformer))) and unwrapped_model.guidance_embeds:
                model = unwrapped_model
                model.save_pretrained(os.path.join(output_dir, "transformer"))

            # Handle discriminator model
            elif isinstance(unwrapped_model, type(unwrap_model(disc))):
                torch.save(unwrapped_model.heads.state_dict(), os.path.join(output_dir, "disc_heads.pt"))

            if weights:
                weights.pop()
```

Aishwarya0811:
## What does this PR do?

Fixes #12444 

This PR fixes a critical bug in the `train_sana_sprint_diffusers.py` script where the `save_model_hook` incorrectly saves both the trained transformer and the frozen `pretrained_model` to the same checkpoint location, causing the frozen model to overwrite the trained weights.

## The Problem

The original `isinstance()` check was too broad:
python
if isinstance(unwrapped_model, type(unwrap_model(transformer))):
    model.save_pretrained(os.path.join(output_dir, "transformer"))
    Both transformer (trained, with guidance=True) and pretrained_model (frozen reference, with guidance=False) are instances of SanaTrigFlow, so this condition matched both models. The result:

First iteration: Saves trained model ✓
Second iteration: Overwrites with frozen model ✗

Users lose all their training progress during checkpointing.

## The Solution
Added a discriminator to distinguish between the models:
pythonif isinstance(unwrapped_model, type(unwrap_model(transformer))) and getattr(unwrapped_model, 'guidance', False):
    model.save_pretrained(os.path.join(output_dir, "transformer"))
The trained model has guidance=True, while the frozen reference has guidance=False. Now only the trained model is saved.
Also added an explicit skip clause for the frozen model to prevent it from triggering the error handler:
pythonelif isinstance(unwrapped_model, type(unwrap_model(transformer))):
    pass  # Skip frozen pretrained_model
Testing
Created a test script (test.py) that verifies:

✓ Trained transformer is saved correctly
✓ Frozen pretrained_model is skipped
✓ Discriminator heads are saved correctly

Test results:
ALL CHECKS PASSED! The fix is working correctly!
  - Trained transformer: SAVED
  - Frozen model: SKIPPED
  - Discriminator heads: SAVED
  - 
co author : @malakazlan


yiyixuxu:
cc @linoytsaban

linoytsaban:
lgtm, cc @sayakpaul as well

sayakpaul:
@bot /style

sayakpaul:
Cc: @lawrence-cj for awareness.

examples/research_projects/sana/train_sana_sprint_diffusers.py
```diff
@@ -1161,7 +1166,9 @@ def load_model_hook(models, input_dir):
                     model = models.pop()
                     unwrapped_model = unwrap_model(model)
 
-                    if isinstance(unwrapped_model, type(unwrap_model(transformer))):
+                    if isinstance(unwrapped_model, type(unwrap_model(transformer))) and getattr(
+                        unwrapped_model, "guidance", False
+                    ):
```
sayakpaul, line 1171: Can you explain this change?. @Aishwarya0811

examples/research_projects/sana/train_sana_sprint_diffusers.py
```diff
@@ -1145,6 +1145,11 @@ def save_model_hook(models, weights, output_dir):
                     elif isinstance(unwrapped_model, type(unwrap_model(disc))):
                         # Save only the heads
                         torch.save(unwrapped_model.heads.state_dict(), os.path.join(output_dir, "disc_heads.pt"))
+
+                        # Skip frozen pretrained_model
+                    elif isinstance(unwrapped_model, type(unwrap_model(transformer))):
+                        pass
```
sayakpaul, line 1151: This change as well.

examples/research_projects/sana/train_sana_sprint_diffusers.py
linoytsaban: > Added a discriminator to distinguish between the models:
> pythonif isinstance(unwrapped_model, type(unwrap_model(transformer))) and getattr(unwrapped_model, 'guidance', False):
> model.save_pretrained(os.path.join(output_dir, "transformer"))
> The trained model has guidance=True, while the frozen reference has guidance=False. Now only the trained model is saved.

examples/research_projects/sana/train_sana_sprint_diffusers.py
linoytsaban: > Also added an explicit skip clause for the frozen model to prevent it from triggering the error handler:
> pythonelif isinstance(unwrapped_model, type(unwrap_model(transformer))):
> pass # Skip frozen pretrained_model

examples/research_projects/sana/train_sana_sprint_diffusers.py
linoytsaban: taken from the solution summary in the description ^

examples/research_projects/sana/train_sana_sprint_diffusers.py
linoytsaban: taken from the solution summary in the description ^

PR #12450, opened 2025-10-08
[ci] revisit the installations in CI.

sayakpaul:
# What does this PR do?

Follow-ups from https://github.com/huggingface/diffusers/pull/12424

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -38,9 +38,8 @@ jobs:
         run: |
           apt update
           apt install -y libpq-dev postgresql-client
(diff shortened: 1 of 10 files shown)


docker/diffusers-doc-builder/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN apt-get -y update && apt-get install -y bash \
     git-lfs \
```
sayakpaul, line 11: Changes in the Dockerfiles have been built and pushed as Docker containers https://github.com/huggingface/diffusers/actions/runs/18335386344.

DN6 (approved):

PR #12453, opened 2025-10-09
Fix `QwenImageEditPlusPipeline` not using the passed in width/height for the VAE

cakedan:
# What does this PR do?

Fixes `QwenImageEditPlusPipeline` not using the passed in width/height for the VAE.

Currently if you set the width/height to anything other than 1024x1024, for example 512x512, the image will be resized to 1024x1024 before being passed in the VAE then cropped to fit the passed in widthxheight which results in a zoomed in photo.

### DIFF
--- a/docs/source/en/api/pipelines/qwenimage.md
+++ b/docs/source/en/api/pipelines/qwenimage.md
@@ -108,11 +108,45 @@ pipe = QwenImageEditPlusPipeline.from_pretrained(
 image_1 = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/grumpy.jpg")
 image_2 = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/peng.png")
 image = pipe(
-    image=[image_1, image_2], 
-    prompt='''put the penguin and the cat at a game show called "Qwen Edit Plus Games"''', 
+    image=[image_1, image_2],
+    prompt='''put the penguin and the cat at a game show called "Qwen Edit Plus Games"''',
     num_inference_steps=50
 ).images[0]
 ```
+
+## Performance
+
+### torch.compile
+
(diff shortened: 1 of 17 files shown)


JoeGaffney:
I noticed this issue with 16by 9 ratio. Glad you found what it was

JoeGaffney:
Hey,

I got some pretty good results doing.


``` python
def image_edit_call(context: ImageContext):
    # see https://github.com/huggingface/diffusers/pull/12453/files
    import diffusers.pipelines.qwenimage.pipeline_qwenimage_edit_plus as qwen_edit_module

    qwen_edit_module.VAE_IMAGE_SIZE = context.width * context.height

    # gather all possible reference images
    reference_images = []
    if context.color_image:
        reference_images.append(context.color_image)

    for current in context.get_reference_images():
        if current is not None:
            reference_images.append(current)

    prompt_embeds, prompt_embeds_mask = qwen_edit_encode(context.data.prompt, reference_images)
    pipe = get_edit_pipeline("ovedrive/Qwen-Image-Edit-2509-4bit")

    args = {
        "width": context.width,
        "height": context.height,
        "prompt_embeds": prompt_embeds,
        "prompt_embeds_mask": prompt_embeds_mask,
        "negative_prompt": "",
        "image": reference_images,
        "generator": context.generator,
        "num_inference_steps": 8,
        "true_cfg_scale": 1.0,
    }

    processed_image = pipe.__call__(**args).images[0]
    context.cleanup()

    return processed_image
```

Before 
qwen-image_image_to_image - Copy

After
qwen-image_image_to_image

vgabbo:
Is this that causes the crop on an image when it's smaller than 640x640? I noticed that if I try to modify a small image it always crops it :/ Will it be merged?
Thanks for your work.

vgabbo:
I just tried and INDEED it helped a lot, now if I send low res images it does not crop them anymore, thus making it much faster, I'll only have to upscale later, but still now that's 1/3 the time it took me to edit the 1MP images!! Thank you VERY much for your contribution!! @JoeGaffney @cakedan

PR #12455, opened 2025-10-09
[ci] xfail more incorrect transformer imports.

sayakpaul:
# What does this PR do?

Hopefully, the last ones.

### DIFF
--- a/tests/pipelines/test_pipelines.py
+++ b/tests/pipelines/test_pipelines.py
@@ -629,6 +629,7 @@ class DownloadTests(unittest.TestCase):
             # https://huggingface.co/hf-internal-testing/stable-diffusion-broken-variants/tree/main/unet
             assert len(files) == 15, f"We should only download 15 files, not {len(files)}"
 
+    @pytest.mark.xfail(condition=is_transformers_version(">", "4.56.2"), reason="Some import error", strict=False)
     def test_download_bin_only_variant_exists_for_model(self):
         variant = None
         use_safetensors = False


tests/pipelines/test_pipelines.py
```diff
@@ -600,6 +600,7 @@ def test_download_legacy_variants_with_sharded_ckpts_raises_warning(self):
                 )
         assert deprecated_warning_msg in str(cap_logger), "Deprecation warning not found in logs"
 
+    @pytest.mark.xfail(condition=is_transformers_version(">", "4.56.2"), reason="Some import error", strict=True)
     def test_download_safetensors_only_variant_exists_for_model(self):
```
dg845, line 603: Perhaps I am missing something, but it appears that this test (and `test_download_safetensors_variant_does_not_exist_for_model` below) are still marked as failed rather than xfailed in the CI:
```
FAILED tests/pipelines/test_pipelines.py::DownloadTests::test_download_safetensors_only_variant_exists_for_model
FAILED tests/pipelines/test_pipelines.py::DownloadTests::test_download_safetensors_variant_does_not_exist_for_model
```
Is this expected?

sayakpaul: Yeah this is not expected and seeing for the first time. Let me check.

sayakpaul:
Hmm, I am not sure why the Transformers version is showing up to 4.57.0-dev here: https://github.com/huggingface/diffusers/actions/runs/18487464650/job/52673652920?pr=12455#step:5:19 whereas it should have been 5.0.0-dev. My plate is a little full at the moment to dive deeper into it. 

So, I have made the xfail non-strict, which should hopefully unblock us. But I am open to other suggestions.

To double-check, I tried to replicate the exact same CI environment:

1. `docker run -it diffusers/diffusers-pytorch-cpu`
2. Clone diffusers and `cd` into it.
3. Check out this PR branch.
4. Then run these: 
https://github.com/huggingface/diffusers/blob/fa468c5d576a42f50ef467a4dc24d0326b25f656/.github/workflows/pr_tests.yml#L117-L120
5. Then run `pytest <corresponding_test>`.

PR #12463, opened 2025-10-10
[core] reuse `AttentionMixin` for compatible classes

sayakpaul:
# What does this PR do?

Many models use `"# Copied from ..."` implementations of `attn_processors` and `set_attn_processor`. They are basically the same as what we have implemented in 
https://github.com/huggingface/diffusers/blob/693d8a3a52252153dc0f1503ea87db89d2364693/src/diffusers/models/attention.py#L39

This PR makes those models inherit from `AttentionMixin` and removes the copied-over implementations.

I decided to leave `fuse_qkv_projections` and `unfuse_qkv_projections` out of this PR because some models don't have attention processors implemented in a way that would make this seamless. But the methods removed in this PR should be very harmless.

### DIFF
--- a/docs/source/en/api/pipelines/hunyuan_video15.md
+++ b/docs/source/en/api/pipelines/hunyuan_video15.md
@@ -56,8 +56,8 @@ export_to_video(video, "output.mp4", fps=15)
 
 - HunyuanVideo1.5 use attention masks with variable-length sequences. For best performance, we recommend using an attention backend that handles padding efficiently.
 
-    - **H100/H800:** `_flash_3_hub` or `_flash_varlen_3`
-    - **A100/A800/RTX 4090:** `flash_hub` or `flash_varlen`
+    - **H100/H800:** `_flash_3_hub` or `_flash_3_varlen_hub`
+    - **A100/A800/RTX 4090:** `flash_hub` or `flash_varlen_hub`
     - **Other GPUs:** `sage_hub`
 
 Refer to the [Attention backends](../../optimization/attention_backends) guide for more details about using a different backend.
--- a/docs/source/en/optimization/attention_backends.md
+++ b/docs/source/en/optimization/attention_backends.md
@@ -141,10 +141,12 @@ Refer to the table below for a complete list of available attention backends and
 | `flash` | [FlashAttention](https://github.com/Dao-AILab/flash-attention) | FlashAttention-2 |
 | `flash_hub` | [FlashAttention](https://github.com/Dao-AILab/flash-attention) | FlashAttention-2 from kernels |
 | `flash_varlen` | [FlashAttention](https://github.com/Dao-AILab/flash-attention) | Variable length FlashAttention |
+| `flash_varlen_hub` | [FlashAttention](https://github.com/Dao-AILab/flash-attention) | Variable length FlashAttention from kernels |
 | `aiter` | [AI Tensor Engine for ROCm](https://github.com/ROCm/aiter) | FlashAttention for AMD ROCm |
 | `_flash_3` | [FlashAttention](https://github.com/Dao-AILab/flash-attention) | FlashAttention-3 |
 | `_flash_varlen_3` | [FlashAttention](https://github.com/Dao-AILab/flash-attention) | Variable length FlashAttention-3 |
 | `_flash_3_hub` | [FlashAttention](https://github.com/Dao-AILab/flash-attention) | FlashAttention-3 from kernels |
+| `_flash_3_varlen_hub` | [FlashAttention](https://github.com/Dao-AILab/flash-attention) | Variable length FlashAttention-3 from kernels |
 | `sage` | [SageAttention](https://github.com/thu-ml/SageAttention) | Quantized attention (INT8 QK) |
 | `sage_hub` | [SageAttention](https://github.com/thu-ml/SageAttention) | Quantized attention (INT8 QK) from kernels |
 | `sage_varlen` | [SageAttention](https://github.com/thu-ml/SageAttention) | Variable length SageAttention |
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -168,10 +168,11 @@ class AttentionBackendName(str, Enum):
     FLASH = "flash"
     FLASH_HUB = "flash_hub"
     FLASH_VARLEN = "flash_varlen"
+    FLASH_VARLEN_HUB = "flash_varlen_hub"
     _FLASH_3 = "_flash_3"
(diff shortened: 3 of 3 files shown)


src/diffusers/pipelines/audioldm2/modeling_audioldm2.py
```diff
@@ -25,7 +25,6 @@
 from ...models.attention_processor import (
     ADDED_KV_ATTENTION_PROCESSORS,
     CROSS_ATTENTION_PROCESSORS,
-    AttentionProcessor,
```
dg845, line 27: I see that many of the tests in `AudioLDM2PipelineFastTests` currently fail in the CI with the following error, e.g.:
```
FAILED tests/pipelines/audioldm2/test_audioldm2.py::AudioLDM2PipelineFastTests::test_inference_batch_consistent - AttributeError: type object 'ClapConfig' has no attribute 'from_text_audio_configs'
```
This method is called in the `AudioLDM2PipelineFastTests.get_dummy_components`:

https://github.com/huggingface/diffusers/blob/fa468c5d576a42f50ef467a4dc24d0326b25f656/tests/pipelines/audioldm2/test_audioldm2.py#L141-L145

It looks like the `ClapConfig.from_text_audio_configs` method exists in `transformers==4.57.0` but has been removed in `main`. Given that this method will be deprecated, should we replace this call with something like
```python
class AudioLDM2PipelineFastTests(PipelineTesterMixin, unittest.TestCase):
    ...
    def get_dummy_components(self):
        ...
        text_encoder_config = ClapConfig(
            text_config=text_branch_config,
            audio_config=audio_branch_config,
            projection_dim=16,
        )
        ...
    ...
```
?

dg845: Similarly, for the following tests fail due to `CLIPFeatureExtractor` being removed:
```
FAILED tests/pipelines/test_pipelines.py::DownloadTests::test_download_bin_only_variant_exists_for_model - AttributeError: module transformers has no attribute CLIPFeatureExtractor
FAILED tests/pipelines/test_pipelines.py::DownloadTests::test_download_bin_variant_does_not_exist_for_model - AttributeError: module transformers has no attribute CLIPFeatureExtractor
FAILED tests/pipelines/test_pipelines.py::PipelineFastTests::test_wrong_model - AttributeError: module transformers has no attribute CLIPFeatureExtractor
```
Should we replace the calls to `CLIPFeatureExtractor` with `CLIPImageProcessor`, or do you think that should be separated into a new PR?

src/diffusers/models/transformers/auraflow_transformer_2d.py
```diff
@@ -365,66 +364,6 @@ def __init__(
 
         self.gradient_checkpointing = False
 
-    @property
-    # Copied from diffusers.models.unets.unet_2d_condition.UNet2DConditionModel.attn_processors
-    def attn_processors(self) -> Dict[str, AttentionProcessor]:
```
dg845, line 370: I think `AuraFlowTransformer2DModel` should inherit from `AttentionMixin`, since the `attn_processors` property has been removed here but not added back via the mixin, leading to a CI failure when we try to call it:
```
FAILED tests/pipelines/aura_flow/test_pipeline_aura_flow.py::AuraFlowPipelineFastTests::test_fused_qkv_projections - AttributeError: 'AuraFlowTransformer2DModel' object has no attribute 'attn_processors'
```

src/diffusers/pipelines/audioldm2/modeling_audioldm2.py
```diff
@@ -531,66 +530,6 @@ def __init__(
             block_out_channels[0], out_channels, kernel_size=conv_out_kernel, padding=conv_out_padding
         )
 
-    @property
```
dg845, line 538: Should `AudioLDM2UNet2DConditionModel` inherit from `AttentionMixin`? I see that these methods have been removed but no corresponding change to add the mixin.

src/diffusers/models/autoencoders/autoencoder_kl.py
```diff
@@ -167,66 +167,6 @@ def disable_slicing(self):
...
-                else:
-                    module.set_processor(processor.pop(f"{name}.processor"))
-
-            for sub_name, child in module.named_children():
-                fn_recursive_attn_processor(f"{name}.{sub_name}", child, processor)
-
-        for name, module in self.named_children():
-            fn_recursive_attn_processor(name, module, processor)
-
     # Copied from diffusers.models.unets.unet_2d_condition.UNet2DConditionModel.set_default_attn_processor
```
dg845, line 143: Perhaps it's out of the scope for this PR, but I see that a lot of models additionally have a `set_default_attn_processor` method, usually `# Copied from diffusers.models.unets.unet_2d_condition.UNet2DConditionModel.set_default_attn_processor`. Do you think it makes sense to add this method to `AttentionMixin`?

src/diffusers/pipelines/audioldm2/modeling_audioldm2.py
sayakpaul: For https://github.com/huggingface/diffusers/pull/12463#discussion_r2427654483 see https://github.com/huggingface/diffusers/pull/12455


nit: Unrelated to this PR. Prefer discussing these separately.

src/diffusers/models/transformers/auraflow_transformer_2d.py
sayakpaul: Good catch. Added in 95fcb5359.

src/diffusers/models/autoencoders/autoencoder_kl.py
sayakpaul: IMO, not yet since `AttentionMixin` is fairly agnostic to the model-type but `set_default_attn_processor` relies on some custom attention processor types. For UNet2DConditionModel, we have:

https://github.com/huggingface/diffusers/blob/fa468c5d576a42f50ef467a4dc24d0326b25f656/src/diffusers/models/unets/unet_2d_condition.py#L762-L769

However, for AutoencoderKL Temporal Decoder:
https://github.com/huggingface/diffusers/blob/fa468c5d576a42f50ef467a4dc24d0326b25f656/src/diffusers/models/autoencoders/autoencoder_kl_temporal_decoder.py#L269-L274

I'd be down to the refactoring, though. Cc: @DN6

sayakpaul:
Thanks for those catches, @dg845. Should have been fixed by now.

dg845:
LGTM :)

sayakpaul:
@DN6 okay to go?

sayakpaul:
@DN6 a gentle ping.

DN6 (approved):

PR #12473, opened 2025-10-13
[core] `AutoencoderMixin` to abstract common methods

sayakpaul:
# What does this PR do?

Asbtracts common methods like `enable_slicing()`, `disable_slicing()`, `enable_tiling()`, and `disable_tiling()` to a class `AutoencoderMixin`. Not all VAEs implement slicing and tiling and that has been addressed accordingly.

As a consequence, we also reduce a bit of code bloat.


dg845 (approved):
LGTM! Do you think the remaining autoencoders, which I believe are

- `AsymmetricAutoencoderKL`
- `AutoencoderKLTemporalDecoder`
- `VQModel`

should also inherit from `AutoencoderMixin`? It would make the autoencoder interface more consistent and I think it should be safe in the sense that e.g. `AutoencoderMixin.enable_tiling` should raise a `NotImplementedError` for these models and the associated tiling/slicing tests in `AutoencoderTesterMixin` should be properly skipped since these models don't define `use_tiling`/`use_slicing` attributes. (An exception is `AsymmetricAutoencoderKL`, which defines `use_tiling`/`use_slicing` but not `enable_tiling`/`enable_slicing` methods currently.)

dg845 (approved):
LGTM :). I would suggest removing the `use_slicing`/`use_tiling` attributes on `AsymmetricAutoencoderKL`:

https://github.com/huggingface/diffusers/blob/cc65d0a09c8b68f53f04c2f4df109d1b8e0db836/src/diffusers/models/autoencoders/autoencoder_asym_kl.py#L110-L111

Since tiling/slicing isn't implemented for this model, I think it makes sense to skip e.g. `AsymmetricAutoencoderKLTests.test_enable_disable_tiling` (right now it passes since `AsymmetricAutoencoderKL.enable_tiling` is essentially a no-op).

sayakpaul:
> Since tiling/slicing isn't implemented for this model, I think it makes sense to skip e.g. AsymmetricAutoencoderKLTests.test_enable_disable_tiling (right now it passes since AsymmetricAutoencoderKL.enable_tiling is essentially a no-op).

I have removed them. But just a nit, I don't think the tests would actually, they _should_ get skipped rather.

sayakpaul:
Failing tests are unrelated.

PR #12474, opened 2025-10-13
fix CI bug for kandinsky3_img2img case

kaixuanliu:
For full pipeline, related `cut_context` is set to `True` in [L511](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/kandinsky3/pipeline_kandinsky3_img2img.py#L511],) while for  pipe_without_text_encoders, this param is set to `False` by default [L116](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/kandinsky3/pipeline_kandinsky3_img2img.py#L116]),  hence there will be some minor difference between their outputs(in UNet [L492](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/unets/unet_kandinsky3.py#L492],) the mean value w/ and w/o `cut_context` is different). For platforms such as Intel XPU, they use their seperate generator [L145](https://github.com/huggingface/diffusers/blob/main/tests/pipelines/kandinsky3/test_kandinsky3_img2img.py#L145), hence the default tolerance value `1e-4` is not enough. Here we explicitly set `encode_prompt_inputs["_cut_context"] = True` to align with the behavior of full pipeline. @DN6 , pls help review, Thx!

### DIFF
--- a/src/diffusers/pipelines/kandinsky3/pipeline_kandinsky3_img2img.py
+++ b/src/diffusers/pipelines/kandinsky3/pipeline_kandinsky3_img2img.py
@@ -113,7 +113,7 @@ class Kandinsky3Img2ImgPipeline(DiffusionPipeline, StableDiffusionLoraLoaderMixi
         negative_prompt=None,
         prompt_embeds: Optional[torch.Tensor] = None,
         negative_prompt_embeds: Optional[torch.Tensor] = None,
-        _cut_context=False,
+        _cut_context=True,
         attention_mask: Optional[torch.Tensor] = None,
         negative_attention_mask: Optional[torch.Tensor] = None,
     ):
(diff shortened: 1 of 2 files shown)


kaixuanliu:
@a-r-r-o-w @DN6 ， Hi, can you help review? Thx!

regisss:
@kaixuanliu Does this only affect XPU? Or GPU too?
Can you also provide the test command to reproduce it please?

kaixuanliu:
@regisss This PR also affects GPU, although on GPU, the tolerance value is OK.  Steps to reproduce:
`pytest -rA tests/pipelines/kandinsky3/test_kandinsky3_img2img.py::Kandinsky3Img2ImgPipelineFastTests::test_encode_prompt_works_in_isolation`

DN6:
@kaixuanliu better solution here is to change the default value to True in the encode_prompt signature since it's hardcoded in the pipeline anyway.

kaixuanliu:
@DN6 I agree. Have updated the code.

PR #12475, opened 2025-10-13
[core] Refactor hub attn kernels

sayakpaul:
# What does this PR do?

Refactors how we load the attention kernels from the Hub. 

Currently, when a user specifies the `DIFFUSERS_ENABLE_HUB_KERNELS` env var, we _always_ download the supported kernel. Currently, we have FA3, but we have ongoing PRs that support FA and SAGE: https://github.com/huggingface/diffusers/pull/12387 and https://github.com/huggingface/diffusers/pull/12439. So, we will download ALL of them even when they're not required. This is not good. 

This PR makes it so that only the relevant kernel gets downloaded without breaking `torch.compile` compliance (fullgraph and no recompilation triggers).

Cc: @mekkcyber

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -73,6 +73,8 @@ jobs:
         run: |
           uv pip install -e ".[quality]"
           uv pip uninstall accelerate && uv pip install -U accelerate@git+https://github.com/huggingface/accelerate.git
+          #uv pip uninstall transformers huggingface_hub && uv pip install --prerelease allow -U transformers@git+https://github.com/huggingface/transformers.git
+          uv pip uninstall transformers huggingface_hub && uv pip install transformers==4.57.1 
           uv pip install pytest-reportlog
       - name: Environment
         run: |
@@ -84,7 +86,7 @@ jobs:
           CUBLAS_WORKSPACE_CONFIG: :16:8
         run: |
           pytest -n 1 --max-worker-restart=0 --dist=loadfile \
-            -k "not Flax and not Onnx" \
+             -k "not Flax and not Onnx" \
             --make-reports=tests_pipeline_${{ matrix.module }}_cuda \
             --report-log=tests_pipeline_${{ matrix.module }}_cuda.log \
             tests/pipelines/${{ matrix.module }}
@@ -126,6 +128,8 @@ jobs:
         uv pip install -e ".[quality]"
         uv pip install peft@git+https://github.com/huggingface/peft.git
         uv pip uninstall accelerate && uv pip install -U accelerate@git+https://github.com/huggingface/accelerate.git
+        #uv pip uninstall transformers huggingface_hub && uv pip install --prerelease allow -U transformers@git+https://github.com/huggingface/transformers.git
+        uv pip uninstall transformers huggingface_hub && uv pip install transformers==4.57.1 
         uv pip install pytest-reportlog
     - name: Environment
       run: python utils/print_env.py
@@ -190,6 +194,8 @@ jobs:
     - name: Install dependencies
       run: |
         uv pip install -e ".[quality,training]"
+        #uv pip uninstall transformers huggingface_hub && uv pip install --prerelease allow -U transformers@git+https://github.com/huggingface/transformers.git
+        uv pip uninstall transformers huggingface_hub && uv pip install transformers==4.57.1 
     - name: Environment
       run: |
         python utils/print_env.py
@@ -232,6 +238,8 @@ jobs:
           uv pip install -e ".[quality]"
           uv pip install peft@git+https://github.com/huggingface/peft.git
           uv pip uninstall accelerate && uv pip install -U accelerate@git+https://github.com/huggingface/accelerate.git
+          #uv pip uninstall transformers huggingface_hub && uv pip install --prerelease allow -U transformers@git+https://github.com/huggingface/transformers.git
+          uv pip uninstall transformers huggingface_hub && uv pip install transformers==4.57.1 
           uv pip install pytest-reportlog
       - name: Environment
         run: |
@@ -281,6 +289,8 @@ jobs:
           uv pip install -e ".[quality]"
           uv pip install peft@git+https://github.com/huggingface/peft.git
           uv pip uninstall accelerate && uv pip install -U accelerate@git+https://github.com/huggingface/accelerate.git
+          #uv pip uninstall transformers huggingface_hub && uv pip install --prerelease allow -U transformers@git+https://github.com/huggingface/transformers.git
+          uv pip uninstall transformers huggingface_hub && uv pip install transformers==4.57.1 
 
       - name: Environment
         run: |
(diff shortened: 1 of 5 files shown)


src/diffusers/models/attention_dispatch.py
```diff
@@ -452,6 +445,39 @@ def _check_attention_backend_requirements(backend: AttentionBackendName) -> None
...
+_BACKEND_PREPARERS: Dict[AttentionBackendName, Callable[[], None]] = {
+    AttentionBackendName._FLASH_3_HUB: _ensure_flash_attn_3_func_hub_loaded,
+}
+
+
+def _prepare_attention_backend(backend: AttentionBackendName) -> None:
+    preparer = _BACKEND_PREPARERS.get(backend)
+    if preparer is not None:
+        preparer()
+
```
MekkCyber: I'm not sure I understand what the `preparer` does

sayakpaul: https://github.com/huggingface/diffusers/blob/7fd26bc037313980443c97f7bb898866c8f2ead7/src/diffusers/models/attention_dispatch.py#L463

MekkCyber: Sounds good

src/diffusers/models/attention_dispatch.py
```diff
@@ -452,6 +445,39 @@ def _check_attention_backend_requirements(backend: AttentionBackendName) -> None
...
+
+def _ensure_attention_backend_ready(backend: AttentionBackendName) -> None:
```
DN6:
```suggestion
def _set_attention_backend(backend: AttentionBackendName) -> None:
```

DN6:
I think the control flow here is a bit difficult to follow. We should aim to minimize the number of new objects/concepts introduced in this module since there's already quite a lot of routing going on in here. 

My recommendations:
1. Add a `_set_attention_backend` function to `attention_dispatch.py` that handles checking requirements and downloading kernels if the are set. Call this function from modeling_utils when setting the backend

```python
def _set_attention_backend(backend: AttentionBackendName) -> None:
    _check_attention_backend_requirements(backend)
    _maybe_download_kernel_for_backend(backend)
```

In attention dispatch, let's create a `_HubKernelConfig` and `_HubKernelRegistry` 

```python
@dataclass
class _HubKernelConfig:
    """Configuration for downloading and using a hub-based attention kernel."""

    repo_id: str
    function_attr: str
    revision: Optional[str] = None
    kernel_fn: Optional[Callable] = None


# Registry for hub-based attention kernels
_HUB_KERNELS_REGISTRY: Dict["AttentionBackendName", _HubKernelConfig] = {
        repo_id="kernels-community/flash-attn3", function_attr="flash_attn_func", revision="fake-ops-return-probs"
    )
}
``` 

` _maybe_download_kernel_for_backend(backend)` would download the kernel and set the `kernel_fn` for a given backend if the backend is supported otherwise it's a no-op.

Then in your hub function, fetch the downloaded kernel from the registry 

```python
    func = _HUB_KERNELS_REGISTRY[AttentionBackendName._FLASH_3_HUB]._kernel_fn
    out = func(
        q=query,
        ....
```

We shouldn't attempt kernel downloads from the dispatch function. It should already be downloaded/available before hand.

sayakpaul:
Okay but 

* I would maintain a separate `_set_attention_backend()` function within `set_attention_backend()`? Does it not complicate things further? Would rather have `_check_attention_backend_requirements(backend)` and `_maybe_download_kernel_for_backend(backend)` directly in `set_attention_backend()`. That's better no? 
* How does `_maybe_download_kernel_for_backend(backend)` interact with the proposed registry? After downloading the kernel from the given config spec, we set the `_kernel_fn`?

DN6:
> I would maintain a separate _set_attention_backend() function within set_attention_backend()? Does it not complicate things further? Would rather have _check_attention_backend_requirements(backend) and _maybe_download_kernel_for_backend(backend) directly in set_attention_backend(). That's better no?

This is also good 👍🏽 

> How does _maybe_download_kernel_for_backend(backend) interact with the proposed registry? After downloading the kernel from the given config spec, we set the _kernel_fn?

Something like
```python
def _maybe_download_kernel_for_backend(backend: AttentionBackendName) -> None:
    if backend not in _HUB_KERNELS_REGISTRY:
        return
    config = _HUB_KERNELS_REGISTRY[backend]

    if config._kernel_fn is not None:
        return

    try:
        from kernels import get_kernel

        kernel_module = get_kernel(config.repo_id, revision=config.revision)
        kernel_func = getattr(kernel_module, config.function_attr)

        # Cache the downloaded kernel function in the config object
        config._kernel_fn = kernel_func

    except Exception as e:
        raise
  
```

sayakpaul:
@DN6 check now. Your feedback should have been addressed. I was able to completely get rid of `kernels_utils.py` as a consequence of that. I also decided to get rid of the `DIFFUSERS_ENABLE_HUB_KERNELS` environment var.

src/diffusers/models/attention_dispatch.py
```diff
@@ -249,6 +238,25 @@ def _is_context_parallel_enabled(
...
+    kernel_fn: Optional[Callable] = None
+
+
+# Registry for hub-based attention kernels
+_HUB_KERNELS_REGISTRY: Dict["AttentionBackendName", _HubKernelConfig] = {
+    # TODO: temporary revision for now. Remove when merged upstream into `main`.
+    AttentionBackendName._FLASH_3_HUB: _HubKernelConfig(
+        repo_id="kernels-community/flash-attn3", function_attr="flash_attn_func", revision="fake-ops-return-probs"
+    )
+}
```
MekkCyber, line 268: what about the other backends ?

MekkCyber (approved):
Nice clean up! thanks

src/diffusers/models/attention_dispatch.py
sayakpaul: This will be populated as we incorporate others.

MekkCyber: I mean shouldn't FA2 be here already ?

sayakpaul: PRs are not merged yet:
* https://github.com/huggingface/diffusers/pull/12387
* https://github.com/huggingface/diffusers/pull/12439

MekkCyber: ah okay sounds good !

sayakpaul:
@DN6 a gentle ping here.

src/diffusers/utils/constants.py
```diff
@@ -46,7 +46,6 @@
 DEFAULT_HF_PARALLEL_LOADING_WORKERS = 8
 HF_ENABLE_PARALLEL_LOADING = os.environ.get("HF_ENABLE_PARALLEL_LOADING", "").upper() in ENV_VARS_TRUE_VALUES
 DIFFUSERS_DISABLE_REMOTE_CODE = os.getenv("DIFFUSERS_DISABLE_REMOTE_CODE", "false").upper() in ENV_VARS_TRUE_VALUES
-DIFFUSERS_ENABLE_HUB_KERNELS = os.environ.get("DIFFUSERS_ENABLE_HUB_KERNELS", "").upper() in ENV_VARS_TRUE_VALUES
```
DN6: We still need this

PR #12486, opened 2025-10-14
[docs] Attention checks

stevhliu:
Adds docs for how to enable checks when using different attention backends

### DIFF
--- a/docs/source/en/api/pipelines/ltx_video.md
+++ b/docs/source/en/api/pipelines/ltx_video.md
@@ -254,8 +254,8 @@ export_to_video(video, "output.mp4", fps=24)
   pipeline.vae.enable_tiling()
 
   def round_to_nearest_resolution_acceptable_by_vae(height, width):
-      height = height - (height % pipeline.vae_temporal_compression_ratio)
-      width = width - (width % pipeline.vae_temporal_compression_ratio)
+      height = height - (height % pipeline.vae_spatial_compression_ratio)
+      width = width - (width % pipeline.vae_spatial_compression_ratio)
       return height, width
 
   prompt = """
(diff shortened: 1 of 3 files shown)


docs/source/en/optimization/attention_backends.md
```diff
@@ -81,6 +81,44 @@ with attention_backend("_flash_3_hub"):
...
+query = torch.randn(1, 10, 8, 64, dtype=torch.bfloat16, device="cuda")
+key = torch.randn(1, 10, 8, 64, dtype=torch.bfloat16, device="cuda")
+value = torch.randn(1, 10, 8, 64, dtype=torch.bfloat16, device="cuda")
+
+try:
+    with attention_backend("flash"):
+        output = dispatch_attention_fn(query, key, value)
+        print("✓ Flash Attention works with checks enabled")
+except Exception as e:
+    print(f"✗ Flash Attention failed: {e}")
```
sayakpaul, line 112: I think this needs to be set as `export DIFFUSERS_ATTN_CHECKS=yes` before running any execution because of how `DIFFUSERS_ATTN_CHECKS` is used:
https://github.com/huggingface/diffusers/blob/b4e6dc3037e75e2dc16466b914b1be597b06be9a/src/diffusers/models/attention_dispatch.py#L43

https://github.com/huggingface/diffusers/blob/b4e6dc3037e75e2dc16466b914b1be597b06be9a/src/diffusers/utils/constants.py#L45

stevhliu: this should be enabled here https://github.com/huggingface/diffusers/pull/12486/files#diff-249c332f10872ba150d634947de12055f99d9db5c3dbf2bef092ccba775f39bdR98 with `os.environ["DIFFUSERS_ATTN_CHECKS"] = "1"` no?

sayakpaul: Yeah `DIFFUSERS_ATTN_CHECKS` is determined at the module-level so I am not sure setting it that way would be effective.

stevhliu: Ah ok, updated to use `export DIFFUSERS_ATTN_CHECKS=yes` then!

PR #12491, opened 2025-10-15
[CI] Check links

stevhliu:
Adds link checker to CI based on suggestion in https://github.com/huggingface/diffusers/pull/12487


.github/workflows/check_documentation_links.yml
```diff
@@ -0,0 +1,45 @@
...
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+
+      - name: Install doc-builder
+        run: |
+          uv pip install --system git+https://github.com/huggingface/doc-builder.git@main
+
+      - name: Check documentation links
+        run: |
+          uv run doc-builder check-links docs/source/en
```
sayakpaul: We could also combine this here:

https://github.com/huggingface/diffusers/blob/aed636f5f089754425614872e0b4781dab65179f/.github/workflows/build_pr_documentation.yml#L14

So, we would first check for the links and then trigger
https://github.com/huggingface/diffusers/blob/aed636f5f089754425614872e0b4781dab65179f/.github/workflows/build_pr_documentation.yml#L15

WDYT?

stevhliu: Sounds good!

.github/workflows/check_documentation_links.yml
```diff
@@ -0,0 +1,45 @@
+name: Check Documentation Links
```
sayakpaul: Then we don't need this no?

sayakpaul (approved):
Just one comment. Thanks for doing this!

PR #12497, opened 2025-10-16
purge HF_HUB_ENABLE_HF_TRANSFER; promote Xet

Vaibhavs10:
(as the title state - since hf_transfer has no effect w/ xet, let's move to hf_xet)

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -42,18 +42,39 @@ jobs:
           CHANGED_FILES: ${{ steps.file_changes.outputs.all }}
         run: |
           echo "$CHANGED_FILES"
-          for FILE in $CHANGED_FILES; do 
+          ALLOWED_IMAGES=(
+            diffusers-pytorch-cpu
+            diffusers-pytorch-cuda
+            diffusers-pytorch-xformers-cuda
+            diffusers-pytorch-minimum-cuda
+            diffusers-doc-builder
+          )
+
+          declare -A IMAGES_TO_BUILD=()
+
+          for FILE in $CHANGED_FILES; do
             # skip anything that isn't still on disk
-            if [[ ! -f "$FILE" ]]; then
+            if [[ ! -e "$FILE" ]]; then
               echo "Skipping removed file $FILE"
               continue
-            fi           
-            if [[ "$FILE" == docker/*Dockerfile ]]; then
-              DOCKER_PATH="${FILE%/Dockerfile}"
-              DOCKER_TAG=$(basename "$DOCKER_PATH")
-              echo "Building Docker image for $DOCKER_TAG"
-              docker build -t "$DOCKER_TAG" "$DOCKER_PATH"
             fi
+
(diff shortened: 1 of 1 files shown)


docker/diffusers-doc-builder/Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN uv pip install --no-cache-dir "git+https://github.com/huggingface/diffusers.
 RUN uv pip install --no-cache-dir \
     accelerate \
     numpy==1.26.4 \
-    hf_transfer \
+    hf_xet \
```
sayakpaul, line 36: Let's keep the changes in `docker` limited to these ones:

https://github.com/huggingface/diffusers/blob/f072c64bf22c6a259d38435b19a8ef3eb67c96a8/.github/workflows/build_docker_images.yml#L73-L77

.github/workflows/benchmark.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   DIFFUSERS_IS_CI: yes
-  HF_HUB_ENABLE_HF_TRANSFER: 1
+  HF_XET_HIGH_PERFORMANCE: 1
```
sayakpaul, line 10: Will there be any problems stemming from this in case a repo isn't Xet-transferred? I am guessing this is currently the case i.e., not all repos have gone through this, yet. In that case, would using `HF_HUB_ENABLE_HF_TRANSFER` have any advantage?

sayakpaul (commented):
Thanks for starting it

.github/workflows/benchmark.yml
Vaibhavs10: good question, all repos on the hub are powered by `xet`. Hence `HF_HUB_ENABLE_HF_TRANSFER` has no usage whatsoever there

DN6 (approved):
Thanks 👍🏽

sayakpaul:
I think if we address https://github.com/huggingface/diffusers/pull/12497#discussion_r2435951393, the CI won't cry.

Vaibhavs10:
> I think if we address https://github.com/huggingface/diffusers/pull/12497#discussion_r2435951393, the CI won't cry.

Sorry for not understanding this correctly - but the whole point is to remove all mentions of hf_transfer (it has no effect on xet repos i.e. 100% of the hub) - so I'd rather fix why the CI is not happy with xet in the first place.

More so, the CI errors look more related to model specific integration test that hub - let me know if my understanding is incorrect.

yiyixuxu:
cc @DN6 
can you take a look? I think this docker-image fail to build pr=12497https://github.com/huggingface/diffusers/actions/runs/18688419900/job/53287370196?pr=12497

sayakpaul:
I have pushed a change in https://github.com/huggingface/diffusers/pull/12497/commits/708973b7ab96bd5bebd68fa7b9ab1eab22b0808a, which should fix the issue. The concerned Docker containers were previously not used in our CI (they are there for bookkeeping and will likely be removed soon).

sayakpaul:
Failing tests are unrelated.

PR #12498, opened 2025-10-16
GGUF fix for unquantized types when using unquantize kernels

dxqb:
Even if the `qweight_type` is one of the `UNQUANTIZED_TYPES`, qweight still has to be "dequantized" because it is stored as an 8-bit tensor. Without doing so, it is a shape mismatch in the following matmul.

Side notes:
 - https://huggingface.co/Isotr0py/ggml/tree/main/build has no build for torch 2.8 (or the upcoming 2.9). Who can we contact to make such a build?
 - thank you for this GGUF implementation! With torch 2.8 and torch.compile it is fast even without specialized kernels. torch 2.7 fails to compile the native dequantization code for some reason, so I couldn't directly compare torch.compile with the custom kernel.

## Who can review?
@DN6 @Isotr0py


Isotr0py:
> https://huggingface.co/Isotr0py/ggml/tree/main/build has no build for torch 2.8 (or the upcoming 2.9). Who can we contact to make such a build?

In fact, there is a pre-release building for torch 2.8 (https://huggingface.co/Isotr0py/ggml/tree/shmem-mmq/build), but I found there is some regression about kernel size and performance in these kernels.

Anyway, I have found out the root issue about the regression and fixing it, and will make a release with torch2.8 and 2.9 support tonight.

src/diffusers/quantizers/gguf/utils.py
```diff
@@ -79,7 +79,8 @@
 def _fused_mul_mat_gguf(x: torch.Tensor, qweight: torch.Tensor, qweight_type: int) -> torch.Tensor:
     # there is no need to call any kernel for fp16/bf16
     if qweight_type in UNQUANTIZED_TYPES:
-        return x @ qweight.T
+        weight = dequantize_gguf_tensor(qweight)
+        return x @ weight.T
```
Isotr0py, line 83: Seems `dequantize_gguf_tensor` missed implementation for FP16 and FP32 qweight:
https://github.com/huggingface/diffusers/blob/dbe413668dbf9d527944b51ef5f99a60b36a90be/src/diffusers/quantizers/gguf/utils.py#L487-L501

dxqb: it was bf16 in my use case.

fp16 and fp32 would fail in any case, whether native or dequant kernels are used.
this PR therefore currently only fixes the bf16 case for kernel dequant - for native bf16 already works

yiyixuxu:
cc @DN6 
I think it is ok to merge this one in, we can do a new PR if need to update dequantize_gguf_tensor, no?

PR #12500, opened 2025-10-17
adjust unit tests for `test_save_load_float16`

kaixuanliu:
When we run unit test like `pytest -rA tests/pipelines/wan/test_wan_22.py::Wan22PipelineFastTests::test_save_load_float16`, we found that the pipeline runs w/ all fp16 datatype, but after save and reload, some parts of text-encoder in `pipe_loaded` uses fp32, although we set  torch_dtype to fp16 explicitly. Deep investigation found that the root cause is here: [L783](https://github.com/huggingface/transformers/blob/v4.57.1/src/transformers/models/t5/modeling_t5.py#L783). Here we made an adjustment to the test case to manually add the `component = component.to(torch_device).half()` operation to align excatly with the behavior in `pipe`


kaixuanliu:
@a-r-r-o-w @DN6 pls help review, thx!

regisss:
Not sure I understand the issue here. This specific T5 module is kept in fp32 on purpose, why forcing a fp16 cast in the test?

kaixuanliu:
@regisss Hi, the purpose of this test case is to compare the output of pipelines using fp16 dtype(`pipe`) and the output of pipelines loaded from previously saved(`pipe_loaded`), they should be the same. However, all components of `pipe` is set to fp16 dtype in [L1424~L1426](https://github.com/huggingface/diffusers/blob/main/tests/pipelines/test_pipelines_common.py#L1424~L1426), while for `pipe_loaded`, some parts are kept in fp32, which does not match exactly with the computation in `pipe` fwd.

tests/pipelines/test_pipelines_common.py
```diff
@@ -1438,19 +1438,19 @@ def test_save_load_float16(self, expected_max_diff=1e-2):
...
                     component.set_default_attn_processor()
-            pipe_loaded.to(torch_device)
+                if hasattr(component, "dtype"):
+                    self.assertTrue(
+                        component.dtype == torch.float16,
+                        f"`{name}.dtype` switched from `float16` to {component.dtype} after loading.",
+                    )
+                if hasattr(component, "half"):
+                    # Although all components for pipe_loaded should be float16 now, some submodules still use fp32, like in https://github.com/huggingface/transformers/blob/v4.57.1/src/transformers/models/t5/modeling_t5.py#L783, so we need to do the conversion again manally to align with the datatype we use in pipe exactly
+                    component = component.to(torch_device).half()
```
sayakpaul: This doesn't seem right at all. `torch_dtype` should be able to take care of it. I just ran it on my GPU for SD and it worked fine.

kaixuanliu: Hi @sayakpaul , I tested on A100, and when I print `pipe_loaded.text_encoder.encoder.block[0].layer[1].DenseReluDense.wo.weight.dtype` in [L1455](https://github.com/huggingface/diffusers/blob/main/tests/pipelines/test_pipelines_common.py#L1455) , it returns `torch.float32`, not `torch.float16`, and the `max_diff` in [L1456](https://github.com/huggingface/diffusers/blob/main/tests/pipelines/test_pipelines_common.py#L1456) is `np.float16(0.0004883)`. When we apply this PR to  align excatly with the behavior in `pipe`, the `max_diff` is `0`.  I think it's better to adjust the test case to make the output comparison of `pipe` and `pipe_loaded` apple to apple.  WDYT?

sayakpaul: My point is `torch_dtype` in `from_pretrained()` should be enough for the model to be in fp16. Setting it with `half()` after loading the model in the FP16 `torch_dtype` seems erroneous to me. 

I also ran the test on an A100, and it wasn't a problem. So, I am not sure if this test fix is correct at all.

kaixuanliu: I printed `pipe_loaded.text_encoder.encoder.block[0].layer[1].DenseReluDense.wo.weight.dtype` after `pipe_loaded = self.pipeline_class.from_pretrained(tmpdir, torch_dtype=torch.float16)`, and it returns `torch.float32`, it is root caused in [L783](https://github.com/huggingface/transformers/blob/v4.57.1/src/transformers/models/t5/modeling_t5.py#L783), so I manualy add `.half()` to `pipe_loaded`, although it looks a bit wierd... On A100, the tolerance value is OK, but I think from the fundamentals perspective, the output from pipelines loaded from former saved should be exactly the same, that is the `max_diff` should be 0, right?

kaixuanliu:
@sayakpaul Hi, I adjusted the test code to pass `dtype` to `get_dummy_components`, instead of add `.half()` to every component, do you think it's OK now?

tests/pipelines/wan/test_wan_22.py
```diff
@@ -51,19 +51,19 @@ class Wan22PipelineFastTests(PipelineTesterMixin, unittest.TestCase):
     test_xformers_attention = False
     supports_dduf = False
 
-    def get_dummy_components(self):
+    def get_dummy_components(self, dtype=torch.float32):
```
sayakpaul: Why do we need this?

sayakpaul (commented):
I am honestly not sure about the changes introduced in this PR. We have gone over multiple comments and so far, I haven't been able to manually verify myself the failures this PR tries to solve.

tests/pipelines/wan/test_wan_22.py
kaixuanliu: pls refer to [L246-L256](https://github.com/huggingface/diffusers/blob/main/docs/source/zh/using-diffusers/schedulers.md?plain=1#L246-L256) (Sorry I only found Chinese version for this explanation). Using `torch.Tensor.to` method will convert all weights, while using `torch_dtype` parameter with `from_pretrained` will preserve layers in `_keep_in_fp32_modules`. For wan models, all components of [pipe](https://github.com/huggingface/diffusers/blob/v0.35.2/tests/pipelines/test_pipelines_common.py#L1427) will be fp16 dtype while it is not the case for [pipe_loaded](https://github.com/huggingface/diffusers/blob/v0.35.2/tests/pipelines/test_pipelines_common.py#L1439). Here I override `test_save_load_float16` function seperately for wan models.

tests/pipelines/wan/test_wan_22.py
```diff
@@ -155,6 +155,43 @@ def test_inference(self):
     def test_attention_slicing_forward_pass(self):
         pass
 
+    @unittest.skipIf(torch_device not in ["cuda", "xpu"], reason="float16 requires CUDA or XPU")
+    def test_save_load_float16(self, expected_max_diff=1e-2):
```
sayakpaul: I still don't know then how on my end the tests are passing.

kaixuanliu: I think it should be related with the input. When I set all the seed in [get_dummy_components](https://github.com/huggingface/diffusers/blob/v0.35.2/tests/pipelines/wan/test_wan_22.py#L54) to `1`, the [max_diff](https://github.com/huggingface/diffusers/blob/v0.35.2/tests/pipelines/test_pipelines_common.py#L1455) on A100 is `np.float16(0.2366)`, and when set seed to 42, the [output](https://github.com/huggingface/diffusers/blob/v0.35.2/tests/pipelines/test_pipelines_common.py#L1435) will be all `nan` value. After this PR, the `max_diff` will all be 0 for all the seed

tests/pipelines/wan/test_wan_22.py
```diff
@@ -51,51 +52,33 @@ class Wan22PipelineFastTests(PipelineTesterMixin, unittest.TestCase):
...
-            freq_dim=256,
-            ffn_dim=32,
-            num_layers=2,
-            cross_attn_norm=True,
-            qk_norm="rms_norm_across_heads",
-            rope_max_seq_len=32,
+        # Use from_pretrained with a tiny model to ensure proper dtype handling
+        # This ensures _keep_in_fp32_modules and _skip_layerwise_casting_patterns are respected
+        transformer = WanTransformer3DModel.from_pretrained(
+            "Kaixuanliu/tiny-random-wan-transformer",
```
kaixuanliu: pls replace my model space. We have to use `from_pretrained` here to make all the submodules' dtype correctly loaded.

tests/pipelines/wan/test_wan_22.py
```diff
@@ -51,51 +52,33 @@ class Wan22PipelineFastTests(PipelineTesterMixin, unittest.TestCase):
...
-            out_channels=16,
-            text_dim=32,
-            freq_dim=256,
-            ffn_dim=32,
-            num_layers=2,
-            cross_attn_norm=True,
-            qk_norm="rms_norm_across_heads",
-            rope_max_seq_len=32,
+        transformer_2 = WanTransformer3DModel.from_pretrained(
+            "Kaixuanliu/tiny-random-wan-transformer",
```
kaixuanliu: Same as above

kaixuanliu:
CC @yao-matrix

DN6:
I think it would be best to modify the `test_save_load_float16` test to account for keep in FP32 modules. e.g

```py
def test_save_load_float16(self, expected_max_diff=1e-2):
    for name, module in components.items():
        module = module.to(torch_device)
        # Account for components with _keep_in_fp32_modules
        if hasattr(module, "_keep_in_fp32_modules"):
            for name, param in module.named_parameters():
                if any(
                    module_to_keep_in_fp32 in name.split(".")
                    for module_to_keep_in_fp32 in module._keep_in_fp32_modules
                ):
                    param.data = param.data.to(torch.float32)
                else:
                    param.data = param.data.to(torch.float16)

        elif hasattr(module, "half"):
            components[name] = module.to(torch_device).half()
```

kaixuanliu:
@DN6 Hi, I think this is a good advice, looks much better now. Have updated the code following your advice, thx!

kaixuanliu:
@sayakpaul @DN6 Hi, can this PR be merged now?

kaixuanliu:
@sayakpaul Hi, the failed CI cases should have nothing to do with this PR, can you help merge?

sayakpaul:
Yeah will merge shortly. Thanks for your contributions!

PR #12502, opened 2025-10-17
Fix: Use incorrect temporary variable key when replacing adapter name…

FeiXie8:
Fix a variable name typo in the load_lora_adapter function where an incorrect temporary variable key was used instead of k when modifying state dict keys during LoRA adapter hotswapping.
The error occurred in the map_state_dict_for_hotswap helper function, specifically in the condition checking for lora_A.weight and lora_B.weight suffixes. This typo would have caused unexpected behavior when remapping keys for hotswapped adapters, as key was not defined in that scope.
This change ensures consistent use of the loop variable k to correctly modify the state dict keys, maintaining the intended logic for adapter hotswapping.

### DIFF
--- a/src/diffusers/utils/dynamic_modules_utils.py
+++ b/src/diffusers/utils/dynamic_modules_utils.py
@@ -151,8 +151,8 @@ def check_imports(filename):
             missing_packages.append(imp)
 
     if len(missing_packages) > 0:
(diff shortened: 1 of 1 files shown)


FeiXie8:
plz take some time to review it
@stevhliu @sayakpaul

PR #12508, opened 2025-10-19
Fix Chroma attention padding order and update docs to use `lodestones/Chroma1-HD`

josephrocca:
# What does this PR do?
Chroma inference is currently incorrect, since the padding token should be added for the transformer forward pass, not for the T5 encoder forward pass. The T5 embedding step should use the regular attention mask.

This change fixes that to align with [official code](https://github.com/lodestone-rock/flow/blob/1845f16c62b0355fe1c27ddc49c51a74b1309847/src/trainer/train_chroma.py#L532C50-L532C61) and ComfyUI.

I've also snuck in an update to use the final checkpoint in the docs/comments: https://huggingface.co/lodestones/Chroma1-HD

---

Top is before fix, bottom is after fix. I used `lodestones/Chroma1-Base`, since it's what I had on hand. Doesn't seem to be a huge difference, except for first column. Might have a stronger effect for shorter prompts, but I didn't test.

chroma_diffusers_t5_comparison


josephrocca:
Need to run a final test before it's ready for review, since I was testing on an older diffusers version.

josephrocca:
Okay all good. Unsure who to tag, but IIRC @sayakpaul merged Chroma impl originally.

docs/source/en/api/pipelines/chroma.md
```diff
@@ -19,20 +19,20 @@ specific language governing permissions and limitations under the License.
 
 Chroma is a text to image generation model based on Flux.
 
-Original model checkpoints for Chroma can be found [here](https://huggingface.co/lodestones/Chroma).
+Original model checkpoints for Chroma can be found [here](https://huggingface.co/lodestones/Chroma1-HD).
```
sayakpaul: This is intentional. As if there's any other official compatible checkpoint released under https://huggingface.co/lodestones/, the users will likely notice it.

src/diffusers/pipelines/chroma/pipeline_chroma.py
```diff
@@ -233,20 +233,21 @@ def _get_t5_prompt_embeds(
             return_tensors="pt",
         )
         text_input_ids = text_inputs.input_ids
-        attention_mask = text_inputs.attention_mask.clone()
+        tokenizer_mask = text_inputs.attention_mask
```
sayakpaul, line 236: Could we add some comment on how `tokenizer_mask` and `attention_mask` are differentiated from one another and which one is passed where?

docs/source/en/api/pipelines/chroma.md
josephrocca: https://huggingface.co/lodestones/Chroma is actually a deprecated repo, rather than a hub for all Chroma models. I don't think there's currently a 'hub' repo. Did you mean to imply that it should link to `https://huggingface.co/lodestones` rather than `https://huggingface.co/lodestones/Chroma` or `https://huggingface.co/lodestones/Chroma1-HD`?

josephrocca: I made some changes - unsure if this fixes your concerns.

DN6:
@bot /style

PR #12512, opened 2025-10-20
Avoiding graph break by changing the way we infer dtype in vae.decoder

### LINKED ISSUE #12501: VAE Decoder next(iter(..)) causes graph break
### Describe the bug

When doing torch.compile with the decoder part of the VAE, a graph break gets hit. This makes it impossible to compile the model as a whole. 

And these are the exact lines from `diffusers` library that cause the graph break:

`diffusers/src/diffusers/models/autoencoders/vae.py:287-289`

```
        sample = self.conv_in(sample)

        upscale_dtype = next(iter(self.up_blocks.parameters())).dtype
```

Particularly, the next(iter()) call causes the evaluation of the lazy tensors that preceed this part of the graph, forcing conv_in module to compile independently. 

It would be ideal if this upscale_dtype can be inferred in a different way

Thanks!

### Reproduction

This is the repro code:

```
import torch
from diffusers import AutoencoderKL


vae = AutoencoderKL.from_pretrained(
        subfolder="vae",
        torch_dtype=torch.float16
)

model = vae.decoder
model = model.eval()
model = model.to(torch.bfloat16)
device = torch.device("cpu")
sample_img = torch.randn(1, 4, 64, 64, dtype=torch.bfloat16)

model.compile()

sample_img = sample_img.to(device)
model = model.to(device)

with torch.no_grad():
    output = model(sample_img)
```

### Logs

```shell

```

### System Info

 🤗 Diffusers version: 0.35.1
- Platform: Linux-5.4.0-212-generic-x86_64-with-glibc2.35
- Python version: 3.11.13
- PyTorch version (GPU?): 2.7.0+cpu (False)
- Flax version (CPU?/GPU?/TPU?): 0.10.4 (cpu)
- Jax version: 0.7.1
- JaxLib version: 0.7.1
- Transformers version: 4.52.4
- Accelerate version: 1.10.1
- PEFT version: 0.17.1
- Safetensors version: 0.6.2
- Accelerator: NA
- Using GPU in script?: no, just cpu
- Using distributed or parallel set-up in script?: no, single core cpu

### Who can help?

@sayakpaul 
@DN6

ppadjinTT:
…sors

# What does this PR do?

This PR addresses the problem disscused in #12501, where the usage of `upscale_dtype = next(iter(self.up_blocks.parameters())).dtype` to infer the dtype in the forward pass of the `vae.decoder` causes the graph break when compiling the model with torch.compile.

The issue is that the usage of `next(iter(...))` forces the lazy tensors in the initial compiled model pass to materialize, resulting in graph break, which decreases performance. 

This PR proposes a simple fix by infering the `dtype` as:
```
upscale_dtype = self.conv_out.weight.dtype
```

Fixes #12501 

## Who can review?

@sayakpaul

### DIFF
--- a/src/diffusers/guiders/__init__.py
+++ b/src/diffusers/guiders/__init__.py
@@ -17,12 +17,6 @@ from typing import Union
 from ..utils import is_torch_available, logging
 
 
-logger = logging.get_logger(__name__)
-logger.warning(
-    "Guiders are currently an experimental feature under active development. The API is subject to breaking changes in future releases."
-)
-
-
 if is_torch_available():
     from .adaptive_projected_guidance import AdaptiveProjectedGuidance
     from .adaptive_projected_guidance_mix import AdaptiveProjectedMixGuidance
--- a/src/diffusers/guiders/guider_utils.py
+++ b/src/diffusers/guiders/guider_utils.py
@@ -41,6 +41,10 @@ class BaseGuidance(ConfigMixin, PushToHubMixin):
     _identifier_key = "__guidance_identifier__"
 
     def __init__(self, start: float = 0.0, stop: float = 1.0, enabled: bool = True):
+        logger.warning(
+            "Guiders are currently an experimental feature under active development. The API is subject to breaking changes in future releases."
+        )
+
         self._start = start
         self._stop = stop
         self._step: int = None
--- a/src/diffusers/pipelines/stable_cascade/pipeline_stable_cascade.py
+++ b/src/diffusers/pipelines/stable_cascade/pipeline_stable_cascade.py
@@ -21,7 +21,7 @@ from ...models import StableCascadeUNet
 from ...schedulers import DDPMWuerstchenScheduler
 from ...utils import is_torch_version, is_torch_xla_available, logging, replace_example_docstring
 from ...utils.torch_utils import randn_tensor
-from ..pipeline_utils import DiffusionPipeline, ImagePipelineOutput
+from ..pipeline_utils import DeprecatedPipelineMixin, DiffusionPipeline, ImagePipelineOutput
 from ..wuerstchen.modeling_paella_vq_model import PaellaVQModel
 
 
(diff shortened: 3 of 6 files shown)


sayakpaul:
@DN6 WDYT?

ppadjinTT:
I made sure all autoencoder tests are passing locally, I would be very thankful if you can take a look @DN6

src/diffusers/models/autoencoders/vae.py
```diff
@@ -286,7 +286,7 @@ def forward(
 
         sample = self.conv_in(sample)
 
-        upscale_dtype = next(iter(self.up_blocks.parameters())).dtype
+        upscale_dtype = self.up_blocks[0].resnets[0].norm1.weight.dtype
```
DN6: Think current failing tests in the CI are due to the fact that not every decoder block has a norm1 with a weight. Hence the use of the generator here to avoid such cases. 

@ppadjinTT I noticed you initially used `self.conv_out.weight` here? What was the issue you ran into with that?

ppadjinTT: okay, I will change that too, tnx! I intially changed the `self.conv_out.weight` because there are some tests that check what happens when conv_out and upscale_blocks have different dtypes

DN6: Could you point me to those tests? Seems like setting to conv_out is more robust.

ppadjinTT: Yup, these are the tests ```pytest -svvv tests/models/autoencoders/test_models_autoencoder_kl.py```

This is one of the tests from this test set that fails ```tests/models/autoencoders/test_models_autoencoder_kl.py::AutoencoderKLTests::test_layerwise_casting_inference```

ppadjinTT: I added better logic for inferring dtype, to capture the case where it doesn't work

DN6: Hmm I think we can remove `upscale_type` entirely here. I think all tests should still pass without it.

ppadjinTT: okay let's try that, i'm pushing the change

ppadjinTT: Do you think it's okay now? @DN6

ppadjinTT: Any chance you can take a look? Thanks for your effort @DN6

DN6:
Thanks @ppadjinTT 👍🏽

PR #12515, opened 2025-10-21
Add thread-safe wrappers for components in pipeline (examples/server-async/utils/requestscopedpipeline.py)

FredyRivera-dev:
# What does this PR do?

In the file `examples/server-async/utils/requestscopedpipeline.py`, add thread-safe wrappers for the tokenizer, VAE, and image processor. Introduce locking mechanisms to ensure thread safety during concurrent access.

This improves memory safety when using the VAE and ImageProcessor wrappers, helping to avoid memory leaks.

## Who can review?

Core library / Schedulers / Pipelines: @yiyixuxu, @asomoza, @sayakpaul  
General / integrations: @DN6, @stevhliu

### DIFF
--- a/src/diffusers/models/model_loading_utils.py
+++ b/src/diffusers/models/model_loading_utils.py
@@ -47,6 +47,7 @@ from ..utils import (
     is_torch_version,
     logging,
 )
+from ..utils.distributed_utils import is_torch_dist_rank_zero
 
 
 logger = logging.get_logger(__name__)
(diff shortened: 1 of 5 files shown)


FredyRivera-dev:
Ok, I'm going to move the wrappers to a separate file, and it was just an update to add wrappers to the VAE and the ImageProccesor to avoid memory leaks during decoding and saving of the image when the Pipeline is executed (requestscopedpipeline.py), sorry if it wasn't very clear

FredyRivera-dev:
Heeyyy @sayakpaul I already moved the wrappers to wrappers.py, check it out :b

sayakpaul:
@bot /style

PR #12526, opened 2025-10-21
[WIP]Add Wan2.2 Animate Pipeline (Continuation of #12442 by tolgacangoz)

### LINKED ISSUE #12441: Support Wan2.2-Animate
[Wan2.2-Animate-14B](https://humanaigc.github.io/wan-animate), it's a unified model for character animation and replacement, with holistic movement and expression replication.

https://github.com/user-attachments/assets/351227d0-4edc-4f6c-9bf9-053e53f218e4

We would like open to the community, if anyone is interested, to integrate this model with Diffusers. Just take into consideration this points:

1.  Don't integrate the preprocessing, we can help with that using a modular custom block.
2. This issue is for more advanced users than know the diffusers library very well.

Just let me know that you're interested and if you have any doubts, feel free to ask, if you open a PR we can help but just we're currently short in bandwidth so we ask you to be patient.

dg845:
# What does this PR do?

This PR is a continuation of #12442 by @tolgacangoz. It adds a pipeline for the Wan2.2-Animate-14B model [(project page](https://humanaigc.github.io/wan-animate/), [paper](https://arxiv.org/abs/2509.14055), [code](https://github.com/Wan-Video/Wan2.2), [weights](https://huggingface.co/Wan-AI/Wan2.2-Animate-14B)), a SOTA character animation and replacement video model.

Fixes #12441 (the original requesting issue).

## Who can review?

@yiyixuxu
@sayakpaul
@tolgacangoz

### DIFF
--- a/src/diffusers/pipelines/wan/pipeline_wan_animate.py
+++ b/src/diffusers/pipelines/wan/pipeline_wan_animate.py
@@ -65,13 +65,6 @@ EXAMPLE_DOC_STRING = """
         >>> image = load_image(
         ...     "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/astronaut.jpg"
         ... )
-        >>> # Resize image such that it splits evenly into spatial patches after VAE encoding
-        >>> max_area = 720 * 1280
-        >>> aspect_ratio = image.height / image.width
-        >>> mod_value = pipe.vae_scale_factor_spatial * pipe.transformer.config.patch_size[1]
-        >>> height = round(np.sqrt(max_area * aspect_ratio)) // mod_value * mod_value
-        >>> width = round(np.sqrt(max_area / aspect_ratio)) // mod_value * mod_value
-        >>> image = image.resize((width, height))
 
         >>> # Load pose and face videos (preprocessed from reference video)
         >>> # Note: Videos should be preprocessed to extract pose keypoints and face features
@@ -229,8 +222,13 @@ class WanAnimatePipeline(DiffusionPipeline, WanLoraLoaderMixin):
         self.video_processor_for_mask = VideoProcessor(
             vae_scale_factor=self.vae_scale_factor_spatial, do_normalize=False, do_convert_grayscale=True
         )
+        # In case self.transformer is None (e.g. for some pipeline tests)
+        spatial_patch_size = self.transformer.config.patch_size[-2:] if self.transformer is not None else (2, 2)
         self.vae_image_processor = WanAnimateImageProcessor(
-            vae_scale_factor=self.vae_scale_factor_spatial, resample="bilinear", fill_color=0
+            vae_scale_factor=self.vae_scale_factor_spatial,
+            spatial_patch_size=spatial_patch_size,
+            resample="bilinear",
+            fill_color=0,
         )
         self.image_processor = image_processor
 
--- a/src/diffusers/pipelines/wan/image_processor.py
+++ b/src/diffusers/pipelines/wan/image_processor.py
@@ -14,7 +14,9 @@
 
 from typing import Optional, Tuple, Union
 
+import numpy as np
 import PIL.Image
+import torch
 
 from ...configuration_utils import register_to_config
 from ...image_processor import VaeImageProcessor
@@ -30,7 +32,12 @@ class WanAnimateImageProcessor(VaeImageProcessor):
             Whether to downscale the image's (height, width) dimensions to multiples of `vae_scale_factor`. Can accept
             `height` and `width` arguments from [`image_processor.VaeImageProcessor.preprocess`] method.
         vae_scale_factor (`int`, *optional*, defaults to `8`):
-            VAE scale factor. If `do_resize` is `True`, the image is automatically resized to multiples of this factor.
+            VAE (spatial) scale factor. If `do_resize` is `True`, the image is automatically resized to multiples of
+            this factor.
+        vae_latent_channels (`int`, *optional*, defaults to `16`):
+            VAE latent channels.
+        spatial_patch_size (`Tuple[int, int]`, *optional*, defaults to `(2, 2)`):
+            The spatial patch size used by the diffusion transformer. For Wan models, this is typically (2, 2).
         resample (`str`, *optional*, defaults to `lanczos`):
             Resampling filter to use when resizing the image.
         do_normalize (`bool`, *optional*, defaults to `True`):
@@ -52,7 +59,8 @@ class WanAnimateImageProcessor(VaeImageProcessor):
         self,
         do_resize: bool = True,
         vae_scale_factor: int = 8,
-        vae_latent_channels: int = 4,
+        vae_latent_channels: int = 16,
+        spatial_patch_size: Tuple[int, int] = (2, 2),
         resample: str = "lanczos",
         reducing_gap: int = None,
         do_normalize: bool = True,
@@ -123,3 +131,55 @@ class WanAnimateImageProcessor(VaeImageProcessor):
                     )
 
         return res
+
+    def get_default_height_width(
+        self,
+        image: Union[PIL.Image.Image, np.ndarray, torch.Tensor],
+        height: Optional[int] = None,
+        width: Optional[int] = None,
+    ) -> Tuple[int, int]:
+        r"""
+        Returns the height and width of the image, downscaled to the next integer multiple of `vae_scale_factor`.
+
+        Args:
+            image (`Union[PIL.Image.Image, np.ndarray, torch.Tensor]`):
+                The image input, which can be a PIL image, NumPy array, or PyTorch tensor. If it is a NumPy array, it
+                should have shape `[batch, height, width]` or `[batch, height, width, channels]`. If it is a PyTorch
+                tensor, it should have shape `[batch, channels, height, width]`.
+            height (`Optional[int]`, *optional*, defaults to `None`):
+                The height of the preprocessed image. If `None`, the height of the `image` input will be used.
+            width (`Optional[int]`, *optional*, defaults to `None`):
+                The width of the preprocessed image. If `None`, the width of the `image` input will be used.
+
+        Returns:
+            `Tuple[int, int]`:
+                A tuple containing the height and width, both resized to the nearest integer multiple of
+                `vae_scale_factor * spatial_patch_size`.
+        """
+
+        if height is None:
+            if isinstance(image, PIL.Image.Image):
+                height = image.height
+            elif isinstance(image, torch.Tensor):
+                height = image.shape[2]
+            else:
+                height = image.shape[1]
+
+        if width is None:
+            if isinstance(image, PIL.Image.Image):
+                width = image.width
+            elif isinstance(image, torch.Tensor):
+                width = image.shape[3]
+            else:
+                width = image.shape[2]
+
+        max_area = width * height
+        aspect_ratio = height / width
+        mod_value_h = self.config.vae_scale_factor * self.config.spatial_patch_size[0]
+        mod_value_w = self.config.vae_scale_factor * self.config.spatial_patch_size[1]
+
+        # Try to preserve the aspect ratio
+        height = round(np.sqrt(max_area * aspect_ratio)) // mod_value_h * mod_value_h
+        width = round(np.sqrt(max_area / aspect_ratio)) // mod_value_w * mod_value_w
+
+        return height, width


src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,638 @@
...
+        self.conv2d = nn.Conv2d(in_channel, out_channel, kernel_size, stride, padding, bias=bias and not activate)
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        if self.downsample:
+            input = self.blur_conv(input)
+
+        input = self.conv2d(input)
+
+        if self.activate:
+            input = self.act(input + self.bias_leaky_relu) * 2**0.5
```
tolgacangoz: Hi. If fused leaky ReLU is supposed to be used that way, then this line should also be updated accordingly. I was trying to reduce the number of abstractions by removing the `FusedLeakyReLU` class.

tolgacangoz: Also, congrats for joining the `diffusers` team :congratulations:!

a-free-a:
@tolgacangoz @dg845 Congratulations on releasing Wan2.2-Animate-14B-Diffusers!
I attempted to use it for video generation but encountered a runtime failure. Could you please confirm whether additional code changes to the preprocessing stage are required to complete the full workflow?

src/diffusers/models/transformers/transformer_wan_animate.py
dg845: The relevant code has been refactored, so this should no longer apply. Feel free to reopen if there are further questions or comments.

dg845:
Here are some `diffusers` samples starting from preprocessed inputs from the [Wan2.2 official repo](https://github.com/Wan-Video/Wan2.2?tab=readme-ov-file#run-wan-animate).

Animation:

https://github.com/user-attachments/assets/801e40ec-e145-433e-aaaa-195cf6e45387

Replacement:

https://github.com/user-attachments/assets/92de5bf2-1fd5-4813-b4c2-1869529fb3cc

src/diffusers/image_processor.py
```diff
@@ -95,8 +95,11 @@ class VaeImageProcessor(ConfigMixin):
             `height` and `width` arguments from [`image_processor.VaeImageProcessor.preprocess`] method.
         vae_scale_factor (`int`, *optional*, defaults to `8`):
             VAE scale factor. If `do_resize` is `True`, the image is automatically resized to multiples of this factor.
-        resample (`str`, *optional*, defaults to `lanczos`):
-            Resampling filter to use when resizing the image.
+        resample (`str`, *optional*, defaults to `"lanczos"`):
```
yiyixuxu: can we add a new `WanVaeImageProcessor(VaeImageProcessor)` and put into wan folder, under `utils.py` file I think? 
(we start to see more and more custom preprocess methods, almost everyone does and they don't really get reused across models, I think moving forward let's just do this for all new models)

cc @DN6 here too, let me know what you think

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+        # Activation with fused bias, if using
+        if self.use_activation:
+            x = self.act_fn(x, channel_dim=channel_dim)
+        return x
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__name__}({self.weight.shape[1]}, {self.weight.shape[0]},"
+            f" kernel_size={self.weight.shape[2]}, stride={self.stride}, padding={self.padding})"
+        )
```
yiyixuxu, line 183:
```suggestion
```

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+        out = F.linear(input, self.weight * self.scale, bias=self.bias)
+        if self.use_activation:
+            out = self.act_fn(out, channel_dim=channel_dim)
+        return out
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__name__}(in_features={self.weight.shape[1]}, out_features={self.weight.shape[0]},"
+            f" bias={self.bias is not None})"
+        )
```
yiyixuxu, line 222:
```suggestion
```

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+
+        # 2. Patch embedding
+        hidden_states = self.patch_embedding(hidden_states)
+        pose_hidden_states = self.pose_patch_embedding(pose_hidden_states)
+        # Add pose embeddings to hidden states
+        hidden_states[:, :, 1:] = hidden_states[:, :, 1:] + pose_hidden_states
+        hidden_states = hidden_states.flatten(2).transpose(1, 2)
+
+        # 3. Condition embeddings (time, text, image)
+        # timestep shape: batch_size, or batch_size, seq_len (wan 2.2 ti2v)
```
yiyixuxu: I think we can move one of these conditions for animate, no?

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+        )
+
+        # 4. Output norm & projection
+        self.norm_out = FP32LayerNorm(inner_dim, eps, elementwise_affine=False)
+        self.proj_out = nn.Linear(inner_dim, out_channels * math.prod(patch_size))
+        self.scale_shift_table = nn.Parameter(torch.randn(1, 2, inner_dim) / inner_dim**0.5)
+
+        self.gradient_checkpointing = False
+
+    def motion_batch_encode(
```
yiyixuxu: can we move this to `forward`? all the layers (`motion_encoder` here)  should be visible in `forward`

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+        self, face_pixel_values: torch.Tensor, batch_size: int = 8, upcast_to_fp32: bool = True
+    ) -> torch.Tensor:
+        if batch_size >= face_pixel_values.shape[0]:
+            motion_vec = self.motion_encoder(face_pixel_values, upcast_to_fp32)
+        else:
+            motion_vec_batches = []
```
yiyixuxu: do you think we can do something like this to avoid `if else` here? (made up code, not tested)

```py
# Extract motion features using motion encoder
face_batches = torch.split(face_pixel_values, motion_encode_batch_size, ....)
motion_vec = torch.cat([
    self.motion_encoder(...) 
    for batch in face_batches
])
```

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        timestep: torch.LongTensor,
+        encoder_hidden_states: torch.Tensor,
+        encoder_hidden_states_image: Optional[torch.Tensor] = None,
+        pose_hidden_states: Optional[torch.Tensor] = None,
+        face_pixel_values: Optional[torch.Tensor] = None,
+        motion_encode_batch_size: int = 8,
```
yiyixuxu: user should be able to adjust the batch_size based on their memory, but I think it is more of a optimization config they set at runtime, instead of a pipeline argument, so maybe a method set_motion_encode_batch_size or something 
cc @DN6 @sayakpaul here, let me know what you guys think!

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+        upcast_to_fp32: bool = True,
```
yiyixuxu: if we need fp32 to have good result, let's just use fp32, remove this argument - not currently used in the pipeline anyway

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+        # Extract motion features using motion encoder
+        motion_vec = self.motion_batch_encode(face_pixel_values, motion_encode_batch_size, upcast_to_fp32)
+        motion_vec = motion_vec.view(batch_size, num_face_frames, -1)
+
+        # Now get face features from the motion vector
+        motion_vec = self.face_encoder(motion_vec)
+
+        # Add padding at the beginning (prepend zeros)
+        _, T_motion, H_motion, C_motion = motion_vec.shape
+        pad_face = torch.zeros(batch_size, 1, H_motion, C_motion, dtype=motion_vec.dtype, device=motion_vec.device)
```
yiyixuxu: would this work? not tested
```suggestion
       pad_face = torch.zeros_like(motion_vec[:, :1])
```

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+        for block_idx, block in enumerate(self.blocks):
+            if torch.is_grad_enabled() and self.gradient_checkpointing:
+                hidden_states = self._gradient_checkpointing_func(
+                    block, hidden_states, encoder_hidden_states, timestep_proj, rotary_emb
+                )
+            else:
+                hidden_states = block(hidden_states, encoder_hidden_states, timestep_proj, rotary_emb)
+
+            # Face adapter integration: apply after every 5th block (0, 5, 10, 15, ...)
+            if block_idx % self.config.inject_face_latents_blocks == 0:
```
yiyixuxu, line 1261: ohh does it make sense to create a `WanAnimateTransformerBlock` that includes `face_adapter`?
e.g.
```py
# In __init__
self.blocks = nn.ModuleList(
    [
        WanAnimateTransformerBlock( 
            ....,
            add_face_adapter=(i % inject_face_latents_blocks == 0),
        )
        for i in range(num_layers)
    ]
)
```

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+            shift, scale = (self.scale_shift_table.unsqueeze(0).to(temb.device) + temb.unsqueeze(2)).chunk(2, dim=2)
+            shift = shift.squeeze(2)
+            scale = scale.squeeze(2)
+        else:
+            # batch_size, inner_dim
+            shift, scale = (self.scale_shift_table.to(temb.device) + temb.unsqueeze(1)).chunk(2, dim=1)
+
+        hidden_states_original_dtype = hidden_states.dtype
+        hidden_states = self.norm_out(hidden_states.float())
+        # Move the shift and scale tensors to the same device as hidden_states.
```
yiyixuxu, line 1275: ohh let's try to fix it here
I think all we need to do is to pack shift and scale into same layer and add that layere into  `_no_split_modules` attribute

src/diffusers/pipelines/wan/pipeline_wan_animate.py
```diff
@@ -0,0 +1,1199 @@
...
+        ... )
+
+        >>> # Load pose and face videos (preprocessed from reference video)
+        >>> # Note: Videos should be preprocessed to extract pose keypoints and face features
+        >>> # Refer to the Wan-Animate preprocessing documentation for details
+        >>> pose_video = load_video("path/to/pose_video.mp4")
+        >>> face_video = load_video("path/to/face_video.mp4")
+
+        >>> # Calculate optimal dimensions based on VAE constraints
+        >>> max_area = 480 * 832
```
yiyixuxu: if we make a vaeimageprocessor for wan, this can be added there too

src/diffusers/pipelines/wan/pipeline_wan_animate.py
```diff
@@ -0,0 +1,1199 @@
...
+            .to(latents.device, latents.dtype)
+        )
+        latents_recip_std = 1.0 / torch.tensor(self.vae.config.latents_std).view(1, self.vae.config.z_dim, 1, 1, 1).to(
+            latents.device, latents.dtype
+        )
+
+        latents = (latents - latents_mean) * latents_recip_std
+        return latents
+
+    def destandardize_latents(self, latents: torch.Tensor) -> torch.Tensor:
```
yiyixuxu: can we put this back into pipeline, it is always part of pipeline

src/diffusers/pipelines/wan/pipeline_wan_animate.py
```diff
@@ -0,0 +1,1199 @@
...
+
+        if prev_segment_conditioning_frames is not None and (
+            not isinstance(prev_segment_conditioning_frames, int) or prev_segment_conditioning_frames not in (1, 5)
+        ):
+            raise ValueError(
+                f"`prev_segment_conditioning_frames` has to be of type `int` and 1 or 5 but its type is"
+                f" {type(prev_segment_conditioning_frames)} and value is {prev_segment_conditioning_frames}"
+            )
+
+    def standardize_latents(self, latents: torch.Tensor) -> torch.Tensor:
```
yiyixuxu: can we put this back into pipeline? let's try not to have too many small methods. and the normalize/denoramlize is always part of the pipeline code

src/diffusers/models/transformers/transformer_wan_animate.py
dg845: I kind of like having this because if you `print(model)` then `MotionConv2d` will print out info similar to `torch.nn.Conv2d` (same with `MotionLinear` below). But I'm also fine with removing it.

src/diffusers/image_processor.py
dg845: I think the changes which make `_resize_and_fill` and `_resize_and_crop` respect `self.config.resample` should be added to the base `VaeImageProcessor` class; this could also be spun off into its own PR. I agree with moving the other (Wan Animate-specific logic) into its own class.

src/diffusers/models/transformers/transformer_wan_animate.py
dg845: @yiyixuxu my understanding is that the main advantage of creating a `WanAnimateTransformerBlock` would be that if there is an associated face adapter block, it will always stay on the same device as the main transformer block (since we would put `WanAnimateTransformerBlock` in `_no_split_modules`). I think the code could look something like:

```python
class WanAnimateTransformerBlock(nn.Module):
	def __init__(self, ..., add_face_adapter: bool = False):
		super().__init__()
		self.transformer_block = WanTransformerBlock(...)
		self.face_cross_attn = None

		if add_face_adapter:
			self.face_cross_attn = WanAnimateFaceBlockCrossAttention(...)
	
	def forward(self, hidden_states, motion_vec, ...):
		hidden_states = self.transformer_block(hidden_states, ...)
		
		if self.face_cross_attn is not None:
			face_adapter_output = self.face_cross_attn(hidden_states, motion_vec)
			hidden_states = face_adapter_output + hidden_states
		
		return hidden_states
```

I think it is clearest to reuse `WanTransformerBlock` here so that it is easy to tell which logic is inherited from the base Wan 2.1 model and which is Wan Animate-specific. WDYT?

yiyixuxu: ohh so we are moving toward single-file format for models, so basically let's assume people who look through wan animate transformer do not need to know anything about wan transformer

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -0,0 +1,889 @@
...
+from ...configuration_utils import ConfigMixin, register_to_config
+from ...loaders import FromOriginalModelMixin, PeftAdapterMixin
+from ...utils import USE_PEFT_BACKEND, logging, scale_lora_layers, unscale_lora_layers
+from ..attention import AttentionMixin, AttentionModuleMixin
+from ..attention_dispatch import dispatch_attention_fn
+from ..cache_utils import CacheMixin
+from ..modeling_outputs import Transformer2DModelOutput
+from ..modeling_utils import ModelMixin
+from ..normalization import FP32LayerNorm
+from .transformer_wan import (
```
yiyixuxu: I missed a few imports here, I'm ok if you want to re-use `WanTransformerBlock` - but let's copy it here instead of importing from a different model file

src/diffusers/image_processor.py
yiyixuxu: sounds good

src/diffusers/models/transformers/transformer_wan_animate.py
yiyixuxu: ok if you feel strongly about keeping it!

src/diffusers/models/transformers/transformer_wan_animate.py
yiyixuxu: i left commeenets here too https://github.com/huggingface/diffusers/pull/12526/files#r2512642810
ok if you want to re-use the block! but let's not import from the wan tranformer file

src/diffusers/models/transformers/transformer_wan_animate.py
dg845: I have updated the code to use `torch.split` but it appears that using a list comprehension for batched motion encoder inference will use more memory than a `for` loop, so I have kept the `for` loop. (In particular, it appears to tip the memory usage from just fitting in memory on an A100 to an OOM error when in `replace` mode.)

src/diffusers/models/transformers/transformer_wan_animate.py
dg845: Yeah, Wan Animate is based on Wan 2.1, so the Wan2.2 TI2V logic isn't necessary here, and I have removed it.

src/diffusers/models/transformers/transformer_wan_animate.py
dg845: My current implementation sets a default value for the motion encoder batch size when initializing the transformer, then allows this value to be overridden in `WanAnimateTransformer3DModel.forward` and `WanAnimatePipeline.__call__`.

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -213,11 +218,7 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         p_t, p_h, p_w = self.patch_size
         ppf, pph, ppw = num_frames // p_t, height // p_h, width // p_w
 
-        split_sizes = [
-            self.attention_head_dim - 2 * (self.attention_head_dim // 3),
-            self.attention_head_dim // 3,
-            self.attention_head_dim // 3,
-        ]
+        split_sizes = [self.t_dim, self.h_dim, self.w_dim]
```
yiyixuxu, line 221: ohh I just removed the `#Copied from` in main, for sana video, `attention_head_dim` is 112, so the `split_size` is different with this change, not sure if it will affect the results if the model is already trained this way

but if you tested and think it's ok we can keep the change

docs/source/en/api/pipelines/wan.md
```diff
@@ -249,6 +250,220 @@ The code snippets available in [this](https://github.com/huggingface/diffusers/p
...
+)
+pipe.to("cuda")
+
+# Load character image and preprocessed videos
+image = load_image("path/to/character.jpg")
+pose_video = load_video("path/to/pose_video.mp4")  # Preprocessed skeletal keypoints
+face_video = load_video("path/to/face_video.mp4")  # Preprocessed facial features
+
+# Resize image to match VAE constraints
+def aspect_ratio_resize(image, pipe, max_area=720 * 1280):
```
yiyixuxu, line 311: can be in a follow-up PR but all these resize methods, including these for wan2.1 and wan2.2 5b, can be added to wan image processor now

src/diffusers/models/transformers/transformer_sana_video.py
dg845: As far as I can tell the generated samples seem to be the same after the VAE change.

PR #12543, opened 2025-10-25
[chore] Move guiders experimental warning

sayakpaul:
# What does this PR do?

If we run:

```py

pipe = StableDiffusionXLPipeline.from_pretrained("hf-internal-testing/tiny-stable-diffusion-xl-pipe")
```

we get:

```py
Guiders are currently an experimental feature under active development. The API is subject to breaking changes in future releases.
```

Despite the fact, `StableDiffusionXLPipeline` or the code above isn't using the guider component. Moving it to the init of `BaseGuidance` class is one way to solve it (which is what this PR does).


yiyixuxu (approved):
oh thanks for fixing this

PR #12546, opened 2025-10-26
Fix overflow and dtype handling in rgblike_to_depthmap (NumPy + PyTorch)

MohammadSadeghSalehi:
This PR fixes a potential overflow bug in `rgblike_to_depthmap` when computing
depth values from RGB-like depth images.

**Changes:**
- Cast image tensors/arrays to `int32` before performing the 16-bit combination (`*256 +`).
- Prevents overflow when input dtype is `uint8` or small int.
- Supports both PyTorch and NumPy paths consistently.
- Added comments for optional `uint16` casting.

**Rationale:**
The original implementation used `2**8` multiplication directly on small integer tensors,
which could overflow for high byte values (≥128). The fix ensures safe arithmetic and
consistent return types.

cc @patrickvonplaten @multimodalart @psuraj

### DIFF
--- a/src/diffusers/image_processor.py
+++ b/src/diffusers/image_processor.py
@@ -1049,31 +1049,31 @@ class VaeImageProcessorLDM3D(VaeImageProcessor):
         # 1. Cast the tensor to a larger integer type (e.g., int32)
         #    to safely perform the multiplication by 256.
         # 2. Perform the 16-bit combination: High-byte * 256 + Low-byte.
-        # 3. Cast the final result to the desired depth map type (uint16) if needed 
-        #    before returning, though leaving it as int32/int64 is often safer 
+        # 3. Cast the final result to the desired depth map type (uint16) if needed
+        #    before returning, though leaving it as int32/int64 is often safer
         #    for return value from a library function.
-        
+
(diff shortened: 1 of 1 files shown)


src/diffusers/image_processor.py
```diff
@@ -1045,16 +1045,37 @@ def depth_pil_to_numpy(images: Union[List[PIL.Image.Image], PIL.Image.Image]) ->
...
+            # Cast to a safe dtype (e.g., int32 or int64) for the calculation
+            image_safe = image.to(torch.int32) 
+            
+            # Calculate the depth map
+            depth_map = image_safe[:, :, 1] * 256 + image_safe[:, :, 2]
+            
+            # You may want to cast the final result to uint16, but casting to a 
+            # larger int type (like int32) is sufficient to fix the overflow.
+            # depth_map = depth_map.to(torch.uint16) # Uncomment if uint16 is strictly required
+            return depth_map
```
DN6: Change looks good. Could we just store the original dtype and cast back once the computation is done.
```python
original_dtype = image.dtype
# perform compute
return depth_map.to(original_dtype)
```

yiyixuxu:
@bot /style

PR #12548, opened 2025-10-27
[docs] put autopipeline after overview and hunyuanimage in images

sayakpaul:


sayakpaul:
> Thanks, it looks like AutoPipeline is still after the Audio section though instead of Overview

That is happening from `make style && make quality`. Any suggestions to bypass that?

stevhliu:
Hmm, will probably need to modify the `check_doc_toc` util to allow certain titles that can be placed in any order instead of alphabetizing it. I'll take a look and open a PR!

PR #12551, opened 2025-10-27
[ci] don't run sana layerwise casting tests in CI.

sayakpaul:
# What does this PR do?

Tests pass locally perfectly fine on different GPUs (RT 4090, A100, H100).

Example failures:
* https://github.com/huggingface/diffusers/actions/runs/18846080835/job/53770552500?pr=12524#step:7:2966
* https://github.com/huggingface/diffusers/actions/runs/18846080835/job/53770552496?pr=12524#step:7:9304

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -525,6 +525,8 @@
         title: Kandinsky 2.2
       - local: api/pipelines/kandinsky3
         title: Kandinsky 3
+      - local: api/pipelines/kandinsky5
+        title: Kandinsky 5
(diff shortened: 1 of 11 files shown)


sayakpaul:
@dg845 please check now. 

Regarding the Qwen failures, I am not too sure but OOMs in our CI can be also triggered by a prior test failure in the mix. I printed the `torch.cuda.max_memory_allocated()` for those concerned QwenImage tests in GB and it came out to be `0.4499330520629883 GB.`

So, I suspect these OOMs are caused by previous test failures. But LMK if you have an another perspective.

PR #12556, opened 2025-10-29
GGUF: torch.compile cannot trace sets

dxqb:
torch.compile fails to trace
```
s = { ... }
if x in s:
```

it can trace lists
## Who can review?
@Isotr0py

### DIFF
--- a/src/diffusers/quantizers/gguf/utils.py
+++ b/src/diffusers/quantizers/gguf/utils.py
@@ -41,23 +41,23 @@ if can_use_cuda_kernels and is_kernels_available():
 else:
     ops = None
 
-UNQUANTIZED_TYPES = {gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16, gguf.GGMLQuantizationType.BF16}
-STANDARD_QUANT_TYPES = {
+UNQUANTIZED_TYPES = [gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16, gguf.GGMLQuantizationType.BF16]
+STANDARD_QUANT_TYPES = [
     gguf.GGMLQuantizationType.Q4_0,
     gguf.GGMLQuantizationType.Q4_1,
     gguf.GGMLQuantizationType.Q5_0,
     gguf.GGMLQuantizationType.Q5_1,
     gguf.GGMLQuantizationType.Q8_0,
     gguf.GGMLQuantizationType.Q8_1,
-}
-KQUANT_TYPES = {
+]
+KQUANT_TYPES = [
     gguf.GGMLQuantizationType.Q2_K,
     gguf.GGMLQuantizationType.Q3_K,
     gguf.GGMLQuantizationType.Q4_K,
     gguf.GGMLQuantizationType.Q5_K,
     gguf.GGMLQuantizationType.Q6_K,
-}
-IMATRIX_QUANT_TYPES = {
+]
+IMATRIX_QUANT_TYPES = [
     gguf.GGMLQuantizationType.IQ1_M,
     gguf.GGMLQuantizationType.IQ1_S,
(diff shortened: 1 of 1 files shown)


src/diffusers/quantizers/gguf/utils.py
```diff
@@ -41,23 +41,23 @@
 else:
     ops = None
 
-UNQUANTIZED_TYPES = {gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16, gguf.GGMLQuantizationType.BF16}
-STANDARD_QUANT_TYPES = {
+UNQUANTIZED_TYPES = [gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16, gguf.GGMLQuantizationType.BF16]
+STANDARD_QUANT_TYPES = [
```
yiyixuxu, line 45: we use set operations on them like here https://github.com/huggingface/diffusers/blob/df8dd778177c3d2272f74cbebd880d7abd9f5ec9/src/diffusers/quantizers/gguf/utils.py#L74

would this work?

yiyixuxu:
thanks for the PR! could you provide a testing script?

src/diffusers/quantizers/gguf/utils.py
dxqb: you are right, this would not work and another solution is required

dxqb:
> thanks for the PR! could you provide a testing script?

```
import torch
import gguf

UNQUANTIZED_TYPES = {gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16, gguf.GGMLQuantizationType.BF16}

def f(x):
    if x in UNQUANTIZED_TYPES:
        pass

f_compiled = torch.compile(f, fullgraph=True)
f_compiled(gguf.GGMLQuantizationType.F32)

```

```
torch._dynamo.exc.Unsupported: Unsupported method call
  Explanation: Dynamo does not know how to trace method `__contains__` of class `set`
  Hint: Avoid calling `set.__contains__` in your code.
```

DN6:
@dxqb Could you provide an example where compile is failing with GGUF? I'm able to run this snippet without any issues?

```py
import torch
from diffusers import FluxPipeline, FluxTransformer2DModel, GGUFQuantizationConfig

ckpt_path = (
    "https://huggingface.co/city96/FLUX.1-dev-gguf/blob/main/flux1-dev-Q2_K.gguf"
)
transformer = FluxTransformer2DModel.from_single_file(
    ckpt_path,
    torch_dtype=torch.bfloat16,
)
transformer = torch.compile(transformer, fullgraph=True)

    model_id, transformer=transformer, torch_dtype=torch.bfloat16
)
pipe.to("cuda")
images = pipe("A cat holding a sign that says hello").images[0]
images.save("flux-gguf-compile.png")

```

dxqb:
> @dxqb Could you provide an example where compile is failing with GGUF? I'm able to run this snippet without any issues?
> 
the untracable line is here:
https://github.com/huggingface/diffusers/blob/1a1634dc08e7172c3f658dcc52978789d6c3e1e8/src/diffusers/quantizers/gguf/utils.py#L81

That code path is only reached if Isotr0py's kernels are used so it depends on several conditions, here:
https://github.com/huggingface/diffusers/blob/1a1634dc08e7172c3f658dcc52978789d6c3e1e8/src/diffusers/quantizers/gguf/utils.py#L33

When all these conditions are fulfilled, you should see it in your example. I've posted above a simpler test case that contains the same line so you can reproduce it in any case.

dxqb:
I'm going to close this for now, because this PR wasn't the right solution to the problem.
The problem still exists - but only if you use the kernels from huggingface for GGUF dequantization - and those haven't been updated since torch 2.6 so they cannot be used currently.

PR #12566, opened 2025-10-31
[tests] add tests for flux modular (t2i, i2i, kontext)

sayakpaul:
# What does this PR do?

The tests also helped me uncover some bugs and fix them. Some comments are in line.

### DIFF
--- a/src/diffusers/modular_pipelines/modular_pipeline.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline.py
@@ -2130,8 +2130,13 @@ class ModularPipeline(ConfigMixin, PushToHubMixin):
                         component_load_kwargs[key] = value["default"]
             try:
                 components_to_register[name] = spec.load(**component_load_kwargs)
-            except Exception as e:
-                logger.warning(f"Failed to create component '{name}': {e}")
+            except Exception:
+                logger.warning(
+                    f"\nFailed to create component {name}:\n"
+                    f"- Component spec: {spec}\n"
+                    f"- load() called with kwargs: {component_load_kwargs}\n\n"
+                    f"{traceback.format_exc()}"
+                )
 
         # Register all components at once
         self.register_components(**components_to_register)


src/diffusers/modular_pipelines/components_manager.py
```diff
@@ -164,7 +164,13 @@ def __call__(self, hooks, model_id, model, execution_device):
 
         device_type = execution_device.type
         device_module = getattr(torch, device_type, torch.cuda)
-        mem_on_device = device_module.mem_get_info(execution_device.index)[0]
+        try:
+            mem_on_device = device_module.mem_get_info(execution_device.index)[0]
+        except AttributeError:
+            try:
+                mem_on_device = device_module.recommended_max_memory()
```
sayakpaul: For MPS. `torch.mps` doesn't have `mem_get_info()`.

Open to other suggestions to tackle it.

src/diffusers/modular_pipelines/flux/denoise.py
```diff
@@ -59,7 +59,7 @@ def inputs(self) -> List[Tuple[str, Any]]:
             ),
             InputParam(
                 "guidance",
-                required=True,
+                required=False,
```
sayakpaul, line 62: For Flux.1-schnell, `guidance` shouldn't be required.

src/diffusers/modular_pipelines/flux/encoders.py
```diff
@@ -95,7 +95,7 @@ def expected_components(self) -> List[ComponentSpec]:
             ComponentSpec(
                 "image_processor",
                 VaeImageProcessor,
-                config=FrozenDict({"vae_scale_factor": 16}),
+                config=FrozenDict({"vae_scale_factor": 16, "vae_latent_channels": 16}),
```
sayakpaul, line 98: Just to be consistent.

src/diffusers/modular_pipelines/flux/inputs.py
```diff
@@ -112,6 +112,10 @@ def __call__(self, components: FluxModularPipeline, state: PipelineState) -> Pip
         block_state.prompt_embeds = block_state.prompt_embeds.view(
             block_state.batch_size * block_state.num_images_per_prompt, seq_len, -1
         )
+        pooled_prompt_embeds = block_state.pooled_prompt_embeds.repeat(1, block_state.num_images_per_prompt)
+        block_state.pooled_prompt_embeds = pooled_prompt_embeds.view(
+            block_state.batch_size * block_state.num_images_per_prompt, -1
+        )
```
sayakpaul, line 118: Was left out of the expansion incorrectly.

tests/modular_pipelines/flux/test_modular_pipeline_flux.py
```diff
@@ -0,0 +1,154 @@
...
+from diffusers.modular_pipelines.modular_pipeline_utils import InsertableDict
+
+from ...testing_utils import floats_tensor, torch_device
+from ..test_modular_pipelines_common import ModularPipelineTesterMixin
+
+
+# Because we should disable `auto_resize` during tests.
+FluxKontextVaeEncoderBlocks = InsertableDict(
+    [
+        ("preprocess", FluxKontextProcessImagesInputStep(_auto_resize=False)),
```
sayakpaul: For facilitating `_auto_resize=False` in testing. 

We don't take the `_auto_resize` argument as a call argument from the user when using Kontext modular.

tests/modular_pipelines/stable_diffusion_xl/test_modular_pipeline_stable_diffusion_xl.py
```diff
@@ -21,24 +21,12 @@
 import torch
 from PIL import Image
 
-from diffusers import (
```
sayakpaul, line 24: Unrelated changes but I thought I would do it (happy to revert).

src/diffusers/modular_pipelines/components_manager.py
yiyixuxu: oh should we throw an error here too for MPS, since we are not supporting it yet 
https://github.com/huggingface/diffusers/blob/main/src/diffusers/modular_pipelines/components_manager.py#L682

src/diffusers/modular_pipelines/flux/denoise.py
```diff
@@ -182,6 +182,7 @@ def __call__(
             latent_model_input = torch.cat([latent_model_input, image_latents], dim=1)
 
         timestep = t.expand(latents.shape[0]).to(latents.dtype)
+        print(f"{latents.shape=}, {timestep.shape=}")
```
yiyixuxu:
```suggestion
```

src/diffusers/modular_pipelines/components_manager.py
```diff
@@ -166,6 +166,7 @@ def __call__(self, hooks, model_id, model, execution_device):
         device_module = getattr(torch, device_type, torch.cuda)
         mem_on_device = device_module.mem_get_info(execution_device.index)[0]
         mem_on_device = mem_on_device - self.memory_reserve_margin
+
```
yiyixuxu, line 171: ohh, what happended here? 
https://github.com/huggingface/diffusers/pull/12566#discussion_r2482332439
It was a good catch, I think we should also add a warning sooner during `enable_model_offload()`

sayakpaul: My bad for misunderstanding!

I have added the error catching. I added a TODO in `enable_auto_cpu_offload()` to error out early for `mem_get_info()`.

PR #12577, opened 2025-11-02
[modular]pass hub_kwargs to load_config

yiyixuxu:
download the config file to local_dir too, so that this works

```
from diffusers import ModularPipelineBlocks

repo_id = "diffusers-internal-dev/krt"
local_dir = "krea-realtime-video"
blocks_1 = ModularPipelineBlocks.from_pretrained(repo_id, local_dir="krea-realtime-video", trust_remote_code=True)
blocks_2 = ModularPipelineBlocks.from_pretrained(local_dir, trust_remote_code=True)
print(f"blocks_1: {blocks_1}")
print(f"blocks_2: {blocks_2}")
```

### DIFF
--- a/src/diffusers/modular_pipelines/modular_pipeline.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline.py
@@ -313,7 +313,7 @@ class ModularPipelineBlocks(ConfigMixin, PushToHubMixin):
         ]
         hub_kwargs = {name: kwargs.pop(name) for name in hub_kwargs_names if name in kwargs}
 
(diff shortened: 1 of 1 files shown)


haochengxi:
Hi, is `diffusers-internal-dev/krt` available now? I want to try this model, but I can not access `https://huggingface.co/diffusers-internal-dev/krt` which says `Sorry, we can't find the page you are looking for.`

DN6:
@haochengxi Please use via `krea/krea-realtime-video`

PR #12579, opened 2025-11-03
[Modular] Some clean up for Modular tests

DN6:
# What does this PR do?
Some changes to make modular tests more in line with our new testing approach

1. Remove `unittest` dependency 
2. Move `get_pipeline` to ModularPipelineTesterMixin (it's most likely going to be the same across all pipelines)
3. Create distinction between tests and test mixins. Since we don't use `unittest` anymore, we need to be explicit about which objects are Mixins and which objects are actually running the tests and need attributes such as `param`, `repo` etc. 
4. Always use `cpu` based Generator through the `self.get_generator` method in `ModularPipelineTesterMixin` for better reproducibility (GPU based generator can introduce noise)
5. Use `torch.abs().max` for comparing slices. Let's move away from `np` conversions. There isn't much of an advantage and when we run on GPU we can avoid `.detach().cpu().numpy()` calls 
6. Use single level inheritance for easier debugging. 
 

Fixes # (issue)

## Before submitting

## Who can review?


tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -57,9 +47,8 @@ class ModularPipelineTesterMixin:
         ]
     )
 
-    def get_generator(self, seed):
-        device = torch_device if torch_device != "mps" else "cpu"
-        generator = torch.Generator(device).manual_seed(seed)
+    def get_generator(self, seed=0):
+        generator = torch.Generator("cpu").manual_seed(seed)
```
sayakpaul, line 51: (nit): `torch.manual_seed(seed)` is also enough.

sayakpaul:
I will propagate this to #12585! Thanks for this PR

PR #12581, opened 2025-11-03
[LoRA] add support for more Qwen LoRAs

linoytsaban:
add support for [dx8152's](https://huggingface.co/dx8152) Qwen loras (e.g. https://huggingface.co/dx8152/Qwen-Edit-2509-Multiple-angles)

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -529,8 +529,6 @@
         title: Kandinsky 2.2
       - local: api/pipelines/kandinsky3
         title: Kandinsky 3
-      - local: api/pipelines/kandinsky5
-        title: Kandinsky 5
       - local: api/pipelines/kolors
         title: Kolors
       - local: api/pipelines/latent_consistency_models
@@ -638,6 +636,8 @@
         title: HunyuanVideo
       - local: api/pipelines/i2vgenxl
         title: I2VGen-XL
+      - local: api/pipelines/kandinsky5_video
+        title: Kandinsky 5.0 Video
       - local: api/pipelines/latte
         title: Latte
       - local: api/pipelines/ltx_video
--- a/docs/source/en/api/pipelines/kandinsky5_video.md
+++ b/docs/source/en/api/pipelines/kandinsky5_video.md
@@ -7,9 +7,9 @@ an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express o
 specific language governing permissions and limitations under the License.
 -->
 
-# Kandinsky 5.0
+# Kandinsky 5.0 Video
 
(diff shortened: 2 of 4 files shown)


linoytsaban:
@bot /style

src/diffusers/loaders/lora_conversion_utils.py
```diff
@@ -2213,6 +2213,10 @@ def convert_key(key: str) -> str:
 
         state_dict = {convert_key(k): v for k, v in state_dict.items()}
 
+    has_default = any("default." in k for k in state_dict)
```
sayakpaul, line 2216: If it's starting with "default.", then let's be explicit about that:
```suggestion
    has_default = any(k.startswith("default.") for k in state_dict)
```

src/diffusers/loaders/lora_conversion_utils.py
```diff
@@ -2213,6 +2213,10 @@ def convert_key(key: str) -> str:
...
+    if has_default:
+        state_dict = {k.replace("default.", ""): v for k, v in state_dict.items()}
```
sayakpaul, line 2218: So, that it's done as intended:

```suggestion
    if has_default:
        state_dict = {k[len("default."):]: v for k, v in state_dict.items()}
```

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -4940,7 +4940,8 @@ def lora_state_dict(
         has_alphas_in_sd = any(k.endswith(".alpha") for k in state_dict)
         has_lora_unet = any(k.startswith("lora_unet_") for k in state_dict)
         has_diffusion_model = any(k.startswith("diffusion_model.") for k in state_dict)
-        if has_alphas_in_sd or has_lora_unet or has_diffusion_model:
+        has_default = any("default." in k for k in state_dict)
```
sayakpaul, line 4943: Same as above.

sayakpaul (commented):
Thank you! I left some minor yet important nits. Once those are addressed, we can ship!

src/diffusers/loaders/lora_conversion_utils.py
linoytsaban: "default" isn't the prefix though, e.g.`transformer_blocks.0.attn.add_v_proj.lora_A.default.weight`

src/diffusers/loaders/lora_conversion_utils.py
linoytsaban: same comment as [here](https://github.com/huggingface/diffusers/pull/12581#discussion_r2489926544) - 'default' is not in the key's prefix, so this won't be the intended behavior in this case

src/diffusers/loaders/lora_pipeline.py
linoytsaban: same comment as https://github.com/huggingface/diffusers/pull/12581#discussion_r2489926544 - 'default' is not in the key's prefix, so this won't be the intended behavior in this case

PR #11916, opened 2025-07-14
[refactor] Flux/Chroma single file implementation + Attention Dispatcher

a-r-r-o-w:
This PR is a continuation of #11368.

Using `set_attention_backend`:

```python
import torch
from diffusers import FluxPipeline, FluxTransformer2DModel, attention_backend

model_id = "black-forest-labs/FLUX.1-dev"
transformer = FluxTransformer2DModel.from_pretrained(model_id, subfolder="transformer", torch_dtype=torch.bfloat16, device_map="cuda")
transformer.set_attention_backend("sage_varlen")
pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", transformer=transformer, torch_dtype=torch.bfloat16)
pipe.text_encoder.to("cuda")
pipe.text_encoder_2.to("cuda")
pipe.vae.to("cuda")

prompt = "A cat holding a sign that says 'hello world'"

image = pipe(prompt, num_inference_steps=28, guidance_scale=4.0).images[0]
image.save("output.png")
```

Using context manager `attention_backend`:

```python
import torch
from diffusers import FluxPipeline, FluxTransformer2DModel, attention_backend

model_id = "black-forest-labs/FLUX.1-dev"
transformer = FluxTransformer2DModel.from_pretrained(model_id, subfolder="transformer", torch_dtype=torch.bfloat16, device_map="cuda")
pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-dev", transformer=transformer, torch_dtype=torch.bfloat16)
pipe.text_encoder.to("cuda")
pipe.text_encoder_2.to("cuda")
pipe.vae.to("cuda")

prompt = "A cat holding a sign that says 'hello world'"

with attention_backend("_native_cudnn"):
    image = pipe(prompt, num_inference_steps=28, guidance_scale=4.0).images[0]
image.save("output.png")
```

PR #12584, opened 2025-11-04
[SANA-Video] Adding 5s pre-trained 480p SANA-Video inference

lawrence-cj:
# What does this PR do?
This PR add SANA-Video, a new text/image-to-video model from NVIDIA
[Paper](https://arxiv.org/abs/2509.24695)
[Project](https://nvlabs.github.io/Sana/Video)
[HF weight](https://huggingface.co/Efficient-Large-Model/SANA-Video_2B_480p_diffusers)

Cc: @yiyixuxu @asomoza @sayakpaul 
```python
import torch
from diffusers import SanaPipeline, SanaVideoPipeline, UniPCMultistepScheduler, DPMSolverMultistepScheduler
from diffusers import AutoencoderKLWan


model_id = "Efficient-Large-Model/SANA-Video_2B_480p_diffusers"
pipe = SanaVideoPipeline.from_pretrained(model_id, torch_dtype=torch.bfloat16)
# pipe.scheduler = UniPCMultistepScheduler.from_config(pipe.scheduler.config, flow_shift=8.0)
# pipe.scheduler = DPMSolverMultistepScheduler.from_config(pipe.scheduler.config, flow_shift=8.0)
pipe.vae.to(torch.float32)
pipe.to("cuda")
model_score = 30

prompt = "Evening, backlight, side lighting, soft light, high contrast, mid-shot, centered composition, clean solo shot, warm color. A young Caucasian man stands in a forest, golden light glimmers on his hair as sunlight filters through the leaves. He wears a light shirt, wind gently blowing his hair and collar, light dances across his face with his movements. The background is blurred, with dappled light and soft tree shadows in the distance. The camera focuses on his lifted gaze, clear and emotional."
negative_prompt = "A chaotic sequence with misshapen, deformed limbs in heavy motion blur, sudden disappearance, jump cuts, jerky movements, rapid shot changes, frames out of sync, inconsistent character shapes, temporal artifacts, jitter, and ghosting effects, creating a disorienting visual experience."
motion_prompt = f" motion score: {model_score}."
prompt = prompt + motion_prompt

video = pipe(
    prompt=prompt,
    height=480,
    width=832,
    frames=81,
    guidance_scale=6,
    num_inference_steps=50,
).frames[0]

export_to_video(video, "sana_video.mp4", fps=16)
```

Results:

https://github.com/user-attachments/assets/bf914906-b974-4bea-a6ae-cf9914f49a68


src/diffusers/video_processor.py
```diff
@@ -111,3 +112,65 @@ def postprocess_video(
...
+        Returns:
+            `Tuple[int, int]`: The closest binned height and width.
+        """
+        ar = float(height / width)
+        closest_ratio = min(ratios.keys(), key=lambda ratio: abs(float(ratio) - ar))
+        default_hw = ratios[closest_ratio]
+        return int(default_hw[0]), int(default_hw[1])
+
+    @staticmethod
+    def resize_and_crop_tensor(samples: torch.Tensor, new_width: int, new_height: int) -> torch.Tensor:
```
dg845, line 135: I think exposing an interface like `VaeImageProcessor.resize`:

https://github.com/huggingface/diffusers/blob/dcfb18a2d340d8e1f0ff001b06d2931ffa8648da/src/diffusers/image_processor.py#L468-L474

would be more robust, since different video preprocessing pipelines will probably make different choices here. Not blocking, on the `diffusers` side we can follow up to support more video pipelines here.

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -0,0 +1,752 @@
...
+        hidden_states = hidden_states.flatten(1, 2).transpose(1, 2)
+        hidden_states = hidden_states.to(original_dtype)
+
+        hidden_states = attn.to_out[0](hidden_states)
+        hidden_states = attn.to_out[1](hidden_states)
+
+        return hidden_states
+
+
+# copy from https://github.com/huggingface/diffusers/blob/11d22e0e809d1219a067ded8a18f7b0129fc58c7/src/diffusers/models/transformers/transformer_wan.py#L410
```
dg845:
```suggestion
# Copied from diffusers.models.transformers.transformer_wan.WanRotaryPosEmbed
```
I think this is the intended `# Copied from` statement? The link points to `transformer_wan.WanTransformerBlock`, which seems like a typo.

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -0,0 +1,752 @@
...
+        query = query.view(batch_size, -1, attn.heads, head_dim).transpose(1, 2)
+
+        key = key.view(batch_size, -1, attn.heads, head_dim).transpose(1, 2)
+        value = value.view(batch_size, -1, attn.heads, head_dim).transpose(1, 2)
+
+        # the output of sdp = (batch, num_heads, seq_len, head_dim)
+        # TODO: add support for attn.scale when we move to Torch 2.1
+        hidden_states = F.scaled_dot_product_attention(
+            query, key, value, attn_mask=attention_mask, dropout_p=0.0, is_causal=False
+        )
```
dg845: Not a blocker, but if it makes sense for the Sana Video model to use different attention backends (e.g. `F.scaled_dot_product_attention`, Flash Attention, etc.), you could consider using [`dispatch_attention_fn`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention_dispatch.py) here to support them.

You can look at the Wan attention processor for an example:

https://github.com/huggingface/diffusers/blob/dcfb18a2d340d8e1f0ff001b06d2931ffa8648da/src/diffusers/models/transformers/transformer_wan.py#L142-L151

See PR #11916 and the [attention backend docs](https://huggingface.co/docs/diffusers/main/en/optimization/attention_backends) for more info.

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -0,0 +1,752 @@
...
+        hidden_states = attn.to_out[1](hidden_states)
+
+        hidden_states = hidden_states / attn.rescale_output_factor
+
+        return hidden_states
+
+
+class SanaVideoTransformerBlock(nn.Module):
+    r"""
+    Transformer block introduced in [Sana](https://huggingface.co/papers/2410.10629).
```
dg845:
```suggestion
    Transformer block introduced in [Sana-Video](https://huggingface.co/papers/2509.24695).
```
nit: should reference Sana-Video paper rather than original Sana paper?

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -0,0 +1,752 @@
...
+        ff_output = self.ff(norm_hidden_states)
+        ff_output = ff_output.flatten(1, 3)
+        hidden_states = hidden_states + gate_mlp * ff_output
+
+        return hidden_states
+
+
+class SanaVideoTransformer3DModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOriginalModelMixin):
+    r"""
+    A 2D Transformer model introduced in [Sana](https://huggingface.co/papers/2410.10629) family of models.
```
dg845:
```suggestion
    A 3D Transformer model introduced in [Sana-Video](https://huggingface.co/papers/2509.24695) family of models.
```
nit: should reference Sana-Video paper rather than original Sana paper?

src/diffusers/pipelines/sana/pipeline_sana_video.py
```diff
@@ -0,0 +1,1045 @@
...
+                    f"{e}. \n"
+                    f"Try to use VAE tiling for large images. For example: \n"
+                    f"pipe.vae.enable_tiling(tile_sample_min_width=512, tile_sample_min_height=512)"
+                )
+
+            if use_resolution_binning:
+                video = self.video_processor.resize_and_crop_tensor(video, orig_width, orig_height)
+
+        if not output_type == "latent":
+            video = self.video_processor.postprocess_video(video, output_type=output_type)
```
yiyixuxu:
```suggestion
            video = self.video_processor.postprocess_video(video, output_type=output_type)
```

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -0,0 +1,752 @@
...
+
+        norm_hidden_states = norm_hidden_states.unflatten(1, (frames, height, width))
+        ff_output = self.ff(norm_hidden_states)
+        ff_output = ff_output.flatten(1, 3)
+        hidden_states = hidden_states + gate_mlp * ff_output
+
+        return hidden_states
+
+
+class SanaVideoTransformer3DModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOriginalModelMixin):
```
dg845:
```suggestion
class SanaVideoTransformer3DModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOriginalModelMixin, AttentionMixin):
```
[`AttentionMixin`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention.py) implements `attn_processors` and `set_attn_processor` methods, so inheriting from it allows us to not have to reimplement them below.

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -0,0 +1,752 @@
...
+                    module.set_processor(processor)
+                else:
+                    module.set_processor(processor.pop(f"{name}.processor"))
+
+            for sub_name, child in module.named_children():
+                fn_recursive_attn_processor(f"{name}.{sub_name}", child, processor)
+
+        for name, module in self.named_children():
+            fn_recursive_attn_processor(name, module, processor)
+
```
dg845:
```suggestion
```
Coupled with https://github.com/huggingface/diffusers/pull/12584#discussion_r2492402412. Inheriting from `AttentionMixin` should allow us to reuse the `attn_processors` and `set_attn_processor` methods implemented there.

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -0,0 +1,752 @@
...
+        freqs_sin_h = freqs_sin[1][:pph].view(1, pph, 1, -1).expand(ppf, pph, ppw, -1)
+        freqs_sin_w = freqs_sin[2][:ppw].view(1, 1, ppw, -1).expand(ppf, pph, ppw, -1)
+
+        freqs_cos = torch.cat([freqs_cos_f, freqs_cos_h, freqs_cos_w], dim=-1).reshape(1, ppf * pph * ppw, 1, -1)
+        freqs_sin = torch.cat([freqs_sin_f, freqs_sin_h, freqs_sin_w], dim=-1).reshape(1, ppf * pph * ppw, 1, -1)
+
+        return freqs_cos, freqs_sin
+
+
+class SanaModulatedNorm(nn.Module):
```
dg845, line 241:
```suggestion
# Copied from diffusers.models.transformers.sana_transformer.SanaModulatedNorm
class SanaModulatedNorm(nn.Module):
```
nit: if this meant to be the same as in the original Sana transformer (which I think is the case), adding a `# Copied from` statement will help ensure that the implementations are coupled.

src/diffusers/pipelines/sana/pipeline_sana_video.py
```diff
@@ -0,0 +1,1045 @@
...
+        """
+        self.vae.enable_tiling()
+
+    def disable_vae_tiling(self):
+        r"""
+        Disable tiled VAE decoding. If `enable_vae_tiling` was previously enabled, this method will go back to
+        computing decoding in one step.
+        """
+        self.vae.disable_tiling()
+
```
dg845:
```suggestion
```
My understanding is that calling the VAE's `enable_slicing`, etc. methods (`pipe.vae.enable_slicing()`) directly is now preferred over having `enable_slicing`, etc. methods in the pipeline itself. CC @yiyixuxu

src/diffusers/pipelines/sana/pipeline_sana_video.py
```diff
@@ -0,0 +1,1045 @@
...
+            int(height) // self.vae_scale_factor_spatial,
+            int(width) // self.vae_scale_factor_spatial,
+        )
+        if isinstance(generator, list) and len(generator) != batch_size:
+            raise ValueError(
+                f"You have passed a list of generators of length {len(generator)}, but requested an effective batch"
+                f" size of {batch_size}. Make sure the batch size matches the length of the generators."
+            )
+
+        latents = randn_tensor(shape, generator=generator, device=device, dtype=dtype)
```
dg845:
```suggestion
        if latents is None:
            latents = randn_tensor(shape, generator=generator, device=device, dtype=dtype)
        else:
            latents = latents.to(device=device, dtype=dtype)
```
To handle pre-generated latents supplied to `__call__`.

src/diffusers/models/transformers/transformer_sana_video.py
```diff
@@ -0,0 +1,752 @@
...
+        attention_head_dim: int = 112,
+        num_layers: int = 20,
+        num_cross_attention_heads: Optional[int] = 20,
+        cross_attention_head_dim: Optional[int] = 112,
+        cross_attention_dim: Optional[int] = 2240,
+        caption_channels: int = 2304,
+        mlp_ratio: float = 2.5,
+        dropout: float = 0.0,
+        attention_bias: bool = False,
+        sample_size: int = 32,
```
dg845: nit: would it make sense for `sample_size` to default to either `30` or `22` here? I see that `SanaVideoPipeline.__call__` will raise an error if `sample_size` is not `30` or `22`.

lawrence-cj:
```suggestion
        sample_size: int = 30,
```

src/diffusers/pipelines/sana/pipeline_sana_video.py
lawrence-cj: We may add this back if we really need it later. Will remove it first.

src/diffusers/models/transformers/transformer_sana_video.py
lawrence-cj: support it now: https://github.com/huggingface/diffusers/pull/12584/commits/fd5cff2e1c349c43d44e63705c3846b8d7425e2f

src/diffusers/video_processor.py
lawrence-cj: OK, I would let u guys help to finish this part. Thanks!!

lawrence-cj:
I have added markdown we need.

THANKS so much for your support! 
@yiyixuxu @dg845

lawrence-cj:
I also added two test cases for you for reference. Please feel free to modify them

https://github.com/huggingface/diffusers/pull/12584/commits/d98f93c1b4dadacb7a10690ad888570fced8dce5

tests/pipelines/sana/test_sana_video.py
```diff
@@ -0,0 +1,190 @@
...
+            dim_mult=[1, 1, 1, 1],
+            num_res_blocks=1,
+            temperal_downsample=[False, True, True],
+        )
+
+        torch.manual_seed(0)
+        scheduler = DPMSolverMultistepScheduler()
+        
+        text_encoder = AutoModel.from_pretrained("hf-internal-testing/tiny-random-gemma2")
+        tokenizer = AutoTokenizer.from_pretrained("hf-internal-testing/tiny-random-gemma2")
```
dg845: It looks like `hf-internal-testing/tiny-random-gemma2` doesn't currently exist on the HF Hub. The original Sana pipeline tests uses `hf-internal-testing/dummy-gemma` and the following code for fast testing:

https://github.com/huggingface/diffusers/blob/dcfb18a2d340d8e1f0ff001b06d2931ffa8648da/tests/pipelines/sana/test_sana.py#L105-L120

Would this work for Sana-Video as well?

lawrence-cj: I think so. Let me correct it.

lawrence-cj: fixed in new commit;

tests/pipelines/sana/test_sana_video.py
```diff
@@ -0,0 +1,204 @@
...
+            attention_head_dim=12,
+            num_layers=2,
+            num_cross_attention_heads=2,
+            cross_attention_head_dim=12,
+            cross_attention_dim=24,
+            caption_channels=32,
+            mlp_ratio=2.5,
+            dropout=0.0,
+            attention_bias=False,
+            sample_size=8,
```
dg845, line 102: Since the pipeline will only accept a transformer `sample_size` of `22` or `30`, setting `sample_size=8` here causes most of the pipeline tests to raise an exception. Would it be possible to create a version of `ASPECT_RATIO_480_BIN` for much smaller resolutions?

If not, I'm not sure what to do here, since running fast tests on e.g. $`832 \times 480`$ resolution video is likely too heavy for the CI. CC @sayakpaul

tests/pipelines/sana/test_sana_video.py
```diff
@@ -0,0 +1,204 @@
...
+            "negative_prompt": "blurry, low quality",
+            "generator": generator,
+            "num_inference_steps": 2,
+            "guidance_scale": 6.0,
+            "height": 16,
+            "width": 16,
+            "frames": 9,
+            "max_sequence_length": 16,
+            "output_type": "pt",
+            "complex_human_instruction": [],
```
lawrence-cj, line 135:
```suggestion
            "complex_human_instruction": [],
            "use_resolution_binning": False,
```
How about set `use_resolution_binning=False` during test. The error raise in pipe is checked when `use_resolution_binning=True`
@dg845

dg845: Sounds good, I missed that 😅 . We can test with `use_resolution_binning=True` in a slow test on a full-size video.

tests/pipelines/sana/test_sana_video.py
sayakpaul: Yeah I concur the thoughts. It's too heavy for the CI to run with that resolution.

dg845: Can be mitigated by setting `use_resolution_binning=False`; see https://github.com/huggingface/diffusers/pull/12584#discussion_r2493193479.

tests/pipelines/sana/test_sana_video.py
```diff
@@ -0,0 +1,204 @@
...
+        transformer = SanaVideoTransformer3DModel(
+            in_channels=16,
+            out_channels=16,
+            num_attention_heads=2,
+            attention_head_dim=12,
+            num_layers=2,
+            num_cross_attention_heads=2,
+            cross_attention_head_dim=12,
+            cross_attention_dim=24,
+            caption_channels=32,
```
dg845:
```suggestion
            caption_channels=8,
```
The `Gemma2Model` used for testing only outputs embeddings with a dimension of `8`, so set `caption_channels=8` to avoid a shape error.

tests/pipelines/sana/test_sana_video.py
```diff
@@ -0,0 +1,204 @@
...
+        return components
+
+    def get_dummy_inputs(self, device, seed=0):
+        if str(device).startswith("mps"):
+            generator = torch.manual_seed(seed)
+        else:
+            generator = torch.Generator(device=device).manual_seed(seed)
+        inputs = {
+            "prompt": "A cat playing with a ball",
+            "negative_prompt": "blurry, low quality",
```
dg845:
```suggestion
            "prompt": "",
            "negative_prompt": "",
```
The tokenizer defined in `hf-internal-testing/dummy-gemma` has a very small vocabulary so it cannot tokenize most non-empty prompts. The suggested change is also done in other Sana pipeline tests e.g.:

https://github.com/huggingface/diffusers/blob/dcfb18a2d340d8e1f0ff001b06d2931ffa8648da/tests/pipelines/sana/test_sana.py#L136-L138

tests/pipelines/sana/test_sana_video.py
```diff
@@ -0,0 +1,205 @@
...
+        else:
+            generator = torch.Generator(device=device).manual_seed(seed)
+        inputs = {
+            "prompt": "",
+            "negative_prompt": "",
+            "generator": generator,
+            "num_inference_steps": 2,
+            "guidance_scale": 6.0,
+            "height": 16,
+            "width": 16,
```
dg845:
```suggestion
            "height": 32,
            "width": 32,
```
Because the pipeline will check in `check_inputs` to see if the `height` and `width` are divisible by `32`.

tests/pipelines/sana/test_sana_video.py
```diff
@@ -0,0 +1,205 @@
...
+
+        components = self.get_dummy_components()
+        pipe = self.pipeline_class(**components)
+        pipe.to(device)
+        pipe.set_progress_bar_config(disable=None)
+
+        inputs = self.get_dummy_inputs(device)
+        video = pipe(**inputs).frames
+        generated_video = video[0]
+        self.assertEqual(generated_video.shape, (9, 3, 16, 16))
```
dg845:
```suggestion
        self.assertEqual(generated_video.shape, (9, 3, 32, 32))
```
Follow-up change to https://github.com/huggingface/diffusers/pull/12584#discussion_r2493433034

tests/pipelines/sana/test_sana_video.py
```diff
@@ -0,0 +1,205 @@
...
+            pipe_loaded.to(torch_device)
+            pipe_loaded.set_progress_bar_config(disable=None)
+
+        inputs = self.get_dummy_inputs(torch_device)
+        torch.manual_seed(0)
+        output_loaded = pipe_loaded(**inputs)[0]
+
+        max_diff = np.abs(output.detach().cpu().numpy() - output_loaded.detach().cpu().numpy()).max()
+        self.assertLess(max_diff, expected_max_difference)
+
```
dg845, line 185:
```suggestion

    # TODO(aryan): Create a dummy gemma model with smol vocab size
    @unittest.skip(
        "A very small vocab size is used for fast tests. So, any kind of prompt other than the empty default used in other tests will lead to a embedding lookup error. This test uses a long prompt that causes the error."
    )
    def test_inference_batch_consistent(self):
        pass

    @unittest.skip(
        "A very small vocab size is used for fast tests. So, any kind of prompt other than the empty default used in other tests will lead to a embedding lookup error. This test uses a long prompt that causes the error."
    )
    def test_inference_batch_single_identical(self):
        pass

    def test_float16_inference(self):
        # Requires higher tolerance as model seems very sensitive to dtype
        super().test_float16_inference(expected_max_diff=0.08)

    def test_save_load_float16(self):
        # Requires higher tolerance as model seems very sensitive to dtype
        super().test_save_load_float16(expected_max_diff=0.2)
```
Skip some tests that don't work due to the tokenizer vocab being too small. @lawrence-cj, I've loosened the tolerances on the `float16` tests in the suggestion, but should these tests actually be skipped? When I tried to generate an image from the full pipeline in fp16 I got a black video, so it seems like fp16 may not be supported by the pipeline.

src/diffusers/pipelines/sana/pipeline_sana_video.py
```diff
@@ -0,0 +1,1018 @@
...
+    else:
+        scheduler.set_timesteps(num_inference_steps, device=device, **kwargs)
+        timesteps = scheduler.timesteps
+    return timesteps, num_inference_steps
+
+
+class SanaVideoPipeline(DiffusionPipeline, SanaLoraLoaderMixin):
+    r"""
+    Pipeline for text-to-video generation using [Sana](https://huggingface.co/papers/2509.24695).
+
```
dg845:
```suggestion

```
nit: remove whitespace here as the CI does not like it

src/diffusers/video_processor.py
```diff
@@ -111,3 +112,65 @@ def postprocess_video(
...
+        r"""
+        Resizes and crops a tensor of videos to the specified dimensions.
+
+        Args:
+            samples (`torch.Tensor`):
+                A tensor of shape (N, C, T, H, W) where N is the batch size, C is the number of channels, T is the
```
dg845:
```suggestion
                A tensor of shape (N, C, T, H, W) where N is the batch size, C is the number of channels, T is the
```
nit: remove trailing whitespace here as the CI does not like it

lawrence-cj:
> Thanks for the follow up changes! I have made some suggestions that should help the Sana Video pipeline tests pass.
> 
> Sorry for all the small change requests, but could you also do the following?
> 
> 1. Can you run the following to make sure that the CI code quality check is green?
> 
> ```shell
> make style
> make quality
> make fix-copies
> ```
> 
> 2. Can you add the new Sana Video markdown docs to `docs/source/en/_toctree.yml`? For reference, here is how the Sana pipeline docs were added: https://github.com/huggingface/diffusers/blob/dcfb18a2d340d8e1f0ff001b06d2931ffa8648da/docs/source/en/_toctree.yml#L562-L563
>     This change will help the docs build correctly.

Done! Let's test it.

docs/source/en/_toctree.yml
```diff
@@ -373,6 +373,8 @@
         title: QwenImageTransformer2DModel
       - local: api/models/sana_transformer2d
         title: SanaTransformer2DModel
+      - local: api/models/sana_video_transformer3d
+        title: SanaVideoTransformer3DModel
```
dg845, line 377: This will cause an error when building the docs since the `api/models/sana_video_transformer3d` file doesn't currently exist. Could you add a markdown doc for the transformer as well? For reference, here is the documentation for `SanaTransformer2DModel`: https://github.com/huggingface/diffusers/blob/main/docs/source/en/api/models/sana_transformer2d.md

lawrence-cj: Done.

dg845:
@bot /style

dg845:
@lawrence-cj, thanks again for the PR! The CI errors are unrelated to the PR so merging.

lawrence-cj:
Thank you so much for your support! ❤️

Cc @dg845 @sayakpaul @yiyixuxu

sayakpaul:
Congratulations on the release!

vladmandic:
question: example includes `model_score = 30` and it simply appends it to prompt, but there is no explanation of it nor is there any reference to `model_score` anywhere in the code?

lawrence-cj:
@vladmandic Oh, it's a typo! It should be `motion_score`, which is related to how large the motion you want, the higher the score, the larger the motion. I'll correct it in my next PR. Thanks for noticing.

PR #12601, opened 2025-11-06
Fix Qwen Edit Plus modular for multi-image input

sayakpaul:
# What does this PR do?

QwenImage Edit Plus allows us to provide multiple images as inputs. Our previous support for the modular version of the model didn't allow for that, though. This PR changes that.

### DIFF
--- a/src/diffusers/models/transformers/transformer_prx.py
+++ b/src/diffusers/models/transformers/transformer_prx.py
@@ -16,7 +16,6 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 
 import torch
 from torch import nn
-from torch.nn.functional import fold, unfold
 
 from ...configuration_utils import ConfigMixin, register_to_config
 from ...utils import logging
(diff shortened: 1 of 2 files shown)


src/diffusers/modular_pipelines/qwenimage/before_denoise.py
```diff
@@ -640,6 +639,43 @@ def __call__(self, components: QwenImageModularPipeline, state: PipelineState) -
...
+
+        vae_scale_factor = components.vae_scale_factor
+        block_state.img_shapes = [
+            [
+                (1, block_state.height // vae_scale_factor // 2, block_state.width // vae_scale_factor // 2),
+                *[
+                    (1, vae_height // vae_scale_factor // 2, vae_width // vae_scale_factor // 2)
+                    for vae_width, vae_height in block_state.vae_image_sizes
+                ],
+            ]
```
sayakpaul, line 656: Main difference from the existing RoPE block for Edit.

src/diffusers/modular_pipelines/qwenimage/encoders.py
```diff
@@ -873,12 +881,18 @@ def __call__(self, components: QwenImageModularPipeline, state: PipelineState):
...
+            processed_images = []
+            vae_image_sizes = []
+            for img in block_state.vae_image:
+                width, height = img.size
+                vae_width, vae_height, _ = calculate_dimensions(self.vae_image_size, width / height)
+                vae_image_sizes.append((vae_width, vae_height))
+                processed_images.append(
+                    components.image_processor.preprocess(image=img, height=vae_height, width=vae_width)
+                )
+            block_state.processed_image = torch.stack(processed_images, dim=0).squeeze(1)
```
sayakpaul: Each input image can have a different resolution and that is why separate `preprocess` is needed (also following the original implementation). This came into fruition because we started testing with multiple input images.

src/diffusers/modular_pipelines/qwenimage/inputs.py
```diff
@@ -341,6 +346,9 @@ def __call__(self, components: QwenImageModularPipeline, state: PipelineState) -
 
             # 2. Patchify the image latent tensor
             image_latent_tensor = components.pachifier.pack_latents(image_latent_tensor)
+            if self.reshape_to_seq_dim:
+                channels = image_latent_tensor.shape[-1]
+                image_latent_tensor = image_latent_tensor.reshape(1, -1, channels)
```
sayakpaul: Qwen Image Edit Plus specific.

sayakpaul:
@asomoza could you give this your eyes and test a bit? I have eliminated the need to pass `vae_image_sizes`.

asomoza:
they look good to me now, it all works with mellon too:

```
add sunglasses to the dog
```

|image one|result|
|---|---|
|dog_plushie (1)|output (5)|

```
make the subject in image one to look like the image two
```
|image one|image two|result|
|---|---|---|
|yarn-art-pikachu (1)|dog_plushie (1)|output (6)|

```
take subject in image one and the subject in image two and put them together in the background of image three
```

|image one|image two|image three|result|
|---|---|---|---|
|yarn-art-pikachu (1)|dog_plushie (1)|normal_output (1)|output (10)|

PR #12615, opened 2025-11-08
cross attn module for ltx

samuelt0:
# What does this PR do?
Adds the cross attention module to the LTX Attention Class

Fixes # (issue)
Mirrors the previous Attention class and allows us to distinguish between attention instances

### DIFF
--- a/src/diffusers/models/transformers/transformer_ltx.py
+++ b/src/diffusers/models/transformers/transformer_ltx.py
@@ -142,6 +142,7 @@ class LTXAttention(torch.nn.Module, AttentionModuleMixin):
         self.dropout = dropout
         self.out_dim = query_dim
         self.heads = heads
+        self.is_cross_attention: Optional[bool] = None
 
         norm_eps = 1e-5
         norm_elementwise_affine = True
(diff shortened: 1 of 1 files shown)


samuelt0:
@sayakpaul

src/diffusers/models/transformers/transformer_ltx.py
```diff
@@ -142,6 +142,7 @@ def __init__(
         self.dropout = dropout
         self.out_dim = query_dim
         self.heads = heads
+        self.is_cross_attention: Optional[bool] = None
```
sayakpaul, line 145: I think we can set it to `True` / `False` based on the value of `cross_attention_dim` no?

samuelt0: `self.cross_attention_dim` is always 4096 for attn1 and attn2 though.

`self.cross_attention_dim` either gets set to `cross_attention_dim` or `query_dim` which are both 4096 in my runs

sayakpaul: `if cross_attention_dim is None: self.is_cross_attention = False` -- what's up with this?

PR #12619, opened 2025-11-10
Deprecate `upcast_vae` in SDXL based pipelines

DN6:
# What does this PR do?
Removed the type cast in the AutoencoderKL Decoder in this [PR](https://github.com/huggingface/diffusers/pull/12512) in order to address graph breaks. The PR breaks some Mellon Nodes and default SDXL inference. The problem is that the casting is used to address an issue in the base SDXL VAE, which overflows when running in FP16.  

The overflow issue has been fixed for a while and almost all finetunes use the fixed VAE, so this logic that conditionally upcasts the upsample layers of the VAE is no longer needed. The memory savings from this selective layer casting in also quite minimal, so we can safely upcast all layers in the VAE if needed.   

This PR
1. Deprecates the `upcast_vae` method in the SDXL based pipelines that selectively upcasts only the upsample layers of the decoder in favour of upcasting the entire VAE if upcasting is required. 

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion_upscale.py
+++ b/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion_upscale.py
@@ -525,6 +525,11 @@ class StableDiffusionUpscalePipeline(
         return latents
 
     def upcast_vae(self):
+        deprecate(
+            "upcast_vae",
+            "1.0.0",
+            "`upcast_vae` is deprecated. Please use `pipe.vae.to(torch.float32)`. For more details, please refer to: https://github.com/huggingface/diffusers/pull/12619#issue-3606633695.",
+        )
         self.vae.to(dtype=torch.float32)
 
     @torch.no_grad()
--- a/src/diffusers/modular_pipelines/stable_diffusion_xl/decoders.py
+++ b/src/diffusers/modular_pipelines/stable_diffusion_xl/decoders.py
@@ -21,7 +21,7 @@ import torch
 from ...configuration_utils import FrozenDict
 from ...image_processor import VaeImageProcessor
 from ...models import AutoencoderKL
-from ...utils import logging
+from ...utils import deprecate, logging
 from ..modular_pipeline import (
     ModularPipelineBlocks,
     PipelineState,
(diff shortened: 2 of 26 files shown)


DN6:
@bot /style

DN6:
Gentle ping @yiyixuxu

examples/community/lpw_stable_diffusion_xl.py
```diff
@@ -1328,18 +1327,7 @@ def _get_add_time_ids(self, original_size, crops_coords_top_left, target_size, d
...
-            self.vae.decoder.mid_block.attentions[0].processor,
-            (AttnProcessor2_0, XFormersAttnProcessor),
-        )
-        # if xformers or torch_2_0 is used attention block does not need
-        # to be in float32 which can save lots of memory
-        if use_torch_2_0_or_xformers:
-            self.vae.post_quant_conv.to(dtype)
-            self.vae.decoder.conv_in.to(dtype)
-            self.vae.decoder.mid_block.to(dtype)
+        deprecate("`upcast_vae` is deprecated")
```
sayakpaul: We should also guide the users about what they should do instead of using `upcast_vae()`. We should also set the version number of thel lib after which this should be removed.

sayakpaul (commented):
Just one question. We should also add a test so that we can remove the `upcast_vae()` method after the depcrecation cycle is complete.

examples/community/lpw_stable_diffusion_xl.py
DN6: Hmm I assumed that no one would be using this method standalone. But looks like there are a few instances where this method is called directly.
https://github.com/search?q=%22pipe.upcast_vae()%22+language:Python&type=code 

Will update the deprecation message and handle the casting inside the method itself for backwards compatiblity.

jiqing-feng:
Please let me know if you have any progress or anything I can help. Thanks!

examples/community/lpw_stable_diffusion_xl.py
```diff
@@ -1328,18 +1327,8 @@ def _get_add_time_ids(self, original_size, crops_coords_top_left, target_size, d
 
     # Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion_upscale.StableDiffusionUpscalePipeline.upcast_vae
     def upcast_vae(self):
-        dtype = self.vae.dtype
+        deprecate("upcast_vae", "1.0.0", "`upcast_vae` is deprecated. Please use `pipe.vae.to(torch.float32)`")
```
sayakpaul, line 1330: (nit): I think this warrants a better message. We could link to the reasoning being provided in the OP of this PR:

```suggestion
        deprecate("upcast_vae", "1.0.0", "`upcast_vae` is deprecated. Please use `pipe.vae.to(torch.float32)`. For more details, please refer to: https://github.com/huggingface/diffusers/pull/12619#issue-3606633695.")
```

src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion_upscale.py
```diff
@@ -529,21 +525,7 @@ def prepare_latents(self, batch_size, num_channels_latents, height, width, dtype
         return latents
 
     def upcast_vae(self):
-        dtype = self.vae.dtype
```
sayakpaul, line 532: Hmm, should the deprecation be not called from here, too? I am seeing some methods where the deprecation call exists and some where it doesn't (for example, this one). Is this expected?

sayakpaul (approved):
I left two questions. Once resolved, we should merge.

PR #12625, opened 2025-11-10
Set prompt2 & prompt3 default to None as indicated by parameter args

DavidCarlyn:
# What does this PR do?

Sets the default parameters of `StableDiffusion3Pipeline.encode_prompt` for `prompt2` & `prompt3` to `None`. Otherwise, if the `encode_prompt` is directly used, then an error for require arguments will be thrown.

The parameter documentation indicates that they should handled being `None` since they will default to `prompt`:

```
prompt (`str` or `List[str]`, *optional*):
      prompt to be encoded
prompt_2 (`str` or `List[str]`, *optional*):
       The prompt or prompts to be sent to the `tokenizer_2` and `text_encoder_2`. If not defined, `prompt` is
       used in all text-encoders
prompt_3 (`str` or `List[str]`, *optional*):
       The prompt or prompts to be sent to the `tokenizer_3` and `text_encoder_3`. If not defined, `prompt` is
       used in all text-encoders
```

Also confirmed by L429-L434:

```
if prompt_embeds is None:
    prompt_2 = prompt_2 or prompt
    prompt_2 = [prompt_2] if isinstance(prompt_2, str) else prompt_2

    prompt_3 = prompt_3 or prompt
    prompt_3 = [prompt_3] if isinstance(prompt_3, str) else prompt_3
```


yiyixuxu:
can you run `make fix-copies`?

PR #12627, opened 2025-11-11
Fix: allow Flux Fill pipeline to skip image preprocessing when latents are provided

Xingfu-Yi:
# Fix: allow Flux Fill pipeline to skip image preprocessing when latents are provided

## What does this PR do?

Fixes a **critical bug** in the Flux Fill pipeline where `image` was implicitly assumed to exist even when `latents` and `masked_image_latents` were directly provided as inputs.

When both `latents` and `masked_image_latents` are passed, supplying an `image` is not only unnecessary but also **disallowed** by input validation. However, the current implementation still tries to preprocess `image`, which results in a `NoneType` error when `image` is `None`.

This PR adds a minimal and safe fix:
If `image` is `None`, the code now explicitly sets `init_image = None` and skips unnecessary preprocessing. This ensures that downstream logic, such as `self.prepare_latents()`, correctly bypasses latent initialization and uses the user-provided latents as intended.

## Changes

* Added a conditional check to skip `image` preprocessing when `image` is `None`
* Prevents runtime errors when using latent-only initialization
* Maintains compatibility and minimal code change within the existing pipeline logic

## Why is this important?

Without this fix, initializing the Flux Fill pipeline with precomputed latents is impossible due to a `NoneType` dereference during image preprocessing.
This breaks valid workflows that rely on directly supplying `latents` and `masked_image_latents` (for example, when reusing intermediate latent states for iterative or compositional generation).

## Fixes

Fixes # (issue number if applicable)

## Before submitting

* [x] This PR fixes a runtime bug (not a docs or typo fix).
* [x] I have read the [[contributor guideline](https://github.com/huggingface/diffusers/blob/main/CONTRIBUTING.md)](https://github.com/huggingface/diffusers/blob/main/CONTRIBUTING.md).
* [ ] I have added or verified tests for the affected path.
* [ ] I have ensured no behavioral changes when `image` is provided normally.


iwr-redmond:
Pinging to remove stale.

PR #12628, opened 2025-11-11
[modular] add a check 

yiyixuxu:


src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -861,6 +861,10 @@ def __init__(self):
             else:
                 sub_blocks[block_name] = block
         self.sub_blocks = sub_blocks
+        if not len(self.block_names) == len(self.block_classes):
```
yiyixuxu, line 864: adding a check here, we had a similar check to autopipeline blocks, which I got this error message a couple times already so it's helpful

DN6 (approved):

Issue #12635, opened 2025-11-11
The Diffusers MVP 🚀


sayakpaul:
Hello folks 👋

## What❓

We’re excited to bring a contributor-focused program to you! In this program, we want to work and collaborate with serious contributors to Diffusers and reward them for their time and energy spent with us. Keep reading this thread if that sounds interesting to you.

To ease the process, we have compiled a list of issues (more below) where we believe we require the most assistance. These issues also require contributors to consider solid and robust API design that is sustainable and maintainable in the long term.

To be considered for this program, we kindly request that you select only one issue at a time, engage with us by proposing potential design RFCs, and proceed from there. For now, we have a one-contributor-per-one-issue policy, but we can always make exceptions.

## Issue list 🐞

- https://github.com/huggingface/diffusers/issues/12216
- https://github.com/huggingface/diffusers/issues/12550
- [https://github.com/huggingface/diffusers/issues/12075](https://github.com/huggingface/diffusers/issues/12075) (@sambhavnoobcoder: https://github.com/huggingface/diffusers/pull/12643)
- [https://github.com/huggingface/diffusers/issues/12344](https://github.com/huggingface/diffusers/issues/12344) (@kashif: https://github.com/huggingface/diffusers/pull/12702)
- [https://github.com/huggingface/diffusers/issues/12294](https://github.com/huggingface/diffusers/issues/12294) (@cdutr: https://github.com/huggingface/diffusers/pull/12655)
- [https://github.com/huggingface/diffusers/issues/12458](https://github.com/huggingface/diffusers/issues/12458) (@akshan-main: https://github.com/huggingface/diffusers/pull/12698)
- [https://github.com/huggingface/diffusers/issues/12568](https://github.com/huggingface/diffusers/issues/12568) (@Ratish1: https://github.com/huggingface/diffusers/pull/12595)
- [https://github.com/huggingface/diffusers/issues/12569](https://github.com/huggingface/diffusers/issues/12569) (@toilaluan: https://github.com/huggingface/diffusers/pull/12648)
- [https://github.com/huggingface/diffusers/issues/12589](https://github.com/huggingface/diffusers/issues/12589) (@LawJarp-A: https://github.com/huggingface/diffusers/pull/12652)
- https://github.com/huggingface/diffusers/pull/12744 (@AlanPonnachan)
[...]

PR #12648, opened 2025-11-13
[Feat] TaylorSeer Cache

toilaluan:
What does this PR do?

Adding [TaylorSeer Caching](https://arxiv.org/pdf/2503.06923) method to accelerate inference speed mentioned in https://github.com/huggingface/diffusers/issues/12569

Author's codebase: https://github.com/Shenyi-Z/TaylorSeer

This PR structure will heavily mimic FasterCache (https://github.com/huggingface/diffusers/pull/10163/files) behaviour
I prioritze to make it work on image model pipelines (Flux, Qwen Image) for ease of evaluation

### Expected Output

4->5x speeding up by these settings while keep output images are qualified

image

### State Design

Core of this algorithm is about predict features of step `t` by using real computed features from previous step using Taylor Expansion Approximation.
We design a State class, include `predict` & `update` method and `taylor_factors: Tensor` to maintain iteration information. Each feature tensor will be bounded to a state instance (in double stream attention class in Flux & QwenImage, output of this module is image_features & txt_features, we will create 2 state instances for them)

- `update` method will be called from real compute timestep and update `taylor_factors` using math formular referenced to original implementation
- `predict` method will be called to predict feature from current `taylor_factors` using math formular referenced to original implementation

Issue #12569, opened 2025-11-01
[feature] implement TaylorSeer


sayakpaul:
Currently, we support PAB and FasterCache. More details: https://huggingface.co/docs/diffusers/main/en/optimization/cache. 

It'd be cool to support [Accelerating Diffusion Models with TaylorSeers](https://arxiv.org/abs/2503.06923) as it's quite neat and significantly better than others. Code is here:
https://github.com/Shenyi-Z/TaylorSeer

Before proceeding with a PR, please propose a design first in this thread. We can take it from there.

ghost:
Hi @sayakpaul! I am lookin forward to it. But I've a few questions to understand the project:
1. Which pipelines should be prioritized? (SDXL, SD 1.5, Flux or all ?)
2. Should this follow the same pattern as PAB/FasterCache ?
3. Roughly how many files/pipelines would need modification? Is this a small, med or large contribution?
4. Also, can TayloSeer work alongside existing optimizations like PAB, FasterCache? 

Based on your answers, I'll propose a design. Thanks!

sayakpaul:
1. Very good question. In general, our methods tend to be very pipeline-agnostic. For example, `FasterCache` should work for most models. But we can prioritize the pipelines you mentioned (SDXL, Flux, QwenImage, etc.).
2. Yes, following a uniform structure is always helpful.
3. It's a significant contribution. You can get a sense of the needed modifications from this PR: https://github.com/huggingface/diffusers/issues/10163. 
4. That's an interesting question. I think we can tackle that later. First priority, IMO, should be to have the method natively intrgrated.

ghost:
Thanks @sayakpaul! This sounds like a great project. Let me review PR #10163 and the TaylorSeer code to understand the project. I'll get back to you in a few days on whether I can commit to this. Is that okay?

sayakpaul:
Sounds perfect!

toilaluan:
Hi @sayakpaul @Aznix07 
I created a PR for this feature
I have written proposed design in the PR description as well as init code diff can desribe a bit
I expect to have it functionally done in 2 to 3 days for image models

sayakpaul:
Could you check with @Aznix07 as well to see if they would be interested in collaborating? This is so everyone gets a fair chance to contribute something.

toilaluan:
Yeah, since this Pr will be huge, include adding for video models, let's collab on it @Aznix07 
The result from paper is very promising and hope we can onboard it soon to users
[...]

PR #12652, opened 2025-11-13
Implement TeaCache 

### LINKED ISSUE #12589: [feature] implement TeaCache
Currently, we support PAB and FasterCache. More details: https://huggingface.co/docs/diffusers/main/en/optimization/cache. 

It'd be cool to support [Timestep Embedding Tells: It's Time to Cache for Video Diffusion Model](https://liewfeng.github.io/TeaCache/) as it's quite popular in the community. Code is here:
https://github.com/ali-vilab/TeaCache

Before proceeding with a PR, please propose a design first in this thread. We can take it from there.

LawJarp-A:
# What does this PR do?
### What is TeaCache?

[TeaCache](https://huggingface.co/papers/2411.19108) (Timestep Embedding Aware Cache) is a training-free caching technique that speeds up diffusion model inference by **1.5x-2.6x** by reusing transformer block computations when consecutive timestep embeddings are similar.

### Architecture & Design

TeaCache uses a `ModelHook` to intercept transformer forward passes without modifying model code. The algorithm:

1. Extracts modulated input from first transformer block (after norm1 + timestep embedding)
2. Computes relative L1 distance vs previous timestep
3. Applies model-specific polynomial rescaling: `c[0]*x^4 + c[1]*x^3 + c[2]*x^2 + c[3]*x + c[4]`
4. Accumulates rescaled distance across timesteps
5. If accumulated < threshold → Reuses cached residual (FAST)
6. If accumulated >= threshold → Full transformer pass (SLOW, update cache)

**Key Design Features:**

- **Hook-based**: Integrates with `HookRegistry` and `CacheMixin` for lifecycle management
- **State Isolation**: `StateManager` with context-aware state for CFG conditional/unconditional branches
- **Model Auto-Detection**: Detects model type from class name and config path (specific variants checked first)
- **Boundary Guarantee**: First and last timesteps always computed fully for quality
- **Specialized Strategies**: Dual residual caching (CogVideoX), per-sequence-length caching (Lumina2)

### Supported Models

| Model | Coefficients | Status |
|-------|-------------|--------|
| FLUX | Auto-detected | Tested |
| FLUX-Kontext | Auto-detected | Ready |
| Mochi | Auto-detected | Ready |
| Lumina2 | Auto-detected | Ready |
| CogVideoX (2b/5b/1.5-5B) | Auto-detected | Ready |

All models support automatic coefficient detection based on model class name and config path. Custom coefficients can also be provided via `TeaCacheConfig`.

---

### Benchmark Results (FLUX.1-dev)

| Threshold | Time | Speedup |
|-----------|------|---------|
| Baseline | 9.26s | 1.00x |
| 0.2 | 6.85s | **1.35x** |
| 0.4 | 5.24s | **1.77x** |
| 0.6 | 4.64s | **2.00x** |
| 0.8 | 4.18s | **2.22x** |

---

### Benchmark Results (Lumina2)

| Threshold | Time | Speedup |
|-----------|------|---------|
| Baseline | 3.45s | 1.00x |
| 0.2 | 3.07s | **1.12x** |
| 0.4 | 2.27s | **1.52x** |
| 0.6 | 1.84s | **1.88x** |

---

### Benchmark Results (CogVideoX-2b)

| Threshold | Time | Speedup |
|-----------|------|---------|
| Baseline | 26.27s | 1.00x |
| 0.3 | 23.97s | **1.10x** |
| 0.5 | 22.57s | **1.16x** |
| 0.7 | 19.31s | **1.36x** |
| 0.9 | 17.38s | **1.51x** |

---

### Benchmark Results (Mochi)

| Threshold | Time | Speedup |
|-----------|------|---------|
| Baseline | 7.71s | 1.00x |
| 0.05 | 6.27s | **1.23x** |
| 0.06 | 6.03s | **1.28x** |
| 0.08 | 5.73s | **1.35x** |
| 0.10 | 5.41s | **1.42x** |

---

**Test Hardware**: NVIDIA h100
**Framework**: Diffusers with TeaCache hooks  
**All tests**: Same seed (42) for reproducibility

### Usage

```python
from diffusers.hooks import TeaCacheConfig

pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-schnell", torch_dtype=torch.bfloat16)
pipe.to("cuda")

# Enable TeaCache (1.75x speedup with 0.4 threshold)
config = TeaCacheConfig(rel_l1_thresh=0.4)
pipe.transformer.enable_cache(config)

image = pipe("A dragon on a crystal mountain", num_inference_steps=20).images[0]

pipe.transformer.disable_cache()
```

### Configuration Options

The `TeaCacheConfig` supports the following parameters:

- `rel_l1_thresh` (float, default=0.2): Threshold for accumulated relative L1 distance. Recommended values: 0.25 for ~1.5x speedup, 0.4 for ~1.8x, 0.6 for ~2.0x. Mochi models require lower thresholds (0.06-0.09).
- `coefficients` (List[float], optional): Polynomial coefficients for rescaling L1 distance. Auto-detected based on model type if not provided.
- `num_inference_steps` (int, optional): Total inference steps. Ensures first/last timesteps are always computed. Auto-detected if not provided.
- `num_inference_steps_callback` (Callable[[], int], optional): Callback returning total inference steps. Alternative to `num_inference_steps`.
- `current_timestep_callback` (Callable[[], int], optional): Callback returning current timestep. Used for debugging/statistics.

### Files Changed
- `src/diffusers/hooks/teacache.py` - Core implementation with model-specific forward functions
- `src/diffusers/models/cache_utils.py` - CacheMixin integration
- `src/diffusers/hooks/__init__.py` - Export TeaCacheConfig and apply_teacache
- `tests/hooks/test_teacache.py` - Comprehensive unit tests

---

Fixes # (issue)
https://github.com/huggingface/diffusers/issues/12589
https://github.com/huggingface/diffusers/issues/12635

## Before submitting
- [X] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? [Please add a link to it if that's the case.](https://github.com/huggingface/diffusers/issues/12589)

## Who can review?
@sayakpaul @yiyixuxu @DN6


LawJarp-A:
Work done
- [X] Implement teacache for **_FLUX_** architecture using **hooks** (only flux for now)
- [X] add logging
- [X] add compatible tests

Waiting for feedback and review :)
cc: @dhruvrnaik @sayakpaul @yiyixuxu

LawJarp-A:
Hi @sayakpaul @dhruvrnaik any updates?

sayakpaul:
@LawJarp-A sorry about the delay on our end. @DN6 will review it soon.

DN6:
Hi @LawJarp-A I think we would need TeaCache to be implemented in a model agnostic way in order to merge the PR. The First Block Cache implementation is a good reference for this.

LawJarp-A:
> Hi @LawJarp-A I think we would need TeaCache to be implemented in a model agnostic way in order to merge the PR. The First Block Cache implementation is a good reference for this.

Yep @DN6 , I agree, I wanted to first implement it just for a single model and get feedback on that before I work on Model agnostic full implementation. I'm sort of working on it, didn't push it yet. I'll take a look at First block cache for reference as well. 
On the same note, lemme know if there is anything to add to the current implementation

src/diffusers/hooks/flux_teacache.py
```diff
@@ -0,0 +1,474 @@
...
+            )
+        if self.rel_l1_thresh <= 0:
+            raise ValueError(
+                f"rel_l1_thresh must be positive, got {self.rel_l1_thresh}. "
+                f"Based on the TeaCache paper, values between 0.1 and 0.3 work best. "
+                f"Try 0.25 for 1.5x speedup or 0.6 for 2x speedup."
+            )
+        if self.rel_l1_thresh < 0.05:
+            import warnings
+            warnings.warn(
```
DN6: Use `logger.warning`

LawJarp-A:
@DN6 updated it in a more model agnostic way. 
Requesting review and feedback

LawJarp-A:
Added multi model support, testing it thoroughly though.

LawJarp-A:
Hi @DN6 @sayakpaul 
Two questions, I'm almost done testing, I'll update the PR with more descriptive results and changes. And do final cleanup/merging etc
1. Any tests I should write and anything I can refer to for the same?
2. Added support for other models, I'll add pictures comparison with speedup and threshold to the PR as well?

In the meantime any feedback would be appreciated

sayakpaul:
Thanks @LawJarp-A!

> Any tests I should write and anything I can refer to for the same?

You can refer to https://github.com/huggingface/diffusers/issues/12569 for testing

> Added support for other models, I'll add pictures comparison with speedup and threshold to the PR as well?

Yes, I think that is informative for users.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1029 @@
...
+
+from ..utils import logging
+from .hooks import BaseState, HookRegistry, ModelHook, StateManager
+
+
+logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
+
+_TEACACHE_HOOK = "teacache"
+
+# Model-specific polynomial coefficients from TeaCache paper/reference implementations
```
sayakpaul: Do we know if these are just model-agnostic or there's something other dependencies as well (for example `num_inference_steps`, `guidance_scale`, etc.)? 

Also, can we add a calibration step similar to https://github.com/huggingface/diffusers/pull/12648 so that users can log these coefficients for other models?

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1029 @@
...
+    ],  # Default to 5b variant
+    # CogVideoX model variants with specific coefficients
+    "CogVideoX-2b": [-3.10658903e01, 2.54732368e01, -5.92380459e00, 1.75769064e00, -3.61568434e-03],
+    "CogVideoX-5b": [-1.53880483e03, 8.43202495e02, -1.34363087e02, 7.97131516e00, -5.23162339e-02],
+    "CogVideoX1.5-5B": [2.50210439e02, -1.65061612e02, 3.57804877e01, -7.81551492e-01, 3.58559703e-02],
+    "CogVideoX1.5-5B-I2V": [1.22842302e02, -1.04088754e02, 2.62981677e01, -3.06001e-01, 3.71213220e-02],
+}
+
+
+def _flux_modulated_input_extractor(module, hidden_states, timestep_emb):
```
sayakpaul: So, each model has a different feature extractor?

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1029 @@
...
+    return module.layers[0].norm1(hidden_states, timestep_emb)[0]
+
+
+def _cogvideox_modulated_input_extractor(module, hidden_states, timestep_emb):
+    """Extract modulated input for CogVideoX models."""
+    # CogVideoX uses the timestep embedding directly, not from a block
+    return timestep_emb
+
+
+def _auto_detect_extractor(module):
```
sayakpaul: How about maintaining a map and simplifying this a bit instead of if/else?

```py
_EXTRACTOR_REGISTRY = {
    "Flux": _flux_modulated_input_extractor,
    "Mochi": _mochi_modulated_input_extractor,
    "Lumina2": _lumina2_modulated_input_extractor,
    "CogVideoX": _cogvideox_modulated_input_extractor,
}

def _auto_detect_extractor(module):
    """Auto-detect and return appropriate extractor based on model type."""
    module_class_name = module.__class__.__name__

    for key, extractor in _EXTRACTOR_REGISTRY.items():
        if key in module_class_name:
            return extractor

    # Default to FLUX for backward compatibility
    logger.warning(
        f"TeaCache: Unknown model type {module_class_name}, defaulting to FLUX extractor. "
        f"Results may be incorrect. Please provide extract_modulated_input_fn explicitly."
    )
    return _flux_modulated_input_extractor
```

Also, WDYT about raising an error when we don't have the model implemented?

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1029 @@
...
+
+            warnings.warn(
+                f"rel_l1_thresh={self.rel_l1_thresh} is very low and may result in minimal caching. "
+                f"Consider using values between 0.1 and 0.3 for optimal performance.",
+                UserWarning,
+            )
+        if self.rel_l1_thresh > 1.0:
+            import warnings
+
+            warnings.warn(
```
sayakpaul: We use `logger.warning`

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1029 @@
...
+                f"Consider using values between 0.1 and 0.6 for better quality-speed tradeoff.",
+                UserWarning,
+            )
+
+        # Set default coefficients if not provided (will be auto-detected in hook initialization)
+        if self.coefficients is None:
+            # Default to FLUX coefficients (will be overridden by auto-detection if model is recognized)
+            self.coefficients = _MODEL_COEFFICIENTS.get(
+                "Flux", [4.98651651e02, -2.83781631e02, 5.58554382e01, -3.82021401e00, 2.64230861e-01]
+            )
```
sayakpaul: Let's make it more explicit by erroring out no? Flux-based values might not transfer to Qwen, for example, and it can go unnoticed silently.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1029 @@
...
+        # Auto-set coefficients from registry if not explicitly provided
+        if self.config.coefficients is None or (
+            self.model_type in _MODEL_COEFFICIENTS and self.config.coefficients == _MODEL_COEFFICIENTS.get("Flux")
+        ):
+            if self.model_type in _MODEL_COEFFICIENTS:
+                self.config.coefficients = _MODEL_COEFFICIENTS[self.model_type]
+                logger.info(f"TeaCache: Auto-detected {self.model_type} coefficients")
+
+        # Initialize rescale function with final coefficients
+        self.rescale_func = np.poly1d(self.config.coefficients)
```
sayakpaul: Would other rescaling functions work here? If so, could we make it configurable?

sayakpaul (commented):
Some initial feedback. Most important question is it seems like we need to craft different logic based on different model? Can we not keep it model agnostic?

src/diffusers/hooks/teacache.py
LawJarp-A: Yes, had not made changes to these bits when I added new models.
Adding them, and making explicit errors for model map and coefficient as well

src/diffusers/hooks/teacache.py
LawJarp-A: Yep, Teacache works on timestep embeddings which different for each model. 
FLUX/Mochi use norm1(hidden_states, timestep_emb) from the first transformer block, Lumina2 uses layers[0].norm1() after preprocessing, and CogVideoX uses the timestep embedding directly etc

LawJarp-A:
I am trying to think if ways we can avoid having the forward model for each model now. Initially that seemed like th ebe

> Some initial feedback. Most important question is it seems like we need to craft different logic based on different model? Can we not keep it model agnostic?


t was fine when I wrote for flux, but lumina needed multi stage preprocessing. 
I am trying to think how to , but keeping a generic forward might not work very well :/
Firstcache, FirstBlock all work block level, but TeaCache is more model level. 
Defo open to ideas :)

LawJarp-A:
@sayakpaul 
Added flux image example in the PR description. 
Tested it with Lumina, CogVideoX as well
Could not test with Mochi because of GPU constraints. I can try with cpu offloading maybe

LawJarp-A:
@sayakpaul @DN6 I got the core logic working, and tested it for model my GPU can handle
Right now I have gone for a simple monolithic method, each of the models forward handlers, extractors all in one file. I tried to abstract it as much, but since TeaCache works on model level, rather than blocks (like most of the caches right now, taylor, firstblock etc). It's proven a bit difficult to make it model agnostic. 

The current implementation puts all model handlers in a single `teacache.py` file. This works but has scaling concerns:
I was thinking, since we **have** to add model specific functions anyway, make them a bit modular deisgn-wise. 

**Potential refactor: Registry + Handler pattern**

```
diffusers/hooks/
├── teacache/
│   ├── __init__.py           # Public API
│   ├── config.py             # TeaCacheConfig
│   ├── hook.py               # TeaCacheHook (core logic)
│   ├── registry.py           # Handler registry
│   └── handlers/
│       ├── __init__.py       # Auto-imports all handlers
│       ├── base.py           # BaseTeaCacheHandler ABC
│       ├── flux.py
│       ├── mochi.py
│       ├── lumina2.py
│       └── cogvideox.py
```

Each handler self-registers and encapsulates its logic:

```python
# handlers/flux.py
from .base import BaseTeaCacheHandler
from ..registry import register_handler

@register_handler("Flux", "FluxKontext")
class FluxHandler(BaseTeaCacheHandler):
    coefficients = [4.98651651e02, -2.83781631e02, ...]
    
    def extract_modulated_input(self, module, hidden_states, temb):
        return module.transformer_blocks[0].norm1(hidden_states, emb=temb)[0]
    
    def handle_forward(self, module, *args, **kwargs):
        # FLUX-specific forward with ControlNet, LORA, etc.
        ...
```

```python
# registry.py
_HANDLER_REGISTRY = {}

def register_handler(*model_names):
    def decorator(cls):
        for name in model_names:
            _HANDLER_REGISTRY[name] = cls
        return cls
    return decorator

def get_handler(module) -> BaseTeaCacheHandler:
    for name, handler_cls in _HANDLER_REGISTRY.items():
        if name in module.__class__.__name__:
            return handler_cls()
    raise ValueError(f"No TeaCache handler for {module.__class__.__name__}")
```

This is similar to how attention processors and schedulers are organized. Happy to refactor if you think it's worth it, or we can keep it simple like now. Since this has proven a bit more of a challenge to integrate than I thought xD would be happy to know if you guys have some ideas.

LawJarp-A:
Hey @DN6 @sayakpaul , any updates :)

src/diffusers/models/cache_utils.py
```diff
@@ -141,3 +153,23 @@ def cache_context(self, name: str):
         yield
 
         registry._set_context(None)
+
+    def enable_teacache(self, rel_l1_thresh: float = 0.2, num_inference_steps: int = None, **kwargs):
```
DN6: Cacheing should only be enabled through `enable_cache` and passing the relevant config. Cache specific enabling is not supported.

LawJarp-A:
@sayakpaul @DN6 checking in again :)

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1021 @@
...
+        ```python
+        from diffusers import FluxPipeline
+        from diffusers.hooks import TeaCacheConfig
+
+        # Load FLUX pipeline
+        pipe = FluxPipeline.from_pretrained("black-forest-labs/FLUX.1-schnell", torch_dtype=torch.bfloat16)
+        pipe.to("cuda")
+
+        # Enable TeaCache with auto-detection (1.5x speedup)
+        pipe.transformer.enable_teacache(rel_l1_thresh=0.2)
```
DN6: Should be
```python
        pipe.transformer.enable_cache(...)
```

We don't enable specific cacheing methods directly

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1021 @@
...
+            elif self.config.num_inference_steps_callback is not None:
+                state.num_steps = self.config.num_inference_steps_callback()
+            elif hasattr(module, "num_steps"):
+                state.num_steps = module.num_steps
+            
+            if state.num_steps > 0:
+                logger.info(f"TeaCache: Using {state.num_steps} inference steps")
+
+    def initialize_hook(self, module):
+        self.state_manager.set_context("teacache")
```
DN6: Cache context is typically set in the denoising loop? I think in this case, both conditional and unconditional branches would write to the same cache state when using CFG.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1021 @@
...
+    "CogVideoX-2b": [-3.10658903e01, 2.54732368e01, -5.92380459e00, 1.75769064e00, -3.61568434e-03],
+    "CogVideoX-5b": [-1.53880483e03, 8.43202495e02, -1.34363087e02, 7.97131516e00, -5.23162339e-02],
+    "CogVideoX1.5-5B": [2.50210439e02, -1.65061612e02, 3.57804877e01, -7.81551492e-01, 3.58559703e-02],
+    "CogVideoX1.5-5B-I2V": [1.22842302e02, -1.04088754e02, 2.62981677e01, -3.06001e-01, 3.71213220e-02],
+}
+
+
+def _flux_modulated_input_extractor(module, hidden_states, timestep_emb):
+    """Extract modulated input for FLUX models."""
+    return module.transformer_blocks[0].norm1(hidden_states, emb=timestep_emb)[0]
```
DN6: I think these extractor functions can be folded into the adapter functions of each model. They're thin wrappers around a single line of code.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,1021 @@
...
+        # This ensures rescale_func is always valid, even if initialize_hook isn't called (e.g., in tests)
+        default_coeffs = config.coefficients if config.coefficients else _MODEL_COEFFICIENTS["Flux"]
+        self.coefficients = default_coeffs
+        self.rescale_func = self._create_rescale_func(default_coeffs)
+        self.state_manager = StateManager(TeaCacheState, (), {})
+        self.extractor_fn = None
+        self.model_type = None
+
+    @staticmethod
+    def _create_rescale_func(coefficients):
```
DN6: Why do we need to create a rescale func? If we have coefficients set, we should be able to just call the function directly?

```py
def rescale_fn(self):
    return self.coefficients[0] * x**4 + self.coefficients[1] * x**3 + self.coefficients[2] * x**2 + self.coefficients[3] * x + self.coefficients[4]

```

DN6 (commented):
Some high level feedback on the design. The control flow is hard to follow as it switches between the hook object and adapter. The adapters themselves are thin wrappers around a modified forward function, so it would be better to just define them as standalone functions.  e.g.

```python
def _flux_forward(
    state: "TeaCacheState", # pass the state to the function not the hook object
    coefficients: List[float],
    rel_l1_thresh: float,
    module: torch.nn.Module,
    hidden_states: torch.Tensor,
    timestep: torch.Tensor,
    pooled_projections: torch.Tensor,
    txt_ids: torch.Tensor,
    img_ids: torch.Tensor,
    return_dict: bool = True,
    **kwargs,
):

    if _should_use_cache(state, modulated_inp, coefficients, rel_l1_thresh)
        hidden_states = _apply_cached_residual(state, hidden_states, modulated_inp)
    else:
	    # run compute
	    _update_cache(state, hidden_states, original_hidden_states, modulated_inp)
```

Since we're hooking the top level forward of the model, we can map this forward function using the class name during hook initialization. 

```python
    def initialize_hook(self, module):
        """Initialize hook with model-specific configuration."""
        model_config = _MODEL_CONFIG.get(module.__name__)
        if model_config is None:
            raise ValueError

        if self.config.coefficients is not None:
            self.coefficients = self.config.coefficients
        else:
            self.coefficients = model_config["coefficients"]

        # Initialize state
        self.state_manager = StateManager(TeaCacheState)
        self.forward_fn = model_config["forward_func"]

        return module

```

Where `_MODEL_CONFIG` is just a mapping for the forward functions and coefficients

```python
_MODEL_CONFIG = {
    "FluxTransformer2DModel": {
        "forward_func": _flux_forward,
        "coefficients": [4.98651651e02, -2.83781631e02, 5.58554382e01, -3.82021401e00, 2.64230861e-01],
    },
}
```

Similarly, the methods defined in the hook object could also be turned into utility functions.  

```python
def _compute_rescaled_distance(rel_distance: float, coefficients: List[float]) -> float:
    return (
        coefficients[0] * rel_distance**4
        + coefficients[1] * rel_distance**3
        + coefficients[2] * rel_distance**2
        + coefficients[3] * rel_distance
        + coefficients[4]
    )
    
def _should_use_cache(state: "TeaCacheState", ...):
	# Return True or False based on whether to use cache. 
	return 
	
def _update_cache(state: "TeaCacheState)
	return 

def _apply_cached_residual(
    state: "TeaCacheState", input_base: torch.Tensor, modulated_inp: torch.Tensor
) -> torch.Tensor:
    """
    Apply cached residual to input (fast path).
    """
    output = input_base + state.previous_residual
    state.previous_modulated_input = modulated_inp
    state.cnt += 1
    return output
```

Let's remove passing `cache_fn` and `compute_fn` between the hook and the adapter. Use operations directly on the cache state + globally available utility methods. We can also remove the modulation extractors and move that logic into the model specific forward functions.

LawJarp-A:
Thanks for the feedback @DN6 
I'll take this week and rework it. 
I had left some redundant code while trying to figure out the organization, will clean it up

src/diffusers/hooks/teacache.py
LawJarp-A: Yes. Will remove this

LawJarp-A:
The per-model forward code is unavoidable due to different model architectures. The adapter pattern was an attempt to organize this, but I agree standalone functions would be cleaner. I'll refactor.

LawJarp-A:
Hi @DN6 , I've updated the implementation as you requested:

  - Replaced adapter classes with standalone forward functions
  - Created _MODEL_CONFIG mapping for forward functions and coefficients
  - Removed cache_fn/compute_fn closures - now using direct if/else logic in each forward
  - Extracted utility functions: _should_compute(), _update_state(), _apply_cached_residual()
  - Removed enable_teacache() - now only enable_cache(TeaCacheConfig(...))
  - Inlined modulation extractors into forward functions

  This does introduce some code duplication - each forward function now has the same if/else pattern:


```python
  if _should_compute(state, modulated_inp, hook.coefficients, hook.config.rel_l1_thresh):
      # compute full transformer
      _update_state(state, output, original, modulated_inp)
  else:
      output = _apply_cached_residual(state, input, modulated_inp)
```

  But the control flow is now much clearer - you can read each forward function top-to-bottom without jumping between closures and hook methods.

  Let me know if you'd like any further changes!

LawJarp-A:
@DN6 @sayakpaul  I spent the weekend going over the code again to understand and simplify
- I have updated the cache context to be set in the denosing loop itself
- removed redundant code
- tested it with all models on a h100 and updated it in the PR description

I have kept it with per model forward function like you requested instead of the common adapter pattern I was using before. 
Please review it now, I think it addresses all the recent feedback I have recieved

Btw, below are the images generated w and w/o cache

Mochi
image


lumina2
image

flux
image

cogxvideo
image

src/diffusers/hooks/hooks.py
```diff
@@ -40,11 +40,14 @@ def __init__(self, state_cls: BaseState, init_args=None, init_kwargs=None):
...
-        if self._current_context is None:
-            raise ValueError("No context is set. Please set a context before retrieving the state.")
-        if self._current_context not in self._state_cache.keys():
-            self._state_cache[self._current_context] = self._state_cls(*self._init_args, **self._init_kwargs)
-        return self._state_cache[self._current_context]
+        context = self._current_context
+        if context is None:
+            # Fallback to default context for backward compatibility with
+            # pipelines that don't call cache_context()
+            context = "_default"
```
sayakpaul: Should this branch not error out like previous?

src/diffusers/hooks/hooks.py
```diff
@@ -133,7 +136,7 @@ def reset_state(self, module: torch.nn.Module):
         return module
 
     def _set_context(self, module: torch.nn.Module, name: str) -> None:
-        # Iterate over all attributes of the hook to see if any of them have the type `StateManager`. If so, call `set_context` on them.
+        """Set context on all StateManager attributes of this hook."""
```
sayakpaul: Can we please revert the changes unrelated to this PR? Makes reviewing a bit easier since the diff becomes smaller.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,901 @@
...
+    if attention_kwargs is None:
+        return None, 1.0
+    attention_kwargs = attention_kwargs.copy()
+    return attention_kwargs, attention_kwargs.pop("scale", 1.0)
+
+
+def _get_model_config() -> Dict[str, Dict[str, Any]]:
+    """Get model configuration mapping. Order matters: more specific variants before generic ones."""
+    return {
+        "FluxKontext": {
```
sayakpaul: WDYT of using the actual model class names here? Will likely be easier to maintain and read.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,901 @@
...
+            "forward_func": _flux_teacache_forward,
+            "coefficients": [-1.04655119e03, 3.12563399e02, -1.69500694e01, 4.10995971e-01, 3.74537863e-02],
```
sayakpaul: (nit): would make a note on how these values were obtained.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,901 @@
...
+    """Polynomial rescaling: c[0]*x^4 + c[1]*x^3 + c[2]*x^2 + c[3]*x + c[4]"""
+    c = coefficients
+    return c[0] * x**4 + c[1] * x**3 + c[2] * x**2 + c[3] * x + c[4]
+
+
+def _compute_rel_l1_distance(current: torch.Tensor, previous: torch.Tensor) -> float:
+    """Compute relative L1 distance between tensors."""
+    prev_mean = previous.abs().mean()
+    if prev_mean.item() > 1e-9:
+        return ((current - previous).abs().mean() / prev_mean).item()
```
sayakpaul: Do we need to make it data-dependent (`item()` call)? Raising it because it makes `torch.compile` cry.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,901 @@
...
+    return 0.0 if current.abs().mean().item() < 1e-9 else float("inf")
+
+
+@torch.compiler.disable
```
sayakpaul, line 156: Is it because of the `item()` call?

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,901 @@
...
+        attention_kwargs=attention_kwargs,
+        return_dict=return_dict,
+    )
+    hidden_states, encoder_hidden_states, timestep, encoder_attention_mask = args
+    attention_kwargs = kwargs.get("attention_kwargs", attention_kwargs)
+    return_dict = kwargs.get("return_dict", return_dict)
+
+    attention_kwargs, lora_scale = _extract_lora_scale(attention_kwargs)
+    if USE_PEFT_BACKEND:
+        scale_lora_layers(module, lora_scale)
```
sayakpaul, line 658: We should check if the underlying model class inherits from `PeftLoaderMixin` and if so, we should do it.

sayakpaul (commented):
I left some comments. LMK if they make sense.

src/diffusers/hooks/teacache.py
```diff
@@ -0,0 +1,901 @@
...
+        return None, 1.0
+    attention_kwargs = attention_kwargs.copy()
+    return attention_kwargs, attention_kwargs.pop("scale", 1.0)
+
+
+def _get_model_config() -> Dict[str, Dict[str, Any]]:
+    """Get model configuration mapping. Order matters: more specific variants before generic ones."""
+    return {
+        "FluxKontext": {
+            "forward_func": _flux_teacache_forward,
```
sayakpaul: One problem that I see with this is what happens when the core forward method undergoes some changes and we fail to propagate them in these modified forwards.

sayakpaul:
@bot /style

LawJarp-A:
Thanks for the review. Taking a look

src/diffusers/hooks/teacache.py
LawJarp-A: fair enough, i'll port that into a complete tensor ops flow

sayakpaul:
@LawJarp-A I am guessing the Copilot review comments were resolved? There also seems to be a couple of unresolved comments.

sayakpaul:
Cc: @LiewFeng would you like to give this a review as well?

sayakpaul:
@LawJarp-A a gentle ping.

LawJarp-A:
> @LawJarp-A a gentle ping.

Noted @sayakpaul 
Working on a timeline of another week or two.
I'll work on the unresolved comments.

sayakpaul:
@LawJarp-A sounds good. Let us know whenever you're ready.

sayakpaul:
@LawJarp-A let us know.

LawJarp-A:
> @LawJarp-A let us know.

Yessir. Will ping here

LawJarp-A:
Hi @sayakpaul, here’s a quick update addressing the remaining review comments.

**Fixes in the latest commit:**

* Fixed infinite recursion in `_is_peft_adapter`. It now checks `PeftAdapterMixin` correctly using a lazy import.
* Handled a Lumina2 edge case. `_maybe_reset_state_for_new_inference` now re-reads `num_steps` at `cnt=0` and resets if it changes. This covers cases where `cnt` wraps without a normal reset. (In standard usage, `maybe_free_model_hooks()` already resets correctly.)
* Fixed test assertion: updated message from “must be positive” to “must be non-negative” to match the actual error.
* Updated `apply_teacache` docstring to recommend `pipe.transformer.enable_cache(config)` for pipeline usage.

**Responses to open comments:**

* **StateManager fallback (hooks.py:47):** The default fallback is needed since some pipelines (like Lumina2) don’t use `cache_context()` in their loop. Throwing an error would break them. I can add a `logger.debug` if you prefer visibility.
* **Calibration utility:** Custom coefficients can already be passed via `TeaCacheConfig`. A calibration step like #12648 would be useful—I'll open a separate issue.
* **Configurable rescaling:** The current 4th-degree polynomial comes from the paper and works across tested models. We can revisit this if needed.
* **`.item()` and `torch.compile`:** All operations stay in tensor form except the final `.item()` for branching, which is required for Python control flow. This is the minimal graph break.
* **Copilot comments:** All addressed—boundary checks, class name matching, zero threshold validation, and Lumina2 state reset.

joangava (approved):

joangava (approved):

sayakpaul:
> .item() and torch.compile: All operations stay in tensor form except the final .item() for branching, which is required for Python control flow. This is the minimal graph break.

But we can handle through pure tensors no? Problem with .item() is that it becomes data-dependent and that wouldn't be known during runtime.

Sorry for the delay.

LawJarp-A:
> But we can handle through pure tensors no? Problem with .item() is that it becomes data-dependent and that wouldn't be known during runtime.

would torch.cond address this? it keeps the predicae as a 0-dim bool tensor and traces both branches, so no graph break, but it needs state mutation moved out of the branches (gating the residual update with torch.where), which is a non-trivial refactor across the four per-model forwards. Happy to do it if that's the direction.

The lighter alternative is matching what first_block_cache.py:133 and mag_cache.py:300 do today, wrap the decision callsite with @torch.compiler.disable so the rest of the forward still compiles. 
Happy to go either way; let me know which bar you want for TeaCache.
@sayakpaul

joangava (approved):

LawJarp-A:
@sayakpaul @DN6 any more comments to address?

LawJarp-A:
@sayakpaul / @DN6  gentle ping on this one.

I also had some suggestions from claude, which i want to review and implement (full disclosure)

LawJarp-A:
@sayakpaul rebased onto main, and answered the `.item()` question in code.

the decision helpers now carry `@torch.compiler.disable`, so the data-dependent branch is contained and the surrounding forward still compiles. same as `first_block_cache.py:133`, `taylorseer_cache.py:225` and `mag_cache.py:184`/`:300`. checked with `torch._dynamo.explain`: teacache contributes exactly one break, at `_should_compute`. the model's other breaks are `aten.nonzero` from its own mask handling and are there with no cache applied. `torch.cond` would remove that last one but needs cache state threaded through both branches, better as a follow-up across all three hooks.

branch no longer conflicts, 27 teacache tests pass, ruff clean.

one still open: the `StateManager` fallback in `hooks.py` is only there because lumina2's pipeline doesn't wrap its loop in `cache_context()`. i'd rather add that to the lumina2 loop and revert `hooks.py`, if touching a pipeline file is fine.

@AtulJoshi1206 rebase is done, thanks for the nudge.

DN6:
Hi @LawJarp-A. Sorry for the delay on our end. I think one thing about the current design that in hindsight isn't sustainable is the custom forwards defined for each model. Ensuring there is no drift between cached forwards and the ones on main will be difficult, so I think it would be best to use the Head/Block hook approach, similar to MagCache and FirstBlockCache
https://github.com/huggingface/diffusers/blob/ae2e4c7907ccc21e52ebe86b349d67e9b0b9f316/src/diffusers/hooks/mag_cache.py#L171

It would also be good to provide a way for users to calibrate the coefficients for a new model like Magcache.

Issue #12344, opened 2025-09-17
Qwen Image: txt_seq_lens is redundant and not used


dxqb:
### Describe the bug

Here https://github.com/huggingface/diffusers/blob/efb7a299af46d739dec6a57a5d2814165fba24b5/src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py#L675 the parameter `txt_seq_lens` is calculated.

It is set to the length of the text attention mask for each batch sample - but both the attention mask *and*  `txt_seq_lens` are passed to the transformer:
https://github.com/huggingface/diffusers/blob/efb7a299af46d739dec6a57a5d2814165fba24b5/src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py#L695C38-L695C39

This is redundant in principle, but not only that: *both* of them are not used
We have already learned that the attention mask passed to the transformer is ignored: https://github.com/huggingface/diffusers/issues/12294
The only point in code where `txt_seq_lens` is used is here:
https://github.com/huggingface/diffusers/blob/efb7a299af46d739dec6a57a5d2814165fba24b5/src/diffusers/models/transformers/transformer_qwenimage.py#L231

Therefore only the maximum value in the 1D array is used, which is the longest mask. The longest mask when using the entire Qwen Image pipeline is *by definition* the shape of the text embeddings, because they are encoded using the `longest` padding strategy here:
https://github.com/huggingface/diffusers/blob/efb7a299af46d739dec6a57a5d2814165fba24b5/src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py#L203
[`True` is equivalent to `longest`]

Using this padding strategy, you always have at least 1 batch sample in which the attention mask is completely `True`, therefore `max(txt_seq_lens)` is the token length.

If you do not use the entire pipeline but only the transformer, you could pass `txt_seq_lens` where `max(txt_seq_lens)` is not equal to the shape of the embeddings. However, then there is a shape mismatch here:

https://github.com/huggingface/diffusers/blob/efb7a299af46d739dec6a57a5d2814165fba24b5/src/diffusers/models/transformers/transformer_qwenimage.py#L138                               ~~~~~~~~~~^~~~~~~~~~~
`RuntimeError: The size of tensor a (512) must match the size of tensor b (6) at non-singleton dimension 1`
`max(txt_seq_lens)` in this example was 6, while the embeddings shape was `[..., 512, ...]`

Conclusion:
[...]

PR #12655, opened 2025-11-14
Fix qwen encoder hidden states mask

### LINKED ISSUE #12294: [Qwen-image] encoder_hidden_states_mask is not used
### Describe the bug

During text encoding stage, the text encoder will produce encoder_hidden_states_mask to mask some regions. It usually happens during training with batch size more than 1, to make sure that the produce text embedding has the same sequence length.

However, when I check the following code, it seems weird that encoder_hidden_states_mask is not used even though it has been passed to the function.

It indeed works when batch=1. However, I feel it is indeed a bug if we train Qwen-image with batch size more than 1.

### Reproduction

```
class QwenDoubleStreamAttnProcessor2_0:
    """
    Attention processor for Qwen double-stream architecture, matching DoubleStreamLayerMegatron logic. This processor
    implements joint attention computation where text and image streams are processed together.
    """

    _attention_backend = None

    def __init__(self):
        if not hasattr(F, "scaled_dot_product_attention"):
            raise ImportError(
                "QwenDoubleStreamAttnProcessor2_0 requires PyTorch 2.0, to use it, please upgrade PyTorch to 2.0."
            )

    def __call__(
        self,
        attn: Attention,
        hidden_states: torch.FloatTensor,  # Image stream
        encoder_hidden_states: torch.FloatTensor = None,  # Text stream
        encoder_hidden_states_mask: torch.FloatTensor = None,
        attention_mask: Optional[torch.FloatTensor] = None,
        image_rotary_emb: Optional[torch.Tensor] = None,
    ) -> torch.FloatTensor:
        if encoder_hidden_states is None:
            raise ValueError("QwenDoubleStreamAttnProcessor2_0 requires encoder_hidden_states (text stream)")

        seq_txt = encoder_hidden_states.shape[1]

        # Compute QKV for image stream (sample projections)
        img_query = attn.to_q(hidden_states)
        img_key = attn.to_k(hidden_states)
        img_value = attn.to_v(hidden_states)

        # Compute QKV for text stream (context projections)
        txt_query = attn.add_q_proj(encoder_hidden_states)
        txt_key = attn.add_k_proj(encoder_hidden_states)
        txt_value = attn.add_v_proj(encoder_hidden_states)

        # Reshape for multi-head attention
        img_query = img_query.unflatten(-1, (attn.heads, -1))
        img_key = img_key.unflatten(-1, (attn.heads, -1))
        img_value = img_value.unflatten(-1, (attn.heads, -1))

        txt_query = txt_query.unflatten(-1, (attn.heads, -1))
        txt_key = txt_key.unflatten(-1, (attn.heads, -1))
        txt_value = txt_value.unflatten(-1, (attn.heads, -1))

        # Apply QK normalization
        if attn.norm_q is not None:
            img_query = attn.norm_q(img_query)
        if attn.norm_k is not None:
            img_key = attn.norm_k(img_key)
        if attn.norm_added_q is not None:
            txt_query = attn.norm_added_q(txt_query)
            txt_key = attn.norm_added_k(txt_key)

        # Apply RoPE
        if image_rotary_emb is not None:
            img_freqs, txt_freqs = image_rotary_emb
            img_query = apply_rotary_emb_qwen(img_query, img_freqs, use_real=False)
            img_key = apply_rotary_emb_qwen(img_key, img_freqs, use_real=False)
            txt_query = apply_rotary_emb_qwen(txt_query, txt_freqs, use_real=False)
            txt_key = apply_rotary_emb_qwen(txt_key, txt_freqs, use_real=False)

        # Concatenate for joint attention
        # Order: [text, image]
        joint_query = torch.cat([txt_query, img_query], dim=1)
        joint_key = torch.cat([txt_key, img_key], dim=1)
        joint_value = torch.cat([txt_value, img_value], dim=1)

        # Compute joint attention
        joint_hidden_states = dispatch_attention_fn(
            joint_query,
            joint_key,
            joint_value,
            attn_mask=attention_mask,
            dropout_p=0.0,
            is_causal=False,
            backend=self._attention_backend,
        )

        # Reshape back
        joint_hidden_states = joint_hidden_states.flatten(2, 3)
        joint_hidden_states = joint_hidden_states.to(joint_query.dtype)

        # Split attention outputs back
        txt_attn_output = joint_hidden_states[:, :seq_txt, :]  # Text part
        img_attn_output = joint_hidden_states[:, seq_txt:, :]  # Image part

        # Apply output projections
        img_attn_output = attn.to_out[0](img_attn_output)
        if len(attn.to_out) > 1:
            img_attn_output = attn.to_out[1](img_attn_output)  # dropout

        txt_attn_output = attn.to_add_out(txt_attn_output)

        return img_attn_output, txt_attn_output

```

### Logs

```shell

```

### System Info

A100

### Who can help?

@sayakpaul

cdutr:
# What does this PR do?

Fixes the QwenImage encoder to properly apply `encoder_hidden_states_mask` when passed to the model. Previously, the mask parameter was accepted but ignored, causing padding tokens to incorrectly influence attention computation.

## Changes

- **Attention mask application**: Modified `QwenDoubleStreamAttnProcessor2_0` to create a 2D attention mask from the 1D `encoder_hidden_states_mask`, properly masking text padding tokens while keeping all image tokens unmasked
- **RoPE adjustment**: Updated positional embedding computation to use the full padded sequence length when a mask is present, ensuring correct position indices
- **Tests**: Added comprehensive tests validating that:
  - Padding tokens are properly isolated and don't affect outputs
  - Masked outputs differ significantly from unmasked outputs
- **Benchmarks**: Included performance analysis showing acceptable overhead (<20% for inference, ~19% for training scenarios)

## Impact

This fix enables proper Classifier-Free Guidance (CFG) batching with variable-length text sequences, which is common when batching conditional and unconditional prompts together.

### Benchmark Results

| Scenario | Latency (ms) | Peak Memory (MB) | Throughput (iter/s) |
|----------|-------------|------------------|---------------------|
| Baseline (no mask) | 11.68 ± 0.23 | 301.5 | 84.70 |
| Mask all-ones (no padding) | 12.01 ± 0.26 | 301.5 | 82.34 |
| Mask with padding (CFG) | 13.86 ± 0.24 | 301.5 | 71.42 |

**Overhead**: +2.8% for mask processing without padding, +18.7% with actual padding (realistic CFG scenario)

The higher overhead with padding is expected and acceptable as it represents the cost of properly handling variable-length sequences in batched inference. This is a necessary correctness fix rather than an optimization. Test ran on RTX 4070 12GB.

Fixes #12294

---

## Before submitting
- [x] This PR fixes a bug in the code
- [x] This PR adds tests that verify the fix
- [x] This PR includes benchmarks demonstrating performance impact

## Who can review?

@yiyixuxu @sayakpaul - Would appreciate your review, especially regarding the benchmarking approach. I used a custom benchmark rather than `BenchmarkMixin` because:
1. This tests a specific bug fix (mask application) rather than optimization strategies
2. The fix uses synthetic models to isolate the mask handling logic
3. Standard benchmarks focus on pretrained model performance with different quantization/offloading strategies
4. The metrics needed are different (latency distribution, throughput) vs standard format (compile/plain time comparison)

**Note**: The benchmark file is named `benchmarking_qwenimage_mask.py` (with "benchmarking" prefix) rather than `benchmark_qwenimage_mask.py` to prevent it from being picked up by `run_all.py`, since it doesn't use `BenchmarkMixin` and produces a different CSV schema. If you prefer, I can adapt it to use the standard format instead.

Happy to adjust the approach if you have suggestions!

### DIFF
--- a/src/diffusers/models/transformers/transformer_qwenimage.py
+++ b/src/diffusers/models/transformers/transformer_qwenimage.py
@@ -358,7 +358,8 @@ class QwenDoubleStreamAttnProcessor2_0:
             )
 
             joint_attention_mask_1d = torch.cat([text_attention_mask, image_attention_mask], dim=1)
-            attention_mask = joint_attention_mask_1d[:, None, None, :] * joint_attention_mask_1d[:, None, :, None]
+            # broadcastable shape for SDPA
+            attention_mask = joint_attention_mask_1d[:, None, None, :]
 
         # Compute joint attention
         joint_hidden_states = dispatch_attention_fn(


sayakpaul:
@cdutr it's great that you have also included the benchmarking script for fullest transparency. But we can remove that from this PR and instead have that as a GitHub gist. 

The benchmark numbers make sense to me. Some comments:

* Could we also check the performance with `torch.compile`? 
* Could we also see some image outputs with and without the changes introduced in this PR?

Also, I think a natural next step would be see how well this performs when combined with FA varlen. WDYT? 

@naykun what do you think about the changes?

cdutr:
Thanks @sayakpaul! I removed the benchmark script, moved all tests to this [gist](https://gist.github.com/cdutr/bea337e4680268168550292d7819dc2f).

## torch.compile test

Also tested the performance with `torch.compile`, and results were similar, the details are below.

Tested on NVIDIA A100 80GB PCIe:

```
Without compile: 4.70ms per iteration
With compile:    1.93ms per iteration
Speedup:         2.43x
Compilation overhead: 7.25s (one-time cost)
```

Also validated on RTX 4050 6GB (laptop) with similar results (2.38x speedup). The mask implementation is fully compatible with torch.compile.

## Image outputs

Tested End-to-end image generation: Successfully generated images using QwenImagePipeline and pipeline runs without errors, here is the output generated:

test_output

## FA Varlen
FA varlen is the natural next step, yes! I'm interested in working on it. Should I keep iterating in this PR, or should we merge it and create a new issue?

The mask infrastructure this PR adds would translate well to varlen, instead of masking padding tokens, we'd pack only valid tokens using the same sequence length information

sayakpaul:
Thanks for the results! Looks quite nice.

> FA varlen is the natural next step, yes! I'm interested in working on it. Should I keep iterating in this PR, or should we merge it and create a new issue?

I think it's fine to first merge this PR and then we work on it afterwards. We're adding easier support for Sage and FA2 in this PR: https://github.com/huggingface/diffusers/pull/12439, so after that's merged, it will be quite easy to work on that (thanks to the `kernels` lib).

Could we also check if the outputs deviate with and without the masks, i.e., the outputs we get on `main` and this PR branch?

sayakpaul:
@dxqb would you maybe interested in checking this PR out as well?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -330,6 +330,31 @@ def __call__(
...
+                    f"must match encoder_hidden_states sequence length ({text_seq_len})"
+                )
+
+            text_attention_mask = encoder_hidden_states_mask.bool()
+            image_attention_mask = torch.ones(
+                (batch_size, image_seq_len), dtype=torch.bool, device=hidden_states.device
+            )
+
+            joint_attention_mask_1d = torch.cat([text_attention_mask, image_attention_mask], dim=1)
+            attention_mask = joint_attention_mask_1d[:, None, None, :] * joint_attention_mask_1d[:, None, :, None]
```
dxqb: this works, but optimization possible:
this generates a real 2D mask in memory, of size seq_len * seq_len. An attention mask that is broadcastable to the required shape is enough:
`attention_mask_2d = attention_mask[:, None, None, :]`

Why this is enough: all tokens don't attend to the masked tokens anymore. whether the masked tokens attend to any other tokens is irrelevant, because they are masked in all layers and their result is never used.
I have tested this and the results were pixel-identical.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -330,6 +330,31 @@ def __call__(
...
+                    f"encoder_hidden_states_mask batch size ({encoder_hidden_states_mask.shape[0]}) "
+                    f"must match hidden_states batch size ({batch_size})"
+                )
+            if encoder_hidden_states_mask.shape[1] != text_seq_len:
+                raise ValueError(
+                    f"encoder_hidden_states_mask sequence length ({encoder_hidden_states_mask.shape[1]}) "
+                    f"must match encoder_hidden_states sequence length ({text_seq_len})"
+                )
+
+            text_attention_mask = encoder_hidden_states_mask.bool()
```
dxqb, line 355: This works if the `encoder_hidden_states_mask` is already bool, or a float tensor with the same semantics.
bool attention masks are enough for the usual usecase of masking unused text tokens, but if only bool attention masks are supported this should be clearly documented. also maybe change the type hint?

see https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html how float attention masks are interpreted by torch. a float `0.0` is *not* masked, a bool `False` *is* masked.

there are *some* usecases for float attention masks for text sequences, like putting an emphasis/bias on certain tokens. not very common though, so if you decide to only support bool attention masks that makes sense to me - but requires documentation.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -630,7 +655,15 @@ def forward(
             else self.time_text_embed(timestep, guidance, hidden_states)
         )
 
-        image_rotary_emb = self.pos_embed(img_shapes, txt_seq_lens, device=hidden_states.device)
+        # Use padded sequence length for RoPE when mask is present.
+        # The attention mask will handle excluding padding tokens.
+        if encoder_hidden_states_mask is not None:
+            txt_seq_lens_for_rope = [encoder_hidden_states.shape[1]] * encoder_hidden_states.shape[0]
```
dxqb, line 667: could you please read this https://github.com/huggingface/diffusers/issues/12344#issuecomment-3516800940
I don't think this is how the `txt_seq_lens` parameter was intended.

However, your change here might still be a valid (temporary) fix, because it's currently (before the PR) not used as intended either

dxqb: just noticed, this was already discussed in other comments above.

src/diffusers/models/transformers/transformer_qwenimage.py
sayakpaul: Explicit is preferred if it's a micro-optimization. So, I think it's fine as is.

sayakpaul:
@naykun WDYT about this PR?

cdutr:
Hey. I would like to work on the last inputs and comments. I was off last week attending a conference but should go back to it tomorrow. Any additional feedback or request is appreciated, of course

cdutr:
Hi, I am back, will reply the points below:

## @dxqb

### Comment 1: encoder_hidden_states_mask Type Hints

Updated to use `torch.BoolTensor` type hint and added documentation clarifying that only boolean masks are supported. Float masks have different semantics in PyTorch SDPA (0.0 = not masked, False = masked), and supporting both would require additional conversion logic. Bool masks cover the standard use case of excluding padding tokens.

### Comment 2: 2D Attention Mask Optimization

Implemented. Changed from generating full `[batch, seq_len, seq_len]` masks to broadcasting-friendly `[batch, 1, 1, seq_len]` shape. This significantly reduces memory usage (e.g., from 73,728 to 384 elements for batch=2, seq_len=192) while maintaining correctness through broadcasting in SDPA.

## @sayakpaul

### torch.compile Performance

I've tested with `torch.compile()` and the results show no performance regression:

**Without compile:**
- With mask: 0.0439s per iteration
- Without mask: 0.0438s per iteration

**With compile:**
- With mask: 0.0437s per iteration
- Without mask: 0.0436s per iteration

The mask implementation is compile-friendly and maintains performance. See full results in the [gist](https://gist.github.com/cdutr/bea337e4680268168550292d7819dc2f).

### Output Comparison (main vs PR branch)

I ran comprehensive tests comparing outputs between main and this PR branch:

**1. Numerical Comparison:**
- PR branch: Mean difference = 1.08e-04 (mask functional)
- Main branch: Mean difference = 0.00e+00 (mask ignored)

**2. Image Generation Comparison:**
Using batched prompts with different lengths (natural padding scenario):
- Prompt: "A cat" (7 tokens) batched with longer prompt (14 tokens) creates 7 padding tokens
- With mask (PR): Padding properly excluded from attention
- Without mask (simulating bug): Padding treated as real tokens
- Result: 87.17% of pixels differ (mean absolute difference: 9.35/255)

With mask:

with_mask

Without mask:

without_mask

Difference:

difference_map

**3. Main Branch Status:**
Main branch crashes when using masks with batched variable-length prompts:
```
RuntimeError: The size of tensor a (14) must match the size of tensor b (7)
```
This is the exact bug this PR fixes.

All test scripts and results available in the gist: https://gist.github.com/cdutr/bea337e4680268168550292d7819dc2f

sayakpaul:
Thanks for your work here! 

Do you think there might some overlap between https://github.com/huggingface/diffusers/pull/12702 and this PR?

kashif:
@cdutr yes its somewhat duplicate work to what I have, I can remove the stuff from my PR, and then you can potentially point your PR to mine with these changes and that way we can merge both things? what do you think?

cdutr:
Hey @kashif, the comparisons on your PR are really nice\! Yes, I can point my PR to yours. Let me know once you've removed the overlapping parts and I'll rebase on top of your branch.

yiyixuxu:
@kashif 
Is it ok we just go with your PR and make @cdutr a co-author in your PR instead? 
the PR is already under review for qwen team, it is a big PR and will take some work to review & test for them.  Let's not make it hadrer than necessarily 

@cdutr sorry this happened but we are really impressed with work here and will do the MVP rewards accordingly :)

sayakpaul:
Adding on top of @yiyixuxu, 

@cdutr since you have already set up the benchmarking for this setup, maybe you could contribute that @kashif's PR as well? Maybe we could document all of that inside https://huggingface.co/docs/diffusers/main/en/api/pipelines/qwenimage?

Again, we're sorry that this happened. But regardless, we would like to honor your contributions and grant you the MVP rewards as Yiyi mentioned (after the PR is merged).

cdutr:
Thanks a lot for the MVP rewards, I really appreciate it and I'm really enjoying working on this project.

It makes total sense to move forward with merging the other PR since it's already in review. For sure, I can contribute the benchmarks and documentation there. I’ll join the discussion there

kashif:
@cdutr  you should be able to commit dir to my branch

cdutr:
@sayakpaul, I will close this PR, since we are working in the other one, ok?

sayakpaul:
Yes, fine. Once again, thanks for your contributions! I will get back to you regarding the rewards of MVP shortly. We will collaborate more, for sure :)

PR #12688, opened 2025-11-20
Add FluxLoraLoaderMixin to Fibo pipeline

SwayStar123:
The internal implementation checks if isinstance(self, FluxLoraLoaderMixin):, but its not actually implementing it.

It needs to be added to allow loading lora


DN6:
@SwayStar123 Change looks good. Do you happen to have a LoRA available to run a quick test?

DN6 (approved):

SwayStar123:
Yes I have tested it with a lora

DN6:
Possible to share a link the LoRA so I can run a check on my end?

Issue #12096, opened 2025-08-07
WanVACEPipeline - doesn't work with apply_group_offloading


SlimRG:
### Describe the bug

When enable apply_group_offloading:
```
Traceback (most recent call last):
  File "D:\Experiments\Video_Outpaint\2__Outpaint.py", line 165, in <module>
    out = pipe(
          ^^^^^
  File "C:\Users\BBCCA\AppData\Local\Programs\Python\Python312\Lib\site-packages\torch\utils\_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\BBCCA\AppData\Local\Programs\Python\Python312\Lib\site-packages\diffusers\pipelines\wan\pipeline_wan_vace.py", line 873, in __call__
    conditioning_latents = self.prepare_video_latents(video, mask, reference_images, generator, device)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\BBCCA\AppData\Local\Programs\Python\Python312\Lib\site-packages\diffusers\pipelines\wan\pipeline_wan_vace.py", line 532, in prepare_video_latents
    inactive = retrieve_latents(self.vae.encode(inactive), generator, sample_mode="argmax")
                                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\BBCCA\AppData\Local\Programs\Python\Python312\Lib\site-packages\diffusers\utils\accelerate_utils.py", line 46, in wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\BBCCA\AppData\Local\Programs\Python\Python312\Lib\site-packages\diffusers\models\autoencoders\autoencoder_kl_wan.py", line 1191, in encode
    h = self._encode(x)
        ^^^^^^^^^^^^^^^
  File "C:\Users\BBCCA\AppData\Local\Programs\Python\Python312\Lib\site-packages\diffusers\models\autoencoders\autoencoder_kl_wan.py", line 1158, in _encode
    out = self.encoder(x[:, :, :1, :, :], feat_cache=self._enc_feat_map, feat_idx=self._enc_conv_idx)
[...]

PR #12692, opened 2025-11-21
Fix broken group offloading with block_level for models with standalone layers

### LINKED ISSUE #12319: Broken group offloading using block_level
### Describe the bug

pr #12283 introduced group offloading on a pipeline  
but its currently broken when using offload_type="block_level" as it does not handle vae module correctly  
and result is failure with input vs weight device mismatch  
using `offload_type="leaf_level"` it seems to work  

### Reproduction

```py
import torch
import diffusers

model = '/mnt/models/stable-diffusion/mine/tempest-by-vlad-0.1.safetensors'
cache_dir = '/mnt/models/Diffusers'
pipe = diffusers.StableDiffusionXLPipeline.from_single_file(
    model,
    torch_dtype=torch.bfloat16,
    cache_dir=cache_dir,
)
pipe.enable_group_offload(
    onload_device=torch.device('cuda:0'),
    offload_device=torch.device('cpu'),
    offload_type="block_level",
    num_blocks_per_group=1,
    use_stream=True
)
image = pipe(
    prompt='A beautiful painting of a futuristic cityscape at sunset',
    width=1024,
    height=1024,
    num_inference_steps=10,
).images[0]
image.save('/tmp/output.png')
```

### Logs

```shell
...
File: diffusers/models/autoencoders/autoencoder_kl.py", line 294, in _decode
> image = self.vae.decode(latents, return_dict=False)[0]
...
RuntimeError: Input type (CUDABFloat16Type) and weight type (CPUBFloat16Type) should be the same
```

### System Info

python==3.12, diffusers==5e181eddfe7e44c1444a2511b0d8e21d177850a0

### Who can help?

@sayakpaul @yiyixuxu @DN6 @asomoza @a-r-r-o-w

rycerzes:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/issues/12319

This PR fixes the device mismatch error that occurs when using `block_level` group offloading with models containing standalone computational layers (like VAE's `post_quant_conv` and `quant_conv`).

#### Problem

When using `block_level` offloading, the implementation only matched `ModuleList` and `Sequential` containers, leaving standalone layers (like `Conv2d`) unmanaged. These layers remained on CPU while their inputs were on CUDA, causing:
```
RuntimeError: Input type (CUDABFloat16Type) and weight type (CPUBFloat16Type) should be the same
```
Block-level offloading logic in `group_offloading.py` only looked for `ModuleList` and `Sequential` containers when creating groups. Standalone computational layers like the VAE's `post_quant_conv` (a `Conv2d` layer) were not included in any group, so they never received hooks to manage their device placement. This caused them to remain on CPU while their inputs were transferred to CUDA.

The block-level offloading logic has been modified to:
1. Identify standalone computational layers that don't belong to any `ModuleList`/`Sequential` container
2. Group them into an `unmatched_group` that gets proper hook management
3. Apply hooks to ensure proper device placement during forward pass

**Key Changes:**
- Updated `_create_groups_for_block_level_offloading()` function to collect unmatched computational layers
- Added logic to create a group for standalone layers using the same `_GO_LC_SUPPORTED_PYTORCH_LAYERS` filter
- Ensured the unmatched group is properly integrated into the hook chain

#### Testing
**Tested with:**
- SDXL VAE (`AutoencoderKL`) which has standalone `post_quant_conv` and `quant_conv` layers
- Created test cases for models with both standalone and deeply nested layer structures
- Confirmed both streaming and non-streaming modes work correctly

**Test Coverage:**
- `test_group_offloading_models_with_standalone_and_deeply_nested_layers` - Verifies the fix works with complex model architectures
- All existing group offloading tests continue to pass

#### Expected Behavior After Fix
**Before:** Block-level offloading fails with device mismatch error when models have standalone computational layers

**After:** Block-level offloading works correctly with all model architectures, including those with:
- Standalone `Conv2d`, `Linear`, and other computational layers
- Nested `ModuleList`/`Sequential` containers
- Mixed architectures with both standalone and containerized layers

## Before submitting
https://github.com/huggingface/diffusers/issues/12319#issuecomment-3556836262

## Who can review?

@sayakpaul this closes #12319

### DIFF
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -615,7 +615,7 @@ def _apply_group_offloading_block_level(module: torch.nn.Module, config: GroupOf
     done at the top-level blocks and modules specified in block_modules.
 
     When block_modules is provided, only those modules will be treated as blocks for offloading. For each specified
-    module, we either offload the entire submodule or recursively apply block offloading to it.
+    module, recursively apply block offloading to it.
     """
     if config.stream is not None and config.num_blocks_per_group != 1:
         logger.warning(


sayakpaul:
@vladmandic would you be interested in testing this out a bit?

rycerzes:
@sayakpaul this patch should also fix #12096 since both have the same root cause (standalone conv layers not tracked in block-level offloading), and this handles both Conv2d (SDXL) and Conv3d (Wan).

The fix should work for WanVACEPipeline as well.

sayakpaul:
Very cool! Feel free to add some lightweight tests to this PR and the final outputs so that we can also test ourselves.

rycerzes:
> Very cool! Feel free to add some lightweight tests to this PR and the final outputs so that we can also test ourselves.

Yes, I added tests in `test_group_offloading.py` covering the core fix [`test_block_level_stream_with_invocation_order_different_from_initialization_order`](https://github.com/rycerzes/diffusers/blob/59b6b678295214b70f6ecaa3f95129b76baf50d8/tests/hooks/test_group_offloading.py#L434) plus edge cases for VAE-like models with standalone layers, deeply nested structures, and parameter-only modules. 

I also created a [standalone test script](https://gist.githubusercontent.com/rycerzes/681dbf5624505cafff1ac62c215ffb17/raw/3f1bf179c28eb18c0c3f10bf0e7c6cb79015e128/test_sdxl_wan_vae_group_offloading.py) that validates SDXL VAE and AutoencoderKLWan with both block_level and leaf_level offloading. [Output](https://gist.github.com/rycerzes/681dbf5624505cafff1ac62c215ffb17?permalink_comment_id=5874406#gistcomment-5874406) of the script.

Pytest output
```
pytest tests/hooks/test_group_offloading.py -v
==================================================================== test session starts =====================================================================
platform win32 -- Python 3.13.3, pytest-9.0.1, pluggy-1.6.0 -- D:\Github\oss\diffusers\.venv\Scripts\python.exe
cachedir: .pytest_cache
rootdir: D:\Github\oss\diffusers
configfile: pyproject.toml
plugins: anyio-4.11.0, timeout-2.4.0, xdist-3.8.0, requests-mock-1.10.0
collected 20 items                                                                                                                                            

tests/hooks/test_group_offloading.py::GroupOffloadTests::test_block_level_offloading_with_parameter_only_module_group_0_block_level PASSED              [  5%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_block_level_offloading_with_parameter_only_module_group_1_leaf_level PASSED               [ 10%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_block_level_stream_with_invocation_order_different_from_initialization_order PASSED       [ 15%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_error_raised_if_group_offloading_applied_on_model_offloaded_module PASSED                 [ 20%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_error_raised_if_group_offloading_applied_on_sequential_offloaded_module PASSED            [ 25%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_error_raised_if_model_offloading_applied_on_group_offloaded_module PASSED                 [ 30%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_error_raised_if_sequential_offloading_applied_on_group_offloaded_module PASSED            [ 35%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_error_raised_if_streams_used_and_no_accelerator_device PASSED                             [ 40%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_error_raised_if_supports_group_offloading_false PASSED                                    [ 45%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_model_with_deeply_nested_blocks PASSED                                                    [ 50%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_model_with_only_standalone_layers PASSED                                                  [ 55%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_multiple_invocations_with_vae_like_model PASSED                                           [ 60%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_nested_container_parameters_offloading PASSED                                             [ 65%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_offloading_forward_pass PASSED                                                            [ 70%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_standalone_conv_layers_with_both_offload_types_0_block_level PASSED                       [ 75%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_standalone_conv_layers_with_both_offload_types_1_leaf_level PASSED                        [ 80%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_vae_like_model_with_standalone_conv_layers PASSED                                         [ 85%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_vae_like_model_without_streams PASSED                                                     [ 90%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_warning_logged_if_group_offloaded_module_moved_to_accelerator PASSED                      [ 95%]
tests/hooks/test_group_offloading.py::GroupOffloadTests::test_warning_logged_if_group_offloaded_pipe_moved_to_accelerator PASSED                        [100%]

===================================================================== 20 passed in 4.34s =====================================================================
```

sayakpaul:
Thanks for the comprehensive testing! I meant to ask for an even more minimal test script that utilizes group offloading with `block_level` and generates an output as expected. Something like:

```py
import torch 

pipe = DiffusionPipeline.from_pretrained("...", torch_dtype=torch.bfloat16)
pipe.transformer.enable_group_offloading(...)

# move rest of the components to CUDA
...

# inference
pipe(...)
```

src/diffusers/hooks/group_offloading.py
```diff
@@ -589,11 +593,20 @@ def _apply_group_offloading_block_level(module: torch.nn.Module, config: GroupOf
...
+            # Check if this is a computational layer that should be handled individually
+            if isinstance(submodule, _GO_LC_SUPPORTED_PYTORCH_LAYERS):
+                unmatched_computational_layers.append((name, submodule))
+                modules_with_group_offloading.add(name)
+            else:
+                # This is a container module - recursively find computational layers within it
+                _find_and_apply_computational_layer_hooks(submodule, name, config, modules_with_group_offloading)
+                unmatched_modules.append((name, submodule))
+                # Do NOT add the container name to modules_with_group_offloading here, because we need
+                # parameters from non-computational sublayers (like GroupNorm) to be gathered
```
sayakpaul: Could you expand a bit more on this?

tests/hooks/test_group_offloading.py
```diff
@@ -362,3 +502,161 @@ def apply_layer_output_tracker_hook(model: DummyModelWithLayerNorm):
             self.assertLess(
                 cumulated_absmax, 1e-5, f"Output differences for {name} exceeded threshold: {cumulated_absmax:.5f}"
             )
+
+    def test_vae_like_model_with_standalone_conv_layers(self):
```
sayakpaul: We can leverage an existing model implementations that were reported to be problematic with small configs and use them for testing here.

tests/hooks/test_group_offloading.py
```diff
@@ -362,3 +502,161 @@ def apply_layer_output_tracker_hook(model: DummyModelWithLayerNorm):
...
+        """Test that models with standalone Conv2d layers (like VAE) work with block-level offloading."""
+        if torch.device(torch_device).type not in ["cuda", "xpu"]:
+            return
+
+        model = DummyVAELikeModel(in_features=64, hidden_features=128, out_features=64)
```
sayakpaul: Prefer using much smaller configs for tests.

tests/hooks/test_group_offloading.py
```diff
@@ -362,3 +502,161 @@ def apply_layer_output_tracker_hook(model: DummyModelWithLayerNorm):
...
+        model_ref = DummyVAELikeModel(in_features=64, hidden_features=128, out_features=64)
+        model_ref.load_state_dict(model.state_dict(), strict=True)
+        model_ref.to(torch_device)
+
+        model.enable_group_offload(torch_device, offload_type="block_level", num_blocks_per_group=1, use_stream=True)
+
+        x = torch.randn(2, 64).to(torch_device)
+
+        with torch.no_grad():
+            for i in range(5):
```
sayakpaul: (nit): we can reduce the iteration number to 2.

tests/hooks/test_group_offloading.py
```diff
@@ -362,3 +502,161 @@ def apply_layer_output_tracker_hook(model: DummyModelWithLayerNorm):
...
+
+        model_ref = DummyVAELikeModel(in_features=64, hidden_features=128, out_features=64)
+        model_ref.load_state_dict(model.state_dict(), strict=True)
+        model_ref.to(torch_device)
+
+        model.enable_group_offload(torch_device, offload_type=offload_type, num_blocks_per_group=1, use_stream=True)
+
+        x = torch.randn(2, 64).to(torch_device)
+
+        with torch.no_grad():
```
sayakpaul, line 500: Why do test here without iterations like the tests below?

DN6:
Hi @rycerzes thank you for taking the time to put this together. 

I'm not sure this is the most ideal solution for the problem. The solution works for the top level modules in the VAEs, but if we apply this recursive fallback to transformer models like Flux, it would result in hooks for `context_embedder`, `x_embedder` etc. As opposed to having them be in a single unmatched group.
https://github.com/huggingface/diffusers/blob/6f1042e36cd588a7b66498f45c3bb7085e4fa395/src/diffusers/models/transformers/transformer_flux.py#L598-L608

Rather than silently falling back to leaf offloading (which isn't the behaviour expected), I would recommend a simpler approach that directly identifies blocks in the Autoencoder for offloading (since they are a bit of an edge case for block offloading)

I would set an attribute in the Autoencoder models `_group_offload_block_modules`

```python
class AutoencoderKL(...):
   _group_offload_block_modules = ["quant_conv", "post_quant_conv", "encoder", "decoder"]  
```

Then update `apply_group_offloading` to accept a `block_modules` argument

```python
def apply_group_offloading(
    module: torch.nn.Module,
    onload_device: Union[str, torch.device],
    offload_device: Union[str, torch.device] = torch.device("cpu"),
    offload_type: Union[str, GroupOffloadingType] = "block_level",
    num_blocks_per_group: Optional[int] = None,
    non_blocking: bool = False,
    use_stream: bool = False,
    record_stream: bool = False,
    block_modules: Optional[list] = None # accepts list of blocks to offload 
) -> None:
```

Set the modules in `enable_group_offload`

```python
        block_modules = block_modules or self._group_offload_block_modules if hasattr(self, "_group_offload_block_modules") else None
        apply_group_offloading(
            module=self,
            onload_device=onload_device,
            offload_device=offload_device,
            offload_type=offload_type,
            num_blocks_per_group=num_blocks_per_group,
            non_blocking=non_blocking,
            use_stream=use_stream,
            record_stream=record_stream,
            low_cpu_mem_usage=low_cpu_mem_usage,
            offload_to_disk_path=offload_to_disk_path,
            block_modules=block_modules
        )
```

Then in the actual group offloading step I think we can either offload the entire module if it's in `_group_offload_block_modules` (since AEs are typically small, this might be okay)  or  
recursively apply block offloading to the submodule. 

```python
    block_modules = set(config.block_modules) if config.block_modules is not None else set()
    for name, submodule in module.named_children():
        if name in block_modules:
             # offload entire submodule 
             # or recursively apply group offloading to submodule. 
            _apply_group_offloading_block_level(submodule, config)
```

tests/hooks/test_group_offloading.py
rycerzes: Will fix this in the next push.

tests/hooks/test_group_offloading.py
rycerzes: This specific test case was intended to verify a single forward pass works correctly without state carry-over issues, but in hindsight it's redundant if we have the loop tests. I'll clean this up.

tests/hooks/test_group_offloading.py
rycerzes: I'll switch to using AutoencoderKL with a small configuration (similar to what is used in test_models_autoencoder_kl.py)

tests/hooks/test_group_offloading.py
rycerzes: I'll update the tests to use a minimal configuration

src/diffusers/hooks/group_offloading.py
rycerzes: The intent here was that if we mark the container (e.g., encoder) as "having group offloading", the logic later in the file that gathers loose parameters (like those in GroupNorm which aren't Conv2d/Linear) would skip them, assuming they are already handled. Since encoder wasn't getting a hook itself in this implementation (only its children were), those parameters would have been left behind.

However, with the new approach suggested by @DN6, the entire container block will be explicitly offloaded, so this specific logic can be removed.

rycerzes:
Thanks for the detailed review and feedback @DN6 and @sayakpaul!

@DN6, I understand how the explicit opt-in approach is safer and cleaner to avoid unintended side effects on other models. I will refactor the implementation to look for a `_group_offload_block_modules` attribute in the model (or accept it as an argument) and use that to identify the blocks to offload, falling back to the current behavior if not present.

@sayakpaul, I will update the tests to use `AutoencoderKL` with a small configuration instead of the `DummyVAELikeModel` to ensure we are testing against the actual model structure while keeping the tests lightweight. I'm going to refactor the implementation based on @DN6's feedback to use an explicit opt-in mechanism instead of the recursive detection. This will likely change the code significantly, so some of the specific line comments might become moot, but I'll address the testing feedback (using small configs instead of dummy models, cleaning up iterations) in the new iteration.

I'm starting on these changes now.

rycerzes:
Hi @sayakpaul @DN6, the changes have been implemented 
• Switched to explicit opt-in via `_group_offload_block_modules`
• Updated `apply_group_offloading` to support `block_modules`
• Updated tests to use AutoencoderKL & adjusted coverage accordingly
• Refactored per review notes and cleaned up iteration logic

Please let me know if anything else is needed; happy to iterate further!

Cedric-Perauer:
Hi @rycerzes, there still seems to be an issue for KLWan VAE, when applying offloading like this :


```python
list(map(lambda module: apply_group_offloading(
    module,
    onload_device=onload_device,
    offload_device=offload_device,
    offload_type="block_level",
    use_stream=False,
    num_blocks_per_group = 2
), [pipe.text_encoder, pipe.transformer, pipe.vae]))
```

it seems to fail at ```conv_in``` of the encoder. 

``` bash
  File "/home/cedric/Downloads/diffusers/src/diffusers/models/autoencoders/autoencoder_kl_wan.py", line 592, in forward
    x = self.conv_in(x, feat_cache[idx])
  File "/home/cedric/anaconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
  File "/home/cedric/anaconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
  File "/home/cedric/Downloads/diffusers/src/diffusers/models/autoencoders/autoencoder_kl_wan.py", line 175, in forward
    return super().forward(x)
  File "/home/cedric/anaconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/conv.py", line 717, in forward
    return self._conv_forward(input, self.weight, self.bias)
  File "/home/cedric/anaconda3/envs/diffusers/lib/python3.10/site-packages/torch/nn/modules/conv.py", line 712, in _conv_forward
    return F.conv3d(
RuntimeError: Input type (torch.cuda.FloatTensor) and weight type (torch.FloatTensor) should be the same
```

It seems like KLWan VAE is not included in test cases as well. Can you confirm this issue ? Happy to help support this as well in next days if needed. Thank you !

rycerzes:
Hi @Cedric-Perauer, thanks for reporting this! I was able to reproduce the issue with the Wan VAE.

The problem occurs because when you call `apply_group_offloading()` directly without passing `block_modules`, the function doesn't know which submodules should be treated as blocks. The `encoder` and `decoder` in `AutoencoderKLWan` are not `ModuleList` or `Sequential` types, so they end up in the "unmatched modules" group instead of being properly offloaded as blocks. This causes the `encoder.conv_in` layer to stay on CPU while the input is on CUDA → device mismatch error.

As @DN6 requested, I am implementing an explicit opt-in approach where models define `_group_offload_block_modules` to specify which submodules should be treated as blocks. `AutoencoderKLWan` already has this defined: `_group_offload_block_modules = ["quant_conv", "post_quant_conv", "encoder", "decoder"]`.

**Two working solutions:**

1. **Use the high-level API** - it handles this automatically:
```python
pipe.vae.enable_group_offload(
    onload_device=onload_device,
    offload_device=offload_device,
    offload_type="block_level",
    num_blocks_per_group=2,
)
```

2. **Pass `block_modules` explicitly** when using `apply_group_offloading()` directly:
```python
block_modules = getattr(pipe.vae, "_group_offload_block_modules", None)

apply_group_offloading(
    pipe.vae,
    onload_device=onload_device,
    offload_device=offload_device,
    offload_type="block_level",
    num_blocks_per_group=2,
    block_modules=block_modules,
)
```

I've tested both approaches with the Wan VAE and confirmed they work. Let me know if you have any other questions!

>It seems like KLWan VAE is not included in test cases as well. Can you confirm this issue ? Happy to help support this as well in next days if needed. Thank you !

KLWan VAE is not explicitly tested in the group offloading test suite, but the same code paths are tested via `AutoencoderKL`. Do I need to implement explicit tests for KLWan VAE? @sayakpaul @DN6

src/diffusers/hooks/group_offloading.py
```diff
@@ -576,28 +582,123 @@ def _apply_group_offloading(module: torch.nn.Module, config: GroupOffloadingConf
...
     modules_with_group_offloading = set()
     unmatched_modules = []
     matched_module_groups = []
+
     for name, submodule in module.named_children():
-        if not isinstance(submodule, (torch.nn.ModuleList, torch.nn.Sequential)):
+        # Check if this is an explicitly defined block module
+        if name in block_modules:
+            # Apply block offloading to the specified submodule
+            _apply_block_offloading_to_submodule(
```
DN6: @rycerzes Can't we reuse `_apply_group_offloading_block_level` here? I don't think we need a new offloading function no?

src/diffusers/hooks/group_offloading.py
```diff
@@ -576,28 +582,123 @@ def _apply_group_offloading(module: torch.nn.Module, config: GroupOffloadingConf
...
+                submodule, name, config, modules_with_group_offloading, matched_module_groups
+            )
```
DN6:
```suggestion
            )
            modules_with_group_offloading.add(name)
```

DN6:
Looking good @rycerzes 👍🏽 just one comment.

rycerzes:
Thanks for the review @DN6 ! I've applied your suggestions:
1.  Replaced the custom helper with a recursive call to `_apply_group_offloading_block_level`.
2.  Added `modules_with_group_offloading.add(name)` to ensure the submodule is correctly marked as handled.

tests/hooks/test_group_offloading.py
```diff
@@ -362,3 +431,138 @@ def apply_layer_output_tracker_hook(model: DummyModelWithLayerNorm):
...
+
+        x = torch.randn(2, 3, 32, 32).to(torch_device)
+
+        with torch.no_grad():
+            for i in range(2):
+                out_ref = model_ref(x).sample
+                out = model(x).sample
+                self.assertTrue(torch.allclose(out_ref, out, atol=1e-5), f"Outputs do not match at iteration {i}.")
+
+    def test_nested_container_parameters_offloading(self):
```
sayakpaul, line 531: Can we reuse an existing modeling class from the library without using `DummyModelWithDeeplyNestedBlocks`? Not a merge-blocker, though.

sayakpaul (commented):
Thanks so much for your work! 

Not a merge-blocker, but it would be nice to apply this PR on some of the troubling models like Wan and see if we're getting expected outputs?

@asomoza, too since this fixes a couple of problems we have been facing related to group offloading.

sayakpaul:
@bot /style

DN6:
@rycerzes Is it okay for me to push some changes up to your branch to deal with the failing tests?

rycerzes:
Hi @DN6 and @sayakpaul!

> @rycerzes Is it okay for me to push some changes up to your branch to deal with the failing tests?

@DN6 - Yes, absolutely! Please feel free to push changes to fix the failing tests.

> Not a merge-blocker, but it would be nice to apply this PR on some of the troubling models like Wan and see if we're getting expected outputs?

@sayakpaul - Could you specify the models, I did run tests with Wan, do you want me to add to the test script?

> Can we reuse an existing modeling class from the library without using DummyModelWithDeeplyNestedBlocks? Not a merge-blocker, though.

Happy to do this if you have a specific model class in mind that would be a good substitute? Otherwise I can look for an appropriate one in the library.

Thanks for the reviews!

PR #12702, opened 2025-11-23
 Fix QwenImage txt_seq_lens handling

### LINKED ISSUE #12344: Qwen Image: txt_seq_lens is redundant and not used

kashif:
# What does this PR do?

- Removes the redundant `txt_seq_lens` plumbing from all QwenImage pipelines and modular steps; the transformer now infers text length from encoder inputs/masks and validates optional overrides.
- Builds a lightweight broadcastable attention mask from `encoder_hidden_states_mask` inside the double-stream attention, avoiding full `seq_len²` masks while keeping padding tokens masked.
- Adjusts QwenImage Transformer/ControlNet RoPE to take a single text length and documents the fallback behavior.
- Adds regression tests to ensure short `txt_seq_lens` values and encoder masks are handled safely.

Fixes #12344 

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
+++ b/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
@@ -663,6 +663,13 @@ class QwenImageEditPlusPipeline(DiffusionPipeline, QwenImageLoraLoaderMixin):
         else:
             batch_size = prompt_embeds.shape[0]
 
+        # QwenImageEditPlusPipeline does not currently support batch_size > 1
+        if batch_size > 1:
+            raise ValueError(
+                f"QwenImageEditPlusPipeline currently only supports batch_size=1, but received batch_size={batch_size}. "
+                "Please process prompts one at a time."
+            )
+
         device = self._execution_device
         # 3. Preprocess image
         if image is not None and not (isinstance(image, torch.Tensor) and image.size(1) == self.latent_channels):


dxqb:
just a few comments, not a full review:
 - there is some overlap with #12655
 - this code has the same issue mentioned in #12655 (expecting boolean semantics in a FloatTensor - but float attention masks are interpreted differently)
 - Could you clarify what the purpose of this PR is?
If the purpose is to remove the `txt_seq_lens` parameters, and infer the sequence lengths from the attention mask: why is it still a parameter of the transformer model?
If the purpose is towards passing sequence lengths to the attention dispatch (see https://github.com/huggingface/diffusers/issues/12344#issuecomment-3516800940), the sequence lengths for each batch sample must be inferred from the mask and passed to the transformer blocks, not only the max sequence length across all batch samples for RoPE

src/diffusers/models/controlnets/controlnet_qwenimage.py
```diff
@@ -244,7 +247,15 @@ def forward(
...
         temb = self.time_text_embed(timestep, hidden_states)
 
-        image_rotary_emb = self.pos_embed(img_shapes, txt_seq_lens, device=hidden_states.device)
+        batch_size, text_seq_len = encoder_hidden_states.shape[:2]
+        if txt_seq_lens is not None:
+            if len(txt_seq_lens) != batch_size:
+                raise ValueError(f"`txt_seq_lens` must have length {batch_size}, but got {len(txt_seq_lens)} instead.")
+            text_seq_len = max(text_seq_len, max(txt_seq_lens))
+        elif encoder_hidden_states_mask is not None:
+            text_seq_len = max(text_seq_len, int(encoder_hidden_states_mask.sum(dim=1).max().item()))
```
dxqb: This only works if the attention mask is in the form of [True, True, True, ..., False, False, False]. While this is the case in the most common use case of text attention masks, it doesn't have to be the case.

If the mask is [True, False, True, False, True, False], `self.pos_embed` receives an incorrect sequence length

kashif:
thanks @dxqb the idea was to remove `txt_seq_lens` all together and work with any mask pattern

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -570,7 +633,6 @@ def forward(
         encoder_hidden_states_mask: torch.Tensor = None,
         timestep: torch.LongTensor = None,
         img_shapes: Optional[List[Tuple[int, int, int]]] = None,
-        txt_seq_lens: Optional[List[int]] = None,
```
yiyixuxu: we should deprecate this argument

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -340,6 +402,7 @@ def __call__(
             is_causal=False,
             backend=self._attention_backend,
             parallel_config=self._parallel_config,
+            seq_lens=joint_seq_lens,
```
yiyixuxu: ohhh, instead of adding a new seq_lens argument to `dispatch_attnention_fn` do you think it'd makes sense to prepare a `joint_attn_mask` that can be normalized by `_normalize_attn_mask` later?
https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention_dispatch.py#L540C1-L589C21

yiyixuxu (commented):
thanks a ton for the PR! @kashif 
I left one question, let me know!

src/diffusers/models/transformers/transformer_qwenimage.py
kashif: checking

sayakpaul:
https://github.com/huggingface/diffusers/pull/12655 provides some benchmarks on the speed, as well. Possible to provide them here too? @kashif

kashif:
some benchmarks with various backends:

code:
[benchmark_backends_qwen.py](https://github.com/user-attachments/files/24030703/benchmark_backends_qwen.py)


backend_benchmark_complete

src/diffusers/models/attention_dispatch.py
```diff
@@ -1809,6 +1809,20 @@ def _native_attention(
...
+        # attn_mask is [batch_size, seq_len_k] boolean: True means attend, False means mask out
+        # SDPA expects [batch_size, 1, 1, seq_len_k] additive mask: 0.0 for attend, -inf for mask out
+        batch_size, seq_len_k = attn_mask.shape
+        # Ensure it's boolean for torch.where
+        if attn_mask.dtype != torch.bool:
+            attn_mask = attn_mask.bool()
+        # Convert boolean to additive: True -> 0.0, False -> -inf
+        attn_mask = torch.where(attn_mask, 0.0, float("-inf"))
+        # Convert to query dtype and reshape to [batch_size, 1, 1, seq_len_k] for broadcasting
+        attn_mask = attn_mask.to(dtype=query.dtype).view(batch_size, 1, 1, seq_len_k)
```
sayakpaul: Should we create a helper utility like `_prepare_attn_mask_native()` and move this logic there?

src/diffusers/models/attention_dispatch.py
```diff
@@ -2313,10 +2327,36 @@ def _xformers_attention(
         attn_mask = xops.LowerTriangularMask()
     elif attn_mask is not None:
         if attn_mask.ndim == 2:
-            attn_mask = attn_mask.view(attn_mask.size(0), 1, attn_mask.size(1), 1)
+            # Convert 2D boolean mask to 4D for xformers
```
sayakpaul: Same.

src/diffusers/models/attention_dispatch.py
```diff
@@ -2332,3 +2372,7 @@ def _xformers_attention(
         out = out.flatten(2, 3)
 
     return out
+
+
+# Initialize: download kernel for the initial backend set via DIFFUSERS_ATTN_BACKEND
```
sayakpaul: (nit): I think the method name itself is self-sufficient.

src/diffusers/models/controlnets/controlnet_qwenimage.py
```diff
@@ -131,12 +132,11 @@ def forward(
         encoder_hidden_states_mask: torch.Tensor = None,
         timestep: torch.LongTensor = None,
         img_shapes: Optional[List[Tuple[int, int, int]]] = None,
-        txt_seq_lens: Optional[List[int]] = None,
```
sayakpaul, line 134: Hmm, this will be a breaking change. Let's gracefully deprecate, instead?

src/diffusers/models/controlnets/controlnet_qwenimage.py
```diff
@@ -263,7 +273,6 @@ def forward(
...
```
sayakpaul, line 266: Same.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -197,21 +229,21 @@ def rope_params(self, index, dim, theta=10000):
     def forward(
         self,
         video_fhw: Union[Tuple[int, int, int], List[Tuple[int, int, int]]],
-        txt_seq_lens: List[int],
+        txt_seq_len: int,
```
sayakpaul: It's an internal module but not private. So, I see this as a breaking change.

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -91,6 +91,63 @@ def test_gradient_checkpointing_is_applied(self):
...
+
+        inputs["encoder_hidden_states_mask"] = encoder_hidden_states_mask
+
+        with torch.no_grad():
+            output = model(**inputs)
+
+        # The model should infer text_seq_len=2 from the mask for RoPE computation
+        self.assertEqual(output.sample.shape[1], inputs["hidden_states"].shape[1])
+
+    def test_builds_attention_mask_from_encoder_mask(self):
```
sayakpaul: The two tests are same. Is this expected?

sayakpaul (commented):
Ouff, really lovely stuff!

yiyixuxu:
cc @naykun can you take a look here too?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -570,7 +649,6 @@ def forward(
         encoder_hidden_states_mask: torch.Tensor = None,
         timestep: torch.LongTensor = None,
         img_shapes: Optional[List[Tuple[int, int, int]]] = None,
-        txt_seq_lens: Optional[List[int]] = None,
```
yiyixuxu, line 750: I think we need to deprecate it here too

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -197,21 +229,42 @@ def rope_params(self, index, dim, theta=10000):
     def forward(
         self,
         video_fhw: Union[Tuple[int, int, int], List[Tuple[int, int, int]]],
-        txt_seq_lens: List[int],
-        device: torch.device,
+        txt_seq_len: Optional[Union[int, torch.Tensor]] = None,
+        device: torch.device = None,
+        txt_seq_lens: Optional[List[int]] = None,
```
yiyixuxu:
```suggestion
        txt_seq_lens: Optional[List[int]] = None,
        device: torch.device = None,
        txt_seq_len: Optional[Union[int, torch.Tensor]] = None,
```

I moved the new argment to the last otherwise it would silently break if people call this function without using the positional argument. e.g. 
 with this code, the second input would now become text_seq_len, even though it was `text_seq_lens`
```py
pos_emb(image_shapes, [128, 64], device)
```

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -630,7 +722,7 @@ def forward(
             else self.time_text_embed(timestep, guidance, hidden_states)
         )
 
-        image_rotary_emb = self.pos_embed(img_shapes, txt_seq_lens, device=hidden_states.device)
+        image_rotary_emb = self.pos_embed(img_shapes, text_seq_len, device=hidden_states.device)
```
yiyixuxu:
```suggestion
        image_rotary_emb = self.pos_embed(img_shapes, text_seq_len=text_seq_len, device=hidden_states.device)
```

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -330,6 +382,33 @@ def __call__(
         joint_key = torch.cat([txt_key, img_key], dim=1)
         joint_value = torch.cat([txt_value, img_value], dim=1)
 
+        # If an encoder_hidden_states_mask is provided, create a joint attention mask.
+        # The encoder_hidden_states_mask is expected to have 1.0 for valid tokens and 0.0 for padding.
+        # We convert it to a boolean mask where True means "attend" and False means "mask out" (don't attend).
+        if encoder_hidden_states_mask is not None and attention_mask is None:
```
yiyixuxu: should we only do this when the `encoder_hidden_state_mask` contains paddings (or different length)? otherwise it might be less efficient for SDPA than `attention_mask == None`, no? did not test it out though so let me know what you think

kashif: yes you are right, i can add a check

yiyixuxu (approved):
thanks so much for helping us on this issue @kashif
changes looks good to me. 

I asked the qwen team to review it too so we will wait for their feedbacks now :)

cdutr:
Hey @kashif! I've prepared a documentation update with a new Performance section covering:

- Attention backend benchmarks (from your tests)
- torch.compile speedup (~2.4x)
- Variable-length prompt handling with CFG

I am also mentioning the gist with the scripts I used

Can you double check?

Is there anything missing, or something else I can help with?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -232,8 +285,7 @@ def forward(
             else:
                 max_vid_index = max(height, width, max_vid_index)
```
naykun, line 297: If we want to support batch inference in this PR, further modifications are needed. The current implementation assumes all instances have the same size, but in general cases—especially in editing scenarios—input images may vary in size. To handle this, we should compute max_vid_index per instance and slice txt_freqs accordingly.
Otherwise, a warning should be raised to alert users who are running batch inference.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -141,6 +141,38 @@ def apply_rotary_emb_qwen(
...
+
+    position_ids = torch.arange(text_seq_len, device=encoder_hidden_states.device, dtype=torch.long)
+    active_positions = torch.where(encoder_hidden_states_mask, position_ids, position_ids.new_zeros(()))
+    has_active = encoder_hidden_states_mask.any(dim=1)
+    per_sample_len = torch.where(has_active, active_positions.max(dim=1).values + 1, torch.as_tensor(text_seq_len))
+
+    # Keep as tensor to avoid graph breaks in torch.compile
+    # torch.maximum works with mixed tensor/scalar and keeps result as tensor
+    text_seq_len_tensor = torch.tensor(text_seq_len, device=encoder_hidden_states.device, dtype=torch.long)
+    rope_text_seq_len = torch.maximum(text_seq_len_tensor, per_sample_len.max())
```
naykun: Will rope_text_seq_len always be equal to text_seq_len_tensor?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -197,21 +229,42 @@ def rope_params(self, index, dim, theta=10000):
     def forward(
         self,
         video_fhw: Union[Tuple[int, int, int], List[Tuple[int, int, int]]],
-        txt_seq_lens: List[int],
-        device: torch.device,
+        txt_seq_lens: Optional[List[int]] = None,
+        device: torch.device = None,
+        txt_seq_len: Optional[Union[int, torch.Tensor]] = None,
```
naykun: This name could be confusing due to the existing txt_seq_lens parameter. Would max_txt_seq_len be clearer for txt_seq_len?

naykun:
Thank you so much for this excellent PR! It’s clean, well-structured, and addresses several long-standing issues. I’ve left a few questions —we can discuss them further.

src/diffusers/models/transformers/transformer_qwenimage.py
kashif: you are right

src/diffusers/models/transformers/transformer_qwenimage.py
kashif: for now i have added a warning

src/diffusers/models/transformers/transformer_qwenimage.py
kashif: changed

docs/source/en/api/pipelines/qwenimage.md
```diff
@@ -108,12 +108,68 @@ pipe = QwenImageEditPlusPipeline.from_pretrained(
...
+
+pipe = QwenImagePipeline.from_pretrained("Qwen/Qwen-Image", torch_dtype=torch.bfloat16).to("cuda")
+pipe.transformer = torch.compile(pipe.transformer)
+
+# First call triggers compilation (~7s overhead on A100)
+# Subsequent calls see ~2.4x speedup
+image = pipe("a cat", num_inference_steps=50).images[0]
+```
+
+### Batched Inference with Variable-Length Prompts
```
sayakpaul, line 135: Instead of specifying performance numbers on `torch.compile` and other attention backends, maybe we could highlight this point and include with and without `torch.compile` numbers? @cdutr WDYT?

src/diffusers/models/attention_dispatch.py
```diff
@@ -1881,6 +1881,39 @@ def score_mod(score, batch_idx, head_idx, q_idx, kv_idx):
     return out
 
 
+def _prepare_attn_mask_native(
```
sayakpaul: Do we not have to do this for other attention backends? @cdutr

src/diffusers/models/attention_dispatch.py
```diff
@@ -2442,3 +2507,6 @@ def _xformers_attention(
         out = out.flatten(2, 3)
 
     return out
+
+
+_maybe_download_kernel_for_backend(_AttentionBackendRegistry._active_backend)
```
sayakpaul: Why do we need this?

src/diffusers/models/controlnets/controlnet_qwenimage.py
```diff
@@ -147,28 +148,44 @@ def forward(
...
+                "version 0.37.0. The text sequence length is now automatically inferred from `encoder_hidden_states` "
+                "and `encoder_hidden_states_mask`.",
+                standard_warn=False,
+            )
+
         if joint_attention_kwargs is not None:
             joint_attention_kwargs = joint_attention_kwargs.copy()
             lora_scale = joint_attention_kwargs.pop("scale", 1.0)
         else:
+            joint_attention_kwargs = {}
```
sayakpaul: Seems like an unrelated change?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -304,14 +367,39 @@ def rope_params(self, index, dim, theta=10000):
...
-        if self.pos_freqs.device != device:
-            self.pos_freqs = self.pos_freqs.to(device)
-            self.neg_freqs = self.neg_freqs.to(device)
+        # Move to device unconditionally to avoid graph breaks in torch.compile
+        self.pos_freqs = self.pos_freqs.to(device)
+        self.neg_freqs = self.neg_freqs.to(device)
+
+        # Validate batch inference with variable-sized images
+        # In Layer3DRope, the outer list represents batch, inner list/tuple represents layers
+        if isinstance(video_fhw, list) and len(video_fhw) > 1:
```
sayakpaul, line 386: Cc: @naykun good for you?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -778,10 +906,20 @@ def forward(
...
+                "Passing `txt_seq_lens` is deprecated and will be removed in version 0.37.0. "
+                "Please use `txt_seq_len` instead (singular, not plural). "
+                "The new parameter accepts a single int or tensor value instead of a list.",
+                standard_warn=False,
+            )
         if attention_kwargs is not None:
             attention_kwargs = attention_kwargs.copy()
             lora_scale = attention_kwargs.pop("scale", 1.0)
         else:
+            attention_kwargs = {}
```
sayakpaul: Unrelated?

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -101,6 +181,5 @@ def prepare_init_args_and_inputs_for_common(self):
     def prepare_dummy_input(self, height, width):
         return QwenImageTransformerTests().prepare_dummy_input(height=height, width=width)
 
-    @pytest.mark.xfail(condition=True, reason="RoPE needs to be revisited.", strict=True)
```
sayakpaul, line 104: Does it pass now? 👀

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -91,6 +90,87 @@ def test_gradient_checkpointing_is_applied(self):
...
+
+        # Test 4: No mask provided (None case)
+        rope_text_seq_len_none, per_sample_len_none, normalized_mask_none = compute_text_seq_len_from_mask(
+            inputs["encoder_hidden_states"], None
+        )
+        self.assertEqual(rope_text_seq_len_none, inputs["encoder_hidden_states"].shape[1])
+        self.assertIsNone(per_sample_len_none)
+        self.assertIsNone(normalized_mask_none)
+
+    def test_non_contiguous_attention_mask(self):
```
sayakpaul, line 148: Could we also test that the deprecation warning is raised when the deprecated args passed?

tests/models/transformers/test_models_transformer_qwenimage.py
kashif: yes it passes

sayakpaul: Damn! Wow!

sayakpaul: @kashif which versions are you using? I am getting:

```bash
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::QwenImageTransformerCompileTests::test_compile_works_with_aot - torch.fx.experimental.symbolic_shapes.GuardOnDataDependentSymNode: Could not guard on data-dependent expression Eq(u0, 1) (unhinted: E...
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::QwenImageTransformerCompileTests::test_torch_compile_recompilation_and_graph_break - torch._dynamo.exc.Unsupported: Dynamic slicing with Tensor arguments
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::QwenImageTransformerCompileTests::test_torch_compile_repeated_blocks - torch._dynamo.exc.Unsupported: Data-dependent branching
```

With:

```bash
RUN_SLOW=1 RUN_COMPILE=yes pytest tests/models/transformers/test_models_transformer_qwenimage.py::QwenImageTransformerCompileTests
```

Would it be possible to fix the compile compatibility? With `main`, it runs as expected:

```
==================================== 3 passed, 1 skipped, 1 xfailed, 5 warnings in 109.93s (0:01:49) =====================================
```

docs/source/en/api/pipelines/qwenimage.md
cdutr: Good point! I've simplified the Performance section to focus on torch.compile with the before/after numbers, 
removed the attention backend tables since the differences between backends are minimal compared to the torch.compile gains

src/diffusers/models/attention_dispatch.py
cdutr: As I understand it, xformers already uses this helper. Flash Attention, Flash Varlen variants, and Flex Attention either don't support arbitrary masks or handle them through different APIs, so no changes are needed there. That said, the name `_prepare_attn_mask_native` is a bit misleading since it's also used by xformers. Maybe we should rename it to something like `_prepare_additive_attn_mask `or `_bool_to_additive_mask`? WDYT @kashif?

tests/models/transformers/test_models_transformer_qwenimage.py
cdutr: @kashif, I can take this on and implement a `test_deprecated_txt_seq_lens_warning` for it

I noticed that the deprecation version is currently set to 0.37.0, while the current diffusers version is 0.37.0.dev0. Because of this, the deprecation utility treats the deprecation window as expired and raises a ValueError instead of a FutureWarning (since version >= deprecation_version)

I will update the deprecation version to 0.38.0, ok?

kashif: ah no i think it needs to be 0.39?

src/diffusers/models/attention_dispatch.py
kashif: yes good idea renaming to `_prepare_additive_attn_mask`

src/diffusers/models/attention_dispatch.py
sayakpaul: Cc: @kashif

tests/models/transformers/test_models_transformer_qwenimage.py
sayakpaul: Compilation issues still need to be resolved I believe.

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -91,6 +90,174 @@ def test_gradient_checkpointing_is_applied(self):
...
+        """Test QwenImageTransformer2DModel with use_layer3d_rope=True (layered model)."""
+        # Create layered model config
+        init_dict = {
+            "patch_size": 2,
+            "in_channels": 16,
+            "out_channels": 16,
+            "num_layers": 2,
+            "attention_head_dim": 128,
+            "num_attention_heads": 4,
+            "joint_attention_dim": 16,
```
sayakpaul, line 212: Can we use smaller values please? @cdutr 

I also think we should `use_additional_t_cond=True` here?

tests/models/transformers/test_models_transformer_qwenimage.py
kashif:
```
RUN_SLOW=1 RUN_COMPILE=yes pytest tests/models/transformers/test_models_transformer_qwenimage.py::QwenImageTransformerCompileTests
...
============================================================ 4 passed, 1 skipped, 12 warnings in 45.10s =============================================================
```

tests/models/transformers/test_models_transformer_qwenimage.py
cdutr: @kashif already addressed it

tests/models/transformers/test_models_transformer_qwenimage.py
sayakpaul: Yes passing for me as well. What was the major change?

kashif: had to fix device placements and data dependent ifs

src/diffusers/models/attention_dispatch.py
kashif: removed

src/diffusers/models/attention_dispatch.py
```diff
@@ -1900,6 +1936,14 @@ def _native_attention(
 ) -> torch.Tensor:
     if return_lse:
         raise ValueError("Native attention backend does not support setting `return_lse=True`.")
+
+    # Convert 2D boolean mask to 4D additive mask for SDPA
+    if attn_mask is not None and attn_mask.ndim == 2:
+        # attn_mask is [batch_size, seq_len_k] boolean: True means attend, False means mask out
```
dxqb: it should not be assumed that the attention mask is boolean in a generic attention dispatcher.
this might be okay for the Qwen pipeline, but other models might use actual float "additive" attention masks. here, all masks that are 2-d are converted to bool

it's also not necessary to convert boolean attention masks to "additive". torch SDPA accepts both, as long as they have the correct dtype, see https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html

src/diffusers/models/controlnets/controlnet_qwenimage.py
dxqb: I think this is resolved, but I let you close it if you agree

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -454,6 +541,35 @@ def __call__(
         joint_key = torch.cat([txt_key, img_key], dim=1)
         joint_value = torch.cat([txt_value, img_value], dim=1)
 
+        # If an encoder_hidden_states_mask is provided, create a joint attention mask.
+        # The encoder_hidden_states_mask is expected to have 1.0 for valid tokens and 0.0 for padding.
+        # We convert it to a boolean mask where True means "attend" and False means "mask out" (don't attend).
+        # Only create the mask if there's actual padding, otherwise keep attention_mask=None for better SDPA performance.
```
dxqb: I don't think this comment is anywhere executed in code. There is always an attention mask used, even if it's all-True.
if torch.all(mask), a None-mask should be used. I've tested this in a finetuning context and it makes a big difference because of how much faster SDPA is without a mask (~ 30% end-to-end faster)

kashif:
thanks @dxqb can you test now?

dxqb:
> thanks @dxqb can you test now?

sure, will do.
could you have a look at this PR please? https://github.com/huggingface/diffusers/pull/12870
It is based on yours. I've made minor changes to your code:

kashif:
@dxqb you can also send a PR to this branch or fix this PR directly

dxqb:
> @dxqb you can also send a PR to this branch or fix this PR directly

I'll send the small changes in your PR that I needed (like support for text sequence length per sample), but keep the PR overall separate - it's a separate feature

yiyixuxu:
@kashif @dxqb 
is this ready for final round of review?

src/diffusers/models/controlnets/controlnet_qwenimage.py
```diff
@@ -281,7 +311,6 @@ def forward(
                     encoder_hidden_states_mask=encoder_hidden_states_mask,
                     timestep=timestep,
                     img_shapes=img_shapes,
-                    txt_seq_lens=txt_seq_lens,
```
yiyixuxu, line 284: should we also build joint mask for controlnet?
not merge-blocking though, I don't think it breaks anything

src/diffusers/models/attention_dispatch.py
```diff
@@ -1900,6 +1937,14 @@ def _native_attention(
 ) -> torch.Tensor:
     if return_lse:
         raise ValueError("Native attention backend does not support setting `return_lse=True`.")
+
+    # Reshape 2D mask to 4D for SDPA
+    # SDPA accepts both boolean masks (torch.bool) and additive masks (float)
+    if attn_mask is not None and attn_mask.ndim == 2:
```
yiyixuxu: I think a `[seq_len_q, seq_len_k]` is also valid 2D mask SDPA accpets, this will cause error, no? 
```suggestion
    if attn_mask is not None and attn_mask.ndim == 2 and attn_mask.shape[0] == query.shape[0] and attn_mask.shape[1] == key.shape[1]:
```

src/diffusers/models/controlnets/controlnet_qwenimage.py
kashif: good catch i think we should add it

naykun:
Hi @yiyixuxu and @kashif,  

Quick update: this PR passed our internal regression tests with `batch_size=1`. Heads up though — the EditPlus pipeline still doesn’t support `batch_size > 1` yet.  

Since this is a large PR with valuable fixes, I suggest we merge it now and handle batch>1 support separately. What are your thoughts?  

Below is a minimal reproduction script for the EditPlus pipeline error when `batch_size > 1`:

```python
import os
import torch
from diffusers import QwenImageEditPlusPipeline
from PIL import Image
torch.manual_seed(0)

pipeline = QwenImageEditPlusPipeline.from_pretrained("Qwen/Qwen-Image-Edit-2511")
print("pipeline loaded")
pipeline.to(torch.bfloat16)
pipeline.to('cuda')
pipeline.set_progress_bar_config(disable=None)

image0 = Image.open("input1.png").convert("RGB")
image1 = Image.open("input2.png").convert("RGB")
prompt = "merge picture 1 with picture 2"
inputs = {
    "image": [[image0, image1], [image0, image1]],
    "prompt": [prompt, prompt],
    "height": 1024,
    "width": 1024,
    "true_cfg_scale": 4.0,
    "negative_prompt": " ",
    "num_inference_steps": 50,
    "guidance_scale": 1.0,
    "num_images_per_prompt": 1,
}
with torch.inference_mode():
    output = pipeline(**inputs)
    output_image = output.images[0]
    output_image.save("output_image_edit_plus.png")
    print("image saved at", os.path.abspath("output_image_edit_plus.png"))
```

sayakpaul:
Thanks a lot @naykun! WDYT about raising an error from the EditPlus pipeline when batch_size > 1? Would that work for you?

naykun:
> Thanks a lot @naykun! WDYT about raising an error from the EditPlus pipeline when batch_size > 1? Would that work for you?

Sounds good to me! 😊

sayakpaul:
Uff feels like a big burden off everyone's shoulders! Thanks a lot @cdutr @kashif for jamming. @cdutr please obtain your MVP certificate from [here](https://huggingface.co/spaces/diffusers/generate-mvp-certificate). Let us also know about your HF username, so that we can grant you some credits :)

DefTruth:
@sayakpaul @yiyixuxu Have you tested Context Parallel? It seems that this PR has caused an error in Context Parallel. 

```bash
[rank3]:     output = function_reference.forward(*args, **kwargs)
[rank3]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/workspace/dev/vipshop/diffusers/src/diffusers/models/transformers/transformer_qwenimage.py", line 959, in forward
[rank3]:     encoder_hidden_states, hidden_states = block(
[rank3]:                                            ^^^^^^
[rank3]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
[rank3]:     return self._call_impl(*args, **kwargs)
[rank3]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1786, in _call_impl
[rank3]:     return forward_call(*args, **kwargs)
[rank3]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/workspace/dev/vipshop/diffusers/src/diffusers/models/transformers/transformer_qwenimage.py", line 696, in forward
[rank3]:     attn_output = self.attn(
[rank3]:                   ^^^^^^^^^^
[rank3]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
[rank3]:     return self._call_impl(*args, **kwargs)
[rank3]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1786, in _call_impl
[rank3]:     return forward_call(*args, **kwargs)
[rank3]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/workspace/dev/vipshop/diffusers/src/diffusers/models/attention_processor.py", line 605, in forward
[rank3]:     return self.processor(
[rank3]:            ^^^^^^^^^^^^^^^
[rank3]:   File "/workspace/dev/vipshop/diffusers/src/diffusers/models/transformers/transformer_qwenimage.py", line 535, in __call__
[rank3]:     txt_query = apply_rotary_emb_qwen(txt_query, txt_freqs, use_real=False)
[rank3]:                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/workspace/dev/vipshop/diffusers/src/diffusers/models/transformers/transformer_qwenimage.py", line 140, in apply_rotary_emb_qwen
[rank3]:     x_out = torch.view_as_real(x_rotated * freqs_cis).flatten(3)
[rank3]:                                ~~~~~~~~~~^~~~~~~~~~~
[rank3]: RuntimeError: The size of tensor a (109) must match the size of tensor b (27) at non-singleton dimension 1
```
I have checkout to b86bd99ea (the last commit before this pr) and the CP is work.

sayakpaul:
Was it working before in the CP setup? Cc: @kashif

DefTruth:
@sayakpaul @kashif The txt_seq_len is error if context parallel enabled, can we fix it.  

https://github.com/huggingface/diffusers/blob/9d68742214b9491718487b63f25d6e8236ed56ea/src/diffusers/models/transformers/transformer_qwenimage.py#L145-L152

sayakpaul:
Let's please not sound condescending here. I am politely asking if it was working as expected for Qwen before this PR. Could you confirm that? 

If you have a fix, then we appreciate any PR otherwise, I kindly ask you to have patience. It will surely be fixed.

DefTruth:
@sayakpaul Hi~

1. I'm very sorry. I didn't mean to offend. I think I was just describing the problem I found. I'm trying to solve this problem. If it works, I will submit a PR for the fix.  I  really like your works, my friends ~ 

2. I have checkout to https://github.com/huggingface/diffusers/commit/b86bd99eacbee91062fe4ef63f7aa652ca5b2c1a (the last commit before this pr) and the CP is work.

DefTruth:
@sayakpaul Change the CP plan from transformer-level to block-level seems can avoid sequence length mismatch in rope. I will submit a pr to fix this issue after I have conducted thorough tests.
```python
        _cp_plan = {
            "transformer_blocks.0": {
                "hidden_states": ContextParallelInput(
                    split_dim=1, expected_dims=3, split_output=False
                ),
                "encoder_hidden_states": ContextParallelInput(
                    split_dim=1, expected_dims=3, split_output=False
                ),
            },
            "pos_embed": {
                0: ContextParallelInput(split_dim=0, expected_dims=2, split_output=True),
                1: ContextParallelInput(split_dim=0, expected_dims=2, split_output=True),
            },
            "proj_out": ContextParallelOutput(gather_dim=1, expected_dims=3),
        }
```

DefTruth:
@sayakpaul I have make a draft pr at https://github.com/huggingface/diffusers/pull/12970, I'm sorry again for my straightforward comments. I really didn't mean to offend. I've just been paying more attention to diffusers' CP's ability. As you can see, I have also made a lot of efforts to fix the bugs related to CP. I truly hope that diffusers will keep getting better and better, because I am also a loyal user of it and have benefited a lot from its code.

GengarWu:
> some benchmarks with various backends:
> 
> code: [benchmark_backends_qwen.py](https://github.com/user-attachments/files/24030703/benchmark_backends_qwen.py)
> 
> backend_benchmark_complete

Have any of you run into this issue: when running qwen-image-edit-2511 at 1024×1024 resolution on H20, the single-step inference speed dropped from 5.7s in version 0.36.0 to 7.8s? @kashif 
```
import os
import torch
from PIL import Image
from diffusers import QwenImageEditPlusPipeline

pipeline = QwenImageEditPlusPipeline.from_pretrained("/data00/models/Qwen-Image-Edit-2511", torch_dtype=torch.bfloat16)
print("pipeline loaded")

pipeline.to('cuda')
pipeline.set_progress_bar_config(disable=None)
image1 = Image.open("/data00/image_examples/qwen-image-edit-input1.jpg")
image2 = Image.open("/data00/image_examples/qwen-image-edit-input2.jpg")
prompt = "The magician bear is on the left, the alchemist bear is on the right, facing each other in the central park square."
inputs = {
    "image": [image1, image2],
    "prompt": prompt,
    "generator": torch.Generator(device="cuda").manual_seed(0),
    "true_cfg_scale": 4.0,
    "negative_prompt": " ",
    "num_inference_steps": 40,
    "guidance_scale": 1.0,
    "num_images_per_prompt": 1,
    "height": 1024,
    "width": 1024,
}
with torch.inference_mode():
    output = pipeline(**inputs)
    output_image = output.images[0]
    output_image.save("output_image_edit_2511.png")
    print("image saved at", os.path.abspath("output_image_edit_2511.png"))
```

kashif:
@GengarWu would you be able to test with the PR https://github.com/huggingface/diffusers/pull/12987  Thanks!

JoeGaffney:
Hey, gettting errors now i think related to this PR.

image

image

I am using the nunchaku transformer.

``` python
@decorator_global_pipeline_cache
def get_edit_pipeline(model_id) -> QwenImageEditPlusPipeline:
    num_inference_steps = 8  # you can also use the 8-step model to improve the quality
    rank = 128  # you can also use the rank=128 model to improve the quality
    model_path = f"nunchaku-tech/nunchaku-qwen-image-edit-2509/svdq-{get_precision()}_r{rank}-qwen-image-edit-2509-lightningv2.0-{num_inference_steps}steps.safetensors"
    transformer = NunchakuQwenImageTransformer2DModel.from_pretrained(model_path)

    pipe = QwenImageEditPlusPipeline.from_pretrained(
        model_id,
        text_encoder=get_qwen2_5_text_encoder(),
        transformer=transformer,
        scheduler=get_scheduler(),
        torch_dtype=torch.bfloat16,
    )

    return optimize_pipeline(pipe, offload=is_memory_exceeded(23))
```

Is it something to mention to them?

JoeGaffney:
I was able to monkey patch to get the same look and not crash. This is probably sub optimal should i put a ticket in the nunchaku repo and link here?

```python 
import diffusers.models.transformers.transformer_qwenimage as qwen_image_module
import torch
from diffusers.models.transformers.transformer_qwenimage import QwenEmbedRope


def apply_qwen_image_patches():
    """
    Monkey patch for nunchaku compatibility with diffusers.
    See: https://github.com/huggingface/diffusers/pull/12702
    """

    original_qwen_rope_forward = QwenEmbedRope.forward

    def patched_qwen_rope_forward(self, video_fhw, txt_seq_lens=None, device=None, max_txt_seq_len=None):
        # Use a safe large default (e.g. 4096) to ensure we generate enough frequencies.
        # We will slice them down in apply_rotary_emb_qwen.
        if max_txt_seq_len is None and txt_seq_lens is None:
            max_txt_seq_len = 4096
        return original_qwen_rope_forward(self, video_fhw, txt_seq_lens, device, max_txt_seq_len)  # type: ignore

    QwenEmbedRope.forward = patched_qwen_rope_forward

    # Also patch apply_rotary_emb_qwen to handle shape mismatch via slicing
    original_apply_rotary_emb_qwen = qwen_image_module.apply_rotary_emb_qwen

    def patched_apply_rotary_emb_qwen(x, freqs_cis, use_real=True, use_real_unbind_dim=-1):
        # x: [Batch, Seq, Heads, Dim]
        # freqs_cis: Tensor or Tuple
        seq_len = x.shape[1]

        # Handle the specific case crashing in nunchaku/diffusers interaction (use_real=False, tensor freqs)
        if not use_real and isinstance(freqs_cis, torch.Tensor):
            # freqs_cis typically [1, MaxSeq, Dim] or [MaxSeq, Dim]
            if freqs_cis.dim() == 3 and freqs_cis.shape[1] > seq_len:
                freqs_cis = freqs_cis[:, :seq_len]
            elif freqs_cis.dim() == 2 and freqs_cis.shape[0] > seq_len:
                freqs_cis = freqs_cis[:seq_len]

        return original_apply_rotary_emb_qwen(x, freqs_cis, use_real, use_real_unbind_dim)

    qwen_image_module.apply_rotary_emb_qwen = patched_apply_rotary_emb_qwen
```

DefTruth:
> > some benchmarks with various backends:
> > code: [benchmark_backends_qwen.py](https://github.com/user-attachments/files/24030703/benchmark_backends_qwen.py)
> > backend_benchmark_complete
> 
> Have any of you run into this issue: when running qwen-image-edit-2511 at 1024×1024 resolution on H20, the single-step inference speed dropped from 5.7s in version 0.36.0 to 7.8s? @kashif
> 
> ```
> import os
> import torch
> from PIL import Image
> from diffusers import QwenImageEditPlusPipeline
> 
> pipeline = QwenImageEditPlusPipeline.from_pretrained("/data00/models/Qwen-Image-Edit-2511", torch_dtype=torch.bfloat16)
> print("pipeline loaded")
> 
> pipeline.to('cuda')
> pipeline.set_progress_bar_config(disable=None)
> image1 = Image.open("/data00/image_examples/qwen-image-edit-input1.jpg")
> image2 = Image.open("/data00/image_examples/qwen-image-edit-input2.jpg")
> prompt = "The magician bear is on the left, the alchemist bear is on the right, facing each other in the central park square."
> inputs = {
>     "image": [image1, image2],
>     "prompt": prompt,
>     "generator": torch.Generator(device="cuda").manual_seed(0),
>     "true_cfg_scale": 4.0,
>     "negative_prompt": " ",
>     "num_inference_steps": 40,
>     "guidance_scale": 1.0,
>     "num_images_per_prompt": 1,
>     "height": 1024,
>     "width": 1024,
> }
> with torch.inference_mode():
>     output = pipeline(**inputs)
>     output_image = output.images[0]
>     output_image.save("output_image_edit_2511.png")
>     print("image saved at", os.path.abspath("output_image_edit_2511.png"))
> ```

I have also encounter the same performance problem

asomoza:
@DefTruth did you test it with the [PR](https://github.com/huggingface/diffusers/pull/12987) mentioned before?

DefTruth:
> @DefTruth did you test it with the [PR](https://github.com/huggingface/diffusers/pull/12987) mentioned before?

oh~ let me have a try

DefTruth:
> @DefTruth did you test it with the [PR](https://github.com/huggingface/diffusers/pull/12987) mentioned before?

I tested this [PR](https://github.com/huggingface/diffusers/pull/12987)  on L20, and there was an 11% performance improvement in my test

PR #12703, opened 2025-11-23
Add Support for Z-Image Series

JerryWu-code:
# What does this PR do?
This PR introduces **Z-Image** Series into the diffusers library. Z-Image is a powerful and highly efficient **6B**-parameter image generation model that is friendly for consumer-grade hardware, with strong capabilities in photorealistic image generation, accurate rendering of both complex Chinese and English text, and robust adherence to bilingual instructions. The technical report and Z-Image-Turbo checkpoint will be released very soon.

Thanks for the support of @yiyixuxu @apolinario.

### DIFF
--- a/src/diffusers/models/transformers/transformer_z_image.py
+++ b/src/diffusers/models/transformers/transformer_z_image.py
@@ -20,12 +20,11 @@ import torch.nn as nn
 import torch.nn.functional as F
 from torch.nn.utils.rnn import pad_sequence
 
-from diffusers.models.normalization import RMSNorm
-
 from ...configuration_utils import ConfigMixin, register_to_config
 from ...loaders import FromOriginalModelMixin, PeftAdapterMixin
 from ...models.attention_processor import Attention
 from ...models.modeling_utils import ModelMixin
+from ...models.normalization import RMSNorm
 from ...utils.torch_utils import maybe_allow_in_graph
 from ..attention_dispatch import dispatch_attention_fn
 


src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,814 @@
...
+            nn.Linear(
+                mid_size,
+                out_size,
+                bias=True,
+            ),
+        )
+        nn.init.normal_(self.mlp[0].weight, std=0.02)
+        nn.init.zeros_(self.mlp[0].bias)
+        nn.init.normal_(self.mlp[2].weight, std=0.02)
+        nn.init.zeros_(self.mlp[2].bias)
```
yiyixuxu:
```suggestion
```

src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,814 @@
...
+import torch.nn.functional as F
+from einops import rearrange
+
+
+try:
+    from flash_attn import flash_attn_varlen_func
+except ImportError:
+    flash_attn_varlen_func = None
+
+try:
```
yiyixuxu: can we add a `is_apex_available` similar to https://github.com/huggingface/diffusers/blob/main/src/diffusers/utils/import_utils.py#L407

src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,814 @@
...
+            mask = torch.zeros((batch_size, max_seqlen), dtype=torch.bool, device=query.device)
+            for i, l in enumerate(seqlens):
+                mask[i, :l] = True
+
+            attn_mask = torch.zeros((batch_size, 1, 1, max_seqlen), dtype=query.dtype, device=query.device)
+            attn_mask.masked_fill_(~mask[:, None, None, :], torch.finfo(query.dtype).min)
+
+            q_padded = q_padded.transpose(1, 2)
+            k_padded = k_padded.transpose(1, 2)
+            v_padded = v_padded.transpose(1, 2)
```
yiyixuxu: is it possible we focus on SDPA and implemeent it with `dispatch_attention_fn` https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention_dispatch.py#L291

user can then switch to `flash_attn_varlen_func` using 

```
with attention_backend("flash_varlen"):
    image = pipeline(...)
```
see the implmenetation for flash_varlen backend here
https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention_dispatch.py#L1353

our documentation on attention backend here https://huggingface.co/docs/diffusers/main/en/optimization/attention_backends

src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,814 @@
...
+            x_max_item_seqlen,
+            scale_gate_msa,
+            x_src_ids_shard,
+        )
+
+        x_shard = self.ffn_forward(x_shard, scale_gate_mlp, x_src_ids_shard)
+
+        return x_shard
+
+    def attn_forward(
```
yiyixuxu: can we remove this method and move the code back to `forward` method?
let's make sure all the computation are inside `forward` here

src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,814 @@
...
+            attn_out = self.attention(
+                self.attention_norm1(x_shard),
+                image_rotary_emb=x_freqs_cis_shard,
+                x_cu_seqlens=x_cu_seqlens,
+                x_max_item_seqlen=x_max_item_seqlen,
+            )
+            x_shard = x_shard + self.attention_norm2(attn_out)
+        return x_shard
+
+    def ffn_forward(
```
yiyixuxu: same here - let's remove this method and keep all the computation in `forward`

src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,814 @@
...
+        self.t_scale = t_scale
+
+        assert len(all_patch_size) == len(all_f_patch_size)
+
+        all_x_embedder = {}
+        all_final_layer = {}
+        for patch_idx, (patch_size, f_patch_size) in enumerate(zip(all_patch_size, all_f_patch_size)):
+            x_embedder = nn.Linear(f_patch_size * patch_size * patch_size * in_channels, dim, bias=True)
+            nn.init.xavier_uniform_(x_embedder.weight)
+            nn.init.constant_(x_embedder.bias, 0.0)
```
yiyixuxu: we don't need to include initialization logic - itls for inference only

src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,814 @@
...
+        )
+        unified = unified_shard.split(unified_item_seqlens, dim=0)
+        x = [unified[i][cap_item_seqlens[i] :] for i in range(bsz)]
+        assert all(len(x[i]) == x_item_seqlens[i] for i in range(bsz))
+
+        x = self.unpatchify(x, x_size, patch_size, f_patch_size)
+
+        return x, {}
+
+    def parameter_count(self) -> int:
```
yiyixuxu: can we remove this?

src/diffusers/pipelines/z_image/pipeline_z_image.py
```diff
@@ -0,0 +1,628 @@
...
+        processing larger images.
+        """
+        self.vae.enable_tiling()
+
+    def disable_vae_tiling(self):
+        r"""
+        Disable tiled VAE decoding. If `enable_vae_tiling` was previously enabled, this method will go back to
+        computing decoding in one step.
+        """
+        self.vae.disable_tiling()
```
yiyixuxu:
```suggestion
```
we deprecated these methods now

src/diffusers/pipelines/z_image/pipeline_z_image.py
```diff
@@ -0,0 +1,628 @@
...
+            assert isinstance(config[0], (float, int, str))
+            result = config[0]
+            for thresh, val in config[1:]:
+                if t >= thresh:
+                    result = val
+                else:
+                    break
+            return result
+        else:
+            raise ValueError(f"invalid time-aware config {config} of type {type(config)}")
```
yiyixuxu:
```suggestion
```
we only used this method once in the pipeline, and it comes in a very specific format, I think we can remove this method and simplify it a bit insidee the pipeline

```py
current_guidance_scale = self.value_from_time_aware_config(
                        (self.guidance_scale, (self._cfg_truncation, 0.0)), t_norm
                    )
```

src/diffusers/pipelines/z_image/pipeline_z_image.py
```diff
@@ -0,0 +1,628 @@
...
+                # Handle cfg truncation
+                current_guidance_scale = self.guidance_scale
+                if (
+                    self.do_classifier_free_guidance
+                    and self._cfg_truncation is not None
+                    and float(self._cfg_truncation) <= 1
+                ):
+                    current_guidance_scale = self.value_from_time_aware_config(
+                        (self.guidance_scale, (self._cfg_truncation, 0.0)), t_norm
+                    )
```
yiyixuxu:
```suggestion
                if (
                    self.do_classifier_free_guidance
                    and self._cfg_truncation is not None
                    and float(self._cfg_truncation) <= 1 
                    and t_norm >= self._cfg_truncation
                ):
                    current_guidance_scale =  0.0
                    )
```

yiyixuxu (commented):
thanks so much for the PR!
I left some feedbacks, I think the most important change is the attention_backend, we should be able to refactor using `dispatch_attention_fn` and it work out of box with both naive and flash_varlen, insteand of manually handle both

yiyixuxu:
@bot /style

src/diffusers/models/attention_dispatch.py
```diff
@@ -621,22 +653,42 @@ def _wrapped_flash_attn_3(
 ) -> Tuple[torch.Tensor, torch.Tensor]:
     # Hardcoded for now because pytorch does not support tuple/int type hints
     window_size = (-1, -1)
-    out, lse, *_ = flash_attn_3_func(
+    max_seqlen_q = q.shape[2]
```
yiyixuxu: ohh, what's the reason to use `_flash_attn_forward` instead here?

src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,652 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+from typing import List, Optional, Tuple
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from einops import rearrange
```
yiyixuxu: ohh I missed this, can we remove `einops` dependency too?

src/diffusers/models/attention_dispatch.py
JerryWu-code: Sorry that may due to mismatch version of flash-attntion 3 with yours, older version (only return **out** in the default flash_attn_func, Version "**flash-attn-3==3.0.0b1**") was not compatible with **lse & \*_**  return, and we will fix this in the next commit.

src/diffusers/models/transformers/transformer_z_image.py
JerryWu-code: Sure we would fix that ~

src/diffusers/models/attention_dispatch.py
JerryWu-code: Partially fixed flash_attn3 in https://github.com/JerryWu-code/diffusers/blob/8a6cb74e7319433126ab09526288ada496a83523/src/diffusers/models/attention_dispatch.py#L651, but not including in this pull request, easy for merging.

src/diffusers/models/transformers/transformer_z_image.py
JerryWu-code: Fixed in https://github.com/huggingface/diffusers/pull/12703/commits/1dd8f3cfe3e261343f42c386115d2184c489a6fd.

JerryWu-code:
This newest branch https://github.com/huggingface/diffusers/pull/12703/commits/a74a0c43c92188b3c4e3b03ec679a519286cf941 should be ready to merge, already merge with main branch right now~, check whether any further things to fix 😊

ChrisLiu6:
@yiyixuxu 
By the way, while testing the `_flash_3` and `_flash_varlen_3` backends, we noticed that the current implementation in `attention_dispatch.py` is incompatible with the latest Flash Attention 3 APIs.

The recent FA3 commit (Dao-AILab/flash-attention@203b9b3) introduced a `return_attn_probs` argument and changed the default behavior. The functions now return a single output tensor by default (instead of a tuple), which causes the current tuple unpacking logic in `diffusers` to fail:

- https://github.com/huggingface/diffusers/blob/d33d9f6715e0c2430c7b7d871b0e20a11601a05d/src/diffusers/models/attention_dispatch.py#L1547-L1557
- https://github.com/huggingface/diffusers/blob/d33d9f6715e0c2430c7b7d871b0e20a11601a05d/src/diffusers/models/attention_dispatch.py#L628-L645

We have implemented a fix that handles this while maintaining backward compatibility:

https://github.com/JerryWu-code/diffusers/commit/de4c6f1b41ef492afeb17b383afe7b83e38442b7#diff-b027e126a86a26981384b125714e0f3bd9923eaa8322f1ae5f6b53fe3e3481c2

Should we include this fix in the current PR, or would you prefer us to open a separate PR for it?

src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -0,0 +1,655 @@
...
+from ...models.modeling_utils import ModelMixin
+from ...utils.import_utils import is_flash_attn_available
+from ...utils.torch_utils import maybe_allow_in_graph
+from ..attention_dispatch import dispatch_attention_fn
+
+
+if is_flash_attn_available():
+    from flash_attn import flash_attn_varlen_func
+else:
+    flash_attn_varlen_func = None
```
yiyixuxu:
```suggestion
```

tests/pipelines/z_image/__init__.py
yiyixuxu, line 1: let's move this folder for now!
we can add test in a follow-up PR!

yiyixuxu:
@ChrisLiu6 
oh thanks! let's do this in a seperate PR since we are eager to get this one in!
>Should we include this fix in the current PR, or would you prefer us to open a separate PR for it?

also need tests & doc in a follow-up PR too :)

yiyixuxu:
also run `make fix-copies` and `make style`

tests/pipelines/z_image/__init__.py
JerryWu-code: Sure, would do this unit-test for all cases in next commit ~ 😊

ChrisLiu6:
Hi @yiyixuxu , ready for merge😄

tests/pipelines/z_image/__init__.py
JerryWu-code: Another pr for unittest is available @ https://github.com/huggingface/diffusers/pull/12715. Thanks for reviewing yiyi @yiyixuxu 😊 ～

tin2tin:
@JerryWu-code 
Super cool. I would like to test it. Is there some code example somewhere on how to use it with Diffusers?

I found it on HF.

tin2tin:
Another question: it doesn't seem to accept image input in the Diffusers implementation, but does the model accept image input? Someone in the ComfyUI community claims it does, but the HF Spaces doesn't have this feature. 

So, is image input possible or not?

asomoza:
@tin2tin the edit model is `soon to be released`, you can see it in the [github repo](https://github.com/Tongyi-MAI/Z-Image). Not sure why is there someone claiming that the turbo model accepts image inputs, as is a pure t2i model

Volnovik:
SageAttention2 seems to introduce severe artifacts into image generation using Z-Image Base. It is present in both UIs I tested. Is someone looking into that?

Volnovik:
Cc: @RuoyiDu @JerryWu-code

PR #12714, opened 2025-11-26
[lora]: Fix Flux2 LoRA NaN test

sayakpaul:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/actions/runs/19675469483/job/56355392848?pr=12711#step:7:185

### DIFF
--- a/tests/lora/test_lora_layers_flux2.py
+++ b/tests/lora/test_lora_layers_flux2.py
@@ -95,6 +95,8 @@ class Flux2LoRATests(unittest.TestCase, PeftLoraLoaderMixinTests):
 
         return noise, input_ids, pipeline_inputs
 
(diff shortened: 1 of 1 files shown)


tests/lora/test_lora_layers_flux2.py
```diff
@@ -94,6 +95,44 @@ def get_dummy_inputs(self, with_generator=True):
 
         return noise, input_ids, pipeline_inputs
 
+    def test_lora_fuse_nan(self):
```
dg845, line 100:
```suggestion
    # Overriding because (1) text encoder LoRAs are not supported in Flux 2 and (2) because the Flux 2 single block
    # QKV projections are always fused, it has no `to_q` param as expected by the original test.
    def test_lora_fuse_nan(self):
```
I think adding a comment explaining why `test_lora_fuse_nan` is being overridden would be useful. The suggestion contains my understanding of the reasons for overriding.

PR #12718, opened 2025-11-26
[feat]: implement "local" caption upsampling for Flux.2

sayakpaul:
# What does this PR do?

Test code:

```py
import torch 

prompt = "a photo of a forest with mist swirling around the tree trunks. The word 'FLUX.2' is painted over it in big, red brush strokes with visible texture"
image = pipe(
    prompt=prompt,
    height=768,
    width=1360,
    caption_upsample_temperature=0.15,
).images[0]
image.save("upsampling.png")
print(f"{torch.cuda.max_memory_reserved() / (1024 ** 3)=}")
```

<details>
<summary>Generated upsampled prompt:</summary>

```
A serene and atmospheric forest scene, captured in a high-resolution photograph, showcases towering, ancient trees with thick, gnarled trunks and sprawling branches that create a dense canopy overhead. The forest floor is carpeted with a lush layer of moss, ferns, and fallen leaves, adding a sense of depth and texture to the image. Mist swirls gently around the tree trunks, creating a dreamy, ethereal atmosphere. The mist is illuminated by soft, diffused light that filters through the canopy, casting dappled shadows and highlighting the intricate details of the bark and foliage. The overall color palette is muted and natural, with shades of green, brown, and gray dominating the scene. In the foreground, the word "FLUX.2" is painted in bold, red brush strokes with visible texture, standing out against the natural backdrop. The paint appears wet and glossy, with visible brushstrokes and subtle drips, adding a dynamic and artistic element to the image. The text is centrally placed and slightly tilted, drawing the viewer\'s eye and adding a sense of movement to the otherwise tranquil scene.
```

</details>

### Output

| No prompt upsampling | Prompt upsampling |
|---------------------|---------------------|
| Image 1 | Image 2 |

---

### Notes

* I decided to create a `system_messages.py` script under `src/diffusers/pipelines/flux2` so that other pipelines derived from Flux2 can easily use it.
* If `caption_upsample_temperature` is set (defaults to `None`), we perform the process.
* The image processor changes are to accommodate [this method](https://github.com/black-forest-labs/flux2/blob/5a5d316b1b42f6b59a8c9194b77c8256be848432/src/flux2/text_encoder.py#L49) in the original codebase. Open to other suggestions, of course, on how to best accommodate them.
* The original codebase [implements](https://github.com/black-forest-labs/flux2/blob/5a5d316b1b42f6b59a8c9194b77c8256be848432/scripts/cli.py#L38) an OpenRouter client with a larger Pixtral model for doing prompt upsampling remotely. I think we can do that through Inference Endpoints? (cc: @apolinario @ariG23498)

### DIFF
--- a/src/diffusers/pipelines/flux2/system_messages.py
+++ b/src/diffusers/pipelines/flux2/system_messages.py
@@ -1,11 +1,14 @@
+# docstyle-ignore
 """
 These system prompts come from:
 https://github.com/black-forest-labs/flux2/blob/5a5d316b1b42f6b59a8c9194b77c8256be848432/src/flux2/system_messages.py#L54
 """
 
+# docstyle-ignore
 SYSTEM_MESSAGE = """You are an AI that reasons about image descriptions. You give structured responses focusing on object relationships, object
 attribution and actions without speculation."""
 
+# docstyle-ignore
 SYSTEM_MESSAGE_UPSAMPLING_T2I = """You are an expert prompt engineer for FLUX.2 by Black Forest Labs. Rewrite user prompts to be more descriptive while strictly preserving their core subject and intent.
 
 Guidelines:
@@ -15,6 +18,7 @@ Guidelines:
 
 Output only the revised prompt and nothing else."""
 
+# docstyle-ignore
 SYSTEM_MESSAGE_UPSAMPLING_I2I = """You are FLUX.2 by Black Forest Labs, an image-editing expert. You convert editing requests into one concise instruction (50-80 words, ~30 for brief requests).
 
 Rules:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,8 @@
 [tool.ruff]
 line-length = 119
+extend-exclude = [
+    "src/diffusers/pipelines/flux2/system_messages.py",
+]
 
 [tool.ruff.lint]
 # Never enforce `E501` (line length violations).


src/diffusers/pipelines/flux2/system_messages.py
```diff
@@ -0,0 +1,35 @@
+"""
+These system prompts come from:
+https://github.com/black-forest-labs/flux2/blob/5a5d316b1b42f6b59a8c9194b77c8256be848432/src/flux2/system_messages.py#L54
+"""
+
+SYSTEM_MESSAGE = """You are an AI that reasons about image descriptions. You give structured responses focusing on object relationships,
```
sayakpaul: As discussed internally, this new-line character thingy messes the quality a bit. Hence, I have decided to keep these system messages one-to-one same as the original implementation linked above. 

If we run `make style && make quality`, this order will be completely destroyed. We can change the `pyproject.toml` to exclude this path from getting formatted. But before we do that let's see if this is the best we have.

src/diffusers/pipelines/flux2/system_messages.py
```diff
@@ -0,0 +1,29 @@
+"""
+These system prompts come from:
```
sayakpaul, line 3: As discussed internally, this new-line character thingy messes up the quality a bit. Hence, I have decided to keep these system messages one-to-one same as the original implementation linked above.

If we run make style && make quality, this order will be completely destroyed. We can change the `pyproject.toml` to exclude this path from getting formatted. But before we do that, let's see if this is the best we have.

src/diffusers/pipelines/flux2/image_processor.py
```diff
@@ -98,10 +98,15 @@ def check_image_input(
         return image
 
     @staticmethod
-    def _resize_to_target_area(image: PIL.Image.Image, target_area: int = 1024 * 1024) -> Tuple[int, int]:
+    def _resize_to_target_area(
```
yiyixuxu: ohh do you want to add a new method called something like `_resize_if_exceeds_area`? or rename this one if we only use it this way

src/diffusers/pipelines/flux2/pipeline_flux2.py
```diff
@@ -57,24 +58,107 @@
...
+                    "content": [{"type": "text", "text": cleaned_txt[i]}],
+                }
+            )
+
+        return messages
+
+
+# Adapted from
+# https://github.com/black-forest-labs/flux2/blob/5a5d316b1b42f6b59a8c9194b77c8256be848432/src/flux2/text_encoder.py#L49C5-L66C19
+def _validate_and_process_images(
```
yiyixuxu, line 132: can we have a seperate step to validate and process image and then run `format_input`?

ariG23498:
@sayakpaul this is really nice. Do you want me to start working on an Endpoint? We can take this conversation to private slack and see how this works.

Update:

After inspecting, it turns out that the model needs upwards of 300 GBs to run. (from the official [model](https://huggingface.co/mistralai/Pixtral-Large-Instruct-2411) card)

> note that running this model on GPU requires over 300 GB of GPU RAM

Building a free Inference Endpoint does not seem to be feasible by me and @sayakpaul and hence we are benching this project. Another option would be to route through Inference Providers, but we have not seen a need (other than this specific one) to let our providers host this model.

src/diffusers/pipelines/flux2/pipeline_flux2.py
sayakpaul: Yes. We now first `_validate_and_process_images()` and then pass the resultant images to `format_input()`.

src/diffusers/pipelines/flux2/image_processor.py
sayakpaul: Yup. 

I created `_resize_if_exceeds_area()` which is basically:

```py
def _resize_if_exceeds_area(image, target_area=1024 * 1024) -> PIL.Image.Image:
    image_width, image_height = image.size
    pixel_count = image_width * image_height
    if pixel_count <= target_area:
        return image
    return Flux2ImageProcessor._resize_to_target_area(image, target_area)
```

sayakpaul:
Gonna go with https://github.com/huggingface/diffusers/pull/12718/#discussion_r2564335738

PR #12741, opened 2025-11-28
[Z-Image] various small changes, Z-Image transformer tests, etc.

sayakpaul:
# What does this PR do?

* Introduces a dedicated test suite for the Z-Image DiT.
* Adds `is_flaky` decorator to `test_inference()` in the Z-Image pipeline test suite.
* Adds a `return_dict` argument to the `forward()` of Z-Image DiT, following other models in the library.
  * As a consequence of this, I followed the `return` pattern, i.e., return a `Transformer2DModelOutput` type output or something like `return (out,)`.

---

## Notes

* The model accepts the hidden states as a `list[torch.Tensor]` which differs from other models. Output also follows the same type. This is why I had to modify a couple of tests (where it was reasonably easy) to allow this. Tests, where it was not relatively easy, were skipped (such as `test_training`, `test_ema_training, etc.).
* The repeated block in this model is `ZImageTransformerBlock`, which is used for `noise_refiner`, `context_refiner`, and `layers`. As a consequence of this, the inputs recorded for the block would vary during compilation and full compilation with `fullgraph=True` would trigger recompilation at least thrice.
* Some of the group offloading tests were skipped because of states that interfered in between the tests (as also noted [here](https://github.com/huggingface/diffusers/blob/1b91856d0eee7b6fb58340e9b54ea2c3d5424311/tests/pipelines/z_image/test_z_image.py#L45)).
* `x_pad_token` and `cap_pad_token` params within the DiT are initialized with `torch.empty()`, possibly for memory efficiency, but they interfere during test in very weird ways. This is because `torch.empty()` can render NaNs. To prevent this from creeping into the tests, I tried adding `is_flaky()` to some of the tests that got affected by this, but that didn't help (see [this](https://github.com/huggingface/diffusers/actions/runs/19779401325/job/56677452099?pr=12741#step:7:573)). @JerryWu-code, would it be safe to get `x_pad_token` and `cap_pad_token` initialized deterministically, maybe with something like `torch.ones()`? Or do you think it would have memory implications? 

---

## Minor nits

* We usually avoid raw `assert` statements inside the model implementations in favor of properly raising errors. Should we follow something similar here, too?
* There is a `self.scheduler.sigma_min = 0.0` inside the Z-Image pipeline: https://github.com/huggingface/diffusers/blob/1b91856d0eee7b6fb58340e9b54ea2c3d5424311/src/diffusers/pipelines/z_image/pipeline_z_image.py#L477. Maybe I am missing out on something but that seems like an antipattern to me.
* The signature of `forward()` of the DiT has shorthand variable names: `x`, `t`, `cap_feats`, unlike `hidden_states`, `timestep`, and `encoder_hidden_states`.
* Should `_cfg_normalization` and `_cfg_truncation` inside the pipeline be turned into properties like `guidance_scale`?

Maybe we could consider revisiting them (but not a priority perhaps).

Cc: @JerryWu-code

### DIFF
--- a/tests/lora/test_lora_layers_z_image.py
+++ b/tests/lora/test_lora_layers_z_image.py
@@ -244,13 +244,11 @@ class ZImageLoRATests(unittest.TestCase, PeftLoraLoaderMixinTests):
     def test_lora_scale_kwargs_match_fusion(self):
         super().test_lora_scale_kwargs_match_fusion(5e-2, 5e-2)
 
-    unittest.skip("Needs to be debugged.")
-
+    @unittest.skip("Needs to be debugged.")
     def test_set_adapters_match_attention_kwargs(self):
         super().test_set_adapters_match_attention_kwargs()
 
-    unittest.skip("Needs to be debugged.")
-
+    @unittest.skip("Needs to be debugged.")
     def test_simple_inference_with_text_denoiser_lora_and_scale(self):
         super().test_simple_inference_with_text_denoiser_lora_and_scale()
 


src/diffusers/models/transformers/transformer_z_image.py
```diff
@@ -658,4 +630,7 @@ def forward(
         unified = list(unified.unbind(dim=0))
         x = self.unpatchify(unified, x_size, patch_size, f_patch_size)
 
-        return x, {}
+        if not return_dict:
+            return (x,)
+
+        return Transformer2DModelOutput(sample=x)
```
sayakpaul, line 650: Should be a very safe change?

sayakpaul:
Failures in "Fast tests for PRs / Fast PyTorch Models & Schedulers CPU tests (pull_request)" pass even when run with `CUDA_VISIBLE_DEVICES="" pytest tests/models/transformers/test_models_transformer_z_image.py`.

Edit: it likely fails when `CUDA_VISIBLE_DEVICES="" pytest tests/models/` is run.

src/diffusers/models/transformers/transformer_z_image.py
yiyixuxu: yes indeed, is this the only actual change in this file? (others seems just format changes)

yiyixuxu (approved):
thanks!

dg845:
Not blocking, but I think we should think about how to best support and test the `List[torch.Tensor]` `hidden_states` pattern going forward.

My current understanding is that the motivation for using a `List[torch.Tensor]` rather than a single batched `torch.Tensor` is to support ragged tensors. This makes it easier to support a batch of `hidden_states` that would naturally have different shapes (for example, images of different resolutions) without needing extra memory and logic to pad them to the same shape. I'm not sure I fully understand the use cases and advantages/disadvantages for this pattern, so I would greatly appreciate it if someone could shed some more light on it. In particular, what is seen as the main reason(s) to use this pattern for current models?

I could see several support strategies here:

1. Insist that all model inputs are single batched tensors.
2. Allow lists of tensors, and try to support them as much as possible within the existing model tests (e.g. `ModelTesterMixin`).
3. Create a separate test suite to handle models with `List[torch.Tensor]` inputs (e.g. a ragged tensor version of `ModelTesterMixin`).

My current personal preference is for (3), as it allows us to tailor the tests for actual `List[torch.Tensor]` use cases, (1) seems unnecessarily restrictive, and (2) seems like it might struggle to handle truly ragged inputs, and I'm wary that it might make the test suite more complicated and result in a lot of test skipping for tests which it is hard to support lists of tensors in the current implementation. A downside of (3) is that it likely requires the most work of the three.

It seems like PyTorch will probably eventually support ragged tensors with [`torch.nested`](https://docs.pytorch.org/docs/stable/nested.html). It's not obvious to me which solution is most "future-proof", but that may be a consideration as well.

tests/models/transformers/test_models_transformer_z_image.py
```diff
@@ -0,0 +1,163 @@
...
+    @unittest.skip("Group offloading needs to revisited for this model because of state population.")
+    def test_group_offloading(self):
+        super().test_group_offloading()
+
+    @unittest.skip("Group offloading needs to revisited for this model because of state population.")
+    def test_group_offloading_with_disk(self):
+        super().test_group_offloading_with_disk()
+
+
+class Flux2TransformerCompileTests(TorchCompileTesterMixin, unittest.TestCase):
```
dg845:
```suggestion
class ZImageTransformerCompileTests(TorchCompileTesterMixin, unittest.TestCase):
```
nit: naming

tests/models/test_modeling_common.py
```diff
@@ -2086,6 +2162,8 @@ def test_torch_compile_repeated_blocks(self):
         recompile_limit = 1
         if self.model_class.__name__ == "UNet2DConditionModel":
             recompile_limit = 2
+        elif self.model_class.__name__ == "ZImageTransformer2DModel":
+            recompile_limit = 3
```
dg845, line 2163: Not blocking, but I think it makes more sense to refactor this test to have a `recompile_limit` argument:

```python
    def test_torch_compile_repeated_blocks(self, recompile_limit: int = 1):
        ...
```

and then override the test as follows:

```python
    def test_torch_compile_repeated_blocks(self):
        super().test_torch_compile_repeated_blocks(recompile_limit=3)
```

IMO it's more clear this way that the Z-Image model is using special testing logic.

tests/models/transformers/test_models_transformer_z_image.py
```diff
@@ -0,0 +1,163 @@
...
+    model_class = ZImageTransformer2DModel
+    different_shapes_for_compilation = [(4, 4), (4, 8), (8, 8)]
+
+    def prepare_init_args_and_inputs_for_common(self):
+        return ZImageTransformerTests().prepare_init_args_and_inputs_for_common()
+
+    def prepare_dummy_input(self, height, width):
+        return ZImageTransformerTests().prepare_dummy_input(height=height, width=width)
+
+    @unittest.skip("Fullgraph is broken")
```
dg845: Should the skip reason here reflect this?

> The repeated block in this model is ZImageTransformerBlock, which is used for noise_refiner, context_refiner, and layers. As a consequence of this, the inputs recorded for the block would vary during compilation and full compilation with fullgraph=True would trigger recompilation at least thrice.

Or is there another reason why we would expect `fullgraph=True` to fail?

tests/models/test_modeling_common.py
dg845: Or perhaps making `recompile_limit` a class attribute might make sense, since it seems like we could reuse it in `test_torch_compile_recompilation_and_graph_break`. Unless I'm missing something, it seems like if `self.recompilation_limit > 1`, `test_torch_compile_recompilation_and_graph_break` should always fail due to

https://github.com/huggingface/diffusers/blob/edf36f5128abf3e6ecf92b5145115514363c58e6/tests/models/test_modeling_common.py#L2070

src/diffusers/models/transformers/transformer_z_image.py
sayakpaul: Yup.

tests/models/test_modeling_common.py
sayakpaul: I will take care of this in a separate PR. Thanks for the ideas.

sayakpaul:
@dg845 I have some thoughts on https://github.com/huggingface/diffusers/pull/12741#issuecomment-3600194452. Let's talk async and we can post a summary of our discussions here.

tests/lora/test_lora_layers_z_image.py
```diff
@@ -34,13 +30,9 @@
...
 sys.path.append(".")
 
-from .utils import PeftLoraLoaderMixinTests  # noqa: E402
+from .utils import PeftLoraLoaderMixinTests, check_if_lora_correctly_set  # noqa: E402
 
 
-@unittest.skip(
-    "ZImage LoRA tests are skipped due to non-deterministic behavior from complex64 RoPE operations "
-    "and torch.empty padding tokens. LoRA functionality works correctly with real models."
-)
```
sayakpaul, line 43: Cc: @CalamitousFelicitousness

tests/pipelines/z_image/test_z_image.py
```diff
@@ -183,7 +184,7 @@ def test_inference(self):
         self.assertEqual(generated_image.shape, (3, 32, 32))
 
         # fmt: off
-        expected_slice = torch.tensor([0.4521, 0.4512, 0.4693, 0.5115, 0.5250, 0.5271, 0.4776, 0.4688, 0.2765, 0.2164, 0.5656, 0.6909, 0.3831, 0.5431, 0.5493, 0.4732])
+        expected_slice = torch.tensor([0.4622, 0.4532, 0.4714, 0.5087, 0.5371, 0.5405, 0.4492, 0.4479, 0.2984, 0.2783, 0.5409, 0.6577, 0.3952, 0.5524, 0.5262, 0.453])
```
sayakpaul, line 187: Because we're now initializing the padding token to be ones.

tests/models/transformers/test_models_transformer_z_image.py
```diff
@@ -0,0 +1,171 @@
...
+os.environ["CUDA_LAUNCH_BLOCKING"] = "1"
+os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":16:8"
+torch.use_deterministic_algorithms(False)
+torch.backends.cudnn.deterministic = True
+torch.backends.cudnn.benchmark = False
+if hasattr(torch.backends, "cuda"):
+    torch.backends.cuda.matmul.allow_tf32 = False
+
+
+@unittest.skipIf(
```
sayakpaul, line 39: @yiyixuxu just FYI.

Issue #11966, opened 2025-07-21
How about forcing the first and last block on device when groupoffloading is used?


seed93:
**Is your feature request related to a problem? Please describe.**
When group offloading is enabled, the offload and onload cannot be streamed between steps and this is really a big time comsuming problem.

**Describe the solution you'd like.**
Is it possible to add an option that could make the first and last block forced on device to avoid offload and onload?

@a-r-r-o-w Could you please give some help? Thanks so much.

seed93:
I have tried to make it a circle here: https://github.com/huggingface/diffusers/blob/v0.33.1/src/diffusers/hooks/group_offloading.py#L319
However it seems not work.

a-r-r-o-w:
Hi @seed93, forcing particular block(s), or other internal layers, to remain on the device should be possible. I'll try to prototype something later this week. I'd imagine denoting what layers to keep on device will have to be determined via a regex, so that ModuleGroup's of remaining modules can be formed easily. Trying to prefetch the first layer from the last layer might also be possible 🤔

seed93:
Thank you so much. I tried a circular prefetch as I said on the last comment but it seems not work. I checked this using nsys and the memcpy between steps still exists.
As the image shows below, the offload and onload were captured by pre_forward and post_forward of GroupOffloadingHook.
<img width="1790" height="989" alt="Image" src="https://github.com/user-attachments/assets/8eb24d6b-b459-4625-816f-c4d3305f53e4" />

seed93:
```
diff --git a/src/diffusers/hooks/group_offloading.py b/src/diffusers/hooks/group_offloading.py
index 36abf3c41..d1397e237 100644
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -435,11 +435,12 @@ class LazyPrefetchGroupOffloadingHook(ModelHook):
             base_module_group_offloading_hook.next_group = group_offloading_hooks[0].group
             base_module_group_offloading_hook.next_group.onload_self = False
 
-        for i in range(num_executed - 1):
+        for i in range(num_executed):
             name1, _ = self.execution_order[i]
-            name2, _ = self.execution_order[i + 1]
+            next_i = (i + 1) % num_executed
+            name2, _ = self.execution_order[next_i]
             logger.debug(f"Applying lazy prefetch group offloading from {name1} to {name2}")
[...]

PR #12747, opened 2025-11-29
Feature/group offload pinning

Aki-07:
# What does this PR do?

Fixes [#11966](https://github.com/huggingface/diffusers/issues/11966)

## Before submitting

## Who can review?

@sayakpaul

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -25,7 +25,7 @@ jobs:
     if: github.event_name == 'pull_request'
     steps:
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
 
       - name: Check out code
         uses: actions/checkout@v6
@@ -101,14 +101,14 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
       - name: Login to Docker Hub
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         with:
           username: ${{ env.REGISTRY }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Build and push
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v6
         with:
           no-cache: true
           context: ./docker/${{ matrix.image-name }}
--- a/.github/workflows/pr_modular_tests.yml
+++ b/.github/workflows/pr_modular_tests.yml
@@ -75,9 +75,27 @@ jobs:
         if: ${{ failure() }}
         run: |
           echo "Repo consistency check failed. Please ensure the right dependency versions are installed with 'pip install -e .[quality]' and run 'make fix-copies'" >> $GITHUB_STEP_SUMMARY
+  check_auto_docs:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v6
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: |
+          pip install --upgrade pip
+          pip install .[quality]
+      - name: Check auto docs
+        run: make modular-autodoctrings
+      - name: Check if failure
+        if: ${{ failure() }}
+        run: |
+          echo "Auto docstring checks failed. Please run `python utils/modular_auto_docstring.py --fix_and_overwrite`." >> $GITHUB_STEP_SUMMARY
 
   run_fast_tests:
-    needs: [check_code_quality, check_repository_consistency]
+    needs: [check_code_quality, check_repository_consistency, check_auto_docs]
     name: Fast PyTorch Modular Pipeline CPU tests
 
     runs-on:
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -11,4 +11,4 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: typos-action
-        uses: crate-ci/typos@v1.12.4
+        uses: crate-ci/typos@v1.42.1
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,10 @@ fix-copies:
 	python utils/check_copies.py --fix_and_overwrite
 	python utils/check_dummies.py --fix_and_overwrite
 
+# Auto docstrings in modular blocks
+modular-autodoctrings:
+	python utils/modular_auto_docstring.py
+
 # Run tests for the library
 
 test:
--- a/docker/diffusers-pytorch-cuda/Dockerfile
+++ b/docker/diffusers-pytorch-cuda/Dockerfile
@@ -1,8 +1,8 @@
-FROM nvidia/cuda:12.1.0-runtime-ubuntu20.04
+FROM nvidia/cuda:12.9.1-runtime-ubuntu24.04
 LABEL maintainer="Hugging Face"
 LABEL repository="diffusers"
 
-ARG PYTHON_VERSION=3.12
+ARG PYTHON_VERSION=3.10
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get -y update \
@@ -32,10 +32,17 @@ RUN uv venv --python ${PYTHON_VERSION} --seed ${VIRTUAL_ENV}
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 # pre-install the heavy dependencies (these can later be overridden by the deps from setup.py)
+# Install torch, torchvision, and torchaudio together to ensure compatibility
 RUN uv pip install --no-cache-dir \
     torch \
     torchvision \
-    torchaudio
+    torchaudio \
+    --index-url https://download.pytorch.org/whl/cu129
+
+# Install compatible versions of numba/llvmlite for Python 3.10+
+RUN uv pip install --no-cache-dir \
+    "llvmlite>=0.40.0" \
+    "numba>=0.57.0"
 
 RUN uv pip install --no-cache-dir "git+https://github.com/huggingface/diffusers.git@main#egg=diffusers[test]"
 
--- a/docker/diffusers-pytorch-xformers-cuda/Dockerfile
+++ b/docker/diffusers-pytorch-xformers-cuda/Dockerfile
@@ -1,8 +1,8 @@
-FROM nvidia/cuda:12.1.0-runtime-ubuntu20.04
+FROM nvidia/cuda:12.9.1-runtime-ubuntu24.04
 LABEL maintainer="Hugging Face"
 LABEL repository="diffusers"
 
-ARG PYTHON_VERSION=3.12
+ARG PYTHON_VERSION=3.10
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get -y update \
@@ -32,10 +32,17 @@ RUN uv venv --python ${PYTHON_VERSION} --seed ${VIRTUAL_ENV}
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 # pre-install the heavy dependencies (these can later be overridden by the deps from setup.py)
+# Install torch, torchvision, and torchaudio together to ensure compatibility
 RUN uv pip install --no-cache-dir \
     torch \
     torchvision \
-    torchaudio
+    torchaudio \
+    --index-url https://download.pytorch.org/whl/cu129
+
+# Install compatible versions of numba/llvmlite for Python 3.10+
+RUN uv pip install --no-cache-dir \
+    "llvmlite>=0.40.0" \
+    "numba>=0.57.0"
 
 RUN uv pip install --no-cache-dir "git+https://github.com/huggingface/diffusers.git@main#egg=diffusers[test]"
 
--- a/docs/source/en/api/pipelines/ltx2.md
+++ b/docs/source/en/api/pipelines/ltx2.md
@@ -24,6 +24,179 @@ You can find all the original LTX-Video checkpoints under the [Lightricks](https
 
 The original codebase for LTX-2 can be found [here](https://github.com/Lightricks/LTX-2).
 
+## Two-stages Generation
+Recommended pipeline to achieve production quality generation, this pipeline is composed of two stages:
+
+- Stage 1: Generate a video at the target resolution using diffusion sampling with classifier-free guidance (CFG). This stage produces a coherent low-noise video sequence that respects the text/image conditioning.
+- Stage 2: Upsample the Stage 1 output by 2 and refine details using a distilled LoRA model to improve fidelity and visual quality. Stage 2 may apply lighter CFG to preserve the structure from Stage 1 while enhancing texture and sharpness.
+
+Sample usage of text-to-video two stages pipeline
+
+```py
+import torch
+from diffusers import FlowMatchEulerDiscreteScheduler
+from diffusers.pipelines.ltx2 import LTX2Pipeline, LTX2LatentUpsamplePipeline
+from diffusers.pipelines.ltx2.latent_upsampler import LTX2LatentUpsamplerModel
+from diffusers.pipelines.ltx2.utils import STAGE_2_DISTILLED_SIGMA_VALUES
+from diffusers.pipelines.ltx2.export_utils import encode_video
+
+device = "cuda:0"
+width = 768
+height = 512
+
+pipe = LTX2Pipeline.from_pretrained(
+    "Lightricks/LTX-2", torch_dtype=torch.bfloat16
+)
+pipe.enable_sequential_cpu_offload(device=device)
+
+prompt = "A beautiful sunset over the ocean"
+negative_prompt = "shaky, glitchy, low quality, worst quality, deformed, distorted, disfigured, motion smear, motion artifacts, fused fingers, bad anatomy, weird hand, ugly, transition, static."
+
+# Stage 1 default (non-distilled) inference
+frame_rate = 24.0
+video_latent, audio_latent = pipe(
+    prompt=prompt,
+    negative_prompt=negative_prompt,
+    width=width,
+    height=height,
+    num_frames=121,
+    frame_rate=frame_rate,
+    num_inference_steps=40,
+    sigmas=None,
+    guidance_scale=4.0,
+    output_type="latent",
+    return_dict=False,
+)
+
+latent_upsampler = LTX2LatentUpsamplerModel.from_pretrained(
+    "Lightricks/LTX-2",
+    subfolder="latent_upsampler",
+    torch_dtype=torch.bfloat16,
+)
+upsample_pipe = LTX2LatentUpsamplePipeline(vae=pipe.vae, latent_upsampler=latent_upsampler)
+upsample_pipe.enable_model_cpu_offload(device=device)
+upscaled_video_latent = upsample_pipe(
+    latents=video_latent,
+    output_type="latent",
+    return_dict=False,
+)[0]
+
+# Load Stage 2 distilled LoRA
+pipe.load_lora_weights(
+    "Lightricks/LTX-2", adapter_name="stage_2_distilled", weight_name="ltx-2-19b-distilled-lora-384.safetensors"
+)
+pipe.set_adapters("stage_2_distilled", 1.0)
+# VAE tiling is usually necessary to avoid OOM error when VAE decoding
+pipe.vae.enable_tiling()
+# Change scheduler to use Stage 2 distilled sigmas as is
+new_scheduler = FlowMatchEulerDiscreteScheduler.from_config(
+    pipe.scheduler.config, use_dynamic_shifting=False, shift_terminal=None
+)
+pipe.scheduler = new_scheduler
+# Stage 2 inference with distilled LoRA and sigmas
+video, audio = pipe(
+    latents=upscaled_video_latent,
+    audio_latents=audio_latent,
+    prompt=prompt,
+    negative_prompt=negative_prompt,
+    num_inference_steps=3,
+    noise_scale=STAGE_2_DISTILLED_SIGMA_VALUES[0], # renoise with first sigma value https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-pipelines/src/ltx_pipelines/ti2vid_two_stages.py#L218
+    sigmas=STAGE_2_DISTILLED_SIGMA_VALUES,
+    guidance_scale=1.0,
+    output_type="np",
+    return_dict=False,
+)
+video = (video * 255).round().astype("uint8")
+video = torch.from_numpy(video)
+
+encode_video(
+    video[0],
+    fps=frame_rate,
+    audio=audio[0].float().cpu(),
+    audio_sample_rate=pipe.vocoder.config.output_sampling_rate,
+    output_path="ltx2_lora_distilled_sample.mp4",
+)
+```
+
+## Distilled checkpoint generation
+Fastest two-stages generation pipeline using a distilled checkpoint.
+
+```py
+import torch
+from diffusers.pipelines.ltx2 import LTX2Pipeline, LTX2LatentUpsamplePipeline
+from diffusers.pipelines.ltx2.latent_upsampler import LTX2LatentUpsamplerModel
+from diffusers.pipelines.ltx2.utils import DISTILLED_SIGMA_VALUES, STAGE_2_DISTILLED_SIGMA_VALUES
+from diffusers.pipelines.ltx2.export_utils import encode_video
+
+device = "cuda"
+width = 768
+height = 512
+random_seed = 42
+generator = torch.Generator(device).manual_seed(random_seed)
+model_path = "rootonchair/LTX-2-19b-distilled"
+
+pipe = LTX2Pipeline.from_pretrained(
+    model_path, torch_dtype=torch.bfloat16
+)
+pipe.enable_sequential_cpu_offload(device=device)
+
+prompt = "A beautiful sunset over the ocean"
+negative_prompt = "shaky, glitchy, low quality, worst quality, deformed, distorted, disfigured, motion smear, motion artifacts, fused fingers, bad anatomy, weird hand, ugly, transition, static."
+
+frame_rate = 24.0
+video_latent, audio_latent = pipe(
+    prompt=prompt,
+    negative_prompt=negative_prompt,
+    width=width,
+    height=height,
+    num_frames=121,
+    frame_rate=frame_rate,
+    num_inference_steps=8,
+    sigmas=DISTILLED_SIGMA_VALUES,
+    guidance_scale=1.0,
+    generator=generator,
+    output_type="latent",
+    return_dict=False,
+)
+
+latent_upsampler = LTX2LatentUpsamplerModel.from_pretrained(
+    model_path,
+    subfolder="latent_upsampler",
+    torch_dtype=torch.bfloat16,
+)
+upsample_pipe = LTX2LatentUpsamplePipeline(vae=pipe.vae, latent_upsampler=latent_upsampler)
+upsample_pipe.enable_model_cpu_offload(device=device)
+upscaled_video_latent = upsample_pipe(
+    latents=video_latent,
+    output_type="latent",
+    return_dict=False,
+)[0]
+
+video, audio = pipe(
+    latents=upscaled_video_latent,
+    audio_latents=audio_latent,
+    prompt=prompt,
+    negative_prompt=negative_prompt,
+    num_inference_steps=3,
+    noise_scale=STAGE_2_DISTILLED_SIGMA_VALUES[0], # renoise with first sigma value https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-pipelines/src/ltx_pipelines/distilled.py#L178
+    sigmas=STAGE_2_DISTILLED_SIGMA_VALUES,
+    generator=generator,
+    guidance_scale=1.0,
+    output_type="np",
+    return_dict=False,
+)
+video = (video * 255).round().astype("uint8")
+video = torch.from_numpy(video)
+
+encode_video(
+    video[0],
+    fps=frame_rate,
+    audio=audio[0].float().cpu(),
+    audio_sample_rate=pipe.vocoder.config.output_sampling_rate,
+    output_path="ltx2_distilled_sample.mp4",
+)
+```
+
 ## LTX2Pipeline
 
 [[autodoc]] LTX2Pipeline
--- a/docs/source/en/quantization/torchao.md
+++ b/docs/source/en/quantization/torchao.md
@@ -83,25 +83,6 @@ Refer to this [table](https://github.com/huggingface/diffusers/pull/10009#issue-
 > [!TIP]
 > The FP8 post-training quantization schemes in torchao are effective for GPUs with compute capability of at least 8.9 (RTX-4090, Hopper, etc.). FP8 often provides the best speed, memory, and quality trade-off when generating images and videos. We recommend combining FP8 and torch.compile if your GPU is compatible.
 
-## autoquant
-
-torchao provides [autoquant](https://docs.pytorch.org/ao/stable/generated/torchao.quantization.autoquant.html#torchao.quantization.autoquant) an automatic quantization API. Autoquantization chooses the best quantization strategy by comparing the performance of each strategy on chosen input types and shapes. This is only supported in Diffusers for individual models at the moment.
-
-```py
-import torch
-from diffusers import DiffusionPipeline
-from torchao.quantization import autoquant
-
-# Load the pipeline
-pipeline = DiffusionPipeline.from_pretrained(
-    "black-forest-labs/FLUX.1-schnell",
-    torch_dtype=torch.bfloat16,
-    device_map="cuda"
-)
-
-transformer = autoquant(pipeline.transformer)
-```
-
 ## Supported quantization types
 
 torchao supports weight-only quantization and weight and dynamic-activation quantization for int8, float3-float8, and uint1-uint7.
--- a/examples/dreambooth/train_dreambooth_lora_qwen_image.py
+++ b/examples/dreambooth/train_dreambooth_lora_qwen_image.py
@@ -1467,7 +1467,8 @@ def main(args):
                 else:
                     num_repeat_elements = len(prompts)
                     prompt_embeds = prompt_embeds.repeat(num_repeat_elements, 1, 1)
-                    prompt_embeds_mask = prompt_embeds_mask.repeat(num_repeat_elements, 1)
+                    if prompt_embeds_mask is not None:
+                        prompt_embeds_mask = prompt_embeds_mask.repeat(num_repeat_elements, 1)
                 # Convert images to latent space
                 if args.cache_latents:
                     model_input = latents_cache[step].sample()
--- a/scripts/convert_ltx2_to_diffusers.py
+++ b/scripts/convert_ltx2_to_diffusers.py
@@ -63,6 +63,8 @@ LTX_2_0_VIDEO_VAE_RENAME_DICT = {
     "up_blocks.4": "up_blocks.1",
     "up_blocks.5": "up_blocks.2.upsamplers.0",
     "up_blocks.6": "up_blocks.2",
+    "last_time_embedder": "time_embedder",
+    "last_scale_shift_table": "scale_shift_table",
     # Common
     # For all 3D ResNets
     "res_blocks": "resnets",
@@ -372,7 +374,9 @@ def convert_ltx2_connectors(original_state_dict: Dict[str, Any], version: str) -
     return connectors
 
 
-def get_ltx2_video_vae_config(version: str) -> Tuple[Dict[str, Any], Dict[str, Any], Dict[str, Any]]:
+def get_ltx2_video_vae_config(
+    version: str, timestep_conditioning: bool = False
+) -> Tuple[Dict[str, Any], Dict[str, Any], Dict[str, Any]]:
     if version == "test":
         config = {
             "model_id": "diffusers-internal-dev/dummy-ltx2",
@@ -396,7 +400,7 @@ def get_ltx2_video_vae_config(version: str) -> Tuple[Dict[str, Any], Dict[str, A
                 "downsample_type": ("spatial", "temporal", "spatiotemporal", "spatiotemporal"),
                 "upsample_residual": (True, True, True),
                 "upsample_factor": (2, 2, 2),
-                "timestep_conditioning": False,
+                "timestep_conditioning": timestep_conditioning,
                 "patch_size": 4,
                 "patch_size_t": 1,
                 "resnet_norm_eps": 1e-6,
@@ -433,7 +437,7 @@ def get_ltx2_video_vae_config(version: str) -> Tuple[Dict[str, Any], Dict[str, A
                 "downsample_type": ("spatial", "temporal", "spatiotemporal", "spatiotemporal"),
                 "upsample_residual": (True, True, True),
                 "upsample_factor": (2, 2, 2),
-                "timestep_conditioning": False,
+                "timestep_conditioning": timestep_conditioning,
                 "patch_size": 4,
                 "patch_size_t": 1,
                 "resnet_norm_eps": 1e-6,
@@ -450,8 +454,10 @@ def get_ltx2_video_vae_config(version: str) -> Tuple[Dict[str, Any], Dict[str, A
     return config, rename_dict, special_keys_remap
 
 
-def convert_ltx2_video_vae(original_state_dict: Dict[str, Any], version: str) -> Dict[str, Any]:
-    config, rename_dict, special_keys_remap = get_ltx2_video_vae_config(version)
+def convert_ltx2_video_vae(
+    original_state_dict: Dict[str, Any], version: str, timestep_conditioning: bool
+) -> Dict[str, Any]:
+    config, rename_dict, special_keys_remap = get_ltx2_video_vae_config(version, timestep_conditioning)
     diffusers_config = config["diffusers_config"]
 
     with init_empty_weights():
@@ -659,10 +665,15 @@ def get_model_state_dict_from_combined_ckpt(combined_ckpt: Dict[str, Any], prefi
 def get_args():
     parser = argparse.ArgumentParser()
 
+    def none_or_str(value: str):
+        if isinstance(value, str) and value.lower() == "none":
+            return None
+        return value
+
     parser.add_argument(
         "--original_state_dict_repo_id",
         default="Lightricks/LTX-2",
-        type=str,
+        type=none_or_str,
         help="HF Hub repo id with LTX 2.0 checkpoint",
     )
     parser.add_argument(
@@ -682,7 +693,7 @@ def get_args():
     parser.add_argument(
         "--combined_filename",
         default="ltx-2-19b-dev.safetensors",
-        type=str,
+        type=none_or_str,
         help="Filename for combined checkpoint with all LTX 2.0 models (VAE, DiT, etc.)",
     )
     parser.add_argument("--vae_prefix", default="vae.", type=str)
@@ -701,22 +712,25 @@ def get_args():
     parser.add_argument(
         "--text_encoder_model_id",
         default="google/gemma-3-12b-it-qat-q4_0-unquantized",
-        type=str,
+        type=none_or_str,
         help="HF Hub id for the LTX 2.0 base text encoder model",
     )
     parser.add_argument(
         "--tokenizer_id",
         default="google/gemma-3-12b-it-qat-q4_0-unquantized",
-        type=str,
+        type=none_or_str,
         help="HF Hub id for the LTX 2.0 text tokenizer",
     )
     parser.add_argument(
         "--latent_upsampler_filename",
         default="ltx-2-spatial-upscaler-x2-1.0.safetensors",
-        type=str,
+        type=none_or_str,
         help="Latent upsampler filename",
     )
 
+    parser.add_argument(
+        "--timestep_conditioning", action="store_true", help="Whether to add timestep condition to the video VAE model"
+    )
     parser.add_argument("--vae", action="store_true", help="Whether to convert the video VAE model")
     parser.add_argument("--audio_vae", action="store_true", help="Whether to convert the audio VAE model")
     parser.add_argument("--dit", action="store_true", help="Whether to convert the DiT model")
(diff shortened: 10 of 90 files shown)


sayakpaul:
Thanks for your PR. However, it's being worked on in https://github.com/huggingface/diffusers/pull/12721.

src/diffusers/hooks/group_offloading.py
```diff
@@ -296,6 +299,24 @@ def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
...
+            if self.group.onload_leader == module and not self._is_group_on_device():
+                self.group.onload_()
+
+            should_onload_next_group = self.next_group is not None and not self.next_group.onload_self
+            if should_onload_next_group:
+                self.next_group.onload_()
+
+            should_synchronize = (
+                not self.group.onload_self and self.group.stream is not None and not should_onload_next_group
+            )
```
sayakpaul: What if `non_blocking=True`?

src/diffusers/hooks/group_offloading.py
```diff
@@ -324,10 +345,26 @@ def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
...
+    def _is_group_on_device(self) -> bool:
+        tensors = []
+        for group_module in self.group.modules:
+            tensors.extend(list(group_module.parameters()))
+            tensors.extend(list(group_module.buffers()))
+        tensors.extend(self.group.parameters)
+        tensors.extend(self.group.buffers)
+
+        if len(tensors) == 0:
+            return True
```
sayakpaul: This means the group is empty. Why would we return True for this?

src/diffusers/hooks/group_offloading.py
```diff
@@ -549,6 +640,16 @@ def apply_group_offloading(
...
 
+    normalized_pin_groups = pin_groups
+    if isinstance(pin_groups, str):
+        normalized_pin_groups = pin_groups.lower()
+        if normalized_pin_groups not in {"first_last", "all"}:
+            raise ValueError("`pin_groups` must be one of `None`, 'first_last', 'all', or a callable.")
+    elif pin_groups is not None and not callable(pin_groups):
+        raise ValueError("`pin_groups` must be one of `None`, 'first_last', 'all', or a callable.")
+
+    pin_groups = normalized_pin_groups
```
sayakpaul: (nit): would prefer to have a small utility function: `_normalize_pin_groups()`.

src/diffusers/hooks/group_offloading.py
```diff
@@ -576,28 +679,124 @@ def _apply_group_offloading(module: torch.nn.Module, config: GroupOffloadingConf
...
+
+    # Create module groups for ModuleList and Sequential blocks, and explicitly defined block modules
     modules_with_group_offloading = set()
     unmatched_modules = []
     matched_module_groups = []
+
     for name, submodule in module.named_children():
-        if not isinstance(submodule, (torch.nn.ModuleList, torch.nn.Sequential)):
+        # Check if this is an explicitly defined block module
+        if name in block_modules:
```
sayakpaul, line 742:
```suggestion
        if block_modules and name in block_modules:
```

sayakpaul:
Could we resolve conflicts so that it's a bit easier to review? Seems like there's some overlap from https://github.com/huggingface/diffusers/pull/12692.

Aki-07:
Done! Rebased on latest main and resolved conflicts with #12692. Should be much cleaner to review now.

src/diffusers/hooks/group_offloading.py
```diff
@@ -60,8 +60,7 @@ class GroupOffloadingConfig:
     offload_to_disk_path: Optional[str] = None
     stream: Optional[Union[torch.cuda.Stream, torch.Stream]] = None
     block_modules: Optional[List[str]] = None
-    exclude_kwargs: Optional[List[str]] = None
-    module_prefix: Optional[str] = ""
+    pin_groups: Optional[Union[str, Callable]] = None
```
sayakpaul, line 68: This seems like a breaking change. Could you please elaborate?

src/diffusers/hooks/group_offloading.py
```diff
@@ -156,27 +156,27 @@ def _pinned_memory_tensors(self):
         finally:
             pinned_dict = None
 
-    def _transfer_tensor_to_device(self, tensor, source_tensor, default_stream):
+    def _transfer_tensor_to_device(self, tensor, source_tensor):
         tensor.data = source_tensor.to(self.onload_device, non_blocking=self.non_blocking)
         if self.record_stream:
-            tensor.data.record_stream(default_stream)
+            tensor.data.record_stream(self._torch_accelerator_module.current_stream())
```
sayakpaul: Could you elaborate on this change? 
https://github.com/huggingface/diffusers/pull/12721/#issuecomment-3584550990 explains why it's the way it is.

bconstantine:
Thank you for the initial comment! We are working on the solutions right now

src/diffusers/hooks/group_offloading.py
Aki-07: Agreed, that was misleading. Now an empty group returns False so a ‘pinned’ empty group will still onload instead of claiming it’s already on device.

src/diffusers/hooks/group_offloading.py
Aki-07: Added a _normalize_pin_groups helper fn()

src/diffusers/hooks/group_offloading.py
Aki-07: Updated

src/diffusers/hooks/group_offloading.py
Aki-07: Thanks for flagging, have fixed to default it. Sorry for the oversight

src/diffusers/hooks/group_offloading.py
Aki-07: Fixed to the correct behavior from #12721, record the tensor on the consumer/default stream ( captured before entering the transfer stream ) so its lifetime is tied to the forward stream

src/diffusers/hooks/group_offloading.py
Aki-07: Even with non_blocking=True, if a previous group onloaded this one on a side stream, we need a sync before the default stream uses the weights or we risk reading half-copied tensors. I’ve limited the sync to the record_stream=False case, when record_stream=True the tensors are tied to the consumer stream so we can safely skip the sync.

sayakpaul:
@bot /style

Aki-07:
@sayakpaul need help with approval

bconstantine:
@sayakpaul  On my local device there is still a few failed tests on `tests/models/autoencoders/test_models_autoencoder_kl.py` , regarding safetensors I/O serialization error and a few decimal output difference on test_output_pretrained. However I have pulled the recent diffusers main and the same error persists. We are interested to check out whether this error appears on the GitHub workflow tests environment. 

We have fixed the previous error on `tests/models/autoencoders/test_models_autoencoder_wan.py` from the last Github workflow run.

sayakpaul:
Thanks!

Could you enlist the failures you are seeing?

bconstantine:
These were the error logs
```
_____________________________________________________ AutoencoderKLTests.test_layerwise_casting_memory _____________________________________________________

self = <tests.models.autoencoders.test_models_autoencoder_kl.AutoencoderKLTests testMethod=test_layerwise_casting_memory>

    @require_torch_accelerator
    @torch.no_grad()
    def test_layerwise_casting_memory(self):
        MB_TOLERANCE = 0.2
        LEAST_COMPUTE_CAPABILITY = 8.0

        def reset_memory_stats():
            gc.collect()
            backend_synchronize(torch_device)
            backend_empty_cache(torch_device)
            backend_reset_peak_memory_stats(torch_device)

        def get_memory_usage(storage_dtype, compute_dtype):
            torch.manual_seed(0)
            config, inputs_dict = self.prepare_init_args_and_inputs_for_common()
            inputs_dict = cast_maybe_tensor_dtype(inputs_dict, torch.float32, compute_dtype)
            model = self.model_class(**config).eval()
            model = model.to(torch_device, dtype=compute_dtype)
            model.enable_layerwise_casting(storage_dtype=storage_dtype, compute_dtype=compute_dtype)

            reset_memory_stats()
            model(**inputs_dict)
            model_memory_footprint = model.get_memory_footprint()
            peak_inference_memory_allocated_mb = backend_max_memory_allocated(torch_device) / 1024**2

            return model_memory_footprint, peak_inference_memory_allocated_mb

        fp32_memory_footprint, fp32_max_memory = get_memory_usage(torch.float32, torch.float32)
        fp8_e4m3_fp32_memory_footprint, fp8_e4m3_fp32_max_memory = get_memory_usage(torch.float8_e4m3fn, torch.float32)
        fp8_e4m3_bf16_memory_footprint, fp8_e4m3_bf16_max_memory = get_memory_usage(
            torch.float8_e4m3fn, torch.bfloat16
        )

        compute_capability = get_torch_cuda_device_capability() if torch_device == "cuda" else None
        self.assertTrue(fp8_e4m3_bf16_memory_footprint < fp8_e4m3_fp32_memory_footprint < fp32_memory_footprint)
        # NOTE: the following assertion would fail on our CI (running Tesla T4) due to bf16 using more memory than fp32.
        # On other devices, such as DGX (Ampere) and Audace (Ada), the test passes. So, we conditionally check it.
        if compute_capability and compute_capability >= LEAST_COMPUTE_CAPABILITY:
>           self.assertTrue(fp8_e4m3_bf16_max_memory < fp8_e4m3_fp32_max_memory)
E           AssertionError: False is not true

tests\models\test_modeling_common.py:1757: AssertionError
_____________________________________________ AutoencoderKLTests.test_lora_adapter_wrong_metadata_raises_error _____________________________________________ 

self = <tests.models.autoencoders.test_models_autoencoder_kl.AutoencoderKLTests testMethod=test_lora_adapter_wrong_metadata_raises_error>

    @torch.no_grad()
    @unittest.skipIf(not is_peft_available(), "Only with PEFT")
    def test_lora_adapter_wrong_metadata_raises_error(self):
        from peft import LoraConfig

        from diffusers.loaders.lora_base import LORA_ADAPTER_METADATA_KEY
        from diffusers.loaders.peft import PeftAdapterMixin

        init_dict, _ = self.prepare_init_args_and_inputs_for_common()
        model = self.model_class(**init_dict).to(torch_device)

        if not issubclass(model.__class__, PeftAdapterMixin):
            pytest.skip(f"PEFT is not supported for this model ({model.__class__.__name__}).")

        denoiser_lora_config = LoraConfig(
            r=4,
            lora_alpha=4,
            target_modules=["to_q", "to_k", "to_v", "to_out.0"],
            init_lora_weights=False,
            use_dora=False,
        )
        model.add_adapter(denoiser_lora_config)
        self.assertTrue(check_if_lora_correctly_set(model), "LoRA layers not set correctly")

            model.save_lora_adapter(tmpdir)
            model_file = os.path.join(tmpdir, "pytorch_lora_weights.safetensors")
            self.assertTrue(os.path.isfile(model_file))

            # Perturb the metadata in the state dict.
            loaded_state_dict = safetensors.torch.load_file(model_file)
            metadata = {"format": "pt"}
            lora_adapter_metadata = denoiser_lora_config.to_dict()
            lora_adapter_metadata.update({"foo": 1, "bar": 2})
            for key, value in lora_adapter_metadata.items():
                if isinstance(value, set):
                    lora_adapter_metadata[key] = list(value)
            metadata[LORA_ADAPTER_METADATA_KEY] = json.dumps(lora_adapter_metadata, indent=2, sort_keys=True)
>           safetensors.torch.save_file(loaded_state_dict, model_file, metadata=metadata)

tests\models\test_modeling_common.py:1315:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  

tensors = {'decoder.mid_block.attentions.0.to_k.lora_A.weight': tensor([[-0.0836,  0.4591, -0.4989, -0.0175],
        [-0.2300, ...,  0.2900,  0.3015],
        [ 0.2199,  0.0162, -0.3994, -0.0383],
        [ 0.2069,  0.4327, -0.3422, -0.0724]]), ...}
filename = 'C:\\Users\\Bryan\\AppData\\Local\\Temp\\tmpbwk5t9sf\\pytorch_lora_weights.safetensors'
metadata = {'format': 'pt', 'lora_adapter_metadata': '{\n  "alora_invocation_tokens": null,\n  "alpha_pattern": {},\n  "arrow_con...e": null,\n  "trainable_token_indices": null,\n  "use_dora": false,\n  "use_qalora": false,\n  "use_rslora": false\n}'}

    def save_file(
        tensors: Dict[str, torch.Tensor],
        filename: Union[str, os.PathLike],
        metadata: Optional[Dict[str, str]] = None,
    ):
        """
        Saves a dictionary of tensors into raw bytes in safetensors format.

        Args:
            tensors (`Dict[str, torch.Tensor]`):
                The incoming tensors. Tensors need to be contiguous and dense.
            filename (`str`, or `os.PathLike`)):
                The filename we're saving into.
            metadata (`Dict[str, str]`, *optional*, defaults to `None`):
                Optional text only metadata you might want to save in your header.
                For instance it can be useful to specify more about the underlying
                tensors. This is purely informative and does not affect tensor loading.

        Returns:
            `None`

        Example:

        ```python
        import torch

        tensors = {"embedding": torch.zeros((512, 1024)), "attention": torch.zeros((256, 256))}
        save_file(tensors, "model.safetensors")
        ```
        """
>       serialize_file(_flatten(tensors), filename, metadata=metadata)
E       safetensors_rust.SafetensorError: Error while serializing: I/O error: The requested operation cannot be performed on a file with a user-mapped section open. (os error 1224)

C:\Users\Bryan\miniconda3\envs\diffusers_contrib\lib\site-packages\safetensors\torch.py:307: SafetensorError
________________________________________________________ AutoencoderKLTests.test_output_pretrained _________________________________________________________ 

self = <tests.models.autoencoders.test_models_autoencoder_kl.AutoencoderKLTests testMethod=test_output_pretrained>

    def test_output_pretrained(self):
        model = AutoencoderKL.from_pretrained("fusing/autoencoder-kl-dummy")
        model = model.to(torch_device)
        model.eval()

        # Keep generator on CPU for non-CUDA devices to compare outputs with CPU result tensors
        generator_device = "cpu" if not torch_device.startswith(torch_device) else torch_device
        if torch_device != "mps":
            generator = torch.Generator(device=generator_device).manual_seed(0)
        else:
            generator = torch.manual_seed(0)

        image = torch.randn(
            1,
            model.config.in_channels,
            model.config.sample_size,
            model.config.sample_size,
        )
        image = image.to(torch_device)
        with torch.no_grad():
            output = model(image, sample_posterior=True, generator=generator).sample

        output_slice = output[0, -1, -3:, -3:].flatten().cpu()

        # Since the VAE Gaussian prior's generator is seeded on the appropriate device,
        # the expected output slices are not the same for CPU and GPU.
        if torch_device == "mps":
            expected_output_slice = torch.tensor(
                [
                    -4.0078e-01,
                    -3.8323e-04,
                    -1.2681e-01,
                    -1.1462e-01,
                    2.0095e-01,
                    1.0893e-01,
                    -8.8247e-02,
                    -3.0361e-01,
                    -9.8644e-03,
                ]
            )
        elif generator_device == "cpu":
            expected_output_slice = torch.tensor(
                [
                    -0.1352,
                    0.0878,
                    0.0419,
                    -0.0818,
                    -0.1069,
                    0.0688,
                    -0.1458,
                    -0.4446,
                    -0.0026,
                ]
            )
        else:
            expected_output_slice = torch.tensor(
                [
                    -0.2421,
                    0.4642,
                    0.2507,
                    -0.0438,
                    0.0682,
                    0.3160,
                    -0.2018,
                    -0.0727,
                    0.2485,
                ]
            )

>       self.assertTrue(torch_all_close(output_slice, expected_output_slice, rtol=1e-2))

tests\models\autoencoders\test_models_autoencoder_kl.py:171:
E           AssertionError: Max diff is absolute 0.000513467937707901. Diff tensor is tensor([4.3437e-05, 6.3509e-05, 2.4503e-04, 5.1347e-04, 3.8743e-05, 2.0981e-04,
E                   1.7959e-04, 1.4303e-04, 2.2203e-06]).

tests\testing_utils.py:129: AssertionError
------------------------------------------------------------------- Captured stderr call -------------------------------------------------------------------
An error occurred while trying to fetch fusing/autoencoder-kl-dummy: fusing/autoencoder-kl-dummy does not appear to have a file named diffusion_pytorch_model.safetensors.
Defaulting to unsafe serialization. Pass `allow_pickle=False` to raise an error instead.
================================================================= short test summary info ==================================================================
FAILED tests/models/autoencoders/test_models_autoencoder_kl.py::AutoencoderKLTests::test_layerwise_casting_memory - AssertionError: False is not true
FAILED tests/models/autoencoders/test_models_autoencoder_kl.py::AutoencoderKLTests::test_lora_adapter_wrong_metadata_raises_error - safetensors_rust.SafetensorError: Error while serializing: I/O error: The requested operation cannot be performed on a file with a user-mapped section o...
FAILED tests/models/autoencoders/test_models_autoencoder_kl.py::AutoencoderKLTests::test_output_pretrained - AssertionError: Max diff is absolute 0.000513467937707901. Diff tensor is tensor([4.3437e-05, 6.3509e-05, 2.4503e-04, 5.1347e-04, 3.8743e-05, 2.0981e-04,
======================================================== 3 failed, 50 passed, 22 skipped in 19.32s =========================================================
```

where two of them are I/O serialization error, a memory check error (I read the comments that this error usually passes on Ampere and Ada environment, which both are not my current environment), and a slight output difference in test_output_pretrained

bconstantine:
@sayakpaul Also with the current checks, it looks like there is coding style error. Can you help us run the automatic style correction?

sayakpaul:
@bot /style

sayakpaul:
The style bot cannot automatically do it. See:
https://github.com/huggingface/diffusers/actions/runs/20191712208/job/57970027213

I would recommend the following:

1. Create a fresh Python env.
2. Run `pip install - ".[style]"` from the root of the repository directory.
3. Run `make style && make quality`.

Aki-07:
Thanks for the pointer @sayakpaul

sayakpaul:
@Aki-07 @bconstantine I ran those failing tests on my end with this branch and also on `main`. I didn't notice any failures.

bconstantine:
@sayakpaul thankyou for testing! Glad to hear no failures on your environment end.

Aki-07:
Hey @sayakpaul, the WanVACE LoRA failures came from the hook offloading immediately when it was attached. It saved the weights before LoRA was added, then put them back later, so the adapters never took effect. I removed that eager offload so the first offload happens after adapters are loaded. Would need your help to re run the pipelines

src/diffusers/hooks/group_offloading.py
DN6: Not sure I follow here. `_transfer_tensor_to_device` is under context here
https://github.com/huggingface/diffusers/pull/12747/changes#diff-3c991fd8823746cd2455c0fa1334ecc07f407291d31775d617967e83db3c3129R216

So `self._torch_accelerator_module.current_stream()` would be the non-default stream?

src/diffusers/hooks/group_offloading.py
```diff
@@ -709,6 +822,86 @@ def _apply_group_offloading_block_level(module: torch.nn.Module, config: GroupOf
...
+            non_blocking=False,
+            stream=None,
+            record_stream=False,
+            onload_self=True,
+            group_id=f"{config.module_prefix}{module.__class__.__name__}_empty_group",
+        )
+        _apply_group_offloading_hook(module, empty_group, config=config)
+
+
+def _apply_block_offloading_to_submodule(
```
DN6: This was removed in #12692? Why add back in if it's not used anywhere?

src/diffusers/hooks/group_offloading.py
```diff
@@ -337,15 +377,21 @@ def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
                 non_blocking=self.group.non_blocking,
             )
             kwargs.update(moved_kwargs)
-        else:
-            kwargs = send_to_device(kwargs, self.group.onload_device, non_blocking=self.group.non_blocking)
+            return kwargs
+        return send_to_device(kwargs, self.group.onload_device, non_blocking=self.group.non_blocking)
 
-        return args, kwargs
+    def _is_group_on_device(self) -> bool:
```
DN6: Duplicated method names
https://github.com/huggingface/diffusers/pull/12747/changes#diff-3c991fd8823746cd2455c0fa1334ecc07f407291d31775d617967e83db3c3129R361

src/diffusers/hooks/group_offloading.py
```diff
@@ -461,6 +552,17 @@ def pre_forward(self, module, *args, **kwargs):
...
+    if isinstance(pin_groups, str):
+        normalized_pin_groups = pin_groups.lower()
+        if normalized_pin_groups not in {"first_last", "all"}:
+            raise ValueError("`pin_groups` must be one of `None`, 'first_last', 'all', or a callable.")
+        return normalized_pin_groups
+    if pin_groups is not None and not callable(pin_groups):
+        raise ValueError("`pin_groups` must be one of `None`, 'first_last', 'all', or a callable.")
+    return pin_groups
+
+
```
DN6: Not sure there's a huge added benefit with formatting input. We can just raise an error if the pin group is invalid.  

```suggestion
# At top of file
VALID_PIN_GROUPS = {"first_last", "all"}


def _validate_pin_groups(pin_groups: Optional[Union[str, Callable]]) -> Optional[Union[str, Callable]]:
    if pin_groups is None or callable(pin_groups):
        return pin_groups

    if isinstance(pin_groups, str) and pin_groups in VALID_PIN_GROUPS:
        return pin_groups

    raise ValueError(f"`pin_groups` must be None, {', '.join(repr(v) for v in VALID_PIN_GROUPS)}, or a callable.")
```

src/diffusers/hooks/group_offloading.py
```diff
@@ -782,10 +975,32 @@ def _apply_group_offloading_leaf_level(module: torch.nn.Module, config: GroupOff
...
             low_cpu_mem_usage=config.low_cpu_mem_usage,
             onload_self=True,
-            group_id=name,
+            group_id=f"{config.module_prefix}{name}",
         )
         _apply_group_offloading_hook(parent_module, group, config=config)
 
+    # Ensure the top-level module also has a group_offloading hook so hook presence checks pass,
+    # even when it holds no parameters/buffers itself.
+    if config.stream is None:
```
DN6: Why do we need this?

src/diffusers/hooks/group_offloading.py
bconstantine: We have removed this function with the new commit, thanks for pointing this out

src/diffusers/hooks/group_offloading.py
bconstantine: We have erased the duplicate method names for _is_group_on_device

src/diffusers/hooks/group_offloading.py
bconstantine: Thank you for the suggestion! We have changed the normalize input into a simple pin_groups validation

src/diffusers/hooks/group_offloading.py
Aki-07: even when all real groups sit in child modules, the root needs a `group_offloading` hook so the model stays marked as offloaded. That keeps the guardrails working (`_is_group_offload_enabled` still blocks .to()/.cuda() and conflicting offloads, and reapply/remove logic finds the hook). Without it, a wrapper with no params would look un-offloaded and could be moved or re-offloaded into a bad state

src/diffusers/hooks/group_offloading.py
Aki-07: Yes, sorry for the confusion. You are  right inside torch.stream(self.stream) current_stream() is the transfer stream, so we shouldn’t use it. Have fixed it with  capturing the consumer/default stream before entering the transfer stream and passed it down so record_stream(default_stream) ties lifetimes to the forward stream.

bconstantine:
Hi @DN6 @sayakpaul   We’ve updated the fix according to the review. Could you take a quick look and share any feedback when you have a moment? Thank you in advance!

Aki-07:
Hey @DN6 @sayakpaul , As mentioned above, have fixed the comments. Could you help us guide on to the next steps?

src/diffusers/models/autoencoders/autoencoder_kl_wan.py
```diff
@@ -966,6 +966,7 @@ class AutoencoderKLWan(ModelMixin, AutoencoderMixin, ConfigMixin, FromOriginalMo
     # keys toignore when AlignDeviceHook moves inputs/outputs between devices
     # these are shared mutable state modified in-place
     _skip_keys = ["feat_cache", "feat_idx"]
+    _group_offload_block_modules = ["quant_conv", "post_quant_conv", "encoder", "decoder"]
```
sayakpaul: It should also be made a part of `ModelMixin` here no?
https://github.com/huggingface/diffusers/blob/52766e6a6939ac6e74375bde5e19c5e0b90d24c1/src/diffusers/models/modeling_utils.py#L233

We have it for `_skip_keys`:
https://github.com/huggingface/diffusers/blob/52766e6a6939ac6e74375bde5e19c5e0b90d24c1/src/diffusers/models/modeling_utils.py#L254

Preferably initialized to a reasonable default like `None` or something?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1402,6 +1403,9 @@ def enable_group_offload(
                 This option only matters when using streamed CPU offloading (i.e. `use_stream=True`). This can be
                 useful when the CPU memory is a bottleneck but may counteract the benefits of using streams.
             exclude_modules (`Union[str, List[str]]`, defaults to `None`): List of modules to exclude from offloading.
+            pin_groups (`\"first_last\"` | `\"all\"` | `Callable`, *optional*):
+                Optionally keep selected groups on the onload device permanently. See `ModelMixin.enable_group_offload`
```
sayakpaul, line 1432: But we didn't document `pin_groups` in `ModelMixin.enable_group_offload` IIUC? Am I missing something?

tests/hooks/test_group_offloading.py
```diff
@@ -150,72 +150,71 @@ def post_forward(self, module, output):
         return output
 
 
-# Model with only standalone computational layers at top level
-class DummyModelWithStandaloneLayers(ModelMixin):
```
sayakpaul, line 154: Why is this being deleted?

Rest of the diffs in this testing script are a bit difficult to follow honestly. Could we keep this cleaner?

src/diffusers/hooks/group_offloading.py
```diff
@@ -156,7 +158,7 @@ def _pinned_memory_tensors(self):
         finally:
             pinned_dict = None
 
-    def _transfer_tensor_to_device(self, tensor, source_tensor, default_stream):
+    def _transfer_tensor_to_device(self, tensor, source_tensor, default_stream=None):
```
sayakpaul: Why do we have to set the default of `default_stream`?

src/diffusers/hooks/group_offloading.py
```diff
@@ -301,6 +303,27 @@ def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
         if self.group.onload_leader is None:
             self.group.onload_leader = module
 
+        if self.group.pinned:
```
sayakpaul: Let's accompany this block of code under `if self.group.pinned` with an explanatory comment. Additionally, this function can be simplified a bit I think:

```py
def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
    # If there wasn't an onload_leader assigned, assume the first submodule to run forward is the leader.
    if self.group.onload_leader is None:
        self.group.onload_leader = module

    is_leader = (self.group.onload_leader == module)

    # Decide whether this call should run the onload/prefetch/sync orchestration.
    # - pinned: always orchestrate
    # - not pinned: only orchestrate when leader runs
    should_orchestrate = self.group.pinned or is_leader
    should_onload_next_group = self.next_group is not None and not self.next_group.onload_self

    if should_orchestrate:
        # Onload policy differs for pinned vs non-pinned.
        if self.group.pinned:
            if is_leader and not self._is_group_on_device():
                self.group.onload_()
        else:
            if is_leader and self.group.onload_self:
                self.group.onload_()

        if should_onload_next_group:
            self.next_group.onload_()

        should_synchronize = (
            not self.group.onload_self
            and self.group.stream is not None
            and not should_onload_next_group
            and not self.group.record_stream
        )
        if should_synchronize:
            self.group.stream.synchronize()

    args = send_to_device(args, self.group.onload_device, non_blocking=self.group.non_blocking)
    kwargs = self._send_kwargs_to_device(kwargs)
    return args, kwargs

```

src/diffusers/hooks/group_offloading.py
```diff
@@ -291,7 +292,8 @@ def __init__(self, group: ModuleGroup, *, config: GroupOffloadingConfig) -> None
         self.config = config
 
     def initialize_hook(self, module: torch.nn.Module) -> torch.nn.Module:
-        if self.group.offload_leader == module:
+        # For disk offload we materialize the safetensor files upfront so callers can inspect them immediately.
```
sayakpaul: This doesn't seem related to the PR.

src/diffusers/hooks/group_offloading.py
```diff
@@ -301,6 +303,27 @@ def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
         if self.group.onload_leader is None:
             self.group.onload_leader = module
 
+        if self.group.pinned:
+            if self.group.onload_leader == module and not self._is_group_on_device():
+                self.group.onload_()
+
+            should_onload_next_group = self.next_group is not None and not self.next_group.onload_self
```
sayakpaul: Can we reuse https://github.com/huggingface/diffusers/blob/52766e6a6939ac6e74375bde5e19c5e0b90d24c1/src/diffusers/hooks/group_offloading.py#L311?

src/diffusers/hooks/group_offloading.py
```diff
@@ -461,6 +546,19 @@ def pre_forward(self, module, *args, **kwargs):
         return args, kwargs
 
 
+VALID_PIN_GROUPS = {"all", "first_last"}
```
sayakpaul: (nit): could move to the top of the module.

src/diffusers/hooks/group_offloading.py
```diff
@@ -597,11 +701,16 @@ def apply_group_offloading(
         offload_to_disk_path=offload_to_disk_path,
         block_modules=block_modules,
         exclude_kwargs=exclude_kwargs,
+        module_prefix="",
```
sayakpaul, line 703: Why do we have to default to this?

src/diffusers/hooks/group_offloading.py
```diff
@@ -597,11 +701,16 @@ def apply_group_offloading(
...
+        pin_groups=pin_groups,
     )
     _apply_group_offloading(module, config)
 
 
 def _apply_group_offloading(module: torch.nn.Module, config: GroupOffloadingConfig) -> None:
+    registry = HookRegistry.check_if_exists_or_initialize(module)
+    registry._group_offload_pin_groups = config.pin_groups
+
```
sayakpaul: Not sure what the purpose of this is.

src/diffusers/hooks/group_offloading.py
```diff
@@ -709,6 +818,25 @@ def _apply_group_offloading_block_level(module: torch.nn.Module, config: GroupOf
             _apply_group_offloading_hook(module, unmatched_group, config=config)
         else:
             _apply_lazy_group_offloading_hook(module, unmatched_group, config=config)
+    elif config.stream is None and config.offload_to_disk_path is None:
```
sayakpaul: This seems unnecessary. Explain?

src/diffusers/hooks/group_offloading.py
```diff
@@ -735,7 +863,7 @@ def _apply_group_offloading_leaf_level(module: torch.nn.Module, config: GroupOff
             record_stream=config.record_stream,
             low_cpu_mem_usage=config.low_cpu_mem_usage,
             onload_self=True,
-            group_id=name,
+            group_id=f"{config.module_prefix}{name}",
```
sayakpaul, line 842: What's happening here?

tests/hooks/test_group_offloading.py
bconstantine: Thanks for pointing this out. The class was not intentionally deleted. From the git history, this shows up as removed due to branch history / rebase artifacts while integrating changes (rather than a deliberate change to the test itself), which makes the diff noisier than it should be. I’m cleaning this up now: I’ll restore that block and reorganize the commits so the test diffs are more focused/atomic and easier to review.

src/diffusers/models/autoencoders/autoencoder_kl_wan.py
Aki-07: Have added `_group_offload_block_modules = None` to ModelMixin

src/diffusers/pipelines/pipeline_utils.py
Aki-07: You are right pin_groups wasn’t documented on ModelMixin, Have added it

src/diffusers/hooks/group_offloading.py
Aki-07: Done

src/diffusers/hooks/group_offloading.py
Aki-07: done

src/diffusers/hooks/group_offloading.py
Aki-07: done

src/diffusers/hooks/group_offloading.py
Aki-07: It is the same thing as above, we prefix `group_id` with the parent name to avoid collisions (ids) when recursing into block_modules. Root stays empty to preserve existing ids, the prefix only appears when descending into children.

src/diffusers/hooks/group_offloading.py
Aki-07: originally added the empty root hook to tag the top module as offloaded when everything else was matched, but it did not change behaviour, the child hooks already mark the model as group-offloaded and the guardrails rely on those. It just added an empty group and potential extra files, so have removed it to simplify. Functionally nothing depends on it.

src/diffusers/hooks/group_offloading.py
Aki-07: You are right, this was not used.Have dropped it

src/diffusers/hooks/group_offloading.py
Aki-07: it is just an internal label to avoid name collisions when we recurse into explicit block_modules. At the top level we leave it empty so ids stay the same, when we go into a child we prefix its name so two submodules with the same class name don’t clash on group id. It is not exposed to users.

src/diffusers/hooks/group_offloading.py
Aki-07: I didnot understand.. We dropped the eager offload so adapters can load before anything leaves the device. The only eager path left is when `offload_to_disk_path` is set, we write the safetensor files up front so the disk offload flow/tests can see them right after enable_group_offload. In-memory stays lazy. PLease do clarify if my thought process was wrong

src/diffusers/hooks/group_offloading.py
Aki-07: I made it optional because the non-stream path calls _process_tensors_from_modules without a stream, there is nothing to record in that case, and record_stream is gated. None is a safety net for the record call, and it saves passing a placeholder from those call sites. If you prefer the stricter signature, I can keep it required and pass None explicitly where we don’t use streams. please do correct me thru my understanding if this is required to change

Aki-07:
@sayakpaul thank you for all ur comments, sorry for the delay in resolving. All have been answered, Please do let us know ur review

src/diffusers/hooks/group_offloading.py
```diff
@@ -461,6 +530,16 @@ def pre_forward(self, module, *args, **kwargs):
...
 
 
+def _validate_pin_groups(pin_groups: Optional[Union[str, Callable]]) -> Optional[Union[str, Callable]]:
+    if pin_groups is None or callable(pin_groups):
+        return pin_groups
+    if isinstance(pin_groups, str) and pin_groups in VALID_PIN_GROUPS:
+        return pin_groups
+    raise ValueError(
+        f"`pin_groups` must be None, {', '.join(repr(v) for v in sorted(VALID_PIN_GROUPS))}, or a callable."
+    )
```
sayakpaul, line 554:
```suggestion
    elif isinstance(pin_groups, str) and pin_groups not in VALID_PIN_GROUPS:
        raise ValueError(
        f"`pin_groups` must be None, {', '.join(repr(v) for v in sorted(VALID_PIN_GROUPS))}, or a callable."
    )
    return pin_groups
```

src/diffusers/hooks/group_offloading.py
sayakpaul: Let's stick to the existing implementation in this case i.e., a stricter signature.

sayakpaul: Let's stick to a stricter signature.

src/diffusers/hooks/group_offloading.py
sayakpaul: Can you clarify this scenario in the comments as well? And provide a small example that justifies this change?

src/diffusers/hooks/group_offloading.py
```diff
@@ -300,35 +305,48 @@ def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
...
+            if self.group.pinned:
+                if is_leader and not self._is_group_on_device():
+                    self.group.onload_()
+            else:
+                if is_leader and self.group.onload_self:
+                    self.group.onload_()
 
-        # If the current module is the onload_leader of the group, we onload the group if it is supposed
-        # to onload itself. In the case of using prefetching with streams, we onload the next group if
-        # it is not supposed to onload itself.
```
sayakpaul, line 306: (nit): let's not get rid of the important comments.

src/diffusers/hooks/group_offloading.py
```diff
@@ -300,35 +305,48 @@ def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
...
-            should_onload_next_group = self.next_group is not None and not self.next_group.onload_self
             if should_onload_next_group:
                 self.next_group.onload_()
 
             should_synchronize = (
-                not self.group.onload_self and self.group.stream is not None and not should_onload_next_group
+                not self.group.onload_self
+                and self.group.stream is not None
+                and not should_onload_next_group
+                and not self.group.record_stream
```
sayakpaul, line 340: Could I get a clarification on why this condition needs to be modified?

src/diffusers/hooks/group_offloading.py
```diff
@@ -300,35 +305,48 @@ def pre_forward(self, module: torch.nn.Module, *args, **kwargs):
...
+        kwargs = self._send_kwargs_to_device(kwargs)
+        return args, kwargs
+
+    def post_forward(self, module: torch.nn.Module, output):
+        if self.group.pinned:
+            return output
+
+        if self.group.offload_leader == module:
+            self.group.offload_()
+        return output
```
sayakpaul, line 350: This part of the diff reads very confusing to me and hence, a bit hard to confidently review. It seems to me, `post_forward()` was just brought up, `_send_kwargs_to_device()` was added (and I am not sure why) amongst other things. Possible to have a cleaner diff?

src/diffusers/models/modeling_utils.py
```diff
@@ -555,6 +557,11 @@ def enable_group_offload(
             ...     use_stream=True,
             ... )
             ```
+
+        Args:
+            pin_groups (`"first_last"` | `"all"` | `Callable`, *optional*):
+                Optionally keep selected groups on the onload device permanently. See
+                [`~hooks.group_offloading.apply_group_offloading`] for details.
```
sayakpaul: Are we just documenting `pin_groups` here? If so, we should remove that from here as `apply_group_offloading()` should already cover it:

https://github.com/bconstantine/diffusers/blob/335dca80fb2cdbdab7c2daa298c3ca934d7107b6/src/diffusers/models/modeling_utils.py#L542

src/diffusers/models/autoencoders/autoencoder_kl_wan.py
sayakpaul: Let's also add a comment on how these modules were chosen to be included here.

sayakpaul:
@seed93 would you like to test it?

src/diffusers/hooks/group_offloading.py
Aki-07: did not change the condition, those same four checks were already present in both branches. I consolidated them into one place to avoid duplication. We still only sync when the group did not onload itself, we are using a stream, there is no pending prefetch, and `record_stream` is not handling lifetime tracking.

src/diffusers/models/modeling_utils.py
Aki-07: removed redundant docstring

src/diffusers/models/autoencoders/autoencoder_kl_wan.py
Aki-07: sure, added comment

src/diffusers/hooks/group_offloading.py
Aki-07: done

src/diffusers/hooks/group_offloading.py
Aki-07: Sure

src/diffusers/hooks/group_offloading.py
Aki-07: done

src/diffusers/hooks/group_offloading.py
Aki-07: Thanks for the flag, I simplified the diff ( have removed  `_send_kwargs_to_device` and the kwargs handling is back inline in `pre_forward` as before )

Aki-07:
Thanks again @sayakpaul for the detailed review! Have addressed all the points

Aki-07:
@sayakpaul Would u help us on the next steps in this PR

Aki-07:
Thanks @sayakpaul for running CI. Based on the logs, the failures don’t touch the group‑offloading changes in this PR:

1. The WanVACE LoRA failures happen in `test_lora_layers_wanvace.py` and never call `enable_group_offload/apply_group_offloading`. So they’re unrelated to this diff (likely a separate LoRA regression).
2. The VQGAN failure is a timeout in test_vqgan.py::test_vqmodel_checkpointing during accelerate launch, which looks like infra/runtime rather than a code change here.

@DN6 could you please guide on the next steps in the PR, with this been there for a while

bconstantine:
@sayakpaul @DN6 may we ask a quick confirmation whether the current CI failures are independent of the group_offloading changes in this PR?

sayakpaul:
Yes, those are unrelated.

bconstantine:
Hi @sayakpaul , could you share the next steps for this PR?

DN6:
Hi @bconstantine. Apologies for the delay. I was caught up with a few other tasks. I will review this tomorrow.

src/diffusers/hooks/group_offloading.py
```diff
@@ -291,7 +295,12 @@ def __init__(self, group: ModuleGroup, *, config: GroupOffloadingConfig) -> None
         self.config = config
 
     def initialize_hook(self, module: torch.nn.Module) -> torch.nn.Module:
-        if self.group.offload_leader == module:
+        # Disk offload only: materialize the safetensor files up front so they exist right after enable_group_offload.
+        # Needed for flows/tests that inspect the offload dir before the first forward
+        #   eg: model.enable_group_offload(..., offload_to_disk_path=tmpdir)
+        #   assert glob.glob(f"{tmpdir}/*.safetensors")
+        # In-memory offload stays lazy to allow adapter loading before the first forward.
+        if self.group.offload_to_disk_path is not None and self.group.offload_leader == module:
```
DN6: What is the test that's failing? If this change is to make a test pass and it is not related to pinning, let's revert it please. We can create a dedicated PR to address that issue.

src/diffusers/hooks/group_offloading.py
DN6: Instead of using this loop to check all tensors, it would be better to introduce an `_is_offloaded` attribute to the `ModuleGroup` which changes to True/False during onload/offload.  Use that to check if the group is one device or not.

src/diffusers/hooks/group_offloading.py
```diff
@@ -634,7 +732,7 @@ def _apply_group_offloading_block_level(module: torch.nn.Module, config: GroupOf
 
     for name, submodule in module.named_children():
         # Check if this is an explicitly defined block module
-        if name in block_modules:
+        if block_modules and name in block_modules:
```
DN6: ? Change not needed I think? If `block_modules` is an empty set this won't evaluate to true.

src/diffusers/hooks/group_offloading.py
DN6: Let's remove this if it isn't relevant to pinning.

src/diffusers/hooks/group_offloading.py
DN6: What is the failing check that requires this change? It doesn't seem related to pinning?

src/diffusers/hooks/group_offloading.py
```diff
@@ -442,6 +474,50 @@ def post_forward(self, module, output):
             group_offloading_hooks[i].next_group = group_offloading_hooks[i + 1].group
             group_offloading_hooks[i].next_group.onload_self = False
 
+        if self.pin_groups is not None and num_executed > 0:
```
DN6: My understanding here is that this logic is responsible for onloading/pinning the groups during the first inference pass. The first pass of offloading with streams will always be slow since we need to figure out execution order.  The complexity introduced with this change is probably not worth the incremental performance benefit. 

I would suggest we factor this into simpler 

```python
  def post_forward(self, module, output):
      # existing prefetch setup code unchanged

      if self.pin_groups is not None:
          self._mark_pinned_groups(group_offloading_hooks)

      return output
```

After marking, subsequent passes should just skip offloading. The group remains pinned after the second pass. 

Also, what happens with pinning if we don't use streams? The prefetch hook doesn't run in that case.

src/diffusers/hooks/group_offloading.py
Aki-07: Agreed. This eager disk offload was added to satisfy a test unrelated to pinning. have reverted it so offload remains lazy. If we still need eager materialization, I will open a separate PR scoped to that issue

src/diffusers/hooks/group_offloading.py
Aki-07: reverted this check back to if name in `block_modules` since block_modules is already a set and the extra guard is redundant

src/diffusers/hooks/group_offloading.py
Aki-07: refactored the pinning into a `_mark_pinned_groups` helper called from post_forward after the prefetch setup, matching the simplified structure you suggested. Also added a warning + doc note,  `pin_groups` is only supported with use_stream=True, it is ignored otherwise since the lazy prefetch hook doesn’t run without streams

Aki-07:
@DN6 Thanks for the review, I have addressed all your comments. Please advise on the next steps for this PR

bconstantine:
Hi @DN6 @sayakpaul , may we ask what are the next steps for this PR?

Issue #12600, opened 2025-11-06
Add KV Cache for Autoregressive Inference


DN6:
Autoregressive Diffusion Techniques such as [Self Forcing](https://github.com/guandeh17/Self-Forcing/tree/main) rely on a rolling KV Cache across video frame chunks to transfer information from past context frames to the current frames being denoised.

This rolling KV Cache design (or variants similar to it) is likely to show up in other types of long video generation/ world models, so it would be good to see if we can support it natively in Diffusers. 

Tasks
- Implement rolling KV Cache seen in [Self Forcing](https://github.com/guandeh17/Self-Forcing/tree/main) using Diffusers' [cache hooks design](https://github.com/huggingface/diffusers/blob/b3e9dfced7c9e8d00f646c710766b532383f04c6/src/diffusers/hooks/first_block_cache.py#L34). 
- Add a Modular Block to Wan Modular Pipelines that uses this rolling KV Cache to perform autoregressive inference.

tolgacangoz:
I was just about opening this issue; it would be great!
@kuantuna, you might be interested in this since your MAGI-1 PR-continuation is related to KV-Caching too.

sayakpaul:
> Add a Modular Block to Wan Modular Pipelines that uses this rolling KV Cache to perform autoregressive inference.


@DN6 for clarity, to be able to perform inference, the one also needs a trained model right that supports rolling KV caching?

11happy:
Hii @sayakpaul , I would like to work on this as a part of Diffusers MVP program.
Thank you : )

11happy:
Hello @sayakpaul , Here's my proposal:
First Before moving to the implementation I wanted to make sure that I understood Rolling KV cache properly , I went through the linked paper for Self-Forcing https://arxiv.org/abs/2506.08009 & also one another which was referenced inside it for inspiration https://arxiv.org/abs/2309.17453 streaming language model with attention sink:
 
the core algorithm ( sink based ) looks like this:
1. compute queries, keys, value for the current step.
2. identify sink tokens (fixed) 
3. if adding new tokens would exceed the kv cache size, we remove the oldest kv entries after the sink region , shift the remainig to left & insert the new tokens

I also went through the self forcing paper linked codebase & found this relavant code section
Link: https://github.com/guandeh17/Self-Forcing/blob/33593df3e81fa3ec10239271dd2c100facac6de1/wan/modules/causal_model.py#L206
```python
[...]

PR #12773, opened 2025-12-02
[WIP] Rolling KV cache for autoregressive generation

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/issues/12600

Functionality-wise the self attention cache seems to work correctly, cross-attention has to be added and verfied. I added Krea to test the cache though I am not getting the same output as the original model yet. From quick debugging, looked to be related to timesteps or rope embeddings. Opening a draft as a reminder to myself to give this feature higher priority

### DIFF
--- a/src/diffusers/hooks/__init__.py
+++ b/src/diffusers/hooks/__init__.py
@@ -24,5 +24,5 @@ if is_torch_available():
     from .layer_skip import LayerSkipConfig, apply_layer_skip
     from .layerwise_casting import apply_layerwise_casting, apply_layerwise_casting_hook
     from .pyramid_attention_broadcast import PyramidAttentionBroadcastConfig, apply_pyramid_attention_broadcast
-    from .rolling_kv_cache import RollingKVCacheConfig, apply_rolling_kv_cache
+    from .rolling_kv_cache import RollingKVCache
     from .smoothed_energy_guidance_utils import SmoothedEnergyGuidanceConfig
--- a/src/diffusers/hooks/rolling_kv_cache.py
+++ b/src/diffusers/hooks/rolling_kv_cache.py
@@ -12,149 +12,112 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dataclasses import dataclass
-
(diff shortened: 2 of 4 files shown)


yiyixuxu (commented):
Thanks for working on this! Some initial thoughts:

I think this two-level pattern (`RollingKVCache` as container + `CacheLayer` as layer cache) is sufficient to support at least all 3 use cases we've seen: SANA/KREA/GLM-Image, so this is good. We want to keep the API consistent across implementations, e.g., all layer cache classes would have `update`/`reset`/`get` methods and the container would have `__getitem__`, `reset_state`, etc. Not sure if we need a base class though - it might be sufficient for maintainers to enforce consistency during PR reviews.

Where should this live? This isn't a hook, so should be moved out. Not sure if we want a central place or keep them in individual model files. some of the caches are pretty model-specific

cc @DN6 - can you take a look and share some initial thoughts too? I'd like to move quickly with a simple & flexible design and we can adapt as we learn more from our use cases.

zucchini-nlp:
> Where should this live? This isn't a hook, so should be moved out.

yeah, it should be moved out. I think with more models released, it can be stored in a one file and make a base class. Right now the base class would be kinda useless if we have only three models and each one is very specific on how caching is done

So, storing in model file and keeping an eye on consistent implementation for new models sounds good to me. We can generalize easily if API is identical for new models

yiyixuxu:
gentle ping @DN6 here

DN6:
I think keeping it model specific makes sense. The implementations are all quite specialised, and cannot be swapped between models. I think we'd only need to decide where the cache objects live. Currently it's split between pipeline files (SANA, Krea) and model files (GLM Image). 

IMO within the model file feels better to me.

zucchini-nlp:
Thanks for review! Then I'll try to keep an eye on new autoregressive model added or released. I think this PR can be closed now since we don't want to add Krea pipeline, I was using it only for testing purposes. The generation quality imo isn't that good, especially in videos longer than 5 sec

Issue #12776, opened 2025-12-02
Flux2: Tensor tuples can cause issues for checkpointing


dxqb:
### Describe the bug

The modulations calculated here...
https://github.com/huggingface/diffusers/blob/edf36f5128abf3e6ecf92b5145115514363c58e6/src/diffusers/models/transformers/transformer_flux2.py#L716

...return tuples of Tensors:
https://github.com/huggingface/diffusers/blob/edf36f5128abf3e6ecf92b5145115514363c58e6/src/diffusers/models/transformers/transformer_flux2.py#L628

These tuples are passed from outside the transformer blocks into the checkpointed transformer blocks.
If the tensors inside the tuples require gradients, this can cause issues for the backward pass:
```
RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.
```

torch checkpointing doesn't identify the tuples as tensors. Only tensors are identified:
https://github.com/pytorch/pytorch/blob/d38164a545b4a4e4e0cf73ce67173f70574890b6/torch/utils/checkpoint.py#L252

### Reproduction

isolated reproduction code is difficult because of the size of the model. but I'll post a draft PR in a minute.

### Logs

```shell

```

### System Info

torch 2.8, diffusers HEAD

### Who can help?

@DN6 @yiyixuxu @sayakpaul

sayakpaul:
Thanks for the investigation! Wonder why the gradient checkpointing tests didn't catch it up. I just ran `pytest tests/models/transformers/test_models_transformer_flux2.py -k "gradient_checkpointing"` and all three tests passed successfully.

dxqb:
not sure. does this test actually run a backward? asking because that would take 128 GB vram at bf16 for parameters and gradients.

sayakpaul:
It does:
https://github.com/huggingface/diffusers/blob/759ea587082aa0e77449952d8f3523f28ddc61f3/tests/models/test_modeling_common.py#L1039

> does this test actually run a backward? asking because that would take 128 GB vram at bf16 for parameters and gradients.

We are using a smaller variant of the model (but representative enough) so memory requirements aren't that much.

@linoytsaban and I also tried gradient checkpointing in our Flux.2 training script and it didn't cause any issues.

ghost:
[...]

PR #12524, opened 2025-10-21
Sunset Python 3.8 & get rid of explicit `typing` exports where possible

sayakpaul:
# What does this PR do?

Python 3.8 has grown quite old, and I don't think any HF library has it as a minimum requirement.

Sunsetting this version also allows us to take advantage of the implicit typing support (`List[float]` -> `list[float]`, for example).

The main changes to review are in:

* `src/diffusers/pipelines/pipeline_utils.py`
* `src/diffusers/utils/typing_utils.py`

PR #12777, opened 2025-12-02
Flux2: Tensor tuples can cause issues for checkpointing

dxqb:
addresses https://github.com/huggingface/diffusers/issues/12776

# What does this PR do?

This PR keeps the tuples, but moves the splitting from tensors into tuples of tensors to the transformer blocks, to avoid issues with checkpointing. By passing a tensor directly, torch.utils.checkpoint() identifies the tensor and saves it accordingly without running a backward through it multiple times.

This is a draft. If you agree with this change I can make it nicer. Among other things:
 - type hints are incorrect
 - splitting might not be necessary anymore, because they are used immediately after

## Who can review?
@yiyixuxu and @asomoza

### DIFF
--- a/src/diffusers/models/transformers/transformer_flux2.py
+++ b/src/diffusers/models/transformers/transformer_flux2.py
@@ -507,7 +507,9 @@ class Flux2TransformerBlock(nn.Module):
 
         # Modulation parameters shape: [1, 1, self.dim]
         (shift_msa, scale_msa, gate_msa), (shift_mlp, scale_mlp, gate_mlp) = Flux2Modulation.split(temb_mod_img, 2)
-        (c_shift_msa, c_scale_msa, c_gate_msa), (c_shift_mlp, c_scale_mlp, c_gate_mlp) = Flux2Modulation.split(temb_mod_txt, 2)
+        (c_shift_msa, c_scale_msa, c_gate_msa), (c_shift_mlp, c_scale_mlp, c_gate_mlp) = Flux2Modulation.split(
+            temb_mod_txt, 2
+        )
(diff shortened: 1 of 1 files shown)


dg845:
Hi @dxqb, thanks for opening this PR and thanks for your patience! This change looks good to me.  As mentioned in https://github.com/huggingface/diffusers/issues/12776#issuecomment-3614294015, it would be nice to have a small script to reproduce/test this behavior.

dxqb:
> Hi @dxqb, thanks for opening this PR and thanks for your patience! This change looks good to me. As mentioned in [#12776 (comment)](https://github.com/huggingface/diffusers/issues/12776#issuecomment-3614294015), it would be nice to have a small script to reproduce/test this behavior.

no repro-code, but it's clear now why this happens. it's documented by pytorch: https://github.com/huggingface/diffusers/issues/12776#issuecomment-3832054530

dg845 (commented):
Thanks for the changes! Can you solve the merge conflicts with `main`? I think they may be a result of #12524, which switches over to using Python 3.9+ style type hints without explicit `typing` imports, including in `transformer_flux2.py`.

dxqb:
> Thanks for the changes! Can you solve the merge conflicts with `main`? I think they may be a result of #12524, which switches over to using Python 3.9+ style type hints without explicit `typing` imports, including in `transformer_flux2.py`.

done and tested using https://github.com/Nerogar/OneTrainer/pull/1279

dg845:
@bot /style

dg845:
Merging as the CI failure is unrelated.

PR #12792, opened 2025-12-04
Z-Image-Turbo ControlNet

### LINKED ISSUE #12769: Z-Image Turbo Controlnet Union is out
Worth considering?
https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union

hlky:
# What does this PR do?

In the [original code](https://github.com/aigc-apps/VideoX-Fun/blob/main/videox_fun/models/z_image_transformer2d_control.py) this is not a typical ControlNet, it is integrated into the `transformer` and relies on operations performed in the `transformer`'s forward. In this PR we implement it as a typical ControlNet by duplicating the necessary operations from the `transformer`'s forward into the ControlNet's forward and pass `transformer` to `ZImageControlNetModel`'s `forward` to access the necessary `transformer` modules, as a result this is perhaps a little slower than the original implementation, but it keeps things clean and in style. `ZImageTransformer2DModel` has minimal changes, `controlnet_block_samples` is introduced, this is a `Dict[int, torch.Tensor]` returned from `ZImageControlNetModel` where the `int` is the `ZImageTransformer2DModel` `layers` index, this is another difference from typical ControlNet where every block has the ControlNet output applied. `ZImageControlNetPipeline` has minimal changes, compared to `ZImagePipeline` it adds `prepare_image` function, adds `control_image` and `controlnet_conditioning_scale` parameters, prepares and encodes `control_image` and calls `controlnet` to obtain `controlnet_block_samples` which are passed to `transformer`. `control_guidance_start`/`control_guidance_end` is not yet implemented.

<details><summary>Test code</summary>
<p>

```sh
python scripts/convert_z_image_controlnet_to_diffusers.py --original_controlnet_repo_id "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union" --filename "Z-Image-Turbo-Fun-Controlnet-Union.safetensors" --output_path "z-image-controlnet-hf"
```

```python
import torch
from diffusers import ZImageControlNetPipeline
from diffusers import ZImageControlNetModel

controlnet_model = "z-image-controlnet-hf"
controlnet = ZImageControlNetModel.from_pretrained(
    controlnet_model, torch_dtype=torch.bfloat16
)
pipe = ZImageControlNetPipeline.from_pretrained(
    "Tongyi-MAI/Z-Image-Turbo", controlnet=controlnet, torch_dtype=torch.bfloat16
)
pipe = pipe.to("cuda")
control_image = load_image("https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union/resolve/main/asset/pose.jpg?download=true")
prompt = "一位年轻女子站在阳光明媚的海岸线上，白裙在轻拂的海风中微微飘动。她拥有一头鲜艳的紫色长发，在风中轻盈舞动，发间系着一个精致的黑色蝴蝶结，与身后柔和的蔚蓝天空形成鲜明对比。她面容清秀，眉目精致，透着一股甜美的青春气息；神情柔和，略带羞涩，目光静静地凝望着远方的地平线，双手自然交叠于身前，仿佛沉浸在思绪之中。在她身后，是辽阔无垠、波光粼粼的大海，阳光洒在海面上，映出温暖的金色光晕。"
image = pipe(
    prompt,
    control_image=control_image,
    controlnet_conditioning_scale=0.75,
    height=1728,
    width=992,
    num_inference_steps=9,
    guidance_scale=0.0,
    generator=torch.Generator("cuda").manual_seed(43),
).images[0]
image.save("zimage.png")

```

</p>
</details> 

<details><summary>Output</summary>
<p>

| PR | Original |
|--------|--------|
| z-image_controlnet |  |

</p>
</details> 

Fixes #12769

## Who can review?

### DIFF
--- a/src/diffusers/pipelines/z_image/pipeline_z_image_controlnet.py
+++ b/src/diffusers/pipelines/z_image/pipeline_z_image_controlnet.py
@@ -51,22 +51,22 @@ EXAMPLE_DOC_STRING = """
 
         >>> # 2.1
         >>> # controlnet = ZImageControlNetModel.from_single_file(
-        ... #     hf_hub_download(
-        ... #         "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
-        ... #         filename="Z-Image-Turbo-Fun-Controlnet-Union-2.1.safetensors",
-        ... #     ),
-        ... #     torch_dtype=torch.bfloat16,
-        ... # )
+        >>> #     hf_hub_download(
+        >>> #         "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
+        >>> #         filename="Z-Image-Turbo-Fun-Controlnet-Union-2.1.safetensors",
+        >>> #     ),
+        >>> #     torch_dtype=torch.bfloat16,
+        >>> # )
 
         >>> # 2.0 - `config` is required
         >>> # controlnet = ZImageControlNetModel.from_single_file(
-        ... #     hf_hub_download(
-        ... #         "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
-        ... #         filename="Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
-        ... #     ),
-        ... #     torch_dtype=torch.bfloat16,
-        ... #     config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
-        ... # )
+        >>> #     hf_hub_download(
+        >>> #         "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
+        >>> #         filename="Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
+        >>> #     ),
+        >>> #     torch_dtype=torch.bfloat16,
+        >>> #     config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
+        >>> # )
 
         >>> pipe = ZImageControlNetPipeline.from_pretrained(
         ...     "Tongyi-MAI/Z-Image-Turbo", controlnet=controlnet, torch_dtype=torch.bfloat16
--- a/src/diffusers/__init__.py
+++ b/src/diffusers/__init__.py
@@ -667,8 +667,8 @@ else:
             "WuerstchenCombinedPipeline",
             "WuerstchenDecoderPipeline",
             "WuerstchenPriorPipeline",
-            "ZImageControlNetPipeline",
             "ZImageControlNetInpaintPipeline",
+            "ZImageControlNetPipeline",
             "ZImageImg2ImgPipeline",
             "ZImagePipeline",
         ]
--- a/src/diffusers/pipelines/z_image/pipeline_z_image_controlnet_inpaint.py
+++ b/src/diffusers/pipelines/z_image/pipeline_z_image_controlnet_inpaint.py
@@ -52,13 +52,13 @@ EXAMPLE_DOC_STRING = """
 
         >>> # 2.0 - `config` is required
         >>> # controlnet = ZImageControlNetModel.from_single_file(
-        ... #     hf_hub_download(
-        ... #         "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
-        ... #         filename="Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
-        ... #     ),
-        ... #     torch_dtype=torch.bfloat16,
-        ... #     config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
-        ... # )
+        >>> #     hf_hub_download(
+        >>> #         "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
+        >>> #         filename="Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
+        >>> #     ),
+        >>> #     torch_dtype=torch.bfloat16,
+        >>> #     config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
+        >>> # )
 
         >>> pipe = ZImageControlNetInpaintPipeline.from_pretrained(
         ...     "Tongyi-MAI/Z-Image-Turbo", controlnet=controlnet, torch_dtype=torch.bfloat16


src/diffusers/models/controlnets/controlnet_z_image.py
```diff
@@ -0,0 +1,459 @@
...
+            all_image_size,
+            all_image_pos_ids,
+            all_cap_pos_ids,
+            all_image_pad_mask,
+            all_cap_pad_mask,
+        )
+
+    def forward(
+        self,
+        transformer: ZImageTransformer2DModel,
```
yiyixuxu: let's not pass transformer as an input

given that there are some shared layers, we can consider these two alternative design:

* option1: pre-computed shared stuff inside the pipeline, you can add a method to the `ZImageTransformer2DModel` to be used by controlnet if it makes things easier (but no need to change the transformer code) e.g. inside pipeline

```py
... = self.transformer.prepare_inputs(...)
controlnet_block_samples = self.controlnet(control_image=control_image, ...) 
```

* Option2: we can try to inject controlnet into transformer inside `__init__`. similar to https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/animatediff/pipeline_animatediff.py#L140. Basically create a model that combine controlnet + transformer

hlky: Which option would you prefer?

yiyixuxu: let's try option 2

hlky: Done

elismasilva: Hi @hlky , I managed to get your PR working locally, but I modified it to work differently. I unified the weights of the base model's transformer with the transformer from the controlnet model. Initially, I did this because I wanted to generate a unified gguf, since when you load the gguf from_single_file you wouldn't be able to load two transformers at the same time. So I unified them and tested with a single gguf and also with a single pretrained model, and it worked. If this is an easier strategy for users, it will only be necessary to upload a different model to Spaces, such as z-image-turbo-control-hf for e.g.

hlky: Hi @elismasilva, I see, currently if we used gguf for the transformer and controlnet this would not be applied to the combined model when it's created in the pipeline `__init__`, I think we can detect the `quantization_config` and pass that along to the combined model which should fix that, we can also add `from_single_file` support, a conversion script, etc. if users prefer to have a single weight file, but I did find that injecting the controlnet in `__init__` as suggested by @yiyixuxu has increased cpu ram usage from creating the combined transformer+controlnet while the non-combined versions are also loaded, so I wonder if we should go for option 1 instead, we can still add `from_single_file` support to the controlnet for gguf in that case. Let's wait for some more input from YiYi on how best to proceed.

elismasilva: I think the __init__ technique in this case is only valid if the person is going to use the pipeline as a whole. Consider that the ControlNet transformer is not an independent model like those in SDXL, for example; it's an additional weighting of the transformer. So I believe it should be the transformer model's responsibility to append these weights, in case someone tries to load the transformer model in isolation, as is possible, and also considering modular diffusers.

e1ijah1:
How do we support control modes other than Canny? Other control modes still produce poor results—here’s what I got when I tried the HED example from the demo with prompt “A man holding a bottle” and [input image](https://huggingface.co/spaces/AiSudo/ZIT-Controlnet/resolve/main/examples/bottle.jpg).
z2

elismasilva:
> How do we support control modes other than Canny? Other control modes still produce poor results—here’s what I got when I tried the HED example from the demo with prompt “A man holding a bottle” and [input image](https://huggingface.co/spaces/AiSudo/ZIT-Controlnet/resolve/main/examples/bottle.jpg). z2

### Hi, I ran a test with your image and got the following result:

GT: 
image


HED: 
image

Result 1:
Steps: 9
CFG: 0
Control Scale: 0.7
Prompt: A man holding a bottle
image

Result 2:
Steps: 9
CFG: 2.5
Control Scale: 0.75 
Prompt: raw photo, portrait of a handsome Asian man sitting at a wooden table, holding a green glass bottle, wearing a black sweater, wristwatch, highly detailed skin texture, realistic pores, serious gaze, soft cinematic lighting, rim lighting, balanced exposure, 8k uhd, dslr, sharp focus, wood grain texture.
Negative prompt: underexposed, crushed blacks, too dark, heavy shadows, makeup, smooth skin, plastic, wax, cartoon, illustration, distorted hands, bad anatomy, blur, haze, flat lighting.

image

To achieve a realistic effect, you will need to apply the Hires.Fix technique to the image after it has been generated:
like this:
image

scripts/convert_z_image_controlnet_to_diffusers.py
```diff
@@ -0,0 +1,70 @@
...
+    elif args.checkpoint_path is not None:
+        ckpt_path = args.checkpoint_path
+    else:
+        raise ValueError(" please provide either `original_controlnet_repo_id` or a local `checkpoint_path`")
+
+    original_state_dict = safetensors.torch.load_file(ckpt_path)
+    return original_state_dict
+
+
+def convert_z_image_controlnet_checkpoint_to_diffusers(original_state_dict):
```
yiyixuxu: there is no conversion going on? maybe we could just load from the original repo using from_single_file so we don't need to host this seperately

src/diffusers/models/controlnets/controlnet_z_image.py
```diff
@@ -0,0 +1,123 @@
...
+from ...configuration_utils import ConfigMixin, register_to_config
+from ...loaders import PeftAdapterMixin
+from ..controlnets.controlnet import zero_module
+from ..modeling_utils import ModelMixin
+from ..transformers.transformer_z_image import (
+    ZImageTransformerBlock,
+)
+
+
+class ZImageControlTransformerBlock(ZImageTransformerBlock):
```
yiyixuxu: let's not inherit from `ZImageTransformerBlock`, we can just copy over, modify and make a notes about it

src/diffusers/pipelines/z_image/pipeline_z_image_controlnet.py
```diff
@@ -0,0 +1,675 @@
...
+        if isinstance(transformer, ZImageTransformer2DModel):
+            transformer = ZImageControlTransformer2DModel.from_controlnet(transformer, controlnet)
+
+        self.register_modules(
+            vae=vae,
+            text_encoder=text_encoder,
+            tokenizer=tokenizer,
+            scheduler=scheduler,
+            transformer=transformer,
+            controlnet=controlnet,
```
yiyixuxu, line 212: we don't need to keep the original controlnet loaded in the pipeline no?

src/diffusers/models/transformers/transformer_z_image_control.py
```diff
@@ -0,0 +1,836 @@
...
+        config = dict(transformer.config)
+        config["_class_name"] = cls.__name__
+
+        config["control_layers_places"] = controlnet.config["control_layers_places"]
+        config["control_in_dim"] = controlnet.config["control_in_dim"]
+
+        expected_kwargs, optional_kwargs = cls._get_signature_keys(cls)
+        config = FrozenDict({k: config.get(k) for k in config if k in expected_kwargs or k in optional_kwargs})
+        config["_class_name"] = cls.__name__
+        model = cls.from_config(config)
```
yiyixuxu:
```suggestion
        with init_empty_weights():
            model = cls.from_config(config)
```

src/diffusers/models/transformers/transformer_z_image_control.py
```diff
@@ -0,0 +1,836 @@
...
+
+        if not load_weights:
+            return model
+
+        for key, all_x_embedder in transformer.all_x_embedder.items():
+            model.all_x_embedder[key].load_state_dict(all_x_embedder.state_dict())
```
yiyixuxu: Can you directly assign these modules instead? This avoids additional memory at loading since we discard the original transformer/controlnet after this anyway.

src/diffusers/models/controlnets/controlnet_z_image.py
yiyixuxu: >in case someone tries to load the transformer model in isolation, as is possible, and also considering modular diffusers.

are you talking about this?

```py
transformer = ZImageTransformer2DModel.from_pretrained(...)
controlnet = ZImageControlNetModel.from_pretrained(...)
pipe = ZImageControlNetPipeline.from_pretrained(
    "Tongyi-MAI/Z-Image-Turbo", controlnet=controlnet, transforfmer=transformer, torch_dtype=torch.bfloat16
)
```
I believe this PR already handle this use case, no? looking at its `__init__`

```py
 if isinstance(transformer, ZImageTransformer2DModel):
            transformer = ZImageControlTransformer2DModel.from_controlnet(transformer, controlnet)
```


the memory situation isn't ideal, but I think we should be able avoid it, I made a few suggestsion in the PR let's test it out to see if it works

src/diffusers/pipelines/z_image/pipeline_z_image_controlnet.py
yiyixuxu, line 212:
```suggestion
```

yiyixuxu:
ohm I thought about this a bit more. I think we can should try an Option 3 that's a middle ground between Option 1 and 2: instead of combining everything into one model, the controlnet only load shared layers from the transformer. so your `from_transfomer` would look lsomething like this

```py
class ZImageControlNetModel:
    @classmethod
    def from_transformer(cls, controlnet, transformer):
        ....
        controlnet.t_embedder = transformer.t_embedder
        controlnet.all_x_embedder = transformer.all_x_embedder
        controlnet.cap_embedder = transformer.cap_embedder
        return controlnet
```

in pipeline, we still have both controlnet and transformer components and should work similarly to our other contorlnet pipelines

what do you think?

hlky:
@yiyixuxu Option 3 sounds good to me, I've made those changes here [a00f104](https://github.com/huggingface/diffusers/pull/12792/commits/a00f1048e0a214fd09764cff5dd3afb0300fba1d). Let me know if you have any further comments, I will add support for `from_single_file` next.

src/diffusers/loaders/single_file_model.py
```diff
@@ -369,6 +374,9 @@ def from_single_file(cls, pretrained_model_link_or_path_or_dict: Optional[str] =
             diffusers_model_config = config_mapping_fn(
                 original_config=original_config, checkpoint=checkpoint, **config_mapping_kwargs
             )
+        elif "config_create_fn" in mapping_functions:
+            config_create_fn = mapping_functions["config_create_fn"]
+            diffusers_model_config = config_create_fn()
```
hlky: Normally the config is from the Hub, without hosting the Diffusers format weights this seemed a good option, WDYT? Could be useful in other circumstances, like deriving the config from the weights. Alternative would perhaps be making PR to [Hub repo](https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union) with the config?

hlky:
`from_single_file`:

```python
import torch
from diffusers import ZImageControlNetModel

controlnet = ZImageControlNetModel.from_single_file(
    hf_hub_download(
        "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union",
        filename="Z-Image-Turbo-Fun-Controlnet-Union.safetensors",
    ),
    torch_dtype=torch.bfloat16,
)
```

src/diffusers/models/controlnets/controlnet_z_image.py
elismasilva: > >in case someone tries to load the transformer model in isolation, as is possible, and also considering modular diffusers.
> 
> are you talking about this?
> 
> 
Ah yes, but I had something else in mind; I was thinking about eliminating the "controlnet" object as an intermediary for loading in the transformer's from_controlnet(), I thought that ControlNet could just be another internal module in the same model, but for that to happen we would have to rewrite practically the entire `from_pretrained()` function to load all the weights in meta device (I believe that's why there's an increase in memory usage during later loading). But anyway, the suggestion in option 3 is also good; you'll remove the things that are currently mixed together, will give a forward() call and a purpose for the controlnet model to exist, but in this case, there would now be two predictions in the pipeline, and the transformer would receive the samples from the controlnet.

elismasilva:
> `from_single_file`:
> 
> ```python
> import torch
> from diffusers import ZImageControlNetModel
> from huggingface_hub import hf_hub_download
> 
> controlnet = ZImageControlNetModel.from_single_file(
>     hf_hub_download(
>         "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union",
>         filename="Z-Image-Turbo-Fun-Controlnet-Union.safetensors",
>     ),
>     torch_dtype=torch.bfloat16,
> )
> ```
@hlky If you need a ControlNet gguf to test the loading as well, I can try to generate one for you. I'm currently testing this model and trying build an image restoration pipeline, Here is the link to the unified model I mentioned earlier If you want to check it out, go to https://huggingface.co/elismasilva/z-image-control-turbo-unified, but I'll switch to using your implementation later.

src/diffusers/models/controlnets/controlnet_z_image.py
```diff
@@ -0,0 +1,509 @@
...
+from ...configuration_utils import ConfigMixin, register_to_config
+from ...loaders import PeftAdapterMixin
+from ...loaders.single_file_model import FromOriginalModelMixin
+from ...models.attention_processor import Attention
+from ...models.normalization import RMSNorm
+from ...utils.torch_utils import maybe_allow_in_graph
+from ..controlnets.controlnet import zero_module
+from ..modeling_utils import ModelMixin
+from ..transformers.transformer_z_image import (
+    ADALN_EMBED_DIM,
```
yiyixuxu: Can we use `#Copied from` for these shared layers, instead imports?

at least for now, in the future, we may restructure how we arrange our model code and put the controlnet et all in the same folder as the base model and then maybe we can do it differently

src/diffusers/models/controlnets/controlnet_z_image.py
```diff
@@ -0,0 +1,509 @@
...
+        controlnet.cap_embedder = transformer.cap_embedder
+        controlnet.rope_embedder = transformer.rope_embedder
+        controlnet.noise_refiner = transformer.noise_refiner
+        controlnet.context_refiner = transformer.context_refiner
+        controlnet.x_pad_token = transformer.x_pad_token
+        controlnet.cap_pad_token = transformer.cap_pad_token
+        return controlnet
+
+    @staticmethod
+    def create_coordinate_grid(size, start=None, device=None):
```
yiyixuxu, line 493: can we add a #Coped from?

src/diffusers/models/controlnets/controlnet_z_image.py
```diff
@@ -0,0 +1,509 @@
...
+        if start is None:
+            start = (0 for _ in size)
+
+        axes = [torch.arange(x0, x0 + span, dtype=torch.int32, device=device) for x0, span in zip(start, size)]
+        grids = torch.meshgrid(axes, indexing="ij")
+        return torch.stack(grids, dim=-1)
+
+    def patchify_and_embed(
```
yiyixuxu, line 502: here too if it's copied from the z-image transformer

yiyixuxu:
cc @DN6 can you take a look for the single file?

hlky:
I've added `# Copied from` statement and removed imports from z-image transformer

hlky:
[`alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0`](https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0) is released, https://github.com/huggingface/diffusers/pull/12792/commits/04388f4698b303785b26bec6179a55aea652a388 should be ok for the [modeling changes](https://github.com/aigc-apps/VideoX-Fun/commit/3c0f159e016c96fad9b4cab9967aaf627ffdfedd), will add inpaint pipeline and test inference later.

Loading v2 checkpoint is tested:

```python
import torch
from diffusers import ZImageControlNetModel

controlnet = ZImageControlNetModel.from_single_file(
    hf_hub_download(
        "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
        filename="Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
    ),
    torch_dtype=torch.bfloat16,
)
```

Note: This demonstrates expanded usage of introduced `config_create_fn`, `create_z_image_controlnet_config` now uses `checkpoint` to find shape of specific layer and return appropriate fixed config, in this case fixed config makes sense as we cannot reliably determine the exact `control_layers_places` or `control_refiner_layers_places`, but in other cases `config_create_fn` functions could potentially produce the entire config dynamically by checking for existence of certain keys, dimensions of certain weights, number of layers with a certain prefix, etc.

hlky:
<details><summary>1.0</summary>
<p>

```python
import torch
from diffusers import ZImageControlNetPipeline
from diffusers import ZImageControlNetModel

controlnet = ZImageControlNetModel.from_single_file(
    hf_hub_download(
        "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union",
        filename="Z-Image-Turbo-Fun-Controlnet-Union.safetensors",
    ),
    torch_dtype=torch.bfloat16,
)
pipe = ZImageControlNetPipeline.from_pretrained(
    "Tongyi-MAI/Z-Image-Turbo", controlnet=controlnet, torch_dtype=torch.bfloat16
)
pipe = pipe.to("cuda")
control_image = load_image("https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union/resolve/main/asset/pose.jpg?download=true")
prompt = "一位年轻女子站在阳光明媚的海岸线上，白裙在轻拂的海风中微微飘动。她拥有一头鲜艳的紫色长发，在风中轻盈舞动，发间系着一个精致的黑色蝴蝶结，与身后柔和的蔚蓝天空形成鲜明对比。她面容清秀，眉目精致，透着一股甜美的青春气息；神情柔和，略带羞涩，目光静静地凝望着远方的地平线，双手自然交叠于身前，仿佛沉浸在思绪之中。在她身后，是辽阔无垠、波光粼粼的大海，阳光洒在海面上，映出温暖的金色光晕。"
image = pipe(
    prompt,
    control_image=control_image,
    controlnet_conditioning_scale=0.75,
    height=1728,
    width=992,
    num_inference_steps=9,
    guidance_scale=0.0,
    generator=torch.Generator("cuda").manual_seed(43),
).images[0]
image.save("z-image_controlnet-1.png")

```

</p>
</details> 

<details><summary>2.0 T2I</summary>
<p>

Note: 2.0 requires more inference steps, using same prompt as 1.0 here so it's different than the official example which changed prompt between 1.0 and 2.0

```python
import torch
from diffusers import ZImageControlNetPipeline
from diffusers import ZImageControlNetModel

controlnet = ZImageControlNetModel.from_single_file(
    hf_hub_download(
        "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
        filename="Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
    ),
    torch_dtype=torch.bfloat16,
)
pipe = ZImageControlNetPipeline.from_pretrained(
    "Tongyi-MAI/Z-Image-Turbo", controlnet=controlnet, torch_dtype=torch.bfloat16
)
pipe = pipe.to("cuda")
control_image = load_image("https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/resolve/main/asset/pose.jpg?download=true")
prompt = "一位年轻女子站在阳光明媚的海岸线上，白裙在轻拂的海风中微微飘动。她拥有一头鲜艳的紫色长发，在风中轻盈舞动，发间系着一个精致的黑色蝴蝶结，与身后柔和的蔚蓝天空形成鲜明对比。她面容清秀，眉目精致，透着一股甜美的青春气息；神情柔和，略带羞涩，目光静静地凝望着远方的地平线，双手自然交叠于身前，仿佛沉浸在思绪之中。在她身后，是辽阔无垠、波光粼粼的大海，阳光洒在海面上，映出温暖的金色光晕。"
image = pipe(
    prompt,
    control_image=control_image,
    controlnet_conditioning_scale=0.75,
    height=1728,
    width=992,
    num_inference_steps=25,
    guidance_scale=0.0,
    generator=torch.Generator("cuda").manual_seed(43),
).images[0]
image.save("z-image_controlnet-2.png")

```

</p>
</details> 

<details><summary>2.0 Inpaint</summary>
<p>

Note: Using the same prompt as official example, different than prompt in above examples

```python
import torch
from diffusers import ZImageControlNetInpaintPipeline
from diffusers import ZImageControlNetModel

controlnet = ZImageControlNetModel.from_single_file(
    hf_hub_download(
        "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
        filename="Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
    ),
    torch_dtype=torch.bfloat16,
)
pipe = ZImageControlNetInpaintPipeline.from_pretrained(
    "Tongyi-MAI/Z-Image-Turbo", controlnet=controlnet, torch_dtype=torch.bfloat16
)
pipe.to("cuda")
image = load_image(
    "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/resolve/main/asset/inpaint.jpg?download=true"
)
mask_image = load_image(
    "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/resolve/main/asset/mask.jpg?download=true"
)
control_image = load_image(
    "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/resolve/main/asset/pose.jpg?download=true"
)
prompt = "一位年轻女子站在阳光明媚的海岸线上，画面为全身竖构图，身体微微侧向右侧，左手自然下垂，右臂弯曲扶在腰间，她的手指清晰可见，站姿放松而略带羞涩。她身穿轻盈的白色连衣裙，裙摆在海风中轻轻飘动，布料半透、质感柔软。女子拥有一头鲜艳的及腰紫色长发，被海风吹起，在身侧轻盈飞舞，发间系着一个精致的黑色蝴蝶结，与发色形成对比。她面容清秀，眉目精致，肤色白皙细腻，表情温柔略显羞涩，微微低头，眼神静静望向远处的海平线，流露出甜美的青春气息与若有所思的神情。背景是辽阔无垠的海洋与蔚蓝天空，阳光从侧前方洒下，海面波光粼粼，泛着温暖的金色光晕，天空清澈明亮，云朵稀薄，整体色调清新唯美。"
image = pipe(
    prompt,
    image=image,
    mask_image=mask_image,
    control_image=control_image,
    controlnet_conditioning_scale=0.75,
    height=1728,
    width=992,
    num_inference_steps=25,
    guidance_scale=0.0,
    generator=torch.Generator("cuda").manual_seed(43),
).images[0]
image.save("zimage-inpaint.png")

```

</p>
</details> 

| 1.0 | 2.0 | Inpaint |
|--------|--------|--------|
| z-image_controlnet-1 | z-image_controlnet-2 | zimage-inpaint |

All working.

hlky:
We can remove `control_noise_refiner.` from 2.0 state_dict and model as they are unused to save some vram, this did require changing `_should_convert_state_dict_to_diffusers` though as `set(model_state_dict.keys())` is a subset of `set(checkpoint_state_dict.keys())`, this doesn't usually happen as from_single_file checkpoints typically have different keys, so I added a check for exact match of `set(model_state_dict.keys())` and `set(checkpoint_state_dict.keys())` as well. I don't think this will cause any issues with other checkpoint/model types.

elismasilva:
> We can remove `control_noise_refiner.` from 2.0 state_dict and model as they are unused to save some vram, this did require changing `_should_convert_state_dict_to_diffusers` though as `set(model_state_dict.keys())` is a subset of `set(checkpoint_state_dict.keys())`, this doesn't usually happen as from_single_file checkpoints typically have different keys, so I added a check for exact match of `set(model_state_dict.keys())` and `set(checkpoint_state_dict.keys())` as well. I don't think this will cause any issues with other checkpoint/model types.

It seems that this version 2.0 of Z-Image's controlnet will soon be replaced by another version, since they made a mistake training the model without using the model's control_noise_refiner. So I think it's not worth investing much in this version; they are training it again.

iwr-redmond:
See [this note](https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/discussions/4#693eaa4ee4573bfbec779564) for more details.

src/diffusers/pipelines/z_image/pipeline_z_image_controlnet.py
```diff
@@ -0,0 +1,706 @@
...
+        >>> from diffusers import ZImageControlNetPipeline
+        >>> from diffusers import ZImageControlNetModel
+        >>> from diffusers.utils import load_image
+        >>> from huggingface_hub import hf_hub_download
+
+        >>> controlnet = ZImageControlNetModel.from_single_file(
+        ...     hf_hub_download(
+        ...         "alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union",
+        ...         filename="Z-Image-Turbo-Fun-Controlnet-Union.safetensors",
+        ...     ),
```
DN6, line 48: Nit. You can just pass the URL to the file directly here
https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union/blob/main/Z-Image-Turbo-Fun-Controlnet-Union.safetensors

src/diffusers/loaders/single_file_model.py
```diff
@@ -172,11 +174,19 @@
...
 }
 
 
 def _should_convert_state_dict_to_diffusers(model_state_dict, checkpoint_state_dict):
-    return not set(model_state_dict.keys()).issubset(set(checkpoint_state_dict.keys()))
+    model_state_dict_keys = set(model_state_dict.keys())
+    checkpoint_state_dict_keys = set(checkpoint_state_dict.keys())
+    is_subset = model_state_dict_keys.issubset(checkpoint_state_dict_keys)
+    is_match = model_state_dict_keys == checkpoint_state_dict_keys
+    return not (is_subset and is_match)
```
DN6, line 187: Just a question. Is this to handle cases where you'd have transformer + controlnet in a single checkpoint?

src/diffusers/loaders/single_file_utils.py
```diff
@@ -3885,3 +3889,52 @@ def update_state_dict(state_dict: dict[str, object], old_key: str, new_key: str)
             handler_fn_inplace(key, converted_state_dict)
 
     return converted_state_dict
+
+
+def create_z_image_controlnet_config(checkpoint, **kwargs):
```
DN6: Can we host these (either on the official model repo or create a community repo for them) and fetch from a model repo like we do for all the other configs? Why hardcode in src?

src/diffusers/loaders/single_file_model.py
hlky: See https://github.com/huggingface/diffusers/pull/12792#issuecomment-3649491719

src/diffusers/loaders/single_file_utils.py
hlky: See https://github.com/huggingface/diffusers/pull/12792#issuecomment-3649224557 and https://github.com/huggingface/diffusers/pull/12792#discussion_r2608997735
Hosting them is an option, if we can get it merged to the official model repo, and someone could create a community repo.
The idea here was to avoid that:
a) merging PRs to official model repos can take time
b) there would be no conversion of keys for Diffusers format weights so it's better for the community to just use `from_single_file` with the official checkpoint, a community repo with just the configs might cause confusion in that case if users tried it with `from_pretrained`
Also, as mentioned, this `create_config` functionality could be useful in the future with other cases.

hlky:
Thanks @elismasilva @iwr-redmond, I saw that, we will have to see what happens, for now 2.0 works - if another version is released I'll update this PR or make another.

Also, if you're curious about the potential speed difference of using `control_refiner_layers` - revert [dd9775c](https://github.com/huggingface/diffusers/pull/12792/commits/dd9775caf0906c6727fcbe2797cf6dc60cc38f45) and edit `self.control_layers` to `self.control_refiner_layers` https://github.com/huggingface/diffusers/blob/dd9775caf0906c6727fcbe2797cf6dc60cc38f45/src/diffusers/models/controlnets/controlnet_z_image.py#L690-L697

In my testing it was not much faster, so unless they change anything else don't expect a big improvement to inference times.

tin2tin:
2.1 is up: https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/tree/main

hlky:
[24f454c](https://github.com/huggingface/diffusers/pull/12792/commits/24f454ced2ca7366146d269b7ff0fd2bb03f744a)

I've changed `add_control_noise_refiner` from `bool` to `Literal["control_layers", "control_noise_refiner"]` where `control_layers` is `2.0` and `control_noise_refiner` is `2.1`.

Keys in the weights are the same between 2.0 and 2.1, we have no way to distinguish between them other than passing a config, so I've created [hlky/Z-Image-Turbo-Fun-Controlnet-Union](https://huggingface.co/hlky/Z-Image-Turbo-Fun-Controlnet-Union), [hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0](https://huggingface.co/hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0) and [hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.1](https://huggingface.co/hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.1). These include the weights, so we can also use `from_pretrained`. For `from_single_file` 1.0 vs 2.x is detected, and passing `config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0"` is only required for 2.0, the default config is for 2.1. `config_create_fn` related code is removed.

```python
import torch
from diffusers import ZImageControlNetModel

controlnet = ZImageControlNetModel.from_single_file(
    "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union/blob/main/Z-Image-Turbo-Fun-Controlnet-Union.safetensors",
    torch_dtype=torch.bfloat16,
)


controlnet = ZImageControlNetModel.from_single_file(
    "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/blob/main/Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
    torch_dtype=torch.bfloat16,
    config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
)


controlnet = ZImageControlNetModel.from_single_file(
    "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/blob/main/Z-Image-Turbo-Fun-Controlnet-Union-2.1.safetensors",
    torch_dtype=torch.bfloat16,
)
```

elismasilva:
> [24f454c](https://github.com/huggingface/diffusers/pull/12792/commits/24f454ced2ca7366146d269b7ff0fd2bb03f744a)
> 
> I've changed `add_control_noise_refiner` from `bool` to `Literal["control_layers", "control_noise_refiner"]` where `control_layers` is `2.0` and `control_noise_refiner` is `2.1`.
> 
> Keys in the weights are the same between 2.0 and 2.1, we have no way to distinguish between them other than passing a config, so I've created [hlky/Z-Image-Turbo-Fun-Controlnet-Union](https://huggingface.co/hlky/Z-Image-Turbo-Fun-Controlnet-Union), [hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0](https://huggingface.co/hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0) and [hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.1](https://huggingface.co/hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.1). These include the weights, so we can also use `from_pretrained`. For `from_single_file` 1.0 vs 2.x is detected, and passing `config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0"` is only required for 2.0, the default config is for 2.1. `config_create_fn` related code is removed.
> 
> ```python
> import torch
> from diffusers import ZImageControlNetModel
> 
> controlnet = ZImageControlNetModel.from_single_file(
>     "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union/blob/main/Z-Image-Turbo-Fun-Controlnet-Union.safetensors",
>     torch_dtype=torch.bfloat16,
> )
> 
> 
> controlnet = ZImageControlNetModel.from_single_file(
>     "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/blob/main/Z-Image-Turbo-Fun-Controlnet-Union-2.0.safetensors",
>     torch_dtype=torch.bfloat16,
>     config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0",
> )
> 
> 
> controlnet = ZImageControlNetModel.from_single_file(
>     "https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/blob/main/Z-Image-Turbo-Fun-Controlnet-Union-2.1.safetensors",
>     torch_dtype=torch.bfloat16,
> )
> ```

ive made a commentary here https://github.com/aigc-apps/VideoX-Fun/pull/404.

I think he's willing to make the model available correctly according to the standard; if I'm wrong, you can help out there.

hlky:
@elismasilva I don't think there's any need for them to make changes to their own repo, at most we should make a PR with the config only to their Hub repos, but there'd still need to be another Hub repo for the 2.0 vs 2.1 config.

For now, I think it is fine, we have Hub repos for all versions which can be used with `from_pretrained`, and the original weights can be used with `from_single_file`, the only caveat is that users need to pass `config="hlky/Z-Image-Turbo-Fun-Controlnet-Union-2.0"` for 2.0.

Edit: We could detect 2.0 vs 2.1 based on the contents of weights then we could PR a config to the original Hub repo and switch the 2.1 or 2.0 config based on weight content detection, but I think it's ok the way it is just uploading configs to the Hub for each version, we get `from_pretrained` support this way too.

elismasilva:
@hlky Ah yes, actually I only suggested it in case he also wants to align with that standard. Locally I did what you said, I removed the weights from the 2.0 refiner just to make it lighter. But I think in his case I don't see the need for a 2.1, just the updated 2.0 with the correct weights.

hlky:
Prompt: `一位年轻女子站在阳光明媚的海岸线上，白裙在轻拂的海风中微微飘动。她拥有一头鲜艳的紫色长发，在风中轻盈舞动，发间系着一个精致的黑色蝴蝶结，与身后柔和的蔚蓝天空形成鲜明对比。她面容清秀，眉目精致，透着一股甜美的青春气息；神情柔和，略带羞涩，目光静静地凝望着远方的地平线，双手自然交叠于身前，仿佛沉浸在思绪之中。在她身后，是辽阔无垠、波光粼粼的大海，阳光洒在海面上，映出温暖的金色光晕。`
[Control image](https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/resolve/main/asset/pose.jpg?download=true)

Note: Parameters (and seed) match original code, but the same prompt is used here, in the original code examples the prompt is changed between versions.

1.0: 9 steps

2.x: 25 steps

Scale: 0.75

A40 (runpod)

| 1.0 | 2.0 | 2.1 |
|--------|--------|--------|
| z-image_controlnet-1 0 | z-image_controlnet-2 0 | z-image_controlnet-2 1 |
| [00:18<00:00,  2.02s/it] | [01:21<00:00,  3.24s/it] | [01:04<00:00,  2.57s/it] | 

Prompt: `一位年轻女子站在阳光明媚的海岸线上，白裙在轻拂的海风中微微飘动。她拥有一头鲜艳的紫色长发，在风中轻盈舞动，发间系着一个精致的黑色蝴蝶结，与身后柔和的蔚蓝天空形成鲜明对比。她面容清秀，眉目精致，透着一股甜美的青春气息；神情柔和，略带羞涩，目光静静地凝望着远方的地平线，双手自然交叠于身前，仿佛沉浸在思绪之中。在她身后，是辽阔无垠、波光粼粼的大海，阳光洒在海面上，映出温暖的金色光晕。`
[Control image](https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/resolve/main/asset/pose.jpg?download=true)
[Inpaint image](https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/resolve/main/asset/inpaint.jpg?download=true)
[Mask image](https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/resolve/main/asset/mask.jpg?download=true)

| 2.0 | 2.1 |
|--------|--------|
| zimage-inpaint-2 0 | zimage-inpaint-2 1 |
| [01:22<00:00,  3.29s/it] | [01:05<00:00,  2.60s/it] | 


Relatively there is not much speed difference between 2.0 and 2.1, for text-to-image case both 2.0 and 2.1 seem to perform poorly with hands, same for inpaint case but 2.1 does better colour matching.

Also just to note these results do not match the results from the [Hub repo](https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0) but neither do the results when using the original code - I assume a different prompt, scale or number of inference steps was used for the Hub results.

yiyixuxu:
@bot /style

PR #12808, opened 2025-12-09
[Modular]z-image

yiyixuxu:
```py
import torch


# 1. Load the pipeline
# Use bfloat16 for optimal performance on supported GPUs
pipe = ModularPipeline.from_pretrained("Tongyi-MAI/Z-Image-Turbo")
print(f" pipe: {pipe}")
pipe.load_components(torch_dtype=torch.bfloat16)
pipe.to("cuda")
# [Optional] Attention Backend
# Diffusers uses SDPA by default. Switch to Flash Attention for better efficiency if supported:
# pipe.transformer.set_attention_backend("flash")    # Enable Flash-Attention-2
# pipe.transformer.set_attention_backend("_flash_3") # Enable Flash-Attention-3

# [Optional] Model Compilation
# Compiling the DiT model accelerates inference, but the first run will take longer to compile.
# pipe.transformer.compile()

# [Optional] CPU Offloading
# Enable CPU offloading for memory-constrained devices.
# pipe.enable_model_cpu_offload()

prompt = "Young Chinese woman in red Hanfu, intricate embroidery. Impeccable makeup, red floral forehead pattern. Elaborate high bun, golden phoenix headdress, red flowers, beads. Holds round folding fan with lady, trees, bird. Neon lightning-bolt lamp (⚡️), bright yellow glow, above extended left palm. Soft-lit outdoor night background, silhouetted tiered pagoda (西安大雁塔), blurred colorful distant lights."

# 2. Generate Image
## text-to-image
out1 = pipe(
    prompt=prompt,
    height=1024,
    width=1024,
    generator=torch.Generator("cuda").manual_seed(42),
).images[0]

out1.save("t2i_output.png")

## image-to-image
prompt = "Young Chinese woman in red Hanfu, holding an apple"
strength = 0.8
out2 = pipe(
    image=out1,
    prompt=prompt,
    strength=strength,
    generator=torch.Generator("cuda").manual_seed(42),
).images[0]
out2.save(f"i2i_output_strength_{strength}.png")
```
yiyi_test_7_2_t2i_output
yiyi_test_7_2_i2i_output_strength_0 8

### DIFF
--- a/src/diffusers/modular_pipelines/z_image/before_denoise.py
+++ b/src/diffusers/modular_pipelines/z_image/before_denoise.py
@@ -43,32 +43,32 @@ def repeat_tensor_to_batch_size(
 ) -> torch.Tensor:
     """Repeat tensor elements to match the final batch size.
 
-    This function expands a tensor's batch dimension to match the final batch size (batch_size * num_videos_per_prompt)
+    This function expands a tensor's batch dimension to match the final batch size (batch_size * num_images_per_prompt)
     by repeating each element along dimension 0.
 
     The input tensor must have batch size 1 or batch_size. The function will:
-    - If batch size is 1: repeat each element (batch_size * num_videos_per_prompt) times
-    - If batch size equals batch_size: repeat each element num_videos_per_prompt times
+    - If batch size is 1: repeat each element (batch_size * num_images_per_prompt) times
+    - If batch size equals batch_size: repeat each element num_images_per_prompt times
 
     Args:
         input_name (str): Name of the input tensor (used for error messages)
         input_tensor (torch.Tensor): The tensor to repeat. Must have batch size 1 or batch_size.
         batch_size (int): The base batch size (number of prompts)
-        num_videos_per_prompt (int, optional): Number of videos to generate per prompt. Defaults to 1.
+        num_images_per_prompt (int, optional): Number of images to generate per prompt. Defaults to 1.
 
     Returns:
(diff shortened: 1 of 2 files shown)


src/diffusers/modular_pipelines/z_image/modular_blocks.py
```diff
@@ -0,0 +1,191 @@
...
+        )
+
+
+## auto blocks
+class ZImageAutoDenoiseStep(AutoPipelineBlocks):
+    block_classes = [
+        ZImageImage2ImageCoreDenoiseStep,
+        ZImageCoreDenoiseStep,
+    ]
+    block_names = ["image2image", "text2image"]
```
asomoza, line 105: here the previous models had a denoise in the name of the blocks, this is useful for adding the preview block with Mellon

asomoza: actually after looking more into it, all the other models had just one CoreDenoiseStep but this one has a separate one for t2i and i2i, so that's also a problem. I'll skip the preview block for now

yiyixuxu: yeah we will make it work, it all has a `denoise` block, just have to do a search until we find it. it is better if we goive more flexibility in terms on how one should pack the blocks, currently, the only constraint is to have top-level auto blocks like this

```py
AUTO_BLOCKS = InsertableDict(
    [
        ("text_encoder", ...),
        ("vae_image_encoder", ...),
        ("denoise", ...),
        ("decode", ...),
    ]
)
```

PR #12810, opened 2025-12-09
[docs] improve distributed inference cp docs.

sayakpaul:
# What does this PR do?

Our CP docs could benefit from a bit more detail and ready-to-run examples.

* Change to model checkpoint Flux.1-Dev because Qwen isn't supported for parallel yet (https://github.com/huggingface/diffusers/issues/12568).
* Add a command that should be used to invoke parallel inference.
* Simplify the example code snippet.
* Expand on how to set `parallel_config`.

### DIFF
--- a/docs/source/en/training/distributed_inference.md
+++ b/docs/source/en/training/distributed_inference.md
@@ -237,8 +237,7 @@ By selectively loading and unloading the models you need at a given stage and sh
 
 Use [`~ModelMixin.set_attention_backend`] to switch to a more optimized attention backend. Refer to this [table](../optimization/attention_backends#available-backends) for a complete list of available backends.
 
-> [!NOTE] 
-> Most attention backends are compatible with context parallelism. If one is not compatibel with context parallelism, please [file a feature request](https://github.com/huggingface/diffusers/issues/new).
+Most attention backends are compatible with context parallelism. Open an [issue](https://github.com/huggingface/diffusers/issues/new) if a backend is not compatible.
 
 ### Ring Attention
 
@@ -296,7 +295,11 @@ if __name__ == "__main__":
     main()
 ```
 
-The script above needs to be run with a distributed launcher that is compatible with PyTorch. You can use `torchrun` for this: `torchrun --nproc-per-node 2 above_script.py`. `--nproc-per-node` depends on the number of GPUs available.
+The script above needs to be run with a distributed launcher, such as [torchrun](https://docs.pytorch.org/docs/stable/elastic/run.html), that is compatible with PyTorch. `--nproc-per-node` is set to the number of GPUs available.
+
+/```shell
+`torchrun --nproc-per-node 2 above_script.py`. 
+/```
 
 ### Ulysses Attention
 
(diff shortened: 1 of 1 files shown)


docs/source/en/training/distributed_inference.md
```diff
@@ -237,6 +237,9 @@ By selectively loading and unloading the models you need at a given stage and sh
 
 Use [`~ModelMixin.set_attention_backend`] to switch to a more optimized attention backend. Refer to this [table](../optimization/attention_backends#available-backends) for a complete list of available backends.
 
+> [!NOTE] 
+> Most attention backends are compatible with context parallelism. If one is not compatibel with context parallelism, please [file a feature request](https://github.com/huggingface/diffusers/issues/new).
```
stevhliu: If most are compatible, I wonder if it'll be easier to note which ones aren't compatible? I think this way, users will have a better experience if they know ahead of time.

```suggestion
Most attention backends are compatible with context parallelism. Open an [issue](https://github.com/huggingface/diffusers/issues/new) if a backend is not compatible.
```

docs/source/en/training/distributed_inference.md
```diff
@@ -245,40 +248,56 @@ Pass a [`ContextParallelConfig`] to the `parallel_config` argument of the transf
...
+
+    if dist.is_initialized():
+        dist.destroy_process_group()
+
+
+if __name__ == "__main__":
+    main()
 ```
 
+The script above needs to be run with a distributed launcher that is compatible with PyTorch. You can use `torchrun` for this: `torchrun --nproc-per-node 2 above_script.py`. `--nproc-per-node` depends on the number of GPUs available.
```
stevhliu:
```suggestion
The script above needs to be run with a distributed launcher, such as [torchrun](https://docs.pytorch.org/docs/stable/elastic/run.html), that is compatible with PyTorch. `--nproc-per-node` is set to the number of GPUs available.

/```shell
`torchrun --nproc-per-node 2 above_script.py`. 
/```
```

docs/source/en/training/distributed_inference.md
```diff
@@ -288,5 +307,26 @@ finally:
 Pass the [`ContextParallelConfig`] to [`~ModelMixin.enable_parallelism`].
 
 ```py
+# Depending on the number of GPUs available.
 pipeline.transformer.enable_parallelism(config=ContextParallelConfig(ulysses_degree=2))
+```
+
+### parallel_config
+
+It's possible to pass a `ContextParallelConfig` to `parallel_config` during initializing a model and a pipeline:
```
stevhliu: It'd be nice to clarify _why_ a user would want to initialize a model/pipeline with context parallelism versus doing it later. For example, less overhead to reconfigure the model earlier?

```suggestion
Pass `parallel_config` during model initialization to enable context parallelism.
```

stevhliu (commented):
Thanks for improving!

docs/source/en/training/distributed_inference.md
sayakpaul: It's just a preference, I guess. I don't think it will have to do much with overhead.

PR #12820, opened 2025-12-11
[wip] [core] introduce pipeline-specific mixins to hold common methods

sayakpaul:
# What does this PR do?

As discussed internally, this PR introduces pipeline-specific Mixin classes that hold common methods shared across different task-specific pipelines. The actual pipelines (such as `StableDiffusionPipeline`, `QwenImagePipeline`, `QwenImageImg2ImgPipeline`, etc.) would then subclass from these mixins.

Examples of such common methods vary from pipeline to pipeline, but some examples include `encode_prompt()`, properties such as `guidance_scale`, etc.

Additionally, we have a couple of methods like `retrieve_latents()` and `retrieve_timesteps()` originated in `stable_diffusion` and we copied them over to other pipelines that use these methods. Same for `calculate_shift` that originated in `flux`. This PR also considers this situation and follows a reasonable approach to treat it (more below).

I know we also discussed moving the specific LoraLoaderMixins to their respective pipeline modules. The PR only does that for Stable Diffusion to gather feedback. Please keep on reading.

> [!NOTE]
> This PR introduces Mixin classes for four popular pipelines: Flux, Qwen, SDXL, and SD. Once the PR is merged, we could work with the community to do this for other influential pipelines.

---

## Guiding principles

A sample structure for `src/diffusers/pipelines/flux`:

```bash
.
├── __init__.py
├── modeling_flux.py
├── pipeline_flux_control_img2img.py
├── pipeline_flux_control_inpaint.py
├── pipeline_flux_control.py
├── pipeline_flux_controlnet_image_to_image.py
├── pipeline_flux_controlnet_inpainting.py
├── pipeline_flux_controlnet.py
├── pipeline_flux_fill.py
├── pipeline_flux_img2img.py
├── pipeline_flux_inpaint.py
├── pipeline_flux_kontext_inpaint.py
├── pipeline_flux_kontext.py
├── pipeline_flux_prior_redux.py
├── pipeline_flux_utils.py <------ holds the Mixin
├── lora_utils.py <------ holds the `FluxLoraLoaderMixin`
├── pipeline_flux.py
└── pipeline_output.py
```

`pipeline_flux_utils.py` here has the Mixin class, holding the common methods. It additionally has the `calculate_shift()`  method, which is imported by the other pipelines under `src/diffusers/pipelines/flux`. Those pipelines won't have to maintain `# Copied from ...` versions of the `calculate_shift()` method anymore. They can just import like `from .pipeline_flux_utils import calculate_shift`. This is how we reduce LOCs, further on an intra-pipeline basis.

However, if a new pipeline (other than Flux), wants to leverage `calculate_shift()`, it should still follow `# Copied from ...` versions of the `calculate_shift()` instead of a direct import. This way, we can let individual pipeline-level methods evolve independently without introducing complex dependency patterns. For example, `pipeline_flux_utils.py` uses the `# Copied from ...` versions of methods like `retrieve_latents()`.

### LoRA

For LoRA, I have limited the changes to only Flux. Once we settle on that, I can propagate that change for other pipelines tackled in this PR.

---

## Notes

* In the Qwen mixin, kept `encode_prompt()` varying while unifying the `_get_qwen_prompt_embeds()` method in `QwenImageMixin`. Comments are in line.
* There are a few methods that are shared by 2/3 pipelines such as `prepare_image()` is shared by `QwenImageControlNetInpaintPipeline` and `QwenImageControlNetPipeline` but not others. We can choose to move them to `pipeline_qwen_utils.py` but I am not strongly opinionated on this. The same applies to methods like `prepare_mask_latents()`. My preference here is to keep them as is for now.

We can take similar principles for Flux, too. Flux Control and ControlNet pipelines have a `prepare_image()` method. So, we could have something like `FluxControlMixin(FluxMixin)` and include `prepare_image()` there. But I felt like it was getting complex so, decided to skip.

---

## Additional questions

We have 
https://github.com/huggingface/diffusers/blob/6bf668c4d217ebc96065e673d8a257fd79950d34/src/diffusers/pipelines/pipeline_utils.py#L2175
This PR also has a similar Mixin class but without the methods from the above Mixin. To avoid confusion, I have named it `SDMixin`. 

WDYT about collating the two classes (`StableDiffusionMixin` and `SDMixin` (introduced in this PR)) into one?

### DIFF
--- a/examples/community/pipeline_flux_differential_img2img.py
+++ b/examples/community/pipeline_flux_differential_img2img.py
@@ -526,7 +526,7 @@ class FluxDifferentialImg2ImgPipeline(DiffusionPipeline, FluxLoraLoaderMixin):
         return latents
 
     @staticmethod
-    # Copied from diffusers.pipelines.flux.pipeline_flux.FluxPipeline._unpack_latents
+    # Copied from diffusers.pipelines.flux.pipeline_flux_utils.FluxMixin._unpack_latents
     def _unpack_latents(latents, height, width, vae_scale_factor):
         batch_size, num_patches, channels = latents.shape
 
--- a/examples/community/pipeline_flux_kontext_multiple_images.py
+++ b/examples/community/pipeline_flux_kontext_multiple_images.py
@@ -591,7 +591,7 @@ class FluxKontextPipeline(
         return latents
 
     @staticmethod
-    # Copied from diffusers.pipelines.flux.pipeline_flux.FluxPipeline._unpack_latents
+    # Copied from diffusers.pipelines.flux.pipeline_flux_utils.FluxMixin._unpack_latents
     def _unpack_latents(latents, height, width, vae_scale_factor):
         batch_size, num_patches, channels = latents.shape
 
--- a/examples/community/pipeline_flux_semantic_guidance.py
+++ b/examples/community/pipeline_flux_semantic_guidance.py
@@ -664,7 +664,7 @@ class FluxSemanticGuidancePipeline(
         return latents
 
     @staticmethod
-    # Copied from diffusers.pipelines.flux.pipeline_flux.FluxPipeline._unpack_latents
+    # Copied from diffusers.pipelines.flux.pipeline_flux_utils.FluxMixin._unpack_latents
     def _unpack_latents(latents, height, width, vae_scale_factor):
         batch_size, num_patches, channels = latents.shape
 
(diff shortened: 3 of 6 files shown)


sayakpaul:
@yiyixuxu @dn6 I believe this is ready for an initial review.

If the PR is too heavy to review, I can also simplify its scope just to do something like

```bash
|-- src/diffusers/pipelines/stable_diffusion/utils.py (containing `retrieve_latents`, `retrieve_timesteps`, etc.)
|-- ...
|-- src/diffusers/pipelines/flux/utils.py (containing `calculate_shift`, etc.)
```

Now, the pipelines within the `src/diffusers/pipelines/stable_diffusion/` module will import the functions from `utils`. Pipelines outside the module will continue to maintain the "Copied from ..." versions.

I think this is a much more reasonable starting point?

PR #12829, opened 2025-12-12
support CP in native flash attention

sywangyi:
# What does this PR do?

- General functionalities: @sayakpaul @yiyixuxu @DN6

native flash attention could support both Ulysses and ring attention

PR #12832, opened 2025-12-12
[core] gracefully error out when attn-backend x cp combo isn't supported.

sayakpaul:
# What does this PR do?

We should be able to error out when an attention backend isn't supported with CP.

Refer to https://github.com/huggingface/diffusers/pull/12829#issuecomment-3645237672 and https://github.com/huggingface/diffusers/pull/12829#issuecomment-3645582823.

Additionally, when we specify `parallel_config` to `from_pretrained()` and then set an attention backend through `set_attention_backend()` (which is CP-incompatible), it goes through silently. 

This PR eliminates those and gracefully errors out.

### DIFF
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -630,7 +630,7 @@ class ModelMixin(torch.nn.Module, PushToHubMixin):
             raise ValueError(
                 f"Context parallelism is enabled but current attention backend '{backend.value}' "
                 f"does not support context parallelism. "
-                f"Please set a compatible attention backend: {compatible_backends} using `model.set_attention_backend()`"
+                f"Please set a compatible attention backend: {compatible_backends} using `model.set_attention_backend()`."
             )
 
         _check_attention_backend_requirements(backend)


src/diffusers/models/attention_dispatch.py
```diff
@@ -348,6 +352,18 @@ def dispatch_attention_fn(
             check(**kwargs)
 
     kwargs = {k: v for k, v in kwargs.items() if k in _AttentionBackendRegistry._supported_arg_names[backend_name]}
+
+    if "_parallel_config" in kwargs and kwargs["_parallel_config"] is not None:
```
DN6: I don't think we need to run this check every time dispatch is called. I think we can iterate over the modules just once in the `set_attention_backend` method and raise an error if the processor is attempting to use an incompatible backend

sayakpaul: Addressed.

src/diffusers/models/modeling_utils.py
```diff
@@ -617,6 +618,15 @@ def set_attention_backend(self, backend: str) -> None:
             raise ValueError(f"`{backend=}` must be one of the following: " + ", ".join(available_backends))
 
         backend = AttentionBackendName(backend)
+        if not _AttentionBackendRegistry._is_context_parallel_available(backend):
```
DN6: This will raise an error if the backend doesn't support context parallel (regardless of whether you want to use CP or not) 

We need to first check if the attention processor has a `_parallel_config` set and then check if the backend supports context parallel.

sayakpaul: That should be fixed now. Thanks for flagging.

sayakpaul:
@DN6 a gentle ping.

src/diffusers/models/modeling_utils.py
```diff
@@ -607,17 +608,35 @@ def set_attention_backend(self, backend: str) -> None:
...
             raise ValueError(f"`{backend=}` must be one of the following: " + ", ".join(available_backends))
 
         backend = AttentionBackendName(backend)
+        if parallel_config_set and not _AttentionBackendRegistry._is_context_parallel_available(backend):
+            compatible_backends = sorted(_AttentionBackendRegistry._supports_context_parallel)
+            raise ValueError(
+                f"Context parallelism is enabled but backend '{backend.value}' "
+                f"which does not support context parallelism. "
+                f"Please set a compatible attention backend: {compatible_backends} using `model.set_attention_backend()` before "
+                f"calling `model.enable_parallelism()`."
```
DN6: Since this change allows `set_attention_backend` and `enable_paralleism` to be invoked in any order, I think we can  adjust the error message slightly. 

```suggestion
                f"Context parallelism is enabled but current attention backend '{backend.value}' "
                f"does not support context parallelism. "
                f"Please set a compatible attention backend: {compatible_backends} using `model.set_attention_backend()`"
```

DN6 (approved):
Small nit for error message.

PR #12839, opened 2025-12-15
[qwen-image] edit 2511 support

naykun:
🎄 Under the hush of a winter night, this PR bestows Qwen Image Edit with enigmatic new powers.

cc @sayakpaul @yiyixuxu

### DIFF
--- a/src/diffusers/models/transformers/transformer_qwenimage.py
+++ b/src/diffusers/models/transformers/transformer_qwenimage.py
@@ -364,7 +364,13 @@ class QwenDoubleStreamAttnProcessor2_0:
 @maybe_allow_in_graph
 class QwenImageTransformerBlock(nn.Module):
     def __init__(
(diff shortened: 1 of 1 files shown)


src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -618,6 +659,13 @@ def forward(
         hidden_states = self.img_in(hidden_states)
 
         timestep = timestep.to(hidden_states.dtype)
+
+        if self.zero_cond_t:
```
sayakpaul, line 669: Totally optional. But it would be great to leave a comment, briefly explaining what this is doing.

sayakpaul (approved):
Looking forward to seeing this!

Should we also update [docs](https://huggingface.co/docs/diffusers/main/en/api/pipelines/qwenimage) with the new checkpoint?

sayakpaul:
@bot /style

PR #12841, opened 2025-12-15
fix the warning torch_dtype is deprecated

### LINKED ISSUE #12840: Fix the warning `torch_dtype` is deprecated
## Description

When loading `QwenImageEditPipeline` (or other pipelines with Transformers models) using `torch_dtype` parameter, a deprecation warning is displayed:

```python
pipeline = QwenImageEditPipeline.from_pretrained(
    "Qwen/Qwen-Image-Edit",
    torch_dtype=torch.bfloat16,
)
```
Warning message:
```
`torch_dtype` is deprecated! Use `dtype` instead!
```
However, if users try to use `dtype` instead of `torch_dtype` as suggested by the warning, it causes an error because the diffusers' `from_pretrained` method doesn't accept `dtype` as a parameter.

## Root Cause
The deprecation warning originates from the Transformers library, which has deprecated `torch_dtype` in favor of `dtype`. When diffusers loads Transformers models (like text encoders in QwenImageEditPipeline), it passes `torch_dtype` to the Transformers' `from_pretrained` method, triggering this warning.

## Proposed Solution
Internally convert `torch_dtype` to `dtype` when loading Transformers models within diffusers, so users can continue using the existing API without warnings.

msdsm:
## Motivation

When users load pipelines like `QwenImageEditPipeline` with the `torch_dtype` parameter, they see a deprecation warning from the Transformers library. However, changing to `dtype` (as the warning suggests) causes errors because diffusers' `from_pretrained` doesn't accept `dtype` as a parameter.

This creates a confusing user experience where:
1. Users see a deprecation warning
2. Following the warning's advice breaks their code
3. There's no way to avoid the warning without modifying the diffusers library

## Solution

This PR internally converts `torch_dtype` to `dtype` when loading Transformers models (transformers >= 4.20.0), while keeping `torch_dtype` for diffusers models and other components.

Fixes #12840 

## Who can review?

### DIFF
--- a/src/diffusers/pipelines/pipeline_loading_utils.py
+++ b/src/diffusers/pipelines/pipeline_loading_utils.py
@@ -801,7 +801,7 @@ def load_sub_model(
     # add kwargs to loading method
     diffusers_module = importlib.import_module(__name__.split(".")[0])
     loading_kwargs = {}
-    
+
     is_diffusers_model = issubclass(class_obj, diffusers_module.ModelMixin)
 
     if is_transformers_available():
@@ -814,7 +814,7 @@ def load_sub_model(
         and issubclass(class_obj, PreTrainedModel)
         and transformers_version >= version.parse("4.20.0")
     )
-    
+
     # For transformers models >= 4.56.0, use 'dtype' instead of 'torch_dtype' to avoid deprecation warnings
     if issubclass(class_obj, torch.nn.Module):
         if is_transformers_model and transformers_version >= version.parse("4.56.0"):


src/diffusers/pipelines/pipeline_loading_utils.py
```diff
@@ -820,6 +814,17 @@ def load_sub_model(
         and issubclass(class_obj, PreTrainedModel)
         and transformers_version >= version.parse("4.20.0")
     )
+    
+    # For transformers models, use 'dtype' instead of 'torch_dtype' to avoid deprecation warnings
+    if issubclass(class_obj, torch.nn.Module):
+        if is_transformers_model:
```
DN6: Think we would also need to check if transformers version is greater than `4.56.0`.

msdsm:
@DN6 Thanks for the review! I've updated the code to check for transformers version >= 4.56.0 before using the `dtype` parameter.

DN6:
@bot /style

msdsm:
@DN6 Linting errors have been fixed. CI is waiting for workflow approval.

PR #12846, opened 2025-12-16
Add AutoPipeline For Text2Video

naomili0924:
# What does this PR do?

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/tests/pipelines/test_pipelines_auto.py
+++ b/tests/pipelines/test_pipelines_auto.py
@@ -27,6 +27,7 @@ from diffusers import (
     AutoPipelineForImage2Image,
     AutoPipelineForInpainting,
     AutoPipelineForText2Image,
+    AutoPipelineForText2Video,
     ControlNetModel,
     DiffusionPipeline,
 )
@@ -454,6 +455,15 @@ class AutoPipelineFastTest(unittest.TestCase):
 
         pipe = AutoPipelineForText2Image.from_pipe(pipe, image_encoder=None)
         assert pipe.image_encoder is None
+
+    def test_from_pretrained_text_to_video(self):
+        repo = "hf-internal-testing/tiny-stable-diffusion-pipe"
+
+        pipe = AutoPipelineForText2Video.from_pretrained(repo)
+        assert pipe.__class__.__name__ == "TextToVideoSDPipeline"
+
+        pipe = AutoPipelineForText2Video.from_pipe(pipe)
+        assert pipe.__class__.__name__ == "TextToVideoSDPipeline"
 
 
 @slow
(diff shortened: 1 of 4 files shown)


tests/pipelines/test_pipelines_auto.py
```diff
@@ -455,6 +456,15 @@ def test_from_pipe_optional_components(self):
...
         assert pipe.image_encoder is None
 
+    def test_from_pretrained_text_to_video(self):
+        repo = "hf-internal-testing/tiny-stable-diffusion-pipe"
+
+        pipe = AutoPipelineForText2Video.from_pretrained(repo)
+        assert pipe.__class__.__name__ == "TextToVideoSDPipeline"
+
+        pipe = AutoPipelineForText2Video.from_pipe(pipe)
+        assert pipe.__class__.__name__ == "TextToVideoSDPipeline"
```
IlyasMoutawwakil, line 466: does this test pass with "hf-internal-testing/tiny-stable-diffusion-pipe" ?

sayakpaul:
@yiyixuxu could you give this a look?

IlyasMoutawwakil:
@yiyixuxu the idea is to have a single entrypoint for the text2video models / task

yiyixuxu:
hi @IlyasMoutawwakil

The purpose of AutoPipeline is to enable users to switch between tasks using the same checkpoint. For video models, though, different tasks typically require different checkpoints (see [Wan 2.1](https://huggingface.co/collections/Wan-AI/wan21-702312c64fbb76a39ce5ba49) for example - their checkpoints all have the task encoded in their names: "T2V" for text-to-video, "I2V" for image-to-video, plus other tasks like audio-to-video, first-last-frame-to-video, etc.). We're also seeing some model makers (e.g. LTX) release checkpoints that handle multiple tasks, which we support with a single pipeline (LTXConditionPipeline).

The only exception where checkpoint sharing happens is text-to-video vs video-to-video, but IMO it's not worth adding and maintaining two auto pipelines for such a limited use case.

I'd like to understand the motivation a bit more - is there a specific project you're working on that requires this? If the goal is just to have a single entry point to load different video pipelines without knowing the class name, DiffusionPipeline.from_pretrained() should already handle that.

IlyasMoutawwakil:
Thanks @yiyixuxu i think i misunderstood the rationale behind AutoPipelines. The reason why we prefer having a task-specific AutoPipeline is because it makes task-specific testing in optimum-onnx and optimum-intel easy.
DiffusionPipeline does not guarantee which pipeline you will get when you load a checkpoint, while AutoPipelineForTask will always return the pipeline for the task it's related to.
I wasn't aware that video checkpoints only support one task for now, thanks for the clarification, i think we can use DiffusionPipeline with video models in our testing.

PR #12852, opened 2025-12-17
Cosmos Predict2.5 Base: inference pipeline, scheduler & chkpt conversion

miguelmartin75:
# What does this PR do?

This PR adds Cosmos Predict2.5 Base. It has been tested using the 2B model checkpoint [official HF checkpoint is here](https://huggingface.co/nvidia/Cosmos-Predict2.5-2B). The converted checkpoints have yet to be uploaded to HF.

This change is largely based off the previous predict1/predict2 support done by @a-r-r-o-w 

Testing:
- Aside from the unit tests, the examples in this [README from cosmos predict2.5]() have been checked against, and the inference pipeline produces similar outputs. 
- These are included in the docstring for the pipeline

## Additions

Pipelines:
  - A base pipeline which handles all modes for Predict2.5 Base checkpoint: Text2World, Image2World, Video2World
    - This pipeline loads the Reason1 checkpoint via `Qwen2_5_VLForConditionalGeneration`
    - In fact this pipeline supports image output if providing `num_frames=1`
    - Limitations: the pipeline assumes `batch_size == 1`
  - Three derivative pipelines are made based on this pipeline, to handle the official modes: Text2World, Image2World, Video2World
  - Unit tests are added, based on tests present for cosmos predict2

Scheduler:
  - ~~`FlowUniPCMultistepScheduler`: is a new scheduler introduced, which uses the EDM noise schedule (Karras sigmas) using the UniPC algorithm as predict2.5 uses flow matching. This name can be changed.~~
  - The above is integrated into the existing `UniPCMultistepScheduler` scheduler via supporting `use_karras_sigmas=True` and `use_flow_sigmas=True`
  - This is done to match the [predict2.5 codebase](https://github.com/nvidia-cosmos/cosmos-predict2.5)

Model changes:
- Modified `CosmosTransformer` to accept an optional cross-attention projection layer (used for text embeddings from Reason1)

Scripts:
- Extended `scripts/convert_cosmos_to_diffusers.py` to support Predict2.5

## Who can review?

- @yiyixuxu - pipeline and scheduler
- @sayakpaul - Documentation review
- @a-r-r-o-w

### DIFF
--- a/src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
+++ b/src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
@@ -661,9 +661,6 @@ class Cosmos2_5_PredictBasePipeline(DiffusionPipeline):
         else:
             batch_size = prompt_embeds.shape[0]
 
-        if batch_size != 1:
-            raise ValueError("batch_size must be 1")
-
         # Encode input prompt
         (
             prompt_embeds,
@@ -684,6 +681,9 @@ class Cosmos2_5_PredictBasePipeline(DiffusionPipeline):
 
         num_frames_in = None
         if image is not None:
+            if batch_size != 1:
+                raise ValueError(f"batch_size must be 1 for image input (given {batch_size})")
+
             image = torchvision.transforms.functional.to_tensor(image).unsqueeze(0)
             video = torch.cat([image, torch.zeros_like(image).repeat(num_frames - 1, 1, 1, 1)], dim=0)
             video = video.unsqueeze(0)
@@ -693,6 +693,9 @@ class Cosmos2_5_PredictBasePipeline(DiffusionPipeline):
             num_frames_in = 0
         else:
             num_frames_in = len(video)
+
+            if batch_size != 1:
+                raise ValueError(f"batch_size must be 1 for video input (given {batch_size})")
 
         assert video is not None
         video = self.video_processor.preprocess_video(video, height, width)


tests/pipelines/cosmos/test_cosmos2_5_predict.py
```diff
@@ -0,0 +1,301 @@
...
+            dim_mult=[1, 1, 1, 1],
+            num_res_blocks=1,
+            temperal_downsample=[False, True, True],
+        )
+
+        torch.manual_seed(0)
+        scheduler = FlowUniPCMultistepScheduler()
+
+        # NOTE: using Qwen2 VL instead for tests (reason1 is based on 2.5)
+        text_encoder = Qwen2VLForConditionalGeneration.from_pretrained(
```
miguelmartin75: Is there an internal Qwen2_5_VL model to test with?

src/diffusers/models/transformers/transformer_cosmos.py
```diff
@@ -485,6 +488,13 @@ def __init__(
             hidden_size, patch_size[0] * patch_size[1] * patch_size[2] * out_channels, bias=False
         )
 
+        self.use_crossattn_projection = use_crossattn_projection
```
sayakpaul: We don't have to assign the variables like this because `register_to_config` will have them included in the `model.config`:

https://github.com/huggingface/diffusers/blob/5e48f466b9c0d257f2650e8feec378a0022f2402/src/diffusers/models/transformers/transformer_cosmos.py#L426

```suggestion
```

src/diffusers/models/transformers/transformer_cosmos.py
```diff
@@ -546,6 +557,9 @@ def forward(
         else:
             assert False
 
+        if self.use_crossattn_projection:
```
sayakpaul:
```suggestion
        if self.config.use_crossattn_projection:
```

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,1295 @@
...
+        ... ).frames[0]
+        >>> export_to_video(video, "video2world.mp4", fps=16)
+
+        >>> # To produce a single-frame image instead of a world (video) clip, set num_frames=1 and
+        >>> # save the first frame: pipe(..., num_frames=1).frames[0][0].
+        ```
+"""
+
+
+class Cosmos2_5_PredictBase(DiffusionPipeline):
```
sayakpaul:
```suggestion
class Cosmos2_5_PredictBasePipeline(DiffusionPipeline):
```

And elsewhere

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,1295 @@
...
+            )
+
+            # duplicate text embeddings for each generation per prompt, using mps friendly method
+            _, seq_len, _ = negative_prompt_embeds.shape
+            negative_prompt_embeds = negative_prompt_embeds.repeat(1, num_videos_per_prompt, 1)
+            negative_prompt_embeds = negative_prompt_embeds.view(batch_size * num_videos_per_prompt, seq_len, -1)
+
+        return prompt_embeds, negative_prompt_embeds
+
+    def prepare_latents(
```
sayakpaul, line 390: Just checking if this is similar to any of the other existing Cosmos pipelines? If so, maybe this can be supplemented with a `"# Copied from ..." comment?

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,1295 @@
...
+                if i == len(timesteps) - 1 or ((i + 1) > num_warmup_steps and (i + 1) % self.scheduler.order == 0):
+                    progress_bar.update()
+
+                if XLA_AVAILABLE:
+                    xm.mark_step()
+
+        self._current_timestep = None
+
+        if not output_type == "latent":
+            assert self.latents_mean is not None and self.latents_std is not None, (
```
sayakpaul: We avoid `assert` statements like this in the modules. If we really want to raise, let's raise `Error`s instead.

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,1295 @@
...
+                "VAE configuration must define `latents_mean` and `latents_std`."
+            )
+            latents_mean = self.latents_mean.to(latents.device, latents.dtype)
+            latents_std = self.latents_std.to(latents.device, latents.dtype)
+            latents = latents * latents_std + latents_mean
+            video = self.vae.decode(latents.to(self.vae.dtype), return_dict=False)[0]
+
+            assert self.safety_checker is not None
+            self.safety_checker.to(device)
```
sayakpaul, line 812: We don't have to do it in this PR, but we could have a little utility like `run_safety_checker()` inside the pipelines and copy it over all the cosmos pipelines that require it (much akin to `encode_prompt()`, for example).

But this is not merge-blocking.

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,1295 @@
...
+        # Offload all models
+        self.maybe_free_model_hooks()
+
+        if not return_dict:
+            return (video,)
+
+        return CosmosPipelineOutput(frames=video)
+
+
+class Cosmos2_5_PredictText2World(Cosmos2_5_PredictBase):
```
sayakpaul: This feels quite odd. We usually maintain one pipeline class per pipeline file. That's been followed in the existing cosmos implementations as well:
https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/cosmos

We don't inherit from pipelines like `Cosmos2_5_PredictText2World(Cosmos2_5_PredictBase):`, either. We, instead, use`"# Copied from ..."` (we will unbloat quite a bit of that in https://github.com/huggingface/diffusers/pull/12820 but that's a separate thing).

tests/pipelines/cosmos/test_cosmos2_5_predict.py
sayakpaul: Would something like this work?
https://github.com/huggingface/diffusers/blob/5e48f466b9c0d257f2650e8feec378a0022f2402/tests/pipelines/qwenimage/test_qwenimage.py#L116

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
DN6: Could we follow the same structure as Cosmos2 with separate files for each pipeline. 
https://github.com/huggingface/diffusers/tree/main/src/diffusers/pipelines/cosmos

src/diffusers/schedulers/scheduling_flow_unipc_multistep.py
```diff
@@ -0,0 +1,723 @@
...
+    else:
+        raise ValueError(f"`final_sigmas_type` must be 'zero' or 'sigma_min' but got {final_sigmas_type}")
+
+    timesteps = torch.from_numpy(sigmas * num_train_steps).to(torch.int64)
+    sigmas = np.concatenate([sigmas, [sigma_last]]).astype(np.float32)
+    sigmas = torch.from_numpy(sigmas).to(dtype=torch.float32)
+    return sigmas, timesteps
+
+
+class FlowUniPCMultistepScheduler(SchedulerMixin, ConfigMixin):
```
yiyixuxu: the existing unipc multi also support flow matching, https://github.com/huggingface/diffusers/blob/main/src/diffusers/schedulers/scheduling_unipc_multistep.py#L212
do you think we will be able to reuse it?

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,1295 @@
...
+            output_type=output_type,
+            return_dict=return_dict,
+            callback_on_step_end=callback_on_step_end,
+            callback_on_step_end_tensor_inputs=callback_on_step_end_tensor_inputs,
+            max_sequence_length=max_sequence_length,
+            conditional_frame_timestep=conditional_frame_timestep,
+        )
+
+
+class Cosmos2_5_PredictImage2World(Cosmos2_5_PredictBase):
```
yiyixuxu: all our pipelines should inherit from `DIffusionPipeline`
so In the future, we could consider supporting cosmos from Modular Diffusers https://huggingface.co/docs/diffusers/main/en/modular_diffusers/overview

for this PR, to get it in ASAP, I think the best option is to keep single `Cosmos2_5_PredictPipeline` that handles all 3 use cases

miguelmartin75: I will remove these three specialized pipelines, the base pipeline is more flexible and relevant default arguments for each mode in 2World

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
miguelmartin75: I will remove these pipelines

src/diffusers/schedulers/scheduling_flow_unipc_multistep.py
miguelmartin75: This class doesn't output the sigma values needed when setting `use_flow_sigmas=True`, basically I want is `use_karras_sigmas=True` and `use_flow_sigmas=True` to work at the same time (with arguments for sigma_max and sigma_min). 

After reading the code in unipc multi, I have modified the class to support the above. I've updated the PR with this change (it's ~10 LOC) - see [the change here](https://github.com/huggingface/diffusers/pull/12852/changes/94d2b41eac91ada7b78cdac9fa8744a4bfb4ca96). Should I remove the flowunipc scheduler and keep this small change with an explicit test case for this setting?

Here is the outputs from a script for 35 steps (with the change to the branch). Without the change, setting both to True is equivalent to use_karras_sigmas=True, as this branch is hit first in the if/elif block
```
FlowUniPCMultistepScheduler:
Timesteps (len=35)
tensor([995, 994, 993, 991, 990, 988, 986, 983, 980, 976, 971, 965, 957, 948,
        936, 920, 901, 877, 847, 809, 763, 708, 642, 568, 486, 402, 319, 243,
        176, 122,  81,  51,  31,  18,   9])
Sigmas: (len=36)
tensor([0.9950, 0.9942, 0.9932, 0.9920, 0.9905, 0.9887, 0.9865, 0.9839, 0.9806,
        0.9766, 0.9717, 0.9655, 0.9578, 0.9482, 0.9360, 0.9208, 0.9016, 0.8775,
        0.8473, 0.8098, 0.7637, 0.7081, 0.6427, 0.5683, 0.4870, 0.4025, 0.3195,
        0.2430, 0.1768, 0.1229, 0.0817, 0.0519, 0.0315, 0.0181, 0.0099, 0.0000])
UniPCMultistepScheduler(**{'use_flow_sigmas': True}):
Timesteps (len=35)
tensor([999, 970, 941, 913, 884, 856, 827, 799, 770, 742, 713, 685, 656, 627,
        599, 570, 542, 513, 485, 456, 428, 399, 371, 342, 313, 285, 256, 228,
        199, 171, 142, 114,  85,  57,  28])
Sigmas: (len=36)
tensor([0.9990, 0.9705, 0.9419, 0.9134, 0.8848, 0.8563, 0.8277, 0.7992, 0.7707,
        0.7421, 0.7136, 0.6850, 0.6565, 0.6279, 0.5994, 0.5709, 0.5423, 0.5138,
        0.4852, 0.4567, 0.4281, 0.3996, 0.3711, 0.3425, 0.3140, 0.2854, 0.2569,
        0.2283, 0.1998, 0.1713, 0.1427, 0.1142, 0.0856, 0.0571, 0.0285, 0.0000])
torch.allclose?:  False

UniPCMultistepScheduler(**{'use_karras_sigmas': True}):
Timesteps (len=35)
tensor([999, 999, 991, 975, 957, 939, 920, 900, 880, 858, 835, 810, 784, 757,
        727, 695, 661, 623, 583, 538, 488, 434, 375, 312, 248, 188, 136,  94,
         62,  39,  23,  12,   6,   2,   0])
Sigmas: (len=36)
tensor([2.0000e+02, 1.7084e+02, 1.4539e+02, 1.2327e+02, 1.0410e+02, 8.7544e+01,
        7.3297e+01, 6.1086e+01, 5.0662e+01, 4.1801e+01, 3.4303e+01, 2.7989e+01,
        2.2698e+01, 1.8289e+01, 1.4636e+01, 1.1626e+01, 9.1637e+00, 7.1625e+00,
        5.5482e+00, 4.2564e+00, 3.2315e+00, 2.4258e+00, 1.7988e+00, 1.3162e+00,
        9.4913e-01, 6.7355e-01, 4.6961e-01, 3.2106e-01, 2.1476e-01, 1.4017e-01,
        8.8994e-02, 5.4745e-02, 3.2478e-02, 1.8473e-02, 1.0000e-02, 0.0000e+00])
torch.allclose?:  False

UniPCMultistepScheduler(**{'use_karras_sigmas': True, 'use_flow_sigmas': True}):
Timesteps (len=35)
tensor([995, 994, 993, 991, 990, 988, 986, 983, 980, 976, 971, 965, 957, 948,
        936, 920, 901, 877, 847, 809, 763, 708, 642, 568, 486, 402, 319, 243,
        176, 122,  81,  51,  31,  18,   9])
Sigmas: (len=36)
tensor([0.9950, 0.9942, 0.9932, 0.9920, 0.9905, 0.9887, 0.9865, 0.9839, 0.9806,
        0.9766, 0.9717, 0.9655, 0.9578, 0.9482, 0.9360, 0.9208, 0.9016, 0.8775,
        0.8473, 0.8098, 0.7637, 0.7081, 0.6427, 0.5683, 0.4870, 0.4025, 0.3195,
        0.2430, 0.1768, 0.1229, 0.0817, 0.0519, 0.0315, 0.0181, 0.0099, 0.0000])
torch.allclose?:  True
```

Note I have a bug in `FlowUniPCMultistepScheduler` as it outputs len=37 for sigmas when I set num_inference_steps=35, so to compare, I set `set_timesteps(num_inference_steps -1)` to compare

---

For additional context:

I created a new scheduler this out as the UniPC class was forked in the predic2.5 codebase (which I refactored). I did look into trying to re-use unipc, but due to the above settings and the complexity of the class, I decided to just clean-up this fork. In my opinion, the UniPC class is quite complex due to:
1. It is initialized with VP-type noise schedule, but this schedule is not enforced for inference: https://github.com/huggingface/diffusers/blob/main/src/diffusers/schedulers/scheduling_unipc_multistep.py#L251-L255
    - e.g. `use_flow_sigmas=True` or `use_karras_sigmas=True` this constraint is not enforced, i.e. `use_flow_sigmas=True` is only applied for `set_timesteps` which is for inference
2. `betas` and other parameters are not needed
3. KDM noise schedule, which can be achieved via `use_karras_sigmas=True` - but this is in conflict with `use_flow_sigmas=True`
4. Many parameters

In my opinion, a simpler design for this class could be composition to get sigmas that UniPC can operate on. The class does use some composition already, e.g. the argument `solver_p` [overrides the predictor step](https://github.com/huggingface/diffusers/blob/main/src/diffusers/schedulers/scheduling_unipc_multistep.py#L781-L783), but it doesn't do so to get the sigmas/timesteps to perform unipc on. But I do understand the philosophy behind diffusers, copying the code is preferred.

miguelmartin75: We discussed and concluded to remove the scheduler I introduced, which will be my next update to this PR.

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
miguelmartin75: It's a combination of two sources, I will say Modified from ...

miguelmartin75:
Updated PR to address comments, docs should be here: https://moon-ci-docs.huggingface.co/docs/diffusers/pr_12852/en/api/pipelines/cosmos when they are updated. I can update the main example to the latest model once we have uploaded the converted checkpoint to huggingface.

sayakpaul:
@bot /style

sayakpaul:
Could you run `make fix-copies`?

src/diffusers/pipelines/cosmos/__init__.py
```diff
@@ -22,6 +22,12 @@
 
     _dummy_objects.update(get_objects_from_module(dummy_torch_and_transformers_objects))
 else:
+    _import_structure["pipeline_cosmos2_5_predict"] = [
+        "Cosmos2_5_PredictBasePipeline",
+        "Cosmos2_5_PredictImage2World",
+        "Cosmos2_5_PredictText2World",
+        "Cosmos2_5_PredictVideo2World",
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/cosmos/__init__.py
```diff
@@ -35,6 +41,12 @@
     except OptionalDependencyNotAvailable:
         from ...utils.dummy_torch_and_transformers_objects import *
     else:
+        from .pipeline_cosmos2_5_predict import (
+            Cosmos2_5_PredictBasePipeline,
+            Cosmos2_5_PredictImage2World,
+            Cosmos2_5_PredictText2World,
+            Cosmos2_5_PredictVideo2World,
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,845 @@
...
+            video = torch.cat([image, torch.zeros_like(image).repeat(num_frames - 1, 1, 1, 1)], dim=0)
+            video = video.unsqueeze(0)
+            num_frames_in = 1
+        elif video is None:
+            video = torch.zeros(batch_size, num_frames, 3, height, width, dtype=torch.uint8)
+            num_frames_in = 0
+        else:
+            num_frames_in = len(video)
+
+        assert video is not None
```
yiyixuxu, line 700:
```suggestion
```
we already made sure video is not None I think

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,845 @@
...
+            max_sequence_length=max_sequence_length,
+        )
+
+        vae_dtype = self.vae.dtype
+        transformer_dtype = self.transformer.dtype
+
+        num_frames_in = None
+        if image is not None:
+            # TODO: handle batch_size > 1
+            assert batch_size == 1, "batch_size must be 1 for image input"
```
yiyixuxu:
```suggestion
            if batch_size != 1:
                raise ValueError("batch_size must be 1 for image input")
```

what would be the use cases we support for batch_size > 1, if it is possible, we can throw error for unsupported cases in the same place

but it is non merge-blocking

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,845 @@
...
+        image: PipelineImageInput | None = None,
+        video: List[PipelineImageInput] | None = None,
+        prompt: Union[str, List[str]] | None = None,
+        negative_prompt: Optional[Union[str, List[str]]] = None,
+        height: int = 704,
+        width: int = 1280,
+        num_frames: int = 93,
+        num_inference_steps: int = 36,
+        guidance_scale: float = 7.0,
+        fps: int = 16,
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
```diff
@@ -0,0 +1,845 @@
...
+            num_inference_steps (`int`, defaults to `35`):
+                The number of denoising steps. More denoising steps usually lead to a higher quality image at the
+                expense of slower inference.
+            guidance_scale (`float`, defaults to `7.0`):
+                Guidance scale as defined in [Classifier-Free Diffusion
+                Guidance](https://huggingface.co/papers/2207.12598). `guidance_scale` is defined as `w` of equation 2.
+                of [Imagen Paper](https://huggingface.co/papers/2205.11487). Guidance scale is enabled by setting
+                `guidance_scale > 1`.
+            fps (`int`, defaults to `16`):
+                The frames per second of the generated video.
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_predict.py
miguelmartin75: yes, this could be removed. will do in a follow-up pr for additional model conversions.

PR #12853, opened 2025-12-17
Qwen Image Layered Support

naykun:
🚀 This PR introduces Qwen Image Layered—a groundbreaking vision model that dissects images into rich, structured layers (think foreground, background, objects, and more)!

By unlocking pixel-perfect, semantically aware decomposition, we’re not just enabling smarter image editing—we’re igniting a whole new playground for creators, developers, and AI artists. Imagine remixing scenes like multitrack audio, editing objects in isolation, or generating dynamic compositions with unprecedented control.

The future of image generation is layered, modular, and collaborative—and it starts right here. Let’s build it together! 🎨✨

cc @sayakpaul @yiyixuxu

### DIFF
--- a/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
+++ b/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
@@ -393,7 +393,6 @@ the image\n<|vision_start|><|image_pad|><|vision_end|><|im_end|>\n<|im_start|>as
             raise ValueError(f"`max_sequence_length` cannot be greater than 1024 but is {max_sequence_length}")
 
     @staticmethod
-    # Copied from diffusers.pipelines.qwenimage.pipeline_qwenimage.QwenImagePipeline._pack_latents
     def _pack_latents(latents, batch_size, num_channels_latents, height, width, layers):
         latents = latents.view(batch_size, layers, num_channels_latents, height // 2, 2, width // 2, 2)
         latents = latents.permute(0, 1, 3, 5, 2, 4, 6)
@@ -402,7 +401,6 @@ the image\n<|vision_start|><|image_pad|><|vision_end|><|im_end|>\n<|im_start|>as
         return latents
 
     @staticmethod
-    # Copied from diffusers.pipelines.qwenimage.pipeline_qwenimage.QwenImagePipeline._unpack_latents
     def _unpack_latents(latents, height, width, layers, vae_scale_factor):
         batch_size, num_patches, channels = latents.shape
 
--- a/src/diffusers/utils/dummy_torch_and_transformers_objects.py
+++ b/src/diffusers/utils/dummy_torch_and_transformers_objects.py
@@ -2297,6 +2297,21 @@ class QwenImageInpaintPipeline(metaclass=DummyObject):
         requires_backends(cls, ["torch", "transformers"])
 
 
+class QwenImageLayeredPipeline(metaclass=DummyObject):
+    _backends = ["torch", "transformers"]
+
+    def __init__(self, *args, **kwargs):
+        requires_backends(self, ["torch", "transformers"])
+
+    @classmethod
+    def from_config(cls, *args, **kwargs):
+        requires_backends(cls, ["torch", "transformers"])
+
+    @classmethod
+    def from_pretrained(cls, *args, **kwargs):
+        requires_backends(cls, ["torch", "transformers"])
+
+
 class QwenImagePipeline(metaclass=DummyObject):
     _backends = ["torch", "transformers"]
 


src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -143,17 +143,26 @@ def apply_rotary_emb_qwen(
...
 
         self.time_proj = Timesteps(num_channels=256, flip_sin_to_cos=True, downscale_freq_shift=0, scale=1000)
         self.timestep_embedder = TimestepEmbedding(in_channels=256, time_embed_dim=embedding_dim)
+        self.additional_t_cond = additional_t_cond
+        if additional_t_cond:
+            self.addition_t_embedding = nn.Embedding(2, embedding_dim)
+            self.addition_t_embedding.weight.data.zero_()
 
-    def forward(self, timestep, hidden_states):
+    def forward(self, timestep, hidden_states, addition_t_cond=None):
```
sayakpaul, line 155: Maybe we could keep the default consistent across different signature? 

```suggestion
    def forward(self, timestep, hidden_states, addition_t_cond=False):
```

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -143,17 +143,26 @@ def apply_rotary_emb_qwen(
...
-    def __init__(self, embedding_dim):
+    def __init__(self, embedding_dim, additional_t_cond=False):
         super().__init__()
 
         self.time_proj = Timesteps(num_channels=256, flip_sin_to_cos=True, downscale_freq_shift=0, scale=1000)
         self.timestep_embedder = TimestepEmbedding(in_channels=256, time_embed_dim=embedding_dim)
+        self.additional_t_cond = additional_t_cond
+        if additional_t_cond:
+            self.addition_t_embedding = nn.Embedding(2, embedding_dim)
+            self.addition_t_embedding.weight.data.zero_()
```
sayakpaul: Not a critical thing, but since this concerns init, maybe we could delegate this to a proper init method like `self.init()` or something that would handle all kinds of inits like this. But this is totally optional for this PR.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -143,17 +143,26 @@ def apply_rotary_emb_qwen(
...
 
-    def forward(self, timestep, hidden_states):
+    def forward(self, timestep, hidden_states, addition_t_cond=None):
         timesteps_proj = self.time_proj(timestep)
         timesteps_emb = self.timestep_embedder(timesteps_proj.to(dtype=hidden_states.dtype))  # (N, D)
 
         conditioning = timesteps_emb
+        if self.additional_t_cond:
+            assert addition_t_cond is not None, "When additional_t_cond is True, addition_t_cond must be provided."
```
sayakpaul: Could we raise `ValueError`, instead?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -259,6 +268,122 @@ def _compute_video_freqs(self, frame: int, height: int, width: int, idx: int = 0
...
+        )
+        self.neg_freqs = torch.cat(
+            [
+                self.rope_params(neg_index, self.axes_dim[0], self.theta),
+                self.rope_params(neg_index, self.axes_dim[1], self.theta),
+                self.rope_params(neg_index, self.axes_dim[2], self.theta),
+            ],
+            dim=1,
+        )
+        self.rope_cache = {}
```
sayakpaul: (nit): Seems like it's not used? Hence can be removed?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -259,6 +268,122 @@ def _compute_video_freqs(self, frame: int, height: int, width: int, idx: int = 0
...
+
+        # DO NOT USING REGISTER BUFFER HERE, IT WILL CAUSE COMPLEX NUMBERS LOSE ITS IMAGINARY PART
```
sayakpaul: (nit): I think we can remove this comment since `scale_rope` seems to be a `bool`?

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -578,14 +703,21 @@ def __init__(
...
     ):
         super().__init__()
         self.out_channels = out_channels or in_channels
         self.inner_dim = num_attention_heads * attention_head_dim
 
-        self.pos_embed = QwenEmbedRope(theta=10000, axes_dim=list(axes_dims_rope), scale_rope=True)
+        if not use_layer3d_rope:
+            self.pos_embed = QwenEmbedRope(theta=10000, axes_dim=list(axes_dims_rope), scale_rope=True)
+        else:
+            self.pos_embed = QwenEmbedLayer3DRope(theta=10000, axes_dim=list(axes_dims_rope), scale_rope=True)
```
sayakpaul, line 714: Better to keep the two RoPE layers separated into classes like this to avoid unncessary complexity.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -621,6 +753,7 @@ def forward(
         guidance: torch.Tensor = None,  # TODO: this should probably be removed
         attention_kwargs: Optional[Dict[str, Any]] = None,
         controlnet_block_samples=None,
+        additional_t_cond=None,
```
sayakpaul, line 754: Just a suggestion.

`additional_t_cond` is already passed to `QwenTimestepProjEmbeddings`. I think we could assign `self.additional_t_cond = additional_t_cond` inside `QwenTimestepProjEmbeddings`. This way, we probably won't have to propagate `additional_t_cond` during the forward() calls, no? Or am I missing something?

This will also probably simplify things a bit.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+        >>> import torch
+        >>> from PIL import Image
+        >>> from diffusers import QwenImageLayeredPipeline
+        >>> from diffusers.utils import load_image
+
+        >>> pipe = QwenImageLayeredPipeline.from_pretrained("Qwen/Qwen-Image-Layered", torch_dtype=torch.bfloat16)
+        >>> pipe.to("cuda")
+        >>> image = load_image(
+        ...     "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/yarn-art-pikachu.png"
+        ... ).convert("RGBA")
```
sayakpaul, line 55: Interesting. Do we have to convert to the RGBA space in order for this model to function properly? 

(This is purely out of my curiosity)

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+        return encoder_output.latent_dist.sample(generator)
+    elif hasattr(encoder_output, "latent_dist") and sample_mode == "argmax":
+        return encoder_output.latent_dist.mode()
+    elif hasattr(encoder_output, "latents"):
+        return encoder_output.latents
+    else:
+        raise AttributeError("Could not access latents of provided encoder_output")
+
+
+def calculate_dimensions(target_area, ratio):
```
sayakpaul, line 164: Can we use "Copied from ..." here as well?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+    # Copied from diffusers.pipelines.qwenimage.pipeline_qwenimage.QwenImagePipeline._extract_masked_hidden
+    def _extract_masked_hidden(self, hidden_states: torch.Tensor, mask: torch.Tensor):
+        bool_mask = mask.bool()
+        valid_lengths = bool_mask.sum(dim=1)
+        selected = hidden_states[bool_mask]
+        split_result = torch.split(selected, valid_lengths.tolist(), dim=0)
+
+        return split_result
+
+    def _get_qwen_prompt_embeds(
```
sayakpaul, line 251: Just checking if we could use any "Copied from ...` here?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+        )
+        encoder_attention_mask = torch.stack(
+            [torch.cat([u, u.new_zeros(max_seq_len - u.size(0))]) for u in attn_mask_list]
+        )
+
+        prompt_embeds = prompt_embeds.to(dtype=dtype, device=device)
+
+        return prompt_embeds, encoder_attention_mask
+
+    def encode_prompt(
```
sayakpaul, line 292: Same for this one, too. Could we comment with "Copied from ..."?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+
+        latents = latents.view(batch_size, layers + 1, height // 2, width // 2, channels // 4, 2, 2)
+        latents = latents.permute(0, 1, 4, 2, 5, 3, 6)
+
+        latents = latents.reshape(batch_size, layers + 1, channels // (2 * 2), height, width)
+        latents = latents.permute(0, 2, 1, 3, 4)  # (b, c, f, h, w)
+
+        return latents
+
+    def _encode_vae_image(self, image: torch.Tensor, generator: torch.Generator):
```
sayakpaul, line 421: Possible to use "Copied from ..." here (perhaps from QwenImageEdit)?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+        """
+        depr_message = f"Calling `enable_vae_slicing()` on a `{self.__class__.__name__}` is deprecated and this method will be removed in a future version. Please use `pipe.vae.enable_slicing()`."
+        deprecate(
+            "enable_vae_slicing",
+            "0.40.0",
+            depr_message,
+        )
+        self.vae.enable_slicing()
+
+    def disable_vae_slicing(self):
```
sayakpaul: We can safely remove these two methods since they are deprecated.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+        """
+        depr_message = f"Calling `enable_vae_tiling()` on a `{self.__class__.__name__}` is deprecated and this method will be removed in a future version. Please use `pipe.vae.enable_tiling()`."
+        deprecate(
+            "enable_vae_tiling",
+            "0.40.0",
+            depr_message,
+        )
+        self.vae.enable_tiling()
+
+    def disable_vae_tiling(self):
```
sayakpaul: Same for these two as well.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+                z = z.unsqueeze(2)  # b c f h w
+                image.append(self.vae.decode(z, return_dict=False)[0])
+
+            image = torch.cat(image, dim=2)  # b c f h w
+            image = image.permute(0, 2, 3, 4, 1)  # b f h w c
+            image = (image * 0.5 + 0.5).clamp(0, 1).cpu().float().numpy()
+            image = (image * 255).round().astype("uint8")
+            images = []
+            for layers in image:
+                images.append([Image.fromarray(layer) for layer in layers])
```
sayakpaul: Wondering if we can leverage `self.image_processor.postprocess` here? This is what we usually follow for the pipelines? If we need to have a different postprocess method, we could implement a separate ImageProcessor within the `qwenimage` module and use it. 

Example:
https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/hunyuan_video1_5/image_processor.py

sayakpaul (commented):
Thanks a lot for this PR! My comments should mostly be minor. 

LMK if anything is unclear.

Let's also update with a test and the docs?

src/diffusers/models/transformers/transformer_qwenimage.py
naykun: any change to apply here?

src/diffusers/models/transformers/transformer_qwenimage.py
naykun: The parameter name `additional_t_cond` can be confusing because it serves two different purposes: in the __init__ method, it acts as a boolean flag, while in the forward method, it’s expected to be a tensor (with a default value of None). To avoid ambiguity, I’ve renamed the initialization argument to `use_additional_t_cond`, which better reflects its role as a control flag.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
naykun: a bit different here

src/diffusers/models/transformers/transformer_qwenimage.py
sayakpaul: Not at all. I was just noting that the current separation is good.

src/diffusers/models/transformers/transformer_qwenimage.py
sayakpaul: Ah that makes sense!

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+            prompt_image = image
+            image = self.image_processor.preprocess(image, calculated_height, calculated_width)
+            image = image.unsqueeze(2)
+            image = image.to(dtype=self.text_encoder.dtype)
+
+        if prompt is None or prompt == "" or prompt == " ":
+            prompt = self.get_image_caption(prompt_image, use_en_prompt=use_en_prompt, device=device)
+            print(f"Generated prompt: {prompt}")
+        else:
+            print(f"User prompt: {prompt}")
```
sayakpaul: We cab safely remove them.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+            )
+            latents_std = 1.0 / torch.tensor(self.vae.config.latents_std).view(1, self.vae.config.z_dim, 1, 1, 1).to(
+                latents.device, latents.dtype
+            )
+            latents = latents / latents_std + latents_mean
+            latents = torch.unbind(latents, 2)
+            image = []
+            for z in latents[1:]:
+                z = z.unsqueeze(2)  # b c f h w
+                image.append(self.vae.decode(z, return_dict=False)[0])
```
sayakpaul: Just wondering if we can decode one-shot? Or is it done to prevent memory problems?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,940 @@
...
+
+    @torch.no_grad()
+    @replace_example_docstring(EXAMPLE_DOC_STRING)
+    def __call__(
+        self,
+        image: Optional[PipelineImageInput] = None,
+        prompt: Union[str, List[str]] = None,
+        negative_prompt: Union[str, List[str]] = None,
+        true_cfg_scale: float = 4.0,
+        height: Optional[int] = None,
```
yiyixuxu: height/width are not used here, no?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
naykun: Yes for now the input should be converted to 4 channel images

src/diffusers/models/transformers/transformer_qwenimage.py
yiyixuxu: it'a an actual model input, not a flag 
no change needed here

src/diffusers/models/transformers/transformer_qwenimage.py
yiyixuxu: it's already two classes so no changes needed

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
yiyixuxu: we don't need a separate image processor 
I think we can probably use existing one by looping over the layers and process one layer each time
if not, ok to keep the code here

yiyixuxu (approved):
thanks! 
looks really good to me & we can merge this soon!

src/diffusers/models/transformers/transformer_qwenimage.py
sayakpaul: Yes, it can be supplemented as a model input. This comment solves my concern:
https://github.com/huggingface/diffusers/pull/12853/files#r2626129408

Previously, I thought it can just be a flag like it's used in the init.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
naykun: Thanks the advice, I've got a better implementation in the new commit.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
naykun: yes we can

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
naykun: removed!

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -143,17 +143,26 @@ def apply_rotary_emb_qwen(
...
 class QwenTimestepProjEmbeddings(nn.Module):
-    def __init__(self, embedding_dim):
+    def __init__(self, embedding_dim, use_additional_t_cond=False):
         super().__init__()
 
         self.time_proj = Timesteps(num_channels=256, flip_sin_to_cos=True, downscale_freq_shift=0, scale=1000)
         self.timestep_embedder = TimestepEmbedding(in_channels=256, time_embed_dim=embedding_dim)
+        self.use_additional_t_cond = use_additional_t_cond
+        if use_additional_t_cond:
+            self.addition_t_embedding = nn.Embedding(2, embedding_dim)
```
sayakpaul, line 153: Much better.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
```diff
@@ -0,0 +1,885 @@
...
+            latents = latents.permute(0, 2, 1, 3, 4).view(-1, c, 1, h, w)
+
+            image = self.vae.decode(latents, return_dict=False)[0]  # (b f) c 1 h w
+
+            image = image.squeeze(2)
+
+            image = self.image_processor.postprocess(image, output_type=output_type)
+            images = []
+            for bidx in range(b):
+                images.append(image[bidx * f : (bidx + 1) * f])
```
sayakpaul, line 897: This is clean! Thanks for this :)

sayakpaul:
@bot /style

sayakpaul:
@naykun I opened https://github.com/naykun/diffusers/pull/1 to help fix the consistency problems. Could you please check and merge?

sayakpaul:
Failing tests are unrelated. Looking forward to seeing Qwen ImageLayer a grand success! Also, hopefully, this kind of starts a new paradigm in image generation!

gluttony-10:
@naykun Although image is an optional input in pipeline parameters. But without entering an image, it cannot run successfully. Looking forward to repairing.

Best regards！

PR #12858, opened 2025-12-18
[tests] QwenImage Layer tests

sayakpaul:
# What does this PR do?

Adds tests for the newly added QwenImage layer model.

First, this PR fixes how we return the outputs; otherwise, we run into:

```bash
FAILED tests/pipelines/qwenimage/test_qwenimage_layered.py::QwenImageLayeredPipelineFastTests::test_callback_inputs - UnboundLocalError: local variable 'images' referenced before assignment
```

Second, we need to decide how to tackle this situation (which is fairly common across all failures):

```bash
FAILED tests/pipelines/qwenimage/test_qwenimage_layered.py::QwenImageLayeredPipelineFastTests::test_attention_slicing_forward_pass - TypeError: unsupported operand type(s) for -: 'list' and 'list'
```

Unlike other pipelines, we're always returning a `list[Image.Image | np.ndarray | torch.Tensor]`. One solution could be to apply `np.stack()` or `torch.stack` when `output_type="np"` or `output_type="pt"`.

Cc: @yiyixuxu @naykun

### DIFF
--- a/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
+++ b/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_layered.py
@@ -882,27 +882,21 @@ the image\n<|vision_start|><|image_pad|><|vision_end|><|im_end|>\n<|im_start|>as
             latents = latents / latents_std + latents_mean
 
             b, c, f, h, w = latents.shape
-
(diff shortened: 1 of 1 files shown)


sayakpaul:
A gentle ping @yiyixuxu

Issue #12859, opened 2025-12-18
load_lora_weights(), unload_lora_weights these doesn't work with LongCatImagePipeline.


ovi054:
### Model/Pipeline/Scheduler description

load_lora_weights(), unload_lora_weights these doesn't work with LongCatImagePipeline. Please add lora support for this model in diffusers.

<img width="784" height="141" alt="Image" src="https://github.com/user-attachments/assets/ebf55ed2-26d9-4cb2-b771-050cc3993cbf" />

### Open source status

- [ ] The model implementation is available.
- [ ] The model weights are available (Only relevant if addition is not a scheduler).

surajyadav-research:
Hi! I’d love to work on this issue. I’ve gone through the description and can start right away, please let me know if it’s okay to proceed.

tdvipp:
can you provide example code?

PR #12867, opened 2025-12-21
Fix LongCat LoRA load/unload and add regression test

surajyadav-research:
## What does this PR do?

This PR fixes LoRA integration for `LongCatImagePipeline` so that `load_lora_weights()` properly applies the adapter during inference and `unload_lora_weights()` cleanly restores the base (non-LoRA) behavior.

It also adds a slow regression test that:
1) runs the pipeline without LoRA (baseline),
2) loads a LoRA and verifies the output changes,
3) unloads the LoRA and verifies the output returns close to the baseline.

## Why?

Addresses the reported LoRA load/unload issue for LongCat:
-  #12859

## Tests

- `RUN_SLOW=yes pytest -q tests/pipelines/longcat_image/test_longcat_lora.py`


surajyadav-research:
Hi @sayakpaul 
CI is currently “awaiting approval from a maintainer”. Could you please approve the workflow runs?

surajyadav-research:
Hi @sayakpaul,
I’ve fixed the issue in the test file and pushed the update. Would appreciate your review when convenient.

src/diffusers/pipelines/longcat_image/pipeline_longcat_image.py
```diff
@@ -202,7 +202,7 @@ def retrieve_timesteps(
     return timesteps, num_inference_steps
 
 
-class LongCatImagePipeline(DiffusionPipeline, FromSingleFileMixin):
+class LongCatImagePipeline(DiffusionPipeline, FluxLoraLoaderMixin, FromSingleFileMixin):
```
sayakpaul: This seems quite incorrect to me.

Flux has two LoRA loadable modules:
https://github.com/huggingface/diffusers/blob/f6b6a7181eb44f0120b29cd897c129275f366c2a/src/diffusers/loaders/lora_pipeline.py#L1493

For LongCat, it uses a different text encoder (Flux uses two text encoder, let along) and rest of its components also seems to be different from Flux:

https://github.com/huggingface/diffusers/blob/f6b6a7181eb44f0120b29cd897c129275f366c2a/src/diffusers/pipelines/longcat_image/pipeline_longcat_image.py#L214-L222

So, could you please explain how using the `FluxLoraLoaderMixin` is appropriate here?

Instead, I suggest we write a dedicated LoRA loader mixin class for LongCat -- `LongCatLoraLoaderMixin`. You can refer to 
https://github.com/huggingface/diffusers/blob/f6b6a7181eb44f0120b29cd897c129275f366c2a/src/diffusers/loaders/lora_pipeline.py#L4775

as an example.

tests/pipelines/longcat_image/test_longcat_lora.py
```diff
@@ -0,0 +1,107 @@
+# Copyright 2025 The HuggingFace Team.
```
sayakpaul: This is not needed. Please try to consult the existing testing structure for pipeline-level LoRA testing c.f. https://github.com/huggingface/diffusers/tree/main/tests/lora/

surajyadav-research:
Thanks for the clarification @sayakpaul 
I agree the current inheritance from `FluxLoraLoaderMixin` is misleading here.

While implementing this, my reasoning was that `FluxLoraLoaderMixin` mainly reuses the generic LoRA load/fuse path and simply assumes the loadable entry points are `["transformer", "text_encoder"]`. But since LongCat’s `text_encoder` is `Qwen2_5_VLForConditionalGeneration` and the transformer is `LongCatImageTransformer2DModel` (i.e., not Flux’s text-encoder setup), this can easily lead to subtle issues like incorrect key routing, mismatched target modules, or silent partial loads.

So I’ll switch to a dedicated `LongCatLoraLoaderMixin` to make the intent explicit and handle any LongCat-specific routing cleanly.

PR #12869, opened 2025-12-21
Split attention backend

dxqb:
# What does this PR do?

Most recent models have been using variable lengths captions (Qwen, Chroma, Z-Image, ...) and require attention masking if batch size > 1 with multiple captions.
torch SDPA only uses its internal flash attention algorithm if there is no attention mask. Otherwise it falls back to another algorithm that is significantly slower, especially for high sequence lengths.

This PR implements an attention backend that splits up the attention batch into individual samples. Even though attention has to be called multiple times then, it is still faster than masked attention (tested up to batch size 8).

This PR also lays the groundwork for efficiently using "flash varlen" and other varlen attention backends, which are already implemented but not efficiently (see code comment).

This PR is based on @kashif and @cdutr 's work in this PR: https://github.com/huggingface/diffusers/pull/12702
To see only my changes, relative to their PR, follow this link: https://github.com/dxqb/diffusers/pull/3/files

## Benchmarks

Training benchmarks using OneTrainer: especially training in higher resolution benefits:
grafik

Inference benchmark using diffusers Qwen example script (but with regional compilation):
Inference benefits when comparing apples to apples, which is BS2 for CFG. However, the current pipeline already avoids attention masks by calling the transformer twice with BS1, so there is only a slight practical improvement for inference:
grafik

## Who can review?

@yiyixuxu and @sayakpaul 
CC @kashif and @cdutr for feedback

### DIFF
--- a/src/diffusers/models/transformers/transformer_qwenimage.py
+++ b/src/diffusers/models/transformers/transformer_qwenimage.py
@@ -492,7 +492,7 @@ class QwenDoubleStreamAttnProcessor2_0:
         encoder_hidden_states_mask: torch.FloatTensor = None,
         attention_mask: Optional[torch.FloatTensor] = None,
         image_rotary_emb: Optional[torch.Tensor] = None,
-        encoder_hidden_states_len: Optional[torch.Tensor] = None,
+        encoder_hidden_states_len: Optional[List[int]] = None,
     ) -> torch.FloatTensor:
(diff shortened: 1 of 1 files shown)


dxqb:
renaming a branch apparently closes the PR. re-opened here: https://github.com/huggingface/diffusers/pull/12870

PR #12870, opened 2025-12-21
Split attention backends

dxqb:
# What does this PR do?

Most recent models have been using variable lengths captions (Qwen, Chroma, Z-Image, ...) and require attention masking if batch size > 1 with multiple captions.
torch SDPA only uses its internal flash attention algorithm if there is no attention mask. Otherwise it falls back to another algorithm that is significantly slower, especially for high sequence lengths.

This PR implements an attention backend that splits up the attention batch into individual samples. Even though attention has to be called multiple times then, it is still faster than masked attention (tested up to batch size 8).

This PR also lays the groundwork for efficiently using "flash varlen" and other varlen attention backends, which are already implemented but not efficiently (see code comment).

This PR is based on @kashif and @cdutr 's work in this PR: https://github.com/huggingface/diffusers/pull/12702

## Benchmarks

Training benchmarks using OneTrainer: especially training in higher resolution benefits:
grafik

Inference benchmark using diffusers Qwen example script (but with regional compilation):
Inference benefits when comparing apples to apples, which is BS2 for CFG. However, the current pipeline already avoids attention masks by calling the transformer twice with BS1, so there is only a slight practical improvement for inference:
grafik

## Who can review?

@yiyixuxu and @sayakpaul 
CC @kashif and @cdutr for feedback

contains https://github.com/huggingface/diffusers/pull/12892


dxqb:
- [X] added backends to split flash attention. This is useful especially for Windows users, because native torch SDPA doesn't use flash internally on Windows
- [ ] tried to add Z-Image support but ran into this issue: https://github.com/huggingface/diffusers/issues/12893

dxqb:
Yes. 

Next steps:
 - I'll merge with main when https://github.com/huggingface/diffusers/pull/12987 is merged
 - ask for review then

src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py
```diff
@@ -468,6 +468,7 @@ def __call__(
         callback_on_step_end: Callable[[int, int], None] | None = None,
         callback_on_step_end_tensor_inputs: list[str] = ["latents"],
         max_sequence_length: int = 512,
+        batch_negative: bool = False, #TODO remove, only for testing
```
dxqb, line 471: all changes in this file are only for testing - should be reverted before merge

dxqb:
> Yes.
> 
> Next steps:
> 
>     * I'll merge with main when [[Qwen] avoid creating attention masks when there is no padding #12987](https://github.com/huggingface/diffusers/pull/12987) is merged
> 
>     * ask for review then

done

src/diffusers/models/attention_dispatch.py
```diff
@@ -514,7 +518,7 @@ def _prepare_for_flash_attn_or_sage_varlen_without_mask(
     cu_seqlens_k = torch.zeros(batch_size + 1, dtype=torch.int32, device=device)
     cu_seqlens_q[1:] = torch.cumsum(seqlens_q, dim=0)
     cu_seqlens_k[1:] = torch.cumsum(seqlens_k, dim=0)
-    max_seqlen_q = seqlens_q.max().item()
+    max_seqlen_q = seqlens_q.max().item() #TODO item() is inefficient and breaks torch.compile graphs. Use 'seq_len' parameter instead (see split attention backend)
```
dxqb, line 586: I have benchmarked the new varlen attention in torch 2.10:
https://docs.pytorch.org/docs/stable/nn.attention.varlen.html

Performance isn't bad, but in my test case, split attention was always faster - if you considered the preparation of the varlen tokens. Even with torch.compiling the preparation.

dxqb:
Pipelines and pipeline callbacks: @yiyixuxu and @asomoza

could I please have some feedback on this PR, whether you want this attention backend in `diffusers`?
If so, I could add support for more models, now that https://github.com/huggingface/diffusers/issues/12893 is resolved.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -164,8 +164,12 @@ def compute_text_seq_len_from_mask(
...
     active_positions = torch.where(encoder_hidden_states_mask, position_ids, position_ids.new_zeros(()))
     has_active = encoder_hidden_states_mask.any(dim=1)
-    per_sample_len = torch.where(has_active, active_positions.max(dim=1).values + 1, torch.as_tensor(text_seq_len))
-    return text_seq_len, per_sample_len.tolist(), encoder_hidden_states_mask
+    per_sample_len = torch.where(
+        has_active,
+        active_positions.max(dim=1).values + 1,
+        torch.as_tensor(text_seq_len, device=encoder_hidden_states.device),
+    )
+    return text_seq_len, per_sample_len, encoder_hidden_states_mask
```
dxqb: thanks for the merge, but it think you missed one here. for this PR, per_sample_len has to be returned as list[int], not as tensor

kashif: opps my bad, can you add it back?

PR #12873, opened 2025-12-22
[core] remove unneeded autoencoder methods when subclassing from `AutoencoderMixin`

sayakpaul:
# What does this PR do?

Less is better.


yiyixuxu (approved):
thanks!

Issue #12605, opened 2025-11-07
dispatch_attention_fn silently ignores attn_mask for certain backends


zzlol63:
### Describe the bug

When using the [attention backends feature](https://huggingface.co/docs/diffusers/main/en/optimization/attention_backends), it's possible to pass an attention mask to the dispatch_attention_fn to a backend which does not support them (ie. attn_mask is not in the method signature registered to that particular backend), and have no error thrown.

This may cause unintended side-effects during training or inference where the attention will be computed as a full-attention scenario. 

This is not validated either at the as part of the optional checks used for debugging purposes (`DIFFUSERS_ATTN_CHECKS=yes`)

### Reproduction

```
import os
os.environ["DIFFUSERS_ATTN_CHECKS"] = "yes"

import torch
from diffusers.models.attention_dispatch import dispatch_attention_fn, attention_backend

query = torch.randn(1, 10, 8, 64, dtype=torch.bfloat16, device="cuda")
key = torch.randn(1, 10, 8, 64, dtype=torch.bfloat16, device="cuda")
value = torch.randn(1, 10, 8, 64, dtype=torch.bfloat16, device="cuda")
attn_mask = torch.zeros((1, 1, 10, 10), dtype=torch.bool, device="cuda")

with attention_backend("native"):
    output = dispatch_attention_fn(query, key, value, attn_mask)
    output2 = dispatch_attention_fn(query, key, value)

    assert not torch.equal(output, output2), "native: These outputs should not be equal!"

with attention_backend("flash"):
    output = dispatch_attention_fn(query, key, value, attn_mask)
    output2 = dispatch_attention_fn(query, key, value)

    assert not torch.equal(output, output2), "flash: These outputs should not be equal!"
```

```
Traceback (most recent call last):
  File "C:\repro.py", line 22, in <module>
    assert not torch.equal(output, output2), "flash: These outputs should not be equal!"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: flash: These outputs should not be equal!
```

### System Info

- 🤗 Diffusers version: 0.35.2
- Platform: Windows-11-10.0.26100-SP0
- Running on Google Colab?: No
- Python version: 3.12.12
- PyTorch version (GPU?): 2.8.0+cu128 (True)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Huggingface_hub version: 1.1.2
[...]

PR #12892, opened 2025-12-26
Check for attention mask in backends that don't support it

dxqb:
# What does this PR do?

Using an attention backend (https://huggingface.co/docs/diffusers/main/optimization/attention_backends) with a model that passes attention masks yields incorrect results.

This is already checked in parallel backends...
https://github.com/huggingface/diffusers/blob/f6b6a7181eb44f0120b29cd897c129275f366c2a/src/diffusers/models/attention_dispatch.py#L978
...but not yet in the regular ones.

This PR changes that.
Fixes https://github.com/huggingface/diffusers/issues/12605

## Who can review?

@yiyixuxu and @asomoza
CC @zzlol63 @tolgacangoz

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,22 @@
+---
+name: CodeQL Security Analysis For Github Actions
+
+on:
+  push:
+    branches: ["main"]
+  workflow_dispatch:
+  # pull_request:
+
+jobs:
+  codeql:
+    name: CodeQL Analysis
+    uses: huggingface/security-workflows/.github/workflows/codeql-reusable.yml@v1
+    permissions:
+      security-events: write
+      packages: read
+      actions: read
+      contents: read
+    with:
+      languages: '["actions","python"]'
+      queries: 'security-extended,security-and-quality'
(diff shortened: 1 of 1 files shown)


yiyixuxu (approved):
thanks!

yiyixuxu:
@bot /style

dxqb:
> Thanks!
> 
> But we don't even support passing `attn_mask` in the respective methods. So, I am not sure how this is applicable.

True, but that is the point: Those backends don't have an 'attn_mask' parameter currently. Therefore, when the caller passes an 'attn_mask' argument, the attention mask is **silently** ignored here: https://github.com/huggingface/diffusers/blob/05014a343836fbbef92f71e1feae18a3fe6b3810/src/diffusers/models/attention_dispatch.py#L343

if checks are disabled, which they are by default.

Ignoring extra arguments might be valid behaviour for other paremters, but not for attention masks. Ignoring an attention mask yields incorrect results. It should fail instead of returning incorrect results - that's what this PR does.

@yiyixuxu your style bot changed a line I didn't even mean to insert. Fixed.

PR #12911, opened 2026-01-04
Discrete diffusion in diffusers

kashif:
# What does this PR do?

Add experimental support for discrete token diffusion methods and pipeline

moved llada2 to its own PR: https://github.com/huggingface/diffusers/pull/13226

## Before submitting

## Who can review?


yiyixuxu:
Thanks for this PR! 
cc @dg845, can you take a look here? it's related to Dream 7B #12091 you are working on

dg845:
Thanks for the PR! Some preliminary design questions and comments:

1. I think it could be useful to have a natural place to implement logic which is common to discrete diffusion models. Would something like a `DiscreteDiffusionPipelineMixin` make sense? For example, I think `_resolve_start_token_id`, `_normalize_prefix_ids`, `_top_p_filtering`, etc. could be candidates as mixin methods. (A possible alternative could be to put the methods in [`DiffusionPipeline`](https://github.com/huggingface/diffusers/blob/d4f97d19210d4abc32dac78fe7080cd8f5f0809c/src/diffusers/pipelines/pipeline_utils.py#L183), but it feels a little weird to put the methods there because they aren't applicable to continuous diffusion models.) But maybe this is premature, since we might not know what logic will end up being useful for all (or most) discrete diffusion models.
    1. One motivation for this is that we often want to do semi-autoregressive (SAR) sampling for discrete diffusion models, so it would be useful to have autoregressive sampling techniques such as top-$`p`$ sampling, top-$`k`$ sampling, etc. So I think it would be nice to have a place where these methods can be implemented and tested once, and then new discrete diffusion models that support SAR sampling can have easy access to them without having to copy them every time.
2. Similarly, would it make sense to have a `TokenizerTextProcessor` class which handles text pre-processing and and post-processing, analogous to how [`VaeImageProcessor`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/image_processor.py) handles image pre- and post-processing? It's probably less necessary as we don't need to do as much normalization as for images, but I could see this being useful for handling e.g. chat templates like in the SDAR and LLaDA 2 pipelines.
    1. As an aside, this could also be useful for existing (continuous) diffusion models, some of which have pretty involved text processing, such as pipelines like `SanaPipeline` that use a `_text_preprocessing` method: https://github.com/huggingface/diffusers/blob/d4f97d19210d4abc32dac78fe7080cd8f5f0809c/src/diffusers/pipelines/sana/pipeline_sana.py#L548-L549
3. Currently it looks like the pipelines only support denoising models with a `transformers`-like interface. But we would probably want to implement some discrete diffusion transformers in `diffusers`, which currently doesn't enforce that interface. So I think we should think about how we can handle both cases gracefully in discrete diffusion pipelines. (One solution could be to simply adopt the `transformers` interface for all discrete denoising models in `diffusers`, but that could be unnecessarily restrictive.)

src/diffusers/pipelines/token_diffusion/pipeline_token_diffusion.py
```diff
@@ -0,0 +1,276 @@
...
+        num_inference_steps: int = 128,
+        inject_start_token: bool = False,
+    ):
+        super().__init__()
+        self.register_modules(model=model, scheduler=scheduler, tokenizer=tokenizer)
+        self.register_to_config(
+            seq_len=seq_len,
+            num_inference_steps=num_inference_steps,
+            inject_start_token=inject_start_token,
+        )
```
dg845: Generally we don't register default `__call__` arguments to the `config`, but rather set them as default arguments to the `__call__` method:

https://github.com/huggingface/diffusers/blob/d4f97d19210d4abc32dac78fe7080cd8f5f0809c/src/diffusers/pipelines/ltx2/pipeline_ltx2.py#L744-L752

src/diffusers/pipelines/token_diffusion/pipeline_token_diffusion.py
```diff
@@ -0,0 +1,276 @@
...
+        prefix_ids = prefix_ids.to(device=device)
+        if prefix_ids.shape[0] == 1 and batch_size > 1:
+            prefix_ids = prefix_ids.expand(batch_size, -1)
+        return prefix_ids
+
+    @torch.no_grad()
+    def __call__(
+        self,
+        *,
+        batch_size: int = 1,
```
dg845: `diffusers` pipelines usually don't set `__call__` arguments to be keyword-only. (That's not to say that there are no arguments for it, but because other pipelines allow positional arguments I think the expectation is that discrete diffusion pipelines will allow them as well.)

src/diffusers/pipelines/token_diffusion/pipeline_token_diffusion.py
```diff
@@ -0,0 +1,276 @@
...
+                `callback_on_step_end(self, step: int, timestep: int, callback_kwargs: Dict)`.
+            callback_on_step_end_tensor_inputs: List of tensor keys to include in `callback_kwargs`.
+            model_kwargs: Forward kwargs passed to `model(...)` (e.g. attention mask overrides).
+        """
+        if seq_len is None:
+            seq_len = int(self.config.seq_len)
+        if num_inference_steps is None:
+            num_inference_steps = int(self.config.num_inference_steps)
+        if inject_start_token is None:
+            inject_start_token = bool(self.config.inject_start_token)
```
dg845: Following up on https://github.com/huggingface/diffusers/pull/12911#discussion_r2719233657, this logic could be removed if we don't register default arguments to the `config`.

src/diffusers/pipelines/token_diffusion/pipeline_token_diffusion.py
```diff
@@ -0,0 +1,276 @@
...
+        input_ids = self._init_latents(batch_size, seq_len, generator=generator, device=device)
+        attention_mask = torch.ones_like(input_ids, dtype=torch.long)
+
+        fixed_mask = None
+        fixed_values = None
+        if infill_mask is not None:
+            if infill_mask.shape != (batch_size, seq_len):
+                raise ValueError(
+                    f"`infill_mask` must have shape {(batch_size, seq_len)}, got {tuple(infill_mask.shape)}."
+                )
```
dg845: I think input checking and exceptions should be moved to a `check_inputs` method, which is the usual practice for `diffusers` pipelines:

https://github.com/huggingface/diffusers/blob/d4f97d19210d4abc32dac78fe7080cd8f5f0809c/src/diffusers/pipelines/flux2/pipeline_flux2.py#L686-L693

src/diffusers/pipelines/token_diffusion/pipeline_token_diffusion.py
```diff
@@ -0,0 +1,276 @@
...
+        tok = getattr(self, "tokenizer", None)
+        if tok is None:
+            return None
+        for attr in ("bos_token_id", "cls_token_id"):
+            token_id = getattr(tok, attr, None)
+            if token_id is not None:
+                return int(token_id)
+        return None
+
+    def _init_latents(
```
dg845: We usually name methods which sample `latents` from the prior distribution `prepare_latents`:

https://github.com/huggingface/diffusers/blob/d4f97d19210d4abc32dac78fe7080cd8f5f0809c/src/diffusers/pipelines/flux2/pipeline_flux2.py#L619

src/diffusers/pipelines/token_diffusion/pipeline_token_diffusion.py
```diff
@@ -0,0 +1,276 @@
...
+                    device=device,
+                    dtype=torch.long,
+                    generator=generator,
+                )
+            vocab_size = int(getattr(self.scheduler, "vocab_size", 0))
+            if vocab_size <= 0:
+                raise ValueError("Scheduler must define `vocab_size` for uniform prior sampling.")
+            return torch.randint(
+                0, vocab_size, (batch_size, seq_len), device=device, dtype=torch.long, generator=generator
+            )
```
dg845: Suggestion: maybe it would be cleaner to define a scheduler method called (say) `sample_prior` which samples from the prior distribution based on the configured `forward_process`? So if `self.forward_process == "uniform"`, we would call `_sample_uniform_tokens` under the hood in `sample_prior` to sample from a uniform prior distribution.

I think this would allow for more graceful support of other possible forward processes, and make the pipeline code cleaner (as most of the logic would be handled inside the scheduler).

src/diffusers/schedulers/scheduling_token_diffusion.py
```diff
@@ -0,0 +1,467 @@
...
+        """
+        if num_inference_steps <= 0:
+            raise ValueError(f"`num_inference_steps` must be > 0, got {num_inference_steps}.")
+        self.num_inference_steps = int(num_inference_steps)
+
+        # Standard diffusers behavior: map inference steps onto training step indices.
+        timesteps = torch.linspace(
+            self.num_train_timesteps - 1, 0, self.num_inference_steps, dtype=torch.float32
+        ).round()
+        self.timesteps = timesteps.to(dtype=torch.long, device=device)
```
dg845, line 184: Suggestion: we could pre-compute the `alpha` schedule here once the timestep discretization is fixed. This could be a little more efficient since I think we use the `alpha` for each timestep twice in the denoising loop (once as `alpha_t` and once as `alpha_prev`).

src/diffusers/schedulers/scheduling_token_diffusion.py
```diff
@@ -0,0 +1,467 @@
...
+
+            if not return_dict:
+                return (x_prev,)
+            return TokenDiffusionSchedulerOutput(prev_sample=x_prev)
+
+        if self.forward_process != "absorbing":
+            raise ValueError(f"Unsupported forward process for `step()`: {self.forward_process!r}")
+
+        # p_denoise = (alpha_prev - alpha_t) / (1 - alpha_t)
+        denom = (1.0 - alpha_t).clamp_min(torch.finfo(torch.float32).eps)
```
dg845: nit: I find this code structure a little confusing, and think that something like

```python
        if self.forward_process == "uniform":
            ...
        elif self.forward_process == "absorbing":
            ...
        else:
             raise ValueError(f"Unsupported forward process for `step()`: {self.forward_process!r}")

        if not return_dict:
            return (x_prev,)
        return TokenDiffusionSchedulerOutput(prev_sample=x_prev)
```

would be more readable.

src/diffusers/schedulers/scheduling_block_token_diffusion.py
```diff
@@ -0,0 +1,95 @@
...
+from __future__ import annotations
+
+from typing import Optional, Tuple, Union
+
+import torch
+
+from .scheduling_token_diffusion import TokenDiffusionScheduler, TokenDiffusionSchedulerOutput
+
+
+class BlockTokenDiffusionScheduler(TokenDiffusionScheduler):
```
dg845: I think we would generally prefer a single file design rather than inheriting from `TokenDiffusionScheduler`. CC @yiyixuxu 

A possible alternative would be to move the `block_mask` logic into the `add_noise` and `step` methods of `TokenDiffusionScheduler`, retaining the current default of noising/denoising over the entire input if `block_mask` is `None`.

src/diffusers/pipelines/token_diffusion/pipeline_token_diffusion.py
```diff
@@ -0,0 +1,265 @@
...
+        # 2. Prepare timesteps
+        device = self._execution_device
+
+        self.scheduler.set_timesteps(num_inference_steps, device=device)
+        timesteps = self.scheduler.timesteps
+        self._num_timesteps = len(timesteps)
+
+        # 3. Prepare latents
+        input_ids = self.prepare_latents(batch_size, seq_len, generator=generator, device=device)
+        attention_mask = torch.ones_like(input_ids, dtype=torch.long)
```
dg845: It looks like `attention_mask` is currently not being used in this pipeline, is this expected?

src/diffusers/pipelines/token_diffusion/pipeline_token_diffusion.py
```diff
@@ -0,0 +1,265 @@
...
+            Sampled token IDs.
+        texts (`list[str]`, *optional*):
+            Decoded texts if a tokenizer was provided and `output_type="text"`.
+    """
+
+    sequences: torch.LongTensor
+    texts: list[str] | None = None
+
+
+class TokenDiffusionPipeline(DiffusionPipeline, DiscreteDiffusionPipelineMixin):
```
dg845, line 61: Can you add a usage example for `TokenDiffusionPipeline`?

src/diffusers/pipelines/hybrid_token_diffusion/pipeline_hybrid_token_diffusion.py
```diff
@@ -0,0 +1,242 @@
...
+            Sampled token IDs.
+        texts (`list[str]`, *optional*):
+            Decoded texts if a tokenizer was provided and `output_type="text"`.
+    """
+
+    sequences: torch.LongTensor
+    texts: list[str] | None = None
+
+
+class HybridTokenDiffusionPipeline(TokenDiffusionPipeline):
```
dg845, line 44: Can we inherit from `DiffusionPipeline` instead of `TokenDiffusionPipeline` and copy over common methods as necessary?

dg845, line 44: My understanding is that `TokenDiffusionPipeline` and `HybridTokenDiffusionPipeline` implement essentially the same logic (but are intended to be used with different schedulers). Would it be possible to consolidate these two pipelines into a single `TokenDiffusionPipeline` which works with both `TokenDiffusionScheduler` and `HybridTokenDiffusionScheduler`?

src/diffusers/pipelines/sdar/pipeline_sdar.py
```diff
@@ -0,0 +1,479 @@
...
+        # 4. Block-wise generation loop
+        block_progress_bar_config = getattr(self, "_progress_bar_config", {}).copy()
+        block_progress_bar_config["position"] = 0
+        block_progress_bar_config["desc"] = "Blocks"
+        for block_idx in tqdm(range(prefill_blocks, int(num_blocks)), **block_progress_bar_config):
+            start = int(block_idx) * int(block_length)
+            end = start + int(block_length)
+            cur_x = x[:, start:end].clone()
+            cur_position_ids = position_ids[:, start:end]
+            cur_attn_mask = attn_mask[start:end, :end].unsqueeze(0)
```
dg845:
```suggestion
            block_x = x[:, start:end].clone()
            block_position_ids = position_ids[:, start:end]
            block_attn_mask = attn_mask[start:end, :end].unsqueeze(0)
```
nit: rename block-level variables to use the prefix `block` (e.g. `cur_x` --> `block_x`) following the LLaDA 2 pipeline.

src/diffusers/pipelines/sdar/pipeline_sdar.py
```diff
@@ -0,0 +1,479 @@
...
+            stop_token_ids = [int(eos_token_id)] if eos_token_id is not None else None
+        if stop_token_ids is not None:
+            stop_token_ids = [int(token_id) for token_id in stop_token_ids]
+
+        self.model.eval()
+        self.scheduler.set_timesteps(int(num_inference_steps), device=device)
+
+        prompt_length = input_ids.shape[1]
+        num_blocks = (prompt_length + int(max_new_tokens) + int(block_length) - 1) // int(block_length)
+        total_length = int(num_blocks) * int(block_length)
```
dg845:
```suggestion
        num_blocks = (prompt_length + max_new_tokens + block_length - 1) // block_length
        total_length = num_blocks * block_length
```
Can we remove the `int(...)` casts here and elsewhere? It makes the code more readable and we are annotating `max_new_tokens` and `block_length` as `int`s so this should be safe.

src/diffusers/pipelines/sdar/pipeline_sdar.py
```diff
@@ -0,0 +1,479 @@
...
+                    temperature=temperature,
+                    top_k=top_k,
+                    top_p=top_p,
+                    generator=generator,
+                    return_dict=True,
+                )
+                cur_x = step_output.prev_sample
+                transfer_index = step_output.transfer_index
+                sampled_tokens = step_output.sampled_tokens
+                sampled_probs = step_output.sampled_probs
```
dg845, line 384:
```suggestion
```
I think we can remove this as these variables are no longer being used.

src/diffusers/pipelines/bd3lm/pipeline_bd3lm.py
```diff
@@ -0,0 +1,382 @@
...
+                if self.scheduler.check_should_stop(x_accum[:, -block_length:], mask_token_id):
+                    progress_bar.update(num_inference_steps - step_idx)
+                    break
+
+                t = self.scheduler.timesteps[step_idx]
+
+                # Get model predictions only when p_x0 cache is invalidated
+                if p_x0_cache is None:
+                    sigma_t = self.scheduler.compute_sigma(t, batch_size)
+                    model_input = x_accum[:, -block_length:]
```
dg845: Would it be possible to refactor the block logic here to be more parallel with other block discrete diffusion pipelines such as `LLaDA2Pipeline`?

https://github.com/huggingface/diffusers/blob/e365d749a1a260e31a1373415255a1288909ef7e/src/diffusers/pipelines/llada2/pipeline_llada2.py#L384-L387

src/diffusers/pipelines/dflash/pipeline_dflash.py
```diff
@@ -0,0 +1,495 @@
...
+    ):
+        super().__init__()
+        if scheduler is None:
+            scheduler = DFlashTokenDiffusionScheduler()
+        self.register_modules(
+            draft_model=draft_model, target_model=target_model, tokenizer=tokenizer, scheduler=scheduler
+        )
+
+    @classmethod
+    def from_pretrained(
```
dg845: Why do we need to override `from_pretrained` for the DFlash pipeline?

src/diffusers/pipelines/dflash/pipeline_dflash.py
```diff
@@ -0,0 +1,495 @@
...
+"""
+
+
+@dataclass
+class DFlashPipelineOutput(BaseOutput):
+    sequences: torch.LongTensor
+    texts: list[str] | None = None
+
+
+def _build_target_layer_ids(num_target_layers: int, num_draft_layers: int) -> list[int]:
```
dg845, line 58: I think inlining this method would be more clear as we only call it once (in `_get_target_layer_ids`).

src/diffusers/pipelines/dflash/pipeline_dflash.py
```diff
@@ -0,0 +1,495 @@
...
+
+        if stop_token_ids is None:
+            eos_token_id = getattr(getattr(self, "tokenizer", None), "eos_token_id", None)
+            stop_token_ids = [int(eos_token_id)] if eos_token_id is not None else None
+        if stop_token_ids is not None:
+            stop_token_ids = [int(token_id) for token_id in stop_token_ids]
+
+        # 3. Setup models and scheduler
+        self.draft_model.eval()
+        self.target_model.eval()
```
dg845:
```suggestion
```
I think the `draft_model` and `target_model` should already be set to `eval` mode, so we don't need to explicitly call it here.

src/diffusers/pipelines/dflash/pipeline_dflash.py
```diff
@@ -0,0 +1,495 @@
...
+
+        texts = None
+        if output_type == "text" and getattr(self, "tokenizer", None) is not None:
+            texts = self.tokenizer.batch_decode(sequences, skip_special_tokens=True)
+
+        if not return_dict:
+            return sequences, texts
+        return DFlashPipelineOutput(sequences=sequences, texts=texts)
+
+    def _get_block_size(self) -> int:
```
dg845, line 462: Do we need to support general `draft` and `target` models in `_get_block_size` and other methods below? My impression is that the DFlash model uses unique modeling logic, so it seems unlikely that we could drop in a random `draft` or `target` model and have it work out of the box. So I think it's reasonable to only support existing DFlash checkpoints such as `z-lab/Qwen3-8B-DFlash-b16`.

src/diffusers/schedulers/scheduling_token_diffusion.py
```diff
@@ -0,0 +1,567 @@
...
+        denom = float(self.num_train_timesteps - 1)
+        return (t_idx / denom).clamp_(0.0, 1.0)
+
+    def _alpha_t(self, t: torch.Tensor) -> torch.Tensor:
+        """
+        Compute alpha(t) for the configured schedule.
+
+        The returned tensor is expected to be in (0, 1] and monotone decreasing in `t`.
+        """
+        if self.alpha_schedule == "log_linear":
```
dg845: Can you give a reference which uses these $`\alpha(t, \epsilon)`$ schedules? The references I can find (e.g. Appendix E.1 of the [MDLM paper](https://arxiv.org/pdf/2406.07524), Table 4 of the [MD4 paper](https://arxiv.org/pdf/2406.04329)) list these alpha schedules without an epsilon argument.

src/diffusers/schedulers/scheduling_token_diffusion.py
```diff
@@ -0,0 +1,567 @@
...
+            )
+        else:
+            raise ValueError(f"Unsupported forward process: {self.forward_process!r}")
+
+        noised = torch.where(replace_positions, replacement, original_samples)
+        if block_mask is not None:
+            noised = torch.where(block_mask.to(device=device), noised, original_samples)
+        return noised
+
+    def enforce_fixed_masks(
```
dg845: I think it might be better to inline `enforce_fixed_masks` in the pipeline code (e.g. in `TokenDiffusionPipeline`), as it is relatively simple and the choice of whether and how to enforce prefix/infill conditioning seems more like a pipeline design choice.

src/diffusers/schedulers/scheduling_hybrid_token_diffusion.py
```diff
@@ -0,0 +1,267 @@
...
+            raise ValueError(f"`num_train_timesteps` must be > 1, got {num_train_timesteps}.")
+        if not (0.0 < t_eps < 0.5):
+            raise ValueError(f"`t_eps` must be in (0, 0.5), got {t_eps}.")
+        if gamma <= 0:
+            raise ValueError(f"`gamma` must be > 0, got {gamma}.")
+
+        self.vocab_size = int(vocab_size)
+        self.mask_token_id = int(mask_token_id)
+        self.num_train_timesteps = int(num_train_timesteps)
+        self.t_eps = float(t_eps)
```
dg845:
```suggestion
```
I think this is unnecessary as `register_to_config` should make these available as `self.config.vocab_size`, `self.config.mask_token_id`, etc.

src/diffusers/schedulers/scheduling_hybrid_token_diffusion.py
```diff
@@ -0,0 +1,267 @@
...
+
+        p_uniform = max(math.exp(-float(clip_noise)), float(p_uniform))
+        log_B = float(gamma) * math.log(2.0) + math.log(p_uniform) - math.log(1.0 - p_uniform)
+        log_B = float(np.clip(log_B, -float(clip_noise), float(clip_noise)))
+        self.log_B = float(log_B)
+        self.log_gamma = float(math.log(float(gamma)))
```
dg845:
```suggestion
        p_uniform = max(math.exp(clip_noise), p_uniform)
        log_B = gamma * math.log(2.0) + math.log(p_uniform) - math.log(1.0 - p_uniform)
        log_B = np.clip(log_B, -clip_noise, clip_noise)
        self.log_B = float(log_B)
        self.log_gamma = math.log(gamma)
```
Can we remove all the `int(...)` and `float(...)` casts here and elsewhere?

src/diffusers/schedulers/scheduling_hybrid_token_diffusion.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@dataclass
+class HybridTokenDiffusionSchedulerOutput(BaseOutput):
+    prev_sample: torch.LongTensor
+
+
+class HybridTokenDiffusionScheduler(SchedulerMixin, ConfigMixin):
+    """
+    Hybrid-transition discrete token diffusion scheduler.
```
dg845, line 37: Can the `__init__` arguments (such as` p_uniform`, `clip_noise`, `gamma`, etc.) be documented in the docstring here, including what they mean and what values might be reasonable for them?

src/diffusers/schedulers/scheduling_hybrid_token_diffusion.py
```diff
@@ -0,0 +1,267 @@
...
+
+    @register_to_config
+    def __init__(
+        self,
+        vocab_size: int,
+        mask_token_id: int,
+        num_train_timesteps: int = 1000,
+        t_eps: float = 1e-4,
+        p_uniform: float = 0.0,
+        clip_noise: float = 20.0,
```
dg845, line 79: The default parameters here set the effective `p_uniform` to $`\exp(-20)`$. If I understand correctly, `p_uniform` corresponds to the maximum probability that tokens will transition to another token uniformly at random instead of to the mask token. Since $`\exp(-20)`$ is very small, I believe the default schedule is thus essentially identical to a normal absorbing/masked diffusion schedule. Is there a reasonable default setting such that this scheduler behaves noticeably different from standard absorbing diffusion (e.g. as implemented by `TokenDiffusionScheduler`)?

src/diffusers/schedulers/scheduling_bd3lm_token_diffusion.py
```diff
@@ -0,0 +1,428 @@
...
+
+        Returns:
+            `torch.Tensor`: Move chance at each timestep value, same shape as *t*.
+        """
+        noise_type = self.config.noise_type
+        eps = 1e-3
+        if noise_type == "loglinear":
+            return t
+        elif noise_type == "cosine":
+            return 1.0 - (1.0 - eps) * torch.cos(t * math.pi / 2.0)
```
dg845, line 123: The alpha schedules here are different from the alpha schedules defined in `TokenDiffusionScheduler`, is this intended? See also https://github.com/huggingface/diffusers/pull/12911#discussion_r3026115270.

src/diffusers/schedulers/scheduling_bd3lm_token_diffusion.py
```diff
@@ -0,0 +1,428 @@
...
+        self.timesteps = torch.linspace(1.0, 0.0, num_inference_steps, device=device)
+        self._dt = 1.0 / num_inference_steps
+
+    # ------------------------------------------------------------------
+    # Noise schedule utilities
+    # ------------------------------------------------------------------
+
+    def _compute_move_chance(self, t: torch.Tensor) -> torch.Tensor:
+        """
+        Compute the probability that a token has been masked (move chance) at continuous time *t*.
```
dg845, line 98: If I understand correctly, the move chance is given by $`1 - \alpha(t)`$. I think documenting the relationship between the `move_chance` and `alpha_t` here would be useful since it would be easier to compare this scheduler with similar schedulers such as `TokenDiffusionScheduler`.

src/diffusers/schedulers/scheduling_bd3lm_token_diffusion.py
```diff
@@ -0,0 +1,428 @@
...
+
+        Args:
+            t (`torch.Tensor`):
+                Continuous timestep values in [0, 1].
+
+        Returns:
+            `torch.Tensor`: Move chance at each timestep value, same shape as *t*.
+        """
+        noise_type = self.config.noise_type
+        eps = 1e-3
```
dg845: I think `eps` should be configurable via `__init__`, following `TokenDiffusionScheduler`.

src/diffusers/schedulers/scheduling_bd3lm_token_diffusion.py
```diff
@@ -0,0 +1,428 @@
...
+        else:
+            t = timestep.to(dtype=torch.float64)
+        if t.dim() == 0:
+            t = t.unsqueeze(0)
+
+        # ------------------------------------------------------------------
+        # Compute move chances at t and s = t - dt
+        # ------------------------------------------------------------------
+        move_chance_t = self._compute_move_chance(t).to(dtype=torch.float64)
+        move_chance_s = self._compute_move_chance(t - dt).to(dtype=torch.float64)
```
dg845: I think getting `s` from the timestep schedule would be better in case we want to support non-`linspace` timestep schedules.

src/diffusers/schedulers/scheduling_bd3lm_token_diffusion.py
```diff
@@ -0,0 +1,428 @@
...
+
+        # ------------------------------------------------------------------
+        # Apply subs parameterization and convert to p(x_0)
+        # ------------------------------------------------------------------
+        logits = model_output[:, -block_size:].to(dtype=torch.float64)
+
+        # Subs parameterization: mask token gets -inf, then log_softmax normalizes.
+        # For unmasked positions, the distribution is forced to be the identity.
+        logits[..., mask_token_id] = -1e9
+        logits = logits - torch.logsumexp(logits, dim=-1, keepdim=True)
```
dg845: My understanding is that this computes a log softmax, would using something like `torch.special.log_softmax` be better here? Also I think it might be more clear to rename this to something like `log_probs`.

src/diffusers/schedulers/scheduling_bd3lm_token_diffusion.py
```diff
@@ -0,0 +1,428 @@
...
+        # Compute posterior q(x_s | x_t, x_0) and sample
+        # ------------------------------------------------------------------
+        # For non-mask tokens: q_xs = p_x0 * (1 - mask_prob), q_xs[mask] = mask_prob
+        q_xs = p_x0 * (1.0 - mask_prob.unsqueeze(-1))
+        q_xs[..., mask_token_id] = mask_prob.squeeze(-1)
+
+        # Gumbel-argmax categorical sampling
+        gumbel_noise = -(torch.rand_like(q_xs, generator=generator) + 1e-10).log()
+        gumbel_noise = (1e-10 + gumbel_noise).clamp(min=1e-30)
+        x_block = (q_xs / gumbel_noise).argmax(dim=-1)
```
dg845: Could we reuse the `_gumbel_argmax` function defined in `scheduling_token_diffusion.py` here?

src/diffusers/schedulers/scheduling_dflash_token_diffusion.py
```diff
@@ -0,0 +1,187 @@
...
+            return torch.argmax(logits, dim=-1)
+        bsz, seq_len, vocab_size = logits.shape
+        flat = logits.view(-1, vocab_size) / float(temperature)
+        probs = torch.softmax(flat, dim=-1)
+        return torch.multinomial(probs, num_samples=1).view(bsz, seq_len)
+
+    def step(
+        self,
+        draft_tokens: torch.LongTensor,
+        target_logits: torch.Tensor,
```
dg845: Would it be possible to refactor the DFlash `step` method such that it follows the standard `step` interface?

https://github.com/huggingface/diffusers/blob/e365d749a1a260e31a1373415255a1288909ef7e/src/diffusers/schedulers/scheduling_block_refinement.py#L166-L170

dg845:
@bot /style

dg845:
@claude, can you review the PR? Please focus on the changes in `/src/diffusers` and their correponding tests in `/tests`.

kashif:
Adding another discrete-diffusion pipeline to the tracker: huggingface/diffusers#13986 adds a DiffusionGemma block-diffusion pipeline, alongside three schedulers (discrete DDIM, entropy bound, and a uniform-corruption mode for `BlockRefinementScheduler`).

It is an encoder-decoder block-diffusion model: the encoder prefills the prompt into a KV cache once and the bidirectional decoder denoises a fixed-size canvas against it, committing confident tokens via the scheduler and renoising the rest. Same `BlockRefinementScheduler` as the LLaDA2 pipeline, and the same caching idea as the dflash pipeline (#13699). Supports text and image prompts, and a `cache_implementation="static"` path so the decoder can be torch.compiled.

PR #12523, opened 2025-10-21
Fix: Add _skip_keys for AutoencoderKLWan

yiyixuxu:
fix https://github.com/huggingface/diffusers/issues/12499

Prevents `feat_cache` and `feat_idx` from being copied during device placement. These are mutable cache objects that must maintain object identity across forward passes for causal conv caching to work.

this fails in `main` but works in this branch

```py
import os
import torch
import numpy as np
from diffusers import WanImageToVideoPipeline
from diffusers.utils import export_to_video, load_image

model_id = "Wan-AI/Wan2.2-I2V-A14B-Diffusers"

dtype = torch.bfloat16
device = "cuda"
device_map = "balanced"

pipe = WanImageToVideoPipeline.from_pretrained(
    model_id, 
    torch_dtype=dtype, 
    device_map="balanced"
)
# pipe.to(device)


image = load_image(
    "https://huggingface.co/datasets/YiYiXu/testing-images/resolve/main/wan_i2v_input.JPG"
)
max_area = 480 * 832
aspect_ratio = image.height / image.width
mod_value = (
    pipe.vae_scale_factor_spatial * pipe.transformer.config.patch_size[1]
)
height = round(np.sqrt(max_area * aspect_ratio)) // mod_value * mod_value
width = round(np.sqrt(max_area / aspect_ratio)) // mod_value * mod_value
image = image.resize((width, height))
prompt = "Summer beach vacation style, a white cat wearing sunglasses sits on a surfboard. The fluffy-furred feline gazes directly at the camera with a relaxed expression. Blurred beach scenery forms the background featuring crystal-clear waters, distant green hills, and a blue sky dotted with white clouds. The cat assumes a naturally relaxed posture, as if savoring the sea breeze and warm sunlight. A close-up shot highlights the feline's intricate details and the refreshing atmosphere of the seaside."

negative_prompt = "色调艳丽，过曝，静态，细节模糊不清，字幕，风格，作品，画作，画面，静止，整体发灰，最差质量，低质量，JPEG压缩残留，丑陋的，残缺的，多余的手指，画得不好的手部，画得不好的脸部，畸形的，毁容的，形态畸形的肢体，手指融合，静止不动的画面，杂乱的背景，三条腿，背景人很多，倒着走"
generator = torch.Generator(device=device).manual_seed(0)
print(f" ")
print(f" running pipeline with device_map: {device_map}")
try:
    output = pipe(
        image=image,
        prompt=prompt,
        negative_prompt=negative_prompt,
        height=height,
[...]

PR #12921, opened 2026-01-07
[GLM-Image] New Models Support

zRzRzRzRzRzRzR:
@yiyixuxu @sayakpaul For check with model


src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,880 @@
...
+            max_sequence_length=max_sequence_length,
+            device=device,
+            dtype=self.dtype,
+        )
+
+        # 4. process images
+        condition_images_prior_token_id = None
+        if image is not None:
+            preprocessed_condition_images = []
+            condition_images_prior_token_id = []
```
yiyixuxu: this needs to be populated no? otherwise, the code path to write cache will be skipped

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,534 @@
...
+
+    The processor supports passing an attention mask for text tokens. The attention mask should have shape (batch_size,
+    text_seq_length) where 1 indicates a non-padded token and 0 indicates a padded token.
+    """
+
+    def __init__(self):
+        if not hasattr(F, "scaled_dot_product_attention"):
+            raise ImportError("GlmImageAttnProcessor requires PyTorch 2.0. To use it, please upgrade PyTorch to 2.0.")
+        self.processor_state = GlmImageAttenProcessorState.ImageGen
+        self.k_cache = None
```
yiyixuxu: so I refactored the cache in this commit: https://github.com/huggingface/diffusers/commit/cfe19a31b9bc14c090c7259c09f3532dfafcd059

Our attention processor should be stateless, so I replaced it with a simple GlmImageKVCache class that gets passed through all the attention layers.
Feel free to cherry-pick it - it should work out of the box (I tested it). But if you want to focus on other things for now (this refactor won't affect the weights), I can help integrate it later once the rest of the code is set.

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: remove now

docs/source/en/api/pipelines/glm_image.md
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+
+# GLM-Image
+
+> [!TIP]
+> Make sure to check out the Schedulers [guide](../../using-diffusers/schedulers) to learn how to explore the tradeoff between scheduler speed and quality, and see the [reuse components across pipelines](../../using-diffusers/loading#reuse-a-pipeline) section to learn how to efficiently load the same components into multiple pipelines.
+
+This pipeline was contributed by [zRzRzRzRzRzRzR](https://github.com/zRzRzRzRzRzRzR). The original codebase can be found [here](https://huggingface.co/zai-org). The original weights can be found under [hf.co/zai-org](https://huggingface.co/zai-org).
```
sayakpaul: Original codebase should likely be a GitHub link?

docs/source/en/api/pipelines/glm_image.md
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+
+# GLM-Image
+
```
sayakpaul, line 17: Let's add a few lines about the model here?

src/diffusers/models/embeddings.py
```diff
@@ -1658,6 +1658,37 @@ def forward(
         return conditioning
 
 
+class GlmImageCombinedTimestepSizeEmbeddings(nn.Module):
```
sayakpaul: We can move this class to `transformer_glm_image.py`. This is what we have been doing for the new models to keep things self-contained.

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,569 @@
...
+            text_attn_mask = text_attn_mask.float().to(query.device)
+            mix_attn_mask = torch.ones((batch_size, text_seq_length + image_seq_length), device=query.device)
+            mix_attn_mask[:, :text_seq_length] = text_attn_mask
+            mix_attn_mask = mix_attn_mask.unsqueeze(2)
+            attn_mask_matrix = mix_attn_mask @ mix_attn_mask.transpose(1, 2)
+            attention_mask = (attn_mask_matrix > 0).unsqueeze(1).to(query.dtype)
+
+        hidden_states = F.scaled_dot_product_attention(
+            query, key, value, attn_mask=attention_mask, dropout_p=0.0, is_causal=False
+        )
```
sayakpaul: Could we make it implemented in a similar way as
https://github.com/huggingface/diffusers/blob/8600b4c10d67b0ce200f664204358747bd53c775/src/diffusers/models/transformers/transformer_flux.py#L118 ?

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,569 @@
...
+            norm_encoder_hidden_states,
+            c_gate_msa,
+            c_shift_mlp,
+            c_scale_mlp,
+            c_gate_mlp,
+        ) = self.norm1(hidden_states, encoder_hidden_states, temb)
+
+        # 2. Attention
+        if attention_kwargs is None:
+            attention_kwargs = {}
```
sayakpaul:
```suggestion
        attention_kwargs = attention_kwargs or {}
```

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,569 @@
...
+        crop_coords: torch.Tensor,
+        attention_kwargs: Optional[Dict[str, Any]] = None,
+        return_dict: bool = True,
+        attention_mask: Optional[torch.Tensor] = None,
+        kv_caches: Optional[GlmImageKVCache] = None,
+        image_rotary_emb: Optional[
+            Union[Tuple[torch.Tensor, torch.Tensor], List[Tuple[torch.Tensor, torch.Tensor]]]
+        ] = None,
+    ) -> Union[Tuple[torch.Tensor], Transformer2DModelOutput]:
+        batch_size, num_channels, height, width = hidden_states.shape
```
sayakpaul, line 564: Could we add this bit of code?
https://github.com/huggingface/diffusers/blob/8600b4c10d67b0ce200f664204358747bd53c775/src/diffusers/models/transformers/transformer_flux.py#L678-L692

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,569 @@
...
+        hidden_states = self.image_projector(hidden_states)
+        encoder_hidden_states = self.glyph_projector(encoder_hidden_states)
+        prior_embedding = self.prior_token_embedding(prior_token_id)
+        prior_embedding[prior_token_drop] *= 0.0
+        prior_hidden_states = self.prior_projector(prior_embedding)
+
+        hidden_states = hidden_states + prior_hidden_states
+
+        temb = self.time_condition_embed(timestep, target_size, crop_coords, hidden_states.dtype)
+        temb = F.silu(temb)
```
sayakpaul: (nit): can this `F.silu()` go inside `time_condition_embed`?

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,569 @@
...
+                    kv_cache=kv_caches[idx] if kv_caches is not None else None,
+                )
+
+        # 4. Output norm & projection
+        hidden_states = self.norm_out(hidden_states, temb)
+        hidden_states = self.proj_out(hidden_states)
+
+        # 5. Unpatchify
+        hidden_states = hidden_states.reshape(batch_size, post_patch_height, post_patch_width, -1, p, p)
+        output = hidden_states.permute(0, 3, 1, 4, 2, 5).flatten(4, 5).flatten(2, 3)
```
sayakpaul, line 617: (nit): would be nice to add a comment on the shapes (the shape we're transitioning from and the shape we're going into).

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+import numpy as np
+import PIL
+import torch
+from transformers import ByT5Tokenizer, GlmImageForConditionalGeneration, GlmImageProcessor, T5EncoderModel
+
+from ...callbacks import MultiPipelineCallbacks, PipelineCallback
+from ...image_processor import VaeImageProcessor
+from ...loaders import CogView4LoraLoaderMixin
```
sayakpaul: Would it work out of the box?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+EXAMPLE_DOC_STRING = """
+    Examples:
+        ```python
+        >>> import torch
+        >>> from diffusers import GlmImagePipeline
+
+        >>> pipe = GlmImagePipeline.from_pretrained("zai-org/GLM-Image", torch_dtype=torch.bfloat16)
+        >>> pipe.to("cuda")
+
+        >>> prompt = "A photo of an astronaut riding a horse on mars<sop>36 24<eop>"
```
sayakpaul: I thought we were handling the "<sop>" part within the `encode_prompt()` method?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+    base_seq_len: int = 256,
+    base_shift: float = 0.25,
+    max_shift: float = 0.75,
+) -> float:
+    m = (image_seq_len / base_seq_len) ** 0.5
+    mu = m * max_shift + base_shift
+    return mu
+
+
+def retrieve_timesteps(
```
sayakpaul, line 78:
```suggestion
# Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion.retrieve_timesteps
def retrieve_timesteps(
```

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+        vision_language_encoder ([`GlmImageForConditionalGeneration`]):
+            The AR model that generates image tokens from text prompts.
+        transformer ([`GlmImageTransformer2DModel`]):
+            A text conditioned transformer to denoise the encoded image latents (DiT).
+        scheduler ([`SchedulerMixin`]):
+            A scheduler to be used in combination with `transformer` to denoise the encoded image latents.
+    """
+
+    _optional_components = []
+    model_cpu_offload_seq = "transformer->vae"
```
sayakpaul: This seems weird. The pattern is usually:
https://github.com/huggingface/diffusers/blob/8600b4c10d67b0ce200f664204358747bd53c775/src/diffusers/pipelines/flux/pipeline_flux.py#L180

Any reason why this is different?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+
+        self.default_sample_size = (
+            self.transformer.config.sample_size
+            if hasattr(self, "transformer")
+            and self.transformer is not None
+            and hasattr(self.transformer.config, "sample_size")
+            else 128
+        )
+
+    def _build_image_grid_thw(
```
sayakpaul: (nit): I don't think there's any use of `self` here. So, probably all the methods that don't need access to `self` could be turned into `staticmethod`s?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+
+    def _extract_large_image_tokens(
+        self, outputs: torch.Tensor, input_length: int, large_image_start_offset: int, large_image_tokens: int
+    ) -> torch.Tensor:
+        generated_tokens = outputs[0][input_length:]
+        large_image_start = large_image_start_offset
+        large_image_end = large_image_start + large_image_tokens
+        return generated_tokens[large_image_start:large_image_end]
+
+    def _upsample_d32_to_d16(self, token_ids: torch.Tensor, token_h: int, token_w: int) -> torch.Tensor:
```
sayakpaul: But it's not changing the data-type to FP16, no? It is resizing the token IDs by a factor of 2. Maybe `_upsample_token_ids` is a better name?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+        input_ids = torch.tensor(
+            [input_ids_ + [self.tokenizer.pad_token_id] * (max_length - len(input_ids_)) for input_ids_ in input_ids],
+            device=device,
+        )
+        outputs = self.text_encoder(input_ids, attention_mask=attention_mask)
+        glyph_embeds = outputs.last_hidden_state[attention_mask.bool()].unsqueeze(0)
+
+        return glyph_embeds.to(device=device, dtype=dtype)
+
+    def encode_prompt(
```
sayakpaul, line 358: So this is exclusively for the glyph text or does it work on others, too?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+                raise TypeError(
+                    f"`negative_prompt` should be the same type to `prompt`, but got {type(negative_prompt)} !="
+                    f" {type(prompt)}."
+                )
+            elif batch_size != len(negative_prompt):
+                raise ValueError(
+                    f"`negative_prompt`: {negative_prompt} has batch size {len(negative_prompt)}, but `prompt`:"
+                    f" {prompt} has batch size {batch_size}. Please make sure that passed `negative_prompt` matches"
+                    " the batch size of `prompt`."
+                )
```
sayakpaul: (nit): We could move all of this to `check_inputs()`?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+                .view(1, self.vae.config.latent_channels, 1, 1)
+                .to(self.vae.device, self.vae.dtype)
+            )
+            empty_glyph_hiddens = torch.zeros_like(prompt_embeds)[:1, :0, ...]
+            for condition_image, condition_image_prior_token_id in zip(image, condition_images_prior_token_id):
+                condition_image = condition_image.to(device=device, dtype=self.vae.dtype)
+                condition_latent = retrieve_latents(
+                    self.vae.encode(condition_image), generator=generator, sample_mode="argmax"
+                )
+                condition_latent = (condition_latent - latents_mean) / latents_std
```
sayakpaul: (nit): Could we move this to a method like `encode_vae_image()` or `encode_image()`?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,876 @@
...
+                    hidden_states=condition_latent,
+                    encoder_hidden_states=empty_glyph_hiddens,
+                    prior_token_id=condition_image_prior_token_id,
+                    prior_token_drop=torch.full_like(condition_image_prior_token_id, False, dtype=torch.bool),
+                    timestep=torch.zeros((1,), device=device),
+                    target_size=torch.tensor([condition_image.shape[-2:]], device=device),
+                    crop_coords=torch.zeros((1, 2), device=device),
+                    attention_kwargs=attention_kwargs,
+                    kv_caches=kv_caches,
+                )
```
sayakpaul, line 695: So, this is just for populating the `kv_cache`? Do we have any other utilization of this?

sayakpaul (commented):
Looking quite good!

I think all the precomputations are in place and the use of caching also reads quite simple.

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: only for glyph text, for others using GlmImageForConditionGeneration

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: This is a cherry-pick of @yiyixuxu's commit.

sayakpaul: That's fine. This was more of a clarification question.

zRzRzRzRzRzRzR: It would be useful to run the reference image through a forward pass at timestep 0 and keep the KV cache.

sayakpaul: Can we add a comment about it then?

docs/source/en/api/pipelines/glm_image.md
zRzRzRzRzRzRzR: Alright, we're filling it in.

docs/source/en/api/pipelines/glm_image.md
zRzRzRzRzRzRzR: Without the original code repository, this initial release only has the diffusers version. github with using diffusers and sglang quick start

src/diffusers/models/transformers/transformer_glm_image.py
zRzRzRzRzRzRzR: This part was directly copied from the CogView4 model. I see many models using this style, but I haven't written in the Flux style. If I have time, I can change it, but I'm not sure if there will be enough time.

sayakpaul: Okay I will try to help out -:)

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: Yes, for DiT, it's a structure, but you probably can't use the weights directly

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: I believe this was a typo and has been resolved

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: Should it be the way I've modified it now?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: Yes, but I think it should serve this class alone?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: The determination of negative_prompt should be directly removed, as GLM-Image cannot support negative_prompt.

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
sayakpaul: I would then suggest removing it because we don't know how it will play out.

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
sayakpaul: Yes. Hence I think they can be made staticmethods.

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,869 @@
...
+        vision_language_encoder ([`GlmImageForConditionalGeneration`]):
+            The AR model that generates image tokens from text prompts.
+        transformer ([`GlmImageTransformer2DModel`]):
+            A text conditioned transformer to denoise the encoded image latents (DiT).
+        scheduler ([`SchedulerMixin`]):
+            A scheduler to be used in combination with `transformer` to denoise the encoded image latents.
+    """
+
+    _optional_components = []
+    model_cpu_offload_seq = "text_encoder->vision_language_encoder->transformer->vae"
```
sayakpaul: This looks good @zRzRzRzRzRzRzR!

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,869 @@
...
+                " only forward one of the two."
+            )
+        elif prompt is None and prompt_embeds is None:
+            raise ValueError(
+                "Provide either `prompt` or `prompt_embeds`. Cannot leave both `prompt` and `prompt_embeds` undefined."
+            )
+        elif prompt is not None and (not isinstance(prompt, str) and not isinstance(prompt, list)):
+            raise ValueError(f"`prompt` has to be of type `str` or `list` but is {type(prompt)}")
+
+        if do_classifier_free_guidance:
```
sayakpaul: Perfect!

src/diffusers/models/transformers/transformer_glm_image.py
zRzRzRzRzRzRzR: Alright, this is exactly the same as cogview4 here.

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,869 @@
...
+        if prompt_embeds is None:
+            prompt_embeds = self._get_glyph_embeds(prompt, max_sequence_length, device, dtype)
+
+        seq_len = prompt_embeds.size(1)
+        prompt_embeds = prompt_embeds.repeat(1, num_images_per_prompt, 1)
+        prompt_embeds = prompt_embeds.view(batch_size * num_images_per_prompt, seq_len, -1)
+
+        negative_prompt_embeds = None
+        if do_classifier_free_guidance:
+            negative_prompt = ""
```
sayakpaul, line 406: So users are not supposed to pass a custom `negative_prompt` as it is always fixed to ""? Should it be `negative_prompt = negative_prompt or ""`?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: ok

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: ok I will add it here

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: Yes, Not allow to use negative_prompt, we did not train negative_prompt.
so it shoud be “”?

sayakpaul: Yes you'd know that better. Then we shouldn't allow users to pass `negative_prompt`, right?

zRzRzRzRzRzRzR: yes, not allow

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,813 @@
...
+        self._current_timestep = None
+        self._interrupt = False
+
+        if prompt is not None and isinstance(prompt, str):
+            batch_size = 1
+        elif prompt is not None and isinstance(prompt, list):
+            batch_size = len(prompt)
+        else:
+            batch_size = prompt_embeds.shape[0]
+        assert batch_size == 1, "batch_size must be 1"
```
yiyixuxu:
```suggestion
        if batch_size != 1:
            raise ValueError(f"batch_size must be 1 due to AR model limitations, got {batch_size}")
```

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,813 @@
...
+        vae ([`AutoencoderKL`]):
+            Variational Auto-Encoder (VAE) Model to encode and decode images to and from latent representations.
+        transformer ([`GlmImageTransformer2DModel`]):
+            A text conditioned transformer to denoise the encoded image latents (DiT).
+        scheduler ([`SchedulerMixin`]):
+            A scheduler to be used in combination with `transformer` to denoise the encoded image latents.
+    """
+
+    _optional_components = []
+    model_cpu_offload_seq = "text_encoder->vision_language_encoder->transformer->vae"
```
yiyixuxu: I think the vision_language_encoder is ran before text_encoder? so the order here is not correct, no?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,813 @@
...
+        assert batch_size == 1, "batch_size must be 1"
+
+        device = self._execution_device
+
+        prior_token_id, prior_token_image_ids, ar_height, ar_width = self.generate_prior_tokens(
+            prompt=prompt[0] if isinstance(prompt, list) else prompt,
+            image=image,
+            height=height,
+            width=width,
+        )
```
yiyixuxu: a few things here:
1. `generate_prior_tokens` will error out if `height = None` and `width = None`
2. `ar_height`/`ar_width` is pretty straightfoward to calculate, let's calculate them seperately for clarity
3. we can update generate_prior_tokens to only return two tokens, this way it is easier for user to skip this stage reusing pre-computed tokens

here is just a suggestion, I'm not completely ure the logic to assign defaut height/width are correct 

```suggestion
        height = height or self.default_sample_size * self.vae_scale_factor
        width = width or self.default_sample_size * self.vae_scale_factor
        
        height = (height // 32) * 32
        width = (width //32) * 32
      
        prior_token_id, prior_token_image_ids = self.generate_prior_tokens(
            prompt=prompt[0] if isinstance(prompt, list) else prompt,
            image=image,
            height=height,
            width=width,
        )
```

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,813 @@
...
+
+        height = height or ar_height
+        width = width or ar_width
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
yiyixuxu: ohh can we use an emtpy tensor then? e.g. `torch.zero_like(prompt_embeds)`

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,813 @@
...
+                image_height, image_width = img.size[::-1] if isinstance(img, PIL.Image.Image) else img.shape[:2]
+                multiple_of = self.vae_scale_factor * self.transformer.config.patch_size
+                image_height = (image_height // multiple_of) * multiple_of
+                image_width = (image_width // multiple_of) * multiple_of
+                img = self.image_processor.preprocess(img, height=image_height, width=image_width)
+                preprocessed_condition_images.append(img)
+                height = height or image_height
+                width = width or image_width
+            image = preprocessed_condition_images
+        height = height or self.default_sample_size * self.vae_scale_factor
```
yiyixuxu: i found the logic for setting default `height`/`width` is a bit confusing - it's scattered in multiple places: e.g. after `generate_prior_tokens`, during image processing, and here again.

But we did `height = height or ar_height` right after `generate_prior_tokens`, so they're already set there - the later assignments never trigger.

Not sure what the correct/intended logic is. Could you clean this up?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,813 @@
...
+            )
+
+        if callback_on_step_end_tensor_inputs is not None and not all(
+            k in self._callback_tensor_inputs for k in callback_on_step_end_tensor_inputs
+        ):
+            raise ValueError(
+                f"`callback_on_step_end_tensor_inputs` has to be in {self._callback_tensor_inputs}, but found {[k for k in callback_on_step_end_tensor_inputs if k not in self._callback_tensor_inputs]}"
+            )
+
+        if prompt is not None and prompt_embeds is not None:
```
yiyixuxu, line 462: so, the current code structure won't work with`prompt=None`, when the `prompt_embeds` is passed, - we still need `prompt`  generate tokens using the AR model

I think we'd need to accept both `prior_token_id` and `prompt_embeds` as inputs if `prompt` is `None`. so something like

```python
if prompt is None:
    if prior_token_id is None or prompt_embeds is None:
        raise ValueError(
            "When `prompt` is not provided, both `prior_token_id` and `prompt_embeds` must be passed."
        )
```
you also need to add the `prior_token_id` to pipeline input

yiyixuxu (commented):
thanks, looking great and super excited about this model 
I left some comments, mostly, I'm a bit confused on the correct logic to set `height`/`width`

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,613 @@
...
+            to determine the resolution as `sample_size * vae_scale_factor => 128 * 8 => 1024`
+    """
+
+    _supports_gradient_checkpointing = True
+    _no_split_modules = [
+        "GlmImageTransformerBlock",
+        "GlmImageImageProjector",
+        "GlmImageImageProjector",
+    ]
+    _skip_layerwise_casting_patterns = ["patch_embed", "norm", "proj_out"]
```
yiyixuxu, line 492:
```suggestion
    _skip_layerwise_casting_patterns = ["patch_embed", "norm", "proj_out"]
    _skip_keys = ["kv_caches"]
```
so that it works with device hooks (see more details https://github.com/huggingface/diffusers/pull/12523)

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: prior_token_id implementation must be generated by AR so prompt must not be none

zRzRzRzRzRzRzR: let me change logic of this

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: Yes, they're actually the same, since the height obtained from AR is also the size of the target image. Let me delete them.

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: Let me add a check to ensure that height and width cannot be None. This is a strict requirement, as these two parameters must be present for the AR model to correctly output tokens

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: > ohh can we use an emtpy tensor then? e.g. torch.zero_like(prompt_embeds)

We can’t use torch.zeros_like(prompt_embeds) here.

In our model, an empty string ("") does not produce a zero tensor. Instead, it results in two padding text embeddings, and their length is not guaranteed to match the positive prompt embeddings. They are also not all zeros.

Because of this:
+ The negative embeddings may have a different sequence length than prompt_embeds
+ Using zeros_like(prompt_embeds) would be semantically incorrect and could break the alignment logic

If we don’t want to consider a negative prompt, the safer approach is to precompute the embeddings for an empty text once and reuse/cache them, rather than creating a zero tensor.

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,614 @@
...
+        self.k_cache = None
+        self.v_cache = None
+        self.mode: Optional[str] = None  # "write", "read", "skip"
+
+    def store(self, k: torch.Tensor, v: torch.Tensor):
+        if self.k_cache is None:
+            self.k_cache = k
+            self.v_cache = v
+        else:
+            self.k_cache = torch.cat([self.k_cache, k], dim=2)
```
yhyang201: Referring to L253, should dim be equal to 1 here?

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,614 @@
...
+            self.v_cache = torch.cat([self.v_cache, v], dim=2)
+
+    def get(self):
```
yiyixuxu: not sure if it should be 1 or 2 but they should be same 

probably better move the logic together like this so mistakes like https://github.com/huggingface/diffusers/pull/12921/files#r2678634789 is less likely to happen

```suggestion
    def get(self, k: torch.Tensor, v: torch.Tensor):
         k_cache = torch.cat([self.k_cache, key], dim=2)
         v_cache = torch.cat([self.v_cache, key], dim=2)
```

src/diffusers/models/transformers/transformer_glm_image.py
```diff
@@ -0,0 +1,614 @@
...
+                key[:, text_seq_length:, :, :], image_rotary_emb, sequence_dim=1, use_real_unbind_dim=-2
+            )
+
+        if kv_cache is not None:
+            if kv_cache.mode == "write":
+                kv_cache.store(key, value)
+            elif kv_cache.mode == "read":
+                k_cache, v_cache = kv_cache.get()
+                key = torch.cat([k_cache, key], dim=1) if k_cache is not None else key
+                value = torch.cat([v_cache, value], dim=1) if v_cache is not None else value
```
yiyixuxu:
```suggestion
                key, value = kv_cache.get(key, value) if kv_cache is not None
```

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,804 @@
...
+        height: Optional[int] = None,
+        width: Optional[int] = None,
+        num_inference_steps: int = 50,
+        timesteps: Optional[List[int]] = None,
+        sigmas: Optional[List[float]] = None,
+        guidance_scale: float = 1.5,
+        num_images_per_prompt: int = 1,
+        generator: Optional[Union[torch.Generator, List[torch.Generator]]] = None,
+        latents: Optional[torch.FloatTensor] = None,
+        prompt_embeds: Optional[torch.FloatTensor] = None,
```
yiyixuxu:
```suggestion
        prompt_embeds: Optional[torch.Tensor] = None,
        negative_prompt_embeds: Optional[torch.Tensor] = None,
        prior_token_ids: Optional[torch.Tensor] = None,
        prior_image_token_ids: Optional[torch.Tensor] = None
```
we should allow them to pre-compute the tokens since it is the most compute expensive part 
we should allow them to pass pre-compute negative_prompt_embeds too because it is fixed

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,804 @@
...
+        elif prompt is not None and isinstance(prompt, list):
+            batch_size = len(prompt)
+        else:
+            batch_size = prompt_embeds.shape[0]
+        if batch_size != 1:
+            raise ValueError(f"batch_size must be 1 due to AR model limitations, got {batch_size}")
+
+        device = self._execution_device
+
+        prior_token_id, prior_token_image_ids = self.generate_prior_tokens(
```
yiyixuxu:
```suggestion
        if prior_token_ids is None:
            prior_token_id, prior_token_image_ids = self.generate_prior_tokens( ...)
```

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,804 @@
...
+        width: int,
+        image: Optional[List[PIL.Image.Image]] = None,
+        factor: int = 32,
+    ) -> Tuple[torch.Tensor, int, int]:
+        device = self.vision_language_encoder.device
+        height = (height // factor) * factor
+        width = (width // factor) * factor
+        is_text_to_image = image is None or len(image) == 0
+        expanded_prompt, token_h, token_w, prev_h, prev_w = self._build_prompt_with_shape(
+            prompt, height, width, is_text_to_image
```
zucchini-nlp: @zRzRzRzRzRzRzR we should have already moved this to `processor.__call__` in transformers, no?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,804 @@
...
+        ).to(device)
+        existing_grid = inputs.get("image_grid_thw")
+        inputs["image_grid_thw"] = self._build_image_grid_thw(
+            token_h,
+            token_w,
+            prev_h,
+            prev_w,
+            existing_grid=existing_grid if not is_text_to_image else None,
+            device=device,
+        )
```
zucchini-nlp: and this one as well @zRzRzRzRzRzRzR

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,804 @@
...
+
+        prior_token_image_ids = None
+        if image is not None:
+            prior_token_image_embed = self.vision_language_encoder.get_image_features(
+                inputs["pixel_values"], existing_grid
+            )
+            prior_token_image_embed = torch.cat(prior_token_image_embed, dim=0)
+            prior_token_image_ids = self.vision_language_encoder.get_image_tokens(
+                prior_token_image_embed, existing_grid
+            )
```
zucchini-nlp, line 299: if there is an image, then the vision encoder+vq will run twice. Once right here and once in `model.generate()`. 

If overriding `generate()` in transformers is too much work (and code), we can also merge input ids here. So after getting image ids, we call `inputs["input_ids"] = self.get_placeholder_mask(input_ids, image_ids)`

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,804 @@
...
+        )
+        content = []
+        if image is not None:
+            for img in image:
+                content.append({"type": "image", "image": img})
+        content.append({"type": "text", "text": expanded_prompt})
+        messages = [{"role": "user", "content": content}]
+        inputs = self.processor.apply_chat_template(
+            messages, add_generation_prompt=True, tokenize=True, return_dict=True, return_tensors="pt"
+        ).to(device)
```
zucchini-nlp, line 284: we can try passing here smaller size by adjusting default values (in the hub - `"min_pixels": 262144, "max_pixels": 4194304`)

I don't know how the downstream quality will be affected, though it will make the resulting `pixel_values` smaller -> vision attention with less tokens -> less image ids at the end in text-image-to-text case

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
sayakpaul: @zRzRzRzRzRzRzR was this addressed?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
sayakpaul: @zRzRzRzRzRzRzR was this address?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
sayakpaul: @zRzRzRzRzRzRzR

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: Isn't this configured through the hub's configuration file, preprocessor.json?

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
zRzRzRzRzRzRzR: Alright, I understand. Let me add this part of the functionality

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
yiyixuxu: I think it does not apply anymore we now accept `prior_token_ids` and `prior_image_token_id`

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,790 @@
...
+                "Provide either `prompt` or `prompt_embeds`. Cannot leave both `prompt` and `prompt_embeds` undefined."
+            )
+        elif prompt is not None and (not isinstance(prompt, str) and not isinstance(prompt, list)):
+            raise ValueError(f"`prompt` has to be of type `str` or `list` but is {type(prompt)}")
+        if negative_prompt is not None and negative_prompt_embeds is not None:
+            raise ValueError(
+                f"Cannot forward both `negative_prompt`: {negative_prompt} and `negative_prompt_embeds`:"
+                f" {negative_prompt_embeds}. Please make sure to only forward one of the two."
+            )
+
```
yiyixuxu:
```suggestion
        if prompt is not None and prior_token_ids is not None:
            raise ValueError(
                f"Cannot forward both `prompt`: {prompt} and `prior_token_ids`: {prior_token_ids}. Please make sure to"
                " only forward one of the two."
         elif prompt is None and prior_token_ids is None:
            raise ValueError(
                "Provide either `prompt` or `prior_token_ids`. Cannot leave both `prompt` and `prior_token_ids` undefined."
            )
```

yiyixuxu (approved):
thanks!

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
sayakpaul: We can configure this from the pipeline implementation. For example, QwenImage Edit Plus does it like this:

https://github.com/huggingface/diffusers/blob/9d68742214b9491718487b63f25d6e8236ed56ea/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py#L682-L691

It's not exactly the same but you get the idea.

sayakpaul:
@zRzRzRzRzRzRzR opened a PR https://github.com/zRzRzRzRzRzRzR/diffusers/pull/5 for resolving the consistency problems.

sayakpaul:
@bot /style

src/diffusers/pipelines/glm_image/pipeline_glm_image.py
```diff
@@ -0,0 +1,823 @@
...
+        if prompt_embeds is None:
+            prompt_embeds = self._get_glyph_embeds(prompt, max_sequence_length, device, dtype)
+
+        seq_len = prompt_embeds.size(1)
+        prompt_embeds = prompt_embeds.repeat(1, num_images_per_prompt, 1)
+        prompt_embeds = prompt_embeds.view(batch_size * num_images_per_prompt, seq_len, -1)
+
+        # For GLM-Image, negative_prompt must be "" instead of None
+        negative_prompt_embeds = None
```
yiyixuxu:
```suggestion
```

zRzRzRzRzRzRzR:
Using as #12973 , merged

PR #12922, opened 2026-01-07
Ltx 2 latent upsample pipeline

sayakpaul:
# What does this PR do?

Upsampling pipeline by @dg845.

We need to open a PR to the LTX-2 repo with the upsampler as well.

Cc: @matanby


dg845 (approved):
Thanks!

PR #12923, opened 2026-01-07
Torchao floatx version guard

howardzhang-cv:
Motivation: TorchAO removing floatx support
Changes summary: added version guard in quantization_config.py
added version guard in quantization tests in test_torchao.py
added new test for version guarding in test_torchao.py (test_floatx_quantization)

Testing:
pytest tests/quantization/torchao/test_torchao.py -k test_floatx_quantization

### DIFF
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -1106,6 +1106,51 @@ def _sage_attention_backward_op(
     raise NotImplementedError("Backward pass is not implemented for Sage attention.")
 
 
+def _npu_attention_forward_op(
+    ctx: torch.autograd.function.FunctionCtx,
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
+    attn_mask: Optional[torch.Tensor] = None,
+    dropout_p: float = 0.0,
+    is_causal: bool = False,
+    scale: Optional[float] = None,
+    enable_gqa: bool = False,
+    return_lse: bool = False,
+    _save_ctx: bool = True,
+    _parallel_config: Optional["ParallelConfig"] = None,
+):
+    if return_lse:
+        raise ValueError("NPU attention backend does not support setting `return_lse=True`.")
+
+    out = npu_fusion_attention(
+        query,
+        key,
+        value,
+        query.size(2),  # num_heads
+        input_layout="BSND",
+        pse=None,
+        scale=1.0 / math.sqrt(query.shape[-1]) if scale is None else scale,
+        pre_tockens=65536,
+        next_tockens=65536,
+        keep_prob=1.0 - dropout_p,
+        sync=False,
+        inner_precise=0,
+    )[0]
+
+    return out
+
+
+# Not implemented yet.
+def _npu_attention_backward_op(
+    ctx: torch.autograd.function.FunctionCtx,
+    grad_out: torch.Tensor,
+    *args,
+    **kwargs,
+):
+    raise NotImplementedError("Backward pass is not implemented for Npu Fusion Attention.")
+
+
 # ===== Context parallel =====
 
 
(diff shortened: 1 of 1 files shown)


howardzhang-cv:
Also wanted to note that it seems like a lot of the tests in test_torchao (like test_quantization for example) were failing in the main branch

tests/quantization/torchao/test_torchao.py
```diff
@@ -256,9 +256,12 @@ def test_quantization(self):
                     # Cutlass fails to initialize for below
                     # ("float8dq_e4m3_row", np.array([0, 0, 0, 0, 0, 0, 0, 0, 0])),
                     # =====
-                    ("fp4", np.array([0.4668, 0.5195, 0.5547, 0.4199, 0.4434, 0.6445, 0.4316, 0.4531, 0.5625])),
-                    ("fp6", np.array([0.4668, 0.5195, 0.5547, 0.4199, 0.4434, 0.6445, 0.4316, 0.4531, 0.5625])),
                 ])
+                if version.parse(importlib.metadata.version("torchao")) <= version.Version("0.14.0"):
```
jerryzh168: use `is_torchao_version` here as well?

src/diffusers/quantizers/quantization_config.py
```diff
@@ -650,18 +657,21 @@ def generate_float8dq_types(dtype: torch.dtype):
                 return types
 
             def generate_fpx_quantization_types(bits: int):
-                types = {}
+                if is_torchao_version("<=", "0.14.0"):
```
jerryzh168: nit: we can put 0.14.1 to be exact I think, same for others

tests/quantization/torchao/test_torchao.py
howardzhang-cv: The import would be a bit ugly for it lives in a different folder, also it seemed like other torchao version checks in that file seemed to use this syntax. Should I still change it?

jerryzh168 (approved):
LGTM

src/diffusers/quantizers/quantization_config.py
```diff
@@ -457,7 +457,7 @@ class TorchAoConfig(QuantizationConfigMixin):
                     - Shorthands: `float8wo`, `float8wo_e5m2`, `float8wo_e4m3`, `float8dq`, `float8dq_e4m3`,
                       `float8_e4m3_tensor`, `float8_e4m3_row`,
 
-                - **Floating point X-bit quantization:**
+                - **Floating point X-bit quantization:** (in torchao <= 0.14.1, not supported in torchao >= 0.15.0)
```
sayakpaul, line 460: Can we also provide a recommendation on what needs to be done if a user wants to use X-bit ops in torchao >= 0.15.0?

tests/quantization/torchao/test_torchao.py
```diff
@@ -271,6 +274,33 @@ def test_quantization(self):
                 )
                 self._test_quant_type(quantization_config, expected_slice, model_id)
 
+    def test_floatx_quantization(self):
```
sayakpaul, line 278: I don't think floatX is used much so we can avoid this test.

src/diffusers/quantizers/quantization_config.py
jerryzh168: they will need to downgrade to 0.14.1 or lower, we are removing this because we didn't see much usage for this feature

sayakpaul: Can we make that clear in the error message then? 

> they will need to downgrade to 0.14.1 or lower

tests/quantization/torchao/test_torchao.py
howardzhang-cv: Added a skip flag for this. We can also just remove it entirely if you think that's better.

src/diffusers/quantizers/quantization_config.py
howardzhang-cv: There's an error message for this in line 541 for if the user is trying to use the floatx in version >0.14.1. Is there a better place for it?

PR #12933, opened 2026-01-09
[LoRA] add LoRA support to LTX-2

sayakpaul:
# What does this PR do?

Adds support for loading non-diffusers LoRA into `LTX2Pipeline`. More specifically, [this checkpoint](https://huggingface.co/Lightricks/LTX-2/blob/main/ltx-2-19b-distilled-lora-384.safetensors) or any other LoRA checkpoints that are obtained with the trainer shipped by the [official codebase](https://github.com/Lightricks/LTX-2/).

The said LoRA is also crucial for reducing the number of inference steps and seems to be also crucial for the two-stage pipeline as implemented in [here](https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-pipelines/src/ltx_pipelines/ti2vid_two_stages.py).

I decided to give [this LoRA](https://huggingface.co/Lightricks/LTX-2/blob/main/ltx-2-19b-distilled-lora-384.safetensors) a try on the single-stage T2V pipeline and I am getting decent results:

https://github.com/user-attachments/assets/7df89b1b-3776-42b1-a6cb-29cb2c8ec7db

<details>
<summary>Code</summary>

```py
import torch

DEFAULT_NEGATIVE_PROMPT = (
    "blurry, out of focus, overexposed, underexposed, low contrast, washed out colors, excessive noise, "
    "grainy texture, poor lighting, flickering, motion blur, distorted proportions, unnatural skin tones, "
    "deformed facial features, asymmetrical face, missing facial features, extra limbs, disfigured hands, "
    "wrong hand count, artifacts around text, inconsistent perspective, camera shake, incorrect depth of "
    "field, background too sharp, background clutter, distracting reflections, harsh shadows, inconsistent "
    "lighting direction, color banding, cartoonish rendering, 3D CGI look, unrealistic materials, uncanny "
    "valley effect, incorrect ethnicity, wrong gender, exaggerated expressions, wrong gaze direction, "
    "mismatched lip sync, silent or muted audio, distorted voice, robotic voice, echo, background noise, "
    "off-sync audio, incorrect dialogue, added dialogue, repetitive speech, jittery movement, awkward "
    "pauses, incorrect timing, unnatural transitions, inconsistent framing, tilted camera, flat lighting, "
    "inconsistent tone, cinematic oversaturation, stylized filters, or AI artifacts."
)

pipe = DiffusionPipeline.from_pretrained("Lightricks/LTX-2", torch_dtype=torch.bfloat16)
pipe.load_lora_weights("Lightricks/LTX-2", weight_name="ltx-2-19b-distilled-lora-384.safetensors")

frame_rate = 24.0
prompt = "a beautiful sunset by the beach with some glimpse of mountains."
video, audio = pipe(
    prompt=prompt,
    negative_prompt=DEFAULT_NEGATIVE_PROMPT,
    width=768,
    height=512,
    num_frames=121,
    frame_rate=frame_rate,
    num_inference_steps=8,
    guidance_scale=1.0,
    output_type="np",
    return_dict=False,
)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
    audio_sample_rate=pipe.vocoder.config.output_sampling_rate,  # should be 24000
    output_path="video_distilled.mp4",
)
```

</details>

Note that we need the following diff on `pipeline_ltx2.py` as well:

```diff
diff --git a/src/diffusers/pipelines/ltx2/pipeline_ltx2.py b/src/diffusers/pipelines/ltx2/pipeline_ltx2.py
index 9cf847926..46600451b 100644
--- a/src/diffusers/pipelines/ltx2/pipeline_ltx2.py
+++ b/src/diffusers/pipelines/ltx2/pipeline_ltx2.py
@@ -959,6 +959,7 @@ class LTX2Pipeline(DiffusionPipeline, FromSingleFileMixin, LTX2LoraLoaderMixin):
 
         # 5. Prepare timesteps
         sigmas = np.linspace(1.0, 1 / num_inference_steps, num_inference_steps)
+        sigmas = np.array([1.0, 0.99375, 0.9875, 0.98125, 0.975, 0.909375, 0.725, 0.421875])
         mu = calculate_shift(
             video_sequence_length,
```

The `sigmas` above are pre-computed and come from [here](https://github.com/Lightricks/LTX-2/blob/628956009ca14446f01976ba4e861e6a9e210a93/packages/ltx-pipelines/src/ltx_pipelines/utils/constants.py#L9).

Cc: @asomoza @linoytsaban for awareness.

### DIFF
--- a/examples/dreambooth/train_dreambooth_lora_flux2.py
+++ b/examples/dreambooth/train_dreambooth_lora_flux2.py
@@ -1228,7 +1228,7 @@ def main(args):
         else {"device": accelerator.device, "dtype": weight_dtype}
     )
 
-    is_fsdp = accelerator.state.fsdp_plugin is not None
+    is_fsdp = getattr(accelerator.state, "fsdp_plugin", None) is not None
     if not is_fsdp:
         transformer.to(**transformer_to_kwargs)
 
--- a/examples/dreambooth/train_dreambooth_lora_flux2_img2img.py
+++ b/examples/dreambooth/train_dreambooth_lora_flux2_img2img.py
@@ -1178,7 +1178,7 @@ def main(args):
         else {"device": accelerator.device, "dtype": weight_dtype}
     )
 
-    is_fsdp = accelerator.state.fsdp_plugin is not None
+    is_fsdp = getattr(accelerator.state, "fsdp_plugin", None) is not None
     if not is_fsdp:
         transformer.to(**transformer_to_kwargs)
 


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -3011,6 +3013,233 @@ def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
...
 
 
+class LTX2LoraLoaderMixin(LoraBaseMixin):
+    r"""
+    Load LoRA layers into [`LTX2VideoTransformer3DModel`]. Specific to [`LTX2Pipeline`].
+    """
+
+    _lora_loadable_modules = ["transformer", "connectors"]
+    transformer_name = TRANSFORMER_NAME
+    connectors_name = LTX2_CONNECTOR_NAME
```
sayakpaul, line 3023: The distilled LoRA checkpoint needs the `connectors` component of the pipeline to be loaded with LoRA too.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -3011,6 +3013,233 @@ def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
...
+            transformer_peft_state_dict,
+            transformer=getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer,
+            adapter_name=adapter_name,
+            metadata=metadata,
+            _pipeline=self,
+            low_cpu_mem_usage=low_cpu_mem_usage,
+            hotswap=hotswap,
+        )
+        if connectors_peft_state_dict:
+            self.load_lora_into_transformer(
```
sayakpaul, line 3134: This should be fine.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -3011,6 +3013,233 @@ def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
...
+                adapter_name=adapter_name,
+                metadata=metadata,
+                _pipeline=self,
+                low_cpu_mem_usage=low_cpu_mem_usage,
+                hotswap=hotswap,
+                prefix=self.connectors_name,
+            )
+
+    @classmethod
+    def load_lora_into_transformer(
```
sayakpaul, line 3148: The signature has the `prefix` argument added which is why it differs from the rest of the bunch.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -3011,6 +3013,233 @@ def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
...
+        if isinstance(pretrained_model_name_or_path_or_dict, dict):
+            pretrained_model_name_or_path_or_dict = pretrained_model_name_or_path_or_dict.copy()
+
+        # First, ensure that the checkpoint is a compatible one and can be successfully loaded.
+        kwargs["return_lora_metadata"] = True
+        state_dict, metadata = self.lora_state_dict(pretrained_model_name_or_path_or_dict, **kwargs)
+
+        is_correct_format = all("lora" in key for key in state_dict.keys())
+        if not is_correct_format:
+            raise ValueError("Invalid LoRA checkpoint.")
```
dg845, line 3118:
```suggestion
        if not is_correct_format:
            raise ValueError("Invalid LoRA checkpoint. Make sure all LoRA param names contain `lora`.")
```
nit, non-blocking: Could we have a more informative error message here? I'm not sure if the suggestion is exactly correct but I think we should give an indication of what a valid LoRA checkpoint should look like.

src/diffusers/loaders/lora_pipeline.py
dg845: Maybe it would make sense to rename `load_lora_into_transformer` to `load_lora_into_modules` or have separate functions `load_lora_into_transformer` and `load_lora_into_connectors` analogous to how `StableDiffusionLoraLoaderMixin` does it? IIUC `load_lora_weights` is the intended entry point, so renaming/refactoring this method shouldn't disrupt how `LTX2LoraLoaderMixin` is used.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -3011,6 +3013,233 @@ def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
...
+            is_main_process=is_main_process,
+            weight_name=weight_name,
+            save_function=save_function,
+            safe_serialization=safe_serialization,
+        )
+
+    # Copied from diffusers.loaders.lora_pipeline.CogVideoXLoraLoaderMixin.fuse_lora
+    def fuse_lora(
+        self,
+        components: List[str] = ["transformer"],
```
dg845, line 3218:
```suggestion
        components: List[str] = ["transformer", "connectors"],
```
Should the default argument for `components` include `"connectors"`?

Also, I think it might make sense to refactor this into something like

```python
    def fuse_lora(
        self,
        components: List[str] = [],
        lora_scale: float = 1.0,
        safe_fusing: bool = False,
        adapter_names: Optional[List[str]] = None,
        **kwargs,
    ):
        if len(components) == 0:
            components = self._lora_loadable_modules
        # Rest of implementation same as before
        ...
```
Furthermore, since `_lora_loadable_modules` exists in the parent class `LoraBaseMixin`, we could push this logic into `LoraBaseMixin.fuse_lora` and not have to keep overriding `fuse_lora` in the child classes.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -3011,6 +3013,233 @@ def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
...
+        super().fuse_lora(
+            components=components,
+            lora_scale=lora_scale,
+            safe_fusing=safe_fusing,
+            adapter_names=adapter_names,
+            **kwargs,
+        )
+
+    # Copied from diffusers.loaders.lora_pipeline.CogVideoXLoraLoaderMixin.unfuse_lora
+    def unfuse_lora(self, components: List[str] = ["transformer"], **kwargs):
```
dg845, line 3236:
```suggestion
    def unfuse_lora(self, components: List[str] = ["transformer", "connectors"], **kwargs):
```
See https://github.com/huggingface/diffusers/pull/12933#discussion_r2677867590.

tests/lora/test_lora_layers_ltx2.py
```diff
@@ -0,0 +1,293 @@
...
+
+    @unittest.skip("Not supported in LTX2.")
+    def test_simple_inference_with_text_denoiser_block_scale_for_all_dict_options(self):
+        pass
+
+    @unittest.skip("Not supported in LTX2.")
+    def test_modify_padding_mode(self):
+        pass
+
+    @unittest.skip("Text encoder LoRA is not supported in LTX2.")
```
dg845, line 271: Perhaps we could have a flag like `supports_text_encoder_lora` in `PeftLoraLoaderMixinTests` that can skip all of the text encoder LoRA tests at once?

sayakpaul: Sure. Do you want to take a crack at that?

src/diffusers/loaders/lora_pipeline.py
sayakpaul: It's not too bad this way I guess because the `connectors` component is a mini transformer in itself. `load_lora_into_modules` would be a bit inexplicit which we want to avoid.

src/diffusers/loaders/lora_pipeline.py
sayakpaul: Suggestion looks correct and it applies module-wide. Maybe this could clubbed with https://github.com/huggingface/diffusers/pull/12933#discussion_r2678292594 if you want to take a crack?

src/diffusers/loaders/lora_pipeline.py
sayakpaul: We can keep the default as is and then follow the rest. But on a second thought, we would rather want to have the users pass the component names they want to fuse explicitly. 

Perhaps, we can add validation checks like:

```py
if not components:
   raise ValueError

if any(c not in self._lora_loadable_modules for c in components):
    raise ValueError
```

Something like this?

sayakpaul: Furthermore, I assume `connectors` shouldn't be a default component to be fused. It's probably only going to be applicable for the distilled LoRA checkpoint but not to other checkpoints (like the IC LoRAs LTX2 made available; [example](https://huggingface.co/Lightricks/LTX-2-19b-IC-LoRA-Canny-Control/tree/main?show_file_info=ltx-2-19b-ic-lora-canny-control.safetensors)).

Does this make sense?

src/diffusers/loaders/lora_pipeline.py
sayakpaul: See https://github.com/huggingface/diffusers/pull/12933#discussion_r2678299373

sayakpaul:
This PR also allows loading IC LoRAs: https://huggingface.co/Lightricks/LTX-2-19b-IC-LoRA-Canny-Control 

@linoytsaban  / @asomoza do you want to give this a try? I think you would first need to make the I2V pipeline class (`LTX2ImageToVideoPipeline`) inherit from the `LTX2LoraLoaderMixin` class and then pass a canny input image to start the video generation process.

src/diffusers/loaders/lora_pipeline.py
dg845: > Perhaps, we can add validation checks like:

I think those validation checks are already implemented:

https://github.com/huggingface/diffusers/blob/02c7adc3568478777f1235b8755863cb1fc36c8c/src/diffusers/loaders/lora_base.py#L593-L601

In my opinion the ideal behavior is that `fuse_lora()` can be called without any arguments, and this should attempt to fuse all (active?) LoRAs in all possible modules; that is, all modules in `_lora_loadable_modules`. If a module in `_lora_loadable_modules` doesn't have any LoRAs which target it, this will be handled gracefully (presumably as a no-op). A user can then explicitly specify `components` to fuse if they want finer control (for example, they only want the LoRAs to be fused on some modules). `unfuse_lora` should then have the analogous behavior.

However, I'm not sure how feasible this is for the current implementation, and it's possible that my view is mistaken (maybe there are some factors I haven't considered?).

sayakpaul: > In my opinion the ideal behavior is that fuse_lora() can be called without any arguments, and this should attempt to fuse all (active?) LoRAs in all possible modules; that is, all modules in _lora_loadable_modules

Actually, there can be several considerations:

* Load one LoRA in say, text encoder.
* Load two LoRAs into the DiT, fuse them, keep them unfused, etc.

This is the reason why we allow the users to pass `components` explicitly. I think this makes the feature more configurable.

> If a module in _lora_loadable_modules doesn't have any LoRAs which target it, this will be handled gracefully (presumably as a no-op).

It will be a no-op but that is an implicit behaviour which I would like to avoid

PR #12958, opened 2026-01-10
[modular] add auto_docstring & more doc related refactors 

yiyixuxu:
This PR adds a utility script `utils/modular_auto_docstring.py` that automatically generates docstrings for modular pipeline block classes from their `doc` property.

### Usage

Mark classes with `# auto_docstring` comment:

```python
# auto_docstring
class QwenImageAutoVaeEncoderStep(AutoPipelineBlocks):
    block_classes = [QwenImageInpaintVaeEncoderStep, QwenImageImg2ImgVaeEncoderStep]
    block_names = ["inpaint", "img2img"]
    block_trigger_inputs = ["mask_image", "image"]

    @property
    def doc(self):
        return (
            "Vae encoder step that encodes image inputs into latent representations.\n"
            "This is an auto pipeline block.\n"
            " - `QwenImageInpaintVaeEncoderStep` (inpaint) is used when `mask_image` is provided.\n"
            " - `QwenImageImg2ImgVaeEncoderStep` (img2img) is used when `image` is provided."
        )
```

Run the script to insert docstrings:

```bash
python utils/modular_auto_docstring.py --fix_and_overwrite
```


src/diffusers/modular_pipelines/modular_pipeline_utils.py
```diff
@@ -342,6 +343,185 @@ class InputParam:
...
+        if hasattr(cls, name) and callable(getattr(cls, name)):
+            return getattr(cls, name)()
+        return None
+
+    # ======================================================
+    # InputParam templates
+    # ======================================================
+
+    @classmethod
+    def prompt(cls) -> "InputParam":
```
yiyixuxu: our pipeline parameter are pretty consistent across different pipelines, e.g. you always have `prompt`, `height`, `width`, `num_inference_steps`, etc. I made template for these common ones, so that it is easier to define 

before you need 
```py
InputParam(
    name="prompt",
    type_hint=str,
    required=True,
    description="The prompt or prompts to guide image generation."
)
```
now you do 
```py
InputParam.prompt()
InputParam.height(default=1024)
InputParam.num_inference_steps(default=28)
InputParam.generator()
```

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -59,11 +55,127 @@
...
 
 # ====================
-# 1. VAE ENCODER
+# 1. TEXT ENCODER
 # ====================
 
 
+# auto_docstring
+class QwenImageAutoTextEncoderStep(AutoPipelineBlocks):
+    """
```
yiyixuxu, line 65: these are all auto generated docstring

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -59,11 +55,127 @@
 
 
 # ====================
-# 1. VAE ENCODER
+# 1. TEXT ENCODER
 # ====================
 
 
+# auto_docstring
```
yiyixuxu, line 63: add this mark and then run
```bash
python utils/modular_auto_docstring.py --fix_and_overwrite
```

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -59,11 +55,127 @@
...
+
+          prompt_template_encode_start_idx (default: 34)
+
+          tokenizer_max_length (default: 1024)
+
+      Inputs:
+
+          prompt (`str`, *optional*):
+              The prompt or prompts to guide image generation.
+
```
DN6: Nit but I would format so that there's no newline between fields.

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -59,11 +55,127 @@
...
 # ====================
-# 1. VAE ENCODER
+# 1. TEXT ENCODER
 # ====================
 
 
+# auto_docstring
+class QwenImageAutoTextEncoderStep(AutoPipelineBlocks):
+    """
+    class QwenImageAutoTextEncoderStep
```
DN6: Would remove this line being generated in the auto doc

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -59,11 +55,127 @@
...
+    class QwenImageInpaintVaeEncoderStep
+
+      This step is used for processing image and mask inputs for inpainting tasks. It:
+       - Resizes the image to the target size, based on `height` and `width`.
+       - Processes and updates `image` and `mask_image`.
+       - Creates `image_latents`.
+
+      Components:
+
+          image_mask_processor (`InpaintProcessor`) [subfolder=]
```
DN6: IMO just the type hint for the component is sufficient. No need for subfolder information

src/diffusers/modular_pipelines/modular_pipeline_utils.py
DN6: I'm a bit apprehensive about introducing dedicated class methods for common parameters in this way. I think the class can become quite large as common inputs expand. 

I would prefer to keep current syntax (IMO this ensures InputParams are defined in a consistent way) and use post init on the dataclass to automatically add a description. e.g

```python
# centralised descriptions would live somewhere like constants.py 
# can be used for modular + non-modular
INPUT_PARAM_TEMPLATES = {
      "prompt": {"type_hint": str, "required": True, "description": "The prompt or prompts to guide image generation."},
      "height": {"type_hint": int, "description": "The height in pixels of the generated image."},
      "width": {"type_hint": int, "description": "The width in pixels of the generated image."},
      "generator": {"type_hint": torch.Generator, "description": "Torch generator for deterministic generation."},
      # ...
  }

  @dataclass
  class InputParam:
      name: str = None
      type_hint: Any = None
      required: bool = False
      default: Any = None
      description: str = None

  def __post_init__(self):
      if not self.name or self.name not in INPUT_PARAM_TEMPLATES:
          return

      template = INPUT_PARAM_TEMPLATES[self.name]
      if self.type_hint is None:
          self.type_hint = template.get("type_hint")
      if self.description is None:
          self.description = template.get("description")
``` 

If we feel that methods for these inputs are necessary, one way to address it without adding individual methods to the InputParam is to use a metaclass. It would result in the InputParam object being less crowded.  

```python
  class InputParamMeta(type):
      def __getattr__(cls, name: str):
          if name in INPUT_PARAM_TEMPLATES:
              def factory(**overrides):
                  return cls(name=name, **overrides)
              return factory
          raise AttributeError(f"No template named '{name}'")


  @dataclass
  class InputParam(metaclass=InputParamMeta):
```

sayakpaul (commented):
I guess it will be better to review the PR after https://github.com/huggingface/diffusers/pull/12958/files#r2689367739 ? 👀

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
sayakpaul: We could add this to `make style` and error out in the CI if it's not the case? 👀

This will help us enforce consistency.

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -59,11 +55,127 @@
...
+# auto_docstring
+class QwenImageAutoTextEncoderStep(AutoPipelineBlocks):
+    """
+    class QwenImageAutoTextEncoderStep
+
+      Text encoder step that encodes the text prompt into a text embedding. This is an auto pipeline block.
+
+      Components:
+
+          text_encoder (`Qwen2_5_VLForConditionalGeneration`): The text encoder to use [subfolder=]
```
sayakpaul: `[subfolder=]` is a bit hard to understand.

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -59,11 +55,127 @@
...
+    model_name = "qwenimage"
+    block_classes = [QwenImageTextEncoderStep()]
+    block_names = ["text_encoder"]
+    block_trigger_inputs = ["prompt"]
+
+    @property
+    def description(self) -> str:
+        return "Text encoder step that encodes the text prompt into a text embedding. This is an auto pipeline block."
+        " - `QwenImageTextEncoderStep` (text_encoder) is used when `prompt` is provided."
+        " - if `prompt` is not provided, step will be skipped."
```
sayakpaul, line 100: Could this be autogenerated as well provided a class name?

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -59,11 +55,127 @@
...
+
+      Components:
+
+          image_mask_processor (`InpaintProcessor`) [subfolder=]
+
+          vae (`AutoencoderKLQwenImage`) [subfolder=]
+
+      Inputs:
+
+          mask_image (`Image`):
```
sayakpaul, line 120: No strong opinion but I find `PIL.Image.Image` a tad bit more informative as a type hint.

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -78,7 +190,41 @@ def description(self) -> str:
...
+              The width in pixels of the generated image.
+
+          generator (`Generator`, *optional*):
+              Torch generator for deterministic generation.
+
+      Outputs:
+
+          processed_image (`None`):
+
+          image_latents (`Tensor`):
```
sayakpaul, line 179: Similarly `torch.Tensor`?

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
```diff
@@ -78,7 +190,41 @@ def description(self) -> str:
...
+          image (`Image`):
+              Input image for img2img, editing, or conditioning.
+
+          height (`int`, *optional*):
+              The height in pixels of the generated image.
+
+          width (`int`, *optional*):
+              The width in pixels of the generated image.
+
+          generator (`Generator`, *optional*):
```
sayakpaul, line 173: `torch.Generator` perhaps?

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
sayakpaul: I think this is fine to keep as the flow of information feels more defined and specific under it.

src/diffusers/modular_pipelines/modular_pipeline_utils.py
```diff
@@ -337,26 +519,66 @@ class InputParam:
...
+        template_kwargs.update(overrides)
+        return cls(name=name, **template_kwargs)
+
 
 @dataclass
 class OutputParam:
     """Specification for an output parameter."""
 
-    name: str
+    name: str = None
```
yiyixuxu:
```suggestion
    name: str 
```

src/diffusers/modular_pipelines/qwenimage/before_denoise.py
```diff
@@ -532,15 +672,17 @@ def expected_components(self) -> List[ComponentSpec]:
     @property
     def inputs(self) -> List[InputParam]:
         return [
-            InputParam("num_inference_steps", default=50, type_hint=int),
-            InputParam("sigmas", type_hint=List[float]),
-            InputParam("image_latents", required=True, type_hint=torch.Tensor),
+            InputParam.template("num_inference_steps"),
+            InputParam.template("sigmas"),
+            InputParam.template("image_latents", note="Can be generated from vae encoder and packed in input step."),
```
yiyixuxu:
```suggestion
            InputParam.template("image_latents"),
```

src/diffusers/modular_pipelines/modular_pipeline_utils.py
yiyixuxu: I removed the class methods! 
also  moved the common parameter definitions out of `InputParam`/`OutputParam` as well. They're currently still in the same file, but we can move them elsewhere later.

To use a predefined param:

```python
InputParam.template("image")
```

i went with this API, instead  of the `__post_init__`  because  you have to explicitly opt into using the template, even though it's slightly more verbose.

The main thing I wanted to avoid is with `__post_init__` auto-filling, if you make a param that exists in our template but actually intend to customize its docstring later, e.g.  you do something like`InputParam(name="image")`, you will silently get the template defaults, and it is hard to notice. 

With `.template()`, it's more clear when the user intends to use the pre-defined definition vs write a custom param. so for the same example, if someone writes `InputParam(name="image")` without filling in fields, the docstring will show "TODO: Add description".

we can also override fields in the template like this:

```python
InputParam.template("image", note="resized")  # appends a note to the description
InputParam.template("image", required=False)
```

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
yiyixuxu: yes, we should do that eventually! 
(won't include in this PR though)

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
yiyixuxu: good idea!
I think it should be doable, I will look into it (not in this PR)

src/diffusers/modular_pipelines/qwenimage/modular_blocks_qwenimage.py
yiyixuxu: I will look into improving them in a follow-up

yiyixuxu:
@bot /style

src/diffusers/modular_pipelines/qwenimage/decoders.py
```diff
@@ -272,15 +391,19 @@ def expected_components(self) -> List[ComponentSpec]:
     @property
     def inputs(self) -> List[InputParam]:
         return [
-            InputParam("images", required=True, description="the generated image from decoders step"),
             InputParam(
-                name="output_type",
-                default="pil",
-                type_hint=str,
-                description="The type of the output images, can be 'pil', 'np', 'pt'",
+                name="images",
```
sayakpaul, line 395: Why not make it a `template(...)` as well?

src/diffusers/modular_pipelines/qwenimage/before_denoise.py
```diff
@@ -118,7 +118,40 @@ def get_timesteps(scheduler, num_inference_steps, strength):
 # ====================
 
 
+# auto_docstring
 class QwenImagePrepareLatentsStep(ModularPipelineBlocks):
+    """
+    Prepare initial random noise for the generation process
```
sayakpaul, line 124: One bit that is not clear to me yet is how we provide this short description.

sayakpaul: I am guessing it's through `doc()`?

src/diffusers/modular_pipelines/qwenimage/decoders.py
yiyixuxu: it is not exactly the final image output, the `image` output in this block is it's the torch tensor fresh out of vae, and we need to apply post-processing still 

that's also one reason I want to use template explicitly, instead of auto-fill. these variables change their definition through the blocks sometimes, another example is `latent` and we only define template for the common pipeline inputs/outputs that our users familar with (we should keep these definition reletively consistent between modular and standard pipelines as well)

src/diffusers/modular_pipelines/qwenimage/before_denoise.py
yiyixuxu: ohh through the `description` property that you define https://github.com/huggingface/diffusers/blob/main/src/diffusers/modular_pipelines/qwenimage/before_denoise.py#L125

`doc` only put stuff together into a docstring, it does not come up with short descriptions on its own:)

yiyixuxu: we will write a doc page on doc later :)

utils/modular_auto_docstring.py
```diff
@@ -0,0 +1,352 @@
...
+    Returns list of classes that need updating.
+    """
+    classes_to_update = find_auto_docstring_classes(filepath)
+
+    if not classes_to_update:
+        return []
+
+    if not overwrite:
+        # Check mode: only verify that docstrings exist
+        # Content comparison is not reliable due to formatting differences
```
yiyixuxu, line 255: @sayakpaul 

I updated the script so that:

when you run `python utils/modular_auto_docstring.py --fix_and_overwrite`, it now runs `ruff check --fix`, `ruff format`, and `doc-builder style` after generating docstrings, so the output is compliant with our formatters.

when you run `python utils/modular_auto_docstring.py` without `--fix_and_overwrite` (as used in CI), it now only verifies that docstrings exist for classes marked with `# auto_docstring`. otherwise the check_auto_docs will throw errors if though we already run the auto docstring and also the make style and such; 

 We can add stricter change detection later if needed. (I don't have too much time in fixing this right now)

sayakpaul: Sounds good!

yiyixuxu:
@DN6 i merged this since I have a couple other PRs built on top of this
let me know if you have any more feedbacks, I can work on a new PR. feel free to open a PR as you see fit too

PR #12972, opened 2026-01-13
Remove 8bit device restriction

SunMarc:
# What does this PR do?

This PR allow to move 8-bit models quantized with bnb across devices. It was added in 0.48.0 version of bnb and we recently upstreamed the changes to accelerate and transformers.

accelerate PR https://github.com/huggingface/accelerate/pull/3907
transformers PR https://github.com/huggingface/transformers/pull/43258

All tests are passing except `tests/quantization/bnb/test_mixed_int8.py::Bnb8BitCompileTests::test_torch_compile` that is failing just like before. The issue is not due to the hook as we don't add them anymore. 

Also, note that if a user is on the latest commit of accelerate but didn't update diffusers, they will get an error when loading a 8-bit model.

### DIFF
--- a/src/diffusers/pipelines/pipeline_utils.py
+++ b/src/diffusers/pipelines/pipeline_utils.py
@@ -445,6 +445,7 @@ class DiffusionPipeline(ConfigMixin, PushToHubMixin):
 
             _, _, is_loaded_in_8bit_bnb = _check_bnb_status(module)
 
+            # https://github.com/huggingface/accelerate/pull/3907
             if is_loaded_in_8bit_bnb and (
                 is_bitsandbytes_version("<", "0.48.0") or is_accelerate_version("<", "1.13.0.dev0")
             ):


src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -444,7 +445,9 @@ def module_is_sequentially_offloaded(module):
 
             _, _, is_loaded_in_8bit_bnb = _check_bnb_status(module)
 
-            if is_loaded_in_8bit_bnb:
+            if is_loaded_in_8bit_bnb and (
+                is_bitsandbytes_version("<", "0.48.0") or is_accelerate_version("<", "1.13.0.dev0")
```
sayakpaul, line 450: Let's try to point to the `accelerate` PR needed here?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -542,6 +546,14 @@ def module_is_offloaded(module):
             # https://github.com/huggingface/transformers/pull/33122. So, we guard this accordingly.
             if is_loaded_in_4bit_bnb and device is not None and is_transformers_version(">", "4.44.0"):
                 module.to(device=device)
+            # added here https://github.com/huggingface/transformers/pull/43258
+            if (
+                is_loaded_in_8bit_bnb
+                and device is not None
+                and is_transformers_version(">", "4.58.0")
```
sayakpaul, line 554: Why is the transformers version relevant here?

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1223,7 +1235,9 @@ def enable_model_cpu_offload(self, gpu_id: Optional[int] = None, device: Union[t
 
             # This is because the model would already be placed on a CUDA device.
             _, _, is_loaded_in_8bit_bnb = _check_bnb_status(model)
-            if is_loaded_in_8bit_bnb:
+            if is_loaded_in_8bit_bnb and (
+                is_transformers_version("<", "4.58.0") or is_bitsandbytes_version("<", "0.48.0")
```
sayakpaul, line 1240: Same.

tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -858,6 +856,5 @@ def test_torch_compile(self):
     def test_torch_compile_with_cpu_offload(self):
         super()._test_torch_compile_with_cpu_offload(torch_dtype=torch.float16)
 
-    @pytest.mark.xfail(reason="Test fails because of an offloading problem from Accelerate with confusion in hooks.")
```
sayakpaul, line 861: Very cool!

tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -848,7 +846,7 @@ def quantization_config(self):
         )
 
     @pytest.mark.xfail(
-        reason="Test fails because of an offloading problem from Accelerate with confusion in hooks."
+        reason="Test fails because of a type change when recompiling."
```
sayakpaul, line 849: What is the error now? 👁️

sayakpaul (commented):
Thanks a lot for this!

tests/quantization/bnb/test_mixed_int8.py
SunMarc: it didn't change, the issue was not with accelerate from the start.

src/diffusers/pipelines/pipeline_utils.py
SunMarc: replicating the same pattern as with 4-bit. Transformers version is relevant because it will raise an error otherwise when using to()

src/diffusers/pipelines/pipeline_utils.py
SunMarc: it's to make sure that we can move the module since we are adding a cpu_offload hook.

src/diffusers/pipelines/pipeline_utils.py
SunMarc: okay !

src/diffusers/pipelines/pipeline_utils.py
sayakpaul: Just question: will there be a stable release after 4.57.5 for `transformers`? But I guess this is strong enough.

SunMarc: probably not, only patches

PR #12987, opened 2026-01-16
[Qwen] avoid creating attention masks when there is no padding

kashif:
# What does this PR do?

This pull request sets all ones masks to None.

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/loaders/peft.py
+++ b/src/diffusers/loaders/peft.py
@@ -478,7 +478,7 @@ class PeftAdapterMixin:
         Args:
             adapter_names (`List[str]` or `str`):
                 The names of the adapters to use.
-            adapter_weights (`Union[List[float], float]`, *optional*):
+            weights (`Union[List[float], float]`, *optional*):
                 The adapter(s) weights to use with the UNet. If `None`, the weights are set to `1.0` for all the
                 adapters.
 
@@ -495,7 +495,7 @@ class PeftAdapterMixin:
             "jbilcke-hf/sdxl-cinematic-1", weight_name="pytorch_lora_weights.safetensors", adapter_name="cinematic"
         )
         pipeline.load_lora_weights("nerijs/pixel-art-xl", weight_name="pixel-art-xl.safetensors", adapter_name="pixel")
-        pipeline.unet.set_adapters(["cinematic", "pixel"], adapter_weights=[0.5, 0.5])
+        pipeline.unet.set_adapters(["cinematic", "pixel"], weights=[0.5, 0.5])
         ```
         """
         if not USE_PEFT_BACKEND:
--- a/src/diffusers/schedulers/scheduling_consistency_decoder.py
+++ b/src/diffusers/schedulers/scheduling_consistency_decoder.py
@@ -14,7 +14,7 @@ from .scheduling_utils import SchedulerMixin
 def betas_for_alpha_bar(
     num_diffusion_timesteps: int,
     max_beta: float = 0.999,
-    alpha_transform_type: Literal["cosine", "exp"] = "cosine",
+    alpha_transform_type: Literal["cosine", "exp", "laplace"] = "cosine",
 ) -> torch.Tensor:
     """
     Create a beta schedule that discretizes the given alpha_t_bar function, which defines the cumulative product of
@@ -28,8 +28,8 @@ def betas_for_alpha_bar(
             The number of betas to produce.
         max_beta (`float`, defaults to `0.999`):
             The maximum beta to use; use values lower than 1 to avoid numerical instability.
-        alpha_transform_type (`"cosine"` or `"exp"`, defaults to `"cosine"`):
-            The type of noise schedule for `alpha_bar`. Choose from `cosine` or `exp`.
+        alpha_transform_type (`str`, defaults to `"cosine"`):
+            The type of noise schedule for `alpha_bar`. Choose from `cosine`, `exp`, or `laplace`.
 
     Returns:
         `torch.Tensor`:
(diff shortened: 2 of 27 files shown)


kashif:
will do!

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_img2img.py
```diff
@@ -217,14 +218,7 @@ def _get_qwen_prompt_embeds(
...
         split_hidden_states = [e[drop_idx:] for e in split_hidden_states]
-        attn_mask_list = [torch.ones(e.size(0), dtype=torch.long, device=e.device) for e in split_hidden_states]
-        max_seq_len = max([e.size(0) for e in split_hidden_states])
-        prompt_embeds = torch.stack(
-            [torch.cat([u, u.new_zeros(max_seq_len - u.size(0), u.size(1))]) for u in split_hidden_states]
-        )
-        encoder_attention_mask = torch.stack(
-            [torch.cat([u, u.new_zeros(max_seq_len - u.size(0))]) for u in attn_mask_list]
-        )
+        prompt_embeds, encoder_attention_mask = build_prompt_embeds_and_mask(split_hidden_states)
```
yiyixuxu: curiously, cant we just do

```py
if encoder_attention_mask.all():
    encoder_attention_mask = None
```

kashif: the issue was that this was giving trouble with torch.compile...

yiyixuxu: ohh, but I think we typically just torch.compile the models (e.g. `pipeline.transformer.compile(...)`, `pipeline.text_encoder.compile()`)
the computations on the pipeline won't be part of it though

kashif: oh ok let me check that then

kashif: added checks for `.all()`

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -276,3 +276,45 @@ def prepare_dummy_input(self, height, width):
...
     def test_torch_compile_recompilation_and_graph_break(self):
         super().test_torch_compile_recompilation_and_graph_break()
+
+    def test_torch_compile_with_and_without_mask(self):
+        """Test that torch.compile works with both None mask and padding mask."""
+        init_dict, inputs = self.prepare_init_args_and_inputs_for_common()
+        model = self.model_class(**init_dict).to(torch_device)
+        model.eval()
+
+        compiled_model = torch.compile(model, mode="default", fullgraph=False)
```
sayakpaul: Some notes:

* Usually, it should be `model.compile()` as it doesn't wrap the underlying `model` into a dynamo wrapper. This way, we don't have to add any extra code to handle it.
* Why is `fullgraph=False` here?

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -276,3 +276,45 @@ def prepare_dummy_input(self, height, width):
...
+
+        # Test 1: Run with None mask (no padding, all tokens are valid)
+        inputs_no_mask = inputs.copy()
+        inputs_no_mask["encoder_hidden_states_mask"] = None
+
+        with torch.no_grad():
+            output_no_mask = compiled_model(**inputs_no_mask)
```
sayakpaul: Does it lead to graph breaks? If not, then we should add additional contexts:
https://github.com/huggingface/diffusers/blob/3996788b602eaae4da41a1d45726b62e662b73cf/tests/models/test_modeling_common.py#L2145

dxqb:
does this PR lack a merge, or is the amount of code changes intentional and really part of this PR only? (+318 - 170)
if intentional, could you explain what it does and why all these changes are necessary?

avoiding masks if they're not necessary has been one line before
`attention_mask = attention_mask if not torch.all(text_attention_mask) else None`
(on a boolean mask)

kashif:
@dxqb i intially mis-calculated and was adding support for the pipeline to be compiled, i will revert and simplify

dxqb:
> @dxqb i intially mis-calculated and was adding support for the pipeline to be compiled, i will revert and simplify

thanks!
regarding compile,
 - [Regional compilation](https://huggingface.co/docs/diffusers/optimization/fp16#regional-compilation) is usually as efficient as full compilation (even if it was possible). Regional compilation compiles the transformer blocks, but not the entire pipeline.
 - whenever you want to branch depending on GPU data, that's a graph break for torch.compile. Either inefficient, or fails depending on `fullgraph`. Or, in less abstract terms:
 you want to set your attention mask to `None` if the entire attention mask is `True`. But the attention mask lives on GPU. Checking whether all values of a tensor on GPU are `True` requires a transfer back to CPU - that's always a graph break and cannot be compiled (efficiently).

Therefore, I'd suggest to
 - check for an all-True mask in the pipeline
 - pass a None-Mask to the transformer block in that case - but don't check any GPU data in the transformer block, so the transformer block can be compiled.

dxqb:
> Thanks!
> 
> Could you update with the following things?
> 
>     * Shed light into what caused the speed regression
> 

Here is a benchmark of the impact of using a mask unnecessarily (second graph): https://github.com/huggingface/diffusers/pull/12870#issue-3751633383
torch SDPA falls back to a non-flash algorithm if a mask is used.

kashif:
yes i also benchmarked the mask and got:

nsight_sdpa_mask_regression

kashif:
thanks @dxqb please check now

DefTruth:
I tested this https://github.com/huggingface/diffusers/pull/12987 on L20, and there was an 11% performance improvement in my test.

sayakpaul:
And do you find that to resolve your concerns?

DefTruth:
I think this has solved my problem. Based on this PR, the performance in my test cases has been working as expected.

dxqb:
> thanks @dxqb please check now

I don't have a simple test case, but code looks simple and good now

yiyixuxu:
thanks again @kashif !!

PR #12991, opened 2026-01-17
Fix QwenImageEditPlus max_sequence_length and transformer text seq len kwargs

tonera:
# What does this PR do?

This PR fixes two issues in QwenImageEditPlusPipeline (pipeline_qwenimage_edit_plus.py) related to text sequence length handling:

- Make max_sequence_length effective: the parameter was accepted by the public API but previously not applied. We now truncate prompt_embeds and prompt_embeds_mask to max_sequence_length.

- Prevent failures in the denoising loop: QwenImage transformer/RoPE needs an explicit text sequence length, but some accelerated/quantized transformer implementations (e.g. nunchaku) don’t accept the new max_txt_seq_len kwarg. The pipeline now adapts to the transformer forward signature:

- [ ] Prefer passing max_txt_seq_len when supported (or when **kwargs is accepted)
- [ ] Otherwise fall back to deprecated txt_seq_lens
- [ ] Otherwise pass nothing (let the implementation handle it)

- This avoids runtime errors such as:

- ValueError: Either max_txt_seq_len or txt_seq_lens ... must be provided
- TypeError: ... forward() got an unexpected keyword argument 'max_txt_seq_len'

- Who can review?

Tagging: @yiyixuxu @asomoza (pipelines)

### DIFF
--- a/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
+++ b/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit_plus.py
@@ -317,6 +317,20 @@ class QwenImageEditPlusPipeline(DiffusionPipeline, QwenImageLoraLoaderMixin):
         if prompt_embeds is None:
             prompt_embeds, prompt_embeds_mask = self._get_qwen_prompt_embeds(prompt, image, device)
 
+        # NOTE:
+        # - QwenImageTransformer2DModel forward (and its RoPE embedder) requires an explicit text seq len
+        #   (`max_txt_seq_len` / deprecated `txt_seq_lens`) for rotary embeddings.
+        # - Pipeline public API exposes `max_sequence_length`, but previous implementation did not apply it.
+        #   Here we enforce it by truncating prompt embeds and masks.
+        if max_sequence_length is not None:
+            try:
+                msl = int(max_sequence_length)
+            except Exception:
+                msl = None
+            if msl is not None and msl > 0:
+                prompt_embeds = prompt_embeds[:, :msl, :]
+                prompt_embeds_mask = prompt_embeds_mask[:, :msl]
+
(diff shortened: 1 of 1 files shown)


vladmandic:
gentle ping @yiyixuxu @sayakpaul @asomoza

yiyixuxu:
hi @tonera thanks for the PR
but first, can you share a reproducible code example(s) that demonstrates the issue? is this something specific to nunchaku?

vladmandic:
not sure why was this pr closed? its definitely an issue since not all code handles changes introduced in diffusers (yes, nunchaku is an example).
also, it cannot be fixed purely on nunchaku side since newer versions of nunchaku require newer torch/cuda which dropped support for rtx1 and rtx2 gpus - which means that older gpus require existing older nunchaku - which used to work correctly until diffusers change made it incompatible.

tonera:
> not sure why was this pr closed? its definitely an issue since not all code handles changes introduced in diffusers (yes, nunchaku is an example). also, it cannot be fixed purely on nunchaku side since newer versions of nunchaku require newer torch/cuda which dropped support for rtx1 and rtx2 gpus - which means that older gpus require existing older nunchaku - which used to work correctly until diffusers change made it incompatible.

The pull request was submitted when version 0.37 was still in beta, and the current version is 0.38. The maintenance team is processing it too slowly.

vladmandic:
@yiyixuxu @sayakpaul this is a real issue and we had a valid pr 3 months ago, now we lost it :(

sayakpaul:
I am also not sure why this was closed. I think @yiyixuxu's ask is valid. Cc: @dg845 as well.

vladmandic:
> I am also not sure why this was closed. I think @yiyixuxu's ask is valid. Cc: @dg845 as well.

ask is valid *if* it came in reasonable time. after 3 months, it makes contributors just give up.

PR #12994, opened 2026-01-19
feat: implement apply_lora_scale to remove boilerplate.

sayakpaul:
# What does this PR do?

Currently, we have this pattern throughout the modeling implementations:

```py
if joint_attention_kwargs is not None:
    joint_attention_kwargs = joint_attention_kwargs.copy()
    lora_scale = joint_attention_kwargs.pop("scale", 1.0)
else:
    lora_scale = 1.0

if USE_PEFT_BACKEND:
    # weight the lora layers by setting `lora_scale` for each PEFT layer
    scale_lora_layers(self, lora_scale)
else:
    if joint_attention_kwargs is not None and joint_attention_kwargs.get("scale", None) is not None:
        logger.warning(
            "Passing `scale` via `joint_attention_kwargs` when not using the PEFT backend is ineffective."
        )


...


if USE_PEFT_BACKEND:
    # remove `lora_scale` from each PEFT layer
    unscale_lora_layers(self, lora_scale)
```

IMO, this is not pretty and should possibly be minimized for a clean and self-contained implementation of the `forward()`. 

Hence, this PR introduces a decorator `apply_lora_scale` that can be used to decorate the forward method of a model supporting LoRA. I think this will help us reduce a bunch of boilerplate code.

For keeping the PR simple, I have only applied the decorator to `src/diffusers/models/transformers/transformer_flux.py` The LoRA tests for it are passing, indicating this direction might be a nice one.


sayakpaul:
@DN6 would love to know your thoughts!

src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -22,7 +22,7 @@
 
 from ...configuration_utils import ConfigMixin, register_to_config
 from ...loaders import FluxTransformer2DLoadersMixin, FromOriginalModelMixin, PeftAdapterMixin
-from ...utils import USE_PEFT_BACKEND, logging, scale_lora_layers, unscale_lora_layers
```
sayakpaul, line 25: This yields 21 LoC deletions. We have this pattern in about 32 files. So, this amounts for a 672 deletions. Not bad, IMO.

src/diffusers/models/transformers/transformer_flux.py
```diff
@@ -634,6 +634,7 @@ def __init__(
 
         self.gradient_checkpointing = False
 
+    @apply_lora_scale("joint_attention_kwargs")
```
DN6, line 637: Nice! 👍🏽 Design looks good to me.

sayakpaul:
@DN6 a gentle ping

src/diffusers/utils/peft_utils.py
```diff
@@ -275,6 +276,59 @@ def get_module_weight(weight_for_adapter, module_name):
...
+            attention_kwargs = kwargs.get(kwargs_name)
+
+            if attention_kwargs is not None:
+                attention_kwargs = attention_kwargs.copy()
+                kwargs[kwargs_name] = attention_kwargs
+                lora_scale = attention_kwargs.pop("scale", 1.0)
+            else:
+                if (
+                    not USE_PEFT_BACKEND
+                    and attention_kwargs is not None
```
DN6: This path won't ever be accessed since to reach here `attention_kwargs` has to be `None`.

I think it's supposed to be
```python
  if attention_kwargs is not None:
      attention_kwargs = attention_kwargs.copy()
      kwargs[kwargs_name] = attention_kwargs
      lora_scale = attention_kwargs.pop("scale", 1.0)

      if not USE_PEFT_BACKEND and lora_scale != 1.0:
          logger.warning(
              f"Passing `scale` via `{kwargs_name}` when not using the PEFT backend is ineffective."
          )

```

sayakpaul:
Merging as discussed with Dhruv.

Issue #12550, opened 2025-10-27
Fal Flashpack


vgabbo:
**Is your feature request related to a problem? Please describe.**
Safetensors loading seems it could be faster.

**Describe the solution you'd like.**
Recently fal has open sourced a new loading method for faster DIT and TE loading. 

**Describe alternatives you've considered.**
[A clear and concise description of any alternative solutions or features you've considered.
](https://github.com/fal-ai/flashpack)

**Additional context.**
I just wanted to ask if anyone will be looking into it. Seems it could enhance the offloading mechanism.

DN6:
Looking into it :)

devanshi00:
hi @sayakpaul @yiyixuxu 

i’m really looking forward to working on this integration!

since adding flashpack support will likely require major changes to files such as 
`modeling_utils.py`, `pipeline_utils.py`, and some model-specific loader files, 
i wanted to clarify a few things before proceeding:

which direction would you prefer to keep the design consistent with existing patterns?
- **option a:** introduce new methods like `from_pretrained_flashpack()` and `save_pretrained_flashpack()`
- **option b:** extend existing methods with a flag, e.g. `from_pretrained(..., use_flashpack=True)`

which model classes should be prioritized initially?
- dit models (e.g. sd3, flux, etc.)
- unet models  
- text encoders (e.g. clip, t5)

thanks! happy to start drafting a proposal or initial pr plan once we align on these points.

sayakpaul:
Thanks @devanshi00! Option b sounds good!

RakshithaKowlikar:
hey @sayakpaul @yiyixuxu @DN6 

i've been exploring this issue as well and have figured out the changes required in `modeling_utils.py` and `pipeline_utils.py` that align with option B, mimicking the safetensors implementation
this includes proper fallback mechanisms when Flashpack isn't available and dtype consistency checks in the Flashpack loading block before calling `assign_from_file()`

@devanshi00 would you be open to collaborating on this?

RakshithaKowlikar:
Hey @devanshi00! I've already drafted the preliminary implementation while exploring this. Would it work if I handle the implementation and you work on the testing? That way we both contribute to important aspects. Let me know if you'd prefer a different split!

devanshi00:
[...]

PR #12700, opened 2025-11-22
FlashPack

hlky:
# What does this PR do?

Adds support for FlashPack

FlashPack could be used as weight format only (see: `load_flashpack_checkpoint`) - keeping only weight format code would be a cleaner integration, the model loading is indeed faster[1] however part of the performance difference seems to be just due to the complexity of existing `from_pretrained` code, for example I noticed that `_caching_allocator_warmup` slows things down, also `empty_device_cache` is called in `_load_pretrained_model` whereas FlashPack's code doesn't thus the empty cache time is excluded from FlashPack's benchmark results.

- [x] Pipeline
- [x] Cleanup (`is_flashpack_available` check logging, etc)
- [x] FlashPack arguments (`silent`, `num_streams`, `use_distributed_loading` etc.)

<details><summary>Benchmark</summary>
<p>

Changes from FlashPack's version:
- Use Diffusers
- Remove `transformers` related code
- Add SD v1.5

```python
import csv
import gc
import os
import shutil
import tempfile
import time

import torch

from diffusers.models import AutoModel as DiffusersAutoModel
from huggingface_hub import snapshot_download


def test_model(
    repo_id: str,
    subfolder: str | None = None,
    accelerate_device: str | torch.device = "cuda",
    flashpack_device: str | torch.device = "cuda",
    dtype: torch.dtype | None = None,
    allow_pattern: str = "*",
) -> tuple[float, float, int]:
    """
    Test a model from a repository.
    """
    allow_patterns = [f"{subfolder}/{allow_pattern}"]
    if allow_pattern != "*":
        allow_patterns = [f"{subfolder}/{allow_pattern}", f"{subfolder}/config.json"]
    repo_dir = snapshot_download(
        repo_id, allow_patterns=None if subfolder is None else allow_patterns
    )
    model_dir = repo_dir if subfolder is None else os.path.join(repo_dir, subfolder)
    saved_flashpack_path = os.path.join(model_dir, "model.flashpack")
    saved_flashpack_config_path = os.path.join(model_dir, "flashpack_config.json")

    with tempfile.TemporaryDirectory() as tmpdir:
        # Make a new model directory with the model in it so it isn't cached
        temp_model_dir = os.path.join(tmpdir, "model")
        flashpack_dir = os.path.join(tmpdir, "flashpack")
        os.makedirs(flashpack_dir, exist_ok=True)
        print("Copying model to temporary directory")
        shutil.copytree(model_dir, temp_model_dir)

        # Load from the temporary model directory
[...]

Issue #12368, opened 2025-09-22
Add WanImage2ImagePipeline


vladmandic:
[Wan-AI/Wan2.2-T2V-A14B](https://huggingface.co/Wan-AI/Wan2.2-I2V-A14B) supported via `WanPipeline` is not only one of the best video models, its also widely popular as *text-to-image* model  
(basically using it to produce video with single frame) due to its very good photo-realistic output.

Ask is to extend support by adding `WanImage2ImagePipeline` to allow *image-to-image* workflows
Work should be limited to setting `timesteps` and modifying `prepare_latents` method to take input `image` and `strength` params to encode input image and add noise to it, same as in pretty much any other image-to-image pipeline.

Note: this request is different than input image support in `WanImageToVideoPipeline` and `WanVACEPipeline` since
- a) those pipelines work with different model weights, so user cannot switch from text-to-image to image-to-image without full model reload
- b) they do not support denoising strength, image is taken as-is and if used in image-to-image workflow it will produce output identical to input.

Bonus: Also create `WanInpaintPipeline` to allow for image/mask combinations as input.

cc @yiyixuxu @sayakpaul @a-r-r-o-w @DN6

devanshi00:
Hello I m interested in contributing to this issue.
To do this, I plan to add a  **WanImage2ImagePipeline** that closely mirrors standard diffusion *image-to-image* workflows while making only minimal, targeted changes:

### Reuse the existing WanPipeline architecture and weights
The new pipeline will share the same tokenizer, text encoder, VAE, transformer(s), and scheduler as `WanPipeline`. This ensures users can switch seamlessly between text-to-image and image-to-image workflows without reloading or swapping model checkpoints.

### Introduce denoising-strength–based timestep control
Image-to-image behavior will be enabled by computing the starting timestep from a user-provided `strength` parameter and running the denoising process only over the remaining timesteps. This follows the behavior of established img2img pipelines in *diffusers* and allows controlled deviation from the input image.

### Extend `prepare_latents` to support image initialization
When an input image is provided:
- The image will be encoded into latent space using the existing VAE  
- Noise will be added to the encoded latents at the selected starting timestep  
- These noisy latents will be used as the initial state for denoising  

If no image is provided, the pipeline will fall back to the existing random latent initialization, preserving current text-to-image behavior.

### Single-frame output 
[...]

Issue #12235, opened 2025-08-25
Support fun control variants for Wan


sayakpaul:
It would be cool to support the control variants of Wan: https://huggingface.co/collections/alibaba-pai/wan22-fun-68958eabec343b948f1225c5.

https://huggingface.co/alibaba-pai/Wan2.2-Fun-A14B-Control has some very nice examples.

Checking with @a-r-r-o-w if this is already in the works.

satani99:
i'd like to work on this if no one's working.

sayakpaul:
All yours. Go ahead.

PR #12999, opened 2026-01-19
[feat] added fal-flashpack support

devanshi00:
# What does this PR do?
This PR adds **FlashPack support** to Diffusers, enabling **significantly faster model loading and improved inference throughput** through an optimized on-disk weight format.

## What is FlashPack?

**FlashPack** is a weight-packing format that serializes model parameters into a contiguous, GPU-friendly layout.  
It reduces:

- Disk I/O overhead during model loading
- CPU overhead from fragmented tensor deserialization
- Runtime memory indirection during inference

As a result, FlashPack provides:
- Faster pipeline initialization
- Faster end-to-end inference
- Zero changes to sampling logic or model architecture

FlashPack is particularly beneficial for **large transformer-based diffusion models**, including video and multi-modal pipelines.

## Architecture & Design

FlashPack is implemented as a **storage-level optimization**, not a runtime execution hook.

### Design principles

- **Non-intrusive**: No changes to model forward passes
- **Opt-in**: Enabled explicitly via `use_flashpack=True`
- **One-time conversion**: Packed weights are generated once and reused
- **Graceful fallback**: Automatically falls back to standard weights if FlashPack artifacts are missing

### Workflow

1. Load a pipeline using standard Diffusers APIs  
2. Convert weights using `save_pretrained(..., use_flashpack=True)`  
3. Reload the pipeline with `use_flashpack=True`  
4. FlashPack weights are used automatically if present  

Note : To push a FlashPack model to the Hugging Face Hub, users must first save with `use_flashpack=True` and then manually upload the resulting directory, as `push_to_hub=True` currently only supports standard PyTorch/safetensors weights.

### Model Load Time Comparison

| Model | Size (GB) | Standard Load (s) | FlashPack Load (s) | Speedup |
|------|-----------|-------------------|--------------------|---------|
| Wan2.1 1.3B DiT | 2.64 | 1.181 | 0.109 | **10.81×** |
| FLUX.1 [dev] 12B DiT | 22.17 | 3.738 | 0.092 | **40.64×** |
| Wan2.1 14B DiT | 26.61 | 4.105 | 0.125 | **32.84×** |
| Qwen-Image-Edit 20B DiT | 38.05 | 5.988 | 0.166 | **36.09×** |

### Effective Weight Loading Throughput & FlashPack Conversion Cost

| Model | Standard Load (GB/s) | FlashPack Load (GB/s) | FlashPack Conversion Time (s) |
|------|----------------------|------------------------|-------------------------------|
| Wan2.1 1.3B DiT | 2.43 | **27.99** | 8.21 |
| FLUX.1 [dev] 12B DiT | 5.94 | **241.07** | 77.44 |
| Wan2.1 14B DiT | 6.52 | **213.27** | 81.85 |
| Qwen-Image-Edit 20B DiT | 6.36 | **229.78** | 129.59 |
benchmark_comparison

benchmark_comparison_final_2

Note: All timings use direct GPU loading (device_map="cuda" or equivalent) to ensure a fair comparison between standard loading and FlashPack.

### Benchmark Setup

- **Hardware**: NVIDIA A100  
- **Precision**: `bfloat16`  
- **Framework**: Diffusers with FlashPack enabled  
- **Measurement**: End-to-end model load time (disk → GPU)  
- **Reproducibility**: Same environment and configuration across runs  

### One-time FlashPack conversion

```python
import torch
from diffusers import WanPipeline

pipe = WanPipeline.from_pretrained(
    "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
    torch_dtype=torch.bfloat16,
)
pipe.to("cuda")

# One-time conversion to FlashPack format
pipe.save_pretrained(
    "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
    use_flashpack=True,
)
```
This produces the FlashPack-packed weights alongside the regular weights in the repo.

### Loading with FlashPack
```python
pipe = WanPipeline.from_pretrained(
    "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
    torch_dtype=torch.bfloat16,
    use_flashpack=True,  # Enable FlashPack loading
)
pipe.to("cuda")

video = pipe(
    prompt="A beautiful sunset over a calm ocean",
    width=832,
    height=480,
    num_inference_steps=16,
).videos[0]
```
FlashPack load/inference falls back gracefully to standard weights if FlashPack output does not exist.

### Inference-Time Results (FlashPack)
| Model                         | Standard Inference (s) | FlashPack Inference (s) |
|-------------------------------|-----------------------|-------------------------|
| Wan2.1-T2V-1.3B-Diffusers     | 65.527                | 64.326                  |

FlashPack optimizes *model loading*, not transformer compute. Inference speedups are expected to be minimal.

FIxes #(issue)
#12550 

## Before submitting
- [] Did you write any new necessary tests?

## Who can review?
@sayakpaul @yiyixuxu @DN6

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -25,7 +25,7 @@ jobs:
     if: github.event_name == 'pull_request'
     steps:
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
 
       - name: Check out code
         uses: actions/checkout@v6
@@ -101,14 +101,14 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
       - name: Login to Docker Hub
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         with:
           username: ${{ env.REGISTRY }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Build and push
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v6
         with:
           no-cache: true
           context: ./docker/${{ matrix.image-name }}
--- a/.github/workflows/pr_modular_tests.yml
+++ b/.github/workflows/pr_modular_tests.yml
@@ -75,9 +75,27 @@ jobs:
         if: ${{ failure() }}
         run: |
           echo "Repo consistency check failed. Please ensure the right dependency versions are installed with 'pip install -e .[quality]' and run 'make fix-copies'" >> $GITHUB_STEP_SUMMARY
+  check_auto_docs:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v6
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: |
+          pip install --upgrade pip
+          pip install .[quality]
+      - name: Check auto docs
+        run: make modular-autodoctrings
+      - name: Check if failure
+        if: ${{ failure() }}
+        run: |
+          echo "Auto docstring checks failed. Please run `python utils/modular_auto_docstring.py --fix_and_overwrite`." >> $GITHUB_STEP_SUMMARY
 
   run_fast_tests:
-    needs: [check_code_quality, check_repository_consistency]
+    needs: [check_code_quality, check_repository_consistency, check_auto_docs]
     name: Fast PyTorch Modular Pipeline CPU tests
 
     runs-on:
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -11,4 +11,4 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: typos-action
-        uses: crate-ci/typos@v1.12.4
+        uses: crate-ci/typos@v1.42.1
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,10 @@ fix-copies:
 	python utils/check_copies.py --fix_and_overwrite
 	python utils/check_dummies.py --fix_and_overwrite
 
+# Auto docstrings in modular blocks
+modular-autodoctrings:
+	python utils/modular_auto_docstring.py
+
 # Run tests for the library
 
 test:
--- a/docker/diffusers-pytorch-cuda/Dockerfile
+++ b/docker/diffusers-pytorch-cuda/Dockerfile
@@ -1,8 +1,8 @@
-FROM nvidia/cuda:12.1.0-runtime-ubuntu20.04
+FROM nvidia/cuda:12.9.1-runtime-ubuntu24.04
 LABEL maintainer="Hugging Face"
 LABEL repository="diffusers"
 
-ARG PYTHON_VERSION=3.12
+ARG PYTHON_VERSION=3.10
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get -y update \
@@ -32,10 +32,17 @@ RUN uv venv --python ${PYTHON_VERSION} --seed ${VIRTUAL_ENV}
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 # pre-install the heavy dependencies (these can later be overridden by the deps from setup.py)
+# Install torch, torchvision, and torchaudio together to ensure compatibility
 RUN uv pip install --no-cache-dir \
     torch \
     torchvision \
-    torchaudio
+    torchaudio \
+    --index-url https://download.pytorch.org/whl/cu129
+
+# Install compatible versions of numba/llvmlite for Python 3.10+
+RUN uv pip install --no-cache-dir \
+    "llvmlite>=0.40.0" \
+    "numba>=0.57.0"
 
 RUN uv pip install --no-cache-dir "git+https://github.com/huggingface/diffusers.git@main#egg=diffusers[test]"
 
--- a/docker/diffusers-pytorch-xformers-cuda/Dockerfile
+++ b/docker/diffusers-pytorch-xformers-cuda/Dockerfile
@@ -1,8 +1,8 @@
-FROM nvidia/cuda:12.1.0-runtime-ubuntu20.04
+FROM nvidia/cuda:12.9.1-runtime-ubuntu24.04
 LABEL maintainer="Hugging Face"
 LABEL repository="diffusers"
 
-ARG PYTHON_VERSION=3.12
+ARG PYTHON_VERSION=3.10
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get -y update \
@@ -32,10 +32,17 @@ RUN uv venv --python ${PYTHON_VERSION} --seed ${VIRTUAL_ENV}
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 # pre-install the heavy dependencies (these can later be overridden by the deps from setup.py)
+# Install torch, torchvision, and torchaudio together to ensure compatibility
 RUN uv pip install --no-cache-dir \
     torch \
     torchvision \
-    torchaudio
+    torchaudio \
+    --index-url https://download.pytorch.org/whl/cu129
+
+# Install compatible versions of numba/llvmlite for Python 3.10+
+RUN uv pip install --no-cache-dir \
+    "llvmlite>=0.40.0" \
+    "numba>=0.57.0"
 
 RUN uv pip install --no-cache-dir "git+https://github.com/huggingface/diffusers.git@main#egg=diffusers[test]"
 
--- a/docs/source/en/api/pipelines/ltx2.md
+++ b/docs/source/en/api/pipelines/ltx2.md
@@ -24,6 +24,179 @@ You can find all the original LTX-Video checkpoints under the [Lightricks](https
 
 The original codebase for LTX-2 can be found [here](https://github.com/Lightricks/LTX-2).
 
+## Two-stages Generation
+Recommended pipeline to achieve production quality generation, this pipeline is composed of two stages:
+
+- Stage 1: Generate a video at the target resolution using diffusion sampling with classifier-free guidance (CFG). This stage produces a coherent low-noise video sequence that respects the text/image conditioning.
+- Stage 2: Upsample the Stage 1 output by 2 and refine details using a distilled LoRA model to improve fidelity and visual quality. Stage 2 may apply lighter CFG to preserve the structure from Stage 1 while enhancing texture and sharpness.
+
+Sample usage of text-to-video two stages pipeline
+
+```py
+import torch
+from diffusers import FlowMatchEulerDiscreteScheduler
+from diffusers.pipelines.ltx2 import LTX2Pipeline, LTX2LatentUpsamplePipeline
+from diffusers.pipelines.ltx2.latent_upsampler import LTX2LatentUpsamplerModel
+from diffusers.pipelines.ltx2.utils import STAGE_2_DISTILLED_SIGMA_VALUES
+from diffusers.pipelines.ltx2.export_utils import encode_video
+
+device = "cuda:0"
+width = 768
+height = 512
+
+pipe = LTX2Pipeline.from_pretrained(
+    "Lightricks/LTX-2", torch_dtype=torch.bfloat16
+)
+pipe.enable_sequential_cpu_offload(device=device)
+
+prompt = "A beautiful sunset over the ocean"
+negative_prompt = "shaky, glitchy, low quality, worst quality, deformed, distorted, disfigured, motion smear, motion artifacts, fused fingers, bad anatomy, weird hand, ugly, transition, static."
+
+# Stage 1 default (non-distilled) inference
+frame_rate = 24.0
+video_latent, audio_latent = pipe(
+    prompt=prompt,
+    negative_prompt=negative_prompt,
+    width=width,
+    height=height,
+    num_frames=121,
+    frame_rate=frame_rate,
+    num_inference_steps=40,
+    sigmas=None,
+    guidance_scale=4.0,
+    output_type="latent",
+    return_dict=False,
+)
+
+latent_upsampler = LTX2LatentUpsamplerModel.from_pretrained(
+    "Lightricks/LTX-2",
+    subfolder="latent_upsampler",
+    torch_dtype=torch.bfloat16,
+)
+upsample_pipe = LTX2LatentUpsamplePipeline(vae=pipe.vae, latent_upsampler=latent_upsampler)
+upsample_pipe.enable_model_cpu_offload(device=device)
+upscaled_video_latent = upsample_pipe(
+    latents=video_latent,
+    output_type="latent",
+    return_dict=False,
+)[0]
+
+# Load Stage 2 distilled LoRA
+pipe.load_lora_weights(
+    "Lightricks/LTX-2", adapter_name="stage_2_distilled", weight_name="ltx-2-19b-distilled-lora-384.safetensors"
+)
+pipe.set_adapters("stage_2_distilled", 1.0)
+# VAE tiling is usually necessary to avoid OOM error when VAE decoding
+pipe.vae.enable_tiling()
+# Change scheduler to use Stage 2 distilled sigmas as is
+new_scheduler = FlowMatchEulerDiscreteScheduler.from_config(
+    pipe.scheduler.config, use_dynamic_shifting=False, shift_terminal=None
+)
+pipe.scheduler = new_scheduler
+# Stage 2 inference with distilled LoRA and sigmas
+video, audio = pipe(
+    latents=upscaled_video_latent,
+    audio_latents=audio_latent,
+    prompt=prompt,
+    negative_prompt=negative_prompt,
+    num_inference_steps=3,
+    noise_scale=STAGE_2_DISTILLED_SIGMA_VALUES[0], # renoise with first sigma value https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-pipelines/src/ltx_pipelines/ti2vid_two_stages.py#L218
+    sigmas=STAGE_2_DISTILLED_SIGMA_VALUES,
+    guidance_scale=1.0,
+    output_type="np",
+    return_dict=False,
+)
+video = (video * 255).round().astype("uint8")
+video = torch.from_numpy(video)
+
+encode_video(
+    video[0],
+    fps=frame_rate,
+    audio=audio[0].float().cpu(),
+    audio_sample_rate=pipe.vocoder.config.output_sampling_rate,
+    output_path="ltx2_lora_distilled_sample.mp4",
+)
+```
+
+## Distilled checkpoint generation
+Fastest two-stages generation pipeline using a distilled checkpoint.
+
+```py
+import torch
+from diffusers.pipelines.ltx2 import LTX2Pipeline, LTX2LatentUpsamplePipeline
+from diffusers.pipelines.ltx2.latent_upsampler import LTX2LatentUpsamplerModel
+from diffusers.pipelines.ltx2.utils import DISTILLED_SIGMA_VALUES, STAGE_2_DISTILLED_SIGMA_VALUES
+from diffusers.pipelines.ltx2.export_utils import encode_video
+
+device = "cuda"
+width = 768
+height = 512
+random_seed = 42
+generator = torch.Generator(device).manual_seed(random_seed)
+model_path = "rootonchair/LTX-2-19b-distilled"
+
+pipe = LTX2Pipeline.from_pretrained(
+    model_path, torch_dtype=torch.bfloat16
+)
+pipe.enable_sequential_cpu_offload(device=device)
+
+prompt = "A beautiful sunset over the ocean"
+negative_prompt = "shaky, glitchy, low quality, worst quality, deformed, distorted, disfigured, motion smear, motion artifacts, fused fingers, bad anatomy, weird hand, ugly, transition, static."
+
+frame_rate = 24.0
+video_latent, audio_latent = pipe(
+    prompt=prompt,
+    negative_prompt=negative_prompt,
+    width=width,
+    height=height,
+    num_frames=121,
+    frame_rate=frame_rate,
+    num_inference_steps=8,
+    sigmas=DISTILLED_SIGMA_VALUES,
+    guidance_scale=1.0,
+    generator=generator,
+    output_type="latent",
+    return_dict=False,
+)
+
+latent_upsampler = LTX2LatentUpsamplerModel.from_pretrained(
+    model_path,
+    subfolder="latent_upsampler",
+    torch_dtype=torch.bfloat16,
+)
+upsample_pipe = LTX2LatentUpsamplePipeline(vae=pipe.vae, latent_upsampler=latent_upsampler)
+upsample_pipe.enable_model_cpu_offload(device=device)
+upscaled_video_latent = upsample_pipe(
+    latents=video_latent,
+    output_type="latent",
+    return_dict=False,
+)[0]
+
+video, audio = pipe(
+    latents=upscaled_video_latent,
+    audio_latents=audio_latent,
+    prompt=prompt,
+    negative_prompt=negative_prompt,
+    num_inference_steps=3,
+    noise_scale=STAGE_2_DISTILLED_SIGMA_VALUES[0], # renoise with first sigma value https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-pipelines/src/ltx_pipelines/distilled.py#L178
+    sigmas=STAGE_2_DISTILLED_SIGMA_VALUES,
+    generator=generator,
+    guidance_scale=1.0,
+    output_type="np",
+    return_dict=False,
+)
+video = (video * 255).round().astype("uint8")
+video = torch.from_numpy(video)
+
+encode_video(
+    video[0],
+    fps=frame_rate,
+    audio=audio[0].float().cpu(),
+    audio_sample_rate=pipe.vocoder.config.output_sampling_rate,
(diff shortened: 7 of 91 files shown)


sayakpaul:
Thanks for the comprehensive PR. To benchmark model loading, are we reporting just the pipeline loading time or just the denoiser?

Could we benchmark just the denoiser loading? Also, during the standard load, let's specify `device_map="cuda"` that should provide faster loading.

src/diffusers/models/modeling_utils.py
```diff
@@ -707,6 +708,9 @@ def save_pretrained(
                 Whether or not to push your model to the Hugging Face Hub after saving it. You can specify the
                 repository you want to push to with `repo_id` (will default to the name of `save_directory` in your
                 namespace).
+            use_flashpack (`bool`, *optional*, defaults to `False`): 
+                Whether to save the model in FlashPack format. FlashPack is a binary format that allows for faster
+                loading. Requires the `flashpack` library to be installed.
```
sayakpaul: Let's also hyperlink this to the official flashpack documentation.

src/diffusers/models/modeling_utils.py
```diff
@@ -734,6 +738,44 @@ def save_pretrained(
...
         os.makedirs(save_directory, exist_ok=True)
+        if use_flashpack:
+            if not is_main_process:
+                return
+
+            try:
+                from flashpack import pack_to_file
+                import json  # Ensure json is imported
+            except ImportError:
+                raise ImportError("The `use_flashpack=True` argument requires the `flashpack` library.")
```
sayakpaul: Let's wrap this under a `is_flashpack_available()` utility like this:
https://github.com/huggingface/diffusers/blob/d7a1c31f4f85bae5a9e01cdce49bd7346bd8ccd6/src/diffusers/utils/import_utils.py#L288

src/diffusers/models/modeling_utils.py
```diff
@@ -734,6 +738,44 @@ def save_pretrained(
...
+                if hasattr(self, "config"):
+                    try:
+                        # Attempt to get dictionary representation
+                        if hasattr(self.config, "to_dict"):
+                            config_data = self.config.to_dict()
+                        else:
+                            config_data = dict(self.config)
+
+                        with open(config_save_path, "w") as f:
+                            json.dump(config_data, f, indent=4)
```
sayakpaul: Let's wrap this under a utility?

src/diffusers/models/modeling_utils.py
```diff
@@ -1200,6 +1248,69 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
             model = load_flax_checkpoint_in_pytorch_model(model, resolved_model_file)
         else:
+            # If we are using `use_flashpack`, we try to load the model from flashpack first.
+            # If flashpack is not available or the file cannot be loaded, we fall back to
+            # the standard loading path (e.g. safetensors or PyTorch).
+            if use_flashpack:
+                try:
+                    from flashpack import assign_from_file
+                except ImportError:
+                    pass
```
sayakpaul: Same as above.

Also, when `use_flashpack` is provided and flashpack ckpt isn't available what's the overhead of running the conversion from safetensors/bin -> flashpack and then loading?

src/diffusers/models/modeling_utils.py
```diff
@@ -1200,6 +1248,69 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
+                            model.eval()
+
+                            if output_loading_info:
+                                loading_info = {
+                                    "missing_keys": [],
+                                    "unexpected_keys": [],
+                                    "mismatched_keys": [],
+                                    "error_msgs": [],
+                                }
+                                return model, loading_info
```
sayakpaul: We shouldn't need this much code, IMO. Code to obtain `flashpack_file` and `assign_from_file(model, flashpack_file)` should be enough apart from any other validation related stuff. Am I mistaken to think that?

src/diffusers/models/modeling_utils.py
devanshi00: Added a new file named `flashpack_utils.py` in utils folder. You can check it.

src/diffusers/models/modeling_utils.py
devanshi00: Done

src/diffusers/models/modeling_utils.py
devanshi00: Refactored this code and removed the redundant part. You can check it in `flashpack_utils.py`

devanshi00:
> Thanks for the comprehensive PR. To benchmark model loading, are we reporting just the pipeline loading time or just the denoiser?
> 
> Could we benchmark just the denoiser loading? Also, during the standard load, let's specify `device_map="cuda"` that should provide faster loading.

For benchmarking model loading, I used the official benchmarking script from the orignal flashpack repository, which measures only the diffusion transformer (DiT / denoiser) loading time, not the full pipeline. The results reported in the table above therefore correspond exclusively to denoiser loading for the respective Diffusers models.

Here is the script
```python 
import csv
import gc
import os
import shutil
import tempfile
import time
import torch

from huggingface_hub import snapshot_download
from diffusers.models import AutoModel as DiffusersAutoModel
def test_model(
    repo_id: str,
    subfolder: str | None = None,
    accelerate_device: str | torch.device = "cuda",
    flashpack_device: str | torch.device = "cuda",
    dtype: torch.dtype | None = None,
    use_transformers: bool = False,
) -> tuple[float, float, int]:
    """
    Test a model from a repository.
    """
    repo_dir = snapshot_download(
        repo_id, allow_patterns=None if subfolder is None else [f"{subfolder}/*"]
    )
    model_dir = repo_dir if subfolder is None else os.path.join(repo_dir, subfolder)
    saved_flashpack_path = os.path.join(model_dir, "model.flashpack")
    saved_flashpack_config_path = os.path.join(model_dir, "flashpack_config.json")
    
        # Make a new model directory with the model in it so it isn't cached
        temp_model_dir = os.path.join(tmpdir, "model")
        flashpack_dir = os.path.join(tmpdir, "flashpack")
        os.makedirs(flashpack_dir, exist_ok=True)
        print("Copying model to temporary directory")
        shutil.copytree(model_dir, temp_model_dir)
        # Load from the temporary model directory
        print("Loading model from temporary directory using from_pretrained")
        start_time = time.time()
        model = DiffusersAutoModel.from_pretrained(
            temp_model_dir,
            torch_dtype=dtype,
            device_map={"": accelerate_device},
        )

        end_time = time.time()
        accelerate_time = end_time - start_time
        print(f"Time taken with from_pretrained: {accelerate_time} seconds")

        if os.path.exists(saved_flashpack_path) and os.path.exists(
            saved_flashpack_config_path
        ):
            print("Copying flashpack to temporary directory")
            shutil.copy(
                saved_flashpack_path, os.path.join(flashpack_dir, "model.flashpack")
            )
            shutil.copy(
                saved_flashpack_config_path, os.path.join(flashpack_dir, "config.json")
            )
        else:
            print("Packing model to flashpack")
            pack_start_time = time.time()
            model.save_pretrained(
                flashpack_dir, target_dtype=dtype, use_flashpack=True
            )
            pack_end_time = time.time()
            print(
                f"Time taken for flashpack packing: {pack_end_time - pack_start_time} seconds"
            )
            # Copy back to the original model directory
            shutil.copy(
                os.path.join(flashpack_dir, "model.flashpack"), saved_flashpack_path
            )
            shutil.copy(
                os.path.join(flashpack_dir, "config.json"), saved_flashpack_config_path
            )

        del model
        sync_and_flush()

        print("Loading model from flashpack directory using from_pretrained_flashpack")
        flashpack_start_time = time.time()

        flashpack_model = DiffusersAutoModel.from_pretrained(
                flashpack_dir, device=flashpack_device, target_dtype=dtype,use_flashpack=True
            )

        flashpack_end_time = time.time()
        flashpack_time = flashpack_end_time - flashpack_start_time
        print(f"Time taken with flashpack loading: {flashpack_time} seconds")

        total_numel = 0
        for param in flashpack_model.parameters():
            total_numel += param.numel()

        total_bytes = total_numel * dtype.itemsize

        del flashpack_model
        sync_and_flush()

        return accelerate_time, flashpack_time, total_bytes


def test_wan_small_transformer() -> tuple[float, float, int]:
    return test_model(
        repo_id="Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
        subfolder="transformer",
        accelerate_device="cuda:0" if torch.cuda.is_available() else "cpu",
        flashpack_device="cuda:1" if torch.cuda.is_available() else "cpu",
        dtype=torch.bfloat16,
    )


def test_wan_large_transformer() -> tuple[float, float, int]:
    return test_model(
        repo_id="Wan-AI/Wan2.1-T2V-14B-Diffusers",
        subfolder="transformer",
        accelerate_device="cuda:0" if torch.cuda.is_available() else "cpu",
        flashpack_device="cuda:1" if torch.cuda.is_available() else "cpu",
        dtype=torch.bfloat16,
    )


def test_flux_transformer() -> tuple[float, float, int]:
    return test_model(
        repo_id="black-forest-labs/FLUX.1-dev",
        subfolder="transformer",
        accelerate_device="cuda:0" if torch.cuda.is_available() else "cpu",
        flashpack_device="cuda:1" if torch.cuda.is_available() else "cpu",
        dtype=torch.bfloat16,
    )


def test_qwen_transformer() -> tuple[float, float, int]:
    return test_model(
        repo_id="Qwen/Qwen-Image-Edit",
        subfolder="transformer",
        accelerate_device="cuda:0" if torch.cuda.is_available() else "cpu",
        flashpack_device="cuda:1" if torch.cuda.is_available() else "cpu",
        dtype=torch.bfloat16,
        use_transformers=True,
    )


def print_test_result(
    model_name: str,
    accelerate_time: float,
    flashpack_time: float,
    total_bytes: int,
) -> None:
    print(f"{model_name}: Accelerate time: {accelerate_time} seconds")
    print(f"{model_name}: Flashpack time: {flashpack_time} seconds")
    accelerate_gbps = (total_bytes / 1000**3) / accelerate_time
    flashpack_gbps = (total_bytes / 1000**3) / flashpack_time
    print(f"{model_name}: Accelerate GB/s: {accelerate_gbps} GB/s")
    print(f"{model_name}: Flashpack GB/s: {flashpack_gbps} GB/s")


def sync_and_flush() -> None:
    torch.cuda.empty_cache()
    gc.collect()
    os.system("sync")
    if os.geteuid() == 0:
        os.system("echo 3 | tee /proc/sys/vm/drop_caches")


if __name__ == "__main__":
    with open("benchmark_results_finall.csv", "a", newline="") as f:
        writer = csv.writer(f)
        writer.writerow(
            ["model", "accelerate_time", "flashpack_time", "total_overhead_time", "total_bytes"]
        )

        for i in range(10):
            for test_model_name, test_func in [
                ("Wan-AI/Wan2.1-T2V-1.3B-Diffusers", test_wan_small_transformer),
                ("Wan-AI/Wan2.1-T2V-14B-Diffusers", test_wan_large_transformer),
                ("black-forest-labs/FLUX.1-dev", test_flux_transformer),
                ("Qwen/Qwen-Image-Edit", test_qwen_transformer),
            ]:
                accelerate_time, flashpack_time, total_bytes = test_func()
                # ---- main benchmark CSV (unchanged, written every run) ----
                writer.writerow(
                    [
                        test_model_name,
                        accelerate_time,
                        flashpack_time,
                        total_bytes,
                    ]
                )

                print_test_result(
                    test_model_name,
                    accelerate_time,
                    flashpack_time,
                    total_bytes,
                )
```

Regarding FlashPack overhead, the total overhead time is the sum of:
FlashPack conversion time (one-time cost), and
FlashPack load time (per load)
These values are approximately as follows, based on the measurements reported earlier in the PR:
| Model                     | Total FlashPack Overhead Time (s) |
|---------------------------|-----------------------------------|
| Wan2.1 1.3B DiT           | 8.319                             |
| FLUX.1 [dev] 12B DiT      | 77.532                            |
| Wan2.1 14B DiT            | 81.975                            |
| Qwen-Image-Edit 20B DiT   | 129.756                           |

Note: All timings use direct GPU loading (device_map="cuda" or equivalent) to ensure a fair comparison between standard loading and FlashPack.

sayakpaul:
Thanks, so it should be `device_map="cuda".

> These values are approximately as follows, based on the measurements reported earlier in the PR:

The table doesn't make it clear the time with `from_pretrained()` when `device_map="cuda"` is used.

src/diffusers/models/modeling_utils.py
```diff
@@ -734,7 +739,15 @@ def save_pretrained(
         )
 
         os.makedirs(save_directory, exist_ok=True)
-
+        if use_flashpack:
+            if not is_main_process:
+                return
```
sayakpaul, line 744: This should be something like so: 

```py
if use_flashpack: 
    if not is_flashpack_available(): 
         raise 
    ...
```

Applies elsewhere as well.

src/diffusers/models/modeling_utils.py
```diff
@@ -1200,6 +1218,47 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
 
             model = load_flax_checkpoint_in_pytorch_model(model, resolved_model_file)
         else:
+            # If we are using `use_flashpack`, we try to load the model from flashpack first.
+            # If flashpack is not available or the file cannot be loaded, we fall back to
+            # the standard loading path (e.g. safetensors or PyTorch).
+            if use_flashpack:
```
sayakpaul, line 1223: Remember to add a check for `is_flashpack_available()`.

src/diffusers/models/modeling_utils.py
```diff
@@ -1200,6 +1218,47 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
+                        token=token,
+                        revision=revision,
+                        subfolder=subfolder,
+                        user_agent=user_agent,
+                        commit_hash=commit_hash,
+                    )
+                except EnvironmentError:
+                    resolved_model_file = None
+                with no_init_weights():
+                    model = cls.from_config(config, **unused_kwargs)
```
sayakpaul: Why do we need to initialize the model here? It is already done here:
https://github.com/huggingface/diffusers/blob/ec376293714f269947f6d9d8a572bd73040bc1a0/src/diffusers/models/modeling_utils.py#L1278

IMO, the flashpack weight name resolving related code can be around the following block:

https://github.com/huggingface/diffusers/blob/ec376293714f269947f6d9d8a572bd73040bc1a0/src/diffusers/models/modeling_utils.py#L1242-L1256

And then the rest of the code can follow.

sayakpaul (commented):
Left some further comments.

sayakpaul:
Also, WDYT of running a conversion on the go from bin / safetensors -> flashpack when a user requests to load in flashpack for a non-flashpack checkpoint?

src/diffusers/models/modeling_utils.py
devanshi00: I have added the check in `flashpack_utils.py` . Is it okay?

sayakpaul: Sure that works too.

devanshi00:
> Thanks, so it should be `device_map="cuda".
> 
> > These values are approximately as follows, based on the measurements reported earlier in the PR:
> 
> The table doesn't make it clear the time with `from_pretrained()` when `device_map="cuda"` is used.

I hope it is clear now.

devanshi00:
> Also, WDYT of running a conversion on the go from bin / safetensors -> flashpack when a user requests to load in flashpack for a non-flashpack checkpoint?

I don’t think automatic conversion should be enabled by default. FlashPack conversion can take a long time for large models, so triggering it automatically when `use_flashpack=True` would result in unexpected delays in first run. It may also silently create files on disk or modify the local cache, which can be confusing for users. Keeping conversion explicit avoids these surprises and makes behavior easier to understand.
The current behavior should remain the default, with no automatic conversion. If needed, this can be offered as an explicit opt-in (for example, `flashpack_auto_convert=True`) or via a separate helper utility to generate FlashPack files ahead of time. This keeps loading predictable while still giving advanced users more control.

sayakpaul:
> I don’t think automatic conversion should be enabled by default. FlashPack conversion can take a long time for large models, so triggering it automatically when `use_flashpack=True` would result in unexpected delays in first run. It may also silently create files on disk or modify the local cache, which can be confusing for users. Keeping conversion explicit avoids these surprises and makes behavior easier to understand.
> The current behavior should remain the default, with no automatic conversion. If needed, this can be offered as an explicit opt-in (for example, `flashpack_auto_convert=True`) or via a separate helper utility to generate FlashPack files ahead of time. This keeps loading predictable while still giving advanced users more control.

This makes sense. We can make all of this explicitly documented. Thanks for the context!

> I hope it is clear now.

It is not. I would have expected to see a table with the following columns:

* Model checkpoint id
* Flashpack loading time
* `from_pretrained(device_map="cuda")` timing

LMK if this is still unclear.

devanshi00:
> It is not. I would have expected to see a table with the following columns:
> 
> * Model checkpoint id
> * Flashpack loading time
> * `from_pretrained(device_map="cuda")` timing
> 
> LMK if this is still unclear.

So according to you, the benchmarking script is okay ? I use 
```python
 model = DiffusersAutoModel.from_pretrained(
            temp_model_dir,
            torch_dtype=dtype,
            device_map={"": accelerate_device},
        )
   ```
For the standard loading time . I just need to change the csv header right?


> This makes sense. We can make all of this explicitly documented. Thanks for the context!

So do we need an explicit opt-in (for example, `flashpack_auto_convert=True`) or not?

**Comment by sayakpaul** [comment]
> So do we need an explicit opt-in (for example, flashpack_auto_convert=True) or not?

I think we can skip that for now and advise that users run the conversion with `use_flashpack=True` during `save_pretrained()`. 

> For the standard loading time . I just need to change the csv header right?

I only see two columns in the table provided in https://github.com/huggingface/diffusers/pull/12999#issuecomment-3775425192. It has the flashpack timing but not the `from_pretrained()` with device_map timing unless I am missing out on something obvious.

**Comment by devanshi00** [comment]
> I only see two columns in the table provided in [#12999 (comment)](https://github.com/huggingface/diffusers/pull/12999#issuecomment-3775425192). It has the flashpack timing but not the `from_pretrained()` with device_map timing unless I am missing out on something obvious.

Sorry for the confusion. But I m refering to the tables in the first comment where Standard load is equivalent to from_pretrained() with device_map cuda. I had uploaded those results with the same benchmarking script I mentioned afterwards. 
https://github.com/huggingface/diffusers/pull/12999#issue-3829100536

File: src/diffusers/models/modeling_utils.py (hunk 7 above)
Comment (devanshi00) [reply src/diffusers/models/modeling_utils.py:1223]: Already added in `flashpack_utils.py`

File: src/diffusers/models/modeling_utils.py (hunk 8 above)
Comment (devanshi00) [reply src/diffusers/models/modeling_utils.py]: I’ve changed the `load_pretrained()` function and removed the redundant initialization. Happy to clarify or discuss further if you have any questions.

**Comment by sayakpaul** [comment]
I ran the benchmark script myself on an H100 machine (with this PR branch). Here are my findings and I am awestruck!


model | avg_accelerate_time | std_accelerate_time | avg_flashpack_time | std_flashpack_time | total_bytes
-- | -- | -- | -- | -- | --
Qwen/Qwen-Image-Edit | 20.3753 | 0.7633 | 2.7582 | 0.8035 | 40860802176
Wan-AI/Wan2.1-T2V-1.3B-Diffusers | 7.0213 | 10.5124 | 0.263 | 0.0612 | 2837993600
Wan-AI/Wan2.1-T2V-14B-Diffusers | 26.9982 | 0.8495 | 1.9878 | 0.5714 | 28576983168
black-forest-labs/FLUX.1-dev | 12.3575 | 0.6743 | 1.6569 | 0.4736 | 23802816640

File: src/diffusers/models/modeling_utils.py [hunk 9]
```diff
@@ -1280,6 +1321,28 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
+            from ..utils.flashpack_utils import load_flashpack
+            # Even when using FlashPack, we preserve `low_cpu_mem_usage` behavior by initializing
+            # the model with meta tensors. Since FlashPack cannot write into meta tensors, we
+            # explicitly materialize parameters before loading to ensure correctness and parity
+            # with the standard loading path.
+            if any(p.device.type == "meta" for p in model.parameters()):
+                 model.to_empty(device="cpu")
+            load_flashpack(model, flashpack_file)
+            model.register_to_config(_name_or_path=pretrained_model_name_or_path)
+            model.eval()
```
sayakpaul, line 1339: Why should these have to be conditioned under `if flashpack_file is not None`?

src/diffusers/models/modeling_utils.py
```diff
@@ -1280,6 +1321,28 @@ def from_pretrained(cls, pretrained_model_name_or_path: Optional[Union[str, os.P
...
+
+            if output_loading_info:
+                return model, {
+                    "missing_keys": [],
+                    "unexpected_keys": [],
+                    "mismatched_keys": [],
+                    "error_msgs": [],
+                }
+
+            return model
```
sayakpaul, line 1349: This as well.

src/diffusers/models/modeling_utils.py
devanshi00: With the current early-return behavior, the FlashPack code path is intentionally isolated from the standard loading flow. After instantiating the model and loading weights via FlashPack, we finalize the model (registering _name_or_path and switching to eval mode) and return immediately.
As a result, the model deliberately does not go through the usual checkpoint-loading logic (load_state_dict, _load_pretrained_model), device-map dispatch, sharding handling, quantizer post-processing, or the common end-of-function finalization.

src/diffusers/models/modeling_utils.py
devanshi00: This block exists to preserve the from_pretrained(..., output_loading_info=True) API contract for the FlashPack loading path. Since FlashPack bypasses _load_pretrained_model, none of the usual key-matching or error-collection logic is executed, so there is no meaningful loading diagnostics to report. Returning an empty but well-formed loading_info dict keeps the return type consistent with the standard loading path without implying any missing or mismatched keys were checked.

src/diffusers/models/modeling_utils.py
sayakpaul: Let me look into it and see if I can help with further simplifying.

sayakpaul:
@devanshi00 let me know your thoughts on https://github.com/devanshi00/diffusers/pull/2 when you have a moment.

devanshi00:
### FlashPack Loading Semantics and Limitations

FlashPack currently operates on **fully materialized tensors** and loads them **directly into already-instantiated model parameters**. It does not expose an intermediate loading stage where tensors can be:

- inspected before allocation,
- transformed (e.g., quantized or dequantized),
- or placed conditionally on specific devices.

Loaders that support **quantization, parallelism, or device mapping** fundamentally rely on such an intermediate stage to:

- decide tensor dtype or quantization format,
- replace modules (e.g., with quantized layers),
- or shard parameters across devices before materialization.

Because FlashPack bypasses these mechanisms and assumes the checkpoint already represents the **final, inference-ready model state**, it cannot be combined with loaders that require tensor transformation or module replacement, such as **quantized model loading** or **parallel model loading**.

Supporting these features would require a substantial redesign of FlashPack into a **policy-aware loading system**, which would negate its purpose as a **fast, minimal-overhead loader**. Therefore, FlashPack intentionally trades flexibility for speed and simplicity in the current implementation.

devanshi00:
@sayakpaul  any updates on https://github.com/devanshi00/diffusers/pull/2?

sayakpaul:
@devanshi00 I would like to hear @DN6's thoughts on https://github.com/huggingface/diffusers/pull/12999#issuecomment-3812971384 before making any further updates.

devanshi00:
For further clarification, in the integrations from the official repo also sugggest the same.
For pipeline loading [here](https://github.com/fal-ai/flashpack/blob/main/src/flashpack/integrations/diffusers/pipeline.py)
The decisive block is:

```python
if issubclass(class_obj, FlashPackDiffusersModelMixin) or issubclass(
    class_obj, FlashPackTransformersModelMixin
):
    ...
    return class_obj.from_pretrained_flashpack(
        ...
    ).to(device)
   
```
which implies early return for the flashpack model.
Quantization is only applied later in the standard loading path:
```python
if (
    quantization_config is not None
    and isinstance(quantization_config, PipelineQuantizationConfig)
    and issubclass(class_obj, torch.nn.Module)
):
    model_quant_config = quantization_config._resolve_quant_config(...)
    loading_kwargs["quantization_config"] = model_quant_config
```
However, this code appears after the FlashPack early return and is therefore unreachable for FlashPack models.
Device map and dispatch logic are also skipped
In the standard loading path, models may later reach:
```python
dispatch_model(loaded_sub_model, device_map=device_map, force_hooks=True)
```
But FlashPack models never reach:
```python
loaded_sub_model = load_method(...)
```
because they are returned earlier.

DN6:
Hi @devanshi00, thank you so much for taking the time to work on this. We really do appreciate your efforts in coming up with a solution for Flashpack loading. 

Unfortunately, we have a bit of a conflict here. There's an existing PR (#12700) for FlashPack that is a bit further along and more closely matches our expectations for file loading. Given that overlap, I don't think it makes sense to move forward with this PR. 

I'm sorry this one didn't work out. If you are still interested in contributing to diffusers, please tag us on any of the open issues and we can assign it to you.  

Thank you again for your time.

devanshi00:
I completely understand the overlap with the existing FlashPack PR and appreciate the clarification. It was a great learning experience working on this issue, and I’m grateful for the feedback along the way.
I’d definitely be interested in continuing to contribute to diffusers. In particular, I was looking at issues like https://github.com/huggingface/diffusers/issues/12235, https://github.com/huggingface/diffusers/issues/12368
If either of these falls within the scope of contributions and doesn’t already have an active PR in progress, I’d be happy to take one on. Please let me know if it makes sense for me to work on any of them, or if there are other issues you’d recommend.

DN6:
Sorry for the delay @devanshi00. I'll assign WanImage2ImagePipeline to you.

PR #13002, opened 2026-01-20
[modular]support klein

yiyixuxu:
### testing klein
```py
import torch
import gc

device = "cuda"
dtype = torch.bfloat16

# 4b-base
repo_id = "black-forest-labs/FLUX.2-klein-base-4B"

pipeline = ModularPipeline.from_pretrained(repo_id)


pipeline.load_components(torch_dtype=dtype)
pipeline.to(device)
prompt = "a photo of a forest with mist swirling around the tree trunks. The word 'FLUX.2' is painted over it in big, red brush strokes with visible texture"
height = 768
width = 1360
num_inference_steps = 50
generator = torch.Generator(device=device).manual_seed(42)

out = pipeline(
    prompt=prompt,
    height=height,
    width=width,
    generator=generator
).images[0]

out.save("output_4b_base.png")

del pipeline
gc.collect()
torch.cuda.empty_cache()

# 9b-base
repo_id = "black-forest-labs/FLUX.2-klein-base-9B"

pipeline = ModularPipeline.from_pretrained(repo_id)
pipeline.load_components(torch_dtype=dtype)
pipeline.to(device)
prompt = "a photo of a forest with mist swirling around the tree trunks. The word 'FLUX.2' is painted over it in big, red brush strokes with visible texture"
height = 768
width = 1360
num_inference_steps = 50
generator = torch.Generator(device=device).manual_seed(42)

out = pipeline(
    prompt=prompt,
    height=height,
    width=width,
    generator=generator
).images[0]

out.save("output_9b_base.png")


# 4b

repo_id = "black-forest-labs/FLUX.2-klein-4B"
pipeline = ModularPipeline.from_pretrained(repo_id)
pipeline.load_components(torch_dtype=dtype)
pipeline.to(device)

num_inference_steps = 4
generator = torch.Generator(device=device).manual_seed(42)

out = pipeline(
    prompt=prompt,
    height=height,
    width=width,
    generator=generator
).images[0]

out.save("output_4b.png")

del pipeline
gc.collect()
torch.cuda.empty_cache()

# 9b
repo_id = "black-forest-labs/FLUX.2-klein-9B"
pipeline = ModularPipeline.from_pretrained(repo_id)
pipeline.load_components(torch_dtype=dtype)
pipeline.to(device)

num_inference_steps = 4
generator = torch.Generator(device=device).manual_seed(42)

out = pipeline(
    prompt=prompt,
    height=height,
    width=width,
    generator=generator
).images[0]

out.save("output_9b.png")

del pipeline
gc.collect()
torch.cuda.empty_cache()
```

### testing klein with modular setting
```py
import torch
import gc

device = "cuda"
dtype = torch.bfloat16

prompt = "a photo of a forest with mist swirling around the tree trunks. The word 'FLUX.2' is painted over it in big, red brush strokes with visible texture"
height = 768
width = 1360
# 4b

repo_ids = ["black-forest-labs/FLUX.2-klein-4B", "black-forest-labs/FLUX.2-klein-9B", "black-forest-labs/FLUX.2-klein-base-4B", "black-forest-labs/FLUX.2-klein-base-9B"]
for repo_id in repo_ids:
    print(f" =========================================")
    print(f" {repo_id}")
    blocks = ModularPipeline.from_pretrained(repo_id).blocks

    text_encoder_node = blocks.sub_blocks["text_encoder"].init_pipeline(repo_id)
    vae_encoder_node = blocks.sub_blocks["vae_encoder"].init_pipeline(repo_id)
    decoder_node = blocks.sub_blocks["decode"].init_pipeline(repo_id)
    denoise_node = blocks.sub_blocks["denoise"].init_pipeline(repo_id)


    text_encoder_node.load_components(torch_dtype=dtype)
    text_encoder_node.to(device)
    print(" ")
    print(f" text_encoder block: {text_encoder_node.blocks.doc}")
    text_embeddings = text_encoder_node(prompt=prompt).get_by_kwargs("denoiser_input_fields")

    denoise_node.load_components(torch_dtype=dtype)
    denoise_node.to(device)
    print(" ")
    print(f" denoise block: {denoise_node.blocks.doc}")
    if "base" in repo_id:
        num_inference_steps = 50
    else:
        num_inference_steps = 4

    latents = denoise_node(
        **text_embeddings, 
        height=height, 
        width=width, 
        generator=torch.Generator(device=device).manual_seed(42)
        ).latents

    decoder_node.load_components(torch_dtype=dtype)
    decoder_node.to(device)
    print(" ")
    print(f" decoder block: {decoder_node.blocks.doc}")
    print(" ")
    image = decoder_node(latents=latents).images[0]
    image.save(f"output_{repo_id.split("/")[-1]}.png")
```

### testing flux2-dev

```py
from diffusers import ModularPipeline, ComponentsManager
from transformers import AutoProcessor
import torch

device = "cuda"
dtype = torch.bfloat16
repo_id = "black-forest-labs/FLUX.2-dev"

components = ComponentsManager()
# there is an issue in transformer `sub_folder` arg is ignored 
cached_tokenizer_path = "/local/path/to/tokenizer/folder"
tokenizer = PixtralProcessor.from_pretrained(cached_tokenizer_path)

pipeline = ModularPipeline.from_pretrained(repo_id, components_manager=components)
pipeline.update_components(tokenizer=tokenizer)
pipeline.load_components(torch_dtype=dtype)

components.enable_auto_cpu_offload(device=device)

prompt = "a photo of a forest with mist swirling around the tree trunks. The word 'FLUX.2' is painted over it in big, red brush strokes with visible texture"
height = 768
width = 1360
num_inference_steps = 50
generator = torch.Generator(device=device).manual_seed(42)

print(f" pipeline: {pipeline.blocks.doc}")

out = pipeline(
    prompt=prompt,
    height=height,
    width=width,
    generator=generator
).images[0]

out.save("output_dev.png")
```
 modular seetting

```py
from diffusers import ModularPipeline, ComponentsManager
from transformers import PixtralProcessor
import torch

device = "cuda"
dtype = torch.bfloat16

components = ComponentsManager()
# dev
repo_id = "black-forest-labs/FLUX.2-dev"

blocks = ModularPipeline.from_pretrained(repo_id).blocks
text_node = blocks.sub_blocks["text_encoder"].init_pipeline(repo_id, components_manager=components)
denoise_node = blocks.sub_blocks["denoise"].init_pipeline(repo_id, components_manager=components)
decoder_node = blocks.sub_blocks["decode"].init_pipeline(repo_id, components_manager=components)


cached_tokenizer_path = "/path/to/local/folder/tokenizer"
tokenizer = PixtralProcessor.from_pretrained(cached_tokenizer_path)
print(f" tokenizer: {tokenizer}")

text_node.update_components(tokenizer=tokenizer)
text_node.load_components(torch_dtype=dtype)
denoise_node.load_components(torch_dtype=dtype)
decoder_node.load_components(torch_dtype=dtype)

components.enable_auto_cpu_offload(device=device)

prompt = "a photo of a forest with mist swirling around the tree trunks. The word 'FLUX.2' is painted over it in big, red brush strokes with visible texture"
height = 768
width = 1360
num_inference_steps = 50
generator = torch.Generator(device=device).manual_seed(42)

print(f" text_node: {text_node.blocks.doc}")
text_embeddings = text_node(prompt=prompt).get_by_kwargs("denoiser_input_fields")

print(f" denoise_node: {denoise_node.blocks.doc}")
latents = denoise_node(**text_embeddings, height=height, width=width, num_inference_steps=num_inference_steps, generator=generator).latents
# print(f" latents: {latents}")

print(f" decoder_node: {decoder_node.blocks.doc}")
image = decoder_node(latents=latents).images[0]
print(f" image: {image}")
image.save("output_dev.png")
```

### DIFF
--- a/src/diffusers/modular_pipelines/flux2/encoders.py
+++ b/src/diffusers/modular_pipelines/flux2/encoders.py
@@ -514,7 +514,7 @@ class Flux2KleinBaseTextEncoderStep(ModularPipelineBlocks):
         )
 
         if components.requires_unconditional_embeds:
-            negative_prompt = ""
+            negative_prompt = [""] * len(prompt)
             block_state.negative_prompt_embeds = self._get_qwen3_prompt_embeds(
                 text_encoder=components.text_encoder,
                 tokenizer=components.tokenizer,
--- a/tests/modular_pipelines/flux2/test_modular_pipeline_flux2_klein.py
+++ b/tests/modular_pipelines/flux2/test_modular_pipeline_flux2_klein.py
@@ -1,0 +1,91 @@
+# coding=utf-8
+# Copyright 2025 HuggingFace Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import random
+
+import numpy as np
+import PIL
+import pytest
+
+from diffusers.modular_pipelines import (
+    Flux2KleinAutoBlocks,
+    Flux2KleinModularPipeline,
+)
+
+from ...testing_utils import floats_tensor, torch_device
+from ..test_modular_pipelines_common import ModularPipelineTesterMixin
+
+
+class TestFlux2ModularPipelineFast(ModularPipelineTesterMixin):
+    pipeline_class = Flux2KleinModularPipeline
+    pipeline_blocks_class = Flux2KleinAutoBlocks
+    pretrained_model_name_or_path = "hf-internal-testing/tiny-flux2-klein-modular"
+
+    params = frozenset(["prompt", "height", "width"])
+    batch_params = frozenset(["prompt"])
+
+    def get_dummy_inputs(self, seed=0):
+        generator = self.get_generator(seed)
+        inputs = {
+            "prompt": "A painting of a squirrel eating a burger",
+            # TODO (Dhruv): Update text encoder config so that vocab_size matches tokenizer
+            "max_sequence_length": 8,  # bit of a hack to workaround vocab size mismatch
+            "text_encoder_out_layers": (1,),
+            "generator": generator,
+            "num_inference_steps": 2,
+            "height": 32,
+            "width": 32,
+            "output_type": "pt",
+        }
+        return inputs
+
+    def test_float16_inference(self):
+        super().test_float16_inference(9e-2)
+
+
+class TestFlux2ImageConditionedModularPipelineFast(ModularPipelineTesterMixin):
+    pipeline_class = Flux2KleinModularPipeline
+    pipeline_blocks_class = Flux2KleinAutoBlocks
+    pretrained_model_name_or_path = "hf-internal-testing/tiny-flux2-klein-modular"
+
+    params = frozenset(["prompt", "height", "width", "image"])
+    batch_params = frozenset(["prompt", "image"])
+
+    def get_dummy_inputs(self, seed=0):
+        generator = self.get_generator(seed)
+        inputs = {
+            "prompt": "A painting of a squirrel eating a burger",
+            # TODO (Dhruv): Update text encoder config so that vocab_size matches tokenizer
+            "max_sequence_length": 8,  # bit of a hack to workaround vocab size mismatch
+            "text_encoder_out_layers": (1,),
+            "generator": generator,
+            "num_inference_steps": 2,
+            "height": 32,
+            "width": 32,
+            "output_type": "pt",
+        }
+        image = floats_tensor((1, 3, 64, 64), rng=random.Random(seed)).to(torch_device)
+        image = image.cpu().permute(0, 2, 3, 1)[0]
+        init_image = PIL.Image.fromarray(np.uint8(image * 255)).convert("RGB")
+        inputs["image"] = init_image
+
+        return inputs
+
+    def test_float16_inference(self):
+        super().test_float16_inference(9e-2)
+
+    @pytest.mark.skip(reason="batched inference is currently not supported")
+    def test_inference_batch_single_identical(self, batch_size=2, expected_max_diff=0.0001):
+        return
--- a/tests/modular_pipelines/flux2/test_modular_pipeline_flux2_klein_base.py
+++ b/tests/modular_pipelines/flux2/test_modular_pipeline_flux2_klein_base.py
@@ -1,0 +1,91 @@
+# coding=utf-8
+# Copyright 2025 HuggingFace Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import random
+
+import numpy as np
+import PIL
+import pytest
+
+from diffusers.modular_pipelines import (
+    Flux2KleinBaseAutoBlocks,
+    Flux2KleinModularPipeline,
+)
+
+from ...testing_utils import floats_tensor, torch_device
+from ..test_modular_pipelines_common import ModularPipelineTesterMixin
+
+
+class TestFlux2ModularPipelineFast(ModularPipelineTesterMixin):
+    pipeline_class = Flux2KleinModularPipeline
+    pipeline_blocks_class = Flux2KleinBaseAutoBlocks
+    pretrained_model_name_or_path = "hf-internal-testing/tiny-flux2-klein-base-modular"
+
+    params = frozenset(["prompt", "height", "width"])
+    batch_params = frozenset(["prompt"])
+
+    def get_dummy_inputs(self, seed=0):
+        generator = self.get_generator(seed)
+        inputs = {
+            "prompt": "A painting of a squirrel eating a burger",
+            # TODO (Dhruv): Update text encoder config so that vocab_size matches tokenizer
+            "max_sequence_length": 8,  # bit of a hack to workaround vocab size mismatch
+            "text_encoder_out_layers": (1,),
+            "generator": generator,
+            "num_inference_steps": 2,
+            "height": 32,
+            "width": 32,
+            "output_type": "pt",
+        }
+        return inputs
+
+    def test_float16_inference(self):
+        super().test_float16_inference(9e-2)
+
+
+class TestFlux2ImageConditionedModularPipelineFast(ModularPipelineTesterMixin):
+    pipeline_class = Flux2KleinModularPipeline
+    pipeline_blocks_class = Flux2KleinBaseAutoBlocks
+    pretrained_model_name_or_path = "hf-internal-testing/tiny-flux2-klein-base-modular"
+
+    params = frozenset(["prompt", "height", "width", "image"])
+    batch_params = frozenset(["prompt", "image"])
+
+    def get_dummy_inputs(self, seed=0):
+        generator = self.get_generator(seed)
+        inputs = {
+            "prompt": "A painting of a squirrel eating a burger",
+            # TODO (Dhruv): Update text encoder config so that vocab_size matches tokenizer
+            "max_sequence_length": 8,  # bit of a hack to workaround vocab size mismatch
+            "text_encoder_out_layers": (1,),
+            "generator": generator,
+            "num_inference_steps": 2,
+            "height": 32,
+            "width": 32,
+            "output_type": "pt",
+        }
+        image = floats_tensor((1, 3, 64, 64), rng=random.Random(seed)).to(torch_device)
+        image = image.cpu().permute(0, 2, 3, 1)[0]
+        init_image = PIL.Image.fromarray(np.uint8(image * 255)).convert("RGB")
+        inputs["image"] = init_image
+
+        return inputs
+
+    def test_float16_inference(self):
+        super().test_float16_inference(9e-2)
+
+    @pytest.mark.skip(reason="batched inference is currently not supported")
+    def test_inference_batch_single_identical(self, batch_size=2, expected_max_diff=0.0001):
+        return


src/diffusers/modular_pipelines/flux2/before_denoise.py
```diff
@@ -353,7 +353,7 @@ def description(self) -> str:
     def inputs(self) -> List[InputParam]:
         return [
             InputParam(name="prompt_embeds", required=True),
-            InputParam(name="latent_ids"),
```
yiyixuxu, line 356: removed because latent_ids are not used in this block I think

src/diffusers/modular_pipelines/flux2/encoders.py
```diff
@@ -79,10 +79,8 @@ def expected_components(self) -> List[ComponentSpec]:
     def inputs(self) -> List[InputParam]:
         return [
             InputParam("prompt"),
-            InputParam("prompt_embeds", type_hint=torch.Tensor, required=False),
```
yiyixuxu, line 82: in modular, we can just pop out the `text_encoder` block to use in standalone manner if we want to compute them separately. so I'm trying to remove these arguments everywhere to simplify code a bit (same with image_latents too, it gets really complicated for some inpaiting pipeline)

```py
blocks = ....
text_node = blocks.pop("text_encoder").init_pipeline(repo_id)

pipe = blocks.init_pipeline(repo_id)
prompt_embeds = text_node(prompt = ..)
out = pipe(prompt_embeds = ...)
```

src/diffusers/modular_pipelines/flux2/before_denoise.py
```diff
@@ -353,7 +353,7 @@ def description(self) -> str:
...
+            InputParam(name="negative_prompt_embeds", required=False),
```
sayakpaul, line 396: No strong opinions but WDYT of creating a separate block for Klein altogether? I think this way it will be a bit easier to debug and also separate concerns?

My suggestions mainly comes from the fact that Flux.2-Dev doesn't use  `negative_prompt_embeds` while Flux.2-Klein does. So, maybe that warrants creating separate blocks.

src/diffusers/modular_pipelines/flux2/denoise.py
```diff
@@ -134,6 +137,241 @@ def __call__(
...
+            InputParam(
+                "image_latent_ids",
+                type_hint=torch.Tensor,
+                description="Position IDs for image latents. Shape: (B, img_seq_len, 4)",
+            ),
+            InputParam(
+                "prompt_embeds",
+                required=True,
+                type_hint=torch.Tensor,
+                description="Text embeddings from Mistral3",
```
sayakpaul:
```suggestion
                description="Text embeddings from Qwen3",
```

src/diffusers/modular_pipelines/flux2/encoders.py
```diff
@@ -270,6 +245,153 @@ def __call__(self, components: Flux2ModularPipeline, state: PipelineState) -> Pi
...
+    def intermediate_outputs(self) -> List[OutputParam]:
+        return [
+            OutputParam(
+                "prompt_embeds",
+                kwargs_type="denoiser_input_fields",
+                type_hint=torch.Tensor,
+                description="Text embeddings from qwen3 used to guide the image generation",
+            ),
+            OutputParam(
+                "negative_prompt_embeds",
```
sayakpaul, line 423: Should this also have its `kwargs_type` set as "denoiser_input_fields"?

src/diffusers/modular_pipelines/flux2/denoise.py
```diff
@@ -134,6 +137,241 @@ def __call__(
...
```
sayakpaul:
```suggestion
                description="Text embeddings from Qwen3",
```

src/diffusers/modular_pipelines/flux2/denoise.py
```diff
@@ -134,6 +137,241 @@ def __call__(
...
+                getattr(block_state, "negative_txt_ids", None),
+            ),
+        }
+
+        transformer_args = set(inspect.signature(components.transformer.forward).parameters.keys())
+        additional_cond_kwargs = {}
+        for field_name, field_value in block_state.denoiser_input_fields.items():
+            if field_name in transformer_args and field_name not in guider_inputs:
+                additional_cond_kwargs[field_name] = field_value
+        block_state.additional_cond_kwargs.update(additional_cond_kwargs)
```
sayakpaul: Interesting one. Out of curiosity, do you have an example for these args?

src/diffusers/modular_pipelines/flux2/encoders.py
```diff
@@ -79,10 +79,8 @@ def expected_components(self) -> List[ComponentSpec]:
     def inputs(self) -> List[InputParam]:
         return [
             InputParam("prompt"),
-            InputParam("prompt_embeds", type_hint=torch.Tensor, required=False),
             InputParam("max_sequence_length", type_hint=int, default=512, required=False),
             InputParam("text_encoder_out_layers", type_hint=Tuple[int], default=(10, 20, 30), required=False),
-            InputParam("joint_attention_kwargs"),
```
sayakpaul, line 85: But why remove it? 👀

src/diffusers/modular_pipelines/flux2/modular_pipeline.py
```diff
@@ -55,3 +57,56 @@ def num_channels_latents(self):
...
+    > [!WARNING] > This is an experimental feature and is likely to change in the future.
+    """
+
+    default_blocks_name = "Flux2KleinBaseAutoBlocks"
+
+    def get_default_blocks_name(self, config_dict: Optional[Dict[str, Any]]) -> Optional[str]:
+        if config_dict is not None and "is_distilled" in config_dict and config_dict["is_distilled"]:
+            return "Flux2KleinAutoBlocks"
+        else:
+            return "Flux2KleinBaseAutoBlocks"
```
sayakpaul, line 75: Maybe naming them as `Flux2KleinDistilledBlocks` and `Flux2KleinBaseAutoBlocks` is slightly better?

src/diffusers/modular_pipelines/flux2/modular_blocks_flux2_klein.py
```diff
@@ -0,0 +1,171 @@
...
+
+class Flux2KleinAutoBlocks(SequentialPipelineBlocks):
+    model_name = "flux2-klein"
+    block_classes = [
+        Flux2KleinTextEncoderStep(),
+        Flux2KleinAutoVaeEncoderStep(),
+        Flux2KleinCoreDenoiseStep(),
+        Flux2DecodeStep(),
+    ]
+    block_names = ["text_encoder", "vae_image_encoder", "denoise", "decode"]
```
asomoza:
```suggestion
    block_names = ["text_encoder", "vae_encoder", "denoise", "decode"]
```

src/diffusers/modular_pipelines/flux2/modular_blocks_flux2_klein.py
```diff
@@ -0,0 +1,171 @@
...
+
+class Flux2KleinBaseAutoBlocks(SequentialPipelineBlocks):
+    model_name = "flux2-klein"
+    block_classes = [
+        Flux2KleinTextEncoderStep(),
+        Flux2KleinAutoVaeEncoderStep(),
+        Flux2KleinBaseCoreDenoiseStep(),
+        Flux2DecodeStep(),
+    ]
+    block_names = ["text_encoder", "vae_image_encoder", "denoise", "decode"]
```
asomoza:
```suggestion
    block_names = ["text_encoder", "vae_encoder", "denoise", "decode"]
```

asomoza: I think flux2 dev has the same but haven't tested it yet

src/diffusers/modular_pipelines/flux2/encoders.py
yiyixuxu: i didn't see it used in the block

src/diffusers/modular_pipelines/flux2/encoders.py
yiyixuxu, line 423:
```suggestion
                "negative_prompt_embeds",
                kwargs_type="denoiser_input_fields",
```

src/diffusers/modular_pipelines/flux2/before_denoise.py
yiyixuxu: It's a fair point, but on the other hand, I've personally found that having too many blocks can become overwhelming - each time you need to add something, you still need to go through all of them and understand which ones to use.
I think it makes sense to just add the code in the same blocks here, it is so small and fits in. but this is really a matter of preference, not right or wrong. Maybe we'll know better in the future though after building more pipelines :)

src/diffusers/modular_pipelines/flux2/denoise.py
yiyixuxu: so we have this` kwargs_type="denoiser_input_fields"` to allow users to pass additional arguments to the denoiser model without modifying the block code, as long as they mark it with this `kwargs_type`.

In other words, this is a block takes **kwargs 🤣 

However, I'll remove this logic here for now - it might be over-engineering at this point. Writing a new block isn't that big a deal, and we can add this flexibility later if there's a clear need for it.

src/diffusers/modular_pipelines/flux2/before_denoise.py
yiyixuxu: Actually, I changed my mind - I agree it's better to separate them out. Otherwise negative_prompt_embeds will show up as an optional argument in the auto docstring for both Klein and Dev, which is confusing.
Note that in Qwen (https://github.com/huggingface/diffusers/blob/main/src/diffusers/modular_pipelines/qwenimage/inputs.py#L232), I'm experimenting with more composable blocks for situations like this that you can just reuse. But it also makes the blocks more complex, and I'm not sure if I'm over-engineering. So let's keep them simple here and see how it goes.

sayakpaul: Thank you!

src/diffusers/modular_pipelines/flux2/before_denoise.py
```diff
@@ -183,19 +174,14 @@ def __call__(self, components: Flux2ModularPipeline, state: PipelineState) -> Pi
...
-            block_state.device,
+            device,
             timesteps=timesteps,
             sigmas=sigmas,
             mu=mu,
         )
         block_state.timesteps = timesteps
         block_state.num_inference_steps = num_inference_steps
 
-        batch_size = block_state.batch_size * block_state.num_images_per_prompt
```
yiyixuxu, line 194: separated this to a prepare_guidance block

src/diffusers/modular_pipelines/flux2/decoders.py
```diff
@@ -29,29 +29,16 @@
 logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
 
 
-class Flux2DecodeStep(ModularPipelineBlocks):
+class Flux2UnpackLatentsStep(ModularPipelineBlocks):
```
yiyixuxu, line 32: move the "unpacking latent' out of decode step -> this is just to make it easier for decode step work in standalone manner, i.e. user only need to pass latent (not latent_id)

src/diffusers/modular_pipelines/flux2/modular_blocks_flux2.py
```diff
@@ -72,33 +76,56 @@ def description(self):
...
+                type_hint=torch.Tensor,
+                description="The latents from the denoising step.",
+            )
+        ]
 
 
 AUTO_BLOCKS = InsertableDict(
     [
         ("text_encoder", Flux2TextEncoderStep()),
-        ("text_input", Flux2TextInputStep()),
```
yiyixuxu, line 98: rearrangeed a bit so this works with mellon

src/diffusers/modular_pipelines/flux2/modular_pipeline.py
DN6: I agree. It could get a bit confusing down the line.

src/diffusers/modular_pipelines/flux2/encoders.py
DN6: I think the implication here is if we want to pass `prompt_embeds` to the pipeline, we would always have to pop out the text encoder step before doing so right? 

A snippet like this would end up encoding an empty prompt and using it instead of the passed `prompt_embeds`?  

```python
prompt_embeds = <loaded from somewhere else>
flux2_pipe = ModularPipeline.from_pretrained("...")
flux2_pipe(prompt_embeds=prompt_embeds)
```
IMO it's quite different from how Pipelines behave, so it could lead to some confusion. The change to Flux should be fine, but in cases where a pipeline might reuse the state (Krea RT, Waypoint-1) it could make for an awkward developer experience. 

```python
# create initial state with prompt embeds or latents
state = pipe(prompt="my prompt")
# remove text encoding block
pipe.blocks.pop("text_encoder")
# reuse the state so that latents or prompt embeds are not overwritten
pipe(state)
```  

I'm not against the change. Just flagging a potential point of confusion.

yiyixuxu: >I think the implication here is if we want to pass prompt_embeds to the pipeline, we would always have to pop out the text encoder step before doing so right?

We can also wrap the `text_encoder` block inside an `AutoPipelineBlock` like we did for vae_encoder: https://github.com/huggingface/diffusers/blob/main/src/diffusers/modular_pipelines/flux2/modular_blocks.py#L60
That way it only runs if you pass `prompt`, otherwise it skips. 

So like this:
```py
# here we don't pop out the text_encoder, keep it in the pipeline, but use it seperately 
text_node = pipe.blocks.sub_blocks["text_encoder"].init_pipeline(repo_id)
text_node.update_components(text_encoder=pipe.text_encoder, tokenizer=pipe.tokenizer)
text_embeddings = text_node(prompt="...").get_by_kwargs("denoiser_input_fields")

# this would work
out = pipe(**text_embeddings).images[0]
# this would also work
out = pipe(prompt="...").images[0]
```
That said, I don't feel strongly about keeping the behavior consistent with standard pipelines. Supporting both raw inputs and intermediate inputs adds complexity - you end up with a lot of conditional logic and validation etc. I particularly hate our inpainting pipelines where we had to handle the encoded prompt/image/mask and it became a mess.
I think cleaner separation is better, e.g. text node always processes text and outputs embeddings, denoise block always expects embeddings. Each block has a single responsibility and the data flow is explicit. Except for quick demos, no one wants to encode the same stuff more than once, downstream apps and UIs alike, so I think this is more of a real world use case anyway.

yiyixuxu: For Krea/Waypoint style use cases, if the prompt doesn't change between frames, you could very easily encode it just once outside of the pipeline:

```py
text_node = pipe.blocks.sub_blocks.pop("text_encoder").init_pipeline(...)
text_node.update_components(...)
text_embeddings = text_node(prompt="my prompt").get_by_kwargs("denoiser_input_fields")

state = pipe(**text_embeddings, image=image, button=ctrl().button, mouse=ctrl().mouse)

for _ in tqdm(range(1, num_frames)):
    state = pipe(state, **text_embeddings, button=ctrl().button, mouse=ctrl().mouse, ...)
```

yiyixuxu: >A snippet like this would end up encoding an empty prompt and using it instead of the passed prompt_embeds?

That's not good indeed, but I think it's because we forgot to mark the required inputs properly. We should always mark prompt as required for the text_encoder block and prompt_embeds as required for the input step where denoise begins,and  it would be clearer. Additionally, it's probably better to throw an error on unexpected inputs instead of just warning - in your example it would warn about unexpected inputs getting ignored, but I think erroring is better so users don't silently get wrong results.

I will open a PR for this and we can continue the discussion there:)

src/diffusers/modular_pipelines/flux2/modular_pipeline.py
yiyixuxu: yeah a bit confusing, but they named their models this way though - "Klein" for the distilled version and "Klein Base" for the non-distilled version.

I think we should keep the naming consistent with the official since users will encounter this naming when they figure out which checkpoint to use, so I don't think it helps for us to name them differently

yiyixuxu: actually, thought about this more, all flux are distilled (either guidance distilled or step distilled) - these 2 base models are first non-distilled checkpoint we see so it makes sense for them go with this naming convenstion actually

PR #13013, opened 2026-01-22
Modular[add more tests]

yiyixuxu:
add two more test:
1. test loading from standard non-modular repo
2. test running inference in modular setting: run text_encoder, vae_encoder/vae_decoder standalone 
3. added required_inputs signature tests: e.g `prompt` should be a required input for most pipelines

### DIFF
--- a/tests/modular_pipelines/test_modular_pipelines_common.py
+++ b/tests/modular_pipelines/test_modular_pipelines_common.py
@@ -165,6 +165,7 @@ class ModularPipelineTesterMixin:
         assert "text_encoder" in blocks.sub_blocks, "`text_encoder` block is not present in the pipeline"
         assert "denoise" in blocks.sub_blocks, "`denoise` block is not present in the pipeline"
         assert "decode" in blocks.sub_blocks, "`decode` block is not present in the pipeline"
+
         # manually set the components in the sub_pipe
         # a hack to workaround the fact the default pipeline properties are often incorrect for testing cases,
         # #e.g. vae_scale_factor is ususally not 8 because vae is configured to be smaller for testing


yiyixuxu:
will need to adjust conflicts and fix more stuff after this https://github.com/huggingface/diffusers/pull/13078/ and https://github.com/huggingface/diffusers/pull/13063
but ready for a review now! @DN6 @sayakpaul 

feel free to push directly into this PR too if you see anything you want to change

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -130,6 +139,98 @@ def _check_for_parameters(parameters, expected_parameters, param_type):
 
         _check_for_parameters(self.params, input_parameters, "input")
         _check_for_parameters(self.optional_params, optional_parameters, "optional")
+        _check_for_parameters(self.required_params, required_parameters, "required")
+
+    def test_loading_from_default_repo(self):
+        if self.default_repo_id is None:
+            return
```
sayakpaul, line 146:
```suggestion
    def test_loading_from_default_repo(self):
        if self.default_repo_id is None:
            pytest.mark.skip("The underlying test class doesn't implement `default_repo_id`, skipping test.")
```

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -130,6 +139,98 @@ def _check_for_parameters(parameters, expected_parameters, param_type):
...
+        # create text, denoise, decoder (and optional vae encoder) nodes
+        blocks = self.pipeline_blocks_class()
+
+        assert "text_encoder" in blocks.sub_blocks, "`text_encoder` block is not present in the pipeline"
+        assert "denoise" in blocks.sub_blocks, "`denoise` block is not present in the pipeline"
+        assert "decode" in blocks.sub_blocks, "`decode` block is not present in the pipeline"
+
+        # manually set the components in the sub_pipe
+        # a hack to workaround the fact the default pipeline properties are often incorrect for testing cases,
+        # #e.g. vae_scale_factor is ususally not 8 because vae is configured to be smaller for testing
```
sayakpaul, line 171: Is this something because of how defaults are set? I think we should fix this.

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -130,6 +139,98 @@ def _check_for_parameters(parameters, expected_parameters, param_type):
...
+        denoise_inputs = filter_inputs(denoise_available, denoise_node.blocks.input_names)
+        denoise_output = denoise_node(**denoise_inputs)
+        latents = denoise_output.latents
+
+        # 4. Decoder: takes from inputs + denoise_output
+        decode_available = {**inputs, "latents": latents}
+        decode_inputs = filter_inputs(decode_available, decoder_node.blocks.input_names)
+        modular_output = decoder_node(**decode_inputs).images
+
+        assert modular_output.shape == standard_output.shape, (
```
sayakpaul, line 231: Should we also add a check for the outputs? The outputs should also match right?

PR #13021, opened 2026-01-23
feat: add qkv projection fuse for longcat transformers

lgyStoic:
# What does this PR do?

longcat transformers block can fusing qkv projections to once calculation,
By this method we can get slightly performance enhancement by reduce the kernel launch overhead.

Fixes # (issue)

AttentionMixin has the implementation of `fuse_qkv_projections`， so we just simply inherit from this class, then we can enable fuse for longcat-images

## Before submitting
- [ ✅] Did you read the [contributor guideline](https://github.com/huggingface/diffusers/blob/main/CONTRIBUTING.md)?
- [ ✅] Did you read our [philosophy doc](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md) (important for complex PRs)?

## Who can review?

@yiyixuxu @sayakpaul

### DIFF
--- a/docker/diffusers-pytorch-cuda/Dockerfile
+++ b/docker/diffusers-pytorch-cuda/Dockerfile
@@ -2,7 +2,7 @@ FROM nvidia/cuda:12.1.0-runtime-ubuntu20.04
 LABEL maintainer="Hugging Face"
 LABEL repository="diffusers"
 
-ARG PYTHON_VERSION=3.12
+ARG PYTHON_VERSION=3.11
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get -y update \
@@ -32,10 +32,12 @@ RUN uv venv --python ${PYTHON_VERSION} --seed ${VIRTUAL_ENV}
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 # pre-install the heavy dependencies (these can later be overridden by the deps from setup.py)
+# Install torch, torchvision, and torchaudio together to ensure compatibility
 RUN uv pip install --no-cache-dir \
     torch \
     torchvision \
-    torchaudio
+    torchaudio \
+    --index-url https://download.pytorch.org/whl/cu121
 
 RUN uv pip install --no-cache-dir "git+https://github.com/huggingface/diffusers.git@main#egg=diffusers[test]"
 
--- a/docker/diffusers-pytorch-xformers-cuda/Dockerfile
+++ b/docker/diffusers-pytorch-xformers-cuda/Dockerfile
@@ -2,7 +2,7 @@ FROM nvidia/cuda:12.1.0-runtime-ubuntu20.04
 LABEL maintainer="Hugging Face"
 LABEL repository="diffusers"
 
-ARG PYTHON_VERSION=3.12
+ARG PYTHON_VERSION=3.11
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get -y update \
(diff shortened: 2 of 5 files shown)


sayakpaul:
Can you provide some numbers before and after using QKV fusion?

lgyStoic:
Sure, @sayakpaul 
with torchao quant and non-qkv projection fuse, it cost 4536ms in 20 num_steps
image
with torchao quant and qkv projection fuse, it cost  4164ms in 20 num_steps
image
I didn't import torch compile to compare, because compile may optimized this fusion, but in my experience torch compile is a bit unstable, if we can enable this fuse manually, it can benefit to developers.

sayakpaul:
Cool. I think QKV fusion is particularly also beneficial for quantization. Could we try it out with 4bit quantization (bitsandbytes) and see the results?

lgyStoic:
image
Unfortunately , I cant prove the result, even I increasing the prompt size to 10. I can see minimal time difference. In the image first result is bnb + compile + enable qkv fusion, another is bnb + compile + disable qkv fusion.

Here is my test code, I ran in h200 and h800, could you have a check @sayakpaul 
```
import torch
from diffusers import LongCatImagePipeline
from diffusers.quantizers import PipelineQuantizationConfig

    quant_kwargs={"load_in_4bit": True, "bnb_4bit_quant_type": "nf4", "bnb_4bit_compute_dtype": torch.bfloat16},
)

pipe = LongCatImagePipeline.from_pretrained(
                "meituan-longcat/LongCat-Image", 
                pipeline_quant_config=pipeline_quant_config,
                torch_dtype=torch.bfloat16)
pipe = pipe.to("cuda")

# For batch inference, use a list of prompts
prompts = [prompt] * 10  # batch size of 3

@torch.no_grad()
def profile(pipeline, prompts):
    # Warmup
    for _ in range(3):
        pipeline(
            prompts,
            height=1024,
            width=1024,
            guidance_scale=3.5,
            num_inference_steps=20
        ).images
    
    start.record()
    images = pipeline(
        prompts,
        height=1024,
        width=1024,
        guidance_scale=3.5,
        num_inference_steps=20
    ).images
    end.record()
    torch.cuda.synchronize()
    return start.elapsed_time(end), images

def accpipeline(pipeline:LongCatImagePipeline, use_torch_ao: bool = False, use_bitsandbytes: bool = False)->LongCatImagePipeline:
    # pipeline.transformer.fuse_qkv_projections()
    pipeline.vae.fuse_qkv_projections()
    pipeline.transformer.set_attention_backend("_flash_3_hub")
    pipeline.vae.to(memory_format=torch.channels_last)

    pipeline.transformer.compile_repeated_blocks(fullgraph=True, dynamic=True)
    return pipeline


acc_pipeline = accpipeline(pipe)
pipe_time, pipe_images = profile(acc_pipeline, prompts)
from PIL import Image
import numpy as np

# Save all batch images
for i, pipe_img in enumerate(pipe_images):
    if isinstance(pipe_img, torch.Tensor):
        img = pipe_img.detach().cpu().numpy()
        if img.ndim == 4:
            img = img[0]
        if img.shape[0] in (1, 3, 4) and img.shape[0] <= img.shape[-1]:
            # likely (C, H, W) -- convert to (H, W, C)
            img = np.transpose(img, (1, 2, 0))
        img = np.clip(img * 255, 0, 255).astype(np.uint8)
        if img.shape[-1] == 1:  # grayscale
            img = img.squeeze(-1)
        pil_img = Image.fromarray(img)
    elif isinstance(pipe_img, np.ndarray):
        img = pipe_img
        if img.ndim == 4:
            img = img[0]
        if img.shape[0] in (1, 3, 4) and img.shape[0] <= img.shape[-1]:
            img = np.transpose(img, (1, 2, 0))
        img = np.clip(img * 255, 0, 255).astype(np.uint8)
        if img.shape[-1] == 1:
            img = img.squeeze(-1)
        pil_img = Image.fromarray(img)
    elif isinstance(pipe_img, Image.Image):
        pil_img = pipe_img
    else:
        raise RuntimeError(f"Unexpected image type: {type(pipe_img)}")
    
    pil_img.save(f"output_{i}.png")

print(f"Batch size: {len(prompts)}")
print(f"Total time for batch: {pipe_time:.2f} ms")
print(f"Time per image: {pipe_time / len(prompts):.2f} ms")
```

Anyway, I think fusion QKV have no side-effect at any time.

sayakpaul:
I see this line: `# pipeline.transformer.fuse_qkv_projections()`. Did you comment out QKV fusion?

The code seems a bit haphazard to me in terms of state management, which is probably influencing the results. Could you button it up and implement it in a cleaner manner? 

But I will go ahead and merge the PR since the QKV fusion results are good enough

PR #13036, opened 2026-01-27
fast tok update

itazap:
Following transformers v5, we no longer have "slow" tokenizers that use a Trie - by default we use fast tokenizers. This script assumes always slow, so it is updated to work with fast!

### DIFF
--- a/examples/dreambooth/train_dreambooth_lora_qwen_image.py
+++ b/examples/dreambooth/train_dreambooth_lora_qwen_image.py
@@ -1467,7 +1467,8 @@ def main(args):
                 else:
                     num_repeat_elements = len(prompts)
                     prompt_embeds = prompt_embeds.repeat(num_repeat_elements, 1, 1)
-                    prompt_embeds_mask = prompt_embeds_mask.repeat(num_repeat_elements, 1)
+                    if prompt_embeds_mask is not None:
+                        prompt_embeds_mask = prompt_embeds_mask.repeat(num_repeat_elements, 1)
                 # Convert images to latent space
                 if args.cache_latents:
                     model_input = latents_cache[step].sample()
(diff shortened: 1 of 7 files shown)


itazap:
@sayakpaul here is the fix around the `update_trie()` issue you were facing. in v5 transformers we only kept "fast" tokenizers (which don't have `update_trie()`) --> so we can pin v5 for `transformers` to address the tests ! 😊

src/diffusers/loaders/textual_inversion.py
```diff
@@ -547,23 +549,20 @@ def unload_textual_inversion(
...
-            if token_id > last_special_token_id and token_id > last_special_token_id + key_id:
-                token = tokenizer._added_tokens_decoder[token_id]
-                tokenizer._added_tokens_decoder[last_special_token_id + key_id] = token
-                del tokenizer._added_tokens_decoder[token_id]
-                tokenizer._added_tokens_encoder[token.content] = last_special_token_id + key_id
-                key_id += 1
-        tokenizer._update_trie()
-        # set correct total vocab size after removing tokens
-        tokenizer._update_total_vocab_size()
+        # Fast tokenizers: serialize, filter tokens, reload
```
sayakpaul: Would it work with `transformers` (< v5) as well? 

If not, maybe we could keep maintaining two code paths? One for v5 and another one for < v5? This way, in the next release cycle, we can pin `transformers` ver to >=5.0.0. 

WDYT?

itazap: sounds good! I added it back

sayakpaul (approved):
Thanks a lot!

sayakpaul:
@bot /style

PR #13045, opened 2026-01-28
change to CUDA 12.9.

sayakpaul:
# What does this PR do?

As discussed internally.

PR #13047, opened 2026-01-28
[ci] uniform run times and wheels for pytorch cuda.

sayakpaul:
# What does this PR do?

Sadly the previous [PR](https://github.com/huggingface/diffusers/pull/13045) didn't quite fix it. I did locally build with the updated Dockerfiles on A100 and ran some tests, but nothing like [this](https://github.com/huggingface/diffusers/actions/runs/21437874189/job/61733405919?pr=12976#step:8:18319) showed up there. I even [tried](https://github.com/huggingface/diffusers/actions/runs/21438002088) to completely recreate the run as much as possible and still, nothing showed up. I guess it's one of those scenarios where things show up when you run them in the full roll.

~The issue is that even when we specify CUDA 12.9, PyTorch fails to find any wheel for it because we only have 12.8 and 13.0. It was installing the 12.8 wheel and that is causing a mismatch in the CUDA runtime currently.~

Edit: Changed to a different base image and specified thr 129 CUDA wheel as it was available.

### DIFF
--- a/src/diffusers/models/transformers/transformer_wan_animate.py
+++ b/src/diffusers/models/transformers/transformer_wan_animate.py
@@ -340,7 +340,8 @@ class WanAnimateMotionEncoder(nn.Module):
         weight = self.motion_synthesis_weight + 1e-8
         # Upcast the QR orthogonalization operation to FP32
         original_motion_dtype = motion_feat.dtype
-        motion_feat = motion_feat.to(weight.dtype)
+        motion_feat = motion_feat.to(torch.float32)
+        weight = weight.to(torch.float32)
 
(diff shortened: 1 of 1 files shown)


docker/diffusers-pytorch-cuda/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:12.9.0-runtime-ubuntu20.04
+FROM nvidia/cuda:12.9.1-runtime-ubuntu22.04
```
DN6: Is there no wheel for ubuntu24.04?

sayakpaul: Done

PR #13050, opened 2026-01-28
Add FLUX.2 Klein Inpaint Pipeline

### LINKED ISSUE #13005: Add Flux.2 Klein Inpaint pipeline
The currently implemented `Flux2KleinPipeline` supports txt2img and img2img, however there isn't an option to pass a mask image to it.

A `Flux2KleinInpaintPipeline` would be most useful to address this case and allow to use the model to its full extent.

adi776borate:
# What does this PR do?

Fixes #13005
This PR adds the `Flux2KleinInpaintPipeline` for image inpainting using the FLUX.2 [Klein] model with optional reference image conditioning.

#### Basic Inpainting

| Image | Mask | Prompt | Result |
|-------|------|--------|--------|
|  |  | Face of a yellow cat, high resolution, sitting on a park bench |  |
| | |A young college boy, high resolution, sitting on a park bench |  |

#### Inpainting with Reference Image

| Image | Mask | Reference | Prompt | Result |
|-------|------|-----------|--------|--------|
|  |  |  | Replace this ball |  |

### Known Limitations

1. **Generation quality may vary** - Some outputs may contain artifacts. This can often be mitigated with better prompts and tuning hyperparameters (strength, guidance_scale, num_inference_steps).

2. **Reference image conditioning is experimental** - Inpainting with `image_reference` may not consistently produce desired results.

| Image | Mask | Reference | Prompt | Result |
|-------|------|-----------|--------|--------|
|  |  |  | Replace puppy with panda |  |

These limitations may stem either from a bug in the pipeline implementation by me or from inherent constraints of the model. Feedback is appreciated.

## Before submitting

## Who can review?
@asomoza @sayakpaul 
Anyone in the community is free to review the PR once the tests have passed.


Natans8:
I have to be honest, I am not getting good results at all so far, especially when working with bounding box masks, even without a reference.
With `black-forest-labs/FLUX.2-klein-4B`, since it's a distilled model, `guidance_scale` is ignored, so I essentially have only `strength` and `num_inference_steps` to tweak.

At `strength=1.0` as it often goes with editing models, I get the entire inpainting area overridden edge to edge instead of just the prompted subject, usually with a lot of artefacts.
At lower strengths I get my prompt plain ignored.

On the other hand, I don't have the same issues with the regular `Flux2KleinPipeline` with the same parameter values and prompt, it changes the prompted subject and leaves the rest of the image untouched. Same goes for other model pipelines like `FluxKontextInpaintPipeline`.

If you ask, I can provide examples of the results I get once my GPU frees up.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1221 @@
...
+        guidance_scale=None,
+    ):
+        if strength < 0 or strength > 1:
+            raise ValueError(f"The value of strength should in [0.0, 1.0] but is {strength}")
+
+        if (
+            height is not None
+            and height % (self.vae_scale_factor * 2) != 0
+            or width is not None
+            and width % (self.vae_scale_factor * 2) != 0
```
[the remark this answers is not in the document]
adi776borate: I have followed the pattern from Flux2KleinPipeline and seems correct to me.
In Python, operator precedence is and > or
so, Python will evaluate this as
```py
(height is not None and height % (self.vae_scale_factor * 2) != 0) or
(width is not None and width % (self.vae_scale_factor * 2) != 0)
```
So, there is no precedence bug here.

And

> could raise a TypeError if width is None when trying to perform the modulo operation.

This cannot happen because of short-circuiting.

adi776borate:
I'd missed patchifying the mask latents earlier. With that corrected, here are some new example generations:

| Image | Mask | Prompt | Result |
|-------|------|--------|--------|
|  |  | Face of a yellow cat, high resolution, sitting on a park bench | _Using 4B model_|
|  |  |  | _Using 9B model_|
| | |A young college boy, high resolution, sitting on a park bench | _Using 4B model_|
|  |  |  | _Using 9B model_|
|boy| | A young woman standing with bouquet, natural pose, realistic proportions, consistent lighting and environment |  _Using 4B model_|
|  || A young woman standing with bouquet, natural pose, realistic proportions, consistent lighting and environment|  _Using 4B model_|


Below all are using 9B model with `strength=1.0`:

| Image | Mask | Reference | Prompt | Result |
|-------|------|-----------|--------|--------|
|  |  |  | Replace this ball ||
|  |  |  | Replace puppy with panda | |

@Natans8 this might improve what you observed but I'll suggest wait till maintainers' review.
Thanks!

adi776borate:
@asomoza @yiyixuxu 
Sorry for tagging again, but can you please review this?

asomoza:
oh I was waiting to see if the your changes fixed the issue that @Natans8 was having, I'll give it a test today review the PR

adi776borate:
Hi, @Natans8 can you please try my last commit? Also, it would be helpful, if you provide some examples.
@asomoza ,would appreciate a review when you get time.

Natans8:
> Hi, @Natans8 can you please try my last commit? Also, it would be helpful, if you provide some examples. @asomoza ,would appreciate a review when you get time.

It is still working a little weird to me. These are my results on the latest branch:
For example for the following code on 
```python
pipe = Flux2KleinInpaintPipeline.from_pretrained("black-forest-labs/FLUX.2-klein-4B", torch_dtype=torch.bfloat16)
inpainted_img = pipe(
    prompt="Remove the person, keep the background unchanged",
    image=image,
    mask_image=mask_pil, #rectangular mask
    num_inference_steps=4,
    guidance_scale=1.0,
    generator=generator,
    strength=1.0).images[0]
```
And this image
img
I get the following:
image
Or if I try to change the prompt to "remove all people" I can get something like this:
image
"Change the person's cloths to blue":
image
Same prompt with strength 0.8:
image

For comparison this is `Flux2KleinPipeline` with the prompt "Remove the boy in red":
image
"Change the clothes of the boy in red to blue, keep the others unchanged":
image
I had to fiddle a little in the prompt to specify which person I was referring to, not sure if that comparison is fair, but the point is that there is no artifacts.

When trying to tweak the strength, or the prompt of the inpaint pipeline I don't get much better results, sometimes it's random trees or rooms insensitive to the contents of the image, sometimes it's a black box, a white box, a slice of a wall. Of course I have to compare how it fares with 9B as well, ideally.

There is also the possibility that I'm just doing something incorrectly, I do not have a lot of experience with Diffusers. But i hope this feedback will be useful.

asomoza:
For some reason, the demo examples for inpainting that people always use are very simple which will never fail, for example the dog sitting on the bench from the original SD, any inpainting where there's a clear separation from the subject and the background is very simple and any modern model can do it without any issues, you can even use generative fill from photoshop before the AI era and it will also be good.

For what I've tested, Klein almost never changes the rest of the image if you are specific in the edit, the VAE is also very good which makes the quality loss minimal. 

The use case for this pipeline (IMO) is to be able to separate subjects when they're very similar and you can't just prompt it, in @Natans8 you can prompt to for a specific person but its enough of a good example because it's not a simple image to inpaint and we can prompt for people, also I wouldn't test this with the base model since it's not worth the 50 steps and CFG.

I did myself a test pipeline with differential diffusion to see if the model is capable of doing a good edit with restrictions, I will use for all the test just the prompt `remove the people` and using @Natans8 source image, I got this:



So the model is capable of doing it, so now to test this pipeline:

With the default strength:

|base|distilled|
|---|---|
|flux2klein_base_inpainting|flux2klein_inpainting|

I'll use just the distilled one since the results are similar, so varying the strenght I got these:

|0.8|0.9|
|---|---|
|flux2klein_0 8_inpainting|flux2klein_0 9_inpainting|

With 0.9 you could see it's somewhat decent and it's on pair for what I've seen with other inpainting pipelines.

I also tried with the prompt `change the the color of the shirt to blue while maintaining everything else intact ` but it changed the person too.

flux2klein_prompt_inpainting

@adi776borate does this match your results? if this looks usable to you, we can continue and review it.

adi776borate:
Yes, I also got sub-optimal results, sometimes very poor.
I used 9B model with strength 1.0.

1. "Remove the person" (non-rectangular mask):
students_9b_mask2

2. "Remove the person" (rectangular mask):
students_9b_mask

3. "Change the person's clothes to Blue" (non-rectangular mask):
students_9b_mask_blue

I'll take a look at this again in this week to identify the bug.

adi776borate:
Hi all! Sorry for the long gap.

I have made following changes in the pipeline now:
1. I removed `masked_image_latents` from the transformer input. Previously zeroing out the target region broadcasted "no information" across the entire sequence, which caused the model to collapse and ignore the prompt. Instead, now I explicitly encode the `init_image` and feed it into the sequence as clean context at every denoising step. (This improved the results a lot!)

2.  I changed mask latent preparation. Now it interpolate the mask directly to the final packed spatial size (`height // (vae_scale_factor * 2)`, `width // (vae_scale_factor * 2)`) by skipping patchifying.

Here are some results produced with the base 9B model:


<table style="width:100%;border-collapse:collapse;margin-bottom:16px;">
  <tr>
    <td colspan="3" style="padding:8px;border:1px solid #ddd;">
      <strong>Prompt:</strong> Face of a yellow cat with straight ears, high resolution, sitting on a park bench      &nbsp;&nbsp;&nbsp;
      <strong>Strength:</strong> 0.8
    </td>
  </tr>
  <tr>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Image</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Mask</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Result</th>
  </tr>
  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
dog_img
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
dog_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
flux2klein_inpainting_strength0 8
</td>
  </tr>
</table>


<table style="width:100%;border-collapse:collapse;margin-bottom:16px;">
  <tr>
    <td colspan="3" style="padding:8px;border:1px solid #ddd;">
      <strong>Prompt:</strong> Remove the person
      &nbsp;&nbsp;&nbsp;
      <strong>Strength:</strong> 1.0
    </td>
  </tr>
  <tr>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Image</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Mask</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Result</th>
  </tr>
  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
students
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
students_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
students_9b_mask_strength1 0
</td>
  </tr>
</table>


<table style="width:100%;border-collapse:collapse;margin-bottom:16px;">
  <tr>
    <td colspan="3" style="padding:8px;border:1px solid #ddd;">
      <strong>Prompt:</strong> Change the cloth color to blue 
      &nbsp;&nbsp;&nbsp;
      <strong>Strength:</strong> 1.0
    </td>
  </tr>
  <tr>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Image</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Mask</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Result</th>
  </tr>
  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
students
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
students_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
students_9b_mask_blue_strength1 0
</td>
  </tr>
</table>


<table style="width:100%;border-collapse:collapse;margin-bottom:16px;">
  <tr>
    <td colspan="3" style="padding:8px;border:1px solid #ddd;">
      <strong>Prompt:</strong> Put a fine diamond necklace around her neck
      &nbsp;&nbsp;&nbsp;
      <strong>Strength:</strong> 0.6
    </td>
  </tr>
  <tr>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Image</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Mask</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Result</th>
  </tr>
  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
woman_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
woman_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
woman_9b_necklace_0 6
</td>
  </tr>
</table>


<table style="width:100%;border-collapse:collapse;margin-bottom:16px;">
  <tr>
    <td colspan="3" style="padding:8px;border:1px solid #ddd;">
      <strong>Prompt:</strong> The woman is holding a hugging face emoji balloon with a thread
      &nbsp;&nbsp;&nbsp;
      <strong>Strength:</strong> 1.0
    </td>
  </tr>
  <tr>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Image</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Mask</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Result</th>
  </tr>
  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
woman_on_yatch_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
woman_on_yatch_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
balloon_9b_strength1 0
</td>
  </tr>
</table>


<table style="width:100%;border-collapse:collapse;margin-bottom:16px;">
  <tr>
    <td colspan="4" style="padding:8px;border:1px solid #ddd;">
      <strong>Prompt:</strong> Replace the ball
      &nbsp;&nbsp;&nbsp;
      <strong>Strength:</strong> 1.0
    </td>
  </tr>
  <tr>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Image</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Mask</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Reference Images</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Result</th>
  </tr>
  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
ball_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
ball_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
ball_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
flux2klein_inpainting_ref
</td>
  </tr>
</table>


<table style="width:100%;border-collapse:collapse;margin-bottom:16px;">
  <tr>
    <td colspan="4" style="padding:8px;border:1px solid #ddd;">
      <strong>Prompt:</strong> Replace the white car with the black car
      &nbsp;&nbsp;&nbsp;
      <strong>Strength:</strong> 1.0
    </td>
  </tr>
  <tr>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Image</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Mask</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Reference Images</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Result</th>
  </tr>
  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
ball_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
modern_car
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
ball_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
vintage_car_ref_strength_1 0
</td>
  </tr>
</table>


<table style="width:100%;border-collapse:collapse;margin-bottom:16px;">
  <tr>
    <td colspan="4" style="padding:8px;border:1px solid #ddd;">
      <strong>Prompt:</strong> the woman from image 1 wearing the shirt form image 2, with the top from image 3
      &nbsp;&nbsp;&nbsp;
      <strong>Strength:</strong> 1.0
    </td>
  </tr>
  <tr>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Image</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Mask</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Reference Images</th>
    <th style="padding:8px;border:1px solid #ddd;text-align:left;">Result</th>
  </tr>
  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
woman_sitting
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
woman_sitting_mask
</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
tshirt

</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">
multiref_strength_1 0
</td>
  </tr>

  <tr>
    <td style="padding:8px;border:1px solid #ddd;height:120px;"></td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;"></td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;">woman_sitting_mask</td>
    <td style="padding:8px;border:1px solid #ddd;height:120px;"></td>
  </tr>
</table>

I noticed no artifacts in several tests I conducted. Many times, a strength of 0.6 left the image unchanged (quite expected because we have now source image encoded and fed too), so I adjusted the default strength to 0.8. This feels like a good balance between making small edits and larger changes.

Ready for review @asomoza .

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1181 @@
...
+
+    @torch.no_grad()
+    @replace_example_docstring(EXAMPLE_DOC_STRING)
+    def __call__(
+        self,
+        prompt: Union[str, List[str]] = None,
+        image: PipelineImageInput = None,
+        image_reference: Optional[PipelineImageInput] = None,
+        mask_image: PipelineImageInput = None,
+        masked_image_latents: PipelineImageInput = None,
```
asomoza: this is not used right? lets remove it from here and the docstring

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1181 @@
...
+            [`~pipelines.flux2.Flux2PipelineOutput`] or `tuple`: [`~pipelines.flux2.Flux2PipelineOutput`] if
+            `return_dict` is True, otherwise a `tuple`. When returning a tuple, the first element is a list with the
+            generated images.
+        """
+
+        height = height or self.default_sample_size * self.vae_scale_factor
+        width = width or self.default_sample_size * self.vae_scale_factor
+
+        # 1. Check inputs. Raise error if not correct
+        self.check_inputs(
```
asomoza, line 958: since you added the `image_reference` we should check for this in the inputs

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1181 @@
...
+            guidance_scale=guidance_scale,
+        )
+
+        self._guidance_scale = guidance_scale
+        self._attention_kwargs = attention_kwargs
+        self._interrupt = False
+
+        # 2. Preprocess image
+        multiple_of = self.vae_scale_factor * 2
+        if image is not None and not (isinstance(image, torch.Tensor) and image.size(1) == self.latent_channels):
```
asomoza: with this check you're not taking into account when we pass the latents instead of an image, we should allow this.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1181 @@
...
+            image_reference_width = image_reference_width // multiple_of * multiple_of
+            image_reference_height = image_reference_height // multiple_of * multiple_of
+            image_reference = self.image_processor.resize(
+                image_reference, image_reference_height, image_reference_width
+            )
+            processed_image_reference = self.image_processor.preprocess(
+                image_reference,
+                image_reference_height,
+                image_reference_width,
+                crops_coords=crops_coords,
```
asomoza: you're cropping the reference image, is this needed?

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1181 @@
...
+            width (`int`, *optional*, defaults to self.default_sample_size * self.vae_scale_factor):
+                The width in pixels of the generated image. This is set to 1024 by default for the best results.
+            padding_mask_crop (`int`, *optional*, defaults to `None`):
+                The size of margin in the crop to be applied to the image and masking. If `None`, no crop is applied to
+                image and mask_image. If `padding_mask_crop` is not `None`, it will first find a rectangular region
+                with the same aspect ration of the image and contains all masked area, and then expand that area based
+                on `padding_mask_crop`. The image and mask_image will then be cropped based on the expanded area before
+                resizing to the original image size for inpainting. This is useful when the masked area is small while
+                the image is large and contain information irrelevant for inpainting, such as background.
+            strength (`float`, *optional*, defaults to 0.6):
```
asomoza:
```suggestion
            strength (`float`, *optional*, defaults to 0.8):
```

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1181 @@
...
+
+        batch_size, seq_len, _ = prompt_embeds.shape
+        prompt_embeds = prompt_embeds.repeat(1, num_images_per_prompt, 1)
+        prompt_embeds = prompt_embeds.view(batch_size * num_images_per_prompt, seq_len, -1)
+
+        text_ids = self._prepare_text_ids(prompt_embeds)
+        text_ids = text_ids.to(device)
+        return prompt_embeds, text_ids
+
+    def _encode_vae_image(self, image: torch.Tensor, generator: torch.Generator):
```
asomoza, line 542: for this one, we can't use the `# copied from` because you added a `to(...)` which to me is correct but this should be a copy of the original. Maybe we can use the copied from and then in a separate PR fix the original one?

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1181 @@
...
+                            timestep=timestep / 1000,
+                            guidance=None,
+                            encoder_hidden_states=negative_prompt_embeds,
+                            txt_ids=negative_text_ids,
+                            img_ids=img_ids,
+                            joint_attention_kwargs=self._attention_kwargs,
+                            return_dict=False,
+                        )[0]
+                    neg_noise_pred = neg_noise_pred[:, : latents.size(1)]
+                    noise_pred = neg_noise_pred + self.guidance_scale * (noise_pred - neg_noise_pred)
```
asomoza: just a nit: here you're using `self.guidance_scale` instead of a local variable, the other pipelines all use the local variable so to keep it consistent we should also do it, same as the base klein pipeline.

asomoza (commented):
looking good, just some minor comments.

Also can you add `_current_timestep`, same as the base klein pipeline, this is needed for some optimizations.

The results are now looking very good.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Done

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Done

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: thanks for pointing this out

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: The base pipeline preprocesses condition images with `resize_mode="crop"`, so I've followed that now.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Corrected

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Right. I'll open a follow-up PR to fix the original one, once this one is ready to merge

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Corrected

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
asomoza: ok, I didn't notice this before but the cropping seems wrong to me, with that we're losing information on the reference image, AFAIK the original code doesn't do that.  I'll look into that later.

in the meantime is fine with just crop, but we also seems to be missing the 1024x1024 cap size, like the one here:

https://github.com/huggingface/diffusers/blob/11a3284cee024bfff017408b303d18e88981d2ea/src/diffusers/pipelines/flux2/pipeline_flux2_klein.py#L767

adi776borate: The base klein pipeline performs the cropping. See [this](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/flux2/pipeline_flux2_klein.py#L774).
Since the code rounds down `image_width // multiple_of`, you lose at most 15 pixels from either dimension. On a 1024x1024 image, 15 pixels is about 1.4% of the image.

I have a doubt regarding 1024x1024 cap size you pointed out. The base klein piepline restricts its `image` arg dtype to PIL images (but wrongly claims in docstring that it supports all other dtypes too) and subsequently uses methods like `_resize_to_target_area()` and `.size()` which work on PIL only.
I am trying to support any dtype and using `get_default_height_width()` method. But, this method already performs the rounding operation, so to faithfully check the raw size I am thinking to implement a helper. I couldn't find any global util for this (please correct me if I am wrong).

asomoza: you're right with both,  the cropping is minimal, so it shouldn't affect the edit. 

About the second, yeah, you can do a small helper function and we can leave it in this file for the meantime, but I guess in a follow up PR we can move it to the image processor and fix the base klein pipeline too, and maybe all other edit ones that have it.

adi776borate: I have made the required changes, please review

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1232 @@
...
+        >>> pipe.to("cuda")
+
+        >>> prompt = "Replace this ball"
+        >>> img_url = "https://images.pexels.com/photos/39362/the-ball-stadion-football-the-pitch-39362.jpeg?auto=compress&cs=tinysrgb&dpr=1&w=500"
+        >>> mask_url = (
+        ...     "https://github.com/ZenAI-Vietnam/Flux-Kontext-pipelines/blob/main/assets/ball_mask.png?raw=true"
+        ... )
+        >>> image_reference_url = (
+        ...     "https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcTah3x6OL_ECMBaZ5ZlJJhNsyC-OSMLWAI-xw&s"
+        ... )
```
asomoza: lets not use images with external links that we don't know when they will stop working, I uploaded these example images here: https://huggingface.co/datasets/diffusers/docs-images/tree/main/klein_inpaint

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1232 @@
...
+        image: PipelineImageInput = None,
+        image_reference: Optional[PipelineImageInput] = None,
+        mask_image: PipelineImageInput = None,
+        height: Optional[int] = None,
+        width: Optional[int] = None,
+        padding_mask_crop: Optional[int] = None,
+        strength: float = 0.8,
+        num_inference_steps: int = 50,
+        sigmas: Optional[List[float]] = None,
+        guidance_scale: Optional[float] = 8.0,
```
asomoza: is this high cfg number required to do a correct inpainting?

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1232 @@
...
+    def current_timestep(self):
+        return self._current_timestep
+
+    @property
+    def interrupt(self):
+        return self._interrupt
+
+    @torch.no_grad()
+    @replace_example_docstring(EXAMPLE_DOC_STRING)
+    def __call__(
```
asomoza, line 821: you're using the old style typing, we're moving to newer python 3,10+ typing, you can use the base klein pipeline as a reference

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1232 @@
...
+        if hasattr(self.scheduler, "set_begin_index"):
+            self.scheduler.set_begin_index(t_start * self.scheduler.order)
+
+        return timesteps, num_inference_steps - t_start
+
+    def check_inputs(
+        self,
+        prompt,
+        image,
+        mask_image,
```
asomoza, line 710: missed this, but you have the args for `image` and `mask_image` which is correct but you're not validating them in the function body.

tests/pipelines/flux2/test_pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,197 @@
...
+)
+from ..test_pipelines_common import PipelineTesterMixin
+
+
+enable_full_determinism()
+
+
+class Flux2KleinInpaintPipelineFastTests(PipelineTesterMixin, unittest.TestCase):
+    pipeline_class = Flux2KleinInpaintPipeline
+    params = frozenset(["prompt", "image", "mask_image", "height", "width", "guidance_scale", "prompt_embeds"])
```
asomoza: you're missing the image_reference

asomoza (commented):
thanks! I added some other small nits and can you run:

`make style`
`make quality`
`make fix-copies`

after these last changes, I think we're ready to merge

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: I am doing the validation in the function body actually.
Check [this](https://github.com/adi776borate/diffusers/blob/b8aaa196fee9cf7d855e0a470ce7ba7e5ec3bef8/src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py#L742-L751). I followed the kontext inpaint pipeline for this.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: For strength=1.0, cfg of 4.0 will also do but at strength<1.0, 4.0 will under-edit. So, it depends on how strongly you want to inpaint (texture-level or full-subject).

This is a small quick exp I just ran where I am doing color edit:
prompt = "Change the clothes color to blue"

| Guidance Scale | Strength | Result |
| :--- | :--- | :--- |
| 4.0| 0.8|students_cfg4 0_strength0 8|
| 4.0| 1.0|students_cfg4 0_strength1 0|
| 8.0| 0.8|students_cfg8 0_strength0 8|
| 8.0| 1.0|students_9b_mask_blue_strength1 0|

In first case, it left image unchanged.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
asomoza: yeah but it's conditioned to the `padding_mask_crop` and also to just pil image, which means that if I don't pass a `padding_mask_crop` and also don't pass an `image` and `mask_image` the check inputs will pass, but this is meaningless for this pipeline no?

This is just me nitpicking though, since it will fail a couple of lines later but for me, we should validate here that they are present since that's what the function is for. 

I leave the decision to you since in kontext is already like this.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
asomoza: oh ok, lets just leave it at 8.0, I don't see that it burns the image or changes it in a bad way.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Done

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Corrected

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: You're right. I have added a check now.

tests/pipelines/flux2/test_pipeline_flux2_klein_inpaint.py
adi776borate: Done

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1245 @@
...
+            height=height,
+            width=width,
+            output_type=output_type,
+            prompt_embeds=prompt_embeds,
+            callback_on_step_end_tensor_inputs=callback_on_step_end_tensor_inputs,
+            padding_mask_crop=padding_mask_crop,
+            guidance_scale=guidance_scale,
+        )
+
+        self._guidance_scale = guidance_scale
```
asomoza, line 973: here we should add:

```
self._current_timestep = None
```

or else, when reusing the pipeline the timestep will be wrong.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1245 @@
...
+        with self.progress_bar(total=num_inference_steps) as progress_bar:
+            for i, t in enumerate(timesteps):
+                if self.interrupt:
+                    continue
+
+                self._current_timestep = t
+                # broadcast to batch dimension in a way that's compatible with ONNX/Core ML
+                timestep = t.expand(latents.shape[0]).to(latents.dtype)
+
+                latent_model_input = torch.cat([latents, condition_image_latents], dim=1)
```
asomoza, line 1172: here we're concatenating the latents and the condition_image_latents, but when called with the methods, you're passing a dtype that can be different.

latents uses `prompt_embeds.dtype` while condition_image_latents uses ` self.vae.dtype`, this means that if the user changes the vae dtype (for example to use a higher precision for encoding or decoding) this will fail.

This is probably something we can also fix in the other PR

tests/pipelines/flux2/test_pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,199 @@
...
+
+enable_full_determinism()
+
+
+class Flux2KleinInpaintPipelineFastTests(PipelineTesterMixin, unittest.TestCase):
+    pipeline_class = Flux2KleinInpaintPipeline
+    params = frozenset(
+        ["prompt", "image", "image_reference", "mask_image", "height", "width", "guidance_scale", "prompt_embeds"]
+    )
+    batch_params = frozenset(["prompt"])
```
asomoza: this should have also `image` and `mask_image` I think

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1245 @@
...
+            num_inference_steps,
+            device,
+            sigmas=sigmas,
+            mu=mu,
+        )
+        timesteps, num_inference_steps = self.get_timesteps(num_inference_steps, strength, device)
+
+        if num_inference_steps < 1:
+            raise ValueError(
+                f"After adjusting the num_inference_steps by strength parameter: {strength}, the number of pipeline"
```
asomoza:
```suggestion
                f"After adjusting the num_inference_steps by strength parameter: {strength}, the number of pipeline "
```

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1245 @@
...
+            else:
+                crops_coords = None
+                resize_mode = "default"
+
+            original_image = image
+            init_image = self.image_processor.preprocess(
+                image, image_height, image_width, crops_coords=crops_coords, resize_mode=resize_mode
+            )
+        else:
+            raise ValueError("image must be provided correctly for inpainting")
```
asomoza: since we're validating in `check_inputs` this path will never happen

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1245 @@
...
+
+        if isinstance(image, PIL.Image.Image):
+            return image.height, image.width
+        elif isinstance(image, torch.Tensor):
+            return image.shape[-2], image.shape[-1]
+        elif isinstance(image, np.ndarray):
+            return (
+                image.shape[-3] if image.ndim > 3 else image.shape[-2],
+                image.shape[-2] if image.ndim > 3 else image.shape[-1],
+            )
```
asomoza: shouldn't this just be `return image.shape[-3], image.shape[-2]`? I think you're returning the channels and the width when `ndim = 3`

asomoza (commented):
left some more comments, don't forget to run:

```
make style 
make quality
```

so we can run the tests.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: `_current_timestep` is already initialized to `None` in `init` and is reset to `None` after the denoising loop at L1216, so reuse is already handled correctly. 
I intentionally initialized in `init` rather than in `call` to handle edge case if a user accesses `pipe.current_timestep` before the first call. Happy to keep it aligned with other pipelines if you prefer.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: In `check_inputs()`, currently I am not enforcing that image is non-None. If I remove this else block, I'll have to add a check there, so code won't crash even if `image` or `mask_image` is not provided. Sounds good?

joangava (approved):

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
asomoza: I prefer if we keep it analogous to the other pipelines, it's easier to maintain and check for bugs.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
asomoza: yeah, sound good

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Right, but for 2D np arrays, we need conditional

adi776borate:
```python
elif isinstance(image, np.ndarray):
            if image.ndim >= 3:
                return image.shape[-3], image.shape[-2]
            return image.shape[-2], image.shape[-1]
```
this should work, right?

asomoza: yeah, that handles also the   grayscale images

adi776borate:
@asomoza I have made the suggested changes and also modified the batching logic as this specific test was failing with earlier code : `test_inference_batch_single_identical`. The previous code was merging all images in a batch into a single sequence. I changed it to have 1-to-1 mapping for batched inputs and broadcasting for single inputs.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1261 @@
...
+            num_channels_latents,
+            height,
+            width,
+            prompt_embeds.dtype,
+            device,
+            generator,
+            latents,
+        )
+
+        ref_images = [init_image]
```
asomoza: maybe a nit, but doing this we are re-encoding again the `init_image` which seems kind of wasteful, but maybe re using the latents will add too much code, WDYT?

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1261 @@
...
+                if self.do_classifier_free_guidance:
+                    with self.transformer.cache_context("uncond"):
+                        neg_noise_pred = self.transformer(
+                            hidden_states=latent_model_input,
+                            timestep=timestep / 1000,
+                            guidance=None,
+                            encoder_hidden_states=negative_prompt_embeds,
+                            txt_ids=negative_text_ids,
+                            img_ids=img_ids,
+                            joint_attention_kwargs=self._attention_kwargs,
```
asomoza:
```suggestion
                            joint_attention_kwargs=self.attention_kwargs,
```

just a nit (I know it comes from the base one)

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1261 @@
...
+        dtype,
+    ):
+        image_latents = []
+        for image in images:
+            image = image.to(device=device, dtype=dtype)
+
+            if image.shape[1] != self.latent_channels:
+                image_latent = self._encode_vae_image(image=image, generator=generator)
+            else:
+                image_latent = self._patchify_latents(image)
```
asomoza: we're using `_patchify_latents` here but not in `prepare_latents` we need to be consistent in both, if we expect them to be preprocessed, they should be already be normalized and patchified no?

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Ah! I see. You're right. I'll then also return pre-packed encoded latent to reuse. It won't add too much code, I think, just have to slip in on extra param.

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Agree. Since the reused latent is now 128ch (fully processed), I'll update the channel check to self.latent_channels * 4 and simplify the else to a passthrough.

adi776borate:
Hi! If everything appears to be in order, shall we proceed with the tests?

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1261 @@
...
+            init_image = self.image_processor.preprocess(
+                image, image_height, image_width, crops_coords=crops_coords, resize_mode=resize_mode
+            )
+
+        init_image = init_image.to(dtype=torch.float32)
+
+        # 2.2 Preprocess reference image
+        processed_image_reference = None
+        if image_reference is not None and not (
+            isinstance(image_reference, torch.Tensor) and image_reference.size(1) == self.latent_channels
```
asomoza: you skip the preprocessing but in 1017 you set the `processed_image_reference ` as None, so if the user passes the `image_reference` as tensors it skips this and never set the `processed_image_reference `

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1261 @@
...
+        if batch_size > image_latents.shape[0] and batch_size % image_latents.shape[0] == 0:
+            # expand init_latents for batch_size
+            additional_image_per_prompt = batch_size // image_latents.shape[0]
+            image_latents = torch.cat([image_latents] * additional_image_per_prompt, dim=0)
+        elif batch_size > image_latents.shape[0] and batch_size % image_latents.shape[0] != 0:
+            raise ValueError(
+                f"Cannot duplicate `image` of batch size {image_latents.shape[0]} to {batch_size} text prompts."
+            )
+        else:
+            image_latents = torch.cat([image_latents], dim=0)
```
asomoza:
```suggestion
```

the `image_latents` are already the same

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,1261 @@
...
+                for _ in range(1, b_i):
+                    t_offset += scale
+                    t = torch.tensor([t_offset]).view(-1)
+                    item_ids.append(
+                        torch.cartesian_prod(t, torch.arange(height), torch.arange(width), torch.arange(1))
+                    )
+                x_ids = torch.cat(item_ids, dim=0)  # (b_i * h * w, 4)
+                x_ids = x_ids.unsqueeze(0).expand(batch_size, -1, -1)
+                all_image_latent_ids.append(x_ids)
+                t_offset += scale
```
asomoza, line 423: is this path ever reached?

tests/pipelines/flux2/test_pipeline_flux2_klein_inpaint.py
```diff
@@ -0,0 +1,199 @@
...
+
+enable_full_determinism()
+
+
+class Flux2KleinInpaintPipelineFastTests(PipelineTesterMixin, unittest.TestCase):
+    pipeline_class = Flux2KleinInpaintPipeline
+    params = frozenset(
+        ["prompt", "image", "image_reference", "mask_image", "height", "width", "guidance_scale", "prompt_embeds"]
+    )
+    batch_params = frozenset(["prompt", "image", "mask_image"])
```
asomoza:
```suggestion
    batch_params = frozenset(["prompt", "image", "image_reference", "mask_image"])
```

asomoza (commented):
thanks, left a few comments, we're close to merge now

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Added an else block to handle it

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Agreed, removing it

src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py
adi776borate: Yes it is reachable when we pass multiple reference images for a single sample.

yiyixuxu:
@claude can you help to do a review here?

yiyixuxu:
@bot /style

joangava (approved):

yiyixuxu:
@adi776borate can you run `make fix-copies`?

adi776borate:
> @adi776borate can you run `make fix-copies`?

I intentionally held off on that. Could you please review #13299 first? I discussed this with @asomoza [here](https://github.com/huggingface/diffusers/pull/13050/changes#r2941755311).

adi776borate:
@asomoza I have addressed your last comments and also removed some redundant dtype casts while I was at it. This should now correctly handle pre-encoded latents as source or ref image input. I think we should also update the docs to tell the user that, we expect **patchified** latents as input.

asomoza:
thanks @adi776borate to be able to merge the PR we need the fix copies, you can either wait for the other PR to be merged or make this PR match the function and add the fix in the other PR, both works but we can't merge this one with a failing test.

> I think we should also update the docs to tell the user that, we expect patchified latents as input.

we use the docstrings as documentation, so it's enough if you just add this in the docstring.

P.S.: Also you will need to fix the conflicts

asomoza:
@bot /style

asomoza (approved):
thanks @adi776borate we only have the fix-copies issue preventing the merge.

adi776borate:
Hey, I have removed the .to() call, I'll update the follow-up PR accordingly.

asomoza:
thanks a lot! also for keeping up with the reviews

adi776borate:
You're welcome! I really enjoyed working on this!

PR #13055, opened 2026-01-29
[wan] fix wan 2.2 when either of the transformers isn't present.

sayakpaul:
# What does this PR do?

Fix https://github.com/huggingface/diffusers/actions/runs/21468140220/job/61834606921?pr=13039#step:6:1381


yiyixuxu (approved):
thanks!

PR #13058, opened 2026-01-30
Add LTX2 Condition Pipeline

### LINKED ISSUE #12926: LTX-2 condition pipeline
https://github.com/huggingface/diffusers/pull/12915 added support for LTX-2. 

## What's supported?
* Single stage T2V and I2V
* Upsampling

Similar to the [LTX condition pipeline](https://huggingface.co/docs/diffusers/main/api/pipelines/ltx_video#diffusers.LTXConditionPipeline), it would be nice to have support for something similar for LTX-2 as well.

Both the [single](https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-pipelines/src/ltx_pipelines/ti2vid_one_stage.py) and [double stage](https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-pipelines/src/ltx_pipelines/ti2vid_two_stages.py) pipelines support passing conditions.

dg845:
# What does this PR do?

This PR adds `LTX2ConditionPipeline`, a pipeline which supports visual conditioning at arbitrary frames for the LTX-2 model ([paper](https://arxiv.org/abs/2601.03233), [code](https://github.com/Lightricks/LTX-2), [weights](https://huggingface.co/Lightricks/LTX-2)), following the [original code](https://github.com/Lightricks/LTX-2). This is an analogue of `LTXConditionPipeline` for LTX-2, as both the original LTX models and LTX-2 support a similar conditioning scheme.

Fixes #12926

## Who can review?

@sayakpaul
@yiyixuxu

### DIFF
--- a/src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
+++ b/src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
@@ -951,7 +951,6 @@ class LTX2ConditionPipeline(DiffusionPipeline, FromSingleFileMixin, LTX2LoraLoad
 
         return latents, conditioning_mask, clean_latents
 
-    # Copied from diffusers.pipelines.ltx2.pipeline_ltx2.LTX2Pipeline.prepare_audio_latents
     def prepare_audio_latents(
         self,
         batch_size: int = 1,


dg845:
Sample first-last-frame-to-video (FLF2V) script:

```python
import torch

from diffusers import LTX2ConditionPipeline
from diffusers.pipelines.ltx2.pipeline_ltx2_condition import LTX2VideoCondition


model_id = "Lightricks/LTX-2"
device = "cuda:0"
dtype = torch.bfloat16
seed = 42

width = 768
height = 512
frame_rate = 24.0

pipe = LTX2ConditionPipeline.from_pretrained(model_id, torch_dtype=dtype)
pipe.vae.enable_tiling()

generator = torch.Generator(device).manual_seed(seed)

prompt = "CG animation style, a small blue bird takes off from the ground, flapping its wings. The bird's feathers are delicate, with a unique pattern on its chest. The background shows a blue sky with white clouds under bright sunshine. The camera follows the bird upward, capturing its flight and the vastness of the sky from a close-up, low-angle perspective."
negative_prompt = "blurry, out of focus, overexposed, underexposed, low contrast, washed out colors, excessive noise, grainy texture, poor lighting, flickering, motion blur, distorted proportions, unnatural skin tones, deformed facial features, asymmetrical face, missing facial features, extra limbs, disfigured hands, wrong hand count, artifacts around text, inconsistent perspective, camera shake, incorrect depth of field, background too sharp, background clutter, distracting reflections, harsh shadows, inconsistent lighting direction, color banding, cartoonish rendering, 3D CGI look, unrealistic materials, uncanny valley effect, incorrect ethnicity, wrong gender, exaggerated expressions, wrong gaze direction, mismatched lip sync, silent or muted audio, distorted voice, robotic voice, echo, background noise, off-sync audio, incorrect dialogue, added dialogue, repetitive speech, jittery movement, awkward pauses, incorrect timing, unnatural transitions, inconsistent framing, tilted camera, flat lighting, inconsistent tone, cinematic oversaturation, stylized filters, or AI artifacts."

first_image = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/flf2v_input_first_frame.png")
last_image = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/flf2v_input_last_frame.png")

first_cond = LTX2VideoCondition(frames=first_image, index=0, strength=1.0)
last_cond = LTX2VideoCondition(frames=last_image, index=-1, strength=1.0)
conditions = [first_cond, last_cond]

video, audio = pipe(
    conditions=conditions,
    prompt=prompt,
    width=width,
    height=height,
    num_frames=121,
    frame_rate=frame_rate,
    num_inference_steps=40,
    guidance_scale=4.0,
    generator=generator,
    output_type="np",
    return_dict=False,
)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
    output_path="ltx2_cond_flf2v.mp4",
)
```

dg845:
Unfortunately the pipeline isn't quite working as of ed52c0d:

Official FLF2V sample:

https://github.com/user-attachments/assets/887918f1-1d18-4b90-9047-4756da467918

Current `diffusers` FLF2V sample (using condition pipeline):

https://github.com/user-attachments/assets/68acf1aa-ae2b-4f99-a710-9e7d5bf9bbe5

Not sure why the video colors are messed up at the first and last frames (where the conditions are), will debug.

dg845:
I think the color issue is now fixed:

https://github.com/user-attachments/assets/32f5c5ff-ccb3-4097-8368-38f0bcdda0ec

dg845:
The condition pipeline also works with the distilled checkpoint:

<details>
    <summary>FLF2V Distilled Script</summary>

```python
import torch

from diffusers import LTX2ConditionPipeline, LTX2LatentUpsamplePipeline
from diffusers.pipelines.ltx2 import LTX2LatentUpsamplerModel
from diffusers.pipelines.ltx2.pipeline_ltx2_condition import LTX2VideoCondition
from diffusers.pipelines.ltx2.utils import DISTILLED_SIGMA_VALUES, STAGE_2_DISTILLED_SIGMA_VALUES


model_id = "rootonchair/LTX-2-19b-distilled"
device = "cuda:0"
dtype = torch.bfloat16
seed = 42

width = 768
height = 512
frame_rate = 24.0

pipe = LTX2ConditionPipeline.from_pretrained(model_id, torch_dtype=dtype)
pipe.vae.enable_tiling()

generator = torch.Generator(device).manual_seed(seed)

prompt = "CG animation style, a small blue bird takes off from the ground, flapping its wings. The bird's feathers are delicate, with a unique pattern on its chest. The background shows a blue sky with white clouds under bright sunshine. The camera follows the bird upward, capturing its flight and the vastness of the sky from a close-up, low-angle perspective."
negative_prompt = "blurry, out of focus, overexposed, underexposed, low contrast, washed out colors, excessive noise, grainy texture, poor lighting, flickering, motion blur, distorted proportions, unnatural skin tones, deformed facial features, asymmetrical face, missing facial features, extra limbs, disfigured hands, wrong hand count, artifacts around text, inconsistent perspective, camera shake, incorrect depth of field, background too sharp, background clutter, distracting reflections, harsh shadows, inconsistent lighting direction, color banding, cartoonish rendering, 3D CGI look, unrealistic materials, uncanny valley effect, incorrect ethnicity, wrong gender, exaggerated expressions, wrong gaze direction, mismatched lip sync, silent or muted audio, distorted voice, robotic voice, echo, background noise, off-sync audio, incorrect dialogue, added dialogue, repetitive speech, jittery movement, awkward pauses, incorrect timing, unnatural transitions, inconsistent framing, tilted camera, flat lighting, inconsistent tone, cinematic oversaturation, stylized filters, or AI artifacts."

first_image = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/flf2v_input_first_frame.png")
last_image = load_image("https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/flf2v_input_last_frame.png")

first_cond = LTX2VideoCondition(frames=first_image, index=0, strength=1.0)
last_cond = LTX2VideoCondition(frames=last_image, index=-1, strength=1.0)
conditions = [first_cond, last_cond]

video_latent, audio_latent = pipe(
    conditions=conditions,
    prompt=prompt,
    width=width,
    height=height,
    num_frames=121,
    frame_rate=frame_rate,
    num_inference_steps=8,
    sigmas=DISTILLED_SIGMA_VALUES,
    guidance_scale=1.0,
    generator=generator,
    output_type="latent",
    return_dict=False,
)

latent_upsampler = LTX2LatentUpsamplerModel.from_pretrained(
    model_id,
    subfolder="latent_upsampler",
    torch_dtype=dtype,
)
upsample_pipe = LTX2LatentUpsamplePipeline(vae=pipe.vae, latent_upsampler=latent_upsampler)
upsample_pipe.enable_model_cpu_offload(device=device)
upscaled_video_latent = upsample_pipe(
    latents=video_latent,
    output_type="latent",
    return_dict=False,
)[0]

video, audio = pipe(
    conditions=conditions,
    latents=upscaled_video_latent,
    audio_latents=audio_latent,
    prompt=prompt,
    width=width * 2,
    height=height * 2,
    num_frames=121,
    frame_rate=frame_rate,
    num_inference_steps=3,
    # noise_scale=STAGE_2_DISTILLED_SIGMA_VALUES[0],
    sigmas=STAGE_2_DISTILLED_SIGMA_VALUES,
    generator=generator,
    guidance_scale=1.0,
    output_type="np",
    return_dict=False,
)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
    output_path="ltx2_cond_flf2v_distilled.mp4",
)
```
</details>

Official FLF2V distilled sample:

https://github.com/user-attachments/assets/72fc2fd6-a688-40cb-a51c-c2b9a2196abd

`diffusers` FLF2V distilled sample:

https://github.com/user-attachments/assets/fc09d9dd-e766-4b94-a148-bccd025f4d91

sayakpaul:
Thanks for the work, @dg845! Looking at the outputs, is there a discrepancy in the resolutions? The main bird subject seems a bit compressed to me in our implementation.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+
+EXAMPLE_DOC_STRING = """
+    Examples:
+        ```py
+        >>> import torch
+        >>> from diffusers import LTX2Pipeline
+        >>> from diffusers.pipelines.ltx2.export_utils import encode_video
+        >>> from diffusers.utils import load_image
+
+        >>> pipe = LTX2ImageToVideoPipeline.from_pretrained("Lightricks/LTX-2", torch_dtype=torch.bfloat16)
```
sayakpaul: I think these pipeline classes need to be updated?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+        image (`PIL.Image.Image` or `List[PIL.Image.Image]` or `np.ndarray` or `torch.Tensor`):
+            The image (or video) to condition the video on. Accepts any type that can be handled by
+            VideoProcessor.preprocess_video.
+        index (`int`, defaults to `0`):
+            The index at which the image or video will conditionally affect the video generation.
+        strength (`float`, defaults to `1.0`):
+            The strength of the conditioning effect. A value of `1.0` means the conditioning effect is fully applied.
+    """
+
+    image: Union[PIL.Image.Image, List[PIL.Image.Image], np.ndarray, torch.Tensor]
```
sayakpaul: I am guessing this can be a list of images (sequence of frames denoting a video clip)? If so, would `frames` be a better name?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+
+        return prompt_embeds, prompt_attention_mask, negative_prompt_embeds, negative_prompt_attention_mask
+
+    def check_inputs(
+        self,
+        conditions,
+        image,
+        video,
+        cond_index,
+        strength,
```
sayakpaul: Should we only accept `condition` as an input to simplify logic? If so, I think `check_inputs()` would then only accept `condition` (which could be a single item or a list of `condition`s). WDYT?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+        Returns:
+            int: updated sequence length
+        """
+        scale_factor = self.vae_temporal_compression_ratio
+        num_frames = min(sequence_num_frames, target_num_frames - start_frame)
+        # Trim down to a multiple of temporal_scale_factor frames plus 1
+        num_frames = (num_frames - 1) // scale_factor * scale_factor + 1
+        return num_frames
+
+    def latent_idx_from_index(self, frame_idx: int, index_type: str = "latent") -> int:
```
sayakpaul: It's currently just a single type. I guess we can just do it inside the caller instead of having a separate function?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+        for i, condition in enumerate(conditions):
+            if isinstance(condition.image, PIL.Image.Image):
+                # Single image, convert to List[PIL.Image.Image]
+                video_like_cond = [condition.image]
+            elif isinstance(condition.image, np.ndarray) and condition.image.ndim == 3:
+                # Image-like ndarray of shape (H, W, C), insert frame dim in first axis
+                video_like_cond = np.expand_dims(condition.image, axis=0)
+            elif isinstance(condition.image, torch.Tensor) and condition.image.ndim == 3:
+                # Image-like tensor of shape (C, H, W), insert frame dim in first dim
+                video_like_cond = condition.image.unsqueeze(0)
```
sayakpaul: Do you think the `video_processor` should handle this logic?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+            logger.warning(
+                f"{self.__class__.__name__} does not support using a list of generators. The first generator in the"
+                f" list will be used for all (pseudo-)random operations."
+            )
+            generator = generator[0]
+
+        condition_frames, condition_strengths, condition_indices = self.preprocess_conditions(
+            conditions, height, width, num_frames, device=device
+        )
+        # TODO: should we first concatenate all of the condition tensors and encode them all together? The advantage
```
sayakpaul: Good question. We could take a flag from the user to determine this?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+            condition_latent = self._normalize_latents(
+                condition_latent, self.vae.latents_mean, self.vae.latents_std
+            ).to(device=device, dtype=dtype)
+            condition_latent = self._pack_latents(
+                condition_latent, self.transformer_spatial_patch_size, self.transformer_temporal_patch_size
+            )
+            condition_latents.append(condition_latent)
+
+        # NOTE: following the I2V pipeline, we return a conditioning mask. The original LTX 2 code uses a denoising
+        # mask, which is the inverse of the conditioning mask (`denoise_mask = 1 - conditioning_mask`)
```
sayakpaul, line 935: Yes good point. This would influence how this mask influences the rest of the process. We must make sure it's not accidentally treated as a denoising mask.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+        height: int = 512,
+        width: int = 768,
+        num_frames: int = 121,
+        frame_rate: float = 24.0,
+        num_inference_steps: int = 40,
+        sigmas: Optional[List[float]] = None,
+        timesteps: List[int] = None,
+        guidance_scale: float = 4.0,
+        guidance_rescale: float = 0.0,
+        noise_scale: Optional[float] = None,
```
sayakpaul: Should probably be documented.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+        num_videos_per_prompt: Optional[int] = 1,
+        generator: Optional[Union[torch.Generator, List[torch.Generator]]] = None,
+        latents: Optional[torch.Tensor] = None,
+        audio_latents: Optional[torch.Tensor] = None,
+        prompt_embeds: Optional[torch.Tensor] = None,
+        prompt_attention_mask: Optional[torch.Tensor] = None,
+        negative_prompt_embeds: Optional[torch.Tensor] = None,
+        negative_prompt_attention_mask: Optional[torch.Tensor] = None,
+        decode_timestep: Union[float, List[float]] = 0.0,
+        decode_noise_scale: Optional[Union[float, List[float]]] = None,
```
sayakpaul: A bit unrelated: I haven't seen the use of `decode_noise_scale` in LTX-2. If that's the case indeed, WDYT of cleaning it off from the LTX-2 pipelines?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1538 @@
...
+                            noise_pred_audio, noise_pred_audio_text, guidance_rescale=self.guidance_rescale
+                        )
+
+                # NOTE: use only the first chunk of conditioning mask in case it is duplicated for CFG
+                bsz = noise_pred_video.size(0)
+                sigma = self.scheduler.sigmas[i]
+                # Convert the noise_pred_video velocity model prediction into a sample (x0) prediction
+                denoised_sample = latents - noise_pred_video * sigma
+                # Apply the (packed) conditioning mask to the denoised (x0) sample, which will blend the conditions
+                # with the denoised sample according to the conditioning strength (a strength of 1.0 means we fully
```
sayakpaul: > a strength of 1.0 means we fully

(nit): However, from the code, it's not clear how this strength is incorporated. Consider expanding on the comment a little bit.

sayakpaul (commented):
Thanks for this work! It looks very clean and the implementation also seems faithful to the original one. I left some comments, LMK if they're clear.

Additionally, I would like to see some extensions being used in the pipeline. For example, multiple conditions (multiple images) with different indices. Would it be possible?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: Yeah, the idea is that it can either be a single frame or a contiguous sequence of frames.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: The current logic follows `LTXConditionPipeline` in also accepting `image` and `video` arguments (and therefore also needing `cond_index` and `strength` arguments). However, I agree that only accepting `conditions` would probably be better because it it less ambiguous. (One reservation I have is that you would have to import `LTX2VideoCondition` every time, but maybe that's not a big deal.)

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: I was also thinking supporting a `"data"` `index_type`, where the `index` is interpreted in data (pixel) space rather than latent space, as `LTXConditionPipeline` appears to support, but I don't quite understand the `frame_index` logic in `LTXConditionPipeline` yet. My current understanding is that the original LTX-2 code only supports latent indices (but I might be mistaken).

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: Yeah, I think ideally `VideoProcessor` should handle this logic. `VideoProcessor.preprocess_video` only supports video-like inputs, but my impression is that treating images as single-frame videos is a common enough use case in video models that we should support it in `preprocess_video` (for example, by accepting 3D `np.ndarray`s and `torch.Tensor`s as done here).

Ambiguity could be an issue but I think that this isn't too big of a problem because users can manually specify tensors in an unambiguous way (e.g. by manually expanding and possibly reshaping into an appropriate 5D tensor) if they need to.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: `noise_scale` is documented here:

https://github.com/huggingface/diffusers/blob/70dff169961bd8001b7bee2182a7defc898ab12d/src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py#L1117-L1120

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: My understanding is that `decode_noise_scale` is used if the video VAE supports timestep conditioning:

https://github.com/huggingface/diffusers/blob/70dff169961bd8001b7bee2182a7defc898ab12d/src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py#L1507-L1516

The [LTX-2 VAE](https://huggingface.co/Lightricks/LTX-2/blob/main/vae/config.json) currently uses `timestep_conditioning=False`. It's unclear to me whether the LTX-2 code intends to support it, as the video decoder model still accepts a `timestep_conditioning` argument:

https://github.com/Lightricks/LTX-2/blob/4f410820b198e05074a1e92de793e3b59e9ab5a0/packages/ltx-core/src/ltx_core/model/video_vae/video_vae.py#L432

but the VAE decoding code doesn't support a `timestep` argument that would be necessary if `timestep_conditioning=True`:

https://github.com/Lightricks/LTX-2/blob/4f410820b198e05074a1e92de793e3b59e9ab5a0/packages/ltx-core/src/ltx_core/model/video_vae/video_vae.py#L813

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: The conditioning strengths are used here:

https://github.com/huggingface/diffusers/blob/70dff169961bd8001b7bee2182a7defc898ab12d/src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py#L866-L869

Perhaps it would more clear to say that the `conditioning_mask` itself specifies the strength with which the conditions are applied?

dg845:
> Additionally, I would like to see some extensions being used in the pipeline. For example, multiple conditions (multiple images) with different indices. Would it be possible?

The FLF2V script in https://github.com/huggingface/diffusers/pull/13058#issuecomment-3844904224 gives an example using multiple conditions that's not possible in `LTX2ImageToVideoPipeline`: here we condition on a last frame as well as an initial frame.

dg845:
> Looking at the outputs, is there a discrepancy in the resolutions? The main bird subject seems a bit compressed to me in our implementation.

I believe the discrepancy is because the original LTX-2 center crops images:

https://github.com/Lightricks/LTX-2/blob/4f410820b198e05074a1e92de793e3b59e9ab5a0/packages/ltx-pipelines/src/ltx_pipelines/utils/media_io.py#L88

but `VideoProcessor.preprocess_video` doesn't expose a `resize_mode` like `VaeImageProcessor.preprocess` does, so we end up using the `"default"` `resize_mode`, which instead resizes the image to fit in a non-aspect-ratio-preserving way.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: I have opened a separate PR to address this and https://github.com/huggingface/diffusers/pull/13058#issuecomment-3850585412: #13084.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
sayakpaul: > (One reservation I have is that you would have to import LTX2VideoCondition every time, but maybe that's not a big deal.)

That's just one-time import no? Then one could create the conditions either as a list of `LTX2VideoCondition` or just `LTX2VideoCondition` (in case of single condition).

If so, I think that's fine?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
sayakpaul: > but I don't quite understand the frame_index logic in LTXConditionPipeline yet. My current understanding is that the original LTX-2 code only supports latent indices (but I might be mistaken).

If so, let's keep it inline then?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
sayakpaul: Yes, I am on the same side. Do you think it could make sense to remote this logic in a separate PR then? It will simplify the code a bit.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
sayakpaul: Yes!

sayakpaul:
> but VideoProcessor.preprocess_video doesn't expose a resize_mode like VaeImageProcessor.preprocess does, so we end up using the "default" resize_mode, which instead resizes the image to fit in a non-aspect-ratio-preserving way.

Should we try to implement this then? Perhaps club it inside https://github.com/huggingface/diffusers/pull/13084. I think that the results are better with a center crop. WDYT?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1484 @@
...
+from dataclasses import dataclass
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+import numpy as np
+import PIL.Image
+import torch
+from transformers import Gemma3ForConditionalGeneration, GemmaTokenizer, GemmaTokenizerFast
+
+from ...callbacks import MultiPipelineCallbacks, PipelineCallback
+from ...loaders import FromSingleFileMixin, LTXVideoLoraLoaderMixin
```
asomoza:
```suggestion
from ...loaders import FromSingleFileMixin, LTX2LoraLoaderMixin
```

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1484 @@
...
+    std_text = noise_pred_text.std(dim=list(range(1, noise_pred_text.ndim)), keepdim=True)
+    std_cfg = noise_cfg.std(dim=list(range(1, noise_cfg.ndim)), keepdim=True)
+    # rescale the results from guidance (fixes overexposure)
+    noise_pred_rescaled = noise_cfg * (std_text / std_cfg)
+    # mix with the original results from guidance by factor guidance_rescale to avoid "plain looking" images
+    noise_cfg = guidance_rescale * noise_pred_rescaled + (1 - guidance_rescale) * noise_cfg
+    return noise_cfg
+
+
+class LTX2ConditionPipeline(DiffusionPipeline, FromSingleFileMixin, LTXVideoLoraLoaderMixin):
```
asomoza:
```suggestion
class LTX2ConditionPipeline(DiffusionPipeline, FromSingleFileMixin, LTX2LoraLoaderMixin):
```

asomoza (commented):
thanks, proposed the same change as this [PR](https://github.com/huggingface/diffusers/pull/13144)

dg845:
`diffusers` FLF2V distilled sample with center cropping like the original LTX-2 code:

https://github.com/user-attachments/assets/ed40baa8-4372-416d-8645-e91bc05f6b27

rodjjo:
For the first and the last frame i tested and it looks OK.

For intermediate frames I think they need to have the index (multiple of 8) + 1
9, 17, 25 etc.

```python
            if (index - 1) % 8 != 0:
                index = index + (8 - (index - 1) % 8)
            index -= 1  # it counts from 0 to num_frames - 1
            if index < 0:
                index = 0
```

It could go on docs and maybe a warning log on the code.

rodjjo:
I added a wrong comment that I removed.

I was testing this again. I get a smooth result when I remove the frames related to the condition index.

```python
    frames2drop = []
    for f in result:
        f = f.index * 8
        for i in range(1, 8):
            frames2drop.append(f - i)
```
^ by removing these frames from the final video, it removes the pauses caused by the condition frames...
The condition frames pauses the video for 8 frames, so I drop 7

Nice work, congrats!

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1489 @@
...
+        # Packed latents of shape [B, S, D] (S is the effective video sequence length, D is the effective feature dimensions)
+        # are unpacked and reshaped into a video tensor of shape [B, C, F, H, W]. This is the inverse operation of
+        # what happens in the `_pack_latents` method.
+        batch_size = latents.size(0)
+        latents = latents.reshape(batch_size, num_frames, height, width, -1, patch_size_t, patch_size, patch_size)
+        latents = latents.permute(0, 4, 1, 5, 2, 6, 3, 7).flatten(6, 7).flatten(4, 5).flatten(2, 3)
+        return latents
+
+    @staticmethod
+    def _normalize_latents(
```
sayakpaul, line 639: Can this be supplemented with a "# Copied from ..." as well?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1490 @@
...
+            scheduler=scheduler,
+        )
+
+        self.vae_spatial_compression_ratio = (
+            self.vae.spatial_compression_ratio if getattr(self, "vae", None) is not None else 32
+        )
+        self.vae_temporal_compression_ratio = (
+            self.vae.temporal_compression_ratio if getattr(self, "vae", None) is not None else 8
+        )
+        # TODO: check whether the MEL compression ratio logic here is corrct
```
yiyixuxu: can we remove this TODO now?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1490 @@
...
+            logger.warning(
+                f"{self.__class__.__name__} does not support using a list of generators. The first generator in the"
+                f" list will be used for all (pseudo-)random operations."
+            )
+            generator = generator[0]
+
+        condition_frames, condition_strengths, condition_indices = self.preprocess_conditions(
+            conditions, height, width, num_frames, device=device
+        )
+        # TODO: should we first concatenate all of the condition tensors and encode them all together? The advantage
```
yiyixuxu: I think we should concatenate and encode once. The memory concern is non-issue if they can have tiling enabled, we can use that by default in our examples

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1490 @@
...
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        latent_height = height // self.vae_spatial_compression_ratio
+        latent_width = width // self.vae_spatial_compression_ratio
+        latent_num_frames = (num_frames - 1) // self.vae_temporal_compression_ratio + 1
+
+        shape = (batch_size, num_channels_latents, latent_num_frames, latent_height, latent_width)
+        mask_shape = (batch_size, 1, latent_num_frames, latent_height, latent_width)
+
+        if latents is not None:
+            # Latents is either shape [B, F, C, H, W] or [B, seq_len, hidden_dim]
```
yiyixuxu: any reason we need to accept both?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
```diff
@@ -0,0 +1,1490 @@
...
+
+        duration_s = num_frames / frame_rate
+        audio_latents_per_second = (
+            self.audio_sampling_rate / self.audio_hop_length / float(self.audio_vae_temporal_compression_ratio)
+        )
+        audio_num_frames = round(duration_s * audio_latents_per_second)
+        if audio_latents is not None:
+            if audio_latents.ndim == 4:
+                logger.info(
+                    "Got audio_latents of shape [batch_size, num_channels, audio_length, mel_bins], `audio_num_frames` will be inferred."
```
yiyixuxu: same here, why do we accept 2 shapes?

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: I think my original claim that (1) concatenating all of the conditions, then VAE encoding all at once and (2) VAE encoding the conditions separately are equivalent is wrong. This is because the LTX-2 video VAE encoder uses 3D spatiotemporal convolutions, so in general the latents of a given frame depend on the frames around it. So we can't concatenate in the frame dimension because this would cause unrelated conditions to influence each other when calculating the latents.

We also cannot concatenate in the batch dimension, because in general the conditions are different shapes (in particular, the number of frames may be different), and we want to encode video conditions together due to the use of spatiotemporal convolutions.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: The reason we accept both shapes here is because we want to support unpacked 5D latent inputs for two-stage inference. Following `LTXPipeline`, we support 3D packed latents from the `latents` argument to `__call__`, but `LTX2LatentUpsamplePipeline` returns 5D unpacked latents, so we want to be able to supply the output of the upsampling pipeline directly to Stage 2 inference:

```python
# Latent upsampling; video_latents_upsampled is 5D
video_latents_upsampled = upsample_pipe(..., output_type="latent", return_dict=False)[0]

# Stage 2 inference
video, audio = pipe(
    latents=video_latents_upsampled,
    audio_latents=audio_latents,
    ...,
)
```

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
dg845: See https://github.com/huggingface/diffusers/pull/13058#discussion_r2886935261.

src/diffusers/pipelines/ltx2/pipeline_ltx2_condition.py
yiyixuxu: I think we can just need to support 5D here, no?

dg845: Do we usually assume that the `latents` argument to `__call__` is 3D? I see that for example `LTXPipeline.prepare_latents` and `FluxPipeline.prepare_latents` both assume that the supplied `latents` are 3D, so that's why I'm supporting 3D latents here as well.

yiyixuxu: ohh, let's just support 5D here. 
I understand why you want to be consistent with LTXPipeline and FluxPipeline, but in this case the main use case is two-stage and stage 1 directly outputs 5D latents,

PR #13066, opened 2026-02-02
Cosmos Transfer2.5 inference pipeline: general/{seg, depth, blur, edge}

miguelmartin75:
# What does this PR do?

This PR introduces Cosmos Transfer2.5 inference pipeline, which extends the existing code in transformer_cosmos.py and introduces a new controlnet class for cosmos. The conversion script is updated to convert the checkpoints too.

I've intentionally split the controlnet from the base predict model to match the rest of the diffusers codebase. To do this, I have had to duplicate some layers/weights from the base model (relating to the patch & timestep embeddings), but I believe SD3 does this.

Similar to predict2.5, I have added documentation and unit tests.

Additional PRs will be submitted for the following features (in order of priority):
1. Auto-regressive inference support, currently inference can only be applied to a fix number of frames. In cosmos-transfer2.5 AR inference is performed.
2. Additional transfer2.5 variants:
   - multi-control (multiple controlnets at once)
   - auto/multiview
3. Image reference

In addition, unfortunately, the guardrails safety model is too aggressive: it currently flags "not safe" for the examples we have on cosmos-transfer2.5 (e.g. edge example for 93 frames is flagged). This guardrail model needs to be updated, but this work is ~orthogonal of this PR.

## Who can review?

Core library:

### DIFF
--- a/src/diffusers/utils/dummy_pt_objects.py
+++ b/src/diffusers/utils/dummy_pt_objects.py
@@ -896,6 +896,21 @@ class ControlNetXSAdapter(metaclass=DummyObject):
         requires_backends(cls, ["torch"])
 
 
+class CosmosControlNetModel(metaclass=DummyObject):
+    _backends = ["torch"]
+
+    def __init__(self, *args, **kwargs):
+        requires_backends(self, ["torch"])
+
+    @classmethod
+    def from_config(cls, *args, **kwargs):
+        requires_backends(cls, ["torch"])
+
+    @classmethod
+    def from_pretrained(cls, *args, **kwargs):
+        requires_backends(cls, ["torch"])
+
+
 class CosmosTransformer3DModel(metaclass=DummyObject):
     _backends = ["torch"]
 


scripts/convert_cosmos_to_diffusers.py
```diff
@@ -78,12 +78,61 @@
...
+
+python scripts/convert_cosmos_to_diffusers.py \
+    --transformer_type Cosmos-2.5-Transfer-General-2B \
+    --transformer_ckpt_path $transformer_ckpt_path \
+    --vae_type wan2.1 \
+    --output_path converted/transfer/2b/general/blur \
+    --save_pipeline
+
+# seg
+transformer_ckpt_path=~/.cache/huggingface/hub/models--nvidia--Cosmos-Transfer2.5-2B/snapshots/eb5325b77d358944da58a690157dd2b8071bbf85/general/seg/5136ef49-6d8d-42e8-8abf-7dac722a304a_ema_bf16.pt
```
yiyixuxu, line 127: ohh does each variant come with its own base transformer?

in diffusers we typically split controlnet from the base model is so that user can mix and match, it this something possible with cosmos?

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -0,0 +1,936 @@
...
+        return encoder_output.latent_dist.sample(generator)
+    elif hasattr(encoder_output, "latent_dist") and sample_mode == "argmax":
+        return encoder_output.latent_dist.mode()
+    elif hasattr(encoder_output, "latents"):
+        return encoder_output.latents
+    else:
+        raise AttributeError("Could not access latents of provided encoder_output")
+
+
+def transfer2_5_forward(
```
yiyixuxu: Can we inline this inside the `__call__` method? We typically only create separate methods for operations users might need to run standalone to pre-compute things (like `encode_prompt`, `encode_video`, etc.). It's also easier to read when you don't have to jump around the file.

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -0,0 +1,936 @@
...
+    _exclude_from_cpu_offload = ["safety_checker"]
+
+    def __init__(
+        self,
+        text_encoder: Qwen2_5_VLForConditionalGeneration,
+        tokenizer: AutoTokenizer,
+        transformer: CosmosTransformer3DModel,
+        vae: AutoencoderKLWan,
+        scheduler: UniPCMultistepScheduler,
+        controlnet: CosmosControlNetModel,
```
yiyixuxu: is controlnet optional here?

yiyixuxu (commented):
Thanks for the PR! The overall structure looks good. I left some minor comments.

One question before I can review further: Are the base transformer weights the same across the different control variants?

This helps us understand whether splitting the controlnet from the transformer makes sense (i.e., can users mix and match?), and also helps me understand whether the controlnet is required for this pipeline etc

scripts/convert_cosmos_to_diffusers.py
miguelmartin75: Each variant should have the same weights as the base transformer, I will double check this, but I split out the controlnet and save the pipeline (saves base transformer + controlnet), such that the pipeline can be loaded directly from a model_id/revision. 

I will look into only loading the controlnet from the converted script.

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
miguelmartin75: Yes. I will change the typehint

scripts/convert_cosmos_to_diffusers.py
yiyixuxu: ohh
I think it's easier just to have 4 repos for 4 controlnets -
so that you can load like this 

```py
controlnet = AutoModel.from_pretrained("nvidia/Cosmos-Controlnet-Edge", ...)
pipe = DiffusionPipeline.from_pretrained("nvidia/Cosmos-Transfer2.5-2B", controlnet=controlnet, ...)
```

miguelmartin75: I will put each controlnet in a different revision, e.g. the code look something like this:

```
controlnet = AutoModel.from_pretrained("nvidia/Cosmos-Transfer2.5-2B", revision="controlnet/general/edge")
pipe = DiffusionPipeline.from_pretrained("nvidia/Cosmos-Transfer2.5-2B", controlnet=controlnet, revision_id="diffusers/general")
```

---

also, as a separate PR, I plan to add multiple control (controlnet) support, so the code would look something like:

```
c1 = AutoModel.from_pretrained("nvidia/Cosmos-Transfer2.5-2B", revision="diffusers/controlnet/general/edge")
c2 = AutoModel.from_pretrained("nvidia/Cosmos-Transfer2.5-2B", revision="diffusers/controlnet/general/depth")

pipe = DiffusionPipeline.from_pretrained("nvidia/Cosmos-Transfer2.5-2B", controlnets=[c1, c2], revision_id="diffusers/general")
```

Is this the standard practice for diffusers?

---

Similar to this breaking change, for auto-regressive the `__call__` will have `max_frames` instead of `num_frames` to generate a maximum number of frames defaulting to None (# frames in input). `num_frames` will be renamed to num_frames_per_chunk. Again this will be a separate PR with breaking change.

---

Is multiple PRs preferred, or should I merge these changes into this PR?

yiyixuxu: On multiple PRs: 
Yes, please keep them as separate PRs, so we can move quickly 

for the multiple controlnet:
we don't really use revision very much anymore, the standard practice is to create seperate repos for them, but if you strongly prefer to keep a single repo, that works as well.

miguelmartin75: I've updated to use revision to be consistent with predict2.5, this may change in the future. Thanks for clarifying on multiple PRs.

miguelmartin75:
Addressed your comment about `transfer2_5_forward` + updated the example code

>Are the base transformer weights the same across the different control variants? ... can users mix and match?

Yes, mix & matching controlnets is possible, but only if an image context reference is **not included**([see here](https://github.com/nvidia-cosmos/cosmos-transfer2.5/blob/main/docs/inference_image.md), including an image reference is **not currently supported** in this PR). Additionally, including multiple controlnets ["multicontrol"](https://github.com/nvidia-cosmos/cosmos-transfer2.5/blob/main/docs/inference.md#multi-control) will be possible (any base transformer can be used; cosmos-transfer2.5 always picks "edge"), but I will need to submit a separate PR for this. Note, multicontrol does not support an image reference. 

To be more specific, the base transformer weights are almost the same. The difference lies in the weights of the cross attention layers for an image reference ([see here](https://github.com/nvidia-cosmos/cosmos-transfer2.5/blob/main/docs/inference_image.md)), i.e. [attn2 in diffusers-land](https://github.com/miguelmartin75/diffusers/blob/ef332d211b889be0d6723fc80651293a7d5e20e0/src/diffusers/models/transformers/transformer_cosmos.py#L365-L374) for [these layers](https://github.com/miguelmartin75/diffusers/blob/ef332d211b889be0d6723fc80651293a7d5e20e0/src/diffusers/models/transformers/transformer_cosmos.py#L310-L312) for all blocks in the base transformer. Without an image reference, all base transformers are functionally same, in this case the img_context tensor is `torch.zeros`; I also qualitatively verified all pairs of base transformer + controlnet as a sanity check and it looks like they output the same results. 

I will need to document this when I have a PR up for image reference feature, (3) in my description

src/diffusers/models/transformers/transformer_cosmos.py
```diff
@@ -204,14 +203,132 @@ def __call__(
             query, key, value, attn_mask=attention_mask, dropout_p=0.0, is_causal=False
         )
         hidden_states = hidden_states.transpose(1, 2).flatten(2, 3).type_as(query)
+        return hidden_states
 
-        # 6. Output projection
+    def __call__(
```
yiyixuxu: can we not have this change?

src/diffusers/models/transformers/transformer_cosmos.py
```diff
@@ -204,14 +203,132 @@ def __call__(
...
+            attention_mask=attention_mask,
+            image_rotary_emb=image_rotary_emb,
+        )
         hidden_states = attn.to_out[0](hidden_states)
         hidden_states = attn.to_out[1](hidden_states)
 
         return hidden_states
 
 
+class CosmosAttnProcessor2_5(CosmosAttnProcessor2_0):
```
yiyixuxu: can we avoid inheriting from `CosmosAttnProcessor2_0`? they should be self-contained classes 
also can you remove the compute_attn methods so the logic lives directly in `__call__`

src/diffusers/models/transformers/transformer_cosmos.py
```diff
@@ -204,14 +203,132 @@ def __call__(
...
+
+        q_img = attn.q_img_norm(q_img)
+        k_img = attn.k_img_norm(k_img)
+
+        q_img_idx = q_img.size(3)
+        k_img_idx = k_img.size(3)
+        v_img_idx = v_img.size(3)
+        k_img = k_img.repeat_interleave(q_img_idx // k_img_idx, dim=3)
+        v_img = v_img.repeat_interleave(q_img_idx // v_img_idx, dim=3)
+        img_out = torch.nn.functional.scaled_dot_product_attention(
```
yiyixuxu: does it make sense to use `dispatch_attention_fn`? https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_flux.py#L256

yiyixuxu (commented):
thanks, I left a few more comments

src/diffusers/models/transformers/transformer_cosmos.py
miguelmartin75: thanks, did not know this existed

src/diffusers/models/transformers/transformer_cosmos.py
```diff
@@ -240,30 +361,59 @@ def __init__(
...
+            )
 
         self.norm3 = CosmosAdaLayerNormZero(in_features=hidden_size, hidden_features=adaln_lora_dim)
         self.ff = FeedForward(hidden_size, mult=mlp_ratio, activation_fn="gelu", bias=out_bias)
 
+        # NOTE: zero conv for CosmosControlNet
+        self.before_proj = None
+        self.after_proj = None
+        if before_proj:
+            # TODO: check hint_dim in i4
```
miguelmartin75: TODO remove

src/diffusers/models/transformers/transformer_cosmos.py
miguelmartin75: applied to both CosmosAttnProcessor2_0 and CosmosAttnProcessor2_5

yiyixuxu (approved):
thanks!

sayakpaul:
@miguelmartin75 would it make sense to update https://huggingface.co/docs/diffusers/main/en/api/pipelines/cosmos docs as well with transfer?

miguelmartin75:
Yes, thanks for the call out on that @sayakpaul - maybe I can do this in https://github.com/huggingface/diffusers/pull/13114

jwpaq:
multi-control (multiple controlnets at once) is supported now ? @miguelmartin75

miguelmartin75:
No, only one controlnet is supported at once in this diffusers port currently. 

Here's relevant code pointers incase you can contribute:
1. https://github.com/nvidia-cosmos/cosmos-transfer2.5/blob/main/cosmos_transfer2/_src/transfer2/networks/minimal_v4_lvg_dit_control_vace.py#L948
2. https://github.com/nvidia-cosmos/cosmos-transfer2.5/blob/main/cosmos_transfer2/_src/transfer2/networks/minimal_v4_lvg_dit_control_vace.py#L987
3. https://github.com/nvidia-cosmos/cosmos-transfer2.5/blob/main/cosmos_transfer2/_src/transfer2/networks/minimal_v4_lvg_dit_control_vace.py#L1084-L1088

It should be possible to support this in diffusers by:
a. Running the forward pass on the set of controlnets you want (independently) 
b. Combining their output tensor similar to codepointer 3 above (i.e. `control_B_T_H_W_D_sum`)

I haven't tested this. There may be details I am missing.

PR #13072, opened 2026-02-03
[modular] change the template modular pipeline card

sayakpaul:
# What does this PR do?

New card: https://huggingface.co/sayakpaul/new-template

### DIFF
--- a/src/diffusers/modular_pipelines/modular_pipeline_utils.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline_utils.py
@@ -972,7 +972,7 @@ def generate_modular_model_card_content(blocks) -> Dict[str, Any]:
     blocks_desc_parts = []
     sub_blocks = getattr(blocks, "sub_blocks", None) or {}
     if sub_blocks:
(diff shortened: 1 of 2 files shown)


src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -1734,6 +1735,159 @@ def from_pretrained(
         )
         return pipeline
 
+    def _generate_modular_model_card_content(self) -> Dict[str, Any]:
```
yiyixuxu: I think we can move this as a function in` modular_pipeline_utils` so the pipeline itself is less crowded, like https://github.com/huggingface/diffusers/blob/main/src/diffusers/modular_pipelines/modular_pipeline_utils.py#L865
 
```py
def generate_modular_model_card_conten(blocks):
    ...
````

sayakpaul:
@yiyixuxu thanks! 

I also added tests to ensure the structure doesn't break. Will merge after the tests are green.

sayakpaul:
Failing tests are unrelated.

PR #13075, opened 2026-02-03
[Context Parallel] fix: support PyTorch < 2.6 in distributed gather

### LINKED ISSUE #13074: Distributed gather crashes on PyTorch < 2.6
### Describe the bug

AttributeError: module 'torch' has no attribute 'accelerator' when running distributed gather on PyTorch versions < 2.6.

This error happens because `gather_size_by_comm` in `src/diffusers/models/_modeling_parallel.py` uses `torch.accelerator.current_accelerator()`, which only exists in PyTorch 2.6+. Diffusers officially supports PyTorch 2.1+, so this causes a crash on versions 2.1–2.5 with `AttributeError: module 'torch' has no attribute 'accelerator'`. 

### Reproduction

Since this is a utility function; it can be triggered directly with a minimal distributed setup:

```py
import torch.distributed as dist
from diffusers.models._modeling_parallel import gather_size_by_comm

dist.init_process_group(
    backend="gloo",
    init_method="file:///tmp/pg",
    rank=0,
    world_size=1,
)

gather_size_by_comm(1, dist.group.WORLD)

```

### Logs

```shell
[rank0]: Traceback (most recent call last):
[rank0]:   File "/home/aja/diffusers/test.py", line 11, in <module>
[rank0]:     gather_size_by_comm(1, dist.group.WORLD)
[rank0]:   File "/home/aja/diffusers/src/diffusers/models/_modeling_parallel.py", line 293, in gather_size_by_comm
[rank0]:     gather_device = "cpu" if "cpu" in comm_backends else torch.accelerator.current_accelerator()
[rank0]:                                                          ^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/aja/diffusers/.venv/lib/python3.11/site-packages/torch/__init__.py", line 2216, in __getattr__
[rank0]:     raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
[rank0]: AttributeError: module 'torch' has no attribute 'accelerator'
```

### System Info

diffuser : `0.37.0.dev0`
torch : `2.4.0`
python : `3.11`
system: Linux 

### Who can help?

@sayakpaul @DN6

akshatvishu:
# What does this PR do?
Replace unsafe `torch.accelerator` usage with an `hasattr` guard for backward compatibility.

Fixes #13074 

Original library don't face this issue as their minimum supported torch version is 2.7 :
https://github.com/vipshop/cache-dit/blob/main/pyproject.toml#L13

## Who can review?

@sayakpaul  @DN6

### DIFF
--- a/src/diffusers/models/_modeling_parallel.py
+++ b/src/diffusers/models/_modeling_parallel.py
@@ -294,14 +294,10 @@ def gather_size_by_comm(size: int, group: dist.ProcessGroup) -> List[int]:
     if "cpu" in comm_backends:
         gather_device = "cpu"
     else:
-        gather_device = get_device()
+        gather_device = torch.device(get_device())
 
-        if gather_device == "cpu":
-            raise RuntimeError(
-                "Ulysses Anything Attention (UAA) requires an accelerator "
-                "(CUDA, NPU, XPU, etc.) to perform distributed gathering, "
-                "but only CPU was detected."
-            )
+        if gather_device.type == "cpu":
+            raise RuntimeError("No suitable accelerator found.")
 
     gathered_sizes = [torch.empty((1,), device=gather_device, dtype=torch.int64) for _ in range(world_size)]
     dist.all_gather(


src/diffusers/models/_modeling_parallel.py
```diff
@@ -290,7 +290,15 @@ def gather_size_by_comm(size: int, group: dist.ProcessGroup) -> List[int]:
...
     # NOTE: e.g., dist.init_process_group(backend="cpu:gloo,cuda:nccl")
-    gather_device = "cpu" if "cpu" in comm_backends else torch.accelerator.current_accelerator()
+    if "cpu" in comm_backends:
+        gather_device = "cpu"
+    elif hasattr(torch, "accelerator"):
+        acc = torch.accelerator.current_accelerator()
+        # Fall back to CUDA when no accelerator is returned.
+        gather_device = acc if acc is not None else torch.device("cuda")
+    else:
+        gather_device = torch.device("cuda")
```
sayakpaul: I don't think it's a sensible fallback option for other accelerators. It could be better to raise an error instead.

Or we could utilize the `torch_device` from here:
https://github.com/huggingface/diffusers/blob/6d4fc6baa06f4dc8a4ff7d6a011005dceb8d5af8/src/diffusers/utils/torch_utils.py#L336

akshatvishu: Thanks for the review !  I’ve updated the code to use `get_device()` and raise a clear `RuntimeError` when an accelerator is required but only CPU is detected.

src/diffusers/models/_modeling_parallel.py
```diff
@@ -290,7 +291,18 @@ def gather_size_by_comm(size: int, group: dist.ProcessGroup) -> List[int]:
     # HACK: Use Gloo backend for all_gather to avoid H2D and D2H overhead
     comm_backends = str(dist.get_backend(group=group))
     # NOTE: e.g., dist.init_process_group(backend="cpu:gloo,cuda:nccl")
-    gather_device = "cpu" if "cpu" in comm_backends else torch.accelerator.current_accelerator()
+    if "cpu" in comm_backends:
+        gather_device = "cpu"
+    else:
+        gather_device = get_device()
```
sayakpaul: `torch.accelerator.current_accelerator()` returns a `torch.device` type (as per the [docs](https://docs.pytorch.org/docs/stable/generated/torch.accelerator.current_accelerator.html)). So this should be:
```suggestion
        gather_device = torch.device(get_device())
```

src/diffusers/models/_modeling_parallel.py
```diff
@@ -290,7 +291,18 @@ def gather_size_by_comm(size: int, group: dist.ProcessGroup) -> List[int]:
...
+
+        if gather_device == "cpu":
+            raise RuntimeError(
+                "Ulysses Anything Attention (UAA) requires an accelerator "
+                "(CUDA, NPU, XPU, etc.) to perform distributed gathering, "
+                "but only CPU was detected."
+            )
```
sayakpaul: I think we can simplify the error message to something like:

"No suitable accelerator found."

akshatvishu: Done. I've simplified the msg!

src/diffusers/models/_modeling_parallel.py
akshatvishu: I've updated the implementation to wrap the get_device() result in torch.device()

PR #13078, opened 2026-02-04
[modular] add modular tests for Z-Image and Wan

sayakpaul:
# What does this PR do?

* Add tests for Z-Image (only T2I for keeping things simple).
* Add tests for Wan modular (just T2V to keep things simple; can add more if needed).
* Modify the `test_modular_pipelines_common.py` to account for `output_type="video"` and other changes that come with video pipelines.

### DIFF
--- a/docs/source/en/api/pipelines/ltx2.md
+++ b/docs/source/en/api/pipelines/ltx2.md
@@ -106,8 +106,6 @@ video, audio = pipe(
     output_type="np",
     return_dict=False,
 )
-video = (video * 255).round().astype("uint8")
-video = torch.from_numpy(video)
 
 encode_video(
     video[0],
@@ -185,8 +183,6 @@ video, audio = pipe(
     output_type="np",
     return_dict=False,
 )
-video = (video * 255).round().astype("uint8")
-video = torch.from_numpy(video)
 
 encode_video(
     video[0],
--- a/src/diffusers/modular_pipelines/modular_pipeline.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline.py
@@ -1598,11 +1598,11 @@ class ModularPipeline(ConfigMixin, PushToHubMixin):
             else:
                 logger.warning(f"`blocks` is `None`, no default blocks class found for {self.__class__.__name__}")
 
-        self.blocks = blocks
+        self._blocks = blocks
         self._components_manager = components_manager
         self._collection = collection
-        self._component_specs = {spec.name: deepcopy(spec) for spec in self.blocks.expected_components}
-        self._config_specs = {spec.name: deepcopy(spec) for spec in self.blocks.expected_configs}
+        self._component_specs = {spec.name: deepcopy(spec) for spec in self._blocks.expected_components}
+        self._config_specs = {spec.name: deepcopy(spec) for spec in self._blocks.expected_configs}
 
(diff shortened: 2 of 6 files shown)


tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -37,6 +38,9 @@ class ModularPipelineTesterMixin:
     optional_params = frozenset(["num_inference_steps", "num_images_per_prompt", "latents", "output_type"])
     # this is modular specific: generator needs to be a intermediate input because it's mutable
     intermediate_params = frozenset(["generator"])
+    # Output type for the pipeline (e.g., "images" for image pipelines, "videos" for video pipelines)
+    # Subclasses can override this to change the expected output type
+    output_type = "images"
```
yiyixuxu: I think it's better to name this "output_name" 
 since`output_type` is a standard pipeline argumnet

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -197,12 +225,27 @@ def test_inference_batch_single_identical(
...
-        assert output_batch.shape[0] == batch_size
-
-        max_diff = torch.abs(output_batch[0] - output[0]).max()
+        output = pipe(**inputs, output=self.output_type)
+        output_batch = pipe(**batched_inputs, output=self.output_type)
+
+        assert self._get_batch_size_from_output(output_batch) == batch_size
+
+        # Convert outputs to tensors for comparison
+        if isinstance(output, list) and isinstance(output_batch, list):
```
yiyixuxu: are we trying to handle 3 different output formats here? (list of tensors, list of numpy arrays, single tensor)
Do we actually need all this? Shouldn't our pipeline outputs more or less standardized? at least when you pass a `output_type`

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -37,6 +38,12 @@ class ModularPipelineTesterMixin:
     optional_params = frozenset(["num_inference_steps", "num_images_per_prompt", "latents", "output_type"])
     # this is modular specific: generator needs to be a intermediate input because it's mutable
     intermediate_params = frozenset(["generator"])
+    # Output type for the pipeline (e.g., "images" for image pipelines, "videos" for video pipelines)
+    # Subclasses can override this to change the expected output type
+    output_name = "images"
+    # Whether the pipeline returns tensors directly with output_type="pt" or needs conversion
+    # Set to True for pipelines that don't support output_type parameter (e.g., WAN)
```
yiyixuxu: ohh it is for wan
actually can you add that argument for wan, probably just here https://github.com/huggingface/diffusers/blob/main/src/diffusers/modular_pipelines/wan/decoders.py#L52

yiyixuxu: ok we added it in this PR https://github.com/huggingface/diffusers/pull/13063
and will merge that in first, there are a few other refactors in addition to `output_type` but it should not affect this PR sinnce we only test t2v here

sayakpaul:
@yiyixuxu I think this should be ready to go. I merged in https://github.com/huggingface/diffusers/pull/13063 and it helped simplify this PR quite a bit. LMK if you have any further comments.

PR #13081, opened 2026-02-04
[core] make qwen hidden states contiguous to make torchao happy.

sayakpaul:
# What does this PR do?

Same as https://github.com/huggingface/diffusers/pull/13068/


yiyixuxu (approved):
thanks!

PR #13082, opened 2026-02-04
[CI] Refactor Wan Model Tests

DN6:
# What does this PR do?
Update Wan tests with new format

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/models/transformers/transformer_wan_animate.py
+++ b/src/diffusers/models/transformers/transformer_wan_animate.py
@@ -54,7 +54,7 @@ def _get_qkv_projections(attn: "WanAttention", hidden_states: torch.Tensor, enco
         encoder_hidden_states = hidden_states
 
     if attn.fused_projections:
-        if attn.cross_attention_dim_head is None:
+        if not attn.is_cross_attention:
             # In self-attention layers, we can fuse the entire QKV projection into a single linear
             query, key, value = attn.to_qkv(hidden_states).chunk(3, dim=-1)
         else:
@@ -685,7 +685,10 @@ class WanAttention(torch.nn.Module, AttentionModuleMixin):
             self.add_v_proj = torch.nn.Linear(added_kv_proj_dim, self.inner_dim, bias=True)
             self.norm_added_k = torch.nn.RMSNorm(dim_head * heads, eps=eps)
 
-        self.is_cross_attention = cross_attention_dim_head is not None
+        if is_cross_attention is not None:
+            self.is_cross_attention = is_cross_attention
+        else:
+            self.is_cross_attention = cross_attention_dim_head is not None
 
         self.set_processor(processor)
 
@@ -693,7 +696,7 @@ class WanAttention(torch.nn.Module, AttentionModuleMixin):
         if getattr(self, "fused_projections", False):
             return
 
-        if self.cross_attention_dim_head is None:
+        if not self.is_cross_attention:
             concatenated_weights = torch.cat([self.to_q.weight.data, self.to_k.weight.data, self.to_v.weight.data])
             concatenated_bias = torch.cat([self.to_q.bias.data, self.to_k.bias.data, self.to_v.bias.data])
             out_features, in_features = concatenated_weights.shape
--- a/src/diffusers/models/transformers/transformer_chronoedit.py
+++ b/src/diffusers/models/transformers/transformer_chronoedit.py
@@ -43,7 +43,7 @@ def _get_qkv_projections(attn: "WanAttention", hidden_states: torch.Tensor, enco
         encoder_hidden_states = hidden_states
 
     if attn.fused_projections:
-        if attn.cross_attention_dim_head is None:
+        if not attn.is_cross_attention:
             # In self-attention layers, we can fuse the entire QKV projection into a single linear
             query, key, value = attn.to_qkv(hidden_states).chunk(3, dim=-1)
         else:
@@ -219,7 +219,10 @@ class WanAttention(torch.nn.Module, AttentionModuleMixin):
             self.add_v_proj = torch.nn.Linear(added_kv_proj_dim, self.inner_dim, bias=True)
             self.norm_added_k = torch.nn.RMSNorm(dim_head * heads, eps=eps)
 
-        self.is_cross_attention = cross_attention_dim_head is not None
+        if is_cross_attention is not None:
+            self.is_cross_attention = is_cross_attention
+        else:
+            self.is_cross_attention = cross_attention_dim_head is not None
 
         self.set_processor(processor)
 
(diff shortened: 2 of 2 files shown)


DN6:
Thanks for flagging @dg845. I've fixed the test issues. There are some GGUF related fixes that should probably go in a different PR (will handle that later)

dg845 (approved):
Thanks!

src/diffusers/models/transformers/transformer_wan.py
```diff
@@ -214,15 +214,18 @@ def __init__(
             self.add_v_proj = torch.nn.Linear(added_kv_proj_dim, self.inner_dim, bias=True)
             self.norm_added_k = torch.nn.RMSNorm(dim_head * heads, eps=eps)
 
-        self.is_cross_attention = cross_attention_dim_head is not None
+        if is_cross_attention is not None:
+            self.is_cross_attention = is_cross_attention
+        else:
+            self.is_cross_attention = cross_attention_dim_head is not None
```
sayakpaul, line 220: Seems like an unrelated change?

src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -502,24 +502,27 @@ def __init__(
...
         eps: float = 1e-6,
         cross_attention_dim_head: Optional[int] = None,
+        bias: bool = True,
         processor=None,
     ):
         super().__init__()
         self.inner_dim = dim_head * heads
         self.heads = heads
-        self.cross_attention_head_dim = cross_attention_dim_head
+        self.cross_attention_dim_head = cross_attention_dim_head
```
sayakpaul, line 511: Same as above. Would be nice if you could explain these changes? Were these flagged by the newly written test suite?

tests/models/testing_utils/common.py
```diff
@@ -446,16 +446,17 @@ def test_getattr_is_correct(self, caplog):
...
         if fp32_modules is None or len(fp32_modules) == 0:
             pytest.skip("Model does not have _keep_in_fp32_modules defined.")
 
-        # Test with float16
-        model.to(torch_device)
-        model.to(torch.float16)
+        # Save the model and reload with float16 dtype
+        # _keep_in_fp32_modules is only enforced during from_pretrained loading
+        model.save_pretrained(tmp_path)
+        model = self.model_class.from_pretrained(tmp_path, torch_dtype=torch.float16).to(torch_device)
```
sayakpaul, line 459: Have we run this test for all the other models to ensure this isn't breaking anything?

tests/models/testing_utils/quantization.py
```diff
@@ -176,15 +176,7 @@ def _test_quantization_inference(self, config_kwargs):
...
 
-        # Get model dtype from first parameter
-        model_dtype = next(model_quantized.parameters()).dtype
-
         inputs = self.get_dummy_inputs()
-        # Cast inputs to model dtype
-        inputs = {
-            k: v.to(model_dtype) if isinstance(v, torch.Tensor) and v.is_floating_point() else v
-            for k, v in inputs.items()
-        }
```
sayakpaul, line 187: Why remove them?

tests/models/testing_utils/quantization.py
```diff
@@ -1021,9 +1015,6 @@ def test_gguf_dequantize(self):
         """Test that dequantize() works correctly."""
         self._test_dequantize({"compute_dtype": torch.bfloat16})
 
-    def test_gguf_quantized_layers(self):
-        self._test_quantized_layers({"compute_dtype": torch.bfloat16})
```
sayakpaul, line 1025: Why remove? Is it a duplicate?

tests/models/transformers/test_models_transformer_wan.py
```diff
@@ -12,57 +12,57 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import unittest
```
sayakpaul, line 15: I am guessing the changes under `tests/models/transformers/` were all auto-generated?

tests/models/transformers/test_models_transformer_wan_vace.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+class TestWanVACETransformer3DAttention(WanVACETransformer3DTesterConfig, AttentionTesterMixin):
+    """Attention processor tests for Wan VACE Transformer 3D."""
+
+
+class TestWanVACETransformer3DCompile(WanVACETransformer3DTesterConfig, TorchCompileTesterMixin):
+    """Torch compile tests for Wan VACE Transformer 3D."""
+
+    def test_torch_compile_repeated_blocks(self):
```
sayakpaul, line 147: I think we can further simplify this test by letting users pass a `recompile_limit`. I will open a PR.

src/diffusers/models/transformers/transformer_wan.py
DN6: This is to ensure fused projections work across all Wan models that use WanAttention.  Wan VACE passes `is_cross_attention` rather than setting `cross_attention_dim_head`

src/diffusers/models/transformers/transformer_wan_animate.py
DN6: This is just to keep the naming convention consistent

tests/models/testing_utils/common.py
DN6: With the current way this test is written, any model that that has `_keep_in_fp32_modules` will fail this test since using `model.to(..)` doesn't respect  `_keep_in_fp32_modules` (nor should it IMO)

This test was newly introduced in the refactor. I didn't catch the issue when merging the PR because Flux doesn't use `_keep_in_fp32_modules`.

tests/models/testing_utils/quantization.py
DN6: Casting here is brittle because it's based on `model_dtype` which we get from `model_dtype = next(model_quantized.parameters()).dtype`. This can lead to different dtypes across different models and different quantization schemes. e.g With Flux + GGUF the test passes because the parameter dtype is the same the input dtype (bfloat16). However with Wan it fails because the parameter dtype is int8.

tests/models/testing_utils/quantization.py
DN6: It's because the test only works properly for models that can be quantized via `from_pretrained`. The test checks whether all applicable Linear layers in a model are quantized. 

With GGUF we only support loading prequantized checkpoints, so the choice of which layers to quantize is entirely up to the person that created the checkpoint.

tests/models/testing_utils/quantization.py
sayakpaul: Makes sense. But does it affect the existing Flux tests? 

> Casting here is brittle because it's based on model_dtype which we get from model_dtype = next(model_quantized.parameters()).dtype

I wonder if using `.dtype` on a model subclassed from `ModelMixin` would alleviate this problem because `dtype` implementation is quite elaborate:
https://github.com/huggingface/diffusers/blob/5bf248ddd8796b4f4958559429071a28f9b2dd3a/src/diffusers/models/modeling_utils.py#L155

DN6: I've added a torch_dtype property to the quantization tests and we cast the inputs directly in get_dummy_inputs. Think it's more clear this way 

Flux TorchAO and BnB tests will fail with this change, but I'll update the Flux2 PR to include fixes to address the change in this test.

sayakpaul: Sounds good. Thanks!

PR #13092, opened 2026-02-07
[Fix]Allow `prompt` and `prior_token_ids` to be provided simultaneously in `GlmImagePipeline`

JaredforReal:
### What does this PR do?

Previously, `GlmImagePipeline.check_inputs()` enforced mutual exclusion between `prompt` and `prior_token_ids`, raising a `ValueError` if both were provided. This was unnecessarily restrictive — there is a valid use case where a user wants to supply pre-computed `prior_token_ids` (to skip the expensive AR generation step) while still passing `prompt` so that `prompt_embeds` (glyph embeddings) can be derived from it automatically.

This PR relaxes that constraint:

- **Removed** the check that raises when both `prompt` and `prior_token_ids` are provided.
- **Relaxed** the `prior_token_ids` requires `prompt_embeds` check — now either `prompt` *or* `prompt_embeds` satisfies the requirement, since `prompt` will be used to generate `prompt_embeds` via `encode_prompt()` downstream.

### Supported input combinations after this change

| Inputs | Behavior | Status |
|---|---|---|
| `prompt` only | AR generates `prior_token_ids`; glyph encoder generates `prompt_embeds` | ✅ (unchanged) |
| `prior_token_ids` + `prompt_embeds` | Both used directly, AR step skipped | ✅ (unchanged) |
| `prompt` + `prior_token_ids` | `prior_token_ids` used directly (AR skipped); `prompt` generates `prompt_embeds` | ✅ **new** |
| Neither `prompt` nor `prior_token_ids` | ❌ `ValueError` | ✅ (unchanged) |
| `prior_token_ids` alone (no `prompt`/`prompt_embeds`) | ❌ `ValueError` | ✅ (unchanged) |
| `prompt` + `prompt_embeds` | ❌ `ValueError` | ✅ (unchanged) |

### Tests

Added two new tests to test_glm_image.py:

- **`test_prompt_with_prior_token_ids`**: End-to-end test that first generates `prior_token_ids` via `generate_prior_tokens()`, then runs the full pipeline with both `prompt` and `prior_token_ids` provided together.
- **`test_check_inputs_rejects_invalid_combinations`**: Unit test verifying that the three invalid input combinations listed above still correctly raise `ValueError`.

All existing tests continue to pass.

### DIFF
--- a/tests/pipelines/glm_image/test_glm_image.py
+++ b/tests/pipelines/glm_image/test_glm_image.py
@@ -297,16 +297,6 @@ class GlmImagePipelineFastTests(PipelineTesterMixin, unittest.TestCase):
 
         # Step 1: Run with prompt only to get prior_token_ids from AR model
         generator = torch.Generator(device=device).manual_seed(0)
-        inputs_prompt_only = {
-            "prompt": "A photo of a cat",
-            "generator": generator,
-            "num_inference_steps": 2,
-            "guidance_scale": 1.5,
-            "height": height,
-            "width": width,
-            "max_sequence_length": 16,
-            "output_type": "pt",
-        }
         prior_token_ids, _, _ = pipe.generate_prior_tokens(
             prompt="A photo of a cat",
             height=height,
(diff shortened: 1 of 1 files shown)


JaredforReal:
@yiyixuxu PTAL, thanks

yiyixuxu (approved):
thanks!

JaredforReal:
@yiyixuxu Sorry to be late, trigger CI, and make quality should be able to pass

yiyixuxu:
@JaredforReal 
thanks for the PR!

PR #13114, opened 2026-02-10
Cosmos Transfer2.5 Auto-Regressive Inference Pipeline

miguelmartin75:
# What does this PR do?

This builds off https://github.com/huggingface/diffusers/pull/13066 by adding auto-regressive inference for Cosmos Transfer2.5. This pipeline does not require the controlnet or controls to be input. From the documentation:

> The call function can be used in two modes: with or without controls.
>        When controls are not provided (`controls is None`), inference works in the same manner as predict2.5 (see
        `Cosmos2_5_PredictPipeline`). This mode strictly uses the base transformer (`self.transformer`) to perform
        inference and accepts as input an optional `image` or `video` along with a `prompt` / `negative_prompt`, and
        can be used in the following ways:
        - **Text2World**: `image=None`, `video=None`, `prompt` provided.
        - **Image2World**: `image` provided, `video=None`, `prompt` provided.
        - **Video2World**: `video` provided, `image=None`, `prompt` provided.
>        When `controls` are provided and a ControlNet is attached, `controls` drive the conditioning and `video` &
        `image` is ignored. Controls are assumed to be pre-processed, e.g. edge maps are pre-computed.
>        Setting `num_frames` will restrict the total number of frames output, if not provided or assigned to None
        (default) then the number of output frames will match the input `video`, `image` or `controls` respectively.
        Auto-regressive inference is supported and thus a sliding window of `num_frames_per_chunk` frames are used per
        denoising loop. In addition, when auto-regressive inference is performed, the previous
        `num_latent_conditional_frames` or `num_conditional_frames` are used to condition the following denoising
        inference loops.

## Who can review?

- Pipelines: @yiyixuxu 

### DIFF
--- a/src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
+++ b/src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
@@ -456,7 +456,7 @@ class Cosmos2_5_TransferPipeline(DiffusionPipeline):
                 cond_indicator,
             )
 
-    # Copied from diffusers.pipelines.cosmos.pipeline_cosmos_text2world.CosmosTextToWorldPipeline.check_inputs
+    # Modified from diffusers.pipelines.cosmos.pipeline_cosmos_text2world.CosmosTextToWorldPipeline.check_inputs
     def check_inputs(
         self,
         prompt,
@@ -466,9 +466,23 @@ class Cosmos2_5_TransferPipeline(DiffusionPipeline):
         callback_on_step_end_tensor_inputs=None,
         num_ar_conditional_frames=None,
         num_ar_latent_conditional_frames=None,
+        num_frames_per_chunk=None,
+        num_frames=None,
+        conditional_frame_timestep=0.1,
     ):
-        if height % 16 != 0 or width % 16 != 0:
-            raise ValueError(f"`height` and `width` have to be divisible by 16 but are {height} and {width}.")
+        if width <= 0 or height <= 0 or height % 16 != 0 or width % 16 != 0:
+            raise ValueError(
+                f"`height` and `width` have to be divisible by 16 (& positive) but are {height} and {width}."
+            )
+
+        if num_frames is not None and num_frames <= 0:
+            raise ValueError(f"`num_frames` has to be a positive integer when provided but is {num_frames}.")
+
+        if conditional_frame_timestep < 0 or conditional_frame_timestep > 1:
+            raise ValueError(
+                "`conditional_frame_timestep` has to be a float in the [0, 1] interval but is "
+                f"{conditional_frame_timestep}."
+            )
 
         if callback_on_step_end_tensor_inputs is not None and not all(
             k in self._callback_tensor_inputs for k in callback_on_step_end_tensor_inputs
@@ -499,6 +513,35 @@ class Cosmos2_5_TransferPipeline(DiffusionPipeline):
             raise ValueError("`num_ar_latent_conditional_frames` must be >= 0.")
         if num_ar_conditional_frames is not None and num_ar_conditional_frames < 0:
             raise ValueError("`num_ar_conditional_frames` must be >= 0.")
+
+        if num_ar_latent_conditional_frames is not None:
+            num_ar_conditional_frames = max(
+                0, (num_ar_latent_conditional_frames - 1) * self.vae_scale_factor_temporal + 1
+            )
+
+        min_chunk_len = self.vae_scale_factor_temporal + 1
+        if num_frames_per_chunk < min_chunk_len:
+            logger.warning(f"{num_frames_per_chunk=} must be larger than {min_chunk_len=}, setting to min_chunk_len")
+            num_frames_per_chunk = min_chunk_len
+
+        max_frames_by_rope = None
+        if getattr(self.transformer.config, "max_size", None) is not None:
+            max_frames_by_rope = max(
+                size // patch
+                for size, patch in zip(self.transformer.config.max_size, self.transformer.config.patch_size)
+            )
+            if num_frames_per_chunk > max_frames_by_rope:
+                raise ValueError(
+                    f"{num_frames_per_chunk=} is too large for RoPE setting ({max_frames_by_rope=}). "
+                    "Please reduce `num_frames_per_chunk`."
+                )
+
+        if num_ar_conditional_frames >= num_frames_per_chunk:
+            raise ValueError(
+                f"{num_ar_conditional_frames=} must be smaller than {num_frames_per_chunk=} for chunked generation."
+            )
+
+        return num_frames_per_chunk
 
     @property
     def guidance_scale(self):
@@ -663,8 +706,7 @@ class Cosmos2_5_TransferPipeline(DiffusionPipeline):
                     raise ValueError("`controls` must contain 3D frames in CHW format.")
                 width = int((height + 16) * (frame.shape[2] / frame.shape[1]))  # NOTE: assuming C H W
 
-        # Check inputs. Raise error if not correct
-        self.check_inputs(
+        num_frames_per_chunk = self.check_inputs(
             prompt,
             height,
             width,
@@ -672,6 +714,9 @@ class Cosmos2_5_TransferPipeline(DiffusionPipeline):
             callback_on_step_end_tensor_inputs,
             num_ar_conditional_frames,
             num_ar_latent_conditional_frames,
+            num_frames_per_chunk,
+            num_frames,
+            conditional_frame_timestep,
         )
 
         if num_ar_latent_conditional_frames is not None:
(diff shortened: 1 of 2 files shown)


src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -535,18 +511,19 @@ def __call__(
         self,
         image: PipelineImageInput | None = None,
         video: List[PipelineImageInput] | None = None,
+        controls: Optional[PipelineImageInput | List[PipelineImageInput]] = None,
```
yiyixuxu: maybe we should make this pipeline strictly about controlnet (i.e. not make it optional) and then remove the image and video argument? this is how other controlnet behave anyways
if they want to use without controlnet, they can switich to the base pipeline

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -660,6 +654,34 @@ def __call__(
             else:
                 width = int((height + 16) * (frame.shape[2] / frame.shape[1]))  # NOTE: assuming C H W
 
+        if num_latent_conditional_frames is not None and num_conditional_frames is not None:
```
yiyixuxu: any reason we need two arguments here? is it possible we only keep num_conditional_frames?

miguelmartin75: this is done to provide the user with the option to provide either one, the official GH uses `num_conditional_frames` for transfer but `num_latent_conditional_frames` for predict so I figured it would be best to provide both

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -647,19 +624,50 @@ def __call__(
...
+                    frame = frame[0]
+
+            if isinstance(frame, PIL.Image.Image):
                 width = int((height + 16) * (frame.width / frame.height))
             else:
+                if frame.ndim != 3:
+                    raise ValueError("`controls` must contain 3D frames in CHW format.")
                 width = int((height + 16) * (frame.shape[2] / frame.shape[1]))  # NOTE: assuming C H W
 
+        if num_latent_conditional_frames is not None and num_conditional_frames is not None:
```
yiyixuxu: Could we simplify this validation logic and move it into `check_inputs`? I think we don't need to try to infer user intent here — let's just document arguments clearly, enforce it strictly and raise otherwise:

```python
if num_latent_conditional_frames is not None and num_conditional_frames is not None:
    raise ValueError(
        "Provide only one of `num_conditional_frames` or `num_latent_conditional_frames`, not both."
    )
if num_latent_conditional_frames is None and num_conditional_frames is None:
    raise ValueError(
        "Provide either `num_conditional_frames` or `num_latent_conditional_frames`."
    )
if num_latent_conditional_frames is not None and num_latent_conditional_frames < 0:
    raise ValueError("`num_latent_conditional_frames` must be >= 0.")
if num_conditional_frames is not None and num_conditional_frames < 0:
    raise ValueError("`num_conditional_frames` must be >= 0.")
```

The fallback computation that derives one from the other can stay in `__call__` after validation.

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -706,102 +714,163 @@ def __call__(
...
+                control_video = control_video.repeat(batch_size, 1, 1, 1, 1)
+            else:
+                raise ValueError(
+                    f"Expected controls batch size {batch_size} to match prompt batch size, but got {control_video.shape[0]}."
+                )
+
+        num_frames_out = control_video.shape[2]
+        if num_frames is not None:
+            num_frames_out = min(num_frames_out, num_frames)
+        if num_frames_out == 0:
```
yiyixuxu: why would num_frames_out == 0 here?

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -706,102 +714,163 @@ def __call__(
...
+            num_frames_out = num_frames_per_chunk
+
+        control_video = _maybe_pad_or_trim_video(control_video, num_frames_out)
 
-                self._current_timestep = t.cpu().item()
+        min_chunk_len = 1 if num_frames_out <= 1 else self.vae_scale_factor_temporal + 1
```
yiyixuxu: why do we single out num_frames_out <=1, what happends if it >1 but less then the `self.vae_scale_factor_temporal + 1`? 
you will also get a chunk larger than the total frames, no?

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -627,6 +600,10 @@ def __call__(
             max_sequence_length (`int`, defaults to `512`):
                 The maximum number of tokens in the prompt. If the prompt exceeds this length, it will be truncated. If
                 the prompt is shorter than this length, it will be padded.
+            num_conditional_frames (`int`, *optional*, defaults to `1`):
```
yiyixuxu: maybe explain a bit more here? as it is not yet a standard variables for our pipelines, 
also `num_ar_conditional_frames` might be a better name because this is specific to AR: e.g. how many frames are used to condition the next chunk

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -706,102 +714,163 @@ def __call__(
...
+                    ]
+                else:
+                    controls_latents = [
+                        retrieve_latents(self.vae.encode(vid.unsqueeze(0)), generator=generator)
+                        for vid in chunk_control_video
+                    ]
+                controls_latents = torch.cat(controls_latents, dim=0).to(transformer_dtype)
+
+                latents_mean = self.latents_mean.to(device=device, dtype=transformer_dtype)
+                latents_std = self.latents_std.to(device=device, dtype=transformer_dtype)
```
yiyixuxu:
```suggestion
```
we already did this before the loop -
we don't need to do this on each iteration, no?

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
miguelmartin75: This is left over when video, image and controls was optional inputs. I will delete this branch.

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
miguelmartin75: yes, but there is a slice operation when returning the output

```python
video_chunks = [
    chunk[:, :, num_conditional_frames:, ...] if chunk_idx != 0 else chunk
    for chunk_idx, chunk in enumerate(video_chunks)
]
video = torch.cat(video_chunks, dim=2)
video = video[:, :, :num_frames_out, ...]
```

miguelmartin75: Doubling checking the first question

miguelmartin75: I've removed the if - it's not needed, just some additional compute, but it is insignificant.

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
miguelmartin75: Added additional context in docs + renamed the args

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
miguelmartin75: Removed

yiyixuxu (approved):
thanks!
let's fix the CI too

src/diffusers/pipelines/cosmos/pipeline_cosmos2_5_transfer.py
```diff
@@ -709,102 +723,161 @@ def __call__(
...
+                logger.warning(
+                    f"{num_frames_per_chunk=} is too large for RoPE setting to maximum ({max_frames_by_rope=})"
                 )
+                num_frames_per_chunk = max_frames_by_rope
 
-                in_latents = cond_mask * cond_latent + (1 - cond_mask) * latents
-                in_latents = in_latents.to(transformer_dtype)
-                in_timestep = cond_indicator * cond_timestep + (1 - cond_indicator) * sigma_t
-                control_blocks = None
-                if controls_latents is not None and self.controlnet is not None:
```
yiyixuxu, line 807: I think this block of code can be moved into check_inputs and raise as well.
Since given the checkpoint, the range is fixed, so we can give the user a clear error message

miguelmartin75: I've moved this block

miguelmartin75:
CI should be fixed now

PR #13116, opened 2026-02-10
[tests] tests for `modules_to_not_convert`

sayakpaul:
# What does this PR do?

The base class has a test for it but I think we didn't actually make it so that the tests would run for different backends.

@DN6 could you test Quanto and GGUF?


tests/models/testing_utils/quantization.py
```diff
@@ -273,7 +263,9 @@ def _test_quantized_layers(self, config_kwargs):
             f"Quantized layer count mismatch: expected {expected_quantized_layers}, got {num_quantized_layers} (total linear layers: {num_linear_layers}, FP32 modules: {num_fp32_modules})"
         )
 
-    def _test_quantization_modules_to_not_convert(self, config_kwargs, modules_to_not_convert):
+    def _test_quantization_modules_to_not_convert(
+        self, config_kwargs, modules_to_not_convert, to_not_convert_key="modules_to_not_convert"
```
DN6: Instead of introducing an abstraction to deal with the fact that BnB does not use the  `modules_to_not_convert` convention, let's just rewrite the test from scratch for BnB.

sayakpaul:
@DN6 a gentle ping here.

sayakpaul:
@DN6 a gentle ping

Issue #13107, opened 2026-02-09
Flux Klein LoRA – Missing Key Errors


mshehzadkhan:
### Describe the bug

It is happeing on most of the loras, i tried from civit ai

The failure occurs during Flux2 LoRA key conversion inside Diffusers, specifically in the function:

_convert_non_diffusers_flux2_lora_to_diffusers

When calling pipe.load_lora_weights(...), Diffusers attempts to convert a non-Diffusers Flux2-style LoRA checkpoint to the internal Diffusers format. During this conversion, the following error is raised:

```
state_dict = _convert_non_diffusers_flux2_lora_to_diffusers(state_dict)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../diffusers/loaders/lora_conversion_utils.py", line 2340, in _convert_non_diffusers_flux2_lora_to_diffusers
    converted_state_dict[
        f"{attn_prefix}.to_qkv_mlp_proj.{lora_key}.weight"
    ] = original_state_dict.pop(
        "single_blocks.0.linear1.lora_A.weight"
    )
KeyError: 'single_blocks.0.linear1.lora_A.weight'
```

### Reproduction

```
import torch
from diffusers import Flux2KleinPipeline

device = "cuda"
dtype = torch.bfloat16

pipe = Flux2KleinPipeline.from_pretrained(
    "black-forest-labs/FLUX.2-klein-4B",
    torch_dtype=dtype
).to(device)

pipe.load_lora_weights("shehzad1/Klein4B_Base_lora")

prompt = "A cat holding a sign that says hello world"

image = pipe(
    prompt=prompt,
    height=1024,
    width=1024,
    guidance_scale=1.0,
    num_inference_steps=4,
    generator=torch.Generator(device=device).manual_seed(0)
).images[0]

image.save("flux-klein.png")

```

### Logs

```shell

```

### System Info

Diffusers: latest@main
Python: 3.12
Torch: 2.9.1+cu128

### Who can help?

@sayakpaul @DN6 @hlky

PR #13119, opened 2026-02-10
[lora] fix non-diffusers lora key handling for flux2

sayakpaul:
# What does this PR do?

Fix https://github.com/huggingface/diffusers/issues/13107

Cc: @mshehzadkhan. Could you give this a try please?


mshehzadkhan:
I’ve tested a bunch of Klein LoRAs for each model from CivitAI, and they all seem to be loading correctly now.

Thanks a lot @sayakpaul for the quick fix 🙌

PR #13120, opened 2026-02-10
[modular] add tests for robust model loading.

sayakpaul:
# What does this PR do?

Add tests for model loading w.r.t https://github.com/huggingface/diffusers/pull/13025.

### DIFF
--- a/src/diffusers/models/transformers/transformer_chronoedit.py
+++ b/src/diffusers/models/transformers/transformer_chronoedit.py
@@ -43,7 +43,7 @@ def _get_qkv_projections(attn: "WanAttention", hidden_states: torch.Tensor, enco
         encoder_hidden_states = hidden_states
 
     if attn.fused_projections:
-        if attn.cross_attention_dim_head is None:
+        if not attn.is_cross_attention:
             # In self-attention layers, we can fuse the entire QKV projection into a single linear
             query, key, value = attn.to_qkv(hidden_states).chunk(3, dim=-1)
         else:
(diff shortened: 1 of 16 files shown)


tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -598,3 +598,69 @@ def test_model_description_includes_block_count(self):
         content = generate_modular_model_card_content(blocks)
 
         assert "5-block architecture" in content["model_description"]
+
+
+class TestAutoModelLoadIdTagging:
```
sayakpaul, line 603: I think it's sufficient to test like this instead of turning this into a Mixin? Not too strongly opinionated, though. Can iterate if needed.

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -598,3 +598,69 @@ def test_model_description_includes_block_count(self):
...
+
+        assert pipe.unet is auto_model
+
+        assert "unet" in pipe._component_specs
+        spec = pipe._component_specs["unet"]
+        assert spec.pretrained_model_name_or_path == "hf-internal-testing/tiny-stable-diffusion-xl-pipe"
+        assert spec.subfolder == "unet"
+
+
+class TestLoadComponentsSkipBehavior:
```
sayakpaul, line 631: Same as above.

tests/modular_pipelines/test_modular_pipelines_common.py
yiyixuxu: i think it's enough like this

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -598,3 +598,69 @@ def test_model_description_includes_block_count(self):
...
+
+    def test_load_components_selective_loading(self):
+        pipe = ModularPipeline.from_pretrained("hf-internal-testing/tiny-stable-diffusion-xl-pipe")
+
+        pipe.load_components(names="unet", torch_dtype=torch.float32)
+
+        # Verify only requested component was loaded.
+        assert hasattr(pipe, "unet")
+        assert pipe.unet is not None
+        if "vae" in pipe._component_specs:
```
yiyixuxu:
```suggestion
```
 we know if we have vae or not no?

PR #13123, opened 2026-02-11
Allow Automodel to use  `from_config` with custom code.

DN6:
# What does this PR do?
AutoModel currently doesn't support creating objects using `from_config`,  but this can be useful in cases where we might want to load a custom component that doesn't have any weights associated with it e.g. Custom scheduler like components for pipelines.

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/models/auto_model.py
+++ b/src/diffusers/models/auto_model.py
@@ -104,26 +104,10 @@ class AutoModel(ConfigMixin):
 
         if isinstance(pretrained_model_name_or_path_or_dict, (str, os.PathLike)):
             pretrained_model_name_or_path = pretrained_model_name_or_path_or_dict
-            cls.config_name = "config.json"
             config = cls.load_config(pretrained_model_name_or_path, subfolder=subfolder, **hub_kwargs)
         else:
             config = pretrained_model_name_or_path_or_dict
-            pretrained_model_name_or_path = config.get("_name_or_path")
-
-        library = None
-        orig_class_name = None
-
-        if "_class_name" in config:
-            orig_class_name = config["_class_name"]
-            library = "diffusers"
-        elif "model_type" in config:
-            orig_class_name = "AutoModel"
-            library = "transformers"
-        else:
-            raise ValueError(
-                f"Couldn't find a model class associated with the config: {config}. Make sure the config "
-                "contains a `_class_name` or `model_type` key."
-            )
+            pretrained_model_name_or_path = config.get("_name_or_path", None)
 
         has_remote_code = "auto_map" in config and cls.__name__ in config["auto_map"]
         trust_remote_code = resolve_trust_remote_code(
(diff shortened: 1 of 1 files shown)


src/diffusers/models/auto_model.py
```diff
@@ -31,10 +31,128 @@ class AutoModel(ConfigMixin):
...
         raise EnvironmentError(
             f"{self.__class__.__name__} is designed to be instantiated "
-            f"using the `{self.__class__.__name__}.from_pretrained(pretrained_model_name_or_path)` or "
+            f"using the `{self.__class__.__name__}.from_pretrained(pretrained_model_name_or_path)`, "
+            f"`{self.__class__.__name__}.from_config(config)`, or "
             f"`{self.__class__.__name__}.from_pipe(pipeline)` methods."
         )
 
+    @classmethod
+    def from_config(
```
yiyixuxu, line 40: ohh I think maybe we should support this in `from_pretrained()`? even for scheduler and such weightless things, we load them using `from_pretrained()`, no?

we normally use `from_config` like this, would be confusing I think 
```py
pipe = DiffusionPipeline.from_pretrained(..)
pipe.scheduler = NEWSCHEDULER.from_config(pipe.config)
```

DN6: This is true. But the SchedulerMixin's `from_pretrained` just runs `from_config` under the hood. 
https://github.com/huggingface/diffusers/blob/985d83c948ab890bac8c4323217b1ea3e833ab04/src/diffusers/schedulers/scheduling_utils.py#L147-L154

I thought perhaps we could make an exception for just the AutoModel class? The `load_config` and then `from_config` flow doesn't work too well when trying to load custom code.

yiyixuxu: ohh thanks for explaining. I think I got it now. 
but is it possible to not re-use the `from_config` API with a different usage pattern?  I think maybe we can add a new arg to `from_pretrained` to support this use case? e.g `load_weights=True`, `init_from_config=False` etc

Even for custom components that come with the weights, we neeed to support for this use case, for example training, or stuff like here we need to initialize it to inspect model structure for quantisation https://github.com/cubiq/Mellon/blob/main/modules/ModularDiffusers/loaders.py#L181

yiyixuxu: or maybe a brand new method if it you prefer

yiyixuxu: init_from_hub? from_pretrained_config? etc

DN6: IMO `from_config` is the preferred option because it works well with ComponentSpec's `default_creation_method` as well. So we don't need to introduce too many changes to modular get this functionality. 

Models currently support passing in a `pretrained_model_name_or_path`. I know that there was a plan to deprecate this behaviour (but I'm not clear as to why the deprecation is needed?)  
https://github.com/huggingface/diffusers/blob/76af013a41fd6ba5f977bc015364a5e7d57fd5ff/src/diffusers/configuration_utils.py#L237-L251

yiyixuxu: >but I'm not clear as to why the deprecation is needed?

I think it's before my time in but if I have to guess, they deprecated it after adding `from_pretrained()` to `scheduler` to make API cleaner: `from_pretrained` handles loading (without or without weights) and `from_config` takes only `config dict` 

so if we undo the deprecation & introduce the from_config on AutoModel, the mental model would be something like this?
* `from_pretrained` create the model with the hub_id with default behavior: load weigths for models, create config for weightless things) and this correspond so the `default_creation_method` in ComponentSpec
* `from_config` is used to explicitly create from config without weights, acccept both a config dict and path

this sounds fine to me

yiyixuxu: >IMO from_config is the preferred option because it works well with ComponentSpec's default_creation_method as wel

One note on this though  — `default_creation_method` is currently aligned with the current `from_pretrained`/`from_config` usage. e.g. the default creation method would be `from_pretrained`  for weightless components as long as it needs to load a config from hub, `image processor` uses `from_config` because it's created from a config dict defined in componenet spec directly

not sure if this complicates anything so just something to keep in mind

PR #13125, opened 2026-02-11
Fix CogVideoX scheduler prev_timestep for non-leading spacing

Mr-Neutr0n:
## Summary

- **Bug:** `CogVideoXDDIMScheduler.step()` and `CogVideoXDPMScheduler.step()` compute `prev_timestep` using `timestep - num_train_timesteps // num_inference_steps`, which only works correctly when `timestep_spacing="leading"`. For `"linspace"` or `"trailing"` spacing, timesteps are not uniformly spaced by that stride, so the formula produces incorrect previous timestep values and wrong `alpha_prod_t_prev` lookups, leading to degraded or incorrect denoising.

- **Fix:** Replace the hardcoded arithmetic with a `previous_timestep()` method that looks up the actual next entry in `self.timesteps`, matching the approach already used by `DDPMScheduler.previous_timestep()`.

## Files Changed

- `src/diffusers/schedulers/scheduling_ddim_cogvideox.py` — use `previous_timestep()` in `step()`, add method
- `src/diffusers/schedulers/scheduling_dpm_cogvideox.py` — use `previous_timestep()` in `step()`, add method

## Test plan

- [ ] Verify existing CogVideoX scheduler tests pass with no regressions
- [ ] Run `CogVideoXDDIMScheduler` with `timestep_spacing="trailing"` and confirm `prev_timestep` values now match the actual timestep schedule
- [ ] Run `CogVideoXDPMScheduler` with `timestep_spacing="linspace"` and confirm correct behavior

### DIFF
--- a/src/diffusers/schedulers/scheduling_ddim_cogvideox.py
+++ b/src/diffusers/schedulers/scheduling_ddim_cogvideox.py
@@ -383,7 +383,7 @@ class CogVideoXDDIMScheduler(SchedulerMixin, ConfigMixin):
         # - pred_prev_sample -> "x_t-1"
 
         # 1. get previous step value (=t-1)
-        prev_timestep = timestep - self.config.num_train_timesteps // self.num_inference_steps
+        prev_timestep = self.previous_timestep(timestep)
 
         # 2. compute alphas, betas
         alpha_prod_t = self.alphas_cumprod[timestep]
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
Friendly ping on this fix. The current hardcoded stride formula `timestep - num_train_timesteps // num_inference_steps` produces incorrect alpha_prod_t_prev lookups when timestep_spacing is not 'leading'. The `previous_timestep()` method added here mirrors the approach already used in DDPMScheduler. Would be great to get a review when you have a moment.

qgallouedec:

PR #13128, opened 2026-02-11
fix: use state_dict pattern in ZImageControlNet.from_transformer to prevent weight sharing

### LINKED ISSUE #13077: ZImageControlNet.from_transformer creates a shallow copy of the transformer weights
### Describe the bug

When using the [.from_transformer](https://github.com/huggingface/diffusers/blob/1b8fc6c5895ec73f29a5a3caf05c4fdc60450085/src/diffusers/models/controlnets/controlnet_z_image.py#L519) classmethod of the [ZImageControlNetModel](https://github.com/huggingface/diffusers/blob/1b8fc6c5895ec73f29a5a3caf05c4fdc60450085/src/diffusers/models/controlnets/controlnet_z_image.py#L432), the layers that are copied from the transformer are shallow copies. This means that any changes done to the controlnet weights after copying will also affect the transformer weights.

I'm unclear on what the best solution would be. Looking at [FluxControlNetModen.from_transformer](https://github.com/huggingface/diffusers/blob/430c557b6a66a3c2b5740fb186324cb8a9f0f2e9/src/diffusers/models/controlnets/controlnet_flux.py#L122), they have already prepared the layers in the `__init__` function, and therefore can directly load in using state dicts, but this is harder here, given how the `__init__` and `from_transformer` functions have been made in `ZImageControlNetModel`. Short term, I've been able to circumvent it by using `copy.deepcopy(...)`, but I think something more elegant is possible here.

### Reproduction

```python
import torch
from diffusers import ZImageControlNetModel, ZImageTransformer2DModel

transformer = ZImageTransformer2DModel.from_pretrained(
    "Tongyi-MAI/Z-Image",
    subfolder="transformer",
    torch_dtype=torch.bfloat16,   
)

controlnet = ZImageControlNetModel(
    control_layers_places=[0, 15, 29],                  # These parameters are not relevant, just for instantiation 
    control_refiner_layers_places=[0, 1],               # ^
    add_control_noise_refiner="control_noise_refiner",  # ^
    control_in_dim=16                                   # ^
)

controlnet = ZImageControlNetModel.from_transformer(
    controlnet=controlnet,
    transformer=transformer
)

# Weights before modification
print("Transformer t_embedder weights before modification:")
print(transformer.t_embedder.mlp[0].weight)

# Fill ControlNet t_embedder weights with constant value
torch.nn.init.constant_(controlnet.t_embedder.mlp[0].weight, 42.0)
print("ControlNet t_embedder weights after modification:")
print(controlnet.t_embedder.mlp[0].weight) # As expected, filled with 42s

# Should remain unchanged, but is also filled with 42s
print("Transformer t_embedder weights after ControlNet modification:")
print(transformer.t_embedder.mlp[0].weight)
```

### Logs

```shell
Transformer t_embedder weights before modification:
Parameter containing:
tensor([[ 0.0069, -0.0051,  0.0036,  ...,  0.0325, -0.0153, -0.0089],
        [ 0.0004, -0.0102,  0.0078,  ...,  0.0262,  0.0229,  0.0505],
        [-0.0024,  0.0032, -0.0040,  ...,  0.0237,  0.0195,  0.0017],
        ...,
        [-0.0059, -0.0013,  0.0008,  ...,  0.0223,  0.0209,  0.0132],
        [ 0.0026, -0.0014, -0.0052,  ...,  0.0021,  0.0142,  0.0630],
        [-0.0004,  0.0034,  0.0032,  ...,  0.0339,  0.0187,  0.0254]],
       dtype=torch.bfloat16, requires_grad=True)
ControlNet t_embedder weights after modification:
Parameter containing:
tensor([[42., 42., 42.,  ..., 42., 42., 42.],
        [42., 42., 42.,  ..., 42., 42., 42.],
        [42., 42., 42.,  ..., 42., 42., 42.],
        ...,
        [42., 42., 42.,  ..., 42., 42., 42.],
        [42., 42., 42.,  ..., 42., 42., 42.],
        [42., 42., 42.,  ..., 42., 42., 42.]], dtype=torch.bfloat16,
       requires_grad=True)
Transformer t_embedder weights after ControlNet modification:
Parameter containing:
tensor([[42., 42., 42.,  ..., 42., 42., 42.],
        [42., 42., 42.,  ..., 42., 42., 42.],
        [42., 42., 42.,  ..., 42., 42., 42.],
        ...,
        [42., 42., 42.,  ..., 42., 42., 42.],
        [42., 42., 42.,  ..., 42., 42., 42.],
        [42., 42., 42.,  ..., 42., 42., 42.]], dtype=torch.bfloat16,
       requires_grad=True)
```

### System Info

- 🤗 Diffusers version: 0.37.0.dev0
- Platform: Linux-5.15.0-91-generic-x86_64-with-glibc2.35
- Python version: 3.11.14
- PyTorch version (GPU?): 2.9.0+cu126 (True)
- Jax version: not installed
- Transformers version: 4.57.1
- Accelerate version: 1.11.0
- PEFT version: 0.17.1
- Bitsandbytes version: 0.48.2
- Safetensors version: 0.6.2
- Accelerator: NVIDIA RTX 6000 Ada Generation, 49140 MiB
NVIDIA RTX A6000, 49140 MiB
NVIDIA RTX A6000, 49140 MiB
- Using GPU in script?: No

### Who can help?

_No response_

jscaldwell55:
## Summary

`ZImageControlNetModel.from_transformer()` uses direct assignment to copy modules from the transformer, creating shared references. Modifying controlnet weights inadvertently mutates the original transformer. This PR replaces the shallow copies with fresh module instantiation + `load_state_dict()`, matching the established pattern used by `FluxControlNetModel` and `QwenImageControlNetModel`.

## Problem

The current implementation:
```python
controlnet.t_embedder = transformer.t_embedder
controlnet.noise_refiner = transformer.noise_refiner
# ... etc
```

This creates shared references; both objects point to the same `nn.Module` instances. Training the controlnet corrupts the transformer weights.

## Solution

For each module category, use the approach appropriate to its type:

| Type | Old (broken) | New (fixed) |
|------|-------------|-------------|
| `nn.Module` (t_embedder, cap_embedder, noise_refiner, etc.) | `controlnet.x = transformer.x` | Create fresh module, then `load_state_dict(transformer.x.state_dict())` |
| `nn.Parameter` (x_pad_token, cap_pad_token) | `controlnet.x = transformer.x` | `nn.Parameter(transformer.x.data.clone())` |
| `RopeEmbedder` (not nn.Module, no learnable weights) | `controlnet.x = transformer.x` | Fresh instance from `transformer.config` |
| Scalar (t_scale) | `controlnet.x = transformer.x` | Unchanged — immutable float |

This follows the exact pattern used by:
- [`FluxControlNetModel.from_transformer`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/controlnets/controlnet_flux.py#L122)  
- [`QwenImageControlNetModel.from_transformer`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/controlnets/controlnet_qwenimage.py#L99)

### Why not `copy.deepcopy()`?

`deepcopy` technically fixes the symptom but is the wrong tool here:
- Copies internal PyTorch bookkeeping (hooks, grad_fn references, autograd metadata) that shouldn't be transferred
- Higher peak memory - duplicates the entire object graph recursively
- Not the established pattern anywhere in the diffusers controlnet codebase
- The `state_dict` approach copies only the tensor data, which is exactly what we want

## Testing

Added 10 regression tests in `tests/models/controlnets/test_models_controlnet_z_image.py` that verify:
- Each shared module is independent (modifying controlnet doesn't affect transformer)
- Weights are correctly copied
- t_scale has correct value

The reproduction from #13077 now passes; modifying controlnet weights no longer affects the transformer:

```python
import torch
from diffusers import ZImageControlNetModel, ZImageTransformer2DModel

transformer = ZImageTransformer2DModel.from_pretrained(
    "Tongyi-MAI/Z-Image", subfolder="transformer", torch_dtype=torch.bfloat16,
)
controlnet = ZImageControlNetModel(
    control_layers_places=[0, 15, 29],
    control_refiner_layers_places=[0, 1],
    add_control_noise_refiner="control_noise_refiner",
    control_in_dim=16,
)
controlnet = ZImageControlNetModel.from_transformer(controlnet=controlnet, transformer=transformer)

# Modify controlnet weights
original_weight = transformer.t_embedder.mlp[0].weight.clone()
torch.nn.init.constant_(controlnet.t_embedder.mlp[0].weight, 42.0)

# Transformer should be unaffected
assert torch.equal(transformer.t_embedder.mlp[0].weight, original_weight), "Transformer weights were corrupted!"
print("✓ Transformer weights remain independent after controlnet modification")
```

Fixes #13077

cc: @yiyixuxu @JerryWu-code

### DIFF
--- a/tests/models/controlnets/test_models_controlnet_z_image.py
+++ b/tests/models/controlnets/test_models_controlnet_z_image.py
@@ -1,0 +1,197 @@
+# coding=utf-8
+# Copyright 2025 HuggingFace Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import torch
+
+from diffusers.models.controlnets.controlnet_z_image import ZImageControlNetModel
+from diffusers.models.transformers.transformer_z_image import ZImageTransformer2DModel
+
+
+class ZImageControlNetFromTransformerTests(unittest.TestCase):
+    """Tests for ZImageControlNetModel.from_transformer weight independence.
+
+    Verifies that from_transformer creates independent copies of weights,
+    so modifying the controlnet does not affect the original transformer.
+    Regression test for https://github.com/huggingface/diffusers/issues/13077
+    """
+
+    def get_transformer_config(self):
+        return {
+            "all_patch_size": (2,),
+            "all_f_patch_size": (1,),
+            "in_channels": 16,
+            "dim": 256,
+            "n_layers": 2,
+            "n_refiner_layers": 2,
+            "n_heads": 8,
+            "n_kv_heads": 8,
+            "cap_feat_dim": 256,
+            "axes_dims": [8, 12, 12],
+            "axes_lens": [64, 64, 64],
+        }
+
+    def get_controlnet_config(self):
+        return {
+            "control_layers_places": [0, 1],
+            "control_refiner_layers_places": [0, 1],
+            "add_control_noise_refiner": "control_noise_refiner",
+            "control_in_dim": 16,
+            "dim": 256,
+            "n_refiner_layers": 2,
+            "n_heads": 8,
+            "n_kv_heads": 8,
+        }
+
+    def test_t_embedder_independence(self):
+        """Modifying controlnet.t_embedder should not affect transformer.t_embedder."""
+        transformer = ZImageTransformer2DModel(**self.get_transformer_config())
+        controlnet = ZImageControlNetModel(**self.get_controlnet_config())
+        controlnet = ZImageControlNetModel.from_transformer(controlnet=controlnet, transformer=transformer)
+
+        original = transformer.t_embedder.mlp[0].weight.clone()
+        torch.nn.init.constant_(controlnet.t_embedder.mlp[0].weight, 42.0)
+
+        self.assertTrue(
+            torch.equal(transformer.t_embedder.mlp[0].weight, original),
+            "Transformer t_embedder weights were corrupted by controlnet modification",
+        )
+
+    def test_cap_embedder_independence(self):
+        """Modifying controlnet.cap_embedder should not affect transformer.cap_embedder."""
+        transformer = ZImageTransformer2DModel(**self.get_transformer_config())
+        controlnet = ZImageControlNetModel(**self.get_controlnet_config())
+        controlnet = ZImageControlNetModel.from_transformer(controlnet=controlnet, transformer=transformer)
+
+        original = transformer.cap_embedder[1].weight.clone()
+        torch.nn.init.constant_(controlnet.cap_embedder[1].weight, 42.0)
+
+        self.assertTrue(
+            torch.equal(transformer.cap_embedder[1].weight, original),
+            "Transformer cap_embedder weights were corrupted by controlnet modification",
+        )
+
+    def test_all_x_embedder_independence(self):
+        """Modifying controlnet.all_x_embedder should not affect transformer.all_x_embedder."""
+        transformer = ZImageTransformer2DModel(**self.get_transformer_config())
+        controlnet = ZImageControlNetModel(**self.get_controlnet_config())
+        controlnet = ZImageControlNetModel.from_transformer(controlnet=controlnet, transformer=transformer)
+
+        first_key = list(transformer.all_x_embedder.keys())[0]
+        original = transformer.all_x_embedder[first_key].weight.clone()
+        torch.nn.init.constant_(controlnet.all_x_embedder[first_key].weight, 42.0)
+
+        self.assertTrue(
+            torch.equal(transformer.all_x_embedder[first_key].weight, original),
+            "Transformer all_x_embedder weights were corrupted by controlnet modification",
+        )
+
+    def test_noise_refiner_independence(self):
+        """Modifying controlnet.noise_refiner should not affect transformer.noise_refiner."""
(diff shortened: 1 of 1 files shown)


hlky:
It is not intended to train these modules, see original training script https://github.com/aigc-apps/VideoX-Fun/blob/main/scripts/z_image_fun/train_turbo_control_2.1.sh#L35
With shallow copy we save on VRAM usage, and furthermore, if these modules were trained following this PR, how will they load afterwards? This is not handled
IMO the simplest solution here is to just not train these modules, don't use `.train()` and instead manually set `requires_grad` on the trainable `control_*` modules as in the original training script

bghira:
literally *can't* train those modules or the controlnet will never converge

jscaldwell55:
@hlky @bghira  I appreciate the feedback. I took a look, but I think there may be some confusion -- that script uses ZImageControlTransformer2DModel from videox_fun.models, which is a different class/codebase from diffusers. This PR addresses ZImageControlNetModel.from_transformer(). 

  The diffusers from_transformer() method creates a controlnet from a transformer for inference/experimentation. The shallow copy causes silent weight corruption if the controlnet is modified (even accidentally). FluxControlNet and QwenImageControlNet both use load_state_dict for this same pattern.

 Happy to add a docstring note that these modules are typically frozen during training workflows, but the API itself should be safe by default.

  See:

  - [FluxControlNetModel.from_transformer](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/controlnets/controlnet_flux.py#L122)
  - [QwenImageControlNetModel.from_transformer](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/controlnets/controlnet_qwenimage.py#L99)

jscaldwell55:
I followed up on this when i had some more time and you both are correct; I didn't realize that the overall intent is that the modules be shared/frozen. I can see now that the practice is to keep those transformer-derived modulesuntrained, and train only control modules.

jscaldwell55:
Closing this PR and opened #13136, which takes a different approach based on @hlky @bghira feedback and maintainer intent.  Instead of copying modules, the new PR keeps the shared references but explicitly freezes them with \`requires_grad_(False)\`. This preserves the memory-efficient design while preventing accidental training of shared modules.

PR #13131, opened 2026-02-13
Fix broken attention masking for inner padding in ZImageTransformer2DModel

### LINKED ISSUE #12893: Z-Image text sequence length issue
### Describe the bug

I think there might be an issue with calculating the sequence length of `cap_feat` (which is the text encoder output), and masking it accordingly.

I'm going to use code links from *before* the omni commit, because it's easier to read - but the issue seems exist in both, before and after the omni commit.

https://github.com/huggingface/diffusers/blob/52766e6a6939ac6e74375bde5e19c5e0b90d24c1/src/diffusers/models/transformers/transformer_z_image.py#L593-L611

In these lines, the sequence length for each `cap_feats` sample is put into `cap_item_seqlens` and is then used to create an attention mask in line 611.

The problem with this is that `cap_feats` has been overwritten before in these lines:
https://github.com/huggingface/diffusers/blob/52766e6a6939ac6e74375bde5e19c5e0b90d24c1/src/diffusers/models/transformers/transformer_z_image.py#L552-L560

`cap_feats` therefore isn't anymore what the caller has passed. instead, `cap_feats` has already been padded by `patchify_and_embed`. Therefore, *all* `cap_item_seqlens` are identical to `cap_max_item_seqlen`.

This leads to unmasked text tokens during attention, which probably wasn't the intention here.

### Reproduction

put a breakpoint in this line:
https://github.com/huggingface/diffusers/blob/52766e6a6939ac6e74375bde5e19c5e0b90d24c1/src/diffusers/models/transformers/transformer_z_image.py#L612

all `cap_items_seqlens` are identical, `cap_attn_mask` is all True - even if multiple different text lengths were passed to the transformer.

### Logs

```shell

```

### System Info

-

### Who can help?

@JerryWu-code @RuoyiDu @yiyixuxu

Mr-Neutr0n:
## What does this PR do?

Fixes #12893

The attention masks in `ZImageTransformer2DModel` were not properly masking inner padding tokens. After `patchify_and_embed` pads each sequence to a multiple of `SEQ_MULTI_OF` (32), the sequence lengths used to build attention masks reflected the *padded* lengths instead of the original token counts. As a result, `cap_attn_mask` was effectively all-True even when caption lengths differed across the batch, meaning padding tokens were never excluded from attention.

**Reproduction:** set a breakpoint after the attention mask is constructed in the forward pass and inspect `cap_attn_mask` with varying caption lengths — all entries are True regardless of the actual caption length.

**Root cause:** `_prepare_sequence` computes `item_seqlens = [len(f) for f in feats]` from the already-padded features, then builds the attention mask using those padded lengths. Similarly, `_build_unified_sequence` inherits these padded lengths and builds a unified mask that also fails to exclude inner padding.

**Fix:** Use the `inner_pad_mask` (already computed by `_pad_with_ids` during patchification) to properly exclude padding positions from the attention mask. Concretely:
- In `_prepare_sequence`: change `attn_mask[i, :seq_len] = 1` to `attn_mask[i, :seq_len] = ~inner_pad_mask[i]`, so inner-padded positions are marked as False.
- Return the batched `inner_pad_mask` from `_prepare_sequence` so downstream code can use it.
- In `_build_unified_sequence`: accept and concatenate the inner pad masks for x, cap, and siglip in the same order as the unified sequence, then use the combined mask to build the unified attention mask.

This ensures that padding tokens added for sequence alignment are properly excluded from attention in all stages: the noise refiner, context refiner, siglip refiner, and the main transformer layers.

## Who can review?

@JerryWu-code @RuoyiDu @yiyixuxu

### DIFF
--- a/src/diffusers/models/transformers/transformer_z_image.py
+++ b/src/diffusers/models/transformers/transformer_z_image.py
@@ -788,10 +788,10 @@ class ZImageTransformer2DModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOr
         feats = pad_sequence(feats, batch_first=True, padding_value=0.0)
         freqs_cis = pad_sequence(freqs_cis, batch_first=True, padding_value=0.0)[:, : feats.shape[1]]
 
-        # Attention mask
+        # Attention mask (exclude inner padding from attention)
         attn_mask = torch.zeros((bsz, max_seqlen), dtype=torch.bool, device=device)
         for i, seq_len in enumerate(item_seqlens):
-            attn_mask[i, :seq_len] = 1
+            attn_mask[i, :seq_len] = ~inner_pad_mask[i]
 
         # Noise mask
         noise_mask_tensor = None
@@ -802,7 +802,10 @@ class ZImageTransformer2DModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOr
                 padding_value=0,
             )[:, : feats.shape[1]]
 
-        return feats, freqs_cis, attn_mask, item_seqlens, noise_mask_tensor
+        # Pad inner_pad_mask to batch for downstream use
+        batched_inner_pad_mask = pad_sequence(inner_pad_mask, batch_first=True, padding_value=True)[:, : feats.shape[1]]
+
+        return feats, freqs_cis, attn_mask, item_seqlens, noise_mask_tensor, batched_inner_pad_mask
 
     def _build_unified_sequence(
         self,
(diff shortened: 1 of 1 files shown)


dxqb:
thanks for the PR!
are you involved in any way with the Z-Image authors? If not, I'd appreciate a comment by the original authors @JerryWu-code @RuoyiDu to confirm the bug and that this fix is what they intended.

dxqb:
I'm getting broken outputs with this PR *even with batch size 1*, calling the transformer twice for positive and negative separately. The samples below should be strictly identical, because the bug only affected the output if batch size > 1.

Could you check what might be the cause?

Simple test case:
positive prompt "man", negative prompt empty
Z-Image non-Turbo
1024px
cfg 3
20 steps

first without this PR, second with this PR
grafik
grafik

Mr-Neutr0n:
Thanks for testing this @dxqb — I don't have access to the Z-Image weights to properly validate generation quality, so I'll close this for now. The masking logic clearly needs more nuance than what I implemented here. Would be great to get input from the original authors on how inner padding should interact with attention in their architecture.

PR #13139, opened 2026-02-13
[docs] kernels

stevhliu:
adds a kernels section in the `Accelerate inference` docs with the results:

- cross-linked to `Attention backends` docs which demonstrates support for loading attention kernels with `set_attention_backend`
- defer to the blog post and pipeline integration guide for more details about implementing non-attention kernels since this is more involved and already well-documented there


docs/source/en/optimization/fp16.md
```diff
@@ -248,6 +248,24 @@ Refer to the [diffusers/benchmarks](https://huggingface.co/datasets/diffusers/be
 
 The [diffusers-torchao](https://github.com/sayakpaul/diffusers-torchao#benchmarking-results) repository also contains benchmarking results for compiled versions of Flux and CogVideoX.
 
+## Kernels
+
+[Kernels](https://huggingface.co/docs/kernels/index) is a library for building, distributing, and loading optimized compute kernels on the [Hub](https://huggingface.co/kernels-community). It supports [attention](./attention_backends#set_attention_backend) kernels and custom CUDA kernels for operations like RMSNorm.
+
+The [Diffusers Pipeline Integration](https://github.com/huggingface/kernels/blob/main/skills/cuda-kernels/references/diffusers-integration.md) guide shows how to integrate a kernel. Create a custom optimized attention processor, patch all modules in the model, and inject the kernel into the pipeline.
```
sayakpaul: The kernel skill basically lets users get an agent to write custom kernels for a model and hardware. It's not specific to the attention processor but also other modules as well such RMSNorm. Should we make it clearer?

docs/source/en/optimization/fp16.md
```diff
@@ -248,6 +248,24 @@ Refer to the [diffusers/benchmarks](https://huggingface.co/datasets/diffusers/be
...
+
+> [!TIP]
+> Install the [add cuda-kernels](https://github.com/huggingface/kernels/blob/main/skills/cuda-kernels/SKILL.md) skill to teach Claude or Codex how to write a kernel. The [Custom kernels for all from Codex and Claude](https://huggingface.co/blog/custom-cuda-kernels-agent-skills) blog post covers this in more detail.
+
+For example, a custom RMSNorm kernel with [torch.compile](#torchcompile) speeds up LTX-Video generation 1.43x on an H100.
```
sayakpaul: It wasn't just RMSNorm but also other modules implemented with custom kernels.

stevhliu: i mention RMSNorm as an example only for the benchmark results below

docs/source/en/optimization/fp16.md
stevhliu: lmk if this is clearer!

PR #13144, opened 2026-02-14
[LTX2] Fix wrong lora mixin

asomoza:
# What does this PR do?

LTX2 has the incorrect lora mixing which prevents using LTX2 lora with it. This PR fixes that.

## Who can review?

@sayakpaul @dg845


sayakpaul (approved):

dg845 (approved):
Thanks!

PR #13132, opened 2026-02-13
[CI] Fix `setuptools` `pkg_resources` Bug for PR GPU Tests

dg845:
# What does this PR do?

This PR is a follow-up to #13129 which attempts to fix the same error for the `pr_tests_gpu.yml` workflow, which performs GPU tests for PRs.

The fix is currently the same as in the previous PR, but I'm not sure how to confirm that the previous fix is indeed working - I manually ran `push_tests.yml` in https://github.com/huggingface/diffusers/actions/runs/21971315738 on current `main` (5f3ea22513d63d0914cf8946d47722cda6180ff6) with that PR merged and it looks like `Torch CUDA Tests (others)` still has the same errors (https://github.com/huggingface/diffusers/actions/runs/21971315738/job/63473305519):

```bash
 =========================== short test summary info ============================
FAILED tests/others/test_dependencies.py::DependencyTester::test_backend_registration - RuntimeError: Failed to import diffusers.pipelines.stable_diffusion_k_diffusion.pipeline_stable_diffusion_k_diffusion because of the following error (look up to see its traceback):
No module named 'pkg_resources'
FAILED tests/others/test_dependencies.py::DependencyTester::test_pipeline_imports - RuntimeError: Failed to import diffusers.pipelines.stable_diffusion_k_diffusion.pipeline_stable_diffusion_k_diffusion because of the following error (look up to see its traceback):
No module named 'pkg_resources'
============ 2 failed, 69 passed, 11 skipped, 14 warnings in 21.42s ============
```

I think the fix is theoretically correct (assuming no mistakes or misunderstandings) but it seems like it might not have worked.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@sayakpaul
@DN6

PR #13129, opened 2026-02-12
[CI] Fix `setuptools` `pkg_resources` Errors

dg845:
# What does this PR do?

`setuptools` removed the `pkg_resources` module in version 81 (see e.g. [this page in the `setuptools` docs](https://setuptools.pypa.io/en/latest/deprecated/pkg_resources.html)), but some dependencies have not migrated away from it yet and thus fail with an error like

```bash
ModuleNotFoundError: No module named 'pkg_resources'
```

when testing dependencies on the CI (example: https://github.com/huggingface/diffusers/actions/runs/21893622907/job/63204915410?pr=12976).

This PR attempts to fix these errors in the CI by installing those dependencies with a previous version of `setuptools` that does contain `pkg_resources`, while installing all other packages as normal.

Packages currently known to have this issue:
- `k-diffusion` (pinned at `k-diffusion==0.0.12`)

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@sayakpaul
@DN6

PR #13149, opened 2026-02-16
[CI] Revert `setuptools` CI Fix as the Failing Pipelines are Deprecated

dg845:
# What does this PR do?

Follow-up to #13129 and #13132 which maybe finally fixes the `ModuleNotFoundError: No module named 'pkg_resources'` CI error.

I think the previous fix didn't work because I assumed the error was triggered when installing the packages with the issue (such as `k-diffusion`), but I think the error actually happens at runtime when we try to import from the packages. This is because the [`clip-anytorch` dependency](https://github.com/crowsonkb/k-diffusion/blob/master/setup.cfg) of `k-diffusion` will attempt to explicitly import `pkg_resources` in `setup.py`:

https://github.com/rom1504/CLIP/blob/4610075309864ea7aaec976c58ccd69048b64c1e/setup.py#L3

So I think the solution is to pin `setuptools` to a version which still has `pkg_resources` so that the import succeeds.

## Who can review?

@sayakpaul
@DN6

### DIFF
--- a/src/diffusers/models/transformers/transformer_flux2.py
+++ b/src/diffusers/models/transformers/transformer_flux2.py
@@ -424,7 +424,7 @@ class Flux2SingleTransformerBlock(nn.Module):
         self,
         hidden_states: torch.Tensor,
         encoder_hidden_states: torch.Tensor | None,
-        temb_mod_params: tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+        temb_mod: torch.Tensor,
         image_rotary_emb: tuple[torch.Tensor, torch.Tensor] | None = None,
         joint_attention_kwargs: dict[str, Any] | None = None,
         split_hidden_states: bool = False,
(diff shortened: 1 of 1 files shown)


dg845:
Ran the `pr_tests_gpu.yml` workflow manually (https://github.com/huggingface/diffusers/actions/runs/22047985704) and it looks like the issue is fixed: https://github.com/huggingface/diffusers/actions/runs/22047985704/job/63700281914

> Is k_diffusion still used?

From a previous CI run with the error (https://github.com/huggingface/diffusers/actions/runs/21981663925/job/63505654736?pr=12524) it looks like it's still being used in `diffusers.pipelines.stable_diffusion_k_diffusion.pipeline_stable_diffusion_k_diffusion`.

sayakpaul:
We should consider deprecating it and removing it from testing IMO like these:
https://huggingface.co/docs/diffusers/main/en/api/pipelines/attend_and_excite

DN6:
https://github.com/huggingface/diffusers/blob/b0dc51da31a112e5cd4e8a3ee9eafff08dab913d/src/diffusers/pipelines/stable_diffusion_k_diffusion/pipeline_stable_diffusion_xl_k_diffusion.py#L86

The pipelines are deprecated. Let's remove it. Opened a PR #13152

PR #13155, opened 2026-02-17
[CI] Add ftfy as a test dependency

DN6:
# What does this PR do?
We had `k-diffusion` as a test dependency, which in turn installed `clip-anytorch` which had `ftfy` as a dependency. This is why all the tests with pipelines using `ftfy` were passing even though we never had it in our test deps. 

We removed `k-diffusion` from the library since it's no longer maintained and this resulted in `ftfy` also getting inadvertently removed.  

Adding it back here as a test dep. 

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/.github/workflows/pr_modular_tests.yml
+++ b/.github/workflows/pr_modular_tests.yml
@@ -117,7 +117,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        uv pip install -e ".[quality]"
+        uv pip install -e ".[quality,test]"
         #uv pip uninstall transformers huggingface_hub && uv pip install --prerelease allow -U transformers@git+https://github.com/huggingface/transformers.git
         uv pip uninstall transformers huggingface_hub && uv pip install transformers==4.57.1
         uv pip uninstall accelerate && uv pip install -U accelerate@git+https://github.com/huggingface/accelerate.git --no-deps


.github/workflows/pr_modular_tests.yml
```diff
@@ -117,7 +117,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        uv pip install -e ".[quality]"
+        uv pip install -e ".[quality,test]"
```
sayakpaul, line 120: What's happening here? 

We already do it right?
https://github.com/huggingface/diffusers/blob/bcbbded7c3fc873343a7c8f8a63d91d5c727a4a3/docker/diffusers-pytorch-cpu/Dockerfile#L28

.github/workflows/pr_tests.yml
```diff
@@ -114,7 +114,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        uv pip install -e ".[quality]"
+        uv pip install -e ".[quality,test]"
```
sayakpaul, line 117: Same.

.github/workflows/pr_modular_tests.yml
dg845, line 120: Should this change also be added to the push workflows such as `push_tests.yml`? For example here:

https://github.com/huggingface/diffusers/blob/42f8cbb85e2a6711686c189933c062b14be22696/.github/workflows/push_tests.yml#L134

DN6: Fine to remove, but we'd have to merge just the changes to `setup.py` with the CI on this PR failing and then run a build and push up the images.

Thought it would be better just remove in a follow up.

DN6: So the reason it's here is to just verify everything is passing. We install test deps when building the Docker images.
https://github.com/huggingface/diffusers/blob/bcbbded7c3fc873343a7c8f8a63d91d5c727a4a3/docker/diffusers-pytorch-cuda/Dockerfile#L47

Will remove this from the workflow once `setup.py` is merged to main and we push up the new images.

PR #13158, opened 2026-02-17
[docs] add docs for qwenimagelayered

stevhliu:
adds some docs for `QwenImageLayeredPipeline`


docs/source/en/api/pipelines/qwenimage.md
```diff
@@ -91,6 +91,46 @@ image.save("qwen_fewsteps.png")
 > [!TIP]
 > The `guidance_scale` parameter in the pipeline is there to support future guidance-distilled models when they come up. Note that passing `guidance_scale` to the pipeline is ineffective. To enable classifier-free guidance, please pass `true_cfg_scale` and `negative_prompt` (even an empty negative prompt like " ") should enable classifier-free guidance computations.
 
+## Decomposing an image with QwenImageLayeredPipeline
+
+[`QwenImageLayeredPipeline`] decomposes an image into multiple RGBA layers, so you can edit each layer independently without affecting the others.
```
sayakpaul: Should we include this docstring or the one included in the pipeline implementation would suffice? That way, just including

```bash
## QwenImageLayeredPipeline

[[autodoc]] QwenImageLayeredPipeline
  - all
  - __call__

```

would be enough?

stevhliu: ohh yes, you're absolutely right! good suggestion

PR #13161, opened 2026-02-18
[attention backends] change to updated repo and version.

sayakpaul:
# What does this PR do?

For standardization purposes, this PR updates the `sage_attention` repo id to `sage-attention` as the former is now deprecated.

The PR also updates the usage to incorporate `version` instead of `revision` which is a better practice.


src/diffusers/models/attention_dispatch.py
```diff
@@ -272,7 +272,7 @@ class _HubKernelConfig:
 _HUB_KERNELS_REGISTRY: dict["AttentionBackendName", _HubKernelConfig] = {
     # TODO: temporary revision for now. Remove when merged upstream into `main`.
     AttentionBackendName._FLASH_3_HUB: _HubKernelConfig(
-        repo_id="kernels-community/flash-attn3", function_attr="flash_attn_func", revision="fake-ops-return-probs"
+        repo_id="kernels-community/flash-attn3", function_attr="flash_attn_func", revision="v1"
```
danieldk:
```suggestion
        repo_id="kernels-community/flash-attn3", function_attr="flash_attn_func", version=1
```

with kernels >= 0.12

src/diffusers/models/attention_dispatch.py
```diff
@@ -286,7 +286,7 @@ class _HubKernelConfig:
         repo_id="kernels-community/flash-attn2", function_attr="flash_attn_varlen_func", revision=None
     ),
     AttentionBackendName.SAGE_HUB: _HubKernelConfig(
-        repo_id="kernels-community/sage_attention", function_attr="sageattn", revision=None
+        repo_id="kernels-community/sage-attention", function_attr="sageattn", revision=None
```
danieldk: @MekkCyber, I also think `version=1` here?

src/diffusers/models/attention_dispatch.py
sayakpaul: Are we then asking the users to always keep `kernels` to `>=0.12`? Personally, I don't have a problem with it, just want to make sure we're on the same page.

src/diffusers/models/attention_dispatch.py
MekkCyber: yes!

PR #12812, opened 2025-12-09
[core] Enable CP for kernels-based attention backends

sayakpaul:
# What does this PR do?

Adds CP support to the `kernels`-based attention backends.

Our CP support is quickly gaining traction. Currently, we have a few attention backends that are fully based on `kernels`. In order for their adoption to grow and make them a bit more complete in terms of feature parity, I think we should make them CP-compatible, too. 

<details>
<summary>Code to test:</summary>

```py
import argparse
import torch
from torch import distributed as dist
from diffusers import DiffusionPipeline, ContextParallelConfig, AutoModel


CKPT_ID = "black-forest-labs/FLUX.1-dev"

def parse_args():
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--cp-backend",
        type=str,
        choices=["ring", "ulysses", "unified"],
        default="ulysses",
        help="Context parallel backend to use.",
    )
    parser.add_argument(
        "--attn-backend",
        type=str,
        choices=["flash_hub", "_flash_3_hub", "sage_hub"],
        default="flash_hub",
        help="Attention backend to use.",
    )
    return parser.parse_args()


def setup_distributed():
    if not dist.is_initialized():
        dist.init_process_group(backend="nccl")
    rank = dist.get_rank()
    device = torch.device(f"cuda:{rank}")
    torch.cuda.set_device(device)
    return device


def main():
    args = parse_args()

    device = setup_distributed()
    world_size = dist.get_world_size()
    if args.cp_backend == "ring":
        cp_config = ContextParallelConfig(ring_degree=world_size)
    elif args.cp_backend == "unified":
        cp_config = ContextParallelConfig(ring_degree=world_size // 2, ulysses_degree=world_size // 2)
    else:
        cp_config = ContextParallelConfig(ulysses_degree=world_size)

    transformer = AutoModel.from_pretrained(
        CKPT_ID, 
        subfolder="transformer", 
        torch_dtype=torch.bfloat16, 
        parallel_config=cp_config
    )

    pipeline = DiffusionPipeline.from_pretrained(
        CKPT_ID, transformer=transformer, torch_dtype=torch.bfloat16,
    ).to(device)
    pipeline.transformer.set_attention_backend(args.attn_backend)

    prompt = """
    cinematic film still of a cat sipping a margarita in a pool in Palm Springs, California
    highly detailed, high budget hollywood movie, cinemascope, moody, epic, gorgeous, film grain
    """

    generator = torch.Generator().manual_seed(42)
[...]

PR #13165, opened 2026-02-19
[attention backends] use dedicated wrappers from fa3 for cp.

sayakpaul:
# What does this PR do?

Great pointer here https://github.com/huggingface/diffusers/pull/12812#discussion_r2827759302 from @vasqu.

### DIFF
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -62,6 +62,8 @@ _REQUIRED_FLEX_VERSION = "2.5.0"
 _REQUIRED_XLA_VERSION = "2.2"
 _REQUIRED_XFORMERS_VERSION = "0.0.29"
 
+logger = get_logger(__name__)  # pylint: disable=invalid-name
+
 _CAN_USE_FLASH_ATTN = is_flash_attn_available() and is_flash_attn_version(">=", _REQUIRED_FLASH_VERSION)
 _CAN_USE_FLASH_ATTN_3 = is_flash_attn_3_available()
 _CAN_USE_AITER_ATTN = is_aiter_available() and is_aiter_version(">=", _REQUIRED_AITER_VERSION)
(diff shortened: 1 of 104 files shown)


src/diffusers/models/attention_dispatch.py
```diff
@@ -276,7 +276,11 @@ class _HubKernelConfig:
 _HUB_KERNELS_REGISTRY: dict["AttentionBackendName", _HubKernelConfig] = {
     # TODO: temporary revision for now. Remove when merged upstream into `main`.
     AttentionBackendName._FLASH_3_HUB: _HubKernelConfig(
-        repo_id="kernels-community/flash-attn3", function_attr="flash_attn_func", revision="fake-ops-return-probs"
+        repo_id="kernels-community/flash-attn3",
+        function_attr="flash_attn_func",
+        revision="fake-ops-return-probs",
```
vasqu, line 334: Ah didnt notice it before, we don't have it merged into main yet?

src/diffusers/models/attention_dispatch.py
```diff
@@ -1275,55 +1305,50 @@ def _flash_attention_3_hub_backward_op(
...
+
+    wrapped_backward_fn(
+        grad_out,
+        query,
+        key,
+        value,
+        out,
+        softmax_lse,
+        None,
+        None,  # cu_seqlens_q, cu_seqlens_k
```
vasqu, line 1384: Just fmi, there is no fa varlen cp yet? I think meta had a version at some point

vasqu (approved):
LGTM, trusting you that this works as it seems we don't have tests? 

Comments are mostly for my interest / nits

sayakpaul:
> trusting you that this works as it seems we don't have tests?

Yes, I manually verified that this works as we're still growing the feature.

src/diffusers/models/attention_dispatch.py
sayakpaul: We're working on it in https://github.com/huggingface/diffusers/pull/13161.

src/diffusers/models/attention_dispatch.py
sayakpaul: We handle varlen separately + varlens aren't that common in diffusion use cases.

https://github.com/huggingface/diffusers/blob/99daaa802da01ef4cff5141f4f3c0329a57fb591/src/diffusers/models/attention_dispatch.py#L177

PR #13166, opened 2026-02-19
[gguf][torch.compile time] Convert to plain tensor earlier in dequantize_gguf_tensor

anijain2305:
Once `dequantize_gguf_tensor` fetches the quant_type attributes from the GGUFParamter tensor subclass, there is no further need of running the actual dequantize operations on the Tensor subclass, we can just convert to plain tensor right away.

This not only makes PyTorch eager faster, but reduces torch.compile tracer compile time from 36 seconds to 10 seconds for auroflow quantized model, because there is lot less code to trace now.

cc @yiyixuxu @sayakpaul

### DIFF
--- a/src/diffusers/quantizers/gguf/utils.py
+++ b/src/diffusers/quantizers/gguf/utils.py
@@ -516,6 +516,9 @@ def dequantize_gguf_tensor(tensor):
 
     block_size, type_size = GGML_QUANT_SIZES[quant_type]
 
+    # Conver to plain tensor to avoid unnecessary __torch_function__ overhead.
+    tensor = tensor.as_tensor()
+
     tensor = tensor.view(torch.uint8)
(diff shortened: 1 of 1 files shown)


src/diffusers/quantizers/gguf/utils.py
```diff
@@ -516,6 +516,9 @@ def dequantize_gguf_tensor(tensor):
 
     block_size, type_size = GGML_QUANT_SIZES[quant_type]
 
+    # Conver to plain tensor to avoid unnecessary __torch_function__ overhead.
+    tensor = tensor.as_tensor()
```
sayakpaul, line 520: Does it have any impact on quality?

DN6 (approved):
Good catch! Thanks @anijain2305 👍🏽

PR #13171, opened 2026-02-22
[attention backends] remove non-hub attention backends.

sayakpaul:
# What does this PR do?

Now that the Hub-based attention backends (Flash, Flash 3, SAGE) have been fully feature-complete (`torch.compile`, CP, etc.), we can confidently remove the non-Hub variants.

This has several advantages:

* More Hub-centric
* Eliminates complex installation nightmares
* Lesser maintainence burden (otherwise we would have two different code paths for each of the features for each attention backend)
* Helps us improve `kernels` as well

Cc: @danieldk @LysandreJik

### DIFF
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -34,12 +34,7 @@ from ..utils import (
     get_logger,
     is_aiter_available,
     is_aiter_version,
-    is_flash_attn_3_available,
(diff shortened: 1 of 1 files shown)


sayakpaul:
I will run the tests in `test_attention_backends.py` as well.

missionfloyd:
I suppose this makes it even more cloud-dependent?

PR #13174, opened 2026-02-23
[tests] add attention backend tests.

sayakpaul:
# What does this PR do?

Add attention backend tests.


tests/models/testing_utils/attention.py
```diff
@@ -14,22 +14,105 @@
...
 )
 
+# All backends under test.
+_ALL_BACKEND_PARAMS = [_PARAM_NATIVE, _PARAM_NATIVE_CUDNN, _PARAM_FLASH_HUB, _PARAM_FLASH_3_HUB]
+
+# Backends that only accept bf16/fp16 inputs; models and inputs must be cast before running them.
+_BF16_REQUIRED_BACKENDS = {
+    AttentionBackendName._NATIVE_CUDNN,
+    AttentionBackendName.FLASH_HUB,
+    AttentionBackendName._FLASH_3_HUB,
```
DN6: Include Sage as well?

tests/models/testing_utils/attention.py
```diff
@@ -179,3 +261,208 @@ def test_attention_processor_count_mismatch_raises_error(self):
...
+            rtol=0,
+            msg=(
+                f"Output from model.set_attention_backend('{backend.value}') should be identical "
+                f"to the output from `with attention_backend('{backend.value}'):`."
+            ),
+        )
+
+    @torch.no_grad()
+    @pytest.mark.parametrize("backend", _ALL_BACKEND_PARAMS)
+    def test_output_close_to_native(self, backend):
```
DN6: Might want to remove native from the parametrize, since you'd just be comparing it with itself.

tests/models/testing_utils/attention.py
```diff
@@ -14,22 +14,105 @@
...
+# Backends that perform non-deterministic operations and therefore cannot run when
+# torch.use_deterministic_algorithms(True) is active (e.g. after enable_full_determinism()).
+_NON_DETERMINISTIC_BACKENDS = {AttentionBackendName._NATIVE_CUDNN}
+
+
+def _maybe_cast_to_bf16(backend, model, inputs_dict):
+    """Cast model and floating-point inputs to bfloat16 when the backend requires it."""
+    if backend not in _BF16_REQUIRED_BACKENDS:
+        return model, inputs_dict
+    model = model.to(dtype=torch.bfloat16)
```
DN6: One thing to note here is this casting will overwrite the `_keep_in_fp32_modules`. Practically shouldn't be an issue since these tests run with dummy models, but might want to account for this when casting the model params.

tests/models/testing_utils/attention.py
```diff
@@ -179,3 +261,208 @@ def test_attention_processor_count_mismatch_raises_error(self):
...
+        with attention_backend(backend):
+            ctx_output = model(**inputs_dict, return_dict=False)[0]
+
+        initial_registry_backend, _ = _AttentionBackendRegistry.get_active_backend()
+
+        try:
+            model.set_attention_backend(backend.value)
+        except Exception as e:
+            logger.warning("Skipping test for backend '%s': %s", backend.value, e)
+            pytest.skip(str(e))
```
DN6: Why skip on exception? It would be better to just raise so that potential issues get surfaced?

tests/models/testing_utils/attention.py
```diff
@@ -14,22 +14,105 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Module-level backend parameter sets for AttentionBackendTesterMixin
+# ---------------------------------------------------------------------------
+
+_CUDA_AVAILABLE = torch.cuda.is_available()
+_KERNELS_AVAILABLE = is_kernels_available()
```
DN6: Nit. `is_kernels_available()` function already returns a cached variable. So caching the output here isn't really necessary.

tests/models/testing_utils/attention.py
```diff
@@ -179,3 +261,208 @@ def test_attention_processor_count_mismatch_raises_error(self):
...
+            )
+
+        restored_backend, _ = _AttentionBackendRegistry.get_active_backend()
+        assert restored_backend == initial_backend, (
+            f"Backend should be restored to {initial_backend} after exiting the context manager, "
+            f"got {restored_backend}."
+        )
+
+    @pytest.mark.parametrize("backend", _ALL_BACKEND_PARAMS)
+    def test_compile(self, backend):
```
DN6: Add `is_compile` decorator so that we can use marks to filter out if needed.

tests/models/testing_utils/attention.py
```diff
@@ -179,3 +261,208 @@ def test_attention_processor_count_mismatch_raises_error(self):
...
+
+    # test_output_close_to_native: alternate backends (flash, cuDNN) may
+    # accumulate small numerical errors vs the reference PyTorch SDPA kernel.
+    backend_vs_native_atol: float = 1e-2
+    backend_vs_native_rtol: float = 1e-2
+
+    # test_compile: regional compilation introduces the same kind of numerical
+    # error as the non-compiled backend path, so the same loose tolerance applies.
+    compile_vs_native_atol: float = 1e-2
+    compile_vs_native_rtol: float = 1e-2
```
DN6: I would just set these as default args within the relevant test function. 
`def test_attention_backend(..., atol=1e-2, rtol=1-2)`

Easier for an inheriting class to change the tolerance as needed.

sayakpaul:
@DN6, all your comments are addressed. LMK if anything is not addressed. I purposefully discarded SAGE to prefer the ones that I feel are more prevalent.

tests/models/testing_utils/attention.py
```diff
@@ -14,22 +14,104 @@
...
+    AttentionBackendName._NATIVE_CUDNN,
+    AttentionBackendName.FLASH_HUB,
+    AttentionBackendName._FLASH_3_HUB,
+}
+
+# Backends that perform non-deterministic operations and therefore cannot run when
+# torch.use_deterministic_algorithms(True) is active (e.g. after enable_full_determinism()).
+_NON_DETERMINISTIC_BACKENDS = {AttentionBackendName._NATIVE_CUDNN}
+
+
```
sayakpaul, line 80:
```suggestion

# TODO: revist this when https://github.com/huggingface/diffusers/pull/13182
# is merged.

```

sayakpaul:
@DN6 a gentle ping.

DN6 (approved):

PR #13190, opened 2026-02-26
Fix None dereference crashes and mutable default feat_idx in Wan/Cosmos autoencoders

jashshah999:
## What does this PR do?

Fixes None dereference crashes and a mutable default argument bug:

**1. `models/model_loading_utils.py`** - `_fetch_remapped_cls_from_config` used chained `.get().get()` without a fallback on the first call. If `previous_class_name` is not in `_CLASS_REMAPPING_DICT`, the first `.get()` returns `None` and the second `.get()` crashes with `AttributeError`.

**2. `models/autoencoders/autoencoder_kl_cosmos.py`** - Both `CosmosPatchEmbed3d` and `CosmosUnpatcher3d` call `_WAVELETS.get(patch_method).clone()`. If `patch_method` is not a recognized wavelet type, `.get()` returns `None` and `.clone()` crashes with `AttributeError`.

**3. `models/autoencoders/autoencoder_kl_wan.py`** - 8 `forward()` methods used `feat_idx=[0]` as a mutable default argument. Since the list is actively mutated via `feat_idx[0] += 1` during forward passes, the shared default list accumulates state across calls, corrupting the cache index on subsequent invocations.

### DIFF
--- a/src/diffusers/models/autoencoders/autoencoder_kl_cosmos.py
+++ b/src/diffusers/models/autoencoders/autoencoder_kl_cosmos.py
@@ -111,7 +111,10 @@ class CosmosPatchEmbed3d(nn.Module):
         self.patch_size = patch_size
         self.patch_method = patch_method
 
-        wavelets = _WAVELETS.get(patch_method).clone()
+        wavelets = _WAVELETS.get(patch_method)
+        if wavelets is None:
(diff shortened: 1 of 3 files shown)


jashshah999:
Friendly ping — just checking if this is ready for review when you get a chance. Happy to address any feedback!

jashshah999:
Gentle ping -- this fixes actual None dereference crashes in the Wan/Cosmos autoencoders. Happy to address any feedback.

PR #13193, opened 2026-02-27
[modular] fallback to default_blocks_name when loading base block classes in ModularPipeline

yiyixuxu:
so that we are able to correctly load repos like https://huggingface.co/diffusers/flux2-modular

When creating a pipeline using `SequentialPipelineBlocks.from_blocks_dict()` and saving it, the `modular_model_index.json` stores `_blocks_class_name` as `"SequentialPipelineBlocks"`. On loading, `from_pretrained` tries to instantiate `SequentialPipelineBlocks()` directly, which has empty `block_classes` and produces a broken pipeline.

```python
pipe = ModularPipeline.from_pretrained("diffusers/flux2-modular")
print(pipe.blocks)
```

under `main`, it won't be able to map to `Flux2AutoBlocks`

```
>>> print(pipe.blocks)
SequentialPipelineBlocks(
  Class: ModularPipelineBlocks

  Description: 


  Sub-Blocks:
)
```

This PR added a fallback behavior in `ModularPipeline.__init__ method` so if the resolved blocks class is not importable from our library or has empty `block_classes` (i.e., it's a base class like `SequentialPipelineBlocks`, `ConditionalPipelineBlocks`, etc.), fall back to the pipeline's `default_blocks_name` .

### DIFF
--- a/tests/modular_pipelines/test_modular_pipelines_common.py
+++ b/tests/modular_pipelines/test_modular_pipelines_common.py
@@ -699,3 +699,27 @@ class TestLoadComponentsSkipBehavior:
 
         # Verify test_component was not loaded
         assert not hasattr(pipe, "test_component") or pipe.test_component is None
+
+
+class TestModularPipelineInitFallback:
+    """Test that ModularPipeline.__init__ falls back to default_blocks_name when
+    _blocks_class_name is a base class (e.g. SequentialPipelineBlocks saved by from_blocks_dict)."""
+
+    def test_init_fallback_when_blocks_class_name_is_base_class(self, tmp_path):
+        # 1. Load pipeline and get a workflow (returns a base SequentialPipelineBlocks)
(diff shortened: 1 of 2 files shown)


tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -699,3 +699,27 @@ def test_load_components_skips_invalid_pretrained_path(self):
 
         # Verify test_component was not loaded
         assert not hasattr(pipe, "test_component") or pipe.test_component is None
+
+
+class TestModularPipelineInitFallback:
```
sayakpaul, line 704: This is perfect test! I will try to make it available to other pipelines in https://github.com/huggingface/diffusers/pull/13192.

sayakpaul (approved):
Was just thinking about this today. Thanks!

PR #13195, opened 2026-02-27
[modular]Update model card to include workflow

yiyixuxu:
before https://huggingface.co/YiYiXu/modular-loader-t2i-0704
after https://huggingface.co/YiYiXu/modular-loader-t2i-20260227

the input/output isn't too meaningful for conditional/auto pipeline so it is better we show them for each workflow

here is an example with pipeline that does not have workflow 
https://huggingface.co/diffusers/Qwen-Image-Layered-modular

### DIFF
--- a/tests/modular_pipelines/test_modular_pipelines_common.py
+++ b/tests/modular_pipelines/test_modular_pipelines_common.py
@@ -629,35 +629,6 @@ class TestModularModelCardContent:
 
         assert content["trigger_inputs_section"] == ""
 
-    def test_blocks_description_with_sub_blocks(self):
-        """Test that blocks with sub-blocks are correctly described."""
-
-        class MockBlockWithSubBlocks:
-            def __init__(self):
-                self.__class__.__name__ = "ParentBlock"
-                self.description = "Parent block"
-                self.sub_blocks = {
-                    "child1": self.create_child_block("ChildBlock1", "Child 1 description"),
-                    "child2": self.create_child_block("ChildBlock2", "Child 2 description"),
-                }
-
-            def create_child_block(self, name, desc):
-                class ChildBlock:
-                    def __init__(self):
-                        self.__class__.__name__ = name
-                        self.description = desc
-
-                return ChildBlock()
-
-        blocks = self.create_mock_blocks()
-        blocks.sub_blocks["parent"] = MockBlockWithSubBlocks()
-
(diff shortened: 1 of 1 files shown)


yiyixuxu:
cc @sayakpaul 
I wanted to include workflow into the modelcard now that we support them
can you take a look here? feel free to push directly into this PR if you see things want to change

src/diffusers/modular_pipelines/modular_pipeline_utils.py
```diff
@@ -1109,63 +1142,74 @@ def generate_modular_model_card_content(blocks) -> dict[str, Any]:
...
+        # Per-workflow details
+        for wf_name, trigger_inputs in workflow_map.items():
+            trigger_input_names = set(trigger_inputs.keys())
+            try:
+                workflow_blocks = blocks.get_workflow(wf_name)
+            except Exception:
+                parts.append(f"<details>\n<summary><strong>{wf_name}</strong></summary>\n")
+                parts.append(f"> **Trigger inputs**: {', '.join(f'`{t}`' for t in trigger_input_names)}\n")
+                parts.append("*Could not resolve workflow blocks.*\n")
+                parts.append("</details>\n")
```
sayakpaul, line 1175: Do we need both? 

[Here](https://huggingface.co/YiYiXu/modular-loader-t2i-20260227), we already have:

```
text2image: requires prompt
```

Then having a collapsible block (essentially with the same info)

Image

seems to be redundant to me.

yiyixuxu: updated, they look like this now https://huggingface.co/YiYiXu/modular-loader-t2i-20260227

src/diffusers/utils/hub_utils.py
```diff
@@ -141,9 +145,17 @@ def load_or_create_model_card(
             " To install it, please run `pip install Jinja2`."
         )
 
+    if update_model_card and not is_modular:
+        raise ValueError("`update_model_card=True` is only supported for modular pipelines (`is_modular=True`).")
+
     try:
         # Check if the model card is present on the remote repo
         model_card = ModelCard.load(repo_id_or_path, token=token)
+        # For modular pipelines, regenerate card content when requested (preserve existing metadata)
```
sayakpaul, line 154: Yes very cool approach!

sayakpaul (approved):
Perfect

PR #13204, opened 2026-03-02
[modular] not pass trust_remote_code to external repos 

yiyixuxu:
based on this discussion
https://github.com/huggingface/blog/pull/3278#discussion_r2873972272

### DIFF
--- a/src/diffusers/modular_pipelines/modular_pipeline.py
+++ b/src/diffusers/modular_pipelines/modular_pipeline.py
@@ -2256,7 +2256,10 @@ class ModularPipeline(ConfigMixin, PushToHubMixin):
                 new_component_spec = current_component_spec
                 if hasattr(self, name) and getattr(self, name) is not None:
                     logger.warning(f"ModularPipeline.update_components: setting {name} to None (spec unchanged)")
-            elif current_component_spec.default_creation_method == "from_pretrained" and getattr(component, "_diffusers_load_id", None) is None
+            elif (
+                current_component_spec.default_creation_method == "from_pretrained"
+                and getattr(component, "_diffusers_load_id", None) is None
+            ):
                 new_component_spec = ComponentSpec(name=name, type_hint=type(component))
             else:
                 new_component_spec = ComponentSpec.from_component(name, component)


src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -2323,17 +2325,45 @@ def load_components(self, names: list[str] | str | None = None, **kwargs):
...
+                    warning_msg += (
+                        f"\nNote: `trust_remote_code=True` was not passed to `{name}` because it comes from "
+                        f"an external repository (`{spec.pretrained_model_name_or_path}`). For safety, `trust_remote_code` is only forwarded to "
+                        f"components from the same repository as the pipeline.\n\n"
+                        f"To load this component manually:\n"
+                        f"  from diffusers import AutoModel\n"
+                        f'  {name} = AutoModel.from_pretrained("{spec.pretrained_model_name_or_path}", trust_remote_code=True)\n'
+                        f"  pipe.update_components({name} = {name})\n"
+                    )
+                logger.warning(warning_msg)
```
sayakpaul: Of course, we cannot generalize to all possible combinations but what if the underlying component is a scheduler or a non-modeling component?

For schedulers, we could have `AutoScheduler` or something so that users won't have to remember a specific class name. For others (such as tokenizers, image processors), we already have "auto" classes.

I guess it's not that big a use case yet, but just wanted to flag.

sayakpaul: Also, should we error instead? If the user-side logger is not configured appropriately then this warning could go unnoticed.

tests/modular_pipelines/test_modular_pipelines_custom_blocks.py
```diff
@@ -192,6 +192,156 @@ def test_custom_block_supported_components(self):
...
+
+            # auto_map will reference the module name derived from ExternalRefBlock.__module__,
+            # which is "test_modular_pipelines_custom_blocks". Write the code file with that name.
+            code_path = os.path.join(pipeline_repo_dir, "test_modular_pipelines_custom_blocks.py")
+            with open(code_path, "w") as f:
+                f.write(EXTERNAL_REF_BLOCK_CODE_STR)
+
+            block = ModularPipelineBlocks.from_pretrained(pipeline_repo_dir, trust_remote_code=True)
+            pipe = block.init_pipeline()
+            pipe.save_pretrained(pipeline_repo_dir)
```
sayakpaul, line 318: Should we also inspect the `modular_model_index.json` file here (checking if the repo paths were serialized as expected)?

sayakpaul (commented):
I like it! 

I left some thoughts to discuss. LMK if anything is unclear.

src/diffusers/modular_pipelines/modular_pipeline.py
yiyixuxu: i updated the message
tried to make it error but reverted back to warning for now: it would be a little annoying that you won't be able to use `load_components()` because one element is not loadable

tests/modular_pipelines/test_modular_pipelines_custom_blocks.py
yiyixuxu: I think if we can load it back, we are fine (we loaded it back next and verified the path too)

src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -2254,6 +2256,10 @@ def update_components(self, **kwargs):
                 new_component_spec = current_component_spec
                 if hasattr(self, name) and getattr(self, name) is not None:
                     logger.warning(f"ModularPipeline.update_components: setting {name} to None (spec unchanged)")
+            elif current_component_spec.default_creation_method == "from_pretrained" and not (
+                hasattr(component, "_diffusers_load_id") and component._diffusers_load_id is not None
+            ):
```
DN6: IMO a bit easier to read. 
```suggestion
            elif current_component_spec.default_creation_method == "from_pretrained" and getattr(component, "_diffusers_load_id", None) is None
```

DN6 (approved):
Thanks 👍🏽

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -687,6 +687,18 @@ def test_load_components_selective_loading(self):
         assert pipe.unet is not None
         assert getattr(pipe, "vae", None) is None
 
+    def test_load_components_selective_loading_incremental(self):
```
sayakpaul, line 690: Do we also test if `pipe` doesn't have other expected components (like `scheduler`)?

src/diffusers/modular_pipelines/modular_pipeline.py
```diff
@@ -2325,17 +2332,49 @@ def load_components(self, names: list[str] | str | None = None, **kwargs):
...
+                        f"`{spec.pretrained_model_name_or_path}`.\n\n"
+                        f"`trust_remote_code=True` was not forwarded to `{name}` because it comes from "
+                        f"a different repository than the pipeline (`{self._pretrained_model_name_or_path}`). "
+                        f"For safety, `trust_remote_code` is only forwarded to components from the same "
+                        f"repository as the pipeline.\n\n"
+                        f"You need to load this component manually with `trust_remote_code=True` and pass it "
+                        f"to the pipeline via `pipe.update_components()`. For example, if it is a custom model:\n\n"
+                        f'  {name} = AutoModel.from_pretrained("{spec.pretrained_model_name_or_path}", trust_remote_code=True)\n'
+                        f"  pipe.update_components({name}={name})\n"
+                    )
```
sayakpaul, line 2367: This is perfect, thank you!

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -749,6 +761,36 @@ def test_save_pretrained_roundtrip_with_local_model(self, tmp_path):
         for key in original_state_dict:
             assert torch.equal(original_state_dict[key], loaded_state_dict[key]), f"Mismatch in {key}"
 
+    def test_save_pretrained_updates_index_for_model_with_no_load_id(self, tmp_path):
+        """testing the workflow of update the pipeline with a custom model and save the pipeline,
+        the modular_model_index.json should point to the save directory."""
+        import json
+
+        from diffusers import UNet2DConditionModel
```
sayakpaul, line 769: (nit): maybe we could shift all the imports to the top.

tests/modular_pipelines/test_modular_pipelines_common.py
```diff
@@ -749,6 +761,36 @@ def test_save_pretrained_roundtrip_with_local_model(self, tmp_path):
         for key in original_state_dict:
             assert torch.equal(original_state_dict[key], loaded_state_dict[key]), f"Mismatch in {key}"
 
+    def test_save_pretrained_updates_index_for_model_with_no_load_id(self, tmp_path):
```
sayakpaul, line 764: Very cool test!

PR #13207, opened 2026-03-04
cogvideo example: Distribute VAE video encoding across processes in CogVideoX LoRA training

jiqing-feng:
Currently, every process encodes the entire dataset through the VAE during preprocessing. This is redundant in multi-GPU training — each process does the same full encoding work independently.

This PR shards the VAE encoding across processes so that each process only encodes 1/num_processes of the videos, then broadcasts the results. This reduces the preprocessing time proportionally to the number of processes (e.g., ~4x faster with 4 GPUs) with no change to training behavior.

Single-GPU training is unaffected.

### DIFF
--- a/examples/cogvideo/train_cogvideox_lora.py
+++ b/examples/cogvideo/train_cogvideox_lora.py
@@ -1232,21 +1232,47 @@ def main(args):
         id_token=args.id_token,
     )
 
-    def encode_video(video, bar):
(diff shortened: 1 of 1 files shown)


sayakpaul (approved):

sayakpaul:
@bot /style

sayakpaul:
Thanks for your contributions!

PR #13208, opened 2026-03-04
Add Helios-14B Video Generation Pipelines

dg845:
# What does this PR do?

This PR adds pipelines for the Helios-14B video generation model ([project page](https://pku-yuangroup.github.io/Helios-Page/), [official code](https://github.com/PKU-YuanGroup/Helios), [model weights](https://huggingface.co/collections/BestWishYsh/helios)). Helios is a 14B parameter video generation model which can generate minute-scale video at `19.5` FPS on a single H100 GPU. Both a full model and distilled model are available, which correspond to `HeliosPipeline` and `HeliosPyramidPipeline` respectively; each pipeline supports text-to-video (T2V), image-to-video (I2V), and video-to-video (V2V) tasks.

T2V inference example:

```python
import torch
from diffusers import AutoModel, HeliosPipeline
from diffusers.hooks.group_offloading import apply_group_offloading
from diffusers.utils import export_to_video

vae = AutoModel.from_pretrained("BestWishYsh/Helios-Base", subfolder="vae", torch_dtype=torch.float32)

# group-offloading
pipeline = HeliosPipeline.from_pretrained(
    "BestWishYsh/Helios-Base",
    vae=vae,
    torch_dtype=torch.bfloat16
)
pipeline.enable_group_offload(
    onload_device=torch.device("cuda"),
    offload_device=torch.device("cpu"),
    offload_type="block_level",
    num_blocks_per_group=1,
    use_stream=True,
    record_stream=True,
)

prompt = """
A vibrant tropical fish swimming gracefully among colorful coral reefs in a clear, turquoise ocean. The fish has bright blue 
and yellow scales with a small, distinctive orange spot on its side, its fins moving fluidly. The coral reefs are alive with 
a variety of marine life, including small schools of colorful fish and sea turtles gliding by. The water is crystal clear, 
allowing for a view of the sandy ocean floor below. The reef itself is adorned with a mix of hard and soft corals in shades 
of red, orange, and green. The photo captures the fish from a slightly elevated angle, emphasizing its lively movements and 
the vivid colors of its surroundings. A close-up shot with dynamic movement.
"""
negative_prompt = """
Bright tones, overexposed, static, blurred details, subtitles, style, works, paintings, images, static, overall gray, worst quality,
low quality, JPEG compression residue, ugly, incomplete, extra fingers, poorly drawn hands, poorly drawn faces, deformed, disfigured,
misshapen limbs, fused fingers, still picture, messy background, three legs, many people in the background, walking backwards
"""

[...]

PR #13211, opened 2026-03-04
Fix group-offloading bug

SHYuanBest:
Related to https://github.com/huggingface/diffusers/pull/13208 and https://github.com/huggingface/diffusers/pull/13210

cc @yiyixuxu
cc @sayakpaul
cc @dg845

### DIFF
--- a/tests/hooks/test_group_offloading.py
+++ b/tests/hooks/test_group_offloading.py
@@ -621,16 +621,34 @@ class ConditionalModuleGroupOffloadTests(GroupOffloadTests):
         )
 
     @parameterized.expand([("leaf_level",), ("block_level",)])
+    @unittest.skipIf(
+        torch.device(torch_device).type not in ["cuda", "xpu"],
+        "Test requires a CUDA or XPU device.",
+    )
     def test_conditional_modules_with_stream(self, offload_type: str):
         """Regression test: conditionally-executed modules must not cause device mismatch when using streams.
 
-        The first forward pass runs WITHOUT optional_input, so optional_proj_1 and optional_proj_2
-        are not executed and are not included in the lazy prefetch execution order. The second forward
-        pass runs WITH optional_input, which means these modules ARE called. Without the fix, their
-        weights remain on CPU while the input tensor is on GPU, causing a RuntimeError.
+        The model contains two optional Linear layers (optional_proj_1, optional_proj_2) that are only
+        executed when `optional_input` is provided. This simulates modules like patch_short/patch_mid/
+        patch_long in HeliosTransformer3DModel, which are only called when history latents are present.
+
+        When using streams, `LazyPrefetchGroupOffloadingHook` traces the execution order on the first
+        forward pass and sets up a prefetch chain so each module pre-loads the next one's weights.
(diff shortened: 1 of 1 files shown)


sayakpaul:
Could we also add a test for this code path in https://github.com/huggingface/diffusers/blob/main/tests/hooks/test_group_offloading.py?

SHYuanBest:
> Could we also add a test for this code path in https://github.com/huggingface/diffusers/blob/main/tests/hooks/test_group_offloading.py?

done.

tests/hooks/test_group_offloading.py
```diff
@@ -566,3 +566,111 @@ def get_autoencoder_kl_config(self, block_out_channels=None, norm_num_groups=Non
...
+        torch.manual_seed(0)
+        return DummyModelWithConditionalModules(
+            in_features=self.in_features,
+            hidden_features=self.hidden_features,
+            out_features=self.out_features,
+            num_layers=self.num_layers,
+        )
+
+    @parameterized.expand([("leaf_level",), ("block_level",)])
+    def test_conditional_modules_with_stream(self, offload_type: str):
```
sayakpaul, line 628: Let's use `unittest.skipIf` marker to skip the tests properly.

tests/hooks/test_group_offloading.py
```diff
@@ -566,3 +566,111 @@ def get_autoencoder_kl_config(self, block_out_channels=None, norm_num_groups=Non
...
+            out_ref_no_opt = model_ref(x, optional_input=None)
+            out_no_opt = model(x, optional_input=None)
+            self.assertTrue(
+                torch.allclose(out_ref_no_opt, out_no_opt, atol=1e-5),
+                f"[{offload_type}] Outputs do not match on first pass (no optional_input).",
+            )
+
+            # Second forward pass WITH optional_input — optional_proj_1/2 ARE now called.
+            # This is the regression case: without the fix, a RuntimeError is raised here
+            # because the weights of optional_proj_1/2 are still on CPU.
```
sayakpaul:
```suggestion
```

tests/hooks/test_group_offloading.py
```diff
@@ -566,3 +566,111 @@ def get_autoencoder_kl_config(self, block_out_channels=None, norm_num_groups=Non
...
+        )
+
+    @parameterized.expand([("leaf_level",), ("block_level",)])
+    def test_conditional_modules_with_stream(self, offload_type: str):
+        """Regression test: conditionally-executed modules must not cause device mismatch when using streams.
+
+        The first forward pass runs WITHOUT optional_input, so optional_proj_1 and optional_proj_2
+        are not executed and are not included in the lazy prefetch execution order. The second forward
+        pass runs WITH optional_input, which means these modules ARE called. Without the fix, their
+        weights remain on CPU while the input tensor is on GPU, causing a RuntimeError.
```
sayakpaul: Can we also add a comment on why we have to invoke the models multiple times to test this properly?

SHYuanBest: have added

tests/hooks/test_group_offloading.py
SHYuanBest: have modified

PR #13217, opened 2026-03-06
Add Support for LTX-2.3 Models

dg845:
# What does this PR do?

This PR adds support for LTX-2.3 ([official code](https://github.com/Lightricks/LTX-2), [model weights](https://huggingface.co/Lightricks/LTX-2.3)), a new model in the LTX-2.X family of audio-video models. LTX-2.3 has improved audio and visual quality and prompt adherence as compared to LTX-2.0. 

<details>
    <summary>T2V Example</summary>

```python
import torch
from diffusers import LTX2Pipeline

model_id = "dg845/LTX-2.3-Diffusers"
device = "cuda"
dtype = torch.bfloat16
seed = 42

frame_rate = 24.0
width = 768
height = 512
num_inference_steps = 30

prompt = (
    "An astronaut hatches from a fragile egg on the surface of the Moon, the shell cracking and peeling apart in "
    "gentle low-gravity motion. Fine lunar dust lifts and drifts outward with each movement, floating in slow arcs "
    "before settling back onto the ground. The astronaut pushes free in a deliberate, weightless motion, small "
    "fragments of the egg tumbling and spinning through the air. In the background, the deep darkness of space subtly "
    "shifts as stars glide with the camera's movement, emphasizing vast depth and scale. The camera performs a "
    "smooth, cinematic slow push-in, with natural parallax between the foreground dust, the astronaut, and the "
    "distant starfield. Ultra-realistic detail, physically accurate low-gravity motion, cinematic lighting, and a "
    "breath-taking, movie-like shot."
)
negative_prompt = (
    "blurry, out of focus, overexposed, underexposed, low contrast, washed out colors, excessive noise, "
    "grainy texture, poor lighting, flickering, motion blur, distorted proportions, unnatural skin tones, "
    "deformed facial features, asymmetrical face, missing facial features, extra limbs, disfigured hands, "
    "wrong hand count, artifacts around text, inconsistent perspective, camera shake, incorrect depth of "
    "field, background too sharp, background clutter, distracting reflections, harsh shadows, inconsistent "
    "lighting direction, color banding, cartoonish rendering, 3D CGI look, unrealistic materials, uncanny "
    "valley effect, incorrect ethnicity, wrong gender, exaggerated expressions, wrong gaze direction, "
    "mismatched lip sync, silent or muted audio, distorted voice, robotic voice, echo, background noise, "
    "off-sync audio, incorrect dialogue, added dialogue, repetitive speech, jittery movement, awkward "
    "pauses, incorrect timing, unnatural transitions, inconsistent framing, tilted camera, flat lighting, "
    "inconsistent tone, cinematic oversaturation, stylized filters, or AI artifacts."
)

pipe = LTX2Pipeline.from_pretrained(model_id, torch_dtype=dtype)
pipe.vae.enable_tiling()

generator = torch.Generator(device).manual_seed(seed)
video, audio = pipe(
    prompt=prompt,
    width=width,
    height=height,
    num_frames=121,
    frame_rate=frame_rate,
    guidance_scale=3.0,
    stg_scale=1.0,
    modality_scale=3.0,
    guidance_rescale=0.7,
    audio_guidance_scale=7.0,
    audio_stg_scale=1.0,
    audio_modality_scale=3.0,
    audio_guidance_rescale=0.7,
    spatio_temporal_guidance_blocks=[28],
    generator=generator,
    output_type="np",
    return_dict=False,
)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
    output_path="ltx2_3_t2v.mp4",
)
```
</details>

<details>
    <summary>I2V Example</summary>

```python
import torch
from diffusers import LTX2ImageToVideoPipeline

model_id = "dg845/LTX-2.3-Diffusers"
device = "cuda"
dtype = torch.bfloat16
seed = 42

frame_rate = 24.0
width = 768
height = 512
num_inference_steps = 30

prompt = (
    "An astronaut hatches from a fragile egg on the surface of the Moon, the shell cracking and peeling apart in "
    "gentle low-gravity motion. Fine lunar dust lifts and drifts outward with each movement, floating in slow arcs "
    "before settling back onto the ground. The astronaut pushes free in a deliberate, weightless motion, small "
    "fragments of the egg tumbling and spinning through the air. In the background, the deep darkness of space subtly "
    "shifts as stars glide with the camera's movement, emphasizing vast depth and scale. The camera performs a "
    "smooth, cinematic slow push-in, with natural parallax between the foreground dust, the astronaut, and the "
    "distant starfield. Ultra-realistic detail, physically accurate low-gravity motion, cinematic lighting, and a "
    "breath-taking, movie-like shot."
)
negative_prompt = (
    "blurry, out of focus, overexposed, underexposed, low contrast, washed out colors, excessive noise, "
    "grainy texture, poor lighting, flickering, motion blur, distorted proportions, unnatural skin tones, "
    "deformed facial features, asymmetrical face, missing facial features, extra limbs, disfigured hands, "
    "wrong hand count, artifacts around text, inconsistent perspective, camera shake, incorrect depth of "
    "field, background too sharp, background clutter, distracting reflections, harsh shadows, inconsistent "
    "lighting direction, color banding, cartoonish rendering, 3D CGI look, unrealistic materials, uncanny "
    "valley effect, incorrect ethnicity, wrong gender, exaggerated expressions, wrong gaze direction, "
    "mismatched lip sync, silent or muted audio, distorted voice, robotic voice, echo, background noise, "
    "off-sync audio, incorrect dialogue, added dialogue, repetitive speech, jittery movement, awkward "
    "pauses, incorrect timing, unnatural transitions, inconsistent framing, tilted camera, flat lighting, "
    "inconsistent tone, cinematic oversaturation, stylized filters, or AI artifacts."
)

image = load_image(
    "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/astronaut.jpg"
)

pipe = LTX2ImageToVideoPipeline.from_pretrained(model_id, torch_dtype=dtype)
pipe.vae.enable_tiling()

generator = torch.Generator(device).manual_seed(seed)
video, audio = pipe(
    image=image,
    prompt=prompt,
    width=width,
    height=height,
    num_frames=121,
    frame_rate=frame_rate,
    guidance_scale=3.0,
    stg_scale=1.0,
    modality_scale=3.0,
    guidance_rescale=0.7,
    audio_guidance_scale=7.0,
    audio_stg_scale=1.0,
    audio_modality_scale=3.0,
    audio_guidance_rescale=0.7,
    spatio_temporal_guidance_blocks=[28],
    generator=generator,
    output_type="np",
    return_dict=False,
)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
    output_path="ltx2_3_i2v.mp4",
)
```

</details>

<details>
    <summary>FLF2V Example</summary>

```python
import torch
from diffusers import LTX2ConditionPipeline
from diffusers.pipelines.ltx2.pipeline_ltx2_condition import LTX2VideoCondition

model_id = "dg845/LTX-2.3-Diffusers"
device = "cuda"
dtype = torch.bfloat16
seed = 42

frame_rate = 24.0
width = 768
height = 512
num_inference_steps = 30

prompt = (
    "CG animation style, a small blue bird takes off from the ground, flapping its wings. The bird's feathers are "
    "delicate, with a unique pattern on its chest. The background shows a blue sky with white clouds under bright "
    "sunshine. The camera follows the bird upward, capturing its flight and the vastness of the sky from a close-up, "
    "low-angle perspective."
)
negative_prompt = (
    "blurry, out of focus, overexposed, underexposed, low contrast, washed out colors, excessive noise, "
    "grainy texture, poor lighting, flickering, motion blur, distorted proportions, unnatural skin tones, "
    "deformed facial features, asymmetrical face, missing facial features, extra limbs, disfigured hands, "
    "wrong hand count, artifacts around text, inconsistent perspective, camera shake, incorrect depth of "
    "field, background too sharp, background clutter, distracting reflections, harsh shadows, inconsistent "
    "lighting direction, color banding, cartoonish rendering, 3D CGI look, unrealistic materials, uncanny "
    "valley effect, incorrect ethnicity, wrong gender, exaggerated expressions, wrong gaze direction, "
    "mismatched lip sync, silent or muted audio, distorted voice, robotic voice, echo, background noise, "
    "off-sync audio, incorrect dialogue, added dialogue, repetitive speech, jittery movement, awkward "
    "pauses, incorrect timing, unnatural transitions, inconsistent framing, tilted camera, flat lighting, "
    "inconsistent tone, cinematic oversaturation, stylized filters, or AI artifacts."
)

first_image = load_image(
    "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/flf2v_input_first_frame.png"
)
last_image = load_image(
    "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/flf2v_input_last_frame.png"
)
first_cond = LTX2VideoCondition(frames=first_image, index=0, strength=1.0)
last_cond = LTX2VideoCondition(frames=last_image, index=-1, strength=1.0)
conditions = [first_cond, last_cond]

pipe = LTX2ConditionPipeline.from_pretrained(model_id, torch_dtype=dtype)
pipe.vae.enable_tiling()

generator = torch.Generator(device).manual_seed(seed)
video, audio = pipe(
    conditions=conditions,
    prompt=prompt,
    width=width,
    height=height,
    num_frames=121,
    frame_rate=frame_rate,
    guidance_scale=3.0,
    stg_scale=1.0,
    modality_scale=3.0,
    guidance_rescale=0.7,
    audio_guidance_scale=7.0,
    audio_stg_scale=1.0,
    audio_modality_scale=3.0,
    audio_guidance_rescale=0.7,
    spatio_temporal_guidance_blocks=[28],
    generator=generator,
    output_type="np",
    return_dict=False,
)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
    output_path="ltx2_3_flf2v.mp4",
)
```

</details>

<details open>
    <summary>I2V Two Stage Example</summary>

```python
import torch
from diffusers import LTX2ImageToVideoPipeline, LTX2LatentUpsamplePipeline
from diffusers.pipelines.ltx2 import LTX2LatentUpsamplerModel
from diffusers.pipelines.ltx2.utils import STAGE_2_DISTILLED_SIGMA_VALUES

model_id = "dg845/LTX-2.3-Diffusers"
device = "cuda"
dtype = torch.bfloat16
seed = 42

frame_rate = 24.0
width = 768
height = 512
num_inference_steps = 30

prompt = (
    "An astronaut hatches from a fragile egg on the surface of the Moon, the shell cracking and peeling apart in "
    "gentle low-gravity motion. Fine lunar dust lifts and drifts outward with each movement, floating in slow arcs "
    "before settling back onto the ground. The astronaut pushes free in a deliberate, weightless motion, small "
    "fragments of the egg tumbling and spinning through the air. In the background, the deep darkness of space subtly "
    "shifts as stars glide with the camera's movement, emphasizing vast depth and scale. The camera performs a "
    "smooth, cinematic slow push-in, with natural parallax between the foreground dust, the astronaut, and the "
    "distant starfield. Ultra-realistic detail, physically accurate low-gravity motion, cinematic lighting, and a "
    "breath-taking, movie-like shot."
)
negative_prompt = (
    "blurry, out of focus, overexposed, underexposed, low contrast, washed out colors, excessive noise, "
    "grainy texture, poor lighting, flickering, motion blur, distorted proportions, unnatural skin tones, "
    "deformed facial features, asymmetrical face, missing facial features, extra limbs, disfigured hands, "
    "wrong hand count, artifacts around text, inconsistent perspective, camera shake, incorrect depth of "
    "field, background too sharp, background clutter, distracting reflections, harsh shadows, inconsistent "
    "lighting direction, color banding, cartoonish rendering, 3D CGI look, unrealistic materials, uncanny "
    "valley effect, incorrect ethnicity, wrong gender, exaggerated expressions, wrong gaze direction, "
    "mismatched lip sync, silent or muted audio, distorted voice, robotic voice, echo, background noise, "
    "off-sync audio, incorrect dialogue, added dialogue, repetitive speech, jittery movement, awkward "
    "pauses, incorrect timing, unnatural transitions, inconsistent framing, tilted camera, flat lighting, "
    "inconsistent tone, cinematic oversaturation, stylized filters, or AI artifacts."
)

image = load_image(
    "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/astronaut.jpg"
)

pipe = LTX2ImageToVideoPipeline.from_pretrained(model_id, torch_dtype=dtype)
pipe.vae.enable_tiling()

generator = torch.Generator(device).manual_seed(seed)
video_latent, audio_latent = pipe(
    image=image,
    prompt=prompt,
    width=width,
    height=height,
    num_frames=121,
    frame_rate=frame_rate,
    guidance_scale=3.0,
    stg_scale=1.0,
    modality_scale=3.0,
    guidance_rescale=0.7,
    audio_guidance_scale=7.0,
    audio_stg_scale=1.0,
    audio_modality_scale=3.0,
    audio_guidance_rescale=0.7,
    spatio_temporal_guidance_blocks=[28],
    use_cross_timestep=True,
    generator=generator,
    output_type="latent",
    return_dict=False,
)

latent_upsampler = LTX2LatentUpsamplerModel.from_pretrained(
    "dg845/LTX-2.3-Spatial-Upsampler-Diffusers",
    subfolder="latent_upsampler",
    torch_dtype=dtype,
)
upsample_pipe = LTX2LatentUpsamplePipeline(vae=pipe.vae, latent_upsampler=latent_upsampler)
upsample_pipe.enable_model_cpu_offload(device=device)
upscaled_video_latent = upsample_pipe(
    latents=video_latent,
    output_type="latent",
    return_dict=False,
)[0]

pipe.load_lora_weights(
    "Lightricks/LTX-2.3",
    adapter_name="stage_2_distilled",
    weight_name="ltx-2.3-22b-distilled-lora-384.safetensors",
)
pipe.set_adapters("stage_2_distilled", 1.0)
# Change scheduler to use Stage 2 distilled sigmas as is
new_scheduler = FlowMatchEulerDiscreteScheduler.from_config(
    pipe.scheduler.config, use_dynamic_shifting=False, shift_terminal=None
)
pipe.scheduler = new_scheduler

video, audio = pipe(
    image=image,
    latents=upscaled_video_latent,
    audio_latents=audio_latent,
    prompt=prompt,
    width=width * 2,
    height=height * 2,
    num_frames=121,
    frame_rate=frame_rate,
    num_inference_steps=3,
    noise_scale=STAGE_2_DISTILLED_SIGMA_VALUES[0],
    sigmas=STAGE_2_DISTILLED_SIGMA_VALUES,
    guidance_scale=1.0,  # For Stage 2 distilled, disable all guidance
    stg_scale=0.0,
    modality_scale=1.0,
    guidance_rescale=0.0,
    audio_guidance_scale=1.0,
    audio_stg_scale=0.0,
    audio_modality_scale=1.0,
    audio_guidance_rescale=0.0,
    spatio_temporal_guidance_blocks=None,
    use_cross_timestep=True,
    generator=generator,
    output_type="np",
    return_dict=False,
)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
    output_path="ltx2_3_i2v_two_stage.mp4",
)
```

</details>

<details>
    <summary>I2V Distilled Example</summary>

```python
import torch
from diffusers import LTX2ImageToVideoPipeline, LTX2LatentUpsamplePipeline
from diffusers.pipelines.ltx2 import LTX2LatentUpsamplerModel
from diffusers.pipelines.ltx2.utils import DISTILLED_SIGMA_VALUES, STAGE_2_DISTILLED_SIGMA_VALUES

model_id = "dg845/LTX-2.3-Distilled-Diffusers"
device = "cuda"
dtype = torch.bfloat16
seed = 42

frame_rate = 24.0
width = 768
height = 512
num_inference_steps = 30

prompt = (
    "An astronaut hatches from a fragile egg on the surface of the Moon, the shell cracking and peeling apart in "
    "gentle low-gravity motion. Fine lunar dust lifts and drifts outward with each movement, floating in slow arcs "
    "before settling back onto the ground. The astronaut pushes free in a deliberate, weightless motion, small "
    "fragments of the egg tumbling and spinning through the air. In the background, the deep darkness of space subtly "
    "shifts as stars glide with the camera's movement, emphasizing vast depth and scale. The camera performs a "
    "smooth, cinematic slow push-in, with natural parallax between the foreground dust, the astronaut, and the "
    "distant starfield. Ultra-realistic detail, physically accurate low-gravity motion, cinematic lighting, and a "
    "breath-taking, movie-like shot."
)
negative_prompt = None

image = load_image(
    "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/astronaut.jpg"
)

pipe = LTX2ImageToVideoPipeline.from_pretrained(model_id, torch_dtype=dtype)
pipe.vae.enable_tiling()

generator = torch.Generator(device).manual_seed(seed)
video_latent, audio_latent = pipe(
    image=image,
    prompt=prompt,
    width=width,
    height=height,
    num_frames=121,
    frame_rate=frame_rate,
    num_inference_steps=8,
    sigmas=DISTILLED_SIGMA_VALUES,
    guidance_scale=1.0,  # Disable all guidance for distilled inference
    stg_scale=0.0,
    modality_scale=1.0,
    guidance_rescale=0.0,
    audio_guidance_scale=1.0,
    audio_stg_scale=0.0,
    audio_modality_scale=1.0,
    audio_guidance_rescale=0.0,
    spatio_temporal_guidance_blocks=None,
    use_cross_timestep=True,
    generator=generator,
    output_type="latent",
    return_dict=False,
)

latent_upsampler = LTX2LatentUpsamplerModel.from_pretrained(
    "dg845/LTX-2.3-Spatial-Upsampler-Diffusers",
    subfolder="latent_upsampler",
    torch_dtype=dtype,
)
upsample_pipe = LTX2LatentUpsamplePipeline(vae=pipe.vae, latent_upsampler=latent_upsampler)
upsample_pipe.enable_model_cpu_offload(device=device)
upscaled_video_latent = upsample_pipe(
    latents=video_latent,
    output_type="latent",
    return_dict=False,
)[0]

video, audio = pipe(
    image=image,
    latents=upscaled_video_latent,
    audio_latents=audio_latent,
    prompt=prompt,
    width=width * 2,
    height=height * 2,
    num_frames=121,
    frame_rate=frame_rate,
    num_inference_steps=3,
    noise_scale=STAGE_2_DISTILLED_SIGMA_VALUES[0],
    sigmas=STAGE_2_DISTILLED_SIGMA_VALUES,
    guidance_scale=1.0,  # Disable all guidance for distilled inference
    stg_scale=0.0,
    modality_scale=1.0,
    guidance_rescale=0.0,
    audio_guidance_scale=1.0,
    audio_stg_scale=0.0,
    audio_modality_scale=1.0,
    audio_guidance_rescale=0.0,
    spatio_temporal_guidance_blocks=None,
    use_cross_timestep=True,
    generator=generator,
    output_type="np",
    return_dict=False,
)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
    output_path="ltx2_3_i2v_distilled.mp4",
)
```

</details>

## Who can review?

@yiyixuxu
@sayakpaul


src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -217,14 +217,113 @@ def __call__(
         return hidden_states
 
 
+class LTX2PerturbedAttnProcessor:
```
yiyixuxu, line 231: I think this is just a guider https://github.com/huggingface/diffusers/blob/main/src/diffusers/guiders/skip_layer_guidance.py

dg845: Thanks! Looking at the code, it's unclear to me whether `SkipLayerGuidance` currently works for LTX-2.3 for the following reasons:

1. **Not attention backend agnostic**: if I understand correctly, STG is implemented through `AttentionProcessorSkipHook`, which uses `AttentionScoreSkipFunctionMode` to intercept calls to `torch.nn.functional.scaled_dot_product_attention` to simply return the `value`: https://github.com/huggingface/diffusers/blob/e747fe4a942ce379d73a975a82f9e4c484c74ba2/src/diffusers/hooks/layer_skip.py#L93 But I think other attention backends like `flash-attn` won't call that function and thus will not work with `SkipLayerGuidance`.
2. **LTX-2.3 does additional computation on the `value`s**: LTX-2.3 additionally processes the `value`s using learned per-head gates before sending it to the attention output projection `to_out`. This is not supported by the current `SkipLayerGuidance` implementation.

I'm not sure whether these issues can be resolved with changes to the `SkipLayerGuidance` implementation or whether something like a new attention processor would make more sense here.

dg845: I have opened a PR with a possible modification to `SkipLayerGuidance` to allow it to better support LTX-2.3 at #13220.

sayakpaul: This is a good callout! From my understanding, guider as a component doesn't change much. LTX-2 is probably an exception. If more models start to do their own form of SLG, we could think of giving them their own guider classes / attention processors. But for now, I think modifications to the existing SLG class make more sense.

yiyixuxu: let's merge LTx2.3 with a special custom attention processor in this PR first ASAP 

the design from the other PR to refator guider is fundamentally wrong - the purpose of hooks (and guider as well) that it modifies behavior from the outside, without the model needing to be aware & implement logic specific to it 
i will look to refactor with guiders in the follow-up modular PR

dg845:
LTX-2.3 `diffusers` converted checkpoint: [`dg845/LTX-2.3-Diffusers`](https://huggingface.co/dg845/LTX-2.3-Diffusers) (may still have bugs).

dg845:
I2V sample using the example above:

https://github.com/user-attachments/assets/74574867-2bf5-425f-8cb9-39496df33edc

This uses [`dg845/LTX-2.3-Diffusers`](https://huggingface.co/dg845/LTX-2.3-Diffusers/tree/main) with CFG + STG + modality guidance with the LTX-2.3 default guidance scales.

tin2tin:
Tried the i2v example, and got this error:
```
ValueError: Spatio-Temporal Guidance (STG) is specified but no STG blocks are supplied.
Please supply a list of block indices at which to apply STG in `spatio_temporal_guidance_blocks`
```

Adding this seems to make it work: 
```
    spatio_temporal_guidance_blocks=[6,7,8,9,10],
```
I don't know if this is the correct way to solve this, but the examples should properly be updated to deal with this problem. 
A 2. step example could be good to include as well. 2. step cleaned up the pixel jitter pattern seen in your astronaut video.

dg845:
LTX-2.3 `diffusers` latent spatial upsampling pipeline: [`dg845/LTX-2.3-Spatial-Upsampler-Diffusers`](https://huggingface.co/dg845/LTX-2.3-Spatial-Upsampler-Diffusers) (may still have bugs).

dg845:
LTX-2.3 `diffusers` distilled pipeline: [`dg845/LTX-2.3-Distilled-Diffusers`](https://huggingface.co/dg845/LTX-2.3-Distilled-Diffusers).

scripts/convert_ltx2_to_diffusers.py
```diff
@@ -17,7 +17,7 @@
     LTX2Pipeline,
     LTX2VideoTransformer3DModel,
 )
-from diffusers.pipelines.ltx2 import LTX2LatentUpsamplerModel, LTX2TextConnectors, LTX2Vocoder
+from diffusers.pipelines.ltx2 import LTX2LatentUpsamplerModel, LTX2TextConnectors, LTX2Vocoder, LTX2VocoderWithBWE
```
sayakpaul, line 20: Any issue in unifying `LTX2Vocoder` and `LTX2VocoderWithBWE`?

src/diffusers/loaders/lora_conversion_utils.py
```diff
@@ -2156,6 +2156,9 @@ def _convert_non_diffusers_ltx2_lora_to_diffusers(state_dict, non_diffusers_pref
             "scale_shift_table_a2v_ca_audio": "audio_a2v_cross_attn_scale_shift_table",
             "q_norm": "norm_q",
             "k_norm": "norm_k",
+            # LTX-2.3
+            "audio_prompt_adaln_single": "audio_prompt_adaln",
```
sayakpaul, line 2160: Where did this pop up? Distillation checkpoint?

src/diffusers/models/autoencoders/autoencoder_kl_ltx2.py
```diff
@@ -580,6 +582,7 @@ def __init__(
         resnet_eps: float = 1e-6,
         resnet_act_fn: str = "swish",
         spatio_temporal_scale: bool = True,
+        upsample_type: str = "spatiotemporal",
```
sayakpaul, line 585: Should this go at the last of init params to prevent backwards breaking in case someone is using positional arguments?

src/diffusers/models/autoencoders/autoencoder_kl_ltx2.py
```diff
@@ -609,17 +612,38 @@ def __init__(
 
         self.upsamplers = None
         if spatio_temporal_scale:
-            self.upsamplers = nn.ModuleList(
-                [
-                    LTXVideoUpsampler3d(
-                        out_channels * upscale_factor,
+            self.upsamplers = nn.ModuleList()
+
```
sayakpaul, line 616: It seems like `stride` is the only factor that varies depending on `upsampler_type`. So, maybe we could do something like:

```py
if upsample_type == "spatial":
    stride = (1, 2, 2)
elif upsample_type == "temporal":
    stride = (2, 1, 1)
elif upsample_type == "spatio_temporal":
    stride = (2, 2, 2)

self.upsamplers.append(..., strides=strides)
```

WDYT?

src/diffusers/models/transformers/transformer_ltx2.py
sayakpaul: The point on guiders not being backend agnostic is a good thing to keep in mind.

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -212,6 +216,112 @@ def __call__(
...
+            hidden_states = hidden_states.flatten(2, 3)
+
+        hidden_states = attn.to_out[0](hidden_states)
+        hidden_states = attn.to_out[1](hidden_states)
+        return hidden_states
+
+
+class LTX2PerturbedAttnProcessor:
+    r"""
+    Processor which implements attention with perturbation masking and per-head gating for LTX-2.X models.
```
sayakpaul, line 233: 2.3 onwards?

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -212,6 +216,112 @@ def __call__(
...
+    """
+
+    _attention_backend = None
+    _parallel_config = None
+
+    def __init__(self):
+        if is_torch_version("<", "2.0"):
+            raise ValueError(
+                "LTX attention processors require a minimum PyTorch version of 2.0. Please upgrade your PyTorch installation."
+            )
```
sayakpaul, line 243: Good one. I think this is how other processors should do it as well. We wrongly check the existence of `scaled_dot_product_attention()` instead of this. Should update in a follow-up PR.

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -321,16 +438,27 @@ def __init__(
         audio_num_attention_heads: int,
         audio_attention_head_dim,
         audio_cross_attention_dim: int,
+        video_gated_attn: bool = False,
```
sayakpaul, line 441: Should this be called `apply_gated_attention`?

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -321,16 +438,27 @@ def __init__(
...
+        video_cross_attn_adaln: bool = False,
+        audio_gated_attn: bool = False,
```
sayakpaul, line 443: Same as above.

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -422,14 +562,36 @@ def __init__(
...
+        video_mod_param_num = 9 if self.video_cross_attn_adaln else 6
+        audio_mod_param_num = 9 if self.audio_cross_attn_adaln else 6
+        self.scale_shift_table = nn.Parameter(torch.randn(video_mod_param_num, dim) / dim**0.5)
+        self.audio_scale_shift_table = nn.Parameter(torch.randn(audio_mod_param_num, audio_dim) / audio_dim**0.5)
+
+        # Prompt cross-attn (attn2) additional modulation params
+        self.cross_attn_adaln = video_cross_attn_adaln or audio_cross_attn_adaln
+        if self.cross_attn_adaln:
+            self.prompt_scale_shift_table = nn.Parameter(torch.randn(2, dim))
+            self.audio_prompt_scale_shift_table = nn.Parameter(torch.randn(2, audio_dim))
```
sayakpaul, line 578: The code tells me that `video_cross_attn_adaln` and `audio_cross_attn_adaln` can be True or False at the same time together. If so, should `audio_prompt_scale_shift_table` be conditioned `audio_cross_attn_adaln` being True?

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -442,143 +604,181 @@ def forward(
...
+        if self.cross_attn_adaln:
+            video_prompt_ada_params = self.get_mod_params(self.prompt_scale_shift_table, temb_prompt, batch_size)
+            shift_text_kv, scale_text_kv = video_prompt_ada_params
+
+            audio_prompt_ada_params = self.get_mod_params(
+                self.audio_prompt_scale_shift_table, temb_prompt_audio, batch_size
+            )
+            audio_shift_text_kv, audio_scale_text_kv = audio_prompt_ada_params
+
+        # 2.1. Video-Text Cross-Attention (Q: Video; K,V: Test)
```
sayakpaul: nit

```suggestion
        # 2.1. Video-Text Cross-Attention (Q: Video; K,V: Text)
```

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -1152,6 +1395,21 @@ def forward(
...
+                Whether to isolate each modality by turning off cross-modality (audio-to-video and video-to-audio)
+                cross attention (for all blocks). Use for modality guidance in LTX-2.3.
+            spatio_temporal_guidance_blocks (`list[int]`, *optional*, defaults to `None`):
+                The transformer block indices at which to apply spatio-temporal guidance (STG), which shortcuts the
+                self-attention operations by simply using the values rather than the full scaled dot-product attention
+                (SDPA) operation. If `None` or empty, STG will not be applied to any block.
+            perturbation_mask (`torch.Tensor`, *optional*):
+                Perturbation mask for STG of shape `(batch_size,)` or `(batch_size, 1, 1)`. Should be 0 at batch
+                elements where STG should be applied and 1 elsewhere. If STG is being used but `peturbation_mask` is
+                not supplied, will default to applying STG (perturbing) all batch elements.
```
sayakpaul, line 1404: (nit): don't you think having 1 as applying STG is better and less confusing?

src/diffusers/pipelines/ltx2/connectors.py
```diff
@@ -260,24 +339,36 @@ class LTX2TextConnectors(ModelMixin, PeftAdapterMixin, ConfigMixin):
...
+        video_gated_attn: bool = False,
+        audio_connector_num_attention_heads: int = 30,
+        audio_connector_attention_head_dim: int = 128,
+        audio_connector_num_layers: int = 2,
+        audio_connector_num_learnable_registers: int | None = 128,
+        audio_gated_attn: bool = False,
+        connector_rope_base_seq_len: int = 4096,
+        rope_theta: float = 10000.0,
+        rope_double_precision: bool = True,
+        causal_temporal_positioning: bool = False,
```
sayakpaul, line 357: Are these values mostly from LTX-2?

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -494,6 +420,46 @@ def encode_prompt(
 
         return prompt_embeds, prompt_attention_mask, negative_prompt_embeds, negative_prompt_attention_mask
 
+    @torch.no_grad()
+    def enhance_prompt(
+        self,
+        prompt: str,
+        system_prompt: str,
+        max_new_tokens: int = 512,
+        seed: int = 10,
```
sayakpaul, line 429: If a `generator` is provided and the pipeline is executed end-to-end, I believe we could extract the seed with `seed()` on the generator. We can even extract the device as well to completely reproduce the generator state and use that to do the `torch.manual_seed()` being done later inside the method.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -268,73 +270,6 @@ def __init__(
             self.tokenizer.model_max_length if getattr(self, "tokenizer", None) is not None else 1024
         )
 
-    @staticmethod
```
sayakpaul, line 271: Where is this going?

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -939,6 +1026,16 @@ def __call__(
         device = self._execution_device
 
         # 3. Prepare text embeddings
+        if system_prompt is not None and prompt is not None:
```
sayakpaul, line 1050: If there's a default system prompt used by the LTX2.3 codebase, we could do something like this:

https://github.com/huggingface/diffusers/blob/e5aa719241f9b74d6700be3320a777799bfab70a/src/diffusers/pipelines/flux2/pipeline_flux2.py#L67

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -1069,11 +1168,6 @@ def __call__(
         self._num_timesteps = len(timesteps)
 
         # 6. Prepare micro-conditions
-        rope_interpolation_scale = (
-            self.vae_temporal_compression_ratio / frame_rate,
-            self.vae_spatial_compression_ratio,
-            self.vae_spatial_compression_ratio,
-        )
```
sayakpaul, line 1076: Has this gone inside the transformer?

Would we not have to compute it redundantly then?

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -1111,41 +1205,175 @@ def __call__(
...
-                        noise_pred_audio_text - noise_pred_audio_uncond
+                    noise_pred_audio_uncond_text, noise_pred_audio = noise_pred_audio.chunk(2)
+                    audio_cfg_delta = (self.audio_guidance_scale - 1) * (
+                        noise_pred_audio - noise_pred_audio_uncond_text
                     )
 
-                    if self.guidance_rescale > 0:
-                        # Based on 3.4. in https://huggingface.co/papers/2305.08891
-                        noise_pred_video = rescale_noise_cfg(
-                            noise_pred_video, noise_pred_video_text, guidance_rescale=self.guidance_rescale
```
sayakpaul, line 1144: How is this implemented then?

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -1111,41 +1205,175 @@ def __call__(
...
+                    guided_std = video_guided_x0.std(dim=list(range(1, video_guided_x0.ndim)), keepdim=True)
+                    rescale_factor = video_rescale * (cond_std / guided_std) + (1 - video_rescale)
+                    video_guided_x0 = video_guided_x0 * rescale_factor
+
+                    # Convert back to velocity space for scheduler
+                    noise_pred_video = (latents - video_guided_x0) / self.scheduler.sigmas[i]
+                else:
+                    noise_pred_video = noise_pred_video_g
+
+                if self.audio_guidance_rescale > 0:
```
sayakpaul, line 1390: Are we unable to use the rescaling utility?

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -1111,41 +1205,175 @@ def __call__(
                         encoder_hidden_states=connector_prompt_embeds,
                         audio_encoder_hidden_states=connector_audio_prompt_embeds,
                         timestep=timestep,
+                        sigma=timestep,  # Used by LTX-2.3
                         encoder_attention_mask=connector_attention_mask,
                         audio_encoder_attention_mask=connector_attention_mask,
+                        self_attention_mask=None,
+                        audio_self_attention_mask=None,
```
sayakpaul: Are these used by the other pipelines, such as I2V?

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -1111,41 +1205,175 @@ def __call__(
...
                 noise_pred_audio = noise_pred_audio.float()
 
                 if self.do_classifier_free_guidance:
-                    noise_pred_video_uncond, noise_pred_video_text = noise_pred_video.chunk(2)
-                    noise_pred_video = noise_pred_video_uncond + self.guidance_scale * (
-                        noise_pred_video_text - noise_pred_video_uncond
-                    )
+                    noise_pred_video_uncond_text, noise_pred_video = noise_pred_video.chunk(2)
+                    # Use delta formulation as it works more nicely with multiple guidance terms
+                    video_cfg_delta = (self.guidance_scale - 1) * (noise_pred_video - noise_pred_video_uncond_text)
```
sayakpaul, line 1257: (note to other reviewers): guidance is computed a bit latter to account for everything that comes before the computation.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -1111,41 +1205,175 @@ def __call__(
...
+                    noise_pred_audio_uncond_stg = noise_pred_audio_uncond_stg.float()
+
+                    video_stg_delta = self.stg_scale * (noise_pred_video - noise_pred_video_uncond_stg)
+                    audio_stg_delta = self.audio_stg_scale * (noise_pred_audio - noise_pred_audio_uncond_stg)
+                else:
+                    video_stg_delta = audio_stg_delta = 0
+
+                if self.do_modality_isolation_guidance:
+                    with self.transformer.cache_context("uncond_modality"):
+                        noise_pred_video_uncond_modality, noise_pred_audio_uncond_modality = self.transformer(
```
sayakpaul, line 1336: Do these calls vary from the previous ones in terms of the inputs? If so, it could be nice to add a small comment about it because the call arg list is pretty long.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -1111,41 +1205,175 @@ def __call__(
...
+                    )
+                else:
+                    video_modality_delta = audio_modality_delta = 0
+
+                # Now apply all guidance terms
+                noise_pred_video_g = noise_pred_video + video_cfg_delta + video_stg_delta + video_modality_delta
+                noise_pred_audio_g = noise_pred_audio + audio_cfg_delta + audio_stg_delta + audio_modality_delta
+
+                # Apply LTX-2.X guidance rescaling
+                if self.guidance_rescale > 0:
```
sayakpaul, line 1383: Are we unable to use the rescaling utility?

src/diffusers/pipelines/ltx2/vocoder.py
```diff
@@ -8,39 +8,269 @@
...
+        self.downsample = DownSample1d(ratio=ratio, kernel_size=kernel_size)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.upsample(x)
+        x = self.act(x)
+        x = self.downsample(x)
+        return x
+
+
+class SnakeBeta(nn.Module):
```
sayakpaul, line 172: TIL.

Should this go to `activations.py`? Okay if not.

sayakpaul (commented):
Thanks so much for the changes and keeping patience.

The amount of changes (also the ability to navigate them) is a bit overwhelming TBH.

I have left a few comments. Let me know if they make sense. We could consider adding a test-suite mirroring the existing LTX-2 pipeline tests but changing the components with changes specific to LTX-2.3?

scripts/convert_ltx2_to_diffusers.py
dg845: I think there is no issue in principle. But because `LTX2VocoderWithBWE` contains two `LTX2Vocoder`s as submodules it was more natural to me to wrap them in a new module (and it's also more parallel to the original code).

src/diffusers/loaders/lora_conversion_utils.py
dg845: The `prompt_adaln` and `audio_prompt_adaln` modules are used by both the full model and distilled model to calculate scale/shift modulation parameters for the text `encoder_hidden_states` for the video and audio modalities respectively. (I believe this is in place of the `caption_projection`s, which were removed in LTX-2.3.)

src/diffusers/models/autoencoders/autoencoder_kl_ltx2.py
dg845: I put `upsample_type` there because it is follows the argument ordering of `LTX2VideoDownBlock3D`, which already used an analogous `downsample_type` argument. I think the positional argument point is valid but IMO there is less risk of it breaking things as I think it's less likely that users are explicitly calling `LTX2VideoUpBlock3d` on its own.

src/diffusers/models/transformers/transformer_ltx2.py
dg845: It sounds like this change might have been added before the `2.3` release, and I believe the current original code allows LTX `2.0` checkpoints to use STG as well. So I think this is probably ok?

src/diffusers/models/transformers/transformer_ltx2.py
dg845: This follows the original code in that the per-transformer-block `prompt_scale_shift_table` and `audio_prompt_scale_shift_table` used to calculate modulation parameters for the text embeddings `K`, `V` (along with the global params discussed in https://github.com/huggingface/diffusers/pull/13217#discussion_r2943714530) are active if _either_ modality uses text cross attention modulation: https://github.com/Lightricks/LTX-2/blob/ae855f8538843825f9015a419cf4ba5edaf5eec2/packages/ltx-core/src/ltx_core/model/transformer/transformer.py#L115-L122

By contrast, the modulation parameters `shift_text_q`/`audio_shift_text_q`, etc. for the video and audio queries (`Q`) for text cross attention (`attn2` and `audio_attn2`) can be set independently of each other, with `video_cross_attn_adaln` and `audio_cross_attn_adaln` respectively (see e.g. https://github.com/Lightricks/LTX-2/blob/ae855f8538843825f9015a419cf4ba5edaf5eec2/packages/ltx-core/src/ltx_core/model/transformer/transformer.py#L59-L60).

(I think the original code is not consistent in this regard, in that the transformer blocks are only configured to have slots for the  `Q` modulation parameters in their scale/shift tables in `__init__` based on modality-independent configs `video.cross_attention_adaln` and `audio.cross_attention_adaln`, but in the inference logic the modulation parameter calculation logic is active if either modality configues it, similar to `prompt_scale_shift_table` above: https://github.com/Lightricks/LTX-2/blob/ae855f8538843825f9015a419cf4ba5edaf5eec2/packages/ltx-core/src/ltx_core/model/transformer/transformer.py#L245.)

src/diffusers/pipelines/ltx2/connectors.py
dg845: Yeah. In general, all default values here and in other modeling code (e.g. the transformer, VAE, vocoder) currently have their default arguments set to the LTX-`2.0` values.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
dg845: This was moved in the connector code in `connectors.py` as the `per_layer_masked_mean_norm` function. This is because LTX-`2.0` and `2.3` use different normalization logic in their text connectors, so I think it makes more sense for the connector code to own that logic rather than the pipeline code.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
dg845: The default system prompts are quite long (see e.g. the T2V system prompt here: https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-core/src/ltx_core/text_encoders/gemma/encoders/prompts/gemma_t2v_system_prompt.txt). At same time, that doesn't really prevent us from putting it in (for example) `utils.py`.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
dg845: The RoPE interpolation logic is not used in the LTX-`2.X` pipelines (and is a holdover from the LTX-`0.X` pipelines). I decided to remove it now as the pipelines are getting more complex so having it there was making it more confusing to follow the code.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
dg845: I think they are not used by any currently implemented pipeline. They might be used in pipelines that are in the LTX-2 code but not yet implemented in `diffusers`.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
dg845: I believe there is already an existing comment:

https://github.com/huggingface/diffusers/blob/6ee66c9d66e4ae57ede4773682f112f68167c7dd/src/diffusers/pipelines/ltx2/pipeline_ltx2.py#L1319-L1320

src/diffusers/pipelines/ltx2/vocoder.py
dg845: I think ideally it should, although I'm not familiar enough with `Snake`/`SnakeBeta` to say whether this is a stable, widely reusable implementation. My impression is that it's more or less standard though (this implementation follows the original LTX-2 code, which itself follows the BigVGAN-V2 implementation).

src/diffusers/models/transformers/transformer_ltx2.py
dg845: Since gated attention depends on additional parameters (a `to_gate_logits` linear projection layer), I think `video_gated_attn` is slightly better here as `apply_gated_attn` IMO suggests that gated attention only changes the forward logic without needing new parameters. But I'm open to another name here.

src/diffusers/models/transformers/transformer_ltx2.py
dg845: See https://github.com/huggingface/diffusers/pull/13217#discussion_r2943881685.

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -1175,6 +1434,32 @@ def forward(
             audio_encoder_attention_mask = (1 - audio_encoder_attention_mask.to(audio_hidden_states.dtype)) * -10000.0
             audio_encoder_attention_mask = audio_encoder_attention_mask.unsqueeze(1)
 
+        if self_attention_mask is not None and self_attention_mask.ndim == 3:
```
yiyixuxu: are we use self_attention_mask? just looked over the pipeline_ltx2 seems it always passed `self_attention_mask = None`

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -18,7 +18,7 @@
 
 import numpy as np
 import torch
-from transformers import Gemma3ForConditionalGeneration, GemmaTokenizer, GemmaTokenizerFast
```
yiyixuxu, line 21: I think we probably just create new pipelines for 2.3 - it's getting a bit overwhemling, no?

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
dg845: The inlined implementation (which follows the implementation in the original code) is equivalent to the implementation in `rescale_noise_cfg`. I have refactored the code to use `rescale_noise_cfg` instead.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
dg845: Do you think something like 

```python
def enhance_prompt(..., generator: torch.Generator | None = None, ...):
    ...
    if generator is not None:
        seed = generator.initial_seed()
    torch.manual_seed(seed)
    ...
```

would be good? The above is close to the original implementation. Or do you think the generator should control the randomness here more strictly (for example, by loading the `generator`'s current RNG state)?

src/diffusers/models/transformers/transformer_ltx2.py
dg845: We don't use `self_attention_mask` in any of the current `diffusers` LTX-2.X pipelines.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
dg845: Yeah, the pipelines are getting pretty complex. If I understand correctly, most of the code added to the pipelines in the current PR (such as the multimodal guidance strategy with STG) is also supported by the [original code](https://github.com/Lightricks/LTX-2) for `LTX-2.0` checkpoints as well. So the current official inference code for `LTX-2.0` and `LTX-2.3` checkpoints is mostly the same, and if we follow that the `2.0` and `2.3` pipelines would also mostly be the same.

src/diffusers/models/transformers/transformer_ltx2.py
yiyixuxu: let's remove these logic then

src/diffusers/pipelines/ltx2/latent_upsampler.py
```diff
@@ -195,7 +195,8 @@ def __init__(
         dims: int = 3,
         spatial_upsample: bool = True,
         temporal_upsample: bool = False,
-        rational_spatial_scale: float | None = 2.0,
+        rational_spatial_scale: float = 2.0,
+        use_rational_resampler: bool = True,
```
yiyixuxu, line 199: is this just a refactor?

yiyixuxu (approved):
thanks!

src/diffusers/pipelines/ltx2/latent_upsampler.py
dg845: It is mostly a refactor, the` LTX-2.0` checkpoint uses the rational resampler but the `LTX-2.3` checkpoint does not. So I think I wanted to decouple these two arguments to make it easier to support checkpoints which do not use the rational resampler.

tin2tin:
In the 2 step and distill examples, this line: `use_cross_timestep=True` gave me an error.

dg845:
Hi @tin2tin, can you provide an example script where you get the error? When I tested the examples with `use_cross_timestep=True` the generation finished without any errors.

tin2tin:
I'm sorry. I realize my version of your branch might have been a day old. If you get no error, it's all good. Thank you for persistently working on improving this patch. The Diffusers community is eagerly awaiting it.

rkazants:
Hi @sayakpaul, can you please tell if ltx2.3 will be included into 0.40.0 release? When do you plan to release it?

Best regards,
Roman

PR #13218, opened 2026-03-06
skip invalid test case for helios pipeline

kaixuanliu:
In helios pipeline, the latents are always in FP32 format: [L798](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/helios/pipeline_helios.py#L798), hence we should skip the test_save_load_float16 test case, rather than adjust the tolerance.

### DIFF
--- a/tests/pipelines/helios/test_helios.py
+++ b/tests/pipelines/helios/test_helios.py
@@ -139,7 +139,7 @@ class HeliosPipelineFastTests(PipelineTesterMixin, unittest.TestCase):
         generated_slice = torch.cat([generated_slice[:8], generated_slice[-8:]])
         self.assertTrue(torch.allclose(generated_slice, expected_slice, atol=1e-3))
 
-    @unittest.skip("Latents are always processed in FP32. Save/Load the entire pipeline in FP16 will result in errors")
+    @unittest.skip("Helios uses a lot of mixed precision internally, which is not suitable for this test case")
     def test_save_load_float16(self):
         pass
 


kaixuanliu:
@DN6 pls help review, thx!

kaixuanliu:
@yiyixuxu Hi, can you help review? Thx!

yiyixuxu:
hey @kaixuanliu 
latents are initially created in float32 and processed in the actual dtype of transformere though https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/helios/pipeline_helios.py#L817

did I miss anything?

kaixuanliu:
@yiyixuxu Thx for the comment! Yes, I am afraid you are right. But for this pipeline, it has a lot of modules using mixed precision, such as: [keep_in_fp32_module](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_helios.py#L548-L555), [FP32LayerNorm](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_helios.py#L393), [L435-L437](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_helios.py#L435-L437) and [HeliosOutputNorm](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_helios.py#L96), hence I suggest to skip this test case like in [WanVACE pipeline](https://github.com/huggingface/diffusers/blob/main/tests/pipelines/wan/test_wan_vace.py#L218-L222), and I have updated the skip reason. Do you think it is OK?

PR #13220, opened 2026-03-07
Refactor `AttentionProcessorSkipHook` to Support Custom STG Logic

dg845:
# What does this PR do?

This PR refactors `AttentionProcessorSkipHook` to support attention processors with a `_skip_attn_scores` flag, which is meant to allow attention processors to specify custom [spatio-temporal guidance (STG)](https://arxiv.org/abs/2411.18664)-like logic. The idea is that an `AttentionProcessor` could then be implemented as follows:

```python
class NewAttnProcessor:
    ...
    _skip_attn_scores = False
    _skip_attn_scores_fn = None

    def __call__(self, attn, hidden_states, ...):
        ...
        query = attn.to_q(hidden_states)
        key = attn.to_k(encoder_hidden_states)
        value = attn.to_v(encoder_hidden_states)
        ...
        if self._skip_attn_scores:
            # Defaults to returning `value` as in standard STG
            hidden_states = self._skip_attn_scores_fn(attn, query, key, value)
        else:
            ...
            hidden_states = dispatch_attention_fn(query, key, value, ...)
        ...
        hidden_states = attn.to_out[0](hidden_states)
        return hidden_states
```

This is motivated by supporting the following use cases:

1. **Supporting all attention backends**: the current implementation intercepts calls to `F.scaled_dot_product_attention`. But other attention backends like `flash-attn` won't call this function. The PR allows `SkipLayerGuidance` to work with any attention backend. (If the `AttnProcessor` doesn't have a `_skip_attn_scores` attribute, we will fall back to the current intercept approach.)
2. **Custom Skip Logic**: we can define custom logic in the `_skip_attn_scores_fn`. For example, LTX-2.3 applies learned per-head gates to the `value`s before passing them to the attention output projection `to_out`. This can be implemented `_skip_attn_scores_fn`.

Inspired by #13217, particularly https://github.com/huggingface/diffusers/pull/13217#discussion_r2898260578.

## Who can review?

@yiyixuxu
@sayakpaul

### DIFF
--- a/src/diffusers/hooks/layer_skip.py
+++ b/src/diffusers/hooks/layer_skip.py
@@ -111,9 +111,9 @@ class AttentionProcessorSkipHook(ModelHook):
     def __init__(
         self,
         skip_processor_output_fn: Callable,
-        skip_attn_scores_fn: Callable | None = None,
         skip_attention_scores: bool = False,
         dropout: float = 1.0,
+        skip_attn_scores_fn: Callable | None = None,
     ):
         super().__init__()
         self.skip_processor_output_fn = skip_processor_output_fn


src/diffusers/hooks/layer_skip.py
```diff
@@ -108,8 +108,17 @@ def __torch_function__(self, func, types, args=(), kwargs=None):
 
 
 class AttentionProcessorSkipHook(ModelHook):
-    def __init__(self, skip_processor_output_fn: Callable, skip_attention_scores: bool = False, dropout: float = 1.0):
+    def __init__(
+        self,
+        skip_processor_output_fn: Callable,
+        skip_attn_scores_fn: Callable | None = None,
```
sayakpaul: Should it go at the end to preserve BC compatibility (in case someone initialized this class with positional arguments only)?

No strong opinions here.

src/diffusers/hooks/layer_skip.py
```diff
@@ -108,8 +108,17 @@ def __torch_function__(self, func, types, args=(), kwargs=None):
...
+        skip_attention_scores: bool = False,
+        dropout: float = 1.0,
+    ):
+        super().__init__()
         self.skip_processor_output_fn = skip_processor_output_fn
+        # STG default: return the values as attention output
+        self.skip_attn_scores_fn = skip_attn_scores_fn or (lambda attn, q, k, v: v)
```
sayakpaul, line 121: Is it safe to assume that the `skip_attn_scores_fn` will only take those four (`attn, q, k, v`) as inputs and always return `v` as the output?

src/diffusers/hooks/layer_skip.py
```diff
@@ -119,8 +128,22 @@ def new_forward(self, module: torch.nn.Module, *args, **kwargs):
...
+                module.processor._skip_attn_scores = True
+                module.processor._skip_attn_scores_fn = self.skip_attn_scores_fn
+            # Use try block in case attn processor raises an exception
+            try:
+                if processor_supports_skip_fn:
+                    output = self.fn_ref.original_forward(*args, **kwargs)
+                else:
+                    # Fallback to torch native SDPA intercept approach
+                    with AttentionScoreSkipFunctionMode():
+                        output = self.fn_ref.original_forward(*args, **kwargs)
```
sayakpaul, line 142: Why do we have to condition like this?

src/diffusers/hooks/layer_skip.py
```diff
@@ -119,8 +128,22 @@ def new_forward(self, module: torch.nn.Module, *args, **kwargs):
...
+            finally:
```
sayakpaul, line 143: Should also not raise an exception for user clarity?

src/diffusers/hooks/layer_skip.py
dg845: `lambda attn, q, k, v: v` is intended to be the default `skip_attn_scores_fn` (unless I made a mistake), which performs standard STG by returning the values `v`. I don't think this constrains the signature or behavior of `skip_attn_scores_fn` if we pass it as an argument (and we can always manually implement the skip logic in the `if self._skip_attn_scores:` branch in the attention processor itself if necessary).

src/diffusers/hooks/layer_skip.py
dg845: We condition on `processor_supports_skip_fn` here in case the attention processor doesn't define a `_skip_attn_scores` attribute. If it doesn't we will fallback to the current behavior, which is to intercept a `torch.nn.functional.scaled_dot_product_attention` call and return the `value` from there. (The `AttentionScoreSkipFunctionMode` context manager performs the interception.)

src/diffusers/hooks/layer_skip.py
dg845: I'm confused by this question, is your suggestion to catch any exceptions in a `except` block? Maybe something like 

```python
            try:
                ...
            except Exception as e:
                logger.error(f"Tried to skip attn scores but got error {e}", exc_info=True)
                raise
            finally:
                # Clean up if necessary
                if processor_supports_skip_fn:
                    module.processor._skip_attn_scores = False
                    module.processor._skip_attn_scores_fn = None
```
?

PR #13226, opened 2026-03-08
[Discrete Diffusion] Add LLaDA2 pipeline

kashif:
Add support for LLaDA2/LLaDA2.1 discrete diffusion text generation:
- BlockRefinementPipeline: block-wise iterative refinement with confidence-based token commitment, supporting editing threshold for LLaDA2.1 models
- LLaDA2Pipeline: convenience wrapper with LLaDA2-specific defaults
- DiscreteDiffusionPipelineMixin: shared SAR sampling utilities (top-k, top-p, temperature) and prompt/prefix helpers
- compute_confidence_aware_loss: CAP-style training loss
- Examples: sampling scripts for LLaDA2 and block refinement, training scripts with Qwen causal LM
- Docs and tests included

# What does this PR do?

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/loaders/lora_conversion_utils.py
+++ b/src/diffusers/loaders/lora_conversion_utils.py
@@ -2443,6 +2443,191 @@ def _convert_non_diffusers_flux2_lora_to_diffusers(state_dict):
     return converted_state_dict
 
 
+def _convert_kohya_flux2_lora_to_diffusers(state_dict):
+    def _convert_to_ai_toolkit(sds_sd, ait_sd, sds_key, ait_key):
+        if sds_key + ".lora_down.weight" not in sds_sd:
+            return
+        down_weight = sds_sd.pop(sds_key + ".lora_down.weight")
+
+        # scale weight by alpha and dim
+        rank = down_weight.shape[0]
+        default_alpha = torch.tensor(rank, dtype=down_weight.dtype, device=down_weight.device, requires_grad=False)
+        alpha = sds_sd.pop(sds_key + ".alpha", default_alpha).item()
+        scale = alpha / rank
+
+        scale_down = scale
+        scale_up = 1.0
+        while scale_down * 2 < scale_up:
+            scale_down *= 2
+            scale_up /= 2
+
+        ait_sd[ait_key + ".lora_A.weight"] = down_weight * scale_down
+        ait_sd[ait_key + ".lora_B.weight"] = sds_sd.pop(sds_key + ".lora_up.weight") * scale_up
+
+    def _convert_to_ai_toolkit_cat(sds_sd, ait_sd, sds_key, ait_keys, dims=None):
+        if sds_key + ".lora_down.weight" not in sds_sd:
+            return
+        down_weight = sds_sd.pop(sds_key + ".lora_down.weight")
+        up_weight = sds_sd.pop(sds_key + ".lora_up.weight")
+        sd_lora_rank = down_weight.shape[0]
+
+        default_alpha = torch.tensor(
+            sd_lora_rank, dtype=down_weight.dtype, device=down_weight.device, requires_grad=False
+        )
+        alpha = sds_sd.pop(sds_key + ".alpha", default_alpha)
+        scale = alpha / sd_lora_rank
+
+        scale_down = scale
+        scale_up = 1.0
+        while scale_down * 2 < scale_up:
+            scale_down *= 2
+            scale_up /= 2
+
+        down_weight = down_weight * scale_down
+        up_weight = up_weight * scale_up
+
+        num_splits = len(ait_keys)
+        if dims is None:
+            dims = [up_weight.shape[0] // num_splits] * num_splits
+        else:
+            assert sum(dims) == up_weight.shape[0]
+
+        # check if upweight is sparse
+        is_sparse = False
+        if sd_lora_rank % num_splits == 0:
+            ait_rank = sd_lora_rank // num_splits
+            is_sparse = True
+            i = 0
+            for j in range(len(dims)):
+                for k in range(len(dims)):
+                    if j == k:
+                        continue
+                    is_sparse = is_sparse and torch.all(
+                        up_weight[i : i + dims[j], k * ait_rank : (k + 1) * ait_rank] == 0
+                    )
+                i += dims[j]
+            if is_sparse:
+                logger.info(f"weight is sparse: {sds_key}")
+
+        ait_down_keys = [k + ".lora_A.weight" for k in ait_keys]
+        ait_up_keys = [k + ".lora_B.weight" for k in ait_keys]
+        if not is_sparse:
+            ait_sd.update(dict.fromkeys(ait_down_keys, down_weight))
+            ait_sd.update({k: v for k, v in zip(ait_up_keys, torch.split(up_weight, dims, dim=0))})  # noqa: C416
+        else:
+            ait_sd.update({k: v for k, v in zip(ait_down_keys, torch.chunk(down_weight, num_splits, dim=0))})  # noqa: C416
+            i = 0
+            for j in range(len(dims)):
+                ait_sd[ait_up_keys[j]] = up_weight[i : i + dims[j], j * ait_rank : (j + 1) * ait_rank].contiguous()
+                i += dims[j]
+
+    # Detect number of blocks from keys
+    num_double_layers = 0
+    num_single_layers = 0
+    for key in state_dict.keys():
+        if key.startswith("lora_unet_double_blocks_"):
+            block_idx = int(key.split("_")[4])
+            num_double_layers = max(num_double_layers, block_idx + 1)
+        elif key.startswith("lora_unet_single_blocks_"):
+            block_idx = int(key.split("_")[4])
+            num_single_layers = max(num_single_layers, block_idx + 1)
+
+    ait_sd = {}
+
+    for i in range(num_double_layers):
+        # Attention projections
+        _convert_to_ai_toolkit(
+            state_dict,
+            ait_sd,
+            f"lora_unet_double_blocks_{i}_img_attn_proj",
+            f"transformer.transformer_blocks.{i}.attn.to_out.0",
+        )
+        _convert_to_ai_toolkit_cat(
+            state_dict,
+            ait_sd,
+            f"lora_unet_double_blocks_{i}_img_attn_qkv",
+            [
+                f"transformer.transformer_blocks.{i}.attn.to_q",
+                f"transformer.transformer_blocks.{i}.attn.to_k",
+                f"transformer.transformer_blocks.{i}.attn.to_v",
+            ],
+        )
+        _convert_to_ai_toolkit(
+            state_dict,
+            ait_sd,
+            f"lora_unet_double_blocks_{i}_txt_attn_proj",
+            f"transformer.transformer_blocks.{i}.attn.to_add_out",
+        )
+        _convert_to_ai_toolkit_cat(
+            state_dict,
+            ait_sd,
+            f"lora_unet_double_blocks_{i}_txt_attn_qkv",
+            [
+                f"transformer.transformer_blocks.{i}.attn.add_q_proj",
+                f"transformer.transformer_blocks.{i}.attn.add_k_proj",
+                f"transformer.transformer_blocks.{i}.attn.add_v_proj",
+            ],
+        )
+        # MLP layers (Flux2 uses ff.linear_in/linear_out)
+        _convert_to_ai_toolkit(
+            state_dict,
+            ait_sd,
+            f"lora_unet_double_blocks_{i}_img_mlp_0",
+            f"transformer.transformer_blocks.{i}.ff.linear_in",
+        )
+        _convert_to_ai_toolkit(
+            state_dict,
+            ait_sd,
+            f"lora_unet_double_blocks_{i}_img_mlp_2",
+            f"transformer.transformer_blocks.{i}.ff.linear_out",
+        )
+        _convert_to_ai_toolkit(
+            state_dict,
+            ait_sd,
+            f"lora_unet_double_blocks_{i}_txt_mlp_0",
+            f"transformer.transformer_blocks.{i}.ff_context.linear_in",
+        )
+        _convert_to_ai_toolkit(
+            state_dict,
+            ait_sd,
+            f"lora_unet_double_blocks_{i}_txt_mlp_2",
+            f"transformer.transformer_blocks.{i}.ff_context.linear_out",
+        )
+
+    for i in range(num_single_layers):
+        # Single blocks: linear1 -> attn.to_qkv_mlp_proj (fused, no split needed)
+        _convert_to_ai_toolkit(
+            state_dict,
+            ait_sd,
+            f"lora_unet_single_blocks_{i}_linear1",
+            f"transformer.single_transformer_blocks.{i}.attn.to_qkv_mlp_proj",
+        )
+        # Single blocks: linear2 -> attn.to_out
+        _convert_to_ai_toolkit(
+            state_dict,
+            ait_sd,
+            f"lora_unet_single_blocks_{i}_linear2",
+            f"transformer.single_transformer_blocks.{i}.attn.to_out",
+        )
+
+    # Handle optional extra keys
+    extra_mappings = {
+        "lora_unet_img_in": "transformer.x_embedder",
+        "lora_unet_txt_in": "transformer.context_embedder",
+        "lora_unet_time_in_in_layer": "transformer.time_guidance_embed.timestep_embedder.linear_1",
+        "lora_unet_time_in_out_layer": "transformer.time_guidance_embed.timestep_embedder.linear_2",
+        "lora_unet_final_layer_linear": "transformer.proj_out",
+    }
+    for sds_key, ait_key in extra_mappings.items():
+        _convert_to_ai_toolkit(state_dict, ait_sd, sds_key, ait_key)
+
+    remaining_keys = list(state_dict.keys())
+    if remaining_keys:
+        logger.warning(f"Unsupported keys for Kohya Flux2 LoRA conversion: {remaining_keys}")
+
+    return ait_sd
+
+
 def _convert_non_diffusers_z_image_lora_to_diffusers(state_dict):
     """
     Convert non-diffusers ZImage LoRA state dict to diffusers format.
--- a/src/diffusers/loaders/lora_pipeline.py
+++ b/src/diffusers/loaders/lora_pipeline.py
@@ -43,6 +43,7 @@ from .lora_conversion_utils import (
     _convert_bfl_flux_control_lora_to_diffusers,
     _convert_fal_kontext_lora_to_diffusers,
     _convert_hunyuan_video_lora_to_diffusers,
+    _convert_kohya_flux2_lora_to_diffusers,
     _convert_kohya_flux_lora_to_diffusers,
     _convert_musubi_wan_lora_to_diffusers,
     _convert_non_diffusers_flux2_lora_to_diffusers,
@@ -5672,6 +5673,13 @@ class Flux2LoraLoaderMixin(LoraBaseMixin):
             warn_msg = "It seems like you are using a DoRA checkpoint that is not compatible in Diffusers at the moment. So, we are going to filter out the keys associated to 'dora_scale` from the state dict. If you think this is a mistake please open an issue https://github.com/huggingface/diffusers/issues/new."
             logger.warning(warn_msg)
             state_dict = {k: v for k, v in state_dict.items() if "dora_scale" not in k}
+
+        is_kohya = any(".lora_down.weight" in k for k in state_dict)
+        if is_kohya:
+            state_dict = _convert_kohya_flux2_lora_to_diffusers(state_dict)
+            # Kohya already takes care of scaling the LoRA parameters with alpha.
+            out = (state_dict, metadata) if return_lora_metadata else state_dict
+            return out
 
         is_peft_format = any(k.startswith("base_model.model.") for k in state_dict)
         if is_peft_format:
--- a/tests/pipelines/test_pipelines_common.py
+++ b/tests/pipelines/test_pipelines_common.py
@@ -1534,14 +1534,18 @@ class PipelineTesterMixin:
         pipe.set_progress_bar_config(disable=None)
 
         pipe.to("cpu")
-        model_devices = [component.device.type for component in components.values() if hasattr(component, "device")]
+        model_devices = [
+            component.device.type for component in components.values() if getattr(component, "device", None)
+        ]
         self.assertTrue(all(device == "cpu" for device in model_devices))
 
         output_cpu = pipe(**self.get_dummy_inputs("cpu"))[0]
         self.assertTrue(np.isnan(output_cpu).sum() == 0)
 
         pipe.to(torch_device)
-        model_devices = [component.device.type for component in components.values() if hasattr(component, "device")]
+        model_devices = [
+            component.device.type for component in components.values() if getattr(component, "device", None)
+        ]
         self.assertTrue(all(device == torch_device for device in model_devices))
 
         output_device = pipe(**self.get_dummy_inputs(torch_device))[0]
@@ -1552,11 +1556,11 @@ class PipelineTesterMixin:
         pipe = self.pipeline_class(**components)
         pipe.set_progress_bar_config(disable=None)
 
-        model_dtypes = [component.dtype for component in components.values() if hasattr(component, "dtype")]
+        model_dtypes = [component.dtype for component in components.values() if getattr(component, "dtype", None)]
         self.assertTrue(all(dtype == torch.float32 for dtype in model_dtypes))
 
         pipe.to(dtype=torch.float16)
-        model_dtypes = [component.dtype for component in components.values() if hasattr(component, "dtype")]
+        model_dtypes = [component.dtype for component in components.values() if getattr(component, "dtype", None)]
         self.assertTrue(all(dtype == torch.float16 for dtype in model_dtypes))
 
     def test_attention_slicing_forward_pass(self, expected_max_diff=1e-3):


yiyixuxu (commented):
Thanks for working on this!

My main concern is we might be over-generalizing it a bit: currently there is a 3-layer hierarchy (`DiscreteDiffusionPipelineMixin` -> `BlockRefinementPipeline` -`L LLaDA2Pipeline`). I think it is at least one layer too much for diffusion style.  i.e., `BlockRefinementPipeline` should be combined with `LLaDA2Pipeline`. we do use a Mixin class to abstract away common pipeline methods, but since this is the first model we support, I also would prefer to wait on that 

Does it maks sense to flatten this into a single `LLaDA2Pipeline(DiffusionPipeline)` for now?  we can look into refactoring later we more discrete diffusion model arrives 

We can also support it in modular pipelines instead, this way you can avoid duplicated code and it works better with remote code. We have a skill coming soon can help with that. But for this PR, I think we can just merge one first?

kashif:
Thanks @yiyixuxu for the suggestion! Fixed.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -2383,3 +2383,180 @@ def unfuse_qkv_projections(self, unet: bool = True, vae: bool = True):
             else:
                 self.vae.unfuse_qkv_projections()
                 self.fusing_vae = False
+
+
+class DiscreteDiffusionPipelineMixin:
```
yiyixuxu: Can we remove this for now and move the methods to `LLaDA2Pipeline`? 
Since right now there is only one ppeline so don't really need a Mixin. we can extracting a mixin later when we have more pipelines and see a common pattern - the refactor is easy and won't break anything.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+        tokenizer: Optional[Any] = None,
+    ):
+        super().__init__()
+        self.register_modules(model=model, scheduler=scheduler, tokenizer=tokenizer)
+
+    @property
+    def num_timesteps(self):
+        return self._num_timesteps
+
+    def _model_forward_logits(
```
yiyixuxu: this is only used once - can you remove the method and put the cod directly inside __call__? 

I think we should remove `try/except` and maybe the `attention_mask_mode` argument altogether - the type of attention mask is determined by the model, no?

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+            )
+
+    @torch.no_grad()
+    @replace_example_docstring(EXAMPLE_DOC_STRING)
+    def __call__(
+        self,
+        prompt: Optional[Union[str, List[str]]] = None,
+        messages: Optional[List[Dict[str, str]]] = None,
+        input_ids: Optional[torch.LongTensor] = None,
+        prompt_ids: Optional[torch.LongTensor] = None,
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+            prompt (`str` or `List[str]`, *optional*):
+                Prompt text. When `use_chat_template` is `True` (default) and a tokenizer with a chat template is
+                available, the prompt is wrapped in a chat message before tokenization.
+            messages (`List[Dict[str, str]]`, *optional*):
+                Chat messages to encode (e.g. `[{"role": "user", "content": "Hello"}]`). Takes precedence over `prompt`
+                when provided. Requires a tokenizer with `apply_chat_template`.
+            input_ids (`torch.LongTensor`, *optional*):
+                Pre-tokenized input IDs. Takes precedence over `prompt` and `messages`.
+            prompt_ids (`torch.LongTensor`, *optional*):
+                Alias for `input_ids` for backward compatibility.
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+                timestep: int, callback_kwargs: Dict)`.
+            callback_on_step_end_tensor_inputs (`List[str]`, *optional*):
+                Tensor keys to pass to the callback. Allowed keys: `cur_x`, `x0`, `x0_p`, `transfer_index`,
+                `confidence`, `active_block`.
+
+        Examples:
+        """
+        # Handle prompt_ids as alias for input_ids (backward compatibility)
+        if prompt_ids is not None and input_ids is None:
+            input_ids = prompt_ids
```
yiyixuxu:
```suggestion
```
it is a new pipeline for us, so we don't need this I think!

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+            threshold=threshold,
+            sampling_method=sampling_method,
+            callback_on_step_end=callback_on_step_end,
+            callback_on_step_end_tensor_inputs=callback_on_step_end_tensor_inputs,
+        )
+
+        model_params = list(self.model.parameters()) if hasattr(self.model, "parameters") else []
+        model_device = model_params[0].device if len(model_params) > 0 else torch.device("cpu")
+
+        prompt_ids_encoded = self._encode_prompt(None, encoded_ids, device=model_device)
```
yiyixuxu: so the _prepare_input_ids return the token ids and now _encode_prompt seems like doing the same?

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+        batch_size, prompt_length = prompt_ids_encoded.shape
+
+        if eos_token_id is None:
+            eos_token_id = getattr(getattr(self, "tokenizer", None), "eos_token_id", None)
```
yiyixuxu:
```suggestion
            eos_token_id = self.eos_token_ids
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+        if mask_token_id is None:
+            mask_token_id = getattr(getattr(self, "tokenizer", None), "mask_token_id", None)
```
yiyixuxu:
```suggestion
            mask_token_id = self.mask_token_id
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+    _callback_tensor_inputs = ["cur_x", "x0", "x0_p", "transfer_index", "confidence", "active_block"]
+
+    def __init__(
+        self,
+        model: Any,
+        scheduler: BlockRefinementScheduler,
+        tokenizer: Optional[Any] = None,
+    ):
+        super().__init__()
+        self.register_modules(model=model, scheduler=scheduler, tokenizer=tokenizer)
```
yiyixuxu, line 83:
```suggestion
        self.register_modules(model=model, scheduler=scheduler, tokenizer=tokenizer)
        self.eos_token_ids = getattr(self.tokenizer, "eos_token_ids", None) if hasattr(self, "tokenizer") else None
        self.mask_token_ids = getattr(self.tokenizer, "mask_token_ids", None) if hasattr(self, "tokenizer") else None
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+            steps=steps,
+            minimal_topk=minimal_topk,
+            threshold=threshold,
+            sampling_method=sampling_method,
+            callback_on_step_end=callback_on_step_end,
+            callback_on_step_end_tensor_inputs=callback_on_step_end_tensor_inputs,
+        )
+
+        model_params = list(self.model.parameters()) if hasattr(self.model, "parameters") else []
+        model_device = model_params[0].device if len(model_params) > 0 else torch.device("cpu")
```
yiyixuxu:
```suggestion
        device = self._execution_device
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
yiyixuxu: we can create default during `__init__`

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+            raise ValueError("`mask_token_id` must be provided (or available on the tokenizer).")
+
+        steps = min(int(steps), int(gen_length) // int(minimal_topk))
+
+        self.scheduler.set_timesteps(steps, device=model_device)
+
+        num_blocks = (prompt_length + int(gen_length) + int(block_length) - 1) // int(block_length)
+        total_length = int(num_blocks) * int(block_length)
+
+        dtype = getattr(self.model, "dtype", torch.float32)
```
yiyixuxu: can we just use float32? it's just 1s and 0s no?

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+        attn_dtype = torch.bfloat16 if dtype in (torch.bfloat16, torch.float16) else torch.float32
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+        attn_mask_4d = self._build_block_attention_mask(
+            num_blocks=num_blocks,
+            block_length=block_length,
+            total_length=total_length,
+            device=model_device,
+            dtype=attn_dtype,
```
yiyixuxu:
```suggestion
            dtype=torch.float32,
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+        try:
+            return _call(attention_mask_4d), "4d"
+        except (TypeError, ValueError, RuntimeError):
+            pass
+        try:
+            return _call(attention_mask_2d), "2d"
+        except (TypeError, ValueError, RuntimeError):
+            return _call(None), "none"
+
+    def _build_block_attention_mask(
```
yiyixuxu: can we inline this method? it's really small

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+
+        if prompt is None:
+            return torch.zeros((1, 0), device=device, dtype=torch.long)
+        if getattr(self, "tokenizer", None) is None:
+            raise ValueError("`prompt` requires a tokenizer, but no tokenizer was provided to the pipeline.")
+
+        encoded = self.tokenizer(prompt, return_tensors="pt", padding=True)
+        return encoded["input_ids"].to(device=device)
+
+    def prepare_latents(
```
yiyixuxu: can we inline this method? it is one line

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+                        if not eos_in_commits:
+                            continue
+                        eos_pos = (cur_x[b] == int(eos_token_id)).nonzero(as_tuple=True)
+                        if len(eos_pos[0]) == 0:
+                            continue
+                        eos_pos = int(eos_pos[0][0].item())
+                        if prompt_length >= eos_pos:
+                            continue
+                        if (cur_x[b, prompt_length:eos_pos] != int(mask_token_id)).all().item():
+                            finished[b] = True
```
yiyixuxu:
```suggestion
                    finished = self._check_eos_finished(...)
```
can we put this into a helper function? not essential to pipeline and bit hard to follow

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+            # Identify which positions in the block are prompt (non-editable).
+            block_start_pos = num_block * int(block_length)
+            prompt_mask_in_block = torch.zeros(int(block_length), device=model_device, dtype=torch.bool)
+            if block_start_pos < prompt_length:
+                prompt_end_in_block = min(prompt_length - block_start_pos, int(block_length))
+                prompt_mask_in_block[:prompt_end_in_block] = True
+
+            post_steps = 0
+            step_idx = 0
+            while step_idx < int(steps) or (editing_enabled and post_steps <= int(max_post_steps)):
```
yiyixuxu:
```suggestion
           should_continue = not finished.all() and (step_idx < int(steps) or (editing_enabled and post_steps <= int(max_post_steps))
            while should_continue:
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+                if finished.all():
+                    break
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+                transfer_index = scheduler_output.transfer_index
+                editing_transfer_index = scheduler_output.editing_transfer_index
+                final_transfer = transfer_index | editing_transfer_index
+
+                if final_transfer.any():
+                    cur_x[:, -int(block_length) :] = scheduler_output.prev_sample
+
+                # Break if no masks remain and no edits were made.
+                if not masks_remaining and not editing_transfer_index.any():
+                    break
```
yiyixuxu:
```suggestion
```
can we update should_continue at same place with all the conditions?

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+                if callback_on_step_end is not None:
+                    callback_kwargs = {}
+                    for k in callback_on_step_end_tensor_inputs:
+                        callback_kwargs[k] = locals()[k]
+                    callback_outputs = callback_on_step_end(self, global_step, step_idx, callback_kwargs)
+                    cur_x = callback_outputs.pop("cur_x", cur_x)
+
+                global_step += 1
+                if masks_remaining:
+                    step_idx += 1
```
yiyixuxu, line 456:
```suggestion
                if masks_remaining:
                    step_idx += 1
                
                ## update should_continue flag at same place if possible
                if finished.all():
                    should_continue = False
                if not mask_remaining and not final_transformer.amy():
                    should_continue = False
                if step_idx < steps or (editing_enabled and post_steps <= max_post_steps):
                    should_continue = False
                 
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,514 @@
...
+
+            x[:, :current_window_end] = cur_x
+            if eos_token_id is not None and (x[:, prompt_length:current_window_end] == int(eos_token_id)).any().item():
+                if eos_early_stop:
```
yiyixuxu:
```suggestion
            if eos_early_stop and finished.all(()
```
would this work? I think we already checked the eos tokens within inner loop and updated `finished` accordingly -

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,563 @@
...
+    Examples:
+        ```python
+        >>> import torch
+        >>> from transformers import AutoModelForCausalLM, AutoTokenizer
+        >>> from diffusers import BlockRefinementScheduler, LLaDA2Pipeline
+
+        >>> model_id = "inclusionAI/LLaDA2.1-mini"
+        >>> model = AutoModelForCausalLM.from_pretrained(
+        ...     model_id, trust_remote_code=True, dtype=torch.bfloat16, device_map="auto"
+        ... )
```
dg845, line 42: I think having a `diffusers`-native implementation of the transformer would be better. CC @yiyixuxu

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,563 @@
...
+        chat_template_kwargs: Optional[Dict[str, object]],
+    ) -> torch.LongTensor:
+        """Convert prompt/messages/input_ids to a [batch, seq] LongTensor."""
+        if input_ids is not None:
+            if input_ids.ndim == 1:
+                input_ids = input_ids.unsqueeze(0)
+            if input_ids.ndim != 2:
+                raise ValueError(f"`input_ids` must be 2D, got shape {tuple(input_ids.shape)}.")
+            if input_ids.dtype != torch.long:
+                raise ValueError(f"`input_ids` must be int64 token IDs, got dtype={input_ids.dtype}.")
```
dg845, line 110: Could we move the `input_ids` checks and the other checks below to `check_inputs`? This would allow us to fail earlier on unexpected inputs. (This assumes that `check_inputs` is called first, but it is currently called after this function. So my full suggestion would be to call `check_inputs` first in `__call__`, and then move all input validation to `check_inputs`.)

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,563 @@
...
+
+        if callback_on_step_end is not None and isinstance(
+            callback_on_step_end, (PipelineCallback, MultiPipelineCallbacks)
+        ):
+            callback_on_step_end_tensor_inputs = callback_on_step_end.tensor_inputs
+        if callback_on_step_end_tensor_inputs is None:
+            callback_on_step_end_tensor_inputs = ["cur_x"]
+
+        # 2. Check inputs
+        self.check_inputs(
```
dg845: Suggestion: call `check_inputs` first in `__call__` so that we can fail as early as possible. See https://github.com/huggingface/diffusers/pull/13226#discussion_r2957938425 for more details.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,563 @@
...
+        if mask_token_id is None:
+            raise ValueError("`mask_token_id` must be provided (or available on the tokenizer).")
+
+        steps = min(int(steps), int(gen_length) // int(minimal_topk))
+
+        self.scheduler.set_timesteps(steps, device=device)
+
+        # 3. Build attention mask and position IDs
+        num_blocks = (prompt_length + int(gen_length) + int(block_length) - 1) // int(block_length)
+        total_length = int(num_blocks) * int(block_length)
```
dg845: nit: as we are already annotating `gen_length`, `block_length`, etc. as `int`s in `__call__`, maybe we can remove the explicit `int(...)` casts here and elsewhere?

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,563 @@
...
+        generated = x[:, : prompt_length + int(gen_length)]
+        sequences = generated[:, prompt_length:]
+        if eos_token_id is not None and batch_size == 1:
+            eos_positions = (sequences[0] == int(eos_token_id)).nonzero(as_tuple=True)[0]
+            if len(eos_positions) > 0:
+                sequences = sequences[:, : int(eos_positions[0].item()) + 1]
+
+        texts = None
+        if return_text and self.tokenizer is not None:
+            texts = self.tokenizer.batch_decode(sequences, skip_special_tokens=True)
```
dg845, line 484: Suggestion: maybe we could define an `output_type` argument analogous to that used by image generation pipelines?

https://github.com/huggingface/diffusers/blob/67613369bb92f120eff221fc571dc2901fe994f6/src/diffusers/pipelines/flux2/pipeline_flux2.py#L758

Here if `output_type=="seq"` we could return the `sequences` and if `output_type=="text"` we could return the decoded `texts`.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,563 @@
...
+
+                # Update should_continue
+                if finished.all():
+                    should_continue = False
+                elif not masks_remaining and not editing_transfer_index.any():
+                    should_continue = False
+                elif masks_remaining and step_idx >= int(steps):
+                    should_continue = False
+                elif not masks_remaining and post_steps > int(max_post_steps):
+                    should_continue = False
```
dg845: I think putting the stopping logic in the scheduler might make more sense as schedulers are responsible for implementing the sampling algorithm. If I understand correctly, deciding when to stop is purely a property of the sampling algorithm in this case.

src/diffusers/schedulers/scheduling_block_refinement.py
```diff
@@ -0,0 +1,214 @@
...
+        out = torch.full((int(num_inference_steps),), base, dtype=torch.long)
+        out[:remainder] += 1
+        return out
+
+    def step(
+        self,
+        sampled_tokens: torch.LongTensor,
+        sampled_probs: torch.Tensor,
+        timestep: Union[int, torch.Tensor],
+        sample: torch.LongTensor,
```
dg845, line 170: Schedulers usually accept the raw model outputs via a `model_output` argument, which for LlaDA 2 would be the `logits` output by the transformer denoising model. I think it would be nice if we could use the usual scheduler step interface:

https://github.com/huggingface/diffusers/blob/67613369bb92f120eff221fc571dc2901fe994f6/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L425-L429

Do you think this would make sense here? I think changing `step` to use the raw model outputs would also imply that the scheduler (rather than the pipeline) would own the logit sampling logic such as `_sample_with_temperature_topk_topp`.

tests/pipelines/test_pipeline_llada2.py
```diff
@@ -0,0 +1,252 @@
+import unittest
```
dg845, line 1: I think the tests should be moved to their own folder in the `pipelines` subdirectory, e.g. `/tests/pipelines/llada2/test_llada2.py`. See for instance the [Flux 2 tests](https://github.com/huggingface/diffusers/tree/main/tests/pipelines/flux2) as an example.

src/diffusers/training_utils.py
```diff
@@ -109,6 +110,92 @@ def compute_snr(noise_scheduler, timesteps):
...
 
+def compute_confidence_aware_loss(
+    logits: torch.Tensor,
+    labels: torch.Tensor,
+    *,
+    lambda_conf: float = 0.0,
+    temperature: float = 1.0,
+    per_token_weights: Optional[torch.Tensor] = None,
+    ignore_index: int = -100,
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```
dg845:
```suggestion
    per_token_weights: torch.Tensor | None = None,
    ignore_index: int = -100,
) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```
nit: use Python 3.9+ style implicit type annotations following #12524.

src/diffusers/training_utils.py
```diff
@@ -7,10 +7,11 @@
 import warnings
 from contextlib import contextmanager
 from functools import partial
-from typing import Any, Iterable
+from typing import Any, Iterable, Optional, Tuple
```
dg845:
```suggestion
from typing import Any, Iterable
```
Follow up change to https://github.com/huggingface/diffusers/pull/13226#discussion_r2958000603.

examples/discrete_diffusion/train_llada2.py
```diff
@@ -0,0 +1,357 @@
...
+        return self.num_samples
+
+    def __getitem__(self, idx):
+        del idx
+        input_ids = torch.randint(0, self.vocab_size, (self.seq_len,), dtype=torch.long)
+        attention_mask = torch.ones_like(input_ids)
+        return {"input_ids": input_ids, "attention_mask": attention_mask}
+
+
+def forward_process_semi_ar(
```
dg845: Suggestion: maybe we could move this to be the `add_noise` method of `BlockRefinementScheduler`? The `add_noise` method usually defines the forward process associated with the scheduler; see `DDIMScheduler` for an example:

https://github.com/huggingface/diffusers/blob/67613369bb92f120eff221fc571dc2901fe994f6/src/diffusers/schedulers/scheduling_ddim.py#L517

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,484 @@
...
+    def __call__(
+        self,
+        prompt: Optional[Union[str, List[str]]] = None,
+        messages: Optional[List[Dict[str, str]]] = None,
+        input_ids: Optional[torch.LongTensor] = None,
+        use_chat_template: bool = True,
+        add_generation_prompt: bool = True,
+        gen_length: int = 2048,
+        block_length: int = 32,
+        steps: int = 32,
```
dg845:
```suggestion
        num_inference_steps: int = 32,
```
nit: rename `steps` to `num_inference_steps`, which follows the `diffusers` convention for number of denoising steps.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,484 @@
...
+        finished = torch.zeros((batch_size,), device=device, dtype=torch.bool)
+        editing_enabled = editing_threshold is not None and editing_threshold >= 0.0
+        global_step = 0
+
+        # 5. Block-wise refinement loop
+        for num_block in range(prefill_blocks, num_blocks):
+            current_window_end = (num_block + 1) * block_length
+            cur_x = x[:, :current_window_end]
+            cur_attn_mask = attn_mask[:, :current_window_end]
+            cur_position_ids = position_ids[:, :current_window_end]
```
dg845:
```suggestion
            block_x = x[:, :current_window_end]
            block_attn_mask = attn_mask[:, :current_window_end]
            block_position_ids = position_ids[:, :current_window_end]
```
nit: I think prefixing all block-level variable names with `block` (e.g. `cur_x` --> `block_x`, `cur_attn_mask` --> `block_attn_mask`, etc.) would be more intuitive and consistent with other names such as `block_logits` below.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,484 @@
...
+
+        prefill_blocks = prompt_length // block_length
+        self._num_timesteps = steps * max(num_blocks - prefill_blocks, 0)
+
+        finished = torch.zeros((batch_size,), device=device, dtype=torch.bool)
+        editing_enabled = editing_threshold is not None and editing_threshold >= 0.0
+        global_step = 0
+
+        # 5. Block-wise refinement loop
+        for num_block in range(prefill_blocks, num_blocks):
```
dg845:
```suggestion
        block_progress_bar_config = getattr(self, "_progress_bar_config", {}).copy()
        block_progress_bar_config["leave"] = False  # Prevent stale bars from stacking up
        for num_block in tqdm(range(prefill_blocks, num_blocks), **block_progress_bar_config):
```
I think having a block level progress bar here would be useful to track inference progress.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,484 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+
+import torch
```
dg845, line 20:
```suggestion
import torch
from tqdm.auto import tqdm
```
Follow up change for block progress bars https://github.com/huggingface/diffusers/pull/13226#discussion_r2968670591.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,484 @@
...
+            prompt_mask_in_block = torch.zeros(block_length, device=device, dtype=torch.bool)
+            if block_start_pos < prompt_length:
+                prompt_end_in_block = min(prompt_length - block_start_pos, block_length)
+                prompt_mask_in_block[:prompt_end_in_block] = True
+
+            post_steps = 0
+            step_idx = 0
+            should_continue = True
+
+            while should_continue:
```
dg845, line 402: Similarly, as with block progress bars (https://github.com/huggingface/diffusers/pull/13226#discussion_r2968670591), I think have a per-block progress bar would also be useful. Something like

```python
progress_bar = self.progress_bar(total=steps)
while should_continue:
    ...
    if masks_remaining:
        step_idx += 1
        progress_bar.update(1)
    ...
```

should work here.

src/diffusers/schedulers/scheduling_block_refinement.py
```diff
@@ -0,0 +1,458 @@
...
+            if len(eos_pos[0]) == 0:
+                continue
+            eos_pos = int(eos_pos[0][0].item())
+            if prompt_length >= eos_pos:
+                continue
+            if (cur_x[b, prompt_length:eos_pos] != mask_token_id).all().item():
+                finished[b] = True
+        return finished
+
+    def check_should_continue(
```
dg845:
```suggestion
    def check_block_should_continue(
```
nit: I think renaming this to indicate that it checks whether denoising should continue within a block would make the code more clear.

src/diffusers/schedulers/scheduling_block_refinement.py
```diff
@@ -0,0 +1,458 @@
...
+        self,
+        step_idx: int,
+        masks_remaining: bool,
+        editing_enabled: bool,
```
dg845, line 355: `editing_enabled` is currently not used in `check_should_continue`, is this intended?

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,484 @@
...
+
+    def _prepare_input_ids(
+        self,
+        *,
+        prompt: Optional[Union[str, List[str]]],
+        messages: Optional[List[Dict[str, str]]],
+        input_ids: Optional[torch.LongTensor],
+        use_chat_template: bool,
+        add_generation_prompt: bool,
+        chat_template_kwargs: Optional[Dict[str, object]],
```
dg845:
```suggestion
        prompt: str | list[str] | None,
        messages: list[dict[str, str]] | None,
        input_ids: torch.LongTensor | None,
        use_chat_template: bool,
        add_generation_prompt: bool,
        chat_template_kwargs: dict[str, Any] | None,
```
nit: follow Python 3.9+ style implicit type annotations. See https://github.com/huggingface/diffusers/pull/13226#discussion_r2958000603.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,484 @@
...
+            return encoded["input_ids"]
+
+        encoded = self.tokenizer(prompt, return_tensors="pt", padding=isinstance(prompt, list))
+        return encoded["input_ids"]
+
+    def check_inputs(
+        self,
+        prompt: Optional[Union[str, List[str]]],
+        messages: Optional[List[Dict[str, str]]],
+        input_ids: Optional[torch.LongTensor],
```
dg845:
```suggestion
        prompt: str | list[str] | None,
        messages: list[dict[str, str]] | None,
        input_ids: torch.LongTensor | None,
```

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,484 @@
...
+        gen_length: int,
+        block_length: int,
+        steps: int,
+        minimal_topk: int,
+        threshold: float,
+        sampling_method: str,
+        output_type: str,
+        callback_on_step_end: Optional[Union[Callable, PipelineCallback, MultiPipelineCallbacks]],
+        callback_on_step_end_tensor_inputs: Optional[List[str]],
```
dg845:
```suggestion
        callback_on_step_end: Callable | PipelineCallback | MultiPipelineCallbacks | None,
        callback_on_step_end_tensor_inputs: list[str] | None,
```

src/diffusers/schedulers/scheduling_block_refinement.py
```diff
@@ -0,0 +1,458 @@
...
+        num_inference_steps: int = 32,
+        threshold: float = 0.95,
+        editing_threshold: Optional[float] = None,
+        minimal_topk: int = 1,
+    ):
+        self.num_inference_steps = num_inference_steps
+        self.timesteps = torch.arange(self.num_inference_steps - 1, -1, -1, dtype=torch.long)
+        self._transfer_schedule: Optional[torch.LongTensor] = None
+
+    def set_timesteps(self, num_inference_steps: int, device: Optional[Union[str, torch.device]] = None) -> None:
```
dg845:
```suggestion
    def set_timesteps(self, num_inference_steps: int, device: str | torch.device | None = None) -> None:
```
See https://github.com/huggingface/diffusers/pull/13226#discussion_r2968681270.

src/diffusers/schedulers/scheduling_block_refinement.py
kashif: yes i was checking it outside, fixed now thanks

docs/source/en/api/pipelines/llada2.md
```diff
@@ -0,0 +1,83 @@
...
+model = AutoModelForCausalLM.from_pretrained(model_id, trust_remote_code=True, dtype=torch.bfloat16, device_map="auto")
+tokenizer = AutoTokenizer.from_pretrained(model_id, trust_remote_code=True)
+scheduler = BlockRefinementScheduler()
+
+pipe = LLaDA2Pipeline(model=model, scheduler=scheduler, tokenizer=tokenizer)
+output = pipe(
+    prompt="Write a short poem about the ocean.",
+    gen_length=256,
+    block_length=32,
+    steps=32,
```
dg845:
```suggestion
    num_inference_steps=32,
```
nit: so that the example works after renaming `steps` --> `num_inference_steps`.

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,489 @@
...
+        num_inference_steps = min(num_inference_steps, gen_length // minimal_topk)
+
+        self.scheduler.set_timesteps(num_inference_steps, device=device)
+
+        # 3. Build attention mask and position IDs
+        num_blocks = (prompt_length + gen_length + block_length - 1) // block_length
+        total_length = num_blocks * block_length
+
+        # 2D attention mask (no padding) — the model handles backend-specific conversion internally.
+        attn_mask = torch.ones((batch_size, total_length), device=device, dtype=torch.long)
```
dg845, line 363: If I try to run the example script from the docs, I get the following error from the [LLaDA-2.1 MoE modeling code](https://huggingface.co/inclusionAI/LLaDA2.1-mini/blob/main/modeling_llada2_moe.py):

```
...
  File "~/.cache/modules/transformers_modules/inclusionAI/LLaDA2_dot_1_hyphen_mini/f21be037104f6e044e1a86b6d8864a6b85cc868e/modeling_llada2_moe.py", line 874, in forward
    raise ValueError(
ValueError: LLaDA2.0 only support block attention mask with shape: (1, 1, 32, 32), the input attention with shape attention_mask.size()=torch.Size([1, 32])!
```

From the error it seems like the original modeling code will only accept 4D (rather than 2D) attention masks?

kashif:
@dg845 converting the models to diffusers will be quite a lot more work and all the d-llms currently use transformers modeling files for their backbone, so yeah i wanted to first judge how useful this will be in diffusers and then see moving forward?

Regarding the attention shape error, their modeling files are still stuck with some transformers 4 api and I have fixed that for them, so in the meantime, test with `inclusionAI/LLaDA2.1-mini` using revision `refs/pr/4` see https://huggingface.co/inclusionAI/LLaDA2.1-mini/discussions/4 for example.

src/diffusers/pipelines/llada2/pipeline_llada2.py
kashif: moved to schedular

yiyixuxu:
>so yeah i wanted to first judge how useful this will be in diffusers and then see moving forward?

sounds good @kashif @dg845

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,489 @@
...
+        prefill_blocks = prompt_length // block_length
+        self._num_timesteps = num_inference_steps * max(num_blocks - prefill_blocks, 0)
+
+        finished = torch.zeros((batch_size,), device=device, dtype=torch.bool)
+        editing_enabled = editing_threshold is not None and editing_threshold >= 0.0
+        global_step = 0
+
+        # 5. Block-wise refinement loop
+        block_progress_bar_config = getattr(self, "_progress_bar_config", {}).copy()
+        block_progress_bar_config["leave"] = False
```
dg845:
```suggestion
        block_progress_bar_config["position"] = 0
        block_progress_bar_config["desc"] = "Blocks"
```
Suggestion: make the nested progress bars nicer

src/diffusers/pipelines/llada2/pipeline_llada2.py
```diff
@@ -0,0 +1,489 @@
...
+            block_start_pos = num_block * block_length
+            prompt_mask_in_block = torch.zeros(block_length, device=device, dtype=torch.bool)
+            if block_start_pos < prompt_length:
+                prompt_end_in_block = min(prompt_length - block_start_pos, block_length)
+                prompt_mask_in_block[:prompt_end_in_block] = True
+
+            post_steps = 0
+            step_idx = 0
+            should_continue = True
+            progress_bar = self.progress_bar(total=num_inference_steps)
```
dg845, line 400:
```suggestion
            self.set_progress_bar_config(position=1, leave=False, desc=f"Block {num_block} Inference Steps")
            progress_bar = self.progress_bar(total=num_inference_steps)
```
Suggestion: make the inner progress bars nicer (and prevent them from stacking up)

Issue #13225, opened 2026-03-08
RAE+DiT support for published checkpoints


plugyawn:
**Is your feature request related to a problem? Please describe.**
`diffusers` has `AutoencoderRAE` support via #13046. There is still no native latent generator path for the published RAE checkpoints. I've been following the research and was excited to see the RAE support; this comes next and would be exciting to have!

Already, the official RAE repo ships a custom Stage-2 DiT-like latent generator and public checkpoints/configs, but today these are not available through a native `diffusers` model + pipeline + checkpoint conversion flow.

**Describe the solution you'd like.**
In two stages, probably: (a) inference support for the checkpoints (b) training support.
I have some code ready that I've been able to use to get the checkpoints workable, could turn it around into a PR.

plugyawn:
@sayakpaul @kashif if y'all prefer, I could spin out a PR with some codexing on my current codebase for part (a).

PR #13231, opened 2026-03-09
Add RAE Diffusion Transformer

plugyawn:
# What does this PR do?
This PR adds support for [Diffusion Transformers with Representation Autoencoders](https://rae-dit.github.io) in Diffusers.

It implements the Stage-2 side of the RAE setup:
- `RAEDiT2DModel`
- `RAEDiTPipeline`
- checkpoint conversion for published upstream Stage-2 checkpoints
- API docs
- a small `examples/research_projects/rae_dit/` training scaffold

This addresses #13225.

Reference implementation: [byteriper's repository](https://github.com/bytetriper/RAE)

## Validation

Inference parity with the official implementation is high. For matched class label / initial latent noise / schedule, I measured:
- `max_abs_error=0.00001717`
- `mean_abs_error=0.00000122`

Qualitative parity artifacts used during validation:
- same published Stage-2 checkpoint
- same class label
- same initial latent noise
- same 25-step shifted Euler schedule

Inference is also slightly faster in the current Diffusers port on a 40GB A100:

| Precision | CFG | Steps | Diffusers sec/img | Upstream sec/img | Diffusers img/s | Delta |
|---|---:|---:|---:|---:|---:|---:|
| bf16 | 1.0 | 25 | 0.817 | 0.913 | 1.225 | +11.8% |
| bf16 | 4.0 | 25 | 0.852 | 0.931 | 1.174 | +9.3% |
| bf16 | 1.0 | 50 | 1.568 | 1.761 | 0.638 | +12.3% |
| bf16 | 4.0 | 50 | 1.649 | 1.853 | 0.606 | +12.4% |

## Notes

- This PR intentionally does not add upstream autoguidance / guidance-model support.
- The training script is a research-project scaffold under `examples/research_projects`, not a claim of full upstream training parity.
- `AutoencoderRAE.from_pretrained()` is used for the Stage-1 component so the packaged `RAEDiTPipeline.from_pretrained(...)` path works with published RAE checkpoints.

## Before submitting


plugyawn:
@kashif @sayakpaul would be great if you could review. Please note the no_init_weights() fix (details in the PR body); if you prefer, that could be a separate PR, but considering `diffusers` is supposed to be an extension to `torch`, I guess it makes sense?

sayakpaul:
Thanks for the PR. To keep the scope manageable, could we break it down into separate PRs?

For example,

>  there is also a change to no_init_weights( ). Specifically: it makes Diffusers’ skip-weight-init behave more like normal PyTorch. Now, when no_init_weights() is active, the torch.nn.init.* functions stop returning the tensor they were called on (for ref: PyTorch does return). Most models never notice this, but the RAE-DiT implementation does rely on the return value during construction, which can make otherwise valid checkpoints fail to load through the standard from_pretrained() path.

could be a separate PR.

examples/research_projects/rae_dit/README.md
```diff
@@ -0,0 +1,102 @@
...
+## What this mirrors
+
+The scaffold is deliberately composed from existing `diffusers` patterns instead of introducing a new training style:
+
+- `examples/research_projects/autoencoder_rae/train_autoencoder_rae.py`
+  for ImageFolder loading, RAE-specific preprocessing, and the experimental research-project placement.
+- `examples/dreambooth/train_dreambooth_flux.py`
+  for the flow-matching training loop structure, checkpoint resume flow, and `accelerate.save_state(...)` hooks.
+- `examples/flux-control/train_control_flux.py`
+  for the transformer-only save layout and SD3-style flow-matching timestep weighting helpers.
```
sayakpaul: Doesn't belong here.

src/diffusers/models/modeling_utils.py
```diff
@@ -215,7 +215,10 @@ def no_init_weights():
     """
 
     def _skip_init(*args, **kwargs):
-        pass
+        # Preserve the `torch.nn.init.*` return contract so third-party model
+        # constructors that chain on the returned tensor still work under
+        # `no_init_weights()`.
+        return args[0] if len(args) > 0 else None
```
sayakpaul: Can you provide an example?

tests/models/transformers/test_models_rae_dit_transformer2d.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def test_gradient_checkpointing_is_applied(self):
+        expected_set = {"RAEDiT2DModel"}
+        super().test_gradient_checkpointing_is_applied(expected_set=expected_set)
+
+    def test_effective_gradient_checkpointing(self):
+        super().test_effective_gradient_checkpointing(loss_tolerance=1e-4)
+
+    @unittest.skip(
+        "RAEDiT initializes the output head to zeros, so cosine-based layerwise casting checks are uninformative."
```
sayakpaul: I don't think this is the case? We can always skip layerwise casting for certain layer or layer groups here:
https://github.com/huggingface/diffusers/blob/a08c274c3326579c69b065701f0a7e1982be632a/src/diffusers/models/modeling_utils.py#L246

tests/models/transformers/test_models_rae_dit_transformer2d.py
```diff
@@ -0,0 +1,206 @@
...
+        block.adaLN_modulation[-1].weight.data.normal_(mean=0.0, std=0.02)
+        block.adaLN_modulation[-1].bias.data.normal_(mean=0.0, std=0.02)
+
+    model.final_layer.adaLN_modulation[-1].weight.data.normal_(mean=0.0, std=0.02)
+    model.final_layer.adaLN_modulation[-1].bias.data.normal_(mean=0.0, std=0.02)
+    model.final_layer.linear.weight.data.normal_(mean=0.0, std=0.02)
+    model.final_layer.linear.bias.data.normal_(mean=0.0, std=0.02)
+
+
+class RAEDiT2DModelTests(ModelTesterMixin, unittest.TestCase):
```
sayakpaul: Test should use the newly added model tester mixins. You can find an example in https://github.com/huggingface/diffusers/pull/13046

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+
+    raise ValueError(
+        "Cannot expand conditioning tokens without preserving their 2D layout: "
+        f"source length {hidden_states.shape[1]} is incompatible with target length {target_length}."
+    )
+
+
+def _ddt_modulate(hidden_states: torch.Tensor, shift: torch.Tensor | None, scale: torch.Tensor) -> torch.Tensor:
+    if shift is None:
+        shift = torch.zeros_like(scale)
```
sayakpaul: This is a small function, which is okay being present in the caller sites inline? 

We also probably don't need `_repeat_to_length()`.

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+
+    shift = _repeat_to_length(shift, hidden_states.shape[1])
+    scale = _repeat_to_length(scale, hidden_states.shape[1])
+    return hidden_states * (1 + scale) + shift
+
+
+def _ddt_gate(hidden_states: torch.Tensor, gate: torch.Tensor) -> torch.Tensor:
```
sayakpaul: Same as above.

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+    gate = _repeat_to_length(gate, hidden_states.shape[1])
+    return hidden_states * gate
+
+
+def _to_pair(value: int | tuple[int, int] | list[int], name: str) -> tuple[int, int]:
```
sayakpaul, line 63: Same as above.

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+    if isinstance(value, int):
+        return value, value
+
+    if len(value) != 2:
+        raise ValueError(f"`{name}` must be an int or a pair, but got {value}.")
+
+    return int(value[0]), int(value[1])
+
+
+def _rotate_half(hidden_states: torch.Tensor) -> torch.Tensor:
```
sayakpaul: Same as above

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+            qkv_bias=True,
+            qk_norm=use_qknorm,
+            use_rmsnorm=use_rmsnorm,
+        )
+
+        mlp_hidden_dim = int(hidden_size * mlp_ratio)
+        if use_swiglu:
+            self.mlp = SwiGLUFFN(hidden_size, int(2 * mlp_hidden_dim / 3))
+        else:
+            self.mlp = _ApproximateGELUMLP(hidden_size, mlp_hidden_dim)
```
sayakpaul: Can we use 
https://github.com/huggingface/diffusers/blob/a08c274c3326579c69b065701f0a7e1982be632a/src/diffusers/models/attention.py#L1682

for the feedforward?

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+        class_labels: torch.LongTensor,
+        train: bool,
+        force_drop_ids: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        if (train and self.dropout_prob > 0) or (force_drop_ids is not None):
+            class_labels = self.token_drop(class_labels, force_drop_ids=force_drop_ids)
+        return self.embedding_table(class_labels)
+
+
+class VisionRotaryEmbeddingFast(nn.Module):
```
sayakpaul: (nit): why is this called Fast?

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+        freqs_cos = self.freqs_cos
+        freqs_sin = self.freqs_sin
+        if repeat != 1:
+            freqs_cos = freqs_cos.repeat_interleave(repeat, dim=0)
+            freqs_sin = freqs_sin.repeat_interleave(repeat, dim=0)
+
+        return hidden_states * freqs_cos + _rotate_half(hidden_states) * freqs_sin
+
+
+class NormAttention(nn.Module):
```
sayakpaul: Let's implement it similar to 
https://github.com/huggingface/diffusers/blob/a08c274c3326579c69b065701f0a7e1982be632a/src/diffusers/models/transformers/transformer_flux.py#L75

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+        nn.init.constant_(self.s_embedder.proj.bias, 0)
+        nn.init.normal_(self.y_embedder.embedding_table.weight, std=0.02)
+        nn.init.normal_(self.t_embedder.mlp[0].weight, std=0.02)
+        nn.init.normal_(self.t_embedder.mlp[2].weight, std=0.02)
+
+        if self.use_pos_embed:
+            pos_embed = get_2d_sincos_pos_embed(
+                self.pos_embed.shape[-1], int(sqrt(self.pos_embed.shape[1])), output_type="pt"
+            )
+            self.pos_embed.data.copy_(pos_embed.float().unsqueeze(0))
```
sayakpaul: Can we use how https://github.com/huggingface/diffusers/pull/13046 initialized the position embeddings?

src/diffusers/models/transformers/transformer_rae_dit.py
```diff
@@ -0,0 +1,568 @@
...
+            raise ValueError("Sequence length must form a square grid for unpatchify.")
+
+        hidden_states = hidden_states.reshape(hidden_states.shape[0], height, width, patch_size, patch_size, channels)
+        hidden_states = torch.einsum("nhwpqc->nchpwq", hidden_states)
+        hidden_states = hidden_states.reshape(
+            hidden_states.shape[0], channels, height * patch_size, width * patch_size
+        )
+        return hidden_states
+
+    def _run_block(
```
sayakpaul: We don't need this. Let's instead follow this pattern:
https://github.com/huggingface/diffusers/blob/a08c274c3326579c69b065701f0a7e1982be632a/src/diffusers/models/transformers/transformer_flux.py#L714

src/diffusers/pipelines/rae_dit/pipeline_rae_dit.py
```diff
@@ -0,0 +1,259 @@
...
+        device: torch.device,
+    ) -> torch.LongTensor:
+        class_labels = torch.as_tensor(class_labels, device=device, dtype=torch.long).reshape(-1)
+
+        if num_images_per_prompt > 1:
+            class_labels = class_labels.repeat_interleave(num_images_per_prompt)
+
+        return class_labels
+
+    def _prepare_latents(
```
sayakpaul: It should be called `prepare_latents()` similar to other pipelines.

src/diffusers/pipelines/rae_dit/pipeline_rae_dit.py
```diff
@@ -0,0 +1,259 @@
...
+        if output_type == "latent":
+            output = latents
+        else:
+            images = self.vae.decode(latents).sample.clamp(0, 1)
+            if output_type == "pt":
+                output = images
+            else:
+                output = images.cpu().permute(0, 2, 3, 1).float().numpy()
+                if output_type == "pil":
+                    output = self.numpy_to_pil(output)
```
sayakpaul: We should use an image processor instead here. See:
https://github.com/huggingface/diffusers/blob/a08c274c3326579c69b065701f0a7e1982be632a/src/diffusers/pipelines/flux/pipeline_flux.py#L1012

src/diffusers/pipelines/rae_dit/pipeline_rae_dit.py
```diff
@@ -0,0 +1,259 @@
...
+
+        self.maybe_free_model_hooks()
+
+        if not return_dict:
+            return (output,)
+
+        return ImagePipelineOutput(images=output)
```
sayakpaul: Let's give this pipeline a separate output class: `RAEDiTPipelineOutput`.

sayakpaul (commented):
Thanks!

I left some initial comments, let me know if they make sense.

src/diffusers/models/transformers/transformer_rae_dit.py
plugyawn: Yeah, that makes sense, will do that.

plugyawn:
@sayakpaul, from what I understand the RAE checkpoint -> DiT checkpoint -> generation pipeline necessarily requires the no_init_weight() change (otherwise the semantics become a bit muddled, imo). 

Would it make more sense to open a PR for handling `no_init_weights()` behavior before this one?

sayakpaul:
Could you explain why that's needed? I am still not sure about that actually. Prefer providing specific examples that fail without the change for init.

plugyawn:
Not sure how to link files, but it seems to be related to changes introduced in #13046.

A specific example,
- `AutoencoderRAE` consturcts DinoV2WithRegistersModel. 
- ModelMixin.from_pretrained() does this construction under no_init_weights( ) first, before low_cpu_mem_usage kicks in (`modelling_utils.py`, around line 1300)
-   AutoencoderRAE constructs Dinov2WithRegistersModel(config) in _build_encoder:84, and
  ModelMixin.from_pretrained() always does that construction under no_init_weights() first, even
  before low_cpu_mem_usage matters; see modeling_utils.py:1270. In current `transformers`, DINOv2-
  with-registers has init code like this in modeling_dinov2_with_registers.py:464:
```
  module.weight.data = nn.init.trunc_normal_(
      module.weight.data.to(torch.float32), mean=0.0, std=self.config.initializer_range
  ).to(module.weight.dtype)
```

  Under today’s no_init_weights(), nn.init.trunc_normal_ is replaced with a stub that just passes
  and returns None, so that becomes None.to(...) and fails with an `AttributeError: 'NoneType' object has no attribute 'to'`.

Codex has a better summary, I think:
  > failing example: AutoencoderRAE builds Dinov2WithRegistersModel(config) in its encoder
  > path, and ModelMixin.from_pretrained() always instantiates models under no_init_weights() first.
  > In current transformers, DINOv2’s _init_weights() assigns the return value of
  > nn.init.trunc_normal_(...) and then calls .to(...) on it. With the current no_init_weights()
  > stub, that return value becomes None, so construction fails with AttributeError: 'NoneType'
  > object has no attribute 'to'. The proposed change keeps skip-init behavior intact, but restores
  > the normal PyTorch return contract so these constructors remain compatible.

  


Re: #13046, note `test_models_autoencoder_rae.py:45`, where the unit tests seem to be a little off, imo. Not sure the tests are aligned.
```
# ---------------------------------------------------------------------------
# Tiny test encoder for fast unit tests (no transformers dependency)
# ---------------------------------------------------------------------------


class _TinyTestEncoderModule(torch.nn.Module):
    """Minimal encoder that mimics the patch-token interface without any HF model."""

    def __init__(self, hidden_size: int = 16, patch_size: int = 8, **kwargs):
        super().__init__()
        self.patch_size = patch_size
        self.hidden_size = hidden_size

    def forward(self, images: torch.Tensor) -> torch.Tensor:
        pooled = F.avg_pool2d(images.mean(dim=1, keepdim=True), kernel_size=self.patch_size, stride=self.patch_size)
        tokens = pooled.flatten(2).transpose(1, 2).contiguous()
        return tokens.repeat(1, 1, self.hidden_size)


def _tiny_test_encoder_forward(model, images):
    return model(images)


def _build_tiny_test_encoder(encoder_type, hidden_size, patch_size, num_hidden_layers):
    return _TinyTestEncoderModule(hidden_size=hidden_size, patch_size=patch_size)


# Monkey-patch the dispatch tables so "tiny_test" is recognised by AutoencoderRAE
_ENCODER_FORWARD_FNS["tiny_test"] = _tiny_test_encoder_forward
_original_build_encoder = _build_encoder


def _patched_build_encoder(encoder_type, hidden_size, patch_size, num_hidden_layers):
    if encoder_type == "tiny_test":
        return _build_tiny_test_encoder(encoder_type, hidden_size, patch_size, num_hidden_layers)
    return _original_build_encoder(encoder_type, hidden_size, patch_size, num_hidden_layers)


_rae_module._build_encoder = _patched_build_encoder

```

I'm new to diffusers idiomatics, but I was confused why this appeared to be a problem only now, and asked GPT:

 
>  no_init_weights() only becomes a problem when all of these are true at once:
>
>  - a diffusers ModelMixin.from_pretrained() call is constructing the model
>  - that model’s __init__() instantiates another model internally
>  - that internal model uses torch.nn.init.* and also relies on its return value
>
>  RAE is unusual because it does exactly that. Inside autoencoder_rae.py, the AutoencoderRAE constructor directly >builds a transformers vision backbone:
>
> - Dinov2WithRegistersModel:98
> - SiglipVisionModel:111
> - ViTMAEModel:124
>
>  That is not how most other diffusers integrations are structured. Most of the repo does one of these instead:
>
>  - native diffusers models in src/diffusers/models, whose init code only relies on side effects
>  - pipelines that accept transformers models as separate top-level components, rather than constructing them inside > a ModelMixin
>
>  So other work usually does not run a transformers constructor inside diffusers’ patched no_init_weights() context.

sayakpaul:
> Not sure how to link files

Yes, we can link files and I think it's better this way. For example, it's much better to refer to specific lines like
https://github.com/huggingface/diffusers/blob/068c6ef6c16a79340d00d1069b3aa22f7acbe937/src/diffusers/models/modeling_utils.py#L129

instead of plain text.

Overall, I think that the explanation you provided in the above comment is that helpful. We need to have some specific (preferably very minimal) code snippet with and without that change to better understand what's happening and why. 

For this kind of PRs, it's an expectation that the contributors will try to take some time to understand the library code.

plugyawn:
@sayakpaul pushed a follow-up cleanup addressing the main scope issues from the earlier thread.

What changed:
- removed the unused `guidance_transformer` / guidance-model surface from the pipeline + converter, since the inference loop never used it
- localized the published Stage-1 loading fix in `AutoencoderRAE.from_pretrained()`, so this PR no longer depends on a cross-cutting `no_init_weights()` change
- updated the Stage-2 trainer + converter to use the stock `AutoencoderRAE.from_pretrained()` path
- added a regression test for the nested-init constructor case that motivated the loader change

Validation on the A100 after the cleanup:
- `68 passed, 6 skipped` across the RAE model / Stage-2 pipeline / converter / training-smoke slice
- `RAEDiTPipeline.from_pretrained(...)` still loads and generates a finite `(1, 3, 256, 256)` sample from the converted published pipeline
- the new `AutoencoderRAE.from_pretrained()` path matches the previous manual loader workaround exactly for generation (`max_abs_error=0.0`, `mean_abs_error=0.0`)

I also updated the PR description so it no longer mentions the old `no_init_weights()` direction or guidance-model support.

plugyawn:
Hi @sayakpaul! My bad, I'll update the PR today.

sayakpaul:
@plugyawn do you want to take another crack?

plugyawn:
Yess, I'm reading through the diffusers codebase in more detail (I had used it quite a bit but not dived as deep as I had for `transformers`, hehe), and waiting for some compute to come through (I ran out, should be back by this week). 

Thank you so much for waiting! I'll update the PR by this week.

plugyawn:
@sayakpaul, I removed old `no_init_weights()` change; as you said, there is a workaround for now. 
The nested-init issue is handled locally in `AutoencoderRAE` now.

The diff is pretty different from the last round, so if you have time, it'd be great if you could review it again! Sorry for the wait!

sayakpaul:
@kashif possible to do a round of review?

plugyawn:
Also, I'm pretty certain the merged RAE implementation has some faulty testing. 
I'm not sure how it works with the no_init_weights( ) problem unfixed; the tests also use a fake encoder. Might have gone under the radar?

plugyawn:
@kashif, @sayakpaul would you prefer a separate PR (or here) to fix the testing issues on the RAE?

kashif:
yes please, seperate!

plugyawn:
Ill take this as a yes for a separate PR!

sayakpaul:
@kashif could you do an initial review please?

plugyawn:
@kashif @sayakpaul is there a timeline on this?

plugyawn:
@sayakpaul, just a small bump. Have a free week, would love to get this to a mergeable state.

sayakpaul:
@claude could you do a review here?

plugyawn:
Good bot.

plugyawn:
Fixed the issues that Claude mentioned, sans the AttentionModuleMixin( ) change to RAEDiTAttention( ). 
@sayakpaul could you perhaps launch another review to confirm?

plugyawn:
@sayakpaul @kashif wanted to bump in case missed.

plugyawn:
Implemented the RAEDiT attention refactor, with QKV fusion support tested. Fixed-seed decoded sample comparison: max_abs_error=0.00006458, mean_abs_error=0.00000212

image

Tested on an A100 and an H100.

plugyawn:
@claude can you review this?

plugyawn:
Hi @kashif @sayakpaul, a small reminder. This is quickly becoming stale, and there's been quite a lot of work over RAE DiTs already now that can be worked on after this is merged.

plugyawn:
@sayakpaul I assume this PR is not in the pipeline?

iwr-redmond:
Are there any models on the Hub that can be used to test this PR with? Could you please provide a code snippet and some examples?

plugyawn:
Hi @iwr-redmond!
You can find a converted Diffusers-format test checkpoint here: https://huggingface.co/plugyawn/rae-dit-s-ep14-diffusers

This was converted from the public upstream assets:
  - Stage-1 RAE: https://huggingface.co/nyu-visionx/RAE-dinov2-wReg-base-ViTXL-n08
  - Stage-2 weights: https://huggingface.co/nyu-visionx/RAE-collections/blob/main/DiTs/Dinov2/wReg_base/ImageNet256/DiTDH-S_ep14/stage2_model.pt

 On this branch, run:

  ```python
  import torch
  from diffusers import RAEDiTPipeline

  pipe = RAEDiTPipeline.from_pretrained(
      "plugyawn/rae-dit-s-ep14-diffusers",
      torch_dtype=torch.bfloat16,
  ).to("cuda")

  generator = torch.Generator(device="cuda").manual_seed(0)

  image = pipe(
      class_labels=207,  # golden retriever
      num_inference_steps=25,
      guidance_scale=1.0,
      generator=generator,
  ).images[0]

  image.save("rae_dit_class207.png")
```
image


  Other ImageNet-256 Stage-2 checkpoints can be tested by converting them first:

```bash
  git clone https://github.com/bytetriper/RAE.git

  python scripts/convert_rae_stage2_to_diffusers.py \
    --repo_or_path nyu-visionx/RAE-collections \
    --config_repo_or_path ./RAE \
    --config_path configs/stage2/training/ImageNet256/DiTDH-S_DINOv2-B.yaml \
    --checkpoint_path DiTs/Dinov2/wReg_base/ImageNet256/DiTDH-S_ep14/stage2_model.pt \
    --vae_model_name_or_path nyu-visionx/RAE-dinov2-wReg-base-ViTXL-n08 \
    --output_path ./rae-dit-s-ep14 \
    --safe_serialization \
    --verify_load
```

Class 207 (in the example) corresponds to a golden retriever; this checkpoint is ImageNet class-conditioned only. 
Text-conditioned RAE models are part of Scale RAE: https://rae-dit.github.io/scale-rae/, which I was thinking of working on next, once this PR is complete.

plugyawn:
@sayakpaul @kashif 👀

kashif:
edit: correcting myself here. i dug into the test failures properly and the autoencoder picture is different (and better) than what i first wrote, so i've rewritten this comment.

thanks @plugyawn, this is in good shape and the parity numbers look great. did a full pass (model + pipeline + conversion + the autoencoder change).

the earlier-round items are all addressed: attention now goes through `RAEDiTAttnProcessor` + `dispatch_attention_fn` with `AttentionModuleMixin` and qkv fusion, `_no_split_modules` is set, the pipeline has `prepare_latents`/`VaeImageProcessor`/`RAEDiTPipelineOutput`, and `no_init_weights` is handled locally via `_preserve_init_return_tensors`. the model also matches the upstream DiTDH backbone (dual depth 28+2, 1152/2048, rmsnorm, qk-norm, swiglu, adaLN, rope+ape), and it's the same stage-2 backbone raev2 uses so it should carry forward.

on tests: run per file (the way CI shards them) only two things actually fail.

1. `tests/pipelines/rae_dit` `test_inference` fails on the hardcoded `expected_slice` (max_diff 0.0376 vs 1e-4). looks like it just wasn't regenerated after the attention/swiglu refactor, so please regenerate it and confirm it still matches upstream.

2. `test_from_pretrained_loads_nested_init_encoder_checkpoint` fails because `_preserve_init_return_tensors` only wraps the three real encoder constructors, not the `_build_encoder` call itself. the test's nested-init encoder slips past it and hits `trunc_normal_(...).to(...)` returning None under `no_init_weights`. wrapping the `_build_encoder(...)` call in `__init__` instead fixes it (tried it locally, the whole autoencoder file goes green) and it also covers custom encoders. while we're there: since this is a general `no_init_weights` gap (`TORCH_INIT_FUNCTIONS` is only used by `no_init_weights` and here, and no other model builds a transformers encoder in its `__init__`), it would be nicer as a small shared helper in `modeling_utils` next to `no_init_weights` rather than living in `autoencoder_rae`.

fwiw the larger batch of failures i first saw was just cross-file test pollution from running all the rae test files in one pytest session (it reproduces with the current main autoencoder too), so that's a test-isolation thing, not your code. apologies for the earlier noise on that.

minor:
- guidance: pipeline does plain CFG (+ interval) and skips autoguidance, which is fine to scope out, but worth a docs note that upstream RAE-v1 needs AutoGuidance since plain CFG barely moves gFID for RAE. the clean follow-up is the raev2 internal-guidance (x-prediction + reuse the REPA head, no aux model / no extra forward).
- model is already the DiTDH backbone raev2 uses; raev2's deltas are stage-1/training (generalized MLS encoder, REPA, dinov3).
- `wo_shift` still allocates `zeros_like` shift tensors every forward, can just branch the modulation.
- `x_pos_embed` is always None so the decoder-APE branch is dead, wire it or drop the attr.
- `_expand_conditioning_tokens` is heavy, is it ever hit by the shipped checkpoints (same patch size = no-op)? if not, simplify.
- one line on why the swiglu projection halves get swapped would help.

nice work, nothing architectural.

kashif:
pushed a small fix to your branch @plugyawn: widened `_preserve_init_return_tensors` to wrap the whole `_build_encoder` call in `__init__` (instead of just the 3 encoder constructors), so the nested-init test passes and custom encoders are covered too. autoencoder + conversion files are green per-file now. the only thing left is regenerating the pipeline `test_inference` expected_slice on your end (it's device-sensitive so i left it to you).

kashif:
tl;dr of the fix i pushed: the `no_init_weights` workaround was only wrapping the three built-in encoders, so any other encoder (including the one the nested-init test uses) slipped past it and hit `trunc_normal_(...).to(...)` on a None. moving the wrap up to the whole `_build_encoder` call covers all of them. autoencoder and conversion tests are green per-file now, just the pipeline `expected_slice` left to refresh on your side. thanks for bearing with the back-and-forth!

plugyawn:
Thanks @kashif! I'll update the PR with fixes for the minor issues you mentioned. 

Re: the tests,
>tests/pipelines/rae_dit test_inference fails on the hardcoded expected_slice (max_diff 0.0376 vs 1e-4). looks like it just wasn't regenerated after the attention/swiglu refactor, so please regenerate it and confirm it still matches upstream.

Will do! 

>test_from_pretrained_loads_nested_init_encoder_checkpoint fails because _preserve_init_return_tensors only wraps the three real encoder constructors, not the _build_encoder call itself. the test's nested-init encoder slips past it and hits trunc_normal_(...).to(...) returning None under no_init_weights. wrapping the _build_encoder(...) call in __init__ instead fixes it (tried it locally, the whole autoencoder file goes green) and it also covers custom encoders. while we're there: since this is a general no_init_weights gap (TORCH_INIT_FUNCTIONS is only used by no_init_weights and here, and no other model builds a transformers encoder in its __init__), it would be nicer as a **small shared helper in modeling_utils next to no_init_weights rather than living in autoencoder_rae**.

This makes sense to me. I'll add this.

Same with the minor fixes suggested.

plugyawn:
@kashif done!
Summary of changes:
  - refreshed the CPU `RAEDiTPipelineFastTests::test_inference` expected slice
  - moved `_preserve_init_return_tensors` next to `no_init_weights()` in `modeling_utils`
  - removed the `wo_shift` zero allocations and dead `x_pos_embed` branch
  - added the SwiGLU half-swap comment
  - added a docs note that this pipeline is ImageNet class-conditioned with plain CFG only

image
Upstream RAE vs Diffusers conversion for DiTDH-S_ep14, class 207, seed 0, 25 Euler steps. Left: upstream output, middle: Diffusers output, right: normalized absolute difference. Sample parity:
  max_abs_error=0.00013733, mean_abs_error=0.00000247.

kashif (approved):

sayakpaul:
@bot /style

sayakpaul:
@dg845 could you review this as well?

plugyawn:
Fixed for ruff checks.

kashif (approved):

plugyawn:
@dg845 @sayakpaul mostly sure this is merge-ready now!

plugyawn:
@dg845 @sayakpaul bump

plugyawn:
Are any changes needed? More than happy to put in some more time to get this to merged.

PR #13234, opened 2026-03-09
[tests] migrate group offloading tests to pytest

sayakpaul:


dg845 (commented):
Thanks for the PR! I get the following test failures when testing locally:

1. `TestGroupOffload`
    1. `test_warning_logged_if_group_offloaded_module_moved_to_accelerator`
    2. `test_warning_logged_if_group_offloaded_pipe_moved_to_accelerator`
2. `TestConditionalModuleGroupOffload`
    1. `test_warning_logged_if_group_offloaded_module_moved_to_accelerator`
    2. `test_warning_logged_if_group_offloaded_pipe_moved_to_accelerator`

I think this might be due to the fact that the two tests above use a `caplog` argument, but no `caplog` fixture is defined in the file.

sayakpaul:
@dg845 thanks for the review! I have fixed the issue. Also, as an FYI, caplog is a built-in pytest fixture — it doesn't need to be defined in the file.

dg845 (approved):
Thanks!

Issue #12613, opened 2025-11-08
wan 2.2 cause nan in latent in i2v


chaowenguo:
### Describe the bug

import asyncio, aiohttp.web, math, pathlib, torch, builtins, transformers, diffusers, PIL, av, io, numpy, uvloop, concurrent, multiprocessing, os, zipfile, sys
from modelscope.hub.snapshot_download import snapshot_download
from modelscope.hub.file_download import model_file_download
#import sd_embed.embedding_funcs

model_dir = snapshot_download('Wan-AI/Wan2.2-I2V-A14B-Diffusers')
onload_device=torch.device('cuda')
pipe = diffusers.WanImageToVideoPipeline.from_pretrained(model_dir, vae=diffusers.AutoencoderKLWan.from_pretrained(model_dir, subfolder='vae', torch_dtype=torch.float32).to(onload_device),
                                                        low_cpu_mem_usage=True,
                                                        torch_dtype=torch.bfloat16,
                                                        quantization_config=diffusers.quantizers.PipelineQuantizationConfig(quant_backend='bitsandbytes_4bit', quant_kwargs={'load_in_4bit':True, 'bnb_4bit_quant_type':'nf4', 'bnb_4bit_compute_dtype':torch.bfloat16}, components_to_quantize=['transformer', 'transformer_2']))
diffusers.hooks.apply_group_offloading(pipe.text_encoder, onload_device=onload_device, offload_type='leaf_level', use_stream=True, record_stream=True, non_blocking=True, low_cpu_mem_usage=True)
pipe.transformer.enable_group_offload(onload_device=onload_device, offload_type='leaf_level', use_stream=True, record_stream=True, non_blocking=True, low_cpu_mem_usage=True)
pipe.transformer.enable_layerwise_casting(storage_dtype=torch.float8_e4m3fn, compute_dtype=torch.bfloat16)
pipe.transformer_2.enable_group_offload(onload_device=onload_device, offload_type='leaf_level', use_stream=True, record_stream=True, non_blocking=True, low_cpu_mem_usage=True)
pipe.transformer_2.enable_layerwise_casting(storage_dtype=torch.float8_e4m3fn, compute_dtype=torch.bfloat16)
pipe.vae.enable_tiling()
pipe.vae.enable_slicing()

async def config(request):
    return aiohttp.web.json_response({1:2})

def decode_tensors(pipe, step, timestep, callback_kwargs):
    latents = callback_kwargs["latents"]
    print(step, latents)
[...]

PR #13238, opened 2026-03-10
fix: restore module buffers during stream-based group offload

Chase-Xuu:
## Description

In `_offload_to_memory()`, when using CUDA streams (`self.stream is not None`), module buffers from `self.modules` were not being restored to their CPU tensor copies during offload. This created an asymmetry with both `_build_cpu_param_dict()` and `_process_tensors_from_modules()` (used during onload), which correctly iterate over `group_module.buffers()`.

### The bug

| Method | `group_module.parameters()` | `group_module.buffers()` | `self.parameters` | `self.buffers` |
|--------|:---:|:---:|:---:|:---:|
| `_build_cpu_param_dict` | ✅ | ✅ | ✅ | ✅ |
| `_process_tensors_from_modules` (onload) | ✅ | ✅ | ✅ | ✅ |
| `_offload_to_memory` (stream path) | ✅ | ❌ **missing** | ✅ | ✅ |
| `_offload_to_memory` (non-stream path) | ✅ (via `.to()`) | ✅ (via `.to()`) | ✅ | ✅ |

The non-stream path uses `group_module.to(self.offload_device)` which correctly moves all parameters and buffers. The stream path manually iterates but was missing the buffer loop.

### Impact

- Module buffers (e.g., `running_mean`/`running_var` in normalization layers) remain on GPU after offload
- On the next onload cycle, stale GPU buffer data may be used instead of the correct CPU copies
- This could contribute to NaN values when using `record_stream=True` with group offloading (related: #12613)
- Minor GPU memory leak from unreleased buffer tensors

### Fix

Added the missing `group_module.buffers()` loop in the stream path of `_offload_to_memory()`, making it symmetric with the onload path.

### Related Issues

- #12613 (NaN latents with `record_stream=True` in group offloading)

### DIFF
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -246,6 +246,8 @@ class ModuleGroup:
             for group_module in self.modules:
                 for param in group_module.parameters():
                     param.data = self.cpu_param_dict[param]
+                for buffer in group_module.buffers():
+                    buffer.data = self.cpu_param_dict[buffer]
             for param in self.parameters:
                 param.data = self.cpu_param_dict[param]
             for buffer in self.buffers:


Chase-Xuu:
Closing to comply with one-PR-per-project policy. Will resubmit this buffer offload fix after #13240 is resolved. The stream-based group offload buffer issue is real — I'll track it separately.

DN6:
@Chase-Xuu Would you consider reopening this for review?

PR #13242, opened 2026-03-10
[tests] migrate test_hooks.py to pytest

sayakpaul:


dg845 (approved):
Thanks for the PR! I opened a PR to this branch with a fix that creates the generator on `torch_device` rather than CPU, which solves device mismatch errors in some tests locally (e.g. `test_inference`, `test_stateful_hook`, etc.): #13871

PR #13247, opened 2026-03-10
[tests] Tests for conditional pipeline blocks

sayakpaul:
# What does this PR do?

Implements a test suite for the conditional blocks for the following functionalities:

* `get_workflow()` works as expected.
* `select_blocks()` works as expected.
* Automatic workflow selection works as expected.

The tests were written with components emulating the structure that we might get in the real world.

Additionally, I moved the custom block requirement-related tests to `test_modular_pipelines_custom_blocks.py`.


yiyixuxu (approved):
thanks

PR #13248, opened 2026-03-10
[docs] updates

stevhliu:
a few maintenance items:

- move "Guiders" out of Modular Diffusers (because its not exclusive to it). i think after "Schedulers" is the most logical place for guiders as they're both swappable components and it has a much higher visibility there
- remove `torch.jit.trace` in favor of `torch.compile`
- update "memory#memory-efficient-attention" with a link to the "Attention backend" docs and `set_attention_backend`


docs/source/en/optimization/memory.md
```diff
@@ -482,144 +482,6 @@ print(
 )  # (2880, 1, 960, 320) having a stride of 1 for the 2nd dimension proves that it works
 ```
 
-## torch.jit.trace
```
sayakpaul, line 485: Should we completely remove or put under a collapsible block?

stevhliu: i think completely removing is better to keep things simple and avoid any confusion about why its in a collapsible block.

  we could keep it in a collapsible block though if you think there are still some legit use cases for `torch.jit.trace` that `torch.compile` doesn't cover

PR #13264, opened 2026-03-13
[core] Flux2 klein kv followups

sayakpaul:
# What does this PR do?

* Introduces `Flux2Transformer2DModelOutput` class as its outputs differ from the rest of the lot.
* Test suite for Klein KV.
* Docs.


yiyixuxu (approved):
thanks!

Issue #13177, opened 2026-02-23
Flux2KleinPipeline doesn't accept tensor format as input image


joansc:
**Is your feature request related to a problem? Please describe.**

I would like to input a tensor in the Flux2KleinPipeline  image argument. Only `image: list[PIL.Image.Image] | PIL.Image.Image | None = None` is accepted although in the comment section below the arguments it states that:

`image (`torch.Tensor`, `PIL.Image.Image`, `np.ndarray`, `List[torch.Tensor]`, `List[PIL.Image.Image]`, or `List[np.ndarray]`):
                `Image`, numpy array or tensor representing an image batch to be used as the starting point. For both
                numpy array and pytorch tensor, the expected value range is between `[0, 1]` If it's a tensor or a list
                or tensors, the expected shape should be `(B, C, H, W)` or `(C, H, W)`. If it is a numpy array or a
                list of arrays, the expected shape should be `(B, H, W, C)` or `(H, W, C)` It can also accept image
                latents as `image`, but if passing latents directly it is not encoded again.`

Thanks in advance,

Joan

abhixb:
can i get this issue assigned

abhixb:
Hi, I'd like to work on this.  I'll add a tensor/numpy → PIL conversion step and update the type annotations.

joansc:
This still hasn't been officially addressed yet...

abhixb:
i came up with a better fix but maintainers told me this issue will be addressed in another fix

asomoza:
Hi @abhixb, I will take this opportunity offer my apology if I sounded too harsh in the other PR, we now have a better guideline for LLMs to work with PRs, and also I forgot to ask the other contributor to fix this issue in the other pipelines. The [inpainting ](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/flux2/pipeline_flux2_klein_inpaint.py)one manages this issue using the diffuser's standard. 

If you want to still contribute it, you can use it as a reference for addressing this issue, just remember that you can't just change a method that other pipelines use without a real reason and a check that it doesn't break the other pipelines.

Also I believe we still need tests for these cases.

abhixb:
hi @asomoza 
your comments about my PR were completely valid dont worry i didnt take it to heart 
besides my first implementation was genuinely wrong and it was my first time making a PR so yeah

about fixing this PR with the inpainting method i will try as i m busy with some other work as well 
Thank you

PR #13265, opened 2026-03-14
 add tensor and numpy array support for image input in Flux2KleinPipeline

abhixb:
Fixed issue #13177

## The problem

If you pass a tensor or numpy array as the image to the Klein pipeline, it errors even though the docstring explicitly says these types are supported. Three things cause this:

- The image processor's validation method has a hard isinstance check for PIL and just rejects everything else on the spot
- The image loop in the pipeline's call method does `image_width, image_height = img.size` which is PIL only on a tensor .size is a method, on a numpy array it's a single int, both explode when you try to unpack them
- The type annotation on the image parameter only mentions PIL types, so your IDE would tell you not to even try

## The fix

- Added a to_pil helper on the image processor that converts tensors and numpy arrays into PIL images, PIL inputs just pass through
- Wired it into the validation method so it converts first, validates second, and returns a guaranteed PIL image
- The call method's image loop now uses that return value instead of throwing it away
- Updated the type annotation to match the docstring

Nothing changes for people already passing PIL images, same code path as before.

## Tests

Added tests for tensor input, numpy input, and a list of tensors for multi-reference editing. Existing PIL test still passes. All 4 green locally.

### DIFF
--- a/src/diffusers/pipelines/flux2/image_processor.py
+++ b/src/diffusers/pipelines/flux2/image_processor.py
@@ -14,7 +14,9 @@
 
 import math
 
+import numpy as np
 import PIL.Image
+import torch
 
 from ...configuration_utils import register_to_config
 from ...image_processor import VaeImageProcessor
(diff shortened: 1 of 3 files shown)


abhixb:
@sayakpaul hello can u please look into/review this

asomoza:
thanks @abhixb, this is not correct and will need a complete rewrite, first you're changing a method in the image processor which this is not the only pipeline that calls it, you're creating unnecessary tests and one of those is completely wrong and you're converting tensors to a pil image, but then the pipeline has to convert it back to tensors...

Please provide a correct solution, and just in case, I won't review this issue again if it's done completely by the llm without you understanding what is it doing and what should be done to fix the issue.

PR #13269, opened 2026-03-15
[agents]support skills

yiyixuxu:


.ai/AGENTS.md
```diff
@@ -24,54 +24,10 @@ Strive to write code as simple and explicit as possible.
 
 ### Models
 - All layer calls should be visible directly in `forward` — avoid helper functions that hide `nn.Module` calls.
-- Try to not introduce graph breaks as much as possible for better compatibility with `torch.compile`. For example, DO NOT arbitrarily insert operations from NumPy in the forward implementations.
-- Attention must follow the diffusers pattern: both the `Attention` class and its processor are defined in the model file. The processor's `__call__` handles the actual compute and must use `dispatch_attention_fn` rather than calling `F.scaled_dot_product_attention` directly. The attention class inherits `AttentionModuleMixin` and declares `_default_processor_cls` and `_available_processors`.
+- Avoid graph breaks for `torch.compile` compatibility — do not insert NumPy operations in forward implementations.
```
sayakpaul:
```suggestion
- Avoid graph breaks for `torch.compile` compatibility — do not insert NumPy operations in forward implementations and any other patterns that can break `torch.compile` compatibility with `fullgraph=True`.
```

.ai/AGENTS.md
```diff
@@ -24,54 +24,10 @@ Strive to write code as simple and explicit as possible.
...
+- See the **model-integration** skill for the attention pattern, pipeline rules, and test setup details.
```
sayakpaul:
```suggestion
- See the **model-integration** skill for the attention pattern, pipeline rules, test setup instructions, and other important details.
```

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+  models/<model>_transformer3d.md                # or appropriate name
+```
+
+### Integration checklist
+
+- [ ] Implement transformer model with `from_pretrained` support
+- [ ] Implement or reuse scheduler
+- [ ] Implement pipeline(s) with `__call__` method
+- [ ] Add LoRA support if applicable
+- [ ] Register all classes in `__init__.py` files (lazy imports)
```
sayakpaul, line 62:
```suggestion
- [ ] Register all classes in `__init__.py` files (lazy imports)
  - [ ] src/diffusers/models/__init__.py
  - [ ] src/schedulers/__init__.py (if applicable)
  - [ ] src/diffusers/pipelines/__init__.py
  - [ ] src/diffusers/loaders/__init__.py
  - [ ] src/diffusers/__init__.py
```

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+- Models use `ModelMixin` with `register_to_config` for config serialization
+- Schedulers use `SchedulerMixin` with `ConfigMixin`
+- Use `@torch.no_grad()` on pipeline `__call__`
+- Support `output_type="latent"` for skipping VAE decode
+- Support `generator` parameter for reproducibility
+- Use `self.progress_bar(timesteps)` for progress tracking
+
+---
+
+## Modular Pipeline Conversion
```
sayakpaul: Should this be a different skill? `modular-pipeline-conversion`?

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+### Common diffusers conventions
+
+- Pipelines inherit from `DiffusionPipeline`
+- Models use `ModelMixin` with `register_to_config` for config serialization
+- Schedulers use `SchedulerMixin` with `ConfigMixin`
+- Use `@torch.no_grad()` on pipeline `__call__`
+- Support `output_type="latent"` for skipping VAE decode
+- Support `generator` parameter for reproducibility
+- Use `self.progress_bar(timesteps)` for progress tracking
+
```
sayakpaul, line 134: `if is_my_dependency_available(): import my_dependency` should be a multi-line code snippet but the suggestion block wouldn't have rendered it well.

```suggestion
### Misc

**Dealing with new dependencies:**

Don't arbitrarily add new dependencies even if the reference code has it. Always try to implement operations with pure PyTorch first. For example, anything implemented with `einops` can be implemented with just PyTorch.

If you need to rely on a particularly dependency its import should be guarded:

`if is_my_dependency_available(): import my_dependency`

```

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+  __init__.py                          # Lazy imports
+  modular_pipeline.py                  # Pipeline class (tiny, mostly config)
+  encoders.py                          # Text encoder + image/video VAE encoder blocks
+  before_denoise.py                    # Pre-denoise setup blocks
+  denoise.py                           # The denoising loop blocks
+  decoders.py                          # VAE decode block
+  modular_blocks_<model>.py            # Block assembly (AutoBlocks)
+```
+
+### Block types decision tree
```
sayakpaul: Should we provide guidance on input processing? I think those are quite prevalent for the QwenImage series?

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+                hidden_states=latents, timestep=timestep,
+                return_dict=False, **cond_kwargs, **shared_kwargs,
+            )[0]
+        components.guider.cleanup_models(components.transformer)
+
+    noise_pred = components.guider(guider_state)[0]
+    latents = components.scheduler.step(noise_pred, t, latents, generator=generator)[0]
+```
+
+### Key pattern: Chunk loops for video models
```
sayakpaul: Perhaps we could also provide a reference for it study if needed. And then take a raw implementation from the non-modular pipeline class and show its relevant modular implementation in this context? 

Or do you think this is sufficient? I guess we can iterate as we go.

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+
+Use `LoopSequentialPipelineBlocks` for outer loop:
+```python
+class ChunkDenoiseStep(LoopSequentialPipelineBlocks):
+    block_classes = [PrepareChunkStep, NoiseGenStep, DenoiseInnerStep, UpdateStep]
+```
+
+Note: blocks inside `LoopSequentialPipelineBlocks` receive `(components, block_state, k)` where `k` is the loop iteration index.
+
+### Key pattern: Workflow selection
```
sayakpaul: Should also add some notes on implementing `select_workflow()` and `select_blocks()`?

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+
+```python
+class AutoDenoise(ConditionalPipelineBlocks):
+    block_classes = [V2VDenoiseStep, I2VDenoiseStep, T2VDenoiseStep]
+    block_trigger_inputs = ["video_latents", "image_latents"]
+    default_block_name = "text2video"
+```
+
+### Standard InputParam/OutputParam templates
```
sayakpaul: Perhaps a few notes about which ones to use when?

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+
+```python
+# Heavy models - loaded from pretrained
+ComponentSpec("transformer", YourTransformerModel)
+ComponentSpec("vae", AutoencoderKL)
+
+# Lightweight objects - created inline from config
+ComponentSpec("guider", ClassifierFreeGuidance,
+              config=FrozenDict({"guidance_scale": 7.5}),
+              default_creation_method="from_config")
```
sayakpaul:
```suggestion
ComponentSpec(
    "guider", 
    ClassifierFreeGuidance,
    config=FrozenDict({"guidance_scale": 7.5}),
    default_creation_method="from_config"
)
```

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+### Conversion checklist
+
+- [ ] Read original pipeline's `__call__` end-to-end, map stages
```
sayakpaul: Let's also add a note on adding tests?

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+- [ ] Create file structure under `modular_pipelines/<model>/`
+- [ ] Write decoder block (simplest)
+- [ ] Write encoder blocks (text, image, video)
+- [ ] Write before_denoise blocks (timesteps, latent prep, noise)
+- [ ] Write denoise block with guider abstraction (hardest)
+- [ ] Create pipeline class with `default_blocks_name`
+- [ ] Assemble blocks in `modular_blocks_<model>.py`
+- [ ] Wire up `__init__.py` with lazy imports
+- [ ] Run `make style`
+- [ ] Test all workflows for parity with reference
```
sayakpaul: Do we want pairty with a non-modular pipeline created with dummy components (easily doable through the existing fast tests)?

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
+---
+name: Parity Testing
+description: >
+  Testing pipeline parity between reference and diffusers implementations:
+  checkpoint mechanism, stage tests (encode/decode/denoise), injection debugging,
+  visual comparison, comparison utilities, and 18 common pitfalls.
+  Trigger: debugging parity, writing conversion tests, investigating divergence.
```
sayakpaul:
```suggestion
  Testing pipeline parity between reference and diffusers implementations:
  checkpoint mechanism, stage tests (encode/decode/denoise), injection debugging,
  visual comparison, comparison utilities, and common pitfalls.
  Trigger: debugging parity, writing conversion tests, investigating divergence.
```

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+---
+
+## Testing Pipeline Parity
+
+Applies to any conversion: research repo -> diffusers, standard -> modular, or research repo -> modular.
+
+## Principles
+
+1. **Don't combine structural changes with behavioral changes.** For research repo -> diffusers, you must restructure code to fit diffusers APIs (ModelMixin, ConfigMixin, etc.) -- that's unavoidable. But don't also "improve" the algorithm, refactor computation order, or rename internal variables for aesthetics. Keep numerical logic as close to the reference as possible, even if it looks ugly. For standard -> modular, this is stricter: copy loop logic verbatim and only restructure into blocks. In both cases, clean up in a separate commit after parity is confirmed.
```
sayakpaul:
```suggestion
1. **Don't combine structural changes with behavioral changes.** For research repo -> diffusers, first, you must restructure code to fit diffusers APIs (ModelMixin, ConfigMixin, etc.) -- that's unavoidable. But don't also "improve" the algorithm, refactor computation order, or rename internal variables for aesthetics. Keep numerical logic as close to the reference as possible, even if it looks unclean. For standard -> modular, this is stricter: copy loop logic verbatim and only restructure into blocks. In both cases, clean up in a separate commit after parity is confirmed.
```

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+
+2. **Match the reference noise generation first.** The way initial noise/latents are constructed (seed handling, generator, randn call order) often differs between reference and diffusers. If the noise doesn't match, nothing downstream will match, making it impossible to isolate other bugs. Strategy: in the first implementation, replicate the reference's exact noise construction to get parity. After everything else is confirmed working, swap to diffusers-style noise generation as a final step.
```
sayakpaul:
```suggestion
2. **Match the reference noise generation first.** The way initial noise/latents are constructed (seed handling, generator, randn call order) often differs between reference and diffusers. If the noise doesn't match, nothing downstream will match, making it impossible to isolate other bugs. Strategy: in the first implementation, replicate the reference's exact noise construction to get parity. After everything else is confirmed working, swap to diffusers-style noise generation as a final step.The same goes for steps, including text encoding, reference image encoding, etc.
```

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+- **`denoise`** (test last): Run both pipelines with realistic `num_steps` (e.g. 30) so the scheduler computes correct sigmas/timesteps, but stop after 2 loop iterations using `after_step_1`. Don't set `num_steps=2` -- that produces unrealistic sigma schedules. Compare the latents after those 2 steps.
+
+```python
+# Encode stage -- stop before the loop, compare ALL inputs:
+ref_ckpts = {"preloop": Checkpoint(save=True, stop=True)}
+run_reference_pipeline(ref_ckpts)
+ref_data = ref_ckpts["preloop"].data
+
+diff_ckpts = {"preloop": Checkpoint(save=True, stop=True)}
+run_diffusers_pipeline(diff_ckpts)
```
sayakpaul, line 87: Should we add some notes about the skeleton of `run_diffusers_pipeline`?

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+Both pipelines generate independently with identical seeds/params. Save outputs and compare visually. If outputs look identical, you're done -- no need for deeper testing.
+
+**Pipeline stage tests -- only if E2E fails and you need to isolate the bug.**
+For small models, run on CPU/float32 for strict comparison. For large models (e.g. 22B params), CPU/float32 is impractical -- use GPU/bfloat16 with `enable_model_cpu_offload()` and relax tolerances (max_diff < 1e-1 for bfloat16 is typical for passing tests; cosine similarity > 0.9999 is a good secondary check).
+
+Test encode and decode stages first -- they're simpler and bugs there are easier to fix. Only debug the denoising loop if encode and decode both pass.
+
+The challenge: pipelines are monolithic `__call__` methods -- you can't just call "the encode part". The solution is a checkpoint mechanism that lets you stop, save, or inject tensors at named locations inside the pipeline.
+
+**Add a `_checkpoints` argument to both pipelines.**
```
sayakpaul:
```suggestion
**Add a `_checkpoints` argument to both the diffusers pipeline and the reference implementation.**
```

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+    ckpt = _checkpoints.get("preloop")
+    if ckpt is not None and ckpt.load:
+        latents = ckpt.data["latents"].to(device=device, dtype=latents.dtype)
+```
+
+For large models, free the source pipeline's GPU memory before loading the target pipeline. Clone injected tensors to CPU, delete everything else, then run the target with `enable_model_cpu_offload()`.
+
+**Per-step accumulation tracing**: When injection confirms the loop is correct but you want to understand *how* a small initial difference compounds, capture `after_step_{i}` for every step and plot the max_diff curve. A healthy curve stays bounded; an exponential blowup in later steps points to an amplification mechanism (see Pitfall #13).
+
+## Debugging technique: Visual comparison via frame extraction
```
sayakpaul, line 118: Very cool! Learned it myself today.

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+    max_diff = diff.max().item()
+    mean_diff = diff.mean().item()
+    cos = torch.nn.functional.cosine_similarity(
+        a.float().flatten().unsqueeze(0), b.float().flatten().unsqueeze(0)
+    ).item()
+    passed = max_diff < tol
+    print(f"  {'PASS' if passed else 'FAIL'} {name}: max={max_diff:.2e}, mean={mean_diff:.2e}, cos={cos:.5f}")
+    return passed
+```
+Cosine similarity is especially useful for GPU/bfloat16 tests where max_diff can be noisy -- `cos > 0.9999` is a strong signal even when max_diff exceeds tolerance.
```
sayakpaul, line 166: `max_diff` shouldn't exceed tolerance, no? Should the statement be: "... even when `max_diff` doesn't exceed `tol`"?

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+
+### 16. Incomplete injection tests -- inject ALL variables or the test is invalid
+
+When doing injection tests (feeding reference tensors into the diffusers pipeline), you must inject **every** divergent input, including sigmas/timesteps. A common mistake: the preloop checkpoint saves sigmas but the injection code only loads latents and embeddings. The test then runs with different sigma schedules, making it impossible to isolate the real cause.
+
+**Prevention**: After writing injection code, verify by listing every variable the injected stage consumes and checking each one is either (a) injected from reference, or (b) confirmed identical between pipelines.
+
+### 17. bf16 connector/encoder divergence -- don't chase it
+
+When running on GPU/bfloat16, multi-layer encoders (e.g. 8-layer connector transformers) accumulate bf16 rounding noise that looks alarming (max_diff 0.3-2.7). Before investigating, re-run the component test on CPU/float32. If it passes (max_diff < 1e-4), the divergence is pure precision noise, not a code bug. Don't spend hours tracing through layers -- confirm on CPU/f32 and move on.
```
sayakpaul: Fair point. Do you think it might get confusing for the agent, especially when the context of the above points, where we are suggesting points like "### 6. Float precision differences -- don't dismiss them"? 

Oftentimes, the outputs of the encoder/connector/decoder components mix with the noisy latents, and they influence the final output.

sayakpaul (commented):
Very useful!

We could also think of adding a skill for doing Krea realtime like modular pipelines.

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+2. **Match the reference noise generation first.** The way initial noise/latents are constructed (seed handling, generator, randn call order) often differs between reference and diffusers. If the noise doesn't match, nothing downstream will match, making it impossible to isolate other bugs. Strategy: in the first implementation, replicate the reference's exact noise construction to get parity. After everything else is confirmed working, swap to diffusers-style noise generation as a final step.
+
+3. **Test from the start.** Have component tests ready BEFORE writing conversion code. Test bottom-up: components first, then pipeline stages, then e2e.
+
+## Test strategy
+
+**Component parity (CPU/float32) -- always run, as you build.**
+Test each component before assembling the pipeline. This is the foundation -- if individual pieces are wrong, the pipeline can't be right. Each component in isolation, strict max_diff < 1e-3. Two modes:
+- **Fresh**: convert from checkpoint weights, compare against reference (catches conversion bugs)
+- **Saved**: load from saved model on disk, compare against reference (catches stale saves)
```
stevhliu:
```suggestion
Test each component before assembling the pipeline. This is the foundation -- if individual pieces are wrong, the pipeline can't be right. Each component in isolation, strict max_diff < 1e-3.

Test freshly converted checkpoints and saved checkpoints.
- **Fresh**: convert from checkpoint weights, compare against reference (catches conversion bugs)
- **Saved**: load from saved model on disk, compare against reference (catches stale saves)
```

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,387 @@
...
+        diff_model = DiffusersModel.from_pretrained(model_path, torch_dtype=torch.float32)
+    diff_model = diff_model.float().eval()
+    diff_out = diff_model(x)
+    del diff_model
+
+    # 4. Compare in same script -- no saving to disk
+    max_diff = (ref_out - diff_out).abs().max().item()
+    assert max_diff < 1e-3, f"FAIL: max_diff={max_diff:.2e}"
+```
+Key points: (a) both sides in one script -- never split into separate scripts that save/load intermediates, (b) deterministic input via seeded generator, (c) load one model at a time to fit in CPU RAM, (d) `.clone()` the reference output before deleting the model.
```
stevhliu:
```suggestion
Key points: (a) both reference and diffusers component in one script -- never split into separate scripts that save/load intermediates, (b) deterministic input via seeded generator, (c) load one model at a time to fit in CPU RAM, (d) `.clone()` the reference output before deleting the model.
```

.ai/skills/parity-testing/SKILL.md
yiyixuxu: yes, we should save some code examples in the folder (will do it in next integration when I have one at hand, not going to make up one now for this:)

.ai/skills/parity-testing/SKILL.md
yiyixuxu: `max_diff` isn't too good a metric when we cannot run & compare in float32 
cos can give you an idea if the output will look visually the same or not

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,284 @@
...
+- Support `output_type="latent"` for skipping VAE decode
+- Support `generator` parameter for reproducibility
+- Use `self.progress_bar(timesteps)` for progress tracking
+### Misc
+
+**Dealing with new dependencies:**
+
+Don't arbitrarily add new dependencies even if the reference code has it. Always try to implement operations with pure PyTorch first. For example, anything implemented with `einops` can be implemented with just PyTorch.
+
+If you need to rely on a particularly dependency its import should be guarded:
```
dg845:
```suggestion
If you need to rely on a particular dependency its import should be guarded:
```
nit: typo

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,284 @@
...
+- [ ] Write test scripts (reference + target) with identical seeds
+- [ ] Create file structure under `modular_pipelines/<model>/`
+- [ ] Write decoder block (simplest)
+- [ ] Write encoder blocks (text, image, video)
+- [ ] Write before_denoise blocks (timesteps, latent prep, noise)
+- [ ] Write denoise block with guider abstraction (hardest)
+- [ ] Create pipeline class with `default_blocks_name`
+- [ ] Assemble blocks in `modular_blocks_<model>.py`
+- [ ] Wire up `__init__.py` with lazy imports
+- [ ] Run `make style`
```
dg845:
```suggestion
- [ ] Run `make style` and `make quality`
```
nit: I think this might be more complete?

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Principles
+
+1. **Don't combine structural changes with behavioral changes.** For research repo -> diffusers, first, you must restructure code to fit diffusers APIs (ModelMixin, ConfigMixin, etc.) -- that's unavoidable. But don't also "improve" the algorithm, refactor computation order, or rename internal variables for aesthetics. Keep numerical logic as close to the reference as possible, even if it looks unclean. For standard -> modular, this is stricter: copy loop logic verbatim and only restructure into blocks. In both cases, clean up in a separate commit after parity is confirmed.
+
+2. **Match the reference noise generation first.** The way initial noise/latents are constructed (seed handling, generator, randn call order) often differs between reference and diffusers. If the noise doesn't match, nothing downstream will match, making it impossible to isolate other bugs. Strategy: in the first implementation, replicate the reference's exact noise construction to get parity. After everything else is confirmed working, swap to diffusers-style noise generation as a final step.The same goes for steps, including text encoding, reference image encoding, etc.
```
dg845:
```suggestion
2. **Match the reference noise generation first.** The way initial noise/latents are constructed (seed handling, generator, randn call order) often differs between reference and diffusers. If the noise doesn't match, nothing downstream will match, making it impossible to isolate other bugs. Strategy: in the first implementation, replicate the reference's exact noise construction to get parity. After everything else is confirmed working, swap to diffusers-style noise generation as a final step. The same goes for other steps, including text encoding, reference image encoding, etc.
```
nit: may be more clear?

.ai/skills/parity-testing/pitfalls.md
```diff
@@ -0,0 +1,116 @@
...
+
+## 16. Incomplete injection tests -- inject ALL variables or the test is invalid
+
+When doing injection tests (feeding reference tensors into the diffusers pipeline), you must inject **every** divergent input, including sigmas/timesteps. A common mistake: the preloop checkpoint saves sigmas but the injection code only loads latents and embeddings. The test then runs with different sigma schedules, making it impossible to isolate the real cause.
+
+**Prevention**: After writing injection code, verify by listing every variable the injected stage consumes and checking each one is either (a) injected from reference, or (b) confirmed identical between pipelines.
+
+## 17. bf16 connector/encoder divergence -- don't chase it
+
+When running on GPU/bfloat16, multi-layer encoders (e.g. 8-layer connector transformers) accumulate bf16 rounding noise that looks alarming (max_diff 0.3-2.7). Before investigating, re-run the component test on CPU/float32. If it passes (max_diff < 1e-4), the divergence is pure precision noise, not a code bug. Don't spend hours tracing through layers -- confirm on CPU/f32 and move on.
```
dg845:
```suggestion
When running on GPU/bfloat16, multi-layer encoders (e.g. 8-layer connector transformers) accumulate bf16 rounding noise that looks alarming (max_diff 0.3-2.7). Before investigating, re-run the component test on CPU/float32. If it passes (max_diff < 1e-4), the divergence is pure precision noise, not a code bug. Don't spend hours tracing through layers -- confirm on CPU/float32 and move on.
```
nit: typo

.ai/skills/parity-testing/SKILL.md
```diff
@@ -0,0 +1,173 @@
...
+2. **Never monkey-patch model internals in tests.** Do not replace `model.forward` or patch internal methods.
+3. **Debugging instrumentation must be non-destructive.** Checkpoint captures for debugging are fine, but must not alter control flow or outputs.
+4. **Prefer CPU/float32 for numerical comparison when practical.** Float32 avoids bfloat16 precision noise that obscures real bugs. But for large models (22B+), GPU/bfloat16 with `enable_model_cpu_offload()` is necessary -- use relaxed tolerances and cosine similarity as a secondary metric.
+5. **Test both fresh conversion AND saved model.** Fresh catches conversion logic bugs; saved catches stale/corrupted weights from previous runs.
+6. **Diff configs before debugging.** Before investigating any divergence, dump and compare all config values. A 30-second config diff prevents hours of debugging based on wrong assumptions.
+7. **Never modify cached/downloaded model configs directly.** Don't edit files in `~/.cache/huggingface/`. Instead, save to a local directory or open a PR on the upstream repo.
+8. **Test decode before denoise.** Always verify the decoder works correctly before spending time on the denoising loop.
```
dg845: Do you think an analogous "test encode before denoise" rule would be useful here? Or is it already covered by the "Stage Test Order" section above?

.ai/skills/model-integration/SKILL.md
```diff
@@ -0,0 +1,167 @@
...
+
+    def forward(self, hidden_states, attention_mask=None, **kwargs):
+        return self.processor(self, hidden_states, attention_mask, **kwargs)
+```
+
+Consult the implementations in `src/diffusers/models/transformers/` if you need further references.
+
+### Implementation rules
+
+- **Don't combine structural changes with behavioral changes.** Restructuring code to fit diffusers APIs (ModelMixin, ConfigMixin, etc.) is unavoidable. But don't also "improve" the algorithm, refactor computation order, or rename internal variables for aesthetics. Keep numerical logic as close to the reference as possible, even if it looks unclean. For standard → modular, this is stricter: copy loop logic verbatim and only restructure into blocks. Clean up in a separate commit after parity is confirmed.
```
dg845: nit: maybe we could number the implementation rules here similar to the "Testing Rules" in `parity-testing/SKILL.md` so that it's easier to refer back to them if needed?

.ai/skills/parity-testing/SKILL.md
yiyixuxu: yeah already covered,  will remove this point too

PR #13270, opened 2026-03-16
fix parallelism case failure in xpu

sywangyi:
fix 
tests/models/transformers/test_models_transformer_flux.py::TestFluxTransformerContextParallel::test_context_parallel_inference[ulysses]
tests/models/transformers/test_models_transformer_flux.py::TestFluxTransformerContextParallel::test_context_parallel_custom_mesh[ulysses-3d-fsdp]
this case in xpu

@sayakpaul 
thanks


tests/models/testing_utils/parallelism.py
```diff
@@ -26,6 +26,7 @@
 from ...testing_utils import (
     is_context_parallel,
     require_torch_multi_accelerator,
+    torch_device,
```
sayakpaul, line 29: Instead, let's device map of the backends and the devices and use it? This way, we won't have to do if-else blocks.

sayakpaul (approved):

Issue #13277, opened 2026-03-17
QwenImagePipeline failed with Ulysses SP and batch inputs


zhtmike:
### Describe the bug

QwenImagePipeline cannot run with Ulysses SP together with batch prompt inputs. It is related to the mask is not correctly broadcasted.

### Reproduction

With the following code snippets
```python
import torch
import torch.distributed as dist
import argparse
import os
from diffusers import QwenImagePipeline
from diffusers import ContextParallelConfig


def parse_args():
    parser = argparse.ArgumentParser(
        description="Test Qwen-Image with Context Parallelism")
    return parser.parse_args()


args = parse_args()

if dist.is_available():
    dist.init_process_group(backend="nccl")
    rank = dist.get_rank()
    device = torch.device("cuda", rank % torch.cuda.device_count())
    world_size = dist.get_world_size()
    torch.cuda.set_device(device)
else:
    rank = 0
    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
    world_size = 1

model_id = "Qwen/Qwen-Image"

pipe = QwenImagePipeline.from_pretrained(
    model_id,
    torch_dtype=torch.bfloat16,
)
pipe.to(device)

pipe.transformer.set_attention_backend("native")
if world_size > 1:
    from diffusers import QwenImageTransformer2DModel
    assert isinstance(pipe.transformer, QwenImageTransformer2DModel)
    pipe.transformer.enable_parallelism(config=ContextParallelConfig(
        ulysses_degree=world_size))

pipe.set_progress_bar_config(disable=rank != 0)

positive_magic = {
    "en": ", Ultra HD, 4K, cinematic composition.",  # for english prompt
    "zh": ", 超清，4K，电影级构图.",  # for chinese prompt
}
prompts = [
    "A coffee shop entrance features a chalkboard sign reading "
    '"Qwen Coffee 😊 $2 per cup," with a neon light beside it '
    'displaying "通义千问". Next to it hangs a poster showing a '
    "beautiful Chinese woman, and beneath the poster is written "
    '"π≈3.1415926-53589793-23846264-33832795-02384197". '
    "Ultra HD, 4K, cinematic composition",
    "A cute cat with long hair sitting on a sofa, Ultra HD, 4K, cinematic composition."
]

inputs = {
    "prompt": [p + positive_magic["en"] for p in prompts],
    "generator": torch.Generator(device="cpu").manual_seed(0),
    "true_cfg_scale": 4.0,
[...]

PR #13182, opened 2026-02-24
[attention backends] fix ring CP for flash and flash 3

sayakpaul:
# What does this PR do?

* Fixes support for ring attention for Flash and Flash 3
* Adds a test suite for CP backends x attention backends

image
(CP x FA3)

image
(CP x FA2)

PR #12563, opened 2025-10-30
ulysses enabling in native attention path

sywangyi:
fix the corrupted output when enable ulysses with native attention. since native attention is widely use in no-cuda plaform. and ulysses does not rely on lse. so ulysses attention still could be used in native attention path.
- General functionalities: @sayakpaul @yiyixuxu @DN6

PR #13278, opened 2026-03-17
Fix the attention mask in ulysses SP for QwenImage

zhtmike:
# What does this PR do?

Fix issue #13277. 

QwenImagePipeline cannot run with Ulysses SP together with batch prompt inputs. It is related to the mask is not correctly broadcasted.
We need to broadcast the attention mask from [B, S] to [B, H, S_q, S_kv] or simply [B, 1, 1, S_kv] before feeding into SDPA.

Before Fix:

We have the error when running the code snippet mentioned in the issue.
```text
RuntimeError: The expanded size of the tensor (4222) must match the existing size (2) at non-singleton dimension 2.  Target sizes: [2, 12, 4222, 4222].  Tensor sizes: [2, 4222]
```

After Fix:

The images are correctly produced.
output_image_ulysses2_0
output_image_ulysses2_1

Fixes # (issue)

## Before submitting

## Who can review?

@sayakpaul 

### DIFF
--- a/src/diffusers/utils/testing_utils.py
+++ b/src/diffusers/utils/testing_utils.py
@@ -29,6 +29,7 @@ from numpy.linalg import norm
 from packaging import version
 
 from .constants import DIFFUSERS_REQUEST_TIMEOUT
+from .deprecation_utils import deprecate
 from .import_utils import (
     BACKENDS_MAPPING,
     is_accelerate_available,
@@ -67,9 +68,11 @@ else:
 global_rng = random.Random()
 
 logger = get_logger(__name__)
-logger.warning(
-    "diffusers.utils.testing_utils' is deprecated and will be removed in a future version. "
-    "Determinism and device backend utilities have been moved to `diffusers.utils.torch_utils`. "
+deprecate(
+    "diffusers.utils.testing_utils",
+    "1.0.0",
+    "diffusers.utils.testing_utils is deprecated and will be removed in a future version. "
+    "Determinism and device backend utilities have been moved to `diffusers.utils.torch_utils`. ",
 )
 _required_peft_version = is_peft_available() and version.parse(
     version.parse(importlib.metadata.version("peft")).base_version
--- a/tests/others/test_utils.py
+++ b/tests/others/test_utils.py
@@ -13,8 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import importlib
 import os
 import unittest
+import warnings
 
 import pytest
 
@@ -181,6 +183,25 @@ class DeprecateTester(unittest.TestCase):
             deprecate(("deprecated_arg", self.higher_version, "This message is better!!!"), standard_warn=False)
         assert str(warning.warning) == "This message is better!!!"
         assert "diffusers/tests/others/test_utils.py" in warning.filename
+
+    def test_deprecate_testing_utils_module(self):
+        import diffusers.utils.testing_utils
+
+        with warnings.catch_warnings(record=True) as caught_warnings:
+            warnings.simplefilter("always")
+            importlib.reload(diffusers.utils.testing_utils)
+
+        deprecation_warnings = [w for w in caught_warnings if issubclass(w.category, FutureWarning)]
+        assert len(deprecation_warnings) >= 1, "Expected at least one FutureWarning from diffusers.utils.testing_utils"
+
+        messages = [str(w.message) for w in deprecation_warnings]
+        assert any("diffusers.utils.testing_utils" in msg for msg in messages), (
+            f"Expected a deprecation warning mentioning 'diffusers.utils.testing_utils', got: {messages}"
+        )
+        assert any(
+            "diffusers.utils.testing_utils is deprecated and will be removed in a future version." in msg
+            for msg in messages
+        ), f"Expected deprecation message substring not found, got: {messages}"
 
 
 # Copied from https://github.com/huggingface/transformers/blob/main/tests/utils/test_expectations.py


src/diffusers/models/attention_dispatch.py
```diff
@@ -813,6 +813,9 @@ def _native_attention_forward_op(
     if return_lse:
         raise ValueError("Native attention does not support return_lse=True")
 
+    if attn_mask is not None and attn_mask.dim() == 2:
+        attn_mask = attn_mask[:, None, None, :]
```
sayakpaul: Is this Qwen specific?

zhtmike: I haven't test with other models. But I think models with a 2D masks input should have the similar problem

sayakpaul: Possible to check out one other? And also run the 
https://github.com/huggingface/diffusers/blob/16e7067647673ffe7e8d75f34c420c4f03dd542f/tests/models/transformers/test_models_transformer_flux.py#L227 ?

zhtmike: Thanks. From a quick scan, most models seem to handle the mask shape correctly in their own implementations. So I’ve limited the modification to QwenImage only.

Should I run any test cases?

sayakpaul: Thanks! Maybe we could add a similar test to https://github.com/huggingface/diffusers/blob/16e7067647673ffe7e8d75f34c420c4f03dd542f/tests/models/transformers/test_models_transformer_flux.py#L227?

I will give you a ping once it's refactored to follow the latest pattern.

sayakpaul:
@naykun if you want to take a look

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -934,6 +934,7 @@ def forward(
             batch_size, image_seq_len = hidden_states.shape[:2]
             image_mask = torch.ones((batch_size, image_seq_len), dtype=torch.bool, device=hidden_states.device)
             joint_attention_mask = torch.cat([encoder_hidden_states_mask, image_mask], dim=1)
+            joint_attention_mask = joint_attention_mask[:, None, None, :]
```
sayakpaul, line 936: Is this okay for non-CP?

zhtmike: Yes. The image is same w/o this change.

src/diffusers/models/attention_dispatch.py
sayakpaul: https://github.com/huggingface/diffusers/pull/13069 is merged. Let's test it.

zhtmike: Thanks. I have tried UTs and get error even in the main branch. 
```bash
pytest -x tests/models/transformers/test_models_transformer_qwenimage.py
```
and get
```text
_ TestQwenImageTransformerContextParallel.test_context_parallel_inference[ring] _

self = <tests.models.transformers.test_models_transformer_qwenimage.TestQwenImageTransformerContextParallel object at 0x1550bafa3560>
cp_type = 'ring_degree'

    @pytest.mark.parametrize("cp_type", ["ulysses_degree", "ring_degree"], ids=["ulysses", "ring"])
    def test_context_parallel_inference(self, cp_type):
        if not torch.distributed.is_available():
            pytest.skip("torch.distributed is not available.")
    
        if not hasattr(self.model_class, "_cp_plan") or self.model_class._cp_plan is None:
    
        world_size = 2
        init_dict = self.get_init_dict()
        inputs_dict = self.get_dummy_inputs()
    
        # Move all tensors to CPU for multiprocessing
        inputs_dict = {k: v.cpu() if isinstance(v, torch.Tensor) else v for k, v in inputs_dict.items()}
        cp_dict = {cp_type: world_size}
    
        # Find a free port for distributed communication
        master_port = _find_free_port()
    
        # Use multiprocessing manager for cross-process communication
        manager = mp.Manager()
        return_dict = manager.dict()
    
        # Spawn worker processes
        mp.spawn(
            _context_parallel_worker,
            args=(world_size, master_port, self.model_class, init_dict, cp_dict, inputs_dict, return_dict),
            nprocs=world_size,
            join=True,
        )
    
>       assert return_dict.get("status") == "success", (
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            f"Context parallel inference failed: {return_dict.get('error', 'Unknown error')}"
        )
E       AssertionError: Context parallel inference failed: Native attention does not support return_lse=True
```

Seems there is some attention backend selection issue in UT?

sayakpaul: Could we try with "native_cudnn"? 

Cc: @DN6 should we modify this test?

sayakpaul: But I remember https://github.com/huggingface/diffusers/pull/12563 fixing native attention x Ulysses 👀 

Ring is known to not work with native IIRC.

zhtmike: > But I remember #12563 fixing native attention x Ulysses 👀

Yes, native attention x Ulysses is perfect for single prompt input. Currently batch inputs have some problem.

zhtmike: > Could we try with "native_cudnn"?

How to enable that?

sayakpaul: Sorry disregard my suggestion on using the CUDNN backend.

> Yes, native attention x Ulysses is perfect for single prompt input. Currently batch inputs have some problem.

Is it the case just for Qwen or the same happens for Flux, as well? Also, the test under consideration -- does it not use a single prompt?

https://github.com/huggingface/diffusers/blob/11a3284cee024bfff017408b303d18e88981d2ea/tests/models/transformers/test_models_transformer_qwenimage.py#L81

zhtmike: > Is it the case just for Qwen or the same happens for Flux, as well? 

So far, I have only found that Qwen has this problem. Other models, such as [Z-Image]( https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_z_image.py#L130), [HunyuanImage](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_hunyuanimage.py#L819)
expand the attention mask in a similar way before entering the attention block. For Flux, I tested with the main branch, and it works fine with both CP and batch inputs.

> Also, the test under consideration -- does it not use a single prompt?

I am wondering whether we should add a batch input test if possible. At the beginning, I think we should first ensure that all unit tests pass without modifying them.

The background of this bug is that we are working on the training engine based on the Diffusers backend, using QwenImage as the first example. Therefore, we may need a combination of batch inputs (for high throughput) as well as Ulysses SP support. This is why we encountered this bug during the forward process.

sayakpaul: Agreed and thanks so much for the context!

> I am wondering whether we should add a batch input test if possible. At the beginning, I think we should first ensure that all unit tests pass without modifying them.

Would you like to take a crack at this? We'll be quick to review.

I think first we need to ensure that the `test_context_parallel_inference()` test is xfailed when ring attention is enabled with the SDPA. https://github.com/huggingface/diffusers/pull/13182 is adding a test suite for CP-backends and attention backends.

And then a test for batched inputs.

Then let's revisit this PR?

WDYT?

zhtmike: Sure NP, I will add a UT test for batch input

zhtmike: Hi @sayakpaul , I have added a PR #13312 for this, could you take a look?

zhtmike:
Following up on #13312: we can drop the `xfail` after fixing the QwenImage mask.

running `pytest tests/models/transformers/test_models_transformer_qwenimage.py`:
```text
=========================== short test summary info ============================
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerLoRAHotSwap::test_hotswapping_compiled_model_linear[11-11]
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerLoRAHotSwap::test_hotswapping_compiled_model_linear[7-13]
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerLoRAHotSwap::test_hotswapping_compiled_model_linear[13-7]
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerLoRAHotSwap::test_hotswapping_compiled_model_both_linear_and_other[11-11]
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerLoRAHotSwap::test_hotswapping_compiled_model_both_linear_and_other[7-13]
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerLoRAHotSwap::test_hotswapping_compiled_model_both_linear_and_other[13-7]
FAILED tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerLoRAHotSwap::test_enable_lora_hotswap_called_after_adapter_added_warning
====== 7 failed, 69 passed, 54 skipped, 36 warnings in 182.34s (0:03:02) =======

```
The error of LoRAHotSwap seems unrelated.

sayakpaul:
I ran the tests on my end and they are passing (CP tests). Hotswapping test failures are not needed.

dg845:
Merging as the CI is green.

PR #13282, opened 2026-03-18
[Helios] Remove lru_cache for better AoTI compatibility and cleaner code

SHYuanBest:
cc @sayakpaul
related to: https://gist.github.com/sayakpaul/aba9d8925a24944f183d67f1ef4e009d


sayakpaul:
As mentioned over Slack, let's maybe try to assess the impact of this change before merging?

PR #13285, opened 2026-03-19
Add low precision attention API from torchao to TorchAoConfig

howardzhang-cv:
# What does this PR do?
Adds low precision attention API from TorchAO to diffusers by updating TorchAoConfig with attn_backend option.
Note: this will require torchao 0.17.0

## Todo:
- [ ] Need to add new necessary tests
- [ ] Need to update the documentation: 
    - [documentation guidelines](https://github.com/huggingface/diffusers/tree/main/docs)
    - [here are tips on formatting docstrings](https://github.com/huggingface/diffusers/tree/main/docs#writing-source-documentation).

#### flux.1-dev with 2048x2048 image size

| Config | Median Time (s) | Speedup |
| --- | --- | --- |
| bf16 baseline + torch.compile | 3.17 | 1.00x |
| fp8_attn + compile | 2.66 | **1.07x** |

#### Wan2.1-14B-Diffusers with 1280x720 frame size and 81 frames

| Config | Median Time (s) | Speedup |
| --- | --- | --- |
| bf16 baseline + torch.compile | 81.32 | 1.00x |
| fp8_attn + compile | 46.88 | **1.18x** |

Note that these results use a naive scheme, where every layer is quantized. Doing so results in subpar results when doing many inference steps (50 for WAN2.1). Using a better scheme (such as skipping early and late layers, quantizing 36/40 total layers) results in **1.12x** speedup with much better quality. The VBench benchmarks for that are below:

image

### DIFF
--- a/src/diffusers/quantizers/quantization_config.py
+++ b/src/diffusers/quantizers/quantization_config.py
@@ -433,14 +433,19 @@ class TorchAoConfig(QuantizationConfigMixin):
     """This is a config class for torchao quantization/sparsity techniques.
 
     Args:
-        quant_type (`AOBaseConfig`):
+        quant_type (`AOBaseConfig` | None):
             An `AOBaseConfig` subclass instance specifying the quantization type. See the [torchao
             documentation](https://docs.pytorch.org/ao/main/api_ref_quantization.html#inference-apis-for-quantize) for
             available config classes (e.g. `Int4WeightOnlyConfig`, `Int8WeightOnlyConfig`, `Float8WeightOnlyConfig`,
             `Float8DynamicActivationFloat8WeightConfig`, etc.).
+            Pass `None` when only `attention_backend` is used without weight quantization.
         modules_to_not_convert (`list[str]`, *optional*, default to `None`):
             The list of modules to not quantize, useful for quantizing models that explicitly require to have some
             modules left in their original precision.
+        attention_backend (`str`, *optional*, default to `None`):
+            Low-precision attention backend to use. Currently supported: `"fp8_fa3"` (FP8 attention using Flash
+            Attention 3, requires Hopper GPU with SM90+). This is orthogonal to weight quantization — you can use
+            either or both. When used with `torch.compile`, RoPE fusion is automatically enabled.
 
     Example:
         ```python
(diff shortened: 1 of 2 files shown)


src/diffusers/quantizers/torchao/torchao_quantizer.py
```diff
@@ -188,8 +188,19 @@ def validate_environment(self, *args, **kwargs):
                         f"In order to use TorchAO pre-quantized model, you need to have torch>=2.5.0. However, the current version is {torch_version}."
                     )
 
+        if self.quantization_config.attention_backend is not None:
+            from torchao.prototype.attention import AttentionBackend
+
+            AttentionBackend(self.quantization_config.attention_backend.upper())
```
sayakpaul, line 194: What happens if users do something like `model.set_attention_backend("_flash_3_hub")`?

Docs: https://huggingface.co/docs/diffusers/main/en/optimization/attention_backends

src/diffusers/quantizers/torchao/torchao_quantizer.py
```diff
@@ -188,8 +188,19 @@ def validate_environment(self, *args, **kwargs):
...
+
     def update_torch_dtype(self, torch_dtype):
-        config_name = self.quantization_config.quant_type.__class__.__name__
```
sayakpaul, line 192: When can `quant_type` be None? Should it be validated?

src/diffusers/quantizers/torchao/torchao_quantizer.py
```diff
@@ -298,6 +315,9 @@ def get_cuda_warm_up_factor(self):
         - Use a division factor of 8 for int4 weights
         - Use a division factor of 4 for int8 weights
         """
+        if self.quantization_config.quant_type is None:
+            return 4
```
sayakpaul, line 319: Is this a safe choice?

src/diffusers/quantizers/torchao/torchao_quantizer.py
```diff
@@ -314,6 +334,13 @@ def _process_model_before_weight_loading(
...
         **kwargs,
     ):
+        model.config.quantization_config = self.quantization_config
+
+        if self.quantization_config.quant_type is None:
+            # Attention-only mode: no weight quantization setup needed
+            self.modules_to_not_convert = []
+            return
+
         self.modules_to_not_convert = self.quantization_config.modules_to_not_convert
```
sayakpaul, line 344: If `modules_to_not_convert` is specified and `quant_type` is None, should we throw a warning to the user?

src/diffusers/quantizers/torchao/torchao_quantizer.py
```diff
@@ -332,11 +359,73 @@ def _process_model_before_weight_loading(
...
         self.modules_to_not_convert = [module for module in self.modules_to_not_convert if module is not None]
 
-        model.config.quantization_config = self.quantization_config
-
     def _process_model_after_weight_loading(self, model: "ModelMixin"):
+        if self.quantization_config.attention_backend is not None:
+            from torchao.prototype.attention import AttentionBackend, apply_low_precision_attention
+
+            backend = AttentionBackend(self.quantization_config.attention_backend.upper())
+            modules_to_not_convert = self.quantization_config.modules_to_not_convert or []
```
sayakpaul, line 367: We shouldn't allow `self.quantization_config.modules_to_not_convert` when attention mode is specified no?

https://github.com/huggingface/diffusers/pull/13285/changes#diff-3db90d3c84b6ed43fde7eecaad399f024129a4bc7f6b872b25c6b0236299894dR339 gives me this impression.

src/diffusers/quantizers/quantization_config.py
```diff
@@ -433,14 +433,19 @@ class TorchAoConfig(QuantizationConfigMixin):
...
             available config classes (e.g. `Int4WeightOnlyConfig`, `Int8WeightOnlyConfig`, `Float8WeightOnlyConfig`,
             `Float8DynamicActivationFloat8WeightConfig`, etc.).
+            Pass `None` when only `attention_backend` is used without weight quantization.
         modules_to_not_convert (`list[str]`, *optional*, default to `None`):
             The list of modules to not quantize, useful for quantizing models that explicitly require to have some
             modules left in their original precision.
+        attention_backend (`str`, *optional*, default to `None`):
+            Low-precision attention backend to use. Currently supported: `"fp8_fa3"` (FP8 attention using Flash
+            Attention 3, requires Hopper GPU with SM90+). This is orthogonal to weight quantization — you can use
+            either or both. When used with `torch.compile`, RoPE fusion is automatically enabled.
```
sayakpaul, line 448: Why is this restricted to Hopper only? FA3 is available for Ampere as well, no?

Also, when `attention_backend` is specified, does it automatically use FA3? The details aren't very clear here.

sayakpaul (commented):
Thanks! I left some comments.

Additionally, it would be nice to:

* Have an example script with a modern image/video generation model that provides reasonable outputs.
* Update the tests in https://github.com/huggingface/diffusers/blob/48f39c2d59e8db444cb37f91e72413a1db9a2dd6/tests/models/testing_utils/quantization.py#L835
* Update the [docs](https://huggingface.co/docs/diffusers/main/en/quantization/torchao) (which you have already noted in the PR description)

PR #13289, opened 2026-03-19
[Neuron] Add AWS Neuron (Trainium/Inferentia) as an officially supported device

JingyaHuang:
# What does this PR do?

This PR adds AWS Neuron (Trainium/Inferentia) as an officially supported compute backend in Diffusers, on par with existing backends like CUDA, MPS, XPU, and MLU.

**Changes**

- import_utils.py — adds is_torch_neuronx_available() detection, following the existing pattern for optional backends.                                 
- torch_utils.py — registers "neuron" in all backend dispatch tables (BACKEND_SUPPORTS_TRAINING, BACKEND_EMPTY_CACHE, BACKEND_DEVICE_COUNT, BACKEND_MANUAL_SEED, etc.) and adds a randn_tensor workaround since Neuron/XLA does not support creating random tensors directly on device (falls back to CPU).        
- utils/__init__.py — exports is_torch_neuronx_available.                                                                                              
- pipeline_utils.py — adds two new DiffusionPipeline methods:                                                                                          
    - enable_neuron_compile(model_names, cache_dir, fullgraph) — wraps pipeline nn.Module components with torch.compile(backend="neuron") for whole-graph NEFF compilation. Supports optional NEFF caching via TORCH_NEURONX_NEFF_CACHE_DIR.                                                                    
    - neuron_warmup(*args, **kwargs) — runs a single dummy forward pass to trigger upfront neuronx-cc compilation before timed inference.                
                                                                                                                                                         
 **Usage**

* Eager mode  
 
```python                                                                                                                                                     
import torch                                                                                                             
import torch_neuronx  # noqa: F401 — registers torch.neuron                                                            
                                                                                                                           
from diffusers import AutoPipelineForText2Image                                                                          
                                                                                                                           
# Load and move to Neuron device                                                                                         
pipe = AutoPipelineForText2Image.from_pretrained(                                                                        
    "stabilityai/sdxl-turbo",                                                                                            
    torch_dtype=torch.bfloat16,                           
    variant="fp16",                                                                                                      
)
pipe = pipe.to(torch.neuron.current_device())                                                                            
                                                                                                                         
# Warmup                                                                   
pipe(prompt="warmup", height=512, width=512, num_inference_steps=1, guidance_scale=0.0)                                                                                                                        
                                                          
# Inference                                                                                               
image = pipe(                                             
    prompt="a golden retriever surfing a wave, photorealistic",                                                          
    height=512,
    width=512,                                                                                                           
    num_inference_steps=1, 
    guidance_scale=0.0,                                                                    
).images[0]                                                                                                              
                                                                                                                         
image.save("output.png") 
```

**Validation**

So far we validated the following models, the idea is to manually validated a part of representative models, and ensure the model coverage with pur CIs afterward.

* pixart
* sdxl
* flux2-klein-4B (able to run within one NeuronCore w/o. tp under eager mode for resolution 1024x1024)
                                                                                                      
                                                                                                                                                         
**Next Steps**                                                                                                                                                  
* Enable `torch.compile` on Neuron device
* Add tensor parallel support for memory-bound devices like neuron
* Tackle the compatibility of diffusers+nki kernels lib to boost the performance on neuron.
* Support sequence parallel on Neuron for video models.


tests/pipelines/stable_diffusion_xl/test_stable_diffusion_xl.py
```diff
@@ -974,3 +977,47 @@ def test_stable_diffusion_lcm(self):
         max_diff = numpy_cosine_similarity_distance(image.flatten(), expected_image.flatten())
 
         assert max_diff < 1e-2
+
+
+@slow
```
sayakpaul: I don't think we want to run this on our slow CLI. So if it's absolutely must, we should like decorate it with `require_torch_neuron`.

sayakpaul (approved):
Nice thank you!

sayakpaul:
@claude-2-serge could you do a review please?

JingyaHuang:
Thanks @sayakpaul, I addressed the review comments, shall I merge?

tests/pipelines/flux2/test_pipeline_flux2_klein.py
```diff
@@ -181,3 +188,56 @@ def test_image_input(self):
...
+            prompt=self.prompt,
+            height=512,
+            width=512,
+            num_inference_steps=4,
+            guidance_scale=1.0,
+            generator=generator,
+            output_type="np",
+        ).images
+
+        image_slice = image[0, -3:, -3:, -1]
```
sayakpaul, line 240: Would it make sense to compare against a value slice or are we not there yet?

tests/pipelines/stable_diffusion_xl/test_stable_diffusion_xl.py
```diff
@@ -974,3 +978,50 @@ def test_stable_diffusion_lcm(self):
...
+
+        image = pipe(
+            self.prompt,
+            num_inference_steps=1,
+            guidance_scale=0.0,
+            generator=generator,
+            output_type="np",
+        ).images
+
+        image_slice = image[0, -3:, -3:, -1]
```
sayakpaul, line 1024: Same for this.

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
```diff
@@ -862,7 +862,8 @@ def __call__(
             prompt_attention_mask = torch.cat([negative_prompt_attention_mask, prompt_attention_mask], dim=0)
 
         # 4. Prepare timesteps
-        if XLA_AVAILABLE:
+        is_neuron_device = hasattr(device, "type") and device.type == "neuron"
```
yiyixuxu:
```suggestion
        is_neuron_device = device.type == "neuron"
```
I think it should always a `torch.device` here, no?  

@claude-2-serge can you trace the code for how `device` is defined here and let me know if there is any code path that `_exeuction_devide` would not return ` torch.device` type?

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
```diff
@@ -862,7 +862,8 @@ def __call__(
...
+        if XLA_AVAILABLE or is_neuron_device:
```
yiyixuxu, line 866: ohh, would it possible for`is_neuron_device==True` but  xla not available?

src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
```diff
@@ -1195,15 +1196,23 @@ def __call__(
                 # expand the latents if we are doing classifier free guidance
                 latent_model_input = torch.cat([latents] * 2) if self.do_classifier_free_guidance else latents
 
-                latent_model_input = self.scheduler.scale_model_input(latent_model_input, t)
+                # For Neuron: scale_model_input on CPU to avoid XLA ops outside the compiled UNet region.
+                # index_for_timestep() uses .nonzero()/.item() which are incompatible with static graphs.
```
yiyixuxu: ohh, if `index_for_timtestep` is the issue here, maybe we can add `set_begin_index(0)` before the denoising loop, like we did in qwen https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py#L685 

would this solve the issue?

yiyixuxu (commented):
thanks! i left some questions/comments

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
yiyixuxu: @claude-2-serge can you trace the code for how device is defined here and let me know if there is any code path that _exeuction_devide would not return  torch.device type?

sayakpaul: Follow-up inline comments are something we're working with @tarekziade. Should land pretty soon.

JingyaHuang: True!

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
JingyaHuang: TorchNeuron will be torch_xla free, so we won't be able to rely on `XLA_AVAILABLE` for neuron.

tests/pipelines/flux2/test_pipeline_flux2_klein.py
JingyaHuang: My concern is that the expected slice has been changing across Neuron SDK versions. I initially preferred a weaker check, but we could add the comparison to see if TorchNeuron is stable enough on this.

src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
JingyaHuang: yeah, that's a better fix! Thanks Yiyi.

JingyaHuang (commented):
addressed the comments! @sayakpaul @yiyixuxu

JingyaHuang:
The failing CI checks are unrelated to this PR, ok for us to merge? @sayakpaul @yiyixuxu

src/diffusers/models/transformers/transformer_flux2.py
```diff
@@ -961,7 +961,8 @@ def forward(self, ids: torch.Tensor) -> torch.Tensor:
         pos = ids.float()
         is_mps = ids.device.type == "mps"
         is_npu = ids.device.type == "npu"
-        freqs_dtype = torch.float32 if (is_mps or is_npu) else torch.float64
```
DN6, line 964: Since this cast seems to be the same across multiple device types, I would say let's introduce a utility function to just handle these casts in a centralized way

e.g. in torch_utils.py
```python
  # torch_utils.py
  _FP64_UNSUPPORTED_DEVICES = frozenset({"mps", "npu", "neuron"})
  _FP64_DOWNCAST = {torch.float64: torch.float32, torch.int64: torch.int32}

  def maybe_adjust_dtype_for_device(dtype: torch.dtype, device: torch.device) -> torch.dtype:
      if device.type in _FP64_UNSUPPORTED_DEVICES:
          return _FP64_DOWNCAST.get(dtype, dtype)
      return dtype

```

JingyaHuang: @DN6 sounds good, added `maybe_adjust_dtype_for_device` and applied to related models.

yiyixuxu:
@askserge can you do a review here? in addition to review-rules, also these 3 commetns from my end
1. for `maybe_adjust_dtype_for_device`, check if backward-compatible everywhere
2. suggest how we update agent doc for maybe_adjust_dtype_for_device 
3. check the sdxl pipeline, of set_begin_index() can be hardcoded at 0 for all scenarios? for example `denoising_end`

yiyixuxu:
@claude can you do a review here? in addition to review-rules, also these 3 commetns from my end
1. for `maybe_adjust_dtype_for_device`, check if backward-compatible everywhere
2. suggest how we update agent doc for maybe_adjust_dtype_for_device 
3. check the sdxl pipeline, of set_begin_index() can be hardcoded at 0 for all scenarios? for example `denoising_end`, please answer this by traceing the code

src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
```diff
@@ -1095,7 +1095,8 @@ def __call__(
         )
 
         # 4. Prepare timesteps
-        if XLA_AVAILABLE:
+        is_neuron_device = hasattr(device, "type") and device.type == "neuron"
```
yiyixuxu:
```suggestion
        is_neuron_device = device.type == "neuron"
```
think it's ok, no?

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
```diff
@@ -914,10 +915,11 @@ def __call__(
                     # This would be a good case for the `match` statement (Python 3.10+)
                     is_mps = latent_model_input.device.type == "mps"
                     is_npu = latent_model_input.device.type == "npu"
+                    is_neuron = latent_model_input.device.type == "neuron"
                     if isinstance(current_timestep, float):
-                        dtype = torch.float32 if (is_mps or is_npu) else torch.float64
+                        dtype = torch.float32 if (is_mps or is_npu or is_neuron) else torch.float64
```
yiyixuxu: why not use the maybe_adjust_dtype_for_device here too?

yiyixuxu (approved):
thanks so much for working on this!
Can you also update our agent.md on the new helper function you added here? https://github.com/huggingface/diffusers/blob/main/.ai/models.md (Claude CI review #2 https://github.com/huggingface/diffusers/pull/13289#issuecomment-4597543804)

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
JingyaHuang: good catch! applied `maybe_adjust_dtype_for_device` here as well.

JingyaHuang:
Thanks @yiyixuxu for the review, just applied the suggestions. Shall we merge?

PR #13290, opened 2026-03-19
[export] Add export-safe LRU cache helper

cbensimon:


src/diffusers/utils/torch_utils.py
```diff
@@ -333,5 +338,21 @@ def disable_full_determinism():
...
 
 
+@functools.wraps(functools.lru_cache)
+def lru_cache_unless_export(maxsize=128, typed=False):
+    def outer_wrapper(fn: Callable[P, T]):
+        cached = functools.lru_cache(maxsize=maxsize, typed=typed)(fn)
+
+        @functools.wraps(fn)
+        def inner_wrapper(*args: P.args, **kwargs: P.kwargs):
+            if torch.compiler.is_exporting():
```
sayakpaul, line 350: Should we check if `is_exporting` exists with `hastattr`?

src/diffusers/utils/torch_utils.py
```diff
@@ -333,5 +338,21 @@ def disable_full_determinism():
     torch.use_deterministic_algorithms(False)
 
 
+@functools.wraps(functools.lru_cache)
+def lru_cache_unless_export(maxsize=128, typed=False):
+    def outer_wrapper(fn: Callable[P, T]):
+        cached = functools.lru_cache(maxsize=maxsize, typed=typed)(fn)
```
sayakpaul, line 344: Why do we have to specify `typed`?

src/diffusers/utils/torch_utils.py
cbensimon: Yes you're right. What's the minimum torch version officially supported by diffusers?

sayakpaul: 2.3

cbensimon: Ok. I did a version check (https://github.com/huggingface/diffusers/pull/13290/changes/5131131220fe11b3ee7486a014b27473a885d10e) since `is_exporting` is available since `2.7.0`. Let me know what you think

src/diffusers/utils/torch_utils.py
cbensimon: Just replicating `functools.lru_cache` in an un-opiniated fashion. I can remove (or replace with `**kwargs`), no string opinion on this

PR #13294, opened 2026-03-20
[ci] hoping to fix is_flaky with wanvace.

sayakpaul:
# What does this PR do?

The WanVACE LoRA test failures are the flakiest in nature. They have been making our CI red for a while now. This is another attempt getting rid of that:
https://github.com/huggingface/diffusers/actions/runs/23308255022/job/67788477319?pr=13290

### DIFF
--- a/tests/lora/test_lora_layers_wanvace.py
+++ b/tests/lora/test_lora_layers_wanvace.py
@@ -28,7 +28,6 @@ from diffusers.utils.import_utils import is_peft_available
 
 from ..testing_utils import (
     floats_tensor,
-    is_flaky,
     require_peft_backend,
     require_peft_version_greater,
     skip_mps,
@@ -46,7 +45,6 @@ from .utils import PeftLoraLoaderMixinTests  # noqa: E402
 
 @require_peft_backend
 @skip_mps
-@is_flaky(max_attempts=10, description="very flaky class")
 class WanVACELoRATests(unittest.TestCase, PeftLoraLoaderMixinTests):
     pipeline_class = WanVACEPipeline
     scheduler_cls = FlowMatchEulerDiscreteScheduler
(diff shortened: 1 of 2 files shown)


sayakpaul:
Failing tests are unrelated.

src/diffusers/utils/testing_utils.py
```diff
@@ -1041,10 +1041,12 @@ def is_flaky(max_attempts: int = 5, wait_before_retry: float | None = None, desc
     def decorator(obj):
         # If decorating a class, wrap each test method on it
         if inspect.isclass(obj):
-            for attr_name, attr_value in list(obj.__dict__.items()):
-                if callable(attr_value) and attr_name.startswith("test"):
-                    # recursively decorate the method
-                    setattr(obj, attr_name, decorator(attr_value))
+            # Use dir() instead of __dict__ to include inherited test methods from
```
dg845: I think this change should also be added to `is_flaky` in `tests/testing_utils.py`, which is the `is_flaky` method which is actually imported by `WanVACELoRATests`.

dg845 (commented):
Thanks! Is there a way we can verify that the tests inherited from `PeftLoraLoaderMixinTests` are indeed not being retried? I looked at a few CI failures involving `WanVACELoRATests` and the methods which fail do seem to all be inherited from `PeftLoraLoaderMixinTests` (for example, https://github.com/huggingface/diffusers/actions/runs/23179604200/job/67349602447?pr=13276) but I'd like to rule out the possibility that the tests are being retried and are failing for a different reason.

sayakpaul:
> I looked at a few CI failures involving WanVACELoRATests and the methods which fail do seem to all be inherited from PeftLoraLoaderMixinTests (for example, https://github.com/huggingface/diffusers/actions/runs/23179604200/job/67349602447?pr=13276) but I'd like to rule out the possibility that the tests are being retried and are failing for a different reason.

You can run the Wanvace test suite locally and it will pass everytime. I cannot think of another setup to verify.

src/diffusers/utils/testing_utils.py
sayakpaul: Good catch! Fixed.

PR #13297, opened 2026-03-20
[ci] claude in ci.

sayakpaul:
# What does this PR do?

Use Claude for PR reviews. Some choices / comments:
                                                                                                                                       
- A GitHub Actions workflow (using `anthropics/claude-code-action`) that auto-reviews PRs touching `src/diffusers/**` and responds to `@claude` mentions from folks with write access to the repo.                   
- An initial `.ai/review-rules.md` file distilled from the existing `AGENTS.md`, model-integration skill, and parity-testing pitfalls. It currently covers: covering code style, dependencies, models, pipelines, and copied code.                                                                    
- A "compounding engineering" workflow where maintainers add new review rules directly from PR comments (e.g., `@claude` add to `review-rules.md` to never use `einops`) — Claude commits the rule update to the PR branch, so rules accumulate over time with zero friction. This is inspired by https://x.com/bcherny/status/2007179842928947333.
- Access control: only users with write access can invoke `@claude`.                                                                 
- Cost control: path-filtered to `src/diffusers/**` only.
                                                                                                                                                                                                                                                                         
## How to use?                                                                          
                                                                                                                                       
- Claude only runs when explicitly invoked via review any PR that changes files under `src/diffusers/**`. 
- Ask Claude a question: Comment `@claude` <your question> on any PR to ask Claude about the code, the diff, or the review rules.          
- Add a new rule: When you spot a recurring mistake, comment something like:    
`@claude add to .ai/review-rules.md to never use numpy operations in forward methods, always use torch equivalents`. Claude will update the rules file and commit it to the PR branch.

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -671,6 +671,10 @@
         title: Z-Image
       title: Image
     - sections:
+      - local: api/pipelines/llada2
+        title: LLaDA2
+      title: Text
+    - sections:
       - local: api/pipelines/allegro
         title: Allegro
       - local: api/pipelines/chronoedit
@@ -718,6 +722,8 @@
   - sections:
     - local: api/schedulers/overview
       title: Overview
+    - local: api/schedulers/block_refinement
+      title: BlockRefinementScheduler
     - local: api/schedulers/cm_stochastic_iterative
       title: CMStochasticIterativeScheduler
     - local: api/schedulers/ddim_cogvideox
--- a/docs/source/en/api/pipelines/llada2.md
+++ b/docs/source/en/api/pipelines/llada2.md
@@ -1,0 +1,90 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+-->
+
+# LLaDA2
+
+[LLaDA2](https://huggingface.co/collections/inclusionAI/llada21) is a family of discrete diffusion language models
+that generate text through block-wise iterative refinement. Instead of autoregressive token-by-token generation,
+LLaDA2 starts with a fully masked sequence and progressively unmasks tokens by confidence over multiple refinement
+steps.
+
+## Usage
+
+```py
+import torch
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from diffusers import BlockRefinementScheduler, LLaDA2Pipeline
+
+model_id = "inclusionAI/LLaDA2.1-mini"
+model = AutoModelForCausalLM.from_pretrained(
+    model_id, trust_remote_code=True, dtype=torch.bfloat16, device_map="auto"
+)
+tokenizer = AutoTokenizer.from_pretrained(model_id, trust_remote_code=True)
+scheduler = BlockRefinementScheduler()
+
+pipe = LLaDA2Pipeline(model=model, scheduler=scheduler, tokenizer=tokenizer)
+output = pipe(
+    prompt="Write a short poem about the ocean.",
+    gen_length=256,
+    block_length=32,
+    num_inference_steps=32,
+    threshold=0.7,
+    editing_threshold=0.5,
+    max_post_steps=16,
+    temperature=0.0,
+)
+print(output.texts[0])
+```
+
+## Callbacks
+
+Callbacks run after each refinement step. Pass `callback_on_step_end_tensor_inputs` to select which tensors are
+included in `callback_kwargs`. In the current implementation, `block_x` (the sequence window being refined) and
+`transfer_index` (mask-filling commit mask) are provided; return `{"block_x": ...}` from the callback to replace the
+window.
+
+```py
+def on_step_end(pipe, step, timestep, callback_kwargs):
+    block_x = callback_kwargs["block_x"]
+    # Inspect or modify `block_x` here.
+    return {"block_x": block_x}
+
+out = pipe(
+    prompt="Write a short poem.",
+    callback_on_step_end=on_step_end,
+    callback_on_step_end_tensor_inputs=["block_x"],
+)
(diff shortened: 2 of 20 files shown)


.ai/review-rules.md
```diff
@@ -0,0 +1,34 @@
+# PR Review Rules
```
yiyixuxu, line 1: there are a lot of overlaps from https://github.com/huggingface/diffusers/blob/main/.ai/skills/model-integration/SKILL.md

I think maybe we have a separate subdoc for `models.md` that contains the gotchas, common conventions, rules, structures etc and we can link it from everywhere: for now both `AGENTS.md` and the model-integration skills

what do you think? cc @stevhliu here too 

these contents that are already in `AGENTS.md` does not need to be included here again, I think claude CI would have read the AGENTS.md. anyways no?

.github/workflows/claude_review.yml
```diff
@@ -0,0 +1,27 @@
+name: Claude PR Review
+
+on:
+  issue_comment:
+    types: [created]
```
yiyixuxu, line 5: can we also be able to @claude on the pr review comment? (not just issue comments) this gives it more context for stuff like we want it to add a rule based on this pattern eetc

.github/workflows/claude_review.yml
```diff
@@ -0,0 +1,27 @@
...
+      (github.event.comment.author_association == 'MEMBER' ||
+       github.event.comment.author_association == 'OWNER' ||
+       github.event.comment.author_association == 'COLLABORATOR')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          claude_args: |
+            --append-system-prompt "Review this PR against the rules in .ai/review-rules.md. Focus on correctness, not style (ruff handles style). Only review changes under src/diffusers/."
```
yiyixuxu: can we add something like 
only commit changes when explicitly asked using words/phrase blah blah - then maybe we can come up with a safe word for it lol

.ai/review-rules.md
stevhliu: yeah i agree here, and i think it'd be cleaner and easier to maintain to provide a path to content that is already in `AGENTS.md`

.ai/review-rules.md
```diff
@@ -0,0 +1,34 @@
...
+
+Rules for Claude to check during PR reviews. Focus on correctness — style is handled by ruff.
+
+## Code style
+- Inline logic — minimize small helper/utility functions. A reader should follow the full flow without jumping between functions.
+- No defensive code or unused code paths — no fallback paths, safety checks, or config options "just in case".
```
dg845: I wonder if this statement might be too strong. For example, it seems like pipeline `check_input` methods break the "no safety checks" rule if we interpret it literally or strictly. Maybe it would be better to break this into separate items for "fallback paths", "safety checks", "config options", etc. with specific guidance for each?

.ai/review-rules.md
```diff
@@ -0,0 +1,34 @@
+# PR Review Rules
+
+Rules for Claude to check during PR reviews. Focus on correctness — style is handled by ruff.
+
+## Code style
+- Inline logic — minimize small helper/utility functions. A reader should follow the full flow without jumping between functions.
```
dg845: I think numbering the items in each section would be useful if we want to refer to them elsewhere (e.g. "See Code Style (1) about inlining logic").

.ai/review-rules.md
sayakpaul: Good idea!

.ai/review-rules.md
sayakpaul: Elaborate more?

sayakpaul:
Thanks for the comments, all. I will work on them and push the updates.

.ai/review-rules.md
dg845: I'm not sure what the actual behavior of Claude will be, but it seems possible to me that it could interpret "no safety checks" to include `check_input` pipeline methods, which are generally standard. (Alternatively, maybe I'm misunderstanding what is referred to as a "safety check".)

So my proposal would be a separate item for "safety checks" with more concrete guidance. A rough example could be something like

```
- Safety checks - for pipelines, all inputs to `__call__` should be validated in the `check_inputs` method. Refer to <good example> as a reference. Assert statements should not be used to check expressions "just in case"; unsupported cases should be caught early and raise a concise exception. See also Code Style #X on fallback paths.
```

We could also do something similar for the other things discussed in the point above such as "fallback paths" (the below example is based on the similar bullet point in [`AGENTS.md`](https://github.com/huggingface/diffusers/blob/main/.ai/AGENTS.md)):

```
- Fallback paths - expected inputs should be made clear in the docstring for each function. When unexpected inputs are supplied, a concise error should be raised rather than implementing complex fallback logic. The code should not try to guess the user's intent and silently correct user behavior.
```

.ai/review-rules.md
sayakpaul: Pushed in 156b0b90b.

.ai/review-rules.md
sayakpaul: I have now explicitly asked it to review the rules we already have under skills and apply them during reviewing. I guess that should work?

.ai/review-rules.md
sayakpaul: Same as above.

sayakpaul:
@yiyixuxu @dg845 in the latest changes, I have made it so:

* `.ai/review-rules.md` instructs Claude to always review the existing rules under `AGENT.md` and the skills. This will help us mitigate overlap.
* We can invoke `@claude` in the PR review comments in between threads as well (refer to https://github.com/huggingface/diffusers/pull/13297#discussion_r2967906805).

LMK if you have any thoughts.

.ai/review-rules.md
```diff
@@ -0,0 +1,11 @@
+# PR Review Rules
+
+Review-specific rules for Claude. Focus on correctness — style is handled by ruff.
+
+Before reviewing, read and apply the guidelines in:
+- [AGENTS.md](AGENTS.md) — coding style, dependencies, copied code, model conventions
```
dg845, line 6: Should we refactor model information into a `models.md` file following https://github.com/huggingface/diffusers/pull/13297#discussion_r2967887307 in this PR?

yiyixuxu: yeah would be better but ok to do it in a seperate PR
cc @stevhliu in case you're interested

yiyixuxu:
cc @DN6 in case if you want to review it
otherwise PR looks good to me, we will iterate as we go

PR #13309, opened 2026-03-22
Add warning when using CUDA random generator (#13298)

### LINKED ISSUE #13298: When generating images, if the generator device is on cuda, it break things
### Describe the bug

This bug took me a while to relize, but when generating images with Flux Schnell, if the random generator device is on cuda, it create weird, blurry, and noisy images. 

### Reproduction

```
"""
Reproduce: same seed, same model, same prompt — cpu vs cuda generator produces different images.
"""
import os
os.environ["CUDA_VISIBLE_DEVICES"] = "1"

import sys
sys.path.insert(0, "Normalized-Attention-Guidance")
import torch

SEED = 134
PROMPT = "A vintage red bicycle leaning against a brick wall, its pedals leading back to a solitary rear gear."

print("Loading FluxPipeline...")
    "black-forest-labs/FLUX.1-schnell", torch_dtype=torch.bfloat16
).to("cuda")

kwargs = dict(
    guidance_scale=0.0,
    num_inference_steps=4,
    max_sequence_length=256,
)

img_cpu = pipe(PROMPT, generator=torch.Generator("cpu").manual_seed(SEED), **kwargs).images[0]
img_cpu.save("gen_cpu.png")
print("Saved gen_cpu.png")

img_cuda = pipe(PROMPT, generator=torch.Generator("cuda").manual_seed(SEED), **kwargs).images[0]
img_cuda.save("gen_cuda.png")
print("Saved gen_cuda.png")

# Check if they're identical
import numpy as np
arr_cpu = np.array(img_cpu)
arr_cuda = np.array(img_cuda)
print(f"Images identical: {np.array_equal(arr_cpu, arr_cuda)}")
print(f"Max pixel diff: {np.abs(arr_cpu.astype(int) - arr_cuda.astype(int)).max()}")

```
On CPU:
Image
On CUDA:

Image

Very interesting and cool bug but took me a while to find out as I was thinking it was my code broke it. And interestingly, only specific seed and prompt triggers it. I understand this could be expected, but I think it will be good to throw an warning 

### Logs

```shell

```

### System Info

- 🤗 Diffusers version: 0.36.0
- Platform: Linux-6.8.0-94-generic-x86_64-with-glibc2.35
- Python version: 3.11.14
- PyTorch version (GPU?): 2.8.0+cu128 (True)
- Jax version: not installed
- Huggingface_hub version: 0.36.2
- Transformers version: 4.57.0
- Accelerate version: 1.12.0
- PEFT version: 0.18.1
- Bitsandbytes version: 0.49.2
- Safetensors version: 0.7.0
- Accelerator: NVIDIA RTX A6000, 49140 MiB
NVIDIA RTX A6000, 49140 MiB
NVIDIA RTX A6000, 49140 MiB

### Who can help?

@DN6 @yiyixuxu @sayakpaul

r266-tech:
## What does this PR do?

Fixes #13298

When users pass a CUDA `torch.Generator` to diffusers pipelines, they may expect the same results as with a CPU generator given the same seed. However, PyTorch uses different random number generation algorithms on CPU and CUDA, so the generated images will differ.

This PR adds a warning in `randn_tensor` when a CUDA generator is used, informing users that:
- CUDA generators produce different random numbers than CPU generators with the same seed
- This is expected PyTorch behavior
- Use a CPU generator if reproducibility across devices is needed

## Before submitting

- [x] Did you read the [philosophy doc](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md)?

## Who can review?

@yiyixuxu @DN6 @sayakpaul

### DIFF
--- a/src/diffusers/utils/torch_utils.py
+++ b/src/diffusers/utils/torch_utils.py
@@ -176,6 +176,12 @@ def randn_tensor(
                 )
         elif gen_device_type != device.type and gen_device_type == "cuda":
             raise ValueError(f"Cannot generate a {device} tensor from a generator of type {gen_device_type}.")
+        elif gen_device_type == "cuda" and device.type == "cuda":
+            logger.warning(
+                "Using a CUDA random generator may produce different results than a CPU generator with the same seed."
+                " This is expected because PyTorch uses different random number generation algorithms on CPU and CUDA."
+                " If you need reproducible results across devices, use a CPU generator (e.g., torch.Generator('cpu'))."
+            )
 
     # make sure generator list of length 1 is treated like a non-list
     if isinstance(generator, list) and len(generator) == 1:


vladmandic:
this just creates a lot of extra noise in the logs for a completely expected behavior.

r266-tech:
Closing this one out. The maintainers indicated in #13298 that a warning isn't desired here since CPU/CUDA generators diverging is expected PyTorch behavior, and as @vladmandic pointed out, this would add log noise for a common, intentional usage pattern. The reproducibility docs (`using-diffusers/reusing_seeds`) already recommend a CPU `Generator` for cross-device consistency, so that guidance has this covered. Thanks for the feedback.

PR #13312, opened 2026-03-23
change QwenImageTransformer UT to batch inputs

zhtmike:
# What does this PR do?
This PR expands the inputs of `QwenImage` UT from a single prompt to multiple prompts. Since `QwenImagePipeline` already supports multi-prompt inputs, we added corresponding test coverage here.

- Tested with following tests cases with success:
TestQwenImageTransformer;TestQwenImageTransformerMemory;TestQwenImageTransformerTraining;TestQwenImageTransformerAttention;TestQwenImageTransformerContextParallel;TestQwenImageTransformerLoRA

- Tested with following tests cases with failure:
TestQwenImageTransformerLoRAHotSwap (Independent issues/bugs that also occurred on the main branch)

For ContextParallel, we keep bs=1 and will address it in another PR.

In this PR, we also add a check to skip the UT for ring attention when the SDPA backend is enabled. See the discussion in #13278.

Fixes # (issue)
- Expand the inputs for `QwenImageTransformer` from a single prompt to multiple prompts in UT
- Skip the ring attention test when the backend is `SDPA` (“native”).

## Before submitting

## Who can review?

@sayakpaul

### DIFF
--- a/tests/models/testing_utils/parallelism.py
+++ b/tests/models/testing_utils/parallelism.py
@@ -175,7 +175,6 @@ class ContextParallelTesterMixin:
 
         world_size = 2
         init_dict = self.get_init_dict()
-
         inputs_dict = self.get_dummy_inputs(batch_size=batch_size)
 
         # Move all tensors to CPU for multiprocessing


tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -78,7 +78,7 @@ def get_init_dict(self) -> dict[str, int | list[int]]:
         }
 
     def get_dummy_inputs(self) -> dict[str, torch.Tensor]:
-        batch_size = 1
+        batch_size = 2
```
sayakpaul: Instead of this, let's make this method `get_dummy_inputs()` configurable with a `batch_size` (defaulted to 1). This will allow us to test for multiple batch size inputs. WDYT?

Also, do we need changes here because from what I understand these tests/changes aren't representative of the failure we originally incurred in https://github.com/huggingface/diffusers/pull/13276. But please correct me.

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -276,6 +276,33 @@ class TestQwenImageTransformerAttention(QwenImageTransformerTesterConfig, Attent
 class TestQwenImageTransformerContextParallel(QwenImageTransformerTesterConfig, ContextParallelTesterMixin):
     """Context Parallel inference tests for QwenImage Transformer."""
 
+    def get_dummy_inputs(self) -> dict[str, torch.Tensor]:
+        batch_size = 1  # TODO: context parallel failed with batch size > 1, need fix
```
sayakpaul: Similar philosophy. 

We should rather do an `xfail` for a batch size of > 1. And then with https://github.com/huggingface/diffusers/pull/13276, we can lift it off. WDYT?

tests/models/transformers/test_models_transformer_qwenimage.py
zhtmike: Thanks!  Updated.

I think we might need this change to ensure that batch inputs can function well without context parallelism, in case we encounter any breaking changes that are incompatible with batch inputs.

tests/models/transformers/test_models_transformer_qwenimage.py
zhtmike: nice suggestion. Changed it to `xfail`

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -174,9 +177,10 @@ def test_non_contiguous_attention_mask(self):
 
         assert output.sample.shape[1] == inputs["hidden_states"].shape[1]
 
-    def test_txt_seq_lens_deprecation(self):
+    @pytest.mark.parametrize("batch_size", [1, 2])
```
sayakpaul: This is for testing deprecation. We can revert these changes.

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -276,6 +280,17 @@ class TestQwenImageTransformerAttention(QwenImageTransformerTesterConfig, Attent
...
 
+    @pytest.mark.parametrize(
+        "batch_size",
+        [
+            1,
+            pytest.param(2, marks=pytest.mark.xfail(reason="Context parallel does not support batch_size > 1")),
+        ],
+    )
+    def test_context_parallel_batch_size(self, batch_size):
+        self.get_dummy_inputs = functools.partial(self.get_dummy_inputs, batch_size=batch_size)
```
sayakpaul: Instead of `partial` let's make the modifications to
https://github.com/huggingface/diffusers/blob/c02c17c6ee7ac508c56925dde4d4a3c587650dc3/tests/models/testing_utils/parallelism.py#L170-L172

For that, we will need to make the test accept a `batch_size` argument. WDYT?

Also, WDYT about making `test_context_parallel_batch_size()` available to https://github.com/huggingface/diffusers/blob/main/tests/models/testing_utils/parallelism.py instead of making it model-specific?

tests/models/transformers/test_models_transformer_qwenimage.py
zhtmike: reverted.

tests/models/transformers/test_models_transformer_qwenimage.py
zhtmike: Agree, moved to ContextParallelTesterMixin.
Tested with both QwenImage & Flux, both are fine

tests/models/testing_utils/parallelism.py
```diff
@@ -194,6 +206,14 @@ def test_context_parallel_inference(self, cp_type):
             f"Context parallel inference failed: {return_dict.get('error', 'Unknown error')}"
         )
 
+    @pytest.mark.xfail(reason="Context parallel may not support batch_size > 1")
```
sayakpaul, line 203: Is it the case for Flux as well? 

Also, let's always require `get_dummy_inputs()` to have `batch_size`. So, we can safely remove the `inspect` stuff from here and elsewhere.

sayakpaul: @zhtmike 👀

zhtmike: > Is it the case for Flux as well?

yes. Flux works fine for bs > 2, will drop `xfail` once the qwenimage is fixed.

> Also, let's always require `get_dummy_inputs()` to have `batch_size`. So, we can safely remove the `inspect` stuff from here and elsewhere.

Done. Add batch size args to newly refactored model: flux & flux2. Tests are passed

sayakpaul:
Failing test is unrelated.

PR #13316, opened 2026-03-23
Stabilize low-precision custom autoencoder RMS normalization

YangKai0616:
# What does this PR do?

This PR improves numerical stability for four custom autoencoder RMS normalization layers.
For `fp16/bf16` inputs, normalization is computed in `fp32` and then cast back to the original dtype.

These layers are implemented on top of `F.normalize`. In low precision, the default `eps` in `F.normalize` can underflow, so zero-norm slices may effectively become `0/0` and produce `NaNs`.

Minimal reproduction:
```
import torch


def old_impl(x):
    return F.normalize(x, dim=1)


def new_impl(x):
    norm_input = x.float() if x.dtype in (torch.float16, torch.bfloat16) else x
    normalized = F.normalize(norm_input, dim=1).to(x.dtype)
    return normalized


device = "cuda"

x = torch.tensor(
    [[[[[0.0, 1.0], [0.0, -2.0]]]]],
    device=device,
    dtype=torch.float16,
)

# without this change
y_old = old_impl(x)
assert torch.isnan(y_old).any().item(), "Expected NaNs from low-precision F.normalize path"

# with this change
y_new = new_impl(x)
assert not torch.isnan(y_new).any().item(), "Expected finite output when normalization is computed in fp32"

print("old_impl_has_nan =", torch.isnan(y_old).any().item())
print("new_impl_has_nan =", torch.isnan(y_new).any().item())
print("OK")
```

Outputs:
```
old_impl_has_nan = True
new_impl_has_nan = False
OK
```

Device: A100
Torch version: 2.10 / 2.12.0.dev20260322+cu128


YangKai0616:
Hi @sayakpaul, please help review, thanks!

sayakpaul:
Thanks for the PR. Could we also add a test script in the PR description that runs assertions with and without this change?

YangKai0616:
> Thanks for the PR. Could we also add a test script in the PR description that runs assertions with and without this change?

Done!

src/diffusers/models/autoencoders/autoencoder_kl_hunyuanimage_refiner.py
```diff
@@ -87,7 +87,10 @@ def __init__(self, dim: int, channel_first: bool = True, images: bool = True, bi
         self.bias = nn.Parameter(torch.zeros(shape)) if bias else 0.0
 
     def forward(self, x):
-        return F.normalize(x, dim=(1 if self.channel_first else -1)) * self.scale * self.gamma + self.bias
+        norm_input = x.float() if x.dtype in (torch.float16, torch.bfloat16) else x
```
dg845: Do you think we should support FP4 dtypes here?

YangKai0616: I updated the change to also cover FP4/FP8-style dtypes.

dg845:
@bot /style

dg845:
Merging as the CI is green. Thanks again for the PR!

PR #13317, opened 2026-03-24
NucleusMoE-Image

sippycoder:
# What does this PR do?

This PR introduces NucleusMoE-Image series into the diffusers library. 

NucleusMoE-Image is a 2B active 17B parameter model trained with efficiency at its core. Our novel architecture highlights the scalability of sparse MoE architecture for Image generation. The technical report will be released very soon.

### DIFF
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -423,7 +423,9 @@ def dispatch_attention_fn(
         **attention_kwargs,
         "_parallel_config": parallel_config,
     }
-    if is_torch_version(">=", "2.5.0"):
+    # Equivalent to `is_torch_version(">=", "2.5.0")` — use module-level constant to avoid
+    # Dynamo tracing into the lru_cache-wrapped `is_torch_version` during torch.compile.
+    if _CAN_USE_FLEX_ATTN:
         kwargs["enable_gqa"] = enable_gqa
 
     if _AttentionBackendRegistry._checks_enabled:
--- a/src/diffusers/pipelines/ddim/pipeline_ddim.py
+++ b/src/diffusers/pipelines/ddim/pipeline_ddim.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
+
 import torch
 
 from ...models import UNet2DModel
@@ -19,6 +21,9 @@ from ...schedulers import DDIMScheduler
 from ...utils import is_torch_xla_available
 from ...utils.torch_utils import randn_tensor
 from ..pipeline_utils import DiffusionPipeline, ImagePipelineOutput
+
+
+logger = logging.getLogger(__name__)
 
 
 if is_torch_xla_available():
@@ -128,6 +133,13 @@ class DDIMPipeline(DiffusionPipeline):
             )
         else:
             image_shape = (batch_size, self.unet.config.in_channels, *self.unet.config.sample_size)
+
+        if not 0.0 <= eta <= 1.0:
+            logger.warning(
+                f"`eta` should be between 0 and 1 (inclusive), but received {eta}. "
+                "A value of 0 corresponds to DDIM and 1 corresponds to DDPM. "
+                "Unexpected results may occur for values outside this range."
+            )
 
         if isinstance(generator, list) and len(generator) != batch_size:
             raise ValueError(
--- a/src/diffusers/utils/torch_utils.py
+++ b/src/diffusers/utils/torch_utils.py
@@ -347,7 +347,17 @@ def lru_cache_unless_export(maxsize=128, typed=False):
 
         @functools.wraps(fn)
         def inner_wrapper(*args: P.args, **kwargs: P.kwargs):
-            if torch.compiler.is_exporting():
+            compiler = getattr(torch, "compiler", None)
+            is_exporting = bool(compiler and hasattr(compiler, "is_exporting") and compiler.is_exporting())
+            is_compiling = bool(compiler and hasattr(compiler, "is_compiling") and compiler.is_compiling())
+
+            # Fallback for older builds where compiler.is_compiling is unavailable.
+            if not is_compiling:
+                dynamo = getattr(torch, "_dynamo", None)
+                if dynamo is not None and hasattr(dynamo, "is_compiling"):
+                    is_compiling = dynamo.is_compiling()
+
+            if is_exporting or is_compiling:
                 return fn(*args, **kwargs)
             return cached(*args, **kwargs)
 
--- a/tests/models/autoencoders/test_models_autoencoder_wan.py
+++ b/tests/models/autoencoders/test_models_autoencoder_wan.py
@@ -13,24 +13,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import unittest
+import pytest
+import torch
 
 from diffusers import AutoencoderKLWan
+from diffusers.utils.torch_utils import randn_tensor
 
-from ...testing_utils import enable_full_determinism, floats_tensor, torch_device
-from ..test_modeling_common import ModelTesterMixin
-from .testing_utils import AutoencoderTesterMixin
+from ...testing_utils import enable_full_determinism, torch_device
+from ..testing_utils import BaseModelTesterConfig, MemoryTesterMixin, ModelTesterMixin, TrainingTesterMixin
+from .testing_utils import NewAutoencoderTesterMixin
 
 
 enable_full_determinism()
 
 
-class AutoencoderKLWanTests(ModelTesterMixin, AutoencoderTesterMixin, unittest.TestCase):
-    model_class = AutoencoderKLWan
-    main_input_name = "sample"
-    base_precision = 1e-2
+class AutoencoderKLWanTesterConfig(BaseModelTesterConfig):
+    @property
+    def model_class(self):
+        return AutoencoderKLWan
 
-    def get_autoencoder_kl_wan_config(self):
+    @property
+    def output_shape(self):
+        return (3, 9, 16, 16)
+
+    @property
+    def generator(self):
+        return torch.Generator("cpu").manual_seed(0)
+
+    def get_init_dict(self):
         return {
             "base_dim": 3,
             "z_dim": 16,
@@ -39,54 +49,40 @@ class AutoencoderKLWanTests(ModelTesterMixin, AutoencoderTesterMixin, unittest.T
             "temperal_downsample": [False, True, True],
         }
 
-    @property
-    def dummy_input(self):
+    def get_dummy_inputs(self):
         batch_size = 2
         num_frames = 9
         num_channels = 3
         sizes = (16, 16)
-        image = floats_tensor((batch_size, num_channels, num_frames) + sizes).to(torch_device)
+        image = randn_tensor(
+            (batch_size, num_channels, num_frames, *sizes), generator=self.generator, device=torch_device
+        )
         return {"sample": image}
 
-    @property
-    def dummy_input_tiling(self):
-        batch_size = 2
-        num_frames = 9
-        num_channels = 3
-        sizes = (128, 128)
-        image = floats_tensor((batch_size, num_channels, num_frames) + sizes).to(torch_device)
-        return {"sample": image}
 
-    @property
-    def input_shape(self):
-        return (3, 9, 16, 16)
+class TestAutoencoderKLWan(AutoencoderKLWanTesterConfig, ModelTesterMixin):
+    base_precision = 1e-2
 
-    @property
-    def output_shape(self):
-        return (3, 9, 16, 16)
 
-    def prepare_init_args_and_inputs_for_common(self):
-        init_dict = self.get_autoencoder_kl_wan_config()
-        inputs_dict = self.dummy_input
-        return init_dict, inputs_dict
+class TestAutoencoderKLWanTraining(AutoencoderKLWanTesterConfig, TrainingTesterMixin):
+    """Training tests for AutoencoderKLWan."""
 
-    def prepare_init_args_and_inputs_for_tiling(self):
-        init_dict = self.get_autoencoder_kl_wan_config()
-        inputs_dict = self.dummy_input_tiling
-        return init_dict, inputs_dict
-
-    @unittest.skip("Gradient checkpointing has not been implemented yet")
+    @pytest.mark.skip(reason="Gradient checkpointing has not been implemented yet")
     def test_gradient_checkpointing_is_applied(self):
         pass
 
-    @unittest.skip("Test not supported")
-    def test_forward_with_norm_groups(self):
+
+class TestAutoencoderKLWanMemory(AutoencoderKLWanTesterConfig, MemoryTesterMixin):
+    """Memory optimization tests for AutoencoderKLWan."""
+
+    @pytest.mark.skip(reason="RuntimeError: fill_out not implemented for 'Float8_e4m3fn'")
+    def test_layerwise_casting_memory(self):
         pass
 
-    @unittest.skip("RuntimeError: fill_out not implemented for 'Float8_e4m3fn'")
-    def test_layerwise_casting_inference(self):
-        pass
-
-    @unittest.skip("RuntimeError: fill_out not implemented for 'Float8_e4m3fn'")
+    @pytest.mark.skip(reason="RuntimeError: fill_out not implemented for 'Float8_e4m3fn'")
     def test_layerwise_casting_training(self):
         pass
+
+
+class TestAutoencoderKLWanSlicingTiling(AutoencoderKLWanTesterConfig, NewAutoencoderTesterMixin):
+    """Slicing and tiling tests for AutoencoderKLWan."""
--- a/tests/models/autoencoders/testing_utils.py
+++ b/tests/models/autoencoders/testing_utils.py
@@ -145,3 +145,138 @@ class AutoencoderTesterMixin:
             output_without_slicing.detach().cpu().numpy().all(),
             output_without_slicing_2.detach().cpu().numpy().all(),
         ), "Without slicing outputs should match with the outputs when slicing is manually disabled."
+
+
+class NewAutoencoderTesterMixin:
+    @staticmethod
+    def _accepts_generator(model):
+        model_sig = inspect.signature(model.forward)
+        accepts_generator = "generator" in model_sig.parameters
+        return accepts_generator
+
+    @staticmethod
+    def _accepts_norm_num_groups(model_class):
+        model_sig = inspect.signature(model_class.__init__)
+        accepts_norm_groups = "norm_num_groups" in model_sig.parameters
+        return accepts_norm_groups
+
+    def test_forward_with_norm_groups(self):
+        if not self._accepts_norm_num_groups(self.model_class):
+            pytest.skip(f"Test not supported for {self.model_class.__name__}")
+        init_dict = self.get_init_dict()
+        inputs_dict = self.get_dummy_inputs()
+
+        init_dict["norm_num_groups"] = 16
+        init_dict["block_out_channels"] = (16, 32)
+
+        model = self.model_class(**init_dict)
+        model.to(torch_device)
+        model.eval()
+
+        with torch.no_grad():
+            output = model(**inputs_dict)
+
+            if isinstance(output, dict):
+                output = output.to_tuple()[0]
+
+        assert output is not None
+        expected_shape = inputs_dict["sample"].shape
+        assert output.shape == expected_shape, "Input and output shapes do not match"
+
+    def test_enable_disable_tiling(self):
+        if not hasattr(self.model_class, "enable_tiling"):
+            pytest.skip(f"Skipping test as {self.model_class.__name__} doesn't support tiling.")
+
+        init_dict = self.get_init_dict()
+        inputs_dict = self.get_dummy_inputs()
+
+        torch.manual_seed(0)
+        model = self.model_class(**init_dict).to(torch_device)
+
+        if not hasattr(model, "use_tiling"):
+            pytest.skip(f"Skipping test as {self.model_class.__name__} doesn't support tiling.")
+
+        inputs_dict.update({"return_dict": False})
+        _ = inputs_dict.pop("generator", None)
+        accepts_generator = self._accepts_generator(model)
+
+        with torch.no_grad():
+            torch.manual_seed(0)
+            if accepts_generator:
+                inputs_dict["generator"] = torch.manual_seed(0)
+            output_without_tiling = model(**inputs_dict)[0]
+            if isinstance(output_without_tiling, DecoderOutput):
+                output_without_tiling = output_without_tiling.sample
+
+            torch.manual_seed(0)
+            model.enable_tiling()
+            if accepts_generator:
+                inputs_dict["generator"] = torch.manual_seed(0)
+            output_with_tiling = model(**inputs_dict)[0]
+            if isinstance(output_with_tiling, DecoderOutput):
+                output_with_tiling = output_with_tiling.sample
+
+            assert (output_without_tiling.cpu() - output_with_tiling.cpu()).max() < 0.5, (
+                "VAE tiling should not affect the inference results"
+            )
+
+            torch.manual_seed(0)
+            model.disable_tiling()
+            if accepts_generator:
+                inputs_dict["generator"] = torch.manual_seed(0)
+            output_without_tiling_2 = model(**inputs_dict)[0]
+            if isinstance(output_without_tiling_2, DecoderOutput):
+                output_without_tiling_2 = output_without_tiling_2.sample
+
+            assert torch.allclose(output_without_tiling.cpu(), output_without_tiling_2.cpu()), (
+                "Without tiling outputs should match with the outputs when tiling is manually disabled."
+            )
+
+    def test_enable_disable_slicing(self):
+        if not hasattr(self.model_class, "enable_slicing"):
+            pytest.skip(f"Skipping test as {self.model_class.__name__} doesn't support slicing.")
+
+        init_dict = self.get_init_dict()
+        inputs_dict = self.get_dummy_inputs()
+
+        torch.manual_seed(0)
+        model = self.model_class(**init_dict).to(torch_device)
+        if not hasattr(model, "use_slicing"):
+            pytest.skip(f"Skipping test as {self.model_class.__name__} doesn't support tiling.")
+
+        inputs_dict.update({"return_dict": False})
+        _ = inputs_dict.pop("generator", None)
+        accepts_generator = self._accepts_generator(model)
+
+        with torch.no_grad():
+            if accepts_generator:
+                inputs_dict["generator"] = torch.manual_seed(0)
+
+            torch.manual_seed(0)
+            output_without_slicing = model(**inputs_dict)[0]
+            if isinstance(output_without_slicing, DecoderOutput):
+                output_without_slicing = output_without_slicing.sample
+
+            torch.manual_seed(0)
+            model.enable_slicing()
+            if accepts_generator:
+                inputs_dict["generator"] = torch.manual_seed(0)
+            output_with_slicing = model(**inputs_dict)[0]
+            if isinstance(output_with_slicing, DecoderOutput):
+                output_with_slicing = output_with_slicing.sample
+
+            assert (output_without_slicing.cpu() - output_with_slicing.cpu()).max() < 0.5, (
+                "VAE slicing should not affect the inference results"
+            )
+
+            torch.manual_seed(0)
+            model.disable_slicing()
+            if accepts_generator:
+                inputs_dict["generator"] = torch.manual_seed(0)
+            output_without_slicing_2 = model(**inputs_dict)[0]
+            if isinstance(output_without_slicing_2, DecoderOutput):
+                output_without_slicing_2 = output_without_slicing_2.sample
+
+            assert torch.allclose(output_without_slicing.cpu(), output_without_slicing_2.cpu()), (
+                "Without slicing outputs should match with the outputs when slicing is manually disabled."
+            )


sippycoder:
cc: @sayakpaul @IlyasMoutawwakil

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+from ..embeddings import TimestepEmbedding, Timesteps
+from ..modeling_outputs import Transformer2DModelOutput
+from ..modeling_utils import ModelMixin
+from ..normalization import AdaLayerNormContinuous, RMSNorm
+
+
+logger = logging.get_logger(__name__)
+
+
+# copied from diffusers.models.transformers.transformer_qwenimage.apply_rotary_emb_qwen
```
dg845:
```suggestion
# Copied from diffusers.models.transformers.transformer_qwenimage.apply_rotary_emb_qwen with qwen->nucleus
```
nit: `# Copied from` mechanism supports renamings with the above syntax

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+            if addition_t_cond is None:
+                raise ValueError("When additional_t_cond is True, addition_t_cond must be provided.")
+            addition_t_emb = self.addition_t_embedding(addition_t_cond)
+            addition_t_emb = addition_t_emb.to(dtype=hidden_states.dtype)
+            conditioning = conditioning + addition_t_emb
+
+        return self.norm(conditioning)
+
+
+# copied from diffusers.models.transformers.transformer_qwenimage.QwenEmbedRope
```
dg845:
```suggestion
# Copied from diffusers.models.transformers.transformer_qwenimage.QwenEmbedRope with Qwen->NucleusMoE
```
See https://github.com/huggingface/diffusers/pull/13317#discussion_r2985636856. Alternatively, if `NucleusMoEEmbedRope` is changed (for example to remove `txt_seq_lens` as suggested in https://github.com/huggingface/diffusers/pull/13317#discussion_r2985653263), the `# Copied from` statement should be removed.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+                The maximum text sequence length for RoPE computation.
+        """
+        if txt_seq_lens is not None:
+            deprecate(
+                "txt_seq_lens",
+                "0.39.0",
+                "Passing `txt_seq_lens` is deprecated and will be removed in version 0.39.0. "
+                "Please use `max_txt_seq_len` instead.",
+                standard_warn=False,
+            )
```
dg845: As this is a new model, can we remove the dependence on the deprecated `txt_seq_lens` argument?

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+    @staticmethod
+    def _rope_params(index, dim, theta=10000):
+        assert dim % 2 == 0
+        freqs = torch.outer(index, 1.0 / torch.pow(theta, torch.arange(0, dim, 2).to(torch.float32).div(dim)))
+        freqs = torch.polar(torch.ones_like(freqs), freqs)
+        return freqs
+
+    def forward(
+        self,
+        video_fhw: tuple[int, int, int, list[tuple[int, int, int]]],
```
dg845:
```suggestion
        video_fhw: tuple[int, int, int] | list[tuple[int, int, int]],
```
nit: fix type annotation

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+        out = self.shared_expert(hidden_states).reshape(bs * slen, dim)
+
+        scatter_idx = global_token_indices.reshape(-1, 1).expand(-1, dim)
+        out = out.scatter_add(dim=0, index=scatter_idx, src=routed_output)
+        out = out.reshape(bs, slen, dim)
+
+        return out
+
+
+@maybe_allow_in_graph
```
dg845: I think we can remove the `maybe_allow_in_graph` decorator as the NucleusMoE-Image transformer compile tests

```bash
RUN_SLOW=1 RUN_COMPILE=1 pytest tests/models/transformers/test_models_transformer_nucleusmoe_image.py::TestNucleusMoEImageTransformerCompile
```

have the same pass/fail pattern with and without it. (Currently, `test_compile_on_different_shapes` fails both with and without `maybe_allow_in_graph`; all other tests pass.)

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        encoder_hidden_states: torch.Tensor,
+        temb: torch.Tensor,
+        image_rotary_emb: tuple[torch.Tensor, torch.Tensor] | None = None,
+        attention_kwargs: dict[str, Any] | None = None,
+    ) -> torch.Tensor:
+        scale1, gate1, scale2, gate2 = self.img_mod(temb).unsqueeze(1).chunk(4, dim=-1)
+        scale1, scale2 = 1 + scale1, 1 + scale2
```
dg845: nit: I think it's more clear if we do the calculation inline as e.g. `img_modulated = img_normed * (1 + scale1)`.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+
+        gate1 = gate1.clamp(min=-2.0, max=2.0)
+        gate2 = gate2.clamp(min=-2.0, max=2.0)
```
dg845, line 692: It seems weird to me that we first clamp the gates to `[-2.0, 2.0]` and then essentially clamp again by squashing with the `tanh` function below. Is this intended?

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+
+        if self.moe_enabled:
+            img_mlp_output = self.img_mlp(img_modulated2, img_normed2, timestep=temb)
+        else:
+            img_mlp_output = self.img_mlp(img_modulated2)
+
+        hidden_states = hidden_states + gate2.tanh() * img_mlp_output
+
+        if hidden_states.dtype == torch.float16:
+            hidden_states = hidden_states.clip(-65504, 65504)
```
dg845:
```suggestion
        if hidden_states.dtype == torch.float16:
            fp16_finfo = torch.finfo(torch.float16)
            hidden_states = hidden_states.clip(fp16_finfo.min, fp16_finfo.max)
```

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+        num_attention_heads: int = 16,
+        num_key_value_heads: int | None = None,
+        joint_attention_dim: int = 3584,
+        axes_dims_rope: tuple[int, int, int] = (16, 56, 56),
+        mlp_ratio: float = 4.0,
+        moe_enabled: bool = True,
+        dense_moe_strategy: str = "leave_first_three_and_last_block_dense",
+        num_experts: int = 128,
+        moe_intermediate_dim: int = 1344,
+        capacity_factors: List[float] = [8.0] * 24,
```
dg845:
```suggestion
        capacity_factors: float | list[float] = 8.0,
```
I think allowing `capacity_factors` to take `float` arguments as well makes the code a little cleaner. We would then expand `float` inputs to a list inside `__init__`:

```python
if isinstance(capacity_factors, float):
    capacity_factors = [capacity_factors] * num_layers
```

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+
+        self.norm_out = AdaLayerNormContinuous(self.inner_dim, self.inner_dim, elementwise_affine=False, eps=1e-6)
+        self.proj_out = nn.Linear(self.inner_dim, patch_size * patch_size * self.out_channels, bias=False)
+
+        self.gradient_checkpointing = False
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        img_shapes: list[tuple[int, int, int]] | None = None,
```
dg845:
```suggestion
        img_shapes: tuple[int, int, int] | list[tuple[int, int, int]],
```
I think allowing `img_shapes` to take `tuple[int, int, int]` arguments as well would be cleaner, similar to https://github.com/huggingface/diffusers/pull/13317#discussion_r2985717765. If I understand correctly, `NucleusMoEEmbedRope` only accepts batches with the same image shape, so this would make it easier to specify such shapes.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+        """
+        if txt_seq_lens is not None:
+            deprecate(
+                "txt_seq_lens",
+                "0.39.0",
+                "Passing `txt_seq_lens` is deprecated and will be removed in version 0.39.0. "
+                "Please use `encoder_hidden_states_mask` instead.",
+                standard_warn=False,
+            )
+
```
dg845: I think we can remove the deprecated `txt_seq_lens` argument here as well. See https://github.com/huggingface/diffusers/pull/13317#discussion_r2985653263.

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+        ... )
+        >>> pipe.to("cuda")
+        >>> prompt = "A cat holding a sign that says hello world"
+        >>> image = pipe(prompt, num_inference_steps=50).images[0]
+        >>> image.save("nucleus_moe.png")
+        ```
+"""
+
+
+def calculate_shift(
```
dg845, line 59:
```suggestion
# Copied from diffusers.pipelines.qwenimage.pipeline_qwenimage.calculate_shift
def calculate_shift(
```

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+    base_shift: float = 0.5,
+    max_shift: float = 1.15,
+):
+    m = (max_shift - base_shift) / (max_seq_len - base_seq_len)
+    b = base_shift - m * base_seq_len
+    mu = image_seq_len * m + b
+    return mu
+
+
+def retrieve_timesteps(
```
dg845, line 73:
```suggestion
# Copied from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion.retrieve_timesteps
def retrieve_timesteps(
```

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+            vae=vae,
+            text_encoder=text_encoder,
+            processor=processor,
+        )
+        self.vae_scale_factor = (
+            2 ** len(self.vae.temperal_downsample) if getattr(self, "vae", None) else 8
+        )
+        self.image_processor = VaeImageProcessor(vae_scale_factor=self.vae_scale_factor * 2)
+        self.default_sample_size = 128
+        self.return_index = -8
```
dg845: Should `default_sample_size` and `return_index` be configurable via `__init__`?

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+    def check_inputs(
+        self,
+        prompt,
+        height,
+        width,
+        negative_prompt=None,
+        prompt_embeds=None,
+        negative_prompt_embeds=None,
+        prompt_embeds_mask=None,
+        negative_prompt_embeds_mask=None,
```
dg845, line 260:
```suggestion
```
nit: remove `prompt_embeds_mask` and `negative_prompt_embeds_mask` as they are not used in `check_inputs`.

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+
+    @staticmethod
+    def _pack_latents(latents, batch_size, num_channels_latents, height, width):
+        latents = latents.view(batch_size, num_channels_latents, height // 2, 2, width // 2, 2)
+        latents = latents.permute(0, 2, 4, 1, 3, 5)
+        latents = latents.reshape(batch_size, (height // 2) * (width // 2), num_channels_latents * 4)
+        return latents
+
+    @staticmethod
+    def _unpack_latents(latents, height, width, vae_scale_factor):
```
dg845: Could we refactor `_pack_latents` and `_unpack_latents` to take a `patch_size` argument instead of hardcoding the patch size to `2`? This would make the code more robust.

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+        num_channels_latents,
+        height,
+        width,
+        dtype,
+        device,
+        generator,
+        latents=None,
+    ):
+        height = 2 * (int(height) // (self.vae_scale_factor * 2))
+        width = 2 * (int(width) // (self.vae_scale_factor * 2))
```
dg845: Could we refactor this to use `self.transformer.config.patch_size` instead of hardcoding the patch size to `2`? See also https://github.com/huggingface/diffusers/pull/13317#discussion_r2985771288.

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+            raise ValueError(
+                f"You have passed a list of generators of length {len(generator)}, but requested an effective batch"
+                f" size of {batch_size}. Make sure the batch size matches the length of the generators."
+            )
+
+        latents = randn_tensor(shape, generator=generator, device=device, dtype=dtype)
+        latents = self._pack_latents(latents, batch_size, num_channels_latents, height, width)
+        return latents
+
+    def enable_vae_slicing(self):
```
dg845: We can remove the VAE slicing/tiling methods here as they are deprecated. Users can always call the corresponding methods on the VAE itself (e.g. `pipe.vae.enable_tiling()`) to enable/disable slicing/tiling.

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+    def interrupt(self):
+        return self._interrupt
+
+    @torch.no_grad()
+    @replace_example_docstring(EXAMPLE_DOC_STRING)
+    def __call__(
+        self,
+        prompt: str | list[str] = None,
+        negative_prompt: str | list[str] = None,
+        true_cfg_scale: float = 4.0,
```
dg845:
```suggestion
        guidance_scale: float = 4.0,
```
nit: rename to `guidance_scale` to follow the `diffusers` CFG naming conventions.

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+            width,
+            prompt_embeds.dtype,
+            device,
+            generator,
+            latents,
+        )
+
+        latent_h = 2 * (int(height) // (self.vae_scale_factor * 2))
+        latent_w = 2 * (int(width) // (self.vae_scale_factor * 2))
+        img_shapes = [(1, latent_h // 2, latent_w // 2)] * (batch_size * num_images_per_prompt)
```
dg845: Similar to https://github.com/huggingface/diffusers/pull/13317#discussion_r2985778659, can we refactor this to use `self.transformer.config.patch_size`?

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+            for i, t in enumerate(timesteps):
+                if self.interrupt:
+                    continue
+
+                self._current_timestep = t
+                timestep = t.expand(latents.shape[0]).to(latents.dtype)
+
+                noise_pred = self.transformer(
+                    hidden_states=latents,
+                    timestep=timestep / 1000,
```
dg845: Instead of hardcoding this at `1000`, could we use `self.scheduler.config.num_train_timesteps` instead?

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
```diff
@@ -0,0 +1,667 @@
...
+                        attention_kwargs=self._attention_kwargs,
+                        return_dict=False,
+                    )[0]
+
+                    comb_pred = neg_noise_pred + true_cfg_scale * (noise_pred - neg_noise_pred)
+                    cond_norm = torch.norm(noise_pred, dim=-1, keepdim=True)
+                    noise_norm = torch.norm(comb_pred, dim=-1, keepdim=True)
+                    noise_pred = comb_pred * (cond_norm / noise_norm)
+
+                noise_pred = -noise_pred
```
dg845, line 597: Why do we need to negate `noise_pred` here?

src/diffusers/hooks/text_kv_cache.py
```diff
@@ -0,0 +1,109 @@
...
+
+class TextKVCacheHook(ModelHook):
+    """Block-level hook that caches (txt_key, txt_value) per unique prompt."""
+
+    _is_stateful = True
+
+    def __init__(self):
+        super().__init__()
+        # Maps encoder_hidden_states.data_ptr() → (txt_key, txt_value)
+        self.kv_cache: dict[int, tuple] = {}
```
dg845:
```suggestion
        self.kv_cache: dict[int, tuple[torch.Tensor, torch.Tensor]] = {}
```
nit: more specific type annotation

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,788 @@
...
+                FeedForward(
+                    dim=hidden_size,
+                    dim_out=hidden_size,
+                    inner_dim=moe_intermediate_dim,
+                    activation_fn="swiglu",
+                    bias=False,
+                )
+                for _ in range(num_experts)
+            ]
+        )
```
IlyasMoutawwakil: you would need the projections to be in packed/contiguous format for `torch.grouped_mm` support (num_experts, dim_in, dim_out), @sayakpaul is that possible ? in Transformers we use the inline weight converter

sayakpaul: Not at the moment because MoEs are still a bit of a special case in this part of world.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
sippycoder: I agree it's weird. :) I used it to stabilize the gradients if the tanh gates get saturated while training. I will evaluate the model performance without it and get back to you!

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
sippycoder: TIL.

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
sippycoder: `default_sample_size` is used when `width` and `height` are None. It can be safely ignored. I reduced it to 64 since we don't support 2048 resolution yet. 
I moved `return_index` to `__call__` arguments

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
sippycoder: I can pack the MoE weights. That's how I originally trained the model with Expert Parallel.

src/diffusers/pipelines/nucleusmoe_image/pipeline_nucleusmoe_image.py
sippycoder: We trained our model with negative velocity. I had to negate it to make it compatible with the scheduler. It's NucleusMoE-Image specific behavior.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,761 @@
...
+        Args:
+            video_fhw (`tuple[int, int, int]` or `list[tuple[int, int, int]]`):
+                A list of 3 integers [frame, height, width] representing the shape of the video.
+            device: (`torch.device`, *optional*):
+                The device on which to perform the RoPE computation.
+            max_txt_seq_len (`int` or `torch.Tensor`, *optional*):
+                The maximum text sequence length for RoPE computation.
+        """
+        if max_txt_seq_len is None:
+            raise ValueError("Either `max_txt_seq_len` must be provided.")
```
dg845, line 194: Would it be possible to provide a reasonable default value for `max_txt_seq_len` instead of raising an error?

tests/models/transformers/test_models_transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,242 @@
...
+            "hidden_states": hidden_states,
+            "encoder_hidden_states": encoder_hidden_states,
+            "encoder_hidden_states_mask": encoder_hidden_states_mask,
+            "timestep": timestep,
+            "img_shapes": img_shapes,
+        }
+
+
+class TestNucleusMoEImageTransformer(NucleusMoEImageTransformerTesterConfig, ModelTesterMixin):
+    def test_txt_seq_lens_deprecation(self):
```
dg845: I think we can remove this test now that `txt_seq_lens` has been removed from the transformer's `forward` method.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
sippycoder: Hey @IlyasMoutawwakil, I packed the MoE weights and add the `SwiGLUExperts` module to facilitate adding Expert Parallel hooks. Can you please review? Thanks!

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,918 @@
...
+        offsets = torch.cumsum(num_tokens_per_expert, dim=0, dtype=torch.int32)
+
+        # Gate path: grouped GEMM + SiLU.  shape: (total_tokens, I)
+        gate = F.silu(
+            torch._grouped_mm(x.bfloat16(), self.w1.bfloat16(), offs=offsets)
+        )
+        # Up path: grouped GEMM, no activation.  shape: (total_tokens, I)
+        up = torch._grouped_mm(x.bfloat16(), self.w3.bfloat16(), offs=offsets)
+        # Down projection; cast output back to input dtype.  shape: (total_tokens, H)
+        out = torch._grouped_mm(gate * up, self.w2.bfloat16(), offs=offsets).type_as(x)
```
IlyasMoutawwakil: great ! you might wanna use the public api instead (torch.nn.functional.grouped_mm)

IlyasMoutawwakil: also not sure if it makes sense to force bf16 here, the kernel works fine on fp32 but fails when compiled with a dtype other than bf16

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,918 @@
...
+        # Exclusive end offsets: offsets[k] is the first row of expert k+1's tokens,
+        # i.e. the exclusive end of expert k's token range.
+        offsets = torch.cumsum(num_tokens_per_expert, dim=0, dtype=torch.int32)
+
+        # Gate path: grouped GEMM + SiLU.  shape: (total_tokens, I)
+        gate = F.silu(
+            torch._grouped_mm(x.bfloat16(), self.w1.bfloat16(), offs=offsets)
+        )
+        # Up path: grouped GEMM, no activation.  shape: (total_tokens, I)
+        up = torch._grouped_mm(x.bfloat16(), self.w3.bfloat16(), offs=offsets)
```
IlyasMoutawwakil: you can also fuse the gate and up expert weight matrices for a fused projection

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
sippycoder: > also not sure if it makes sense to force bf16 here, the kernel works fine on fp32 but fails when compiled with a dtype other than bf16

Ah ok. That makes sense. I will remove casts.

sippycoder: Makes sense for generality. Done.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
sippycoder: Great idea. Done.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,908 @@
...
+                raise ValueError("When additional_t_cond is True, addition_t_cond must be provided.")
+            addition_t_emb = self.addition_t_embedding(addition_t_cond)
+            addition_t_emb = addition_t_emb.to(dtype=hidden_states.dtype)
+            conditioning = conditioning + addition_t_emb
+
+        return self.norm(conditioning)
+
+
+# Copied from diffusers.models.transformers.transformer_qwenimage.QwenEmbedRope with Qwen->NucleusMoE
+class NucleusMoEEmbedRope(nn.Module):
```
dg845:
```suggestion
class NucleusMoEEmbedRope(nn.Module):
```
I think we can remove the `# Copied from` statement here as the implementation has diverged from the QwenImage implementation.

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,908 @@
...
+            vid_freqs.append(video_freq)
+
+            if self.scale_rope:
+                max_vid_index = max(height // 2, width // 2, max_vid_index)
+            else:
+                max_vid_index = max(height, width, max_vid_index)
+
+        max_txt_seq_len_int = int(max_txt_seq_len)
+        txt_freqs = self.pos_freqs.to(device)[max_vid_index : max_vid_index + max_txt_seq_len_int, ...]
+        vid_freqs = torch.cat(vid_freqs, dim=0)
```
dg845, line 230:
```suggestion
        vid_freqs = []
        for idx, fhw in enumerate(video_fhw):
            frame, height, width = fhw
            video_freq = self._compute_video_freqs(frame, height, width, idx, device)
            vid_freqs.append(video_freq)

            max_txt_seq_len_int = int(max_txt_seq_len)
            if self.scale_rope:
                max_vid_index = torch.maximum(
                    torch.tensor(height // 2, device=device, dtype=torch.long),
                    torch.tensor(width // 2, device=device, dtype=torch.long),
                )
            else:
                max_vid_index = torch.maximum(
                    torch.tensor(height, device=device, dtype=torch.long),
                    torch.tensor(width, device=device, dtype=torch.long),
                )

        txt_freqs = self.pos_freqs.to(device)[max_vid_index + torch.arange(max_txt_seq_len_int, device=device)]
        vid_freqs = torch.cat(vid_freqs, dim=0)
```
Claude suggests this refactor to fix the `RecompileError` that occurs in the `test_compile_on_different_shapes` test. My understanding is that the idea here is to avoid Python-level control flow and values in favor of tensor operations and values, which Dynamo can trace into a single graph.

I have confirmed that this refactor makes the test pass locally. It also fixes some (but not all) of the failing `TestNucleusMoEImageTransformerLoRAHotSwap` tests.

src/diffusers/hooks/text_kv_cache.py
```diff
@@ -0,0 +1,146 @@
...
+    The transformer hook writes the stable ``encoder_hidden_states``
+    ``data_ptr()`` (captured *before* ``txt_norm``) so that block hooks
+    can use it as a reliable cache key across denoising steps.
+    """
+
+    def __init__(self):
+        self.key: int | None = None
+
+
+class _TextKVCacheTransformerHook(ModelHook):
```
dg845:
```suggestion
class TextKVCacheTransformerHook(ModelHook):
```
nit: omit the leading underscore, following e.g. `MagCacheHeadHook`.

src/diffusers/hooks/text_kv_cache.py
```diff
@@ -0,0 +1,146 @@
...
+        if encoder_hidden_states is not None:
+            self._shared.key = encoder_hidden_states.data_ptr()
+        return self.fn_ref.original_forward(*args, **kwargs)
+
+    def reset_state(self, module: torch.nn.Module):
+        self._shared.key = None
+        return module
+
+
+class _TextKVCacheBlockHook(ModelHook):
```
dg845:
```suggestion
class TextKVCacheBlockHook(ModelHook):
```
nit: similarly, omit the leading underscore here, following `MagCacheBlockHook`.

src/diffusers/hooks/text_kv_cache.py
```diff
@@ -0,0 +1,146 @@
...
+    Pre-computes per-block text key and value projections once before the
+    denoising loop and reuses them across all steps. Positive and negative
+    prompts are distinguished via a stable cache key captured by a
+    transformer-level hook before any intermediate tensor allocations.
+    """
+
+    pass
+
+
+class _SharedCacheKey:
```
dg845: Would it be possible to refactor `_SharedCacheKey` into a `TextKVCacheState`, similar to how `MagCacheState` and `TaylorSeerState` are implemented?

https://github.com/huggingface/diffusers/blob/7e463ea4cce63063198909950da58107cc0a52cf/src/diffusers/hooks/mag_cache.py#L140

CC @yiyixuxu to confirm that this is the design we want going forward.

src/diffusers/hooks/text_kv_cache.py
```diff
@@ -0,0 +1,146 @@
...
+class _TextKVCacheBlockHook(ModelHook):
+    """Caches ``(txt_key, txt_value)`` per block per unique prompt using
+    the stable cache key from the shared state."""
+
+    _is_stateful = True
+
+    def __init__(self, shared: _SharedCacheKey):
+        super().__init__()
+        self._shared = shared
+        self.kv_cache: dict[int, tuple[torch.Tensor, torch.Tensor]] = {}
```
dg845: Could we refactor this to use `StateManager` instead?

https://github.com/huggingface/diffusers/blob/7e463ea4cce63063198909950da58107cc0a52cf/src/diffusers/hooks/hooks.py#L34

I believe this is the way MagCache and TaylorSeer implement their caches. CC @yiyixuxu

src/diffusers/hooks/text_kv_cache.py
```diff
@@ -0,0 +1,146 @@
...
+        self.kv_cache.clear()
+        return module
+
+
+def apply_text_kv_cache(module: torch.nn.Module, config: TextKVCacheConfig) -> None:
+    from ..models.transformers.transformer_nucleusmoe_image import NucleusMoEImageTransformerBlock
+
+    shared = _SharedCacheKey()
+
+    transformer_hook = _TextKVCacheTransformerHook(shared)
```
dg845:
```suggestion
    transformer_hook = TextKVCacheTransformerHook(shared)
```
Follow up change to https://github.com/huggingface/diffusers/pull/13317#discussion_r3013990264.

src/diffusers/hooks/text_kv_cache.py
```diff
@@ -0,0 +1,146 @@
...
+    registry = HookRegistry.check_if_exists_or_initialize(module)
+    registry.register_hook(transformer_hook, _TEXT_KV_CACHE_TRANSFORMER_HOOK)
+
+    for _, submodule in module.named_modules():
+        if isinstance(submodule, NucleusMoEImageTransformerBlock):
+            hook = _TextKVCacheBlockHook(shared)
```
dg845:
```suggestion
            hook = TextKVCacheBlockHook(shared)
```
Follow up change to https://github.com/huggingface/diffusers/pull/13317#discussion_r3013994825.

sippycoder:
@dg845 I gave you access to the model at `NucleusAI/NucleusMoE-Image`

src/diffusers/hooks/text_kv_cache.py
sippycoder: Sure! I updated the PR with this change.

src/diffusers/hooks/text_kv_cache.py
sippycoder: Done.

dg845:
Hi @sippycoder, it doesn't look like my HF account (also `dg845`) has access to `NucleusAI/NucleusMoE-Image`, not sure if I am missing something. I get a 404 error if I try to access it.

sippycoder:
> Hi @sippycoder, it doesn't look like my HF account (also `dg845`) has access to `NucleusAI/NucleusMoE-Image`, not sure if I am missing something. I get a 404 error if I try to access it.

Looks like I can't give you private repo access unless you are in my org. I just made the repo public! I didn't update the model page yet.

dg845:
@bot /style

src/diffusers/models/transformers/transformer_nucleusmoe_image.py
```diff
@@ -0,0 +1,907 @@
...
+from ..embeddings import TimestepEmbedding, Timesteps
+from ..modeling_outputs import Transformer2DModelOutput
+from ..modeling_utils import ModelMixin
+from ..normalization import AdaLayerNormContinuous, RMSNorm
+
+
+logger = logging.get_logger(__name__)
+
+
+# Copied from diffusers.models.transformers.transformer_qwenimage.apply_rotary_emb_qwen with qwen->nucleus
```
dg845, line 39: Can you run `make fix-copies` to sync the implementation here with the QwenImage implementation (assuming the implementations are intended to be the same, which I believe is the case)?

sippycoder: Done!

sippycoder:
@yiyixuxu Any comments from you for the `text_kv_cache` hook?
@dg845 Do you think we are good to merge this PR?

dg845:
@bot /style

dg845:
Hi @sippycoder, I think this PR is close to merge, the remaining items should be:

1. Confirm that the text KV cache design looks good
2. Ensure that the MoE weights are correctly supported (e.g. https://github.com/huggingface/diffusers/pull/13317#discussion_r2986803910). I think with the new changes they are probably good, CC @IlyasMoutawwakil to confirm.

Additionally, having docs would be nice, but this is not a hard blocker (we can add them in a follow-up PR if necessary).

sippycoder:
Hey @dg845 @yiyixuxu , we are releasing the model report tomorrow! Thanks for all the reviews! I think Expert Parallelism can be a separate PR. Any chance we can merge the PR today?

src/diffusers/hooks/text_kv_cache.py
```diff
@@ -0,0 +1,157 @@
...
+class TextKVCacheBlockHook(ModelHook):
+    """Caches ``(txt_key, txt_value)`` per block per unique prompt using
+    the stable cache key from the shared state."""
+
+    _is_stateful = True
+
+    def __init__(self, state_manager: StateManager):
+        super().__init__()
+        self.state_manager = state_manager
+        self.kv_cache: dict[int, tuple[torch.Tensor, torch.Tensor]] = {}
```
dg845: After looking at the existing cache design more closely, I think `kv_cache` should not be owned by `TextKVCacheBlockHook` but rather refactored into its own `BaseState` subclass, which is what MagCache and TaylorSeer do (both store cached tensors in their state classes rather than on the hook).

This would mean we would have two state classes: one which holds the shared `encoder_hidden_states` tensor and one which holds the KV cache dict for each block:

```python
# Same as before
class TextKVCacheState(BaseState):
    def __init__(self):
        self.key: int | None = None
    ...

# Holds the block-level KV cache
class TextKVCacheBlockState(BaseState):
    def __init__(self):
        self.kv_cache: dict[int, tuple[torch.Tensor, torch.Tensor]] = {}
    ...

# Same as before
class TextKVCacheTransformerHook(ModelHook):
    ...

class TextKVCacheBlockHook(ModelHook):
    # One state manager for shared transformer-level state, one for block-specific state
    def __init__(self, state_manager: StateManager, block_state_manager: StateManager):
        super().__init__()
        self.state_manager = state_manager
        self.block_state_manager = block_state_manager
    ...
```

This would allow us to manage each block-level KV Cache with a `StateManager`, which I think more cleanly follows the current design.

sippycoder: Yea that makes sense! I added a commit for this.

dg845:
Merging as the CI is green.

q5sys:
> Hey @dg845 @yiyixuxu , we are releasing the model report tomorrow! Thanks for all the reviews! I think Expert Parallelism can be a separate PR. Any chance we can merge the PR today?

@sippycoder When is this actually getting released? You said 'tomorrow' two weeks ago to get this PR merged, but your model still isn't released. I'm eager to try this. 

Edit: it was posted the day after my comment

PR #13329, opened 2026-03-25
[ci] simplify release workflow.

sayakpaul:
# What does this PR do?

Simplify release workflow like https://github.com/huggingface/transformers/blob/main/.github/workflows/release.yml.

- Removed Test PyPI — no longer publish to Test PyPI and install from there; instead, install directly from the locally built wheel to validate.             
- Two-job structure: `build-and-test` (find release branch, build, install from wheel, run import tests) -> `publish-to-pypi` (download artifacts, publish).    
- Manual approval gate via GitHub Environment `pypi-release` with required reviewers — publishing pauses until a maintainer approves.
- OIDC trusted publishing using `pypa/gh-action-pypi-publish` — replaces Twine username/password secrets.                                   
- Fixed deprecated `::set-output` — no longer needed since branch resolution happens in the same job.                                       

I have taken care of the following config setups to make it work:

- Create a GitHub Environment called `pypi-release` with required reviewers (Dhruv, Yiyi, and myself).
- Configure PyPI trusted publishing for the `huggingface/diffusers` repo, specifying the `pypi-release` environment.


.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,72 @@
...
-      - name: Install dependencies
+      - name: Install build dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -U setuptools wheel twine
+          pip install -U setuptools wheel
           pip install -U torch --index-url https://download.pytorch.org/whl/cpu
 
       - name: Build the dist files
         run: python setup.py bdist_wheel && python setup.py sdist
```
DN6: Small nit. Running commands like this with setup.py is deprecated in favour of just running a single 
`python -m build` which gives you both the sdist and the wheel
https://setuptools.pypa.io/en/latest/deprecated/commands.html#running-setuptools-commands

Think we can just use that here.

sayakpaul: Oh nice one. Updated.

sayakpaul:
@DN6 a gentle ping.

.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,72 @@
...
 jobs:
-  find-and-checkout-latest-branch:
+  build-and-test:
     runs-on: ubuntu-22.04
-    outputs:
-      latest_branch: ${{ steps.set_latest_branch.outputs.latest_branch }}
     steps:
-      - name: Checkout Repo
+      - name: Checkout repo
         uses: actions/checkout@v6
```
DN6, line 14: Pin to SHA
https://github.com/huggingface/transformers/blob/e737fb8671d903d5128d673abca76c1a26b4fa60/.github/workflows/release.yml#L14-L17

.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,72 @@
...
+        with:
+          name: python-dist
+          path: dist/
+
       - name: Publish to PyPI
-        env:
-          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
-          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-        run: twine upload dist/* -r pypi
+        uses: pypa/gh-action-pypi-publish@release/v1
```
DN6, line 72: Pin to SHA
https://github.com/huggingface/transformers/blob/e737fb8671d903d5128d673abca76c1a26b4fa60/.github/workflows/release.yml#L62

.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,72 @@
-# Adapted from https://blog.deepjyoti30.dev/pypi-release-github-action
-
 name: PyPI release
 
 on:
   workflow_dispatch:
   push:
     tags:
-      - "*"
+      - "v*"
```
DN6, line 8: Would add 

```yaml 
  permissions:
       contents: read
```

.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,72 @@
...
+          python-version: "3.10"
 
-      - name: Fetch latest branch
-        id: fetch_latest_branch
+      - name: Fetch and checkout latest release branch
         run: |
           pip install -U requests packaging
           LATEST_BRANCH=$(python utils/fetch_latest_release_branch.py)
           echo "Latest branch: $LATEST_BRANCH"
-          echo "latest_branch=$LATEST_BRANCH" >> $GITHUB_ENV
```
DN6, line 31: Don't think this is needed since the checkout step would just fetch the tagged commit? 

See transformers:
https://github.com/huggingface/transformers/blob/e737fb8671d903d5128d673abca76c1a26b4fa60/.github/workflows/release.yml#L14

.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,72 @@
...
+      - name: Install build dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -U setuptools wheel twine
+          pip install -U build
           pip install -U torch --index-url https://download.pytorch.org/whl/cpu
 
       - name: Build the dist files
-        run: python setup.py bdist_wheel && python setup.py sdist
+        run: python -m build
```
DN6, line 37: Let's keep the metadata check for the dist with twine

```yaml
  - name: Validate dist metadata
    run: |
      pip install twine
      twine check --strict dist/*
```

.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,72 @@
-# Adapted from https://blog.deepjyoti30.dev/pypi-release-github-action
-
 name: PyPI release
 
 on:
   workflow_dispatch:
   push:
     tags:
-      - "*"
+      - "v*"
```
DN6, line 7: Would be good to be able to run on the branch as well without having to tag. 
https://github.com/huggingface/transformers/blob/e737fb8671d903d5128d673abca76c1a26b4fa60/.github/workflows/release.yml#L3-L7

sayakpaul:
@DN6 all your comments should have been addressed. Check now.

.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,78 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      
+      - name: Download build artifacts
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093  # v4
+        with:
+          name: python-dist
+          path: dist/
+
+      - name: Publish package distributions to TestPyPI
```
DN6, line 75:
```suggestion
      - name: Publish package distributions to PyPI
```

.github/workflows/pypi_publish.yaml
```diff
@@ -1,82 +1,78 @@
...
 name: PyPI release
 
 on:
   workflow_dispatch:
   push:
     tags:
-      - "*"
+      - v*
+    branches:
+      - 'v*-release'
```
DN6, line 9: Just realised we use v*-patch for patch branch names. Not a merge blocker, I would actually prefer it if we just called everything v*-release like transformers moving forward.

DN6 (approved):
Nice 👍🏽

sayakpaul:
Failing tests are so unrelated!

PR #13291, opened 2026-03-19
remove str option for quantization config in torchao

howardzhang-cv:
# What does this PR do?

Remove the deprecated string-based quant_type path from TorchAoConfig, requiring AOBaseConfig instances instead.

- TorchAoConfig.__init__ now only accepts AOBaseConfig subclass instances (e.g. Int8WeightOnlyConfig()) and raises TypeError for strings
- Deleted ~200 lines of dead code: _get_torchao_quant_type_to_method, _is_xpu_or_cuda_capability_atleast_8_9, TorchAoJSONEncoder, and all string-parsing branches in post_init, to_dict, from_dict, get_apply_tensor_subclass
- Simplified torchao_quantizer.py: removed string-based branches in update_torch_dtype, adjust_target_dtype, get_cuda_warm_up_factor; fixed is_trainable which would crash on AOBaseConfig objects
- Converted all test cases from string quant types to their AOBaseConfig equivalents; removed test_floatx_quantization (no replacement for fpx_weight_only)
- Updated docs to show only AOBaseConfig-based usage

## Testing
`python -m pytest tests/quantization/torchao/test_torchao.py -xvs`

## Who can review?

@sayakpaul

PR #13330, opened 2026-03-25
[tests] fix torchao tests

sayakpaul:
# What does this PR do?

Surfaced in https://github.com/huggingface/diffusers/pull/13291#issuecomment-4122867870. Cc: @howardzhang-cv


tests/models/testing_utils/quantization.py
```diff
@@ -177,6 +177,11 @@ def _test_quantization_inference(self, config_kwargs):
         model_quantized.to(torch_device)
 
         inputs = self.get_dummy_inputs()
+        model_dtype = next(model_quantized.parameters()).dtype
```
DN6: This would affect all quantization backends? e.g. With a GGUF backend the dtype could end up as int8 and potentially cast inputs into int8? 

Also prefer to avoid casting inputs post fetching from `self.get_dummy_inputs()` within a test if we can avoid it.

sayakpaul: Where should it go then? Should we implement a custom `get_dummy_inputs()` for torchao tests? I think it's reasonably safe to keep the dtypes of the inputs to bfloat16 there because that will replicate what we do in actual pipelines. LMK.

sayakpaul:
@dn6 I have adjusted the PR to match https://github.com/huggingface/diffusers/pull/13481. LMK.

sayakpaul:
@DN6 a gentle ping.

PR #13332, opened 2026-03-25
[tests] tighten dependency testing.

sayakpaul:
# What does this PR do?

We should be able to catch stuff like https://github.com/huggingface/diffusers/pull/13321.

Additionally, I ported the existing dependency test suite to use `pytest`.


.github/workflows/pr_dependency_test.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
     paths:
       - "src/diffusers/**.py"
+      - "tests/**.py"
```
sayakpaul, line 9: I initially did it to trigger the tests in this PR. Will remove before merging.

sayakpaul:
Failing tests are unrelated.

PR #13334, opened 2026-03-25
[docs] refactor model skill

stevhliu:
follow up to #13297:

- created `models.md`, a shared reference for holding model coding style, dependencies, common conventions, attention patterns, gotchas, etc.
- updated `AGENTS.md`, `model-integration/SKILL.md`, and `review-rules.md` to point to new `models.md`

### DIFF
--- a/.github/workflows/claude_review.yml
+++ b/.github/workflows/claude_review.yml
@@ -32,6 +32,9 @@ jobs:
       )
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
       - uses: anthropics/claude-code-action@v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
--- a/docs/source/en/api/pipelines/cogvideox.md
+++ b/docs/source/en/api/pipelines/cogvideox.md
@@ -41,16 +41,15 @@ The quantized CogVideoX 5B model below requires ~16GB of VRAM.
 
 ```py
 import torch
-from diffusers import CogVideoXPipeline, AutoModel
+from diffusers import CogVideoXPipeline, AutoModel, TorchAoConfig
 from diffusers.quantizers import PipelineQuantizationConfig
 from diffusers.hooks import apply_group_offloading
 from diffusers.utils import export_to_video
+from torchao.quantization import Int8WeightOnlyConfig
 
 # quantize weights to int8 with torchao
 pipeline_quant_config = PipelineQuantizationConfig(
-  quant_backend="torchao",
-  quant_kwargs={"quant_type": "int8wo"},
-  components_to_quantize="transformer"
+  quant_mapping={"transformer": TorchAoConfig(Int8WeightOnlyConfig())}
 )
 
 # fp8 layerwise weight-casting
--- a/docs/source/en/quantization/torchao.md
+++ b/docs/source/en/quantization/torchao.md
@@ -29,24 +29,7 @@ from diffusers import DiffusionPipeline, PipelineQuantizationConfig, TorchAoConf
 from torchao.quantization import Int8WeightOnlyConfig
 
 pipeline_quant_config = PipelineQuantizationConfig(
-    quant_mapping={"transformer": TorchAoConfig(Int8WeightOnlyConfig(group_size=128)))}
-)
-pipeline = DiffusionPipeline.from_pretrained(
-    "black-forest-labs/FLUX.1-dev",
-    quantization_config=pipeline_quant_config,
-    torch_dtype=torch.bfloat16,
-    device_map="cuda"
-)
-```
-
-For simple use cases, you could also provide a string identifier in [`TorchAo`] as shown below.
-
-```py
-import torch
-from diffusers import DiffusionPipeline, PipelineQuantizationConfig, TorchAoConfig
-
-pipeline_quant_config = PipelineQuantizationConfig(
-    quant_mapping={"transformer": TorchAoConfig("int8wo")}
+    quant_mapping={"transformer": TorchAoConfig(Int8WeightOnlyConfig(group_size=128, version=2))}
 )
 pipeline = DiffusionPipeline.from_pretrained(
     "black-forest-labs/FLUX.1-dev",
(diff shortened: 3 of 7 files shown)


.ai/models.md
```diff
@@ -0,0 +1,88 @@
...
+
+    def forward(self, hidden_states, attention_mask=None, **kwargs):
+        return self.processor(self, hidden_states, attention_mask, **kwargs)
+```
+
+Consult the implementations in `src/diffusers/models/transformers/` if you need further references.
+
+## Implementation rules
+
+1. **Don't combine structural changes with behavioral changes.** Restructuring code to fit diffusers APIs (ModelMixin, ConfigMixin, etc.) is unavoidable. But don't also "improve" the algorithm, refactor computation order, or rename internal variables for aesthetics. Keep numerical logic as close to the reference as possible, even if it looks unclean. For standard → modular, this is stricter: copy loop logic verbatim and only restructure into blocks. Clean up in a separate commit after parity is confirmed.
```
yiyixuxu: this rule is specifically written for agent for that skill - I find that it'd be easier/better to instruct them implement a first draft with minimu code style changes and iterate based on human feedbacks from there 

we should not add here because it might be a bit misleading for PR reviews etc, the final model implementation should have been sufficiently refactored

yiyixuxu (approved):
I left one feedback
looking great!

.ai/models.md
stevhliu: ah makes sense, thanks! moved this rule back to the model integration skill :)

.ai/skills/model-integration/SKILL.md
```diff
@@ -65,89 +65,19 @@ docs/source/en/api/
...
-    def forward(self, hidden_states, attention_mask=None, **kwargs):
-        return self.processor(self, hidden_states, attention_mask, **kwargs)
-```
+### Model conventions, attention pattern, and implementation rules
 
-Consult the implementations in `src/diffusers/models/transformers/` if you need further references.
+See [../../models.md](../../models.md) for the attention pattern, implementation rules, common conventions, dependencies, and gotchas. These apply to all model work.
 
-### Implementation rules
+### Implementation rule
```
dg845: Do you think it would make sense to rename this section to something like `Model Integration Specific Rules`? I think this would help differentiate it from the `Implementation Rules` in `models.md`.

sayakpaul: Strong agree!

.ai/skills/model-integration/SKILL.md
```diff
@@ -65,89 +65,19 @@ docs/source/en/api/
...
-
-- Pipelines inherit from `DiffusionPipeline`
-- Models use `ModelMixin` with `register_to_config` for config serialization
-- Schedulers use `SchedulerMixin` with `ConfigMixin`
-- Use `@torch.no_grad()` on pipeline `__call__`
-- Support `output_type="latent"` for skipping VAE decode
-- Support `generator` parameter for reproducibility
-- Use `self.progress_bar(timesteps)` for progress tracking
-
-## Gotchas
```
sayakpaul, line 133: Same.

.ai/AGENTS.md
```diff
@@ -23,11 +19,10 @@ Strive to write code as simple and explicit as possible.
...
-- Avoid graph breaks for `torch.compile` compatibility — do not insert NumPy operations in forward implementations and any other patterns that can break `torch.compile` compatibility with `fullgraph=True`.
-- See the **model-integration** skill for the attention pattern, pipeline rules, test setup instructions, and other important details.
+- See [models.md](models.md) for conventions, attention pattern, implementation rules, dependencies, and gotchas.
+- See the [model-integration](./skills/model-integration/SKILL.md) skill for the full integration workflow, file structure, test setup, and other details.
 
 ## Skills
 
 Task-specific guides live in `.ai/skills/` and are loaded on demand by AI agents.
-Available skills: **model-integration** (adding/converting pipelines), **parity-testing** (debugging numerical parity).
+Available skills: [model-integration](./skills/model-integration/SKILL.md) (adding/converting pipelines), [parity-testing](./skills/parity-testing/SKILL.md) (debugging numerical parity).
```
sayakpaul: Should this be bulleted?

.ai/AGENTS.md
```diff
@@ -23,11 +19,10 @@ Strive to write code as simple and explicit as possible.
 - Remove the header to intentionally break the link
 
 ### Models
-- All layer calls should be visible directly in `forward` — avoid helper functions that hide `nn.Module` calls.
-- Avoid graph breaks for `torch.compile` compatibility — do not insert NumPy operations in forward implementations and any other patterns that can break `torch.compile` compatibility with `fullgraph=True`.
-- See the **model-integration** skill for the attention pattern, pipeline rules, test setup instructions, and other important details.
+- See [models.md](models.md) for conventions, attention pattern, implementation rules, dependencies, and gotchas.
```
sayakpaul: (nit): the implementation rules (sone of them) and gotchas go beyond just models as mentioned before.

```suggestion
- See [models.md](models.md) for guidance conventions, attention pattern, implementation rules, dependencies, and gotchas.
```

.ai/models.md
```diff
@@ -0,0 +1,87 @@
...
+## Coding style
+
+- All layer calls should be visible directly in `forward` — avoid helper functions that hide `nn.Module` calls.
+- Avoid graph breaks for `torch.compile` compatibility — do not insert NumPy operations in forward implementations and any other patterns that can break `torch.compile` compatibility with `fullgraph=True`.
+- No new mandatory dependency without discussion (e.g. `einops`). Optional deps guarded with `is_X_available()` and a dummy in `utils/dummy_*.py`.
+
+## Common diffusers conventions
+
+- Models use `ModelMixin` with `register_to_config` for config serialization
+- Pipelines inherit from `DiffusionPipeline`
```
sayakpaul: It's not a model thing, though.

.ai/AGENTS.md
stevhliu: thanks for clarifying!

.ai/models.md
stevhliu: lmk what you think of this:

- rename to `## Common model conventions` and only keep the model relevant convention
- move these conventions under a new `### Pipelines & Schedulers` section in `AGENTS.md` (we could also create a new `pipeline-scheduler.md` file if you prefer)

sayakpaul: Since the specifics for pipeline and schedule seem relatively small, let's keep them under AGENTS.md but I think it might also make sense to keep them under a separate MD fine. So, let's go with that approach?

stevhliu: sounds good to me! we can keep it under `AGENTS.md` for now and move it to a separate file if it gets larger :)

PR #13339, opened 2026-03-26
Fix: ensure consistent dtype and eval mode in pipeline save/load tests

YangKai0616:
# Issue:
The `test_save_load_float16` test has a hidden non-determinism problem:

1. When the pipeline is constructed directly, its components default to training mode (e.g., `text_encoder` contains `dropout` layers). However, after loading with `from_pretrained`, the components are in eval mode.
2. Certain buffers (such as `blur_kernel` in `MotionConv2d`) may have inconsistent dtypes during the save/load process, leading to inconsistent inference results.

Although the test usually "luckily" passes on CUDA (because dropout errors are diluted by numerical propagation), this is fundamentally a correctness issue in both the test design and the implementation.

# Fix:

1. Set the `text_encoder` component in the pipeline to eval mode, to be consistent with the handling in other `test_save_loadXX` tests such as [test_save_load_local](https://github.com/huggingface/diffusers/blob/41e1003316173218656998281b28f9e0f6fcbcff/tests/pipelines/test_pipelines_common.py#L1161).
2. In the forward pass, automatically cast the `blur_kernel` to the input dtype. This ensures consistency before and after save/load operations and avoids dtype mismatches. (`blur_kernel` is essentially a fixed mathematical constant, not a learnable parameter. It should adapt to the dtype of the computation stream, rather than forcing the computation stream's data to adapt to it.)

Hi @sayakpaul , please help review, thanks!

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -161,6 +161,8 @@
     - local: training/ddpo
       title: Reinforcement learning training with DDPO
     title: Methods
+  - local: training/nemo_automodel
+    title: NeMo Automodel
   title: Training
 - isExpanded: false
   sections:
--- a/docs/source/en/training/nemo_automodel.md
+++ b/docs/source/en/training/nemo_automodel.md
@@ -1,0 +1,378 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+-->
+
+# NeMo Automodel
+
+[NeMo Automodel](https://github.com/NVIDIA-NeMo/Automodel) is a PyTorch DTensor-native training library from NVIDIA for fine-tuning and pretraining diffusion models at scale. It is Hugging Face native — train any Diffusers-format model from the Hub with no checkpoint conversion. The same YAML recipe and hackable training script runs on any scale from 1 GPU to hundreds of nodes, with [FSDP2](https://pytorch.org/docs/stable/fsdp.html) distributed training, multiresolution bucketed dataloading, and pre-encoded latent space training for maximum GPU utilization. It uses [flow matching](https://huggingface.co/papers/2210.02747) for training and is fully open source (Apache 2.0), NVIDIA-supported, and actively maintained.
+
+NeMo Automodel integrates directly with Diffusers. It loads pretrained models from the Hugging Face Hub using Diffusers model classes and generates outputs with the [`DiffusionPipeline`].
+
+The typical workflow is to install NeMo Automodel (pip or Docker), prepare your data by encoding it into `.meta` files, configure a YAML recipe, launch training with `torchrun`, and run inference with the resulting checkpoint.
+
+## Supported models
+
+| Model | Hugging Face ID | Task | Parameters | Use case |
+|-------|----------------|------|------------|----------|
+| Wan 2.1 T2V 1.3B | [Wan-AI/Wan2.1-T2V-1.3B-Diffusers](https://huggingface.co/Wan-AI/Wan2.1-T2V-1.3B-Diffusers) | Text-to-Video | 1.3B | video generation on limited hardware (fits on single 40GB A100) |
+| FLUX.1-dev | [black-forest-labs/FLUX.1-dev](https://huggingface.co/black-forest-labs/FLUX.1-dev) | Text-to-Image | 12B | high-quality image generation |
+| HunyuanVideo 1.5 | [hunyuanvideo-community/HunyuanVideo-1.5-Diffusers-720p_t2v](https://huggingface.co/hunyuanvideo-community/HunyuanVideo-1.5-Diffusers-720p_t2v) | Text-to-Video | 13B | high-quality video generation |
+
+## Installation
+
+### Hardware requirements
+
+| Component | Minimum | Recommended |
+|-----------|---------|-------------|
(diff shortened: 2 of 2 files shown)


src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -166,8 +166,7 @@ def forward(self, x: torch.Tensor, channel_dim: int = 1) -> torch.Tensor:
             # NOTE: the original implementation uses a 2D upfirdn operation with the upsampling and downsampling rates
             # set to 1, which should be equivalent to a 2D convolution
             expanded_kernel = self.blur_kernel[None, None, :, :].expand(self.in_channels, 1, -1, -1)
-            x = x.to(expanded_kernel.dtype)
-            x = F.conv2d(x, expanded_kernel, padding=self.blur_padding, groups=self.in_channels)
+            x = F.conv2d(x, expanded_kernel.to(x.dtype), padding=self.blur_padding, groups=self.in_channels)
```
sayakpaul: How is this relevant?

tests/pipelines/test_pipelines_common.py
```diff
@@ -1447,6 +1447,10 @@ def test_save_load_float16(self, expected_max_diff=1e-2):
             elif hasattr(module, "half"):
                 components[name] = module.to(torch_device).half()
 
+            for key in components:
+                if "text_encoder" in key and hasattr(components[key], "eval"):
+                    components[key].eval()
```
sayakpaul: We can safely apply eval if the component contains the `eval` attr.

src/diffusers/models/transformers/transformer_wan_animate.py
YangKai0616: `blur_kernel` is registered with [`persistent=False`](https://github.com/huggingface/diffusers/blob/1fe2125802ac656c00d7f087869d0e4c15886d8f/src/diffusers/models/transformers/transformer_wan_animate.py#L142), so it's not saved into the checkpoint. After `from_pretrained`, it's always re-initialized in float32, while before saving it was float16 (due to [`.half()`](https://github.com/huggingface/diffusers/blob/1fe2125802ac656c00d7f087869d0e4c15886d8f/tests/pipelines/test_pipelines_common.py#L1445)). This dtype mismatch causes the two inference runs to take different numerical paths. 

We can print `expanded_kernel.dtype` in the test `tests/pipelines/wan/test_wan_animate.py::WanAnimatePipelineFastTests::test_save_load_float16` — it outputs float16 before save and float32 after load.

tests/pipelines/test_pipelines_common.py
YangKai0616: Done.

src/diffusers/models/transformers/transformer_wan_animate.py
sayakpaul: Then I propose we `xfail` that test under test suite of Wan Animate and tackle it in a separate PR. We might want to use `modules_to_not_convert` or something.

YangKai0616: Done.

In addition, regarding the modification of the `blur_kernel`, I think that since it is a fixed mathematical constant, it is reasonable to directly adjust the `dtype` to fit the computational flow. What do you think? If you agree, I will submit a new PR to modify this part and restore the `xfail` in the test.

dg845: Arguably the call to `module.named_parameters()` below in `test_save_load_float16` is too restrictive, as `named_parameters()` will not return buffers:

https://github.com/huggingface/diffusers/blob/1fe2125802ac656c00d7f087869d0e4c15886d8f/tests/pipelines/test_pipelines_common.py#L1438-L1445

Should we modify the test to also cast buffers to `torch.float16`?

dg845: I agree that calling `expanded_kernel.to(x.dtype)` is probably better than `x = x.to(expanded_kernel.dtype)` in the Wan Animate code above. I also agree with https://github.com/huggingface/diffusers/pull/13339#discussion_r2999256730 that any changes to the Wan Animate code should be spun off into a separate PR, and this PR should focus on changes to `test_save_load_float16`.

YangKai0616: > Arguably the call to `module.named_parameters()` below in `test_save_load_float16` is too restrictive, as `named_parameters()` will not return buffers:
> 
> Should we modify the test to also cast buffers to `torch.float16`?

Hey @dg845 , you are right. I modified the test code to cast all buffers to `torch.float16` as well. Except for `rope`, because `rope` [here](https://github.com/huggingface/diffusers/blob/f2be8bd6b3dc4035bd989dc467f15d86bf3c9c12/src/diffusers/models/transformers/transformer_wan_animate.py#L863) is not affected by `from_pretrained(tmpdir, torch_dtype=torch.float16)`.

YangKai0616: > I agree that calling `expanded_kernel.to(x.dtype)` is probably better than `x = x.to(expanded_kernel.dtype)` in the Wan Animate code above. I also agree with [#13339 (comment)](https://github.com/huggingface/diffusers/pull/13339#discussion_r2999256730) that any changes to the Wan Animate code should be spun off into a separate PR, and this PR should focus on changes to `test_save_load_float16`.

I have already made the modification here in the separate PR https://github.com/huggingface/diffusers/pull/13364.

tests/pipelines/test_pipelines_common.py
```diff
@@ -1443,10 +1443,24 @@ def test_save_load_float16(self, expected_max_diff=1e-2):
                         param.data = param.data.to(torch_device).to(torch.float32)
                     else:
                         param.data = param.data.to(torch_device).to(torch.float16)
+                for name, buf in module.named_buffers():
+                    if not buf.is_floating_point():
+                        buf.data = buf.data.to(torch_device)
```
sayakpaul, line 1448: How is this condition relevant for a device placement?

dg845: I think the idea is that if the buffer holds non-floating point (e.g. int or long) dtype data, we still move it to the test device but don't attempt a cast.

sayakpaul:
Thanks for your contributions!

PR #13353, opened 2026-03-27
Fix BucketBatchSampler cache alignment in DreamBooth scripts

azolotenkov:
## What does this PR do?

This PR fixes a bug where `BucketBatchSampler` reshuffled precomputed batches on each `__iter__()` call.

DreamBooth training scripts precompute latents and/or prompt embeddings and later consume them by dataloader step index, so changing batch order between the caching pass and the training pass can misalign the cached tensors with the current batch.

This PR removes the per-iteration `random.shuffle(self.batches)` call from `BucketBatchSampler.__iter__()` and instead shuffles the precomputed batch list once at sampler construction time, keeping the batch order fixed across epochs.

A future rework may still be needed to support true epoch-wise reshuffling, since batch membership is currently fixed once at sampler construction time.

Applied to:
- `examples/dreambooth/train_dreambooth_lora_flux2.py`
- `examples/dreambooth/train_dreambooth_lora_flux2_img2img.py`
- `examples/dreambooth/train_dreambooth_lora_flux2_klein.py`
- `examples/dreambooth/train_dreambooth_lora_flux2_klein_img2img.py`
- `examples/dreambooth/train_dreambooth_lora_z_image.py`

## Before submitting

## Who can review?

Training examples: @sayakpaul


examples/dreambooth/train_dreambooth_lora_flux2_img2img.py
```diff
@@ -1003,8 +1003,7 @@ def __init__(self, dataset: DreamBoothDataset, batch_size: int, drop_last: bool
                 self.sampler_len += 1  # Count the number of batches
```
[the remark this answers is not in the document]
azolotenkov: Good point. I updated the sampler to shuffle `self.batches` once at construction time instead of on every `__iter__()` call.

This keeps the order fixed across epochs for the step-indexed caches, while still preserving one-time global mixing across buckets.

azolotenkov:
Updated based on review: the sampler now shuffles precomputed batches once at construction time, while keeping iteration order fixed across epochs for cache alignment.

linoytsaban:
good catch @azolotenkov! 
I think it would make more sense if we apply this fix only when `--cache_latents` is enabled, since in other cases it's not issue. In light of this bug, ideally it would be best to support per-epoch reshuffle and latent caching but we can leave that for a follow up PR imo.  
@sayakpaul wdyt?

sayakpaul:
@linoytsaban yes, that would be the most appropriate!

azolotenkov:
Updated. I kept the per-epoch reshuffle in the normal path, and only switch to a fixed precomputed batch order when the script relies on step-indexed caches. I scoped that to `args.cache_latents` or `train_dataset.custom_instance_prompts`, since custom prompt embedding caching has the same alignment requirement here. I also pulled that condition into a separate `has_step_indexed_caches` variable since it’s used both for the sampler behavior and for the caching path.
@linoytsaban wdyt?

azolotenkov:
Friendly ping @sayakpaul @linoytsaban
I have already a follow up prepared PR to support per-epoch reshuffle and latent caching.

examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -1585,8 +1603,7 @@ def _encode_single(prompt: str):
     # if cache_latents is set to True, we encode images to latents and store them.
     # Similar to pre-encoding in the case of a single instance prompt, if custom prompts are provided
     # we encode them in advance as well.
-    precompute_latents = args.cache_latents or train_dataset.custom_instance_prompts
-    if precompute_latents:
```
sayakpaul, line 1589: I think `precompute_latents` is a better variable name here. So, perhaps we could do:

```py
has_step_indexed_caches = precompute_latents = args.cache_latents or train_dataset.custom_instance_prompts
```

sayakpaul (commented):
Thanks! Just one comment.

examples/dreambooth/train_dreambooth_lora_flux2.py
azolotenkov: Updated, thanks. I restored the `precompute_latents` name for the caching path while keeping `has_step_indexed_caches` for the sampler behavior:

`has_step_indexed_caches = precompute_latents = args.cache_latents or train_dataset.custom_instance_prompts`

sayakpaul (approved):
Thanks!

azolotenkov:
Friendly ping @sayakpaul

sayakpaul (approved):

PR #13356, opened 2026-03-28
Add examples on how to profile a pipeline

sayakpaul:
# What does this PR do?

TL;DR: Adds a guide on how to profile a pipeline and fix issues like CPU overhead, CPU<->GPU syncs, etc. 

## Motivation 

Since we provide first-class `torch.compile` support, it's important that our pipelines are set up for optimal success with it. This includes spotting any obvious issues that plague the `torch.compile` performance -- CPU overhead, CPU<->GPU syncs, graphbreaks, kernel launch delays, etc.

The best way to spot these bugs is to profile a pipeline, as it gives a granular measurement of where the GPU is spending time and if it is doing so in an expected manner. We can then uncover any unexpected issues and eventually fix them.

## Workflow

The README.md added in the PR has all the descriptions, but in summary:

* take a popular pipeline like Flux/Flux2/QwenImage/Wan/LTX2
* run the profile with 2 inference steps
* load the trace on [Perfetto](https://ui.perfetto.dev/)
* spot the potential suspects 
* piggy that back to Claude along with the trace
  * ask it to attempt a fix
  * review the fix
  * compare the results

With this Workflow, I was able to fix some issues in the Flux2 Klein pipeline and the Wan pipeline. All changes look quite harmless to me.

## Plan

Not only is it helpful to profile pipelines to get a ceiling on performance, but the community could also help us improve our pipelines should this workflow prove to be useful.

## Note to reviewers

Please review the changes in `src/diffusers/*`. And you can skip straight to the "Afterwards" section in the README.md document.

The tutorial is currently available [here](https://github.com/huggingface/diffusers/tree/profiling-workflow/examples/profiling). Some inline comments.


examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+| `FluxPipeline` | text-to-image | `black-forest-labs/FLUX.1-dev` | 2 |
+| `Flux2KleinPipeline` | text-to-image | `black-forest-labs/FLUX.2-klein-base-9B` | 2 |
+| `WanPipeline` | text-to-video | `Wan-AI/Wan2.1-T2V-14B-Diffusers` | 2 |
+| `LTX2Pipeline` | text-to-video | `Lightricks/LTX-2` | 2 |
+| `QwenImagePipeline` | text-to-image | `Qwen/Qwen-Image` | 2 |
+
+> [!NOTE]
+> We use realistic inference call hyperparameters that mimic how these pipelines will be actually used. This
+> include using classifier-free guidance (where applicable), reasonable dimensions such 1024x1024, etc.
+> But we keep the overall running time to a bare minimum (hence 2 `num_inference_steps`).
```
stevhliu:
```suggestion
> But we keep the number of inference steps to a bare minimum. 
```

examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+   - `transformer_forward` and `scheduler_step` annotations visible
+   - CPU and CUDA timelines present
+   - Stack traces visible on CPU events
+4. Run with `--mode compile` and compare trace for fewer/fused CUDA kernels
+
+You can also use the `run_profiling.sh` script to bulk launch runs for different pipelines.
+
+## Interpreting Traces in Perfetto UI
+
+Open the exported `.json` trace at [ui.perfetto.dev](https://ui.perfetto.dev/). The trace has two main rows: **CPU** (top) and **CUDA** (bottom). In Perfetto, the CPU row is typically labeled with the process/thread name (e.g., `python (PID)` or `MainThread`) and appears at the top. The CUDA row is labeled `GPU 0` (or similar) and appears below the CPU rows.
```
stevhliu, line 148: for each of the issues below, it may be nice to include a screenshot of what it looks like

examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+**6. Kernel launch latency**
+
+Each CUDA kernel is launched from the CPU. The CPU-side launch calls (`cudaLaunchKernel`) appear as small slices on the **CPU row** — zoom in closely to a denoising step to see them. The corresponding GPU-side kernel executions appear on the **CUDA row** directly below. You can also use Perfetto's search bar (`/`) and type `cudaLaunchKernel` to find them. The time between the CPU dispatch and the GPU kernel starting should be minimal (single-digit microseconds). If you see consistent delays > 10-20us between launch and execution:
+- The launch queue may be starved because of excessive Python work between ops
+- There may be implicit syncs forcing serialization
+- `torch.compile` should help here by batching launches — compare eager vs compile to confirm
+
+To inspect this: zoom into a single denoising step, select a CUDA kernel on the GPU row, and look at the corresponding CPU-side launch slice directly above it. The horizontal offset between them is the launch latency. In a healthy trace, CPU launch slices should be well ahead of GPU execution (the CPU is "feeding" the GPU faster than it can consume).
+
+### Quick checklist per pipeline
```
stevhliu, line 199: very helpful!

examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+| GPU staying busy? | CUDA row gaps | Back-to-back kernels | Frequent gaps > 100us |
+| CPU blocking on GPU? | `cudaStreamSynchronize` slices | Rare/absent during denoise | Present every step |
+| Scheduler overhead? | `scheduler_step` span duration | < 1% of step time | > 5% of step time |
+| Compile effective? | CUDA kernel count per step | Fewer large kernels | Same as eager |
+| Kernel launch latency? | CPU launch → GPU kernel offset | < 10us, CPU ahead of GPU | > 20us or CPU trailing GPU |
+| Memory stable? | Memory counter track | Flat during denoise loop | Growing per step |
+
+## Afterwards
+
+To keep the profiling iterations fast, we always used [regional compilation](https://pytorch.org/tutorials/recipes/regional_compilation.html). As one would expect the trace with compilation should show
```
stevhliu:
```suggestion
To keep the profiling iterations fast, we always use [regional compilation](https://pytorch.org/tutorials/recipes/regional_compilation.html). As one would expect the trace with compilation should show
```

examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+
+Before:                                                                                                         
+- _set_context total: 21.6ms (8 calls)                                                                          
+- cache_context total: 21.7ms                                                                                   
+- CPU gaps: 5,523us / 8,007us / 5,508us                                                                         
+                                                                                                            
+After:                                                                                                          
+- _set_context total: 0.0ms (8 calls)                                                                           
+- cache_context total: 0.1ms                                                                                    
+- CPU gaps: 158us / 2,777us / 136us
```
stevhliu: this may be more effective as a table for better side-by-side comparison

examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+| Question | Where to look | Healthy | Unhealthy |
+|----------|--------------|---------|-----------|
+| GPU staying busy? | CUDA row gaps | Back-to-back kernels | Frequent gaps > 100us |
+| CPU blocking on GPU? | `cudaStreamSynchronize` slices | Rare/absent during denoise | Present every step |
+| Scheduler overhead? | `scheduler_step` span duration | < 1% of step time | > 5% of step time |
+| Compile effective? | CUDA kernel count per step | Fewer large kernels | Same as eager |
+| Kernel launch latency? | CPU launch → GPU kernel offset | < 10us, CPU ahead of GPU | > 20us or CPU trailing GPU |
+| Memory stable? | Memory counter track | Flat during denoise loop | Growing per step |
+
+## Afterwards
```
stevhliu: this section reads a bit more like a story, which is a bit different from the rest of this doc. maybe consider restructuring it as "here's what profiling revealed and how to fix it" rather than "here's how we paired with Claude"

examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+Issue 2 — cudaStreamSynchronize during last transformer forward:
+- Root cause: _unpack_latents_with_ids() (called right after the denoising loop) computes h = torch.max(h_ids) +
+1 and w = torch.max(w_ids) + 1 on GPU tensors, then uses them as shape args for torch.zeros((h * w, ch), ...).
+This triggers an implicit .item() DtoH sync, blocking the CPU while the GPU is still finishing the last
+transformer forward's kernels.
+- Fix: Added height/width parameters to _unpack_latents_with_ids(), pre-computed from the known pixel dimensions
+at the call site.
+```
+
+The changes looke reasonable based on our past experience. So, we asked Claude to apply these changes to [`pipeline_flux2_klein.py`](../../src/diffusers/pipelines/flux2/pipeline_flux2_klein.py). We then profiled
```
stevhliu:
```suggestion
The changes looked reasonable based on our past experience. So, we asked Claude to apply these changes to [`pipeline_flux2_klein.py`](../../src/diffusers/pipelines/flux2/pipeline_flux2_klein.py). We then profiled
```

examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+the updated pipeline. It still didn't eliminate the gaps as expected so, we fed that back to Claude and
+it spotted something more crucial.
+
+Under the [`cache_context`](https://github.com/huggingface/diffusers/blob/f2be8bd6b3dc4035bd989dc467f15d86bf3c9c12/src/diffusers/pipelines/flux2/pipeline_flux2_klein.py#L842) manager, there is a call to `_set_context()` upon
+enters and exists. It calls `named_modules()` on the entire underlying model (in this case the Flux2 Klein DiT).
```
stevhliu:
```suggestion
enters and exits. It calls `named_modules()` on the entire underlying model (in this case the Flux2 Klein DiT).
```

examples/profiling/README.md
```diff
@@ -0,0 +1,318 @@
...
+|----------|------|-----------|-------|
+| `FluxPipeline` | text-to-image | `black-forest-labs/FLUX.1-dev` | 2 |
+| `Flux2KleinPipeline` | text-to-image | `black-forest-labs/FLUX.2-klein-base-9B` | 2 |
+| `WanPipeline` | text-to-video | `Wan-AI/Wan2.1-T2V-14B-Diffusers` | 2 |
+| `LTX2Pipeline` | text-to-video | `Lightricks/LTX-2` | 2 |
+| `QwenImagePipeline` | text-to-image | `Qwen/Qwen-Image` | 2 |
+
+> [!NOTE]
+> We use realistic inference call hyperparameters that mimic how these pipelines will be actually used. This
+> include using classifier-free guidance (where applicable), reasonable dimensions such 1024x1024, etc.
```
stevhliu:
```suggestion
> We use realistic inference call hyperparameters that mimic how these pipelines will be actually used. This
> includes using classifier-free guidance (where applicable), reasonable dimensions such 1024x1024, etc.
```

examples/profiling/README.md
sayakpaul: I am attaching it when discussing the solutions.

examples/profiling/README.md
sayakpaul: Did you mean renaming the section header to "What profiling revealed and fixes"?

If you meant a different structure entirely, I'd appreciate some help.

Regardless, I think we should keep this section in here to make the workflow complete and self-contained.

examples/profiling/README.md
```diff
@@ -0,0 +1,315 @@
...
+|----------|-----------|--------|-------|-------|
+| Flux | 1024x1024 | — | 2 | `guidance_scale=3.5` |
+| Flux2Klein | 1024x1024 | — | 2 | `guidance_scale=3.5` |
+| Wan | 480x832 | 81 | 2 | — |
+| LTX2 | 768x512 | 121 | 2 | `guidance_scale=4.0` |
+| QwenImage | 1024x1024 | — | 2 | `true_cfg_scale=4.0` |
+
+All configs use `output_type="latent"` by default (skip VAE decode for cleaner denoising-loop traces).
+
+**CLI flags:**
```
vkuzo, line 111: should the per-transformer-block compile setting be here somewhere?

examples/profiling/README.md
```diff
@@ -0,0 +1,315 @@
...
+
+Open the exported `.json` trace at [ui.perfetto.dev](https://ui.perfetto.dev/). The trace has two main rows: **CPU** (top) and **CUDA** (bottom). In Perfetto, the CPU row is typically labeled with the process/thread name (e.g., `python (PID)` or `MainThread`) and appears at the top. The CUDA row is labeled `GPU 0` (or similar) and appears below the CPU rows.
+
+**Navigation:** Use `W` to zoom in, `S` to zoom out, and `A`/`D` to pan left/right. You can also scroll to zoom and click-drag to pan. Use `Shift+scroll` to scroll vertically through rows.
+
+### What to look for
+
+**1. Gaps between CUDA kernels**
+
+Zoom into the CUDA row during the denoising loop. Ideally, GPU kernels should be back-to-back with no gaps. Gaps mean the GPU is idle waiting for the CPU to launch the next kernel. Common causes:
```
vkuzo, line 160: could mention that no bubbles is ideal, but for small shapes (small model, small batch size, or both) some bubbles could be unavoidable

examples/profiling/README.md
```diff
@@ -0,0 +1,315 @@
...
+These appear on the **CPU row** (not the CUDA row) — they are CPU-side blocking calls that wait for the GPU to finish. Look for long slices labeled `cudaStreamSynchronize` or `cudaDeviceSynchronize`. To find them: zoom into the CPU row during a denoising step and look for unusually wide slices, or use Perfetto's search bar (press `/`) and type `cudaStreamSynchronize` to jump directly to matching events. Click on a slice — if `with_stack=True` was enabled, the bottom panel ("Current Selection") shows the Python stack trace pointing to the exact line causing the sync (e.g., a `.item()` call in the scheduler).
+
+**3. Annotated regions**
+
+Our `record_function` annotations (`transformer_forward`, `scheduler_step`, etc.) appear as labeled spans on the CPU row. This lets you quickly:
+- Measure how long each phase takes (click a span to see duration)
+- See if `scheduler_step` is disproportionately expensive relative to `transformer_forward` (it should be negligible)
+- Spot unexpected CPU work between annotated regions
+
+**4. Eager vs compile comparison**
```
vkuzo, line 178: maybe we can mention cuda graphs here?

sayakpaul: Should it be included under "**Smaller CPU gaps**"? If so, it's being mentioned a bit later since I wanted to keep the scope specific.

examples/profiling/README.md
stevhliu: yeah i think renaming it to "What profiling revealed and fixes" is fine. it just wasn't really clear what "Afterwards" meant

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+| Question | Where to look | Healthy | Unhealthy |
+|----------|--------------|---------|-----------|
+| GPU staying busy? | CUDA row gaps | Back-to-back kernels | Frequent gaps > 100us |
+| CPU blocking on GPU? | `cudaStreamSynchronize` slices | Rare/absent during denoise | Present every step |
+| Scheduler overhead? | `scheduler_step` span duration | < 1% of step time | > 5% of step time |
+| Compile effective? | CUDA kernel count per step | Fewer large kernels | Same as eager |
+| Kernel launch latency? | CPU launch → GPU kernel offset | < 10us, CPU ahead of GPU | > 20us or CPU trailing GPU |
+| Memory stable? | Memory counter track | Flat during denoise loop | Growing per step |
+
+## Afterwards
```
sayakpaul:
```suggestion
## What Profiling Revealed and Fixes
```

src/diffusers/pipelines/wan/pipeline_wan.py
```diff
@@ -574,6 +574,10 @@ def __call__(
         num_warmup_steps = len(timesteps) - num_inference_steps * self.scheduler.order
         self._num_timesteps = len(timesteps)
 
+        # We set the index here to remove DtoH sync, helpful especially during compilation.
+        # Check out more details here: https://github.com/huggingface/diffusers/pull/11696
+        self.scheduler.set_begin_index(0)
```
jbschlosser, line 579: looks good, still feel like we need to broaden the scope of that fix within diffusers at some point :) I'll be out on sabbatical for the next month but I can help when I get back

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+### Notes
+
+* As mentioned above, we profiled with regional compilation so it's possible that
+there are still some gaps outside the compiled regions. A full compilation
+will likely mitigate it. In case it doesn't, the above observations could
+be useful to mitigate that.
+* Use of CUDA Graphs can also help mitigate CPU overhead related issues. When
+using "reduce-overhead" and "max-autotune" in `torch.compile` triggers the
+use of CUDA Graphs.
```
jbschlosser: I'm glad you mentioned this here - I wonder if it's worth clarifying in the respective sections of this doc that CUDAGraph usage is the reason why we expect gap removal from using torch.compile

jbschlosser (commented):
Nice work! Love to see this

src/diffusers/pipelines/wan/pipeline_wan.py
sayakpaul: This would be awesome. Created https://github.com/huggingface/diffusers/issues/13375 to track. Thanks for offering to help.

examples/profiling/README.md
sayakpaul: The results and the graph presented in this doc were obtained with the "default" compilation mode (along with regional compilation):
https://github.com/huggingface/diffusers/blob/131831ff20e6815075cdae923f4d5be85c9dd175/examples/profiling/run_profiling.sh#L30

So not sure? There are also being shipped in this PR that helped mitigate the stalling issues 👀

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
+# Profiling a `DiffusionPipeline` with the PyTorch Profiler
+
+Education materials to strategically profile pipelines to potentially improve their
+runtime with `torch.compile`. To set these pipelines up for success with `torch.compile`,
+we often have to get rid of DtoH syncs, CPU overheads, kernel launch delays, and
```
dg845:
```suggestion
we often have to get rid of device-to-host (DtoH) syncs, CPU overheads, kernel launch delays, and
```
I'm not sure if this terminology will be familiar to the target audience, I had to look it up the first time to find out what it meant.

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+* [How the tooling works](#how-the-tooling-works)
+* [Verification](#verification)
+* [Interpretation](#interpreting-traces-in-perfetto-ui)
+* [Taking profiling-guided steps for improvements](#afterwards)
+
+Jump to the "Verification" section to get started right away.
+
+## Context
+
+We want to uncover CPU overhead, CPU-GPU sync points, and other bottlenecks in popular diffusers pipelines — especially issues that become non-trivial under `torch.compile`. The approach is inspired by [flux-fast's run_benchmark.py](https://github.com/huggingface/flux-fast/blob/0a1dcc91658f0df14cd7fce862a5c8842784c6da/run_benchmark.py#L66-L85) which uses `torch.profiler` with method-level annotations, and motivated by issues like [diffusers#11696](https://github.com/huggingface/diffusers/pull/11696) (DtoH sync from scheduler `.item()` call).
```
dg845:
```suggestion
We want to uncover CPU overhead, CPU-GPU sync points, and other bottlenecks in popular diffusers pipelines — especially issues that become non-trivial when using [`torch.compile`](https://docs.pytorch.org/docs/stable/generated/torch.compile.html). The approach is inspired by [flux-fast's run_benchmark.py](https://github.com/huggingface/flux-fast/blob/0a1dcc91658f0df14cd7fce862a5c8842784c6da/run_benchmark.py#L66-L85) which uses [`torch.profiler`](https://docs.pytorch.org/docs/stable/profiler.html) with method-level annotations, and motivated by issues like [diffusers#11696](https://github.com/huggingface/diffusers/pull/11696) (DtoH sync from scheduler `.item()` call).
```
I think adding links to the `torch.compile` and `torch.profiler` docs could be useful for following along, especially if readers aren't familiar with them.

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+| `QwenImagePipeline` | text-to-image | `Qwen/Qwen-Image` | 2 |
+
+> [!NOTE]
+> We use realistic inference call hyperparameters that mimic how these pipelines will be actually used. This
+> includes using classifier-free guidance (where applicable), reasonable dimensions such 1024x1024, etc.
+> But we keep the number of inference steps to a bare minimum. 
+
+## How the Tooling Works
+
+Follow the flux-fast pattern: **annotate key pipeline methods** with `torch.profiler.record_function` wrappers, then run the pipeline under `torch.profiler.profile` and export a Chrome trace.
```
dg845:
```suggestion
Follow the flux-fast pattern: **annotate key pipeline methods** with `torch.profiler.record_function` wrappers, then run the pipeline under `torch.profiler.profile` and export a Chrome JSON trace.
```

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+1. **Scheduler DtoH sync via `nonzero().item()`** — For Flux, this was fixed by adding `scheduler.set_begin_index(0)` before the denoising loop ([diffusers#11696](https://github.com/huggingface/diffusers/pull/11696)). Profiling should reveal whether similar sync points exist in other pipelines.
+
+2. **`modulate_index` tensor rebuilt every forward in `transformer_qwenimage.py`** (line 901-905) — Python list comprehension + `torch.tensor()` each step. Minor but visible in trace.
+
+3. **Any other `.item()`, `.cpu()`, `.numpy()` calls** in the denoising loop hot path — the profiler's `with_stack=True` will surface these as CPU stalls with Python stack traces.
+
+## Verification
+
+1. Run: `python profiling/profiling_pipelines.py --pipeline flux --mode eager --num_steps 2`
```
dg845:
```suggestion
1. Run: `python examples/profiling/profiling_pipelines.py --pipeline flux --mode eager --num_steps 2`
```
I think running the script from the `diffusers` root directory might be more common?

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+   - Stack traces visible on CPU events
+4. Run with `--mode compile` and compare trace for fewer/fused CUDA kernels
+
+You can also use the `run_profiling.sh` script to bulk launch runs for different pipelines.
+
+## Interpreting Traces in Perfetto UI
+
+Open the exported `.json` trace at [ui.perfetto.dev](https://ui.perfetto.dev/). The trace has two main rows: **CPU** (top) and **CUDA** (bottom). In Perfetto, the CPU row is typically labeled with the process/thread name (e.g., `python (PID)` or `MainThread`) and appears at the top. The CUDA row is labeled `GPU 0` (or similar) and appears below the CPU rows.
+
+**Navigation:** Use `W` to zoom in, `S` to zoom out, and `A`/`D` to pan left/right. You can also scroll to zoom and click-drag to pan. Use `Shift+scroll` to scroll vertically through rows.
```
dg845, line 150: I also found `,` and `.`, which select the previous/next track event, useful for e.g. finding the next `transformer_forward` event, or finding the next GPU kernel on the GPU kernel track.

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+- Measure how long each phase takes (click a span to see duration)
+- See if `scheduler_step` is disproportionately expensive relative to `transformer_forward` (it should be negligible)
+- Spot unexpected CPU work between annotated regions
+
+**4. Eager vs compile comparison**
+
+Open both traces side by side (two Perfetto tabs). Key differences to look for:
+- **Fewer, wider CUDA kernels** in compile mode (fused ops) vs many small kernels in eager
+- **Smaller CPU gaps** between kernels in compile mode (less Python dispatch overhead)
+- **CUDA kernel count per step**: to compare, zoom into a single `transformer_forward` span on the CUDA row and count the distinct kernel slices within it. In eager mode you'll typically see many narrow slices (one per op); in compile mode these fuse into fewer, wider slices. A quick way to estimate: select a time range covering one denoising step on the CUDA row — Perfetto shows the number of slices in the selection summary at the bottom. If compile mode shows a similar kernel count to eager, fusion isn't happening effectively (likely due to graph breaks).
```
dg845, line 183: When I tried comparing an `eager` and `compile` trace side-by-side, I found that it was difficult to find corresponding events because if `--mode compile` is used, there appear to be no `transformer_forward` events in the trace, but rather one large `## Call CompiledFxGraph...` event. (I think if regional compilation is used via `--compile_regional`, the `transformer_forward` event does appear again, with the `## CompiledFxGraph...` events under the `transformer_forward` events.)

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+**6. Kernel launch latency**
+
+Each CUDA kernel is launched from the CPU. The CPU-side launch calls (`cudaLaunchKernel`) appear as small slices on the **CPU row** — zoom in closely to a denoising step to see them. The corresponding GPU-side kernel executions appear on the **CUDA row** directly below. You can also use Perfetto's search bar (`/`) and type `cudaLaunchKernel` to find them. The time between the CPU dispatch and the GPU kernel starting should be minimal (single-digit microseconds). If you see consistent delays > 10-20us between launch and execution:
+- The launch queue may be starved because of excessive Python work between ops
+- There may be implicit syncs forcing serialization
+- `torch.compile` should help here by batching launches — compare eager vs compile to confirm
+
+To inspect this: zoom into a single denoising step, select a CUDA kernel on the GPU row, and look at the corresponding CPU-side launch slice directly above it. The horizontal offset between them is the launch latency. In a healthy trace, CPU launch slices should be well ahead of GPU execution (the CPU is "feeding" the GPU faster than it can consume).
```
dg845:
```suggestion
To inspect this: zoom into a single denoising step, select a CUDA kernel on the GPU row, and look at the corresponding CPU-side launch slice directly above it (there should be an arrow pointing from the CPU launch slice to the GPU kernel slice). The horizontal offset between them is the launch latency. In a healthy trace, CPU launch slices should be well ahead of GPU execution (the CPU is "feeding" the GPU faster than it can consume).
```
Not sure about the exact wording, but I think mentioning this is helpful, especially if there is a big temporal gap between the CPU `cudaLaunchKernel` and corresponding GPU kernel execution.

May also be worth mentioning that when the GPU kernel is selected, the corresponding `cudaLaunchKernel` event should be in the "Preceding Flows" section; I believe the "Delay" column then gives the exact launch latency.

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+      <img src="https://huggingface.co/datasets/sayakpaul/torch-profiling-trace-diffusers/resolve/main/Flux2-Klein/Screenshot%202026-03-27%20at%2011.05.06%E2%80%AFAM.png" alt="Image 2"><br>
+      <em>With compile</em>
+    </td>
+  </tr>
+</table>
+
+### Spotting gaps between launches
+
+Then a reasonable next step is to spot frequent gaps between kernel executions. In the compiled
+case, we don't spot any on the surface. But if we zone in, some become apparent.
```
dg845:
```suggestion
case, we don't spot any on the surface. But if we zoom in, some become apparent.
```
nit: typo

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+    </td>
+    <td align="center">
+      <img src="https://huggingface.co/datasets/sayakpaul/torch-profiling-trace-diffusers/resolve/main/Flux2-Klein/Screenshot%202026-03-27%20at%2010.24.34%E2%80%AFAM.png" alt="Image 2"><br>
+      <em>Gaps become more visible when zoomed in</em>
+    </td>
+  </tr>
+</table>
+
+So, we provided the profile trace file (with compilation) to Claude, asked it to find the instances of
+"cudaStreamSynchronize" and "cudaDeviceSynchronize", and to come up with some potential fixes.
```
dg845:
```suggestion
`cudaStreamSynchronize` and `cudaDeviceSynchronize`, and to come up with some potential fixes.
```
nit: formatting

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+      <em>Without compile</em>
+    </td>
+    <td align="center">
+      <img src="https://huggingface.co/datasets/sayakpaul/torch-profiling-trace-diffusers/resolve/main/Flux2-Klein/Screenshot%202026-03-27%20at%2011.05.06%E2%80%AFAM.png" alt="Image 2"><br>
+      <em>With compile</em>
+    </td>
+  </tr>
+</table>
+
+### Spotting gaps between launches
```
dg845, line 229: I find this section somewhat unsatisfying because it glosses over the reasoning behind each step in the profiling process. I think it would be more instructive if the chain of reasoning that led us (and Claude) to

1. Find the `tqdm` and `_unpack_latents_with_ids` fixes
2. Find out that these fixes weren't sufficient, and why they weren't sufficient
3. Discover that `cache_context` was the bottleneck

was discussed at greater length in this section.

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+- Root cause: _unpack_latents_with_ids() (called right after the denoising loop) computes h = torch.max(h_ids) +
+1 and w = torch.max(w_ids) + 1 on GPU tensors, then uses them as shape args for torch.zeros((h * w, ch), ...).
+This triggers an implicit .item() DtoH sync, blocking the CPU while the GPU is still finishing the last
+transformer forward's kernels.
+- Fix: Added height/width parameters to _unpack_latents_with_ids(), pre-computed from the known pixel dimensions
+at the call site.
+```
+
+The changes looked reasonable based on our past experience. So, we asked Claude to apply these changes to [`pipeline_flux2_klein.py`](../../src/diffusers/pipelines/flux2/pipeline_flux2_klein.py). We then profiled
+the updated pipeline. It still didn't eliminate the gaps as expected so, we fed that back to Claude and
```
dg845: I think it would be nice to have a demonstration using the profiling outputs that the `_unpack_latents_with_ids` fix indeed eliminates a DtoH sync because the current wording

> It still didn't eliminate the gaps as expected

makes it unclear if the change is effective.

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+For large models, when they are invoked iteratively like our case, it adds to the latency because it involes traversing hundreds of submodules.
+
+The fix was to build a list of hooked child registries once on the first call and cache it in `_child_registries_cache`. This way, the subsequent calls would return the cached list directly without
+any traversal. With the fix applied, the improvements were visible.
+
+|                        | Before                       | After                       |
+|------------------------|------------------------------|-----------------------------|
+| `_set_context` total   | 21.6ms (8 calls)             | 0.0ms (8 calls)             |
+| `cache_context` total  | 21.7ms                       | 0.1ms                       |
+| CPU gaps               | 5,523us / 8,007us / 5,508us  | 158us / 2,777us / 136us     |
```
dg845, line 290: I think adding another row with the total wall clock time (or another "overall performance" metric) before and after would be useful here because it's not obvious to me that reducing the CPU gaps here necessarily leads to better performance overall.

examples/profiling/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+We also profiled the **Wan** model and uncovered problems related to CPU DtoH syncs. Below is an
+overview.
+
+First, there was a dynamo cache lookup delay making the GPU idle as reported [in this PR](https://github.com/huggingface/diffusers/pull/11696). So, the fix was to call `self.scheduler.set_begin_index(0)` before
+the denoising loop. This tells the scheduler the starting index is 0, so `_init_step_index()` skips the `nonzero().item()` (which was causing the sync) path entirely. This fix eliminated the below ~2.3s GPU idle time completely:
+
+![GPU idle](https://huggingface.co/datasets/sayakpaul/torch-profiling-trace-diffusers/resolve/main/Wan/Screenshot%202026-03-27%20at%205.56.39%E2%80%AFPM.png)
+
+The UniPC scheduler (used in Wan) creates small constant tensors via `torch.tensor([0.5], dtype=x.dtype, device=device)` during `step()`. This triggers a "cudaMemcpyAsync + cudaStreamSynchronize" to copy
```
dg845:
```suggestion
The UniPC scheduler (used in Wan) creates small constant tensors via `torch.tensor([0.5], dtype=x.dtype, device=device)` during `step()`. This triggers a `cudaMemcpyAsync` + `cudaStreamSynchronize` to copy
```
nit: formatting

examples/profiling/README.md
dg845:
```suggestion
* Use of CUDA Graphs can also help mitigate CPU overhead related issues. CUDA Graphs can be enabled by setting the `torch.compile` mode to `"reduce-overhead"` or `"max-autotune"`.
```
nit: I think the wording here is awkward

src/diffusers/pipelines/flux2/pipeline_flux2_klein.py
```diff
@@ -396,8 +396,9 @@ def _pack_latents(latents):
         return latents
 
     @staticmethod
-    # Copied from diffusers.pipelines.flux2.pipeline_flux2.Flux2Pipeline._unpack_latents_with_ids
```
dg845, line 399: Since `_unpack_latents_with_ids` was originally `# Copied from` `Flux2Pipeline`, should the changes here also be propagated to that pipeline?

src/diffusers/schedulers/scheduling_unipc_multistep.py
```diff
@@ -903,8 +903,8 @@ def multistep_uni_p_bh_update(
             rks.append(rk)
             D1s.append((mi - m0) / rk)
 
-        rks.append(1.0)
-        rks = torch.tensor(rks, device=device)
+        rks.append(torch.ones((), device=device))
+        rks = torch.stack(rks)
```
dg845, line 907: Should this change also be explained in the docs (`examples/profiling/README.md`)?

src/diffusers/pipelines/flux2/pipeline_flux2_klein.py
sayakpaul: Didn't do it because I didn't profile that pipeline. So, would like to keep the changes as minimal as possible for this PR.

examples/profiling/README.md
sayakpaul: WASD keys are popular so decided to just include them.

examples/profiling/README.md
sayakpaul: Yes. We're making it clear from the beginning that we're using regional compilation. Anything you would like to edit here?

examples/profiling/README.md
sayakpaul: Indeed. Would you like to suggest the wording, please?

examples/profiling/README.md
sayakpaul: I have followed the double quotes throughout so, would prefer that.

examples/profiling/README.md
sayakpaul: Neither of them (`tqdm` + `_unpack_latents_with_ids` completely eliminated the gap as written. What is unclear about it?

examples/profiling/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Verification
+
+1. Run: `python examples/profiling/profiling_pipelines.py --pipeline flux --mode eager --num_steps 2`
+2. Verify `profiling_results/flux_eager.json` is produced
+3. Open trace in [Perfetto UI](https://ui.perfetto.dev/) — confirm:
+   - `transformer_forward` and `scheduler_step` annotations visible
+   - CPU and CUDA timelines present
+   - Stack traces visible on CPU events
+4. Run with `--mode compile` and compare trace for fewer/fused CUDA kernels
```
dg845:
```suggestion
4. Run with `--mode compile`: `python examples/profiling/profiling_pipelines.py --pipeline flux --mode compile --compile_regional --num_steps 2` and compare trace for fewer/fused CUDA kernels
```
Follow up suggestion for https://github.com/huggingface/diffusers/pull/13356#discussion_r3025491604. We explicitly specify `--compile_regional` as the following section assumes this is used.

examples/profiling/README.md
dg845: I have suggested an edit at https://github.com/huggingface/diffusers/pull/13356#discussion_r3031651459 to make this more clear.

examples/profiling/README.md
dg845: I think if that is the case, saying something like "It still didn't completely eliminate the gaps as expected..." would be more clear. In my opinion, the original wording doesn't specify whether the changes didn't work at all (the gaps were the same as before) or whether it worked, but not completely (some gaps still remained).

sayakpaul: Feel free to make a suggestion directly. I think it's easier that way.

examples/profiling/README.md
```diff
@@ -0,0 +1,325 @@
...
+- Root cause: _unpack_latents_with_ids() (called right after the denoising loop) computes h = torch.max(h_ids) +
+1 and w = torch.max(w_ids) + 1 on GPU tensors, then uses them as shape args for torch.zeros((h * w, ch), ...).
+This triggers an implicit .item() DtoH sync, blocking the CPU while the GPU is still finishing the last
+transformer forward's kernels.
+- Fix: Added height/width parameters to _unpack_latents_with_ids(), pre-computed from the known pixel dimensions
+at the call site.
+```
+
+The changes looked reasonable based on our past experience. So, we asked Claude to apply these changes to [`pipeline_flux2_klein.py`](../../src/diffusers/pipelines/flux2/pipeline_flux2_klein.py). We then profiled
+the updated pipeline. It still didn't eliminate the gaps as expected so, we fed that back to Claude and
```
dg845:
```suggestion
the updated pipeline. It still didn't completely eliminate the gaps as expected so, we fed that back to Claude and
```

examples/profiling/README.md
dg845: Suggestion: https://github.com/huggingface/diffusers/pull/13356#discussion_r3031689026

sayakpaul:
@dg845 thanks for the comments, very helpful. I think you also advocated for the end user which is exactly what I was looking for in the reviews.

I have made the changes as requested. PTAL.

examples/profiling/README.md
```diff
@@ -0,0 +1,342 @@
...
+
+![GPU idle](https://huggingface.co/datasets/sayakpaul/torch-profiling-trace-diffusers/resolve/main/Wan/Screenshot%202026-03-27%20at%205.56.39%E2%80%AFPM.png)
+
+Similar to the above-mentioned PR, the fix was to call `self.scheduler.set_begin_index(0)` before the denoising loop. This tells the scheduler the starting index is 0, so `_init_step_index()` skips the `nonzero().item()` (which was causing the sync) path entirely. This fix eliminated the ~2.3s GPU idle time completely.
+
+The UniPC scheduler (used in Wan) also had two more sync-causing patterns in `multistep_uni_p_bh_update` and `multistep_uni_c_bh_update`:
+
+1. **`torch.tensor(rks, device=device)`** where `rks` is a list containing GPU scalar tensors. `torch.tensor()` pulls each GPU value back to CPU to construct a new tensor, triggering a DtoH sync. 
+
+**Fix**: Replace with `torch.stack(rks)` which concatenates GPU tensors directly on the GPU — no sync needed. The appended Python float `1.0` was also changed to `torch.ones((), device=device)` so the list is homogeneously GPU tensors.
```
dg845:
```suggestion
**Fix**: Replace with `torch.stack(rks)` which concatenates GPU tensors directly on the GPU — no sync needed. The appended Python float `1.0` was also changed to `torch.ones((), device=device)` so the list contains only GPU tensors.
```
nit: I think the original wording is a bit awkward

PR #13367, opened 2026-03-30
fix(ddim): validate eta is in [0, 1] in DDIMPipeline

### LINKED ISSUE #13362: DDIMPipeline does not validate eta range despite documented constraint [0, 1]
### Describe the bug

In DDIMPipeline, the eta parameter is documented to be within the range [0, 1], but there is currently no validation enforcing this constraint.

From the docstring:

>"eta corresponds to η in paper and should be between [0, 1]"

However, users can pass arbitrary values (e.g., negative or >1) without any warning or error, which may lead to undefined or unintended behavior during inference.

This creates a mismatch between:

- documented API contract  
- actual implementation behavior  

I’d be happy to submit a PR to add validation if this is considered appropriate.

### Reproduction

```py
from diffusers import DDIMPipeline

pipe = DDIMPipeline.from_pretrained("google/ddpm-cifar10-32")

# Invalid eta values (should ideally raise error or warning)
image = pipe(eta=-1.0)
image = pipe(eta=2.0)
```

Both calls execute without validation or error.

### Logs

```shell
Either:

- Raise ValueError if eta is outside [0, 1]
- Or explicitly document that values outside the range are allowed (if intentional)
```

### System Info

- diffusers: local dev version (from source)
- Python: 3.13
- OS: Windows
- PyTorch: 2.11.0

### Who can help?

_No response_

NIK-TIGER-BILL:
## What does this PR do?

Adds input validation for the `eta` parameter in `DDIMPipeline.__call__`.

Fixes #13362

### Context

The DDIM paper defines η (eta) as a value that lies strictly in [0, 1]:
- η = 0 → deterministic DDIM
- η = 1 → DDPM

The docstring already documents this constraint:
> *"eta corresponds to η in paper and should be between [0, 1]"*

However, no runtime check existed, so out-of-range values (negative or > 1) were silently accepted and could produce undefined/unintended sampling behaviour.

### Change

Add a `ValueError` guard immediately after the image-shape computation, before the denoising loop:

```python
if not 0.0 <= eta <= 1.0:
    raise ValueError(
        f"`eta` must be between 0 and 1 (inclusive), but received {eta}. "
        "A value of 0 corresponds to DDIM and 1 corresponds to DDPM."
    )
```

## Before submitting
- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? See #13362

## Who can review?

@pcuenca @DN6

### DIFF
--- a/src/diffusers/pipelines/ddim/pipeline_ddim.py
+++ b/src/diffusers/pipelines/ddim/pipeline_ddim.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import warnings
+import logging
 
 import torch
 
@@ -21,6 +21,9 @@ from ...schedulers import DDIMScheduler
 from ...utils import is_torch_xla_available
 from ...utils.torch_utils import randn_tensor
 from ..pipeline_utils import DiffusionPipeline, ImagePipelineOutput
+
+
+logger = logging.getLogger(__name__)
 
 
 if is_torch_xla_available():
@@ -132,12 +135,10 @@ class DDIMPipeline(DiffusionPipeline):
             image_shape = (batch_size, self.unet.config.in_channels, *self.unet.config.sample_size)
 
         if not 0.0 <= eta <= 1.0:
-            warnings.warn(
+            logger.warning(
                 f"`eta` should be between 0 and 1 (inclusive), but received {eta}. "
                 "A value of 0 corresponds to DDIM and 1 corresponds to DDPM. "
-                "Unexpected results may occur for values outside this range.",
-                UserWarning,
-                stacklevel=2,
+                "Unexpected results may occur for values outside this range."
             )
 
         if isinstance(generator, list) and len(generator) != batch_size:


yiyixuxu:
thanks for the PR,  can you change it to a warning?
honestly I think the documentation is sufficient

NIK-TIGER-BILL:
Thanks for the feedback, @yiyixuxu! Changed to a `warnings.warn` (UserWarning) in the latest commit — the validation is still there to inform callers, but no longer raises an exception. Let me know if the wording looks good to you!

src/diffusers/pipelines/ddim/pipeline_ddim.py
```diff
@@ -129,6 +131,15 @@ def __call__(
         else:
             image_shape = (batch_size, self.unet.config.in_channels, *self.unet.config.sample_size)
 
+        if not 0.0 <= eta <= 1.0:
+            warnings.warn(
```
yiyixuxu: @claude what do you think?

yiyixuxu: can we change it to use logger.warning? to be consistent with all other pipelines

NIK-TIGER-BILL: Good call! Updated in the latest commit — switched from `warnings.warn()` to `logger.warning()` to match the convention used in all other diffusers pipelines. Also removed the now-unused `import warnings`.

PR #13368, opened 2026-03-30
[tests] refactor autoencoderkl tests

sayakpaul:


tests/models/autoencoders/test_models_autoencoder_kl.py
```diff
@@ -362,10 +359,7 @@ def test_stable_diffusion_decode_xformers_vs_2_0_fp16(self, seed):
 
     @parameterized.expand([(13,), (16,), (37,)])
     @require_torch_gpu
-    @unittest.skipIf(
-        not is_xformers_available(),
-        reason="xformers is not required when using PyTorch 2.0.",
-    )
+    @pytest.mark.skipif(not is_xformers_available(), reason="xformers is not required when using PyTorch 2.0.")
```
DN6, line 362: Why don't we remove these xformers tests? They're not as relevant and we rely on `set_attention_backend` for this functionality.

sayakpaul:
@DN6 a gentle ping.

tests/models/autoencoders/test_models_autoencoder_kl.py
sayakpaul: Do we support `set_attention_backend()` on AutoencoderKL? I don't think we do.

I don't think this was a merge blocker 😅

sayakpaul:
@DN6 a gentle ping.

DN6 (approved):

PR #13371, opened 2026-03-31
[tests] refactor wan autoencoder tests

sayakpaul:


tests/models/autoencoders/test_models_autoencoder_wan.py
```diff
@@ -39,54 +44,51 @@ def get_autoencoder_kl_wan_config(self):
...
-    @property
-    def dummy_input(self):
+    def get_dummy_inputs(self, seed=0):
+        torch.manual_seed(seed)
         batch_size = 2
         num_frames = 9
         num_channels = 3
         sizes = (16, 16)
-        image = floats_tensor((batch_size, num_channels, num_frames) + sizes).to(torch_device)
+        image = torch.randn(batch_size, num_channels, num_frames, *sizes).to(torch_device)
```
DN6: Would use randn_tensor. You can use a Generator on CPU for better deterministic output.

tests/models/autoencoders/test_models_autoencoder_wan.py
```diff
@@ -39,54 +44,51 @@ def get_autoencoder_kl_wan_config(self):
...
         return {"sample": image}
 
-    @property
-    def dummy_input_tiling(self):
+    # Bridge for AutoencoderTesterMixin which still uses the old interface
```
DN6: I would recommend just creating a new Mixin e.g `AutoencoderTestMixin` that follows the new format and then once all the new AE tests are merged, we can remove the old class.

tests/models/autoencoders/testing_utils.py
```diff
@@ -145,3 +145,138 @@ def test_enable_disable_slicing(self):
             output_without_slicing.detach().cpu().numpy().all(),
             output_without_slicing_2.detach().cpu().numpy().all(),
         ), "Without slicing outputs should match with the outputs when slicing is manually disabled."
+
+
+class NewAutoencoderTesterMixin:
```
sayakpaul, line 150: Basically same as `AutoencoderTesterMixin` but reimplemented to fit the current mixin structures of the model tests.

PR #13379, opened 2026-04-01
fix(qwen): fix CFG failing when passing neg prompt embeds with none mask

### LINKED ISSUE #13377: [Bug] `QwenImagePipeline` silently disables CFG when passing `negative_prompt_embeds` if mask is `None` (which `encode_prompt` returns by default)
### Describe the bug

In `QwenImagePipeline`, when users manually pre-compute prompt embeddings to optimize memory usage (e.g., placing the encoder and transformer on different GPUs), Classifier-Free Guidance (CFG) is silently disabled if `negative_prompt_embeds_mask` is set to `None`. 

However, `encode_prompt` explicitly converts an all-ones mask to `None` as an optimization. This creates a logical contradiction where the pipeline's own encoder returns a valid state (`None`) that the `__call__` method subsequently rejects, causing CFG to fail with a warning.

### Reproduction

If you manually extract embeddings and pass them to the pipeline:

```python
# 1. Manually encode prompts
pos_embeds, pos_mask = pipeline.encode_prompt("A photo of a cat")
neg_embeds, neg_mask = pipeline.encode_prompt("bad quality") 
# Note: neg_mask is often `None` here because `encode_prompt` optimizes `prompt_embeds_mask.all() -> None`

# 2. Pass them to the pipeline
image = pipeline(
    prompt_embeds=pos_embeds,
    prompt_embeds_mask=pos_mask,
    negative_prompt_embeds=neg_embeds,
    negative_prompt_embeds_mask=neg_mask,  # This passes None
    true_cfg_scale=4.0
).images[0]
```

**Output Warning:**
```text
true_cfg_scale is passed as 4.0, but classifier-free guidance is not enabled since no negative_prompt is provided.
```

**Root Cause Analysis:**
In `pipeline_qwenimage.py`:

**1. The `has_neg_prompt` check in `__call__` requires the mask to NOT be `None`:**
```python
has_neg_prompt = negative_prompt is not None or (
    negative_prompt_embeds is not None and negative_prompt_embeds_mask is not None
)
```

**2. But `encode_prompt` intentionally sets the mask to `None` if it's full:**
```python
if prompt_embeds_mask is not None:
    # ... (reshape logic)
    if prompt_embeds_mask.all():
        prompt_embeds_mask = None  # <--- Here!
```

Because `neg_mask` becomes `None`, `has_neg_prompt` evaluates to `False`, and `do_true_cfg` is set to `False`.

**Expected behavior:**
The presence of `negative_prompt_embeds` alone should be sufficient to trigger `has_neg_prompt = True`. The `negative_prompt_embeds_mask` being `None` should simply mean "no masking is required" (all valid), which is consistent with the behavior of `encode_prompt`.

The check in `__call__` should probably be relaxed to:
```python
has_neg_prompt = negative_prompt is not None or negative_prompt_embeds is not None
```

**Temporary Workaround:**
Users currently have to manually fake an all-ones mask before passing it to the transformer:
```python
if neg_mask is None:
    neg_mask = torch.ones(neg_embeds.shape[:2], dtype=torch.long, device=device)
```

### Logs

```shell

```

### System Info

- 🤗 Diffusers version: 0.37.1
- Platform: Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
- Python version: 3.10.20
- PyTorch version (GPU?): 2.11.0+cu130 (True)
- Jax version: not installed
- Huggingface_hub version: 1.8.0
- Transformers version: 5.4.0
- Accelerate version: 1.13.0
- PEFT version: not installed
- Safetensors version: 0.7.0
- Accelerator: NVIDIA GeForce RTX 3060, 12288 MiB
- Using GPU in script?: Yes
- Using distributed or parallel set-up in script?: Yes

### Who can help?

@yiyixuxu @DN6

Sunhill666:
# What does this PR do?

Fixes #13377

This PR fixes a bug in `QwenImagePipeline` and `QwenImageEditPipeline` where Classifier-Free Guidance (CFG) is incorrectly disabled when users manually pass `negative_prompt_embeds` alongside a `None` mask.

**Root cause:**
The `encode_prompt` method naturally optimizes an all-ones `prompt_embeds_mask` by converting and returning it as `None`. However, the `has_neg_prompt` condition in the `__call__` method strictly required `negative_prompt_embeds_mask is not None` to activate CFG. This logical mismatch caused `do_true_cfg` to evaluate to `False` even when valid negative embeddings were provided. 

**Changes:**
- Relaxed the `has_neg_prompt` condition. 
- It now correctly evaluates `has_neg_prompt` to `True` as long as `negative_prompt_embeds is not None`, since a `None` mask simply implies that no padding/masking is required.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? https://github.com/huggingface/diffusers/issues/13377.

## Who can review?

@asomoza @yiyixuxu


asomoza:
@claude can you review this?

asomoza:
thanks! @sayakpaul should we add a test for this?

asomoza (approved):

sayakpaul:
Yes there should be a test for it sure. I think this PR could handle it too.

Sunhill666:
@sayakpaul Thanks for the feedback! I completely agree. 

I've pushed two commit that addresses the issues and adds the requested tests:

1. **Test Coverage**: Added `test_true_cfg_without_negative_prompt_embeds_mask` to all QwenImage `FastTests` (base, controlnet, edit, edit_plus, img2img, inpaint).
2. **Fixed Legacy Mask Constraints**: While adding the tests, I uncovered that some pipeline variants (like `edit`, `edit_plus`, `layered`, and `controlnet_inpaint`) had rigid `raise ValueError` blocks in `check_inputs()` preventing `None` masks. Their `encode_prompt()` also unconditionally called `.repeat()` on the mask, causing `AttributeError` when `mask` was `None`. I've thoroughly removed these blockers and safely guarded the tensor operations.
3. **Code Consistency**: Addressed the review comment by standardizing the `model_inputs` accesses to uniformly use `.get("...")` (e.g., `input_ids`, `attention_mask`, `pixel_values`) in the `edit` pipelines.

Everything passes locally now. Let me know if there's anything else needed!

src/diffusers/modular_pipelines/qwenimage/encoders.py
```diff
@@ -117,15 +117,15 @@ def get_qwen_prompt_embeds_edit(
...
 
     outputs = text_encoder(
-        input_ids=model_inputs.input_ids,
-        attention_mask=model_inputs.attention_mask,
-        pixel_values=model_inputs.pixel_values,
-        image_grid_thw=model_inputs.image_grid_thw,
+        input_ids=model_inputs.get("input_ids"),
+        attention_mask=model_inputs.get("attention_mask"),
+        pixel_values=model_inputs.get("pixel_values"),
+        image_grid_thw=model_inputs.get("image_grid_thw"),
```
sayakpaul, line 123: What is the reasoning behind this?

src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py
```diff
@@ -584,9 +584,7 @@ def __call__(
 
         device = self._execution_device
 
-        has_neg_prompt = negative_prompt is not None or (
-            negative_prompt_embeds is not None and negative_prompt_embeds_mask is not None
-        )
+        has_neg_prompt = negative_prompt is not None or negative_prompt_embeds is not None
```
sayakpaul, line 596: How do we handle `negative_prompt_embeds_mask`? 👀 Is this block affected?

https://github.com/huggingface/diffusers/blob/71a6fd9f0df04d3764dfa999268a05d87903a85a/src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py#L261-L267

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_controlnet_inpaint.py
```diff
@@ -302,11 +302,13 @@ def encode_prompt(
...
-        prompt_embeds_mask = prompt_embeds_mask.view(batch_size * num_images_per_prompt, seq_len)
 
-        if prompt_embeds_mask is not None and prompt_embeds_mask.all():
-            prompt_embeds_mask = None
+        if prompt_embeds_mask is not None:
+            prompt_embeds_mask = prompt_embeds_mask.repeat(1, num_images_per_prompt, 1)
+            prompt_embeds_mask = prompt_embeds_mask.view(batch_size * num_images_per_prompt, seq_len)
+
+            if prompt_embeds_mask.all():
+                prompt_embeds_mask = None
```
sayakpaul, line 314: Seems like an unrelated change? Since it already has 

https://github.com/huggingface/diffusers/blob/71a6fd9f0df04d3764dfa999268a05d87903a85a/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_controlnet_inpaint.py#L272

if we run `make fix-copies`, the changes would be propagated automatically.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_controlnet_inpaint.py
```diff
@@ -353,15 +355,6 @@ def check_inputs(
                 f" {negative_prompt_embeds}. Please make sure to only forward one of the two."
             )
 
-        if prompt_embeds is not None and prompt_embeds_mask is None:
```
sayakpaul, line 356: What is the reasoning behind this?

What happens when users pass the embeds and not the masks? Maybe we should warn them?

tests/pipelines/qwenimage/test_qwenimage.py
```diff
@@ -234,3 +234,29 @@ def test_vae_tiling(self, expected_diff_max: float = 0.2):
             expected_diff_max,
             "VAE tiling should not affect the inference results",
         )
+
+    def test_true_cfg_without_negative_prompt_embeds_mask(self):
+        components = self.get_dummy_components()
+        pipe = self.pipeline_class(**components)
+        pipe.to("cpu")
```
sayakpaul:
```suggestion
        pipe.to(torch_device)
```

tests/pipelines/qwenimage/test_qwenimage.py
```diff
@@ -234,3 +234,29 @@ def test_vae_tiling(self, expected_diff_max: float = 0.2):
...
+        pipe.set_progress_bar_config(disable=None)
+
+        inputs = self.get_dummy_inputs("cpu")
```
sayakpaul:
```suggestion
        inputs = self.get_dummy_inputs(torch_device)
```

tests/pipelines/qwenimage/test_qwenimage.py
```diff
@@ -234,3 +234,29 @@ def test_vae_tiling(self, expected_diff_max: float = 0.2):
...
+            device="cpu",
+            num_images_per_prompt=1,
+            max_sequence_length=inputs.get("max_sequence_length", 16),
+        )
+
+        inputs["prompt_embeds"] = prompt_embeds
+        inputs["prompt_embeds_mask"] = prompt_embeds_mask
+        inputs["negative_prompt_embeds"] = prompt_embeds
+        inputs["negative_prompt"] = None
+        inputs["negative_prompt_embeds_mask"] = None
```
sayakpaul: Should we not be using it because it's the core thing that we're fixing here?

src/diffusers/modular_pipelines/qwenimage/encoders.py
Sunhill666: When `image=None` is passed to the `Qwen2VLProcessor` (for example, when encoding the `negative_prompt`), the processor returns a `BatchFeature` object that does not contain the `pixel_values` and `image_grid_thw` keys. 

Attempting to access them directly via dot notation (e.g. `model_inputs.pixel_values`) raises an `AttributeError` from the underlying `transformers.utils.logging.FeatureExtractionUtils` mapping class. Using `.get()` safely defaults to `None`, which prevents the pipeline from crashing during negative prompt generations. I also updated `input_ids` and `attention_mask` to use `.get()` here to maintain a consistent access pattern within the same function call.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py
Sunhill666: If `negative_prompt_embeds_mask` isn't provided (is `None`), that block is safely skipped, and `encode_prompt` just returns `None` for the mask. The transformer model later receives this `None` mask and handles it natively by treating all tokens as valid.

Interestingly, while working on this, I found that some other variants (like `edit` and `inpaint`) were actually missing this exact `if prompt_embeds_mask is not None:` check and were crashing on `.repeat()`. I've added the same check to those pipelines in this PR as well so they all handle `None` masks gracefully now.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_controlnet_inpaint.py
Sunhill666: I'll revert the manual edits on the copied methods, run make fix-copies to sync them up properly, and update the PR.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage_controlnet_inpaint.py
Sunhill666: I removed the hard ValueError here because the base QwenImagePipeline actually allows passing embeds without masks, and the underlying transformer natively handles a None mask by treating all tokens as valid. Throwing a strict exception here was completely blocking users from doing exactly what this PR is fixing (passing negative_prompt_embeds without a mask to trigger True CFG).

That said, I totally agree we shouldn't just let it pass silently, especially since the text encoder's output often relies on masks for sequence packing. I'll swap these hard exceptions out for a logger.warning to let users know they should probably pass the mask if they have it. I'll get that updated!

tests/pipelines/qwenimage/test_qwenimage.py
Sunhill666: The intention here is explicitly to test the pipeline without providing the negative_prompt_embeds_mask (hence the test name test_true_cfg_without_negative_prompt_embeds_mask), to verify that our fix correctly catches the missing mask, issues the logger.warning, and successfully proceeds with CFG instead of throwing a hard ValueError or silently disabling CFG.

To make this intent clearer and avoid explicitly assigning None to the inputs dict, I will update the test to use inputs.pop(...) so that it relies directly on the pipeline's default None arguments.

Sunhill666:
Hi @sayakpaul,

Thanks for the very helpful feedback! I have updated the PR to address your comments. Here is a summary of the changes I just pushed:

1. Updated `check_inputs` across all `QwenImage` pipeline variants to emit a `logger.warning(...)` when `prompt_embeds_mask` is provided but `negative_prompt_embeds_mask` is missing.
2. Refactored the dummy input setup in `test_true_cfg_without_negative_prompt_embeds_mask`. Instead of explicitly setting `inputs["negative..."] = None`, I'm now using `inputs.pop("...", None)` to more naturally simulate missing arguments without causing confusion.
3. Replaced the hardcoded `"cpu"` explicitly with `torch_device` in the test files.
4. Corrected the `# Copied from` markers which previously contained typos holding back `make fix-copies`. All duplicated code is now properly synchronized.

Let me know if anything else needs adjusting!

sayakpaul:

Sunhill666:
@sayakpaul I've addressed all the points:

1. **Fixed the malformed `# Copied from` annotation** (removed the backticks).
2. **Reverted `.get()` usage for required keys**: Now using `model_inputs["input_ids"]` and `model_inputs["attention_mask"]` directly to avoid silent `NoneType` failures.
3. **Added missing blank lines** for code style consistency.
4. **Simplified tests**: Replaced `.get("image") if "image" in inputs else None` with just `.get("image")`.
5. Ran `make fix-copies` and `make style` successfully to sync the changes.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py
```diff
@@ -311,6 +311,16 @@ def check_inputs(
                 f" {negative_prompt_embeds}. Please make sure to only forward one of the two."
             )
 
+        if prompt_embeds is not None and prompt_embeds_mask is None:
+            logger.warning(
+                "If `prompt_embeds` are provided, `prompt_embeds_mask` also have to be passed. Make sure to generate `prompt_embeds_mask` from the same text encoder that was used to generate `prompt_embeds`."
+            )
```
dg845, line 320:
```suggestion
            logger.warning(
                "`prompt_embeds` is provided and `prompt_embeds_mask` is not provided, so the model will treat all"
                " prompt tokens as valid. If `prompt_embeds` contains padding, you should provide the padding mask as"
                " `prompt_embeds_mask`. Make sure to generate `prompt_embeds_mask` from the same text encoder that was"
                " used to generate `prompt_embeds`."
            )
```
Not sure about the exact wording, but I think the warning here should explain when not passing the corresponding mask is valid, and when passing it is necessary.

src/diffusers/pipelines/qwenimage/pipeline_qwenimage.py
```diff
@@ -311,6 +311,16 @@ def check_inputs(
...
+
+        if negative_prompt_embeds is not None and negative_prompt_embeds_mask is None:
+            logger.warning(
+                "If `negative_prompt_embeds` are provided, `negative_prompt_embeds_mask` also have to be passed. Make sure to generate `negative_prompt_embeds_mask` from the same text encoder that was used to generate `negative_prompt_embeds`."
+            )
```
dg845, line 328:
```suggestion
            logger.warning(
                "`negative_prompt_embeds` is provided and `negative_prompt_embeds_mask` is not provided, so the model"
                " will treat all negative prompt tokens as valid. If `negative_prompt_embeds` contains padding, you"
                " should provide the padding mask as `negative_prompt_embeds_mask`. Make sure to generate"
                " `negative_prompt_embeds_mask` from the same text encoder that was used to generate"
                " `negative_prompt_embeds`."
            )
```
Analogous suggestion to https://github.com/huggingface/diffusers/pull/13379#discussion_r3097494510 for `negative_prompt_embeds`/`negative_prompt_embeds_mask`.

dg845 (commented):
Thanks for the PR! Left some small suggestions :).

Sunhill666:
Hi @dg845, I've updated the warning messages for both `prompt_embeds` and `negative_prompt_embeds` as suggested.

dg845 (approved):
Thanks!

PR #13381, opened 2026-04-01
[Agents] Optimizations skill

asomoza:
# What does this PR do?

Add a skill for all the optimizations that can be done with diffusers. 

This is a lean skill and mostly dependent of the docs to keep it simple. It will look them locally if the repo is cloned and if not, it will go online.

Example:

```
I run the example for klein in this file: test_klein.py but I 
get a cuda OOM, what can we do to make it run?
```

```
Here's the memory picture for this model on your RTX 4090 Laptop (16 GB VRAM, 32 GB RAM):

  ┌──────────────────────┬───────────┐
  │      Component       │ BF16 size │
  ├──────────────────────┼───────────┤
  │ Transformer          │ 16.91 GB  │
  ├──────────────────────┼───────────┤
  │ Text Encoder (Qwen3) │ 15.26 GB  │
  ├──────────────────────┼───────────┤
  │ VAE                  │ 0.16 GB   │
  ├──────────────────────┼───────────┤
  │ Total                │ 32.32 GB  │
  └──────────────────────┴───────────┘

  The problem: both the transformer and text encoder individually exceed your VRAM, so pipe.to("cuda") or even model CPU offloading
  alone won't work (the text encoder alone would need 15.3 GB on-GPU during text encoding). You need to reduce the size of both
  components. Here are your two viable options:

  ---
  Option 1 — Best quality: layerwise casting (fp8) on both components + model CPU offloading
  - Transformer → ~7.6 GB, Text encoder → ~6.9 GB in fp8 storage
  - VRAM: ~10.6 GB | RAM: ~14.5 GB
  - Minimal quality loss (norm/embed layers stay in BF16)

  import torch
  from diffusers import Flux2KleinPipeline
  from diffusers.hooks import apply_layerwise_casting

  pipe = Flux2KleinPipeline.from_pretrained(
      "black-forest-labs/FLUX.2-klein-base-9B", torch_dtype=torch.bfloat16
  )
  pipe.transformer.enable_layerwise_casting(
      storage_dtype=torch.float8_e4m3fn, compute_dtype=torch.bfloat16
  )
  apply_layerwise_casting(
      pipe.text_encoder,
      storage_dtype=torch.float8_e4m3fn,
      compute_dtype=torch.bfloat16,
      skip_modules_pattern=["norm"],
  )

  ---
  Option 2 — Simplest / fastest inference: nf4 quantization on both components
  - Transformer → ~4.2 GB, Text encoder → ~3.8 GB
  - VRAM: ~11.2 GB (all on GPU, no offloading) | RAM: peaks at ~17 GB during loading, ~8 GB after
  - Noticeable quality degradation vs full precision

  import torch
  from diffusers import Flux2KleinPipeline, PipelineQuantizationConfig

      quant_kwargs={"load_in_4bit": True, "bnb_4bit_quant_type": "nf4"},
  )
  pipe = Flux2KleinPipeline.from_pretrained(
      "black-forest-labs/FLUX.2-klein-base-9B",
      quantization_config=pipeline_quant_config,
      torch_dtype=torch.bfloat16,
      device_map="cpu",
  )
  pipe.to("cuda")

  ---
  Which would you prefer? Option 1 gives you near-full quality; Option 2 is a one-liner load but with more quality degradation. I can
  also combine both (nf4 text encoder + layerwise casting transformer) if you want a middle ground.
```

## Who can review?

@sayakpaul


.ai/skills/optimizations/attention-backends.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Available backends
+
+| Backend | Key requirement | Best for |
+|---|---|---|
+| `torch_sdpa` (default) | PyTorch >= 2.0 | General use; auto-selects FlashAttention or memory-efficient kernels |
+| `flash_attention_2` | `flash-attn` package, Ampere+ GPU | Long sequences, training, best raw throughput |
+| `xformers` | `xformers` package | Older GPUs, memory-efficient attention |
+| `flex_attention` | PyTorch >= 2.5 | Custom attention masks, block-sparse patterns |
+| `sage_attention` | `sageattention` package | INT8 quantized attention for inference speed |
```
sayakpaul, line 15: We should take this information from https://huggingface.co/docs/diffusers/main/en/optimization/attention_backends. 

Prefer using `flash_hub` instead of `flash_attention_2`, etc. 

Should also include `_flash_3_hub`.

We could also just refer Claude to `./docs/source/en/optimization/attention_backends.md` so that it always has the latest info. I think this would be better. WDYT?

.ai/skills/optimizations/attention-backends.md
```diff
@@ -0,0 +1,40 @@
...
+## Debugging attention issues
+
+- **NaN outputs**: Check if your attention mask dtype matches the expected dtype. Some backends require `bool`, others require float masks with `-inf` for masked positions.
+- **Speed regression**: Profile with `torch.profiler` to verify the expected kernel is actually being dispatched. SDPA can silently fall back to the math kernel.
+- **Memory spike**: FlashAttention-2 is memory-efficient for long sequences but has overhead for very short ones. For short sequences, `torch_sdpa` with math fallback may use less memory.
+
+## Implementation notes
+
+- Models integrated into diffusers should use `dispatch_attention_fn` (not `F.scaled_dot_product_attention` directly) so that backend switching works automatically.
+- See the attention pattern in the `model-integration` skill for how to implement this in new models.
```
sayakpaul, line 40: I think we won't need this if we link to the attention backends documentation?

.ai/skills/optimizations/layerwise-casting.md
```diff
@@ -0,0 +1,68 @@
+# Layerwise Casting
+
+## Overview
```
sayakpaul, line 3: Same I guess? 

We could refer Claude to https://huggingface.co/docs/diffusers/main/en/optimization/memory? And briefly discuss a few?

.ai/skills/optimizations/memory-calculator.md
```diff
@@ -0,0 +1,298 @@
+# Memory Calculator
```
sayakpaul, line 1: How about using https://github.com/alvarobartt/hf-mem from our very own @alvarobartt?

.ai/skills/optimizations/memory-calculator.md
```diff
@@ -0,0 +1,298 @@
...
+- `S_leaf` = size of the largest leaf module
+- `S_total_lc` = S_total after applying layerwise casting to castable components (measured, not estimated — norm/embed layers are skipped so it's not exactly half)
+- `S_max_lc` = size of the largest component after layerwise casting (measured)
+- `A` = activation memory during forward pass (cannot be measured ahead of time — estimate conservatively):
+  - **Video models**: `A` scales with resolution and number of frames. A 5-second 960x544 video at 24fps can use ~7-8 GB. Higher resolution or more seconds = more activation memory.
+  - **Image models**: `A` scales with image resolution. A 1024x1024 image might use 2-4 GB, but 2048x2048 could use 8-16 GB.
+  - **Edit/inpainting models**: `A` includes the reference image(s) in addition to the generation activations, so budget extra.
+  - When in doubt, estimate conservatively: `A ≈ 5-8 GB` for typical video workloads, `A ≈ 2-4 GB` for typical image workloads. For high-resolution or long video, increase accordingly.
+
+## Step 2: Compute VRAM and RAM per strategy
```
sayakpaul, line 69: This should probably go to a separate `offloading.md` file?

.ai/skills/optimizations/memory-calculator.md
```diff
@@ -0,0 +1,298 @@
...
+| **VRAM** | `num_blocks_per_group * S_block + A` |
+| **RAM** | `S_total` (all weights on CPU, no pinned copy) |
+| **Speed** | Moderate — synchronous transfers per group |
+| **Quality** | Full precision |
+
+Tune `num_blocks_per_group` to fill available VRAM: `floor((VRAM - A) / S_block)`.
+
+### Group offloading: block_level (with stream)
+
+Streams force `num_blocks_per_group=1`. Prefetches the next block while the current one runs.
```
sayakpaul, line 102: Should also mention that the computation is overlapped with data transfer and thereby reducing latency. Referenced to https://huggingface.co/docs/diffusers/main/en/optimization/memory#offloading as well?

.ai/skills/optimizations/memory-calculator.md
```diff
@@ -0,0 +1,298 @@
...
+Combines reduced weight size with offloading. The offloaded weights are in fp8, so transfers are faster and pinned copies smaller.
+
+| | Estimate |
+|---|---|
+| **VRAM** | `num_blocks_per_group * S_block * 0.5 + A` (block_level) or `S_leaf * 0.5 + A` (leaf_level) |
+| **RAM** | `S_total * 0.5` (no stream) or `~S_total` (with stream, pinned copy of fp8 weights) |
+| **Speed** | Good — smaller transfers due to fp8 |
+| **Quality** | Slight degradation from fp8 |
+
+### Quantization (int4/nf4)
```
sayakpaul, line 187: This should probably be only in `quantization.md`?

.ai/skills/optimizations/memory-calculator.md
```diff
@@ -0,0 +1,298 @@
...
+| **RAM** | `~2.5-3 * S_total` (original weights + pinned copies + allocation overhead) |
+| **Speed** | Fast — overlaps transfer and compute |
+| **Quality** | Full precision |
+
+With `low_cpu_mem_usage=True`: RAM drops to `~S_total` (pins tensors on-the-fly instead of pre-pinning), but slower.
+
+With `record_stream=True`: slightly more VRAM (delays memory reclamation), slightly faster (avoids stream synchronization).
+
+> **Note on RAM estimates with streams:** Measured RAM usage is consistently higher than the theoretical `2 * S_total`. Pinned memory allocation, CUDA runtime overhead, and memory fragmentation add ~30-50% on top. Always use `~2.5-3 * S_total` when checking if the user has enough RAM for streamed offloading.
+
```
sayakpaul, line 116: Disk offload is another option which is worth mentioning here?

.ai/skills/optimizations/memory-calculator.md
```diff
@@ -0,0 +1,298 @@
...
+S_total_nf4  = sum of quantized components (nf4) + remaining components (bf16)
+S_max_int8   = max single component after int8 quantization
+S_max_nf4    = max single component after nf4 quantization
+```
+
+RAM requirement for quantization loading: `RAM >= S_largest_component_bf16` (full-precision weights
+must fit during quantization). If this doesn't hold, quantization is not viable unless pre-quantized
+checkpoints are available.
+
+### Quick decision flowchart
```
sayakpaul, line 258: I think this is quite valid for this document. But any guidance related to offloading, quantization, etc. should be moved to their own.

.ai/skills/optimizations/memory-calculator.md
```diff
@@ -0,0 +1,298 @@
...
+                RAM >= 3 * S_total? (enough for pinned copies + overhead)
+                  → YES: group offload leaf_level + stream (fast)
+                  → NO:
+                    RAM >= S_total?
+                      → YES: group offload leaf_level + stream + low_cpu_mem_usage
+                             or group offload block_level (no stream)
+                      → NO: Quantization required to reduce model size, then retry
+```
+
+Quantization path (evaluate in parallel with the above, not as a fallback):
```
sayakpaul, line 284: Should we provide any handholding on which quantization scheme should be used?

For example, if the hardware supports FP8, then it should be preferred over others, etc.?

.ai/skills/optimizations/quantization.md
```diff
@@ -0,0 +1,180 @@
+# Quantization
```
sayakpaul, line 1: Same as above. A lot of it could be offloaded to https://huggingface.co/docs/diffusers/main/en/quantization/overview

.ai/skills/optimizations/quantization.md
```diff
@@ -0,0 +1,180 @@
...
+### torchao (pipeline-level)
+
+```python
+from diffusers import PipelineQuantizationConfig, DiffusionPipeline
+
+quantization_config = PipelineQuantizationConfig(
+    quant_backend="torchao",
+    quant_kwargs={"quant_type": "int8_weight_only"},
+    components_to_quantize=["transformer"],
+)
```
sayakpaul, line 77: Let's switch to using something like:

```py
import torch
from diffusers import DiffusionPipeline, PipelineQuantizationConfig, TorchAoConfig
from torchao.quantization import Int8WeightOnlyConfig

    quant_mapping={"transformer": TorchAoConfig(Int8WeightOnlyConfig(group_size=128, version=2))}
)
    quantization_config=pipeline_quant_config,
    torch_dtype=torch.bfloat16,
    device_map="cuda"
)
```

.ai/skills/optimizations/reduce-memory.md
```diff
@@ -0,0 +1,213 @@
+# Reduce Memory
```
sayakpaul, line 1: We already have thorough usage in https://huggingface.co/docs/diffusers/main/en/optimization/memory. If something needs to be updated, let's update it there. This way, we serve general users of the library as well as the AI agents.

.ai/skills/optimizations/torch-compile.md
```diff
@@ -0,0 +1,72 @@
+# torch.compile
```
sayakpaul, line 1: Same. However, most of this content should be documented formally actually. Cc: @stevhliu could you help here?

Additionally, we could refer it to follow https://huggingface.co/docs/diffusers/main/en/optimization/speed-memory-optims for squeezing the last bit of performance.

.ai/skills/optimizations/SKILL.md
```diff
@@ -0,0 +1,113 @@
+---
+name: optimizations
```
sayakpaul, line 2: Nice. It might be worth including some of the edge cases we have encountered in the past and that way it doesn't run into the same context rot?

sayakpaul (commented):
Thanks a lot!

My main comments are about reusing the existing documentation as much as we can and removing duplicity which might be causing confusion for the agent. 

It'd then be nice to include some examples where this skill was useful so that we can build trust amongst users.

.ai/skills/optimizations/attention-backends.md
asomoza: oh yeah, for sure, the attention backends was mostly added by claude itself, since it wasn't my priority for this PR and also I didn't test this part. It for sure can be a link to the docs.

asomoza: oh wait, if we just install the diffusers wheel, we don't have the documentation, where are you expecting the model to get the docs? Also fetch online will prevent the skill to be used offline.

I also thought that this makes sense, but only if you have cloned the repo, it doesn't work for just a wheel install.

sayakpaul: Then we should have the users clone a diffusers copy as it's just better and far simpler than to duplicate content.

So, if it's offline, then prompt the users to clone a diffusers copy. If it's not then a fetch operation should suffice.

sayakpaul: But for offline agents, how would they access the skill as these are also not packaged right?

.ai/skills/optimizations/memory-calculator.md
asomoza: I'm fine with it but this will add an external dependency/install and also it doesn't take into account RAM usage and other factors like cuda streams etc. I'm fine either way, As a rule I always use the if we can do it without AI, don't use AI, but in this case we also need to calculate other factors that tool doesn't provide, so we can just do it all at the same time.

sayakpaul: Okay then let's not use the tool. Do you think we should do our own util, instead and have it under `utils`? That is better I guess?

.ai/skills/optimizations/attention-backends.md
asomoza: ~yeah, you can run a complete offline agent with claude code and it should use the skill but haven't tested it. Probably to make it simpler lets just assume they will be online and point to the online docs only? Cloning the repo seems a lot more involved and not sure if users will want that, and also we will have to check for git installation, permissions, etc.~

sayakpaul: Sure sounds good. But how can Claude be fully offline, though? Or do mean using something like OpenCode with a local model deployed?

.ai/skills/optimizations/layerwise-casting.md
asomoza: do you think layerwise will change? The statics parts IMO should be on the skill to prevent fetching everything no? If not, this will be just a simple skill that tells the LLMs to just read the docs.

Also the reason I did it separately is because at least claude never suggest using it otherwise, specially because we don't have the `when to use` text in the docs, we can also add that to the docs as a solution.

sayakpaul: I mean layerwise is already in the docs https://huggingface.co/docs/diffusers/main/en/optimization/memory#layerwise-casting.

So, we could introduce Claude to that skill by referring to the docs and perhaps provide a simple intro. WDYT?

sayakpaul:
Don't want to complicate it further but I don't think we sell our cool support for https://huggingface.co/docs/diffusers/main/en/optimization/cache enough.

.ai/skills/optimizations/attention-backends.md
asomoza: I now see what you meant, IMO this is more something that we should publish in this [repo](https://github.com/huggingface/skills) than here. But yeah, if we assume they will use the skill here, we can just link to the .md local docs.

asomoza:
thanks @sayakpaul I think you're right, if we are going to use the skill here, we can mostly link all the info the model needs to the docs, I'll refactor it and do some tests again.

I was thinking more of a installable skill in my mind.

.ai/skills/optimizations/attention-backends.md
sayakpaul: Yeah for now, I think it's valuable keep the skills repo-specific. This way, they are more easily discoverable.

> But yeah, if we assume they will use the skill here, we can just link to the .md local docs.

That's a fair assumption I guess? How is the skill accessed otherwise then? Can we install it or something?

sayakpaul:
Installable skills also sound good. In that case, we can always refer to the links and assume online access?

.ai/skills/optimizations/attention-backends.md
asomoza: yeah, skills can be installed, hf also has a [cli installer](https://github.com/huggingface/hf-skills), you can install skills for a project, for your user or for enterprise. The project ones are the rarest ones and usually for working on that project not for using them.

sayakpaul: Maybe we can ship without having to install first and based on the feedback we can iterate? I don't think things will change too much. WDYT?

asomoza: yeah, lets go with that

.ai/skills/optimizations/memory-calculator.md
asomoza: I used the tool, it makes the skill leaner and with what claude learned with the previous calculations, it can safely `guess` the usage, also we're assuming the user is online, so no issues here.

.ai/skills/optimizations/memory-calculator.md
asomoza: now is leaner and almost all of this section if offloading, so using a separate md file probably is not worth it anymore

.ai/skills/optimizations/memory-calculator.md
asomoza: we can add this when we have support for fp8 and nvfp4 without the need of nightlies? Also this is mostly linked to the docs now. Also fp8 is mostly safe but quality wise sometimes int8 and int4 have better quality especially with `sdnq` than fp8 and nvfp4, for example, LTX2

.ai/skills/optimizations/SKILL.md
asomoza: the ones I encountered are already in some way in this skill, we can add more when we find more? I don't have any other one I can add.

PR #13388, opened 2026-04-02
fix: correct grammar 'same type to' → 'same type as' in error messages

Ricardo-M-L:
160 occurrences across 158 pipeline files use the incorrect phrase "should be the same type to" instead of "should be the same type as" in the negative_prompt type check error message.

Example (before):
```python
f"`negative_prompt` should be the same type to `prompt`, but got {type(negative_prompt)} != {type(prompt)}."
```

After:
```python
f"`negative_prompt` should be the same type as `prompt`, but got {type(negative_prompt)} != {type(prompt)}."
```

2 files in the codebase already used the correct "same type as" form.


Ricardo-M-L:
Closing this batch-typo PR to reduce noise in my open diffusers queue — keeping the substantive bug fixes (#13434 crash fixes, #13449 CFG regression) front and center. Happy to reopen if a maintainer would prefer otherwise.

PR #13405, opened 2026-04-03
Remove references to torchao's AffineQuantizedTensor

andrewor14:
**Summary:** TorchAO recently deprecated AffineQuantizedTensor and related classes (https://github.com/pytorch/ao/issues/2752). These will be removed in the next release. We should remove references of these classes in diffusers before then.

**Test Plan:**
python -m pytest -s -v tests/quantization/torchao/test_torchao.py

### DIFF
--- a/examples/dreambooth/train_dreambooth_lora_flux2.py
+++ b/examples/dreambooth/train_dreambooth_lora_flux2.py
@@ -1749,8 +1749,8 @@ def main(args):
                     model_input = latents_cache[step].mode()
                 else:
                     with offload_models(vae, device=accelerator.device, offload=args.offload):
-                        pixel_values = batch["pixel_values"].to(dtype=vae.dtype)
-                    model_input = vae.encode(pixel_values).latent_dist.mode()
+                        pixel_values = batch["pixel_values"].to(device=accelerator.device, dtype=vae.dtype)
+                        model_input = vae.encode(pixel_values).latent_dist.mode()
 
                 model_input = Flux2Pipeline._patchify_latents(model_input)
                 model_input = (model_input - latents_bn_mean) / latents_bn_std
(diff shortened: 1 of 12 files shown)


andrewor14:
Hi @sayakpaul can you help review this?

src/diffusers/quantizers/torchao/torchao_quantizer.py
```diff
@@ -133,19 +133,10 @@ def fuzzy_match_size(config_name: str) -> str | None:
...
-
-
 def _linear_extra_repr(self):
-    weight = _quantization_type(self.weight)
+    from torchao.utils import TorchAOBaseTensor
+
+    weight = self.weight.__class__.__name__ if isinstance(self.weight, TorchAOBaseTensor) else None
     if weight is None:
         return f"in_features={self.weight.shape[1]}, out_features={self.weight.shape[0]}, weight=None"
     else:
```
sayakpaul, line 142: Is this sufficient?

andrewor14: Yeah should be, all new tensor subclasses are TorchAOBaseTensor

jerryzh168: @andrewor14 I think we can just do this: https://github.com/pytorch/ao/blob/d26bbae1c7910ecc20343d35750408d7fb3c6097/torchao/quantization/utils.py#L765

jerryzh168: seems OK to make this function public as well

andrewor14: hmm I feel that function is a bit internal to torchao, not sure if we should expose it

PR #13412, opened 2026-04-04
[Scheduler] Add monotonicity validation and tests for LTXEulerAncestr…

### LINKED ISSUE #13411: `LTXEulerAncestralRFScheduler.set_timesteps(sigmas=...)` does not validate monotonicity, causing silent incorrect denoising
## Describe the bug

`LTXEulerAncestralRFScheduler.set_timesteps` accepts an externally-supplied `sigmas` argument without validating that the schedule is monotonically non-increasing. When `step()` is called on a non-monotone schedule, the ancestral RF decomposition computes `sigma_down` outside `[0, 1]` and `alpha_down < 0`, which violates the CONST parametrization invariant. The output is numerically finite — no exception, no NaN, no warning — but denoising does not occur. The corruption is silent.

From [`PHILOSOPHY.md`](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md), line 32:

> "Raising concise error messages is preferred to silently correct erroneous input. Diffusers aims at teaching the user, rather than making the library as easy to use as possible."

## Reproduction

```python
import torch
from diffusers import LTXEulerAncestralRFScheduler

scheduler = LTXEulerAncestralRFScheduler()

# Non-monotone schedule: sigma increases at step 0 → 1.
# A misconfigured ComfyUI workflow can emit exactly this kind of list.
scheduler.set_timesteps(sigmas=[0.2, 0.8, 0.5, 0.0])  # should raise — does not

torch.manual_seed(0)
sample       = torch.randn(1, 4, 8, 8)
model_output = torch.zeros_like(sample)  # model predicts zero signal

out = scheduler.step(model_output, scheduler.timesteps[0], sample)

print(f"Input  norm : {sample.norm():.4f}")
print(f"Output norm : {out.prev_sample.norm():.4f}")
print(f"Ratio       : {(out.prev_sample.norm() / sample.norm()).item():.3f}x")
# Expected: small positive ratio converging toward sigma_next/sigma ≈ 0.25
# Actual:   ratio ≈ 1.45, sign-inverted — latent is corrupted, not denoised
```

## Expected behaviour

`set_timesteps` raises `ValueError` immediately when `sigmas` is not monotonically non-increasing, before any internal state is mutated.

## Actual behaviour — full numerical trace

With `sigma = 0.2`, `sigma_next = 0.8`, `eta = 1.0` (from `step()` in `scheduling_ltx_euler_ancestral_rf.py`):

```
downstep_ratio = 1.0 + (sigma_next / sigma - 1.0) * eta
               = 1.0 + (0.8 / 0.2 - 1.0) * 1.0
               = 4.0          ← should be in (0, 1] for a valid denoising step

sigma_down     = sigma_next * downstep_ratio
               = 0.8 * 4.0 = 3.2     ← outside [0, 1]; CONST parametrization undefined

sigma_ratio    = sigma_down / sigma = 3.2 / 0.2 = 16.0

x_euler = 16.0 * sample + (−15.0) * denoised    ← 16× amplification

alpha_down = 1 − sigma_down = 1 − 3.2 = −2.2    ← negative; undefined in CONST space
scale      = alpha_ip1 / alpha_down = 0.2 / −2.2 = −0.091  ← sign flip

# renoise_coeff = sqrt(clamp(sigma_next² − sigma_down² · alpha_ip1² / alpha_down², min=0))
renoise_coeff ≈ 0.745

prev_sample ≈ −0.091 × x_euler + 0.745 × noise
            ≈ −1.45 × sample + 1.36 × denoised + 0.745 × noise
```

The sample coefficient is **−1.45** (sign-inverted, slightly amplified) rather than a small positive value. The output is finite, plausibly shaped, and completely wrong.

## Why `>=` not `>`

The check should use `sigmas[:-1] >= sigmas[1:]` (non-strict). Strict monotonicity would reject plateau steps — consecutive equal sigmas that are intentional in img2img partial schedules using `set_begin_index`. Non-strict preserves that use case while catching every reversed or partially-reversed schedule.

## Proposed fix

```python
# set_timesteps(), after the ndim guard, before the terminal-sigma warning:
if len(sigmas_tensor) > 1 and not (sigmas_tensor[:-1] >= sigmas_tensor[1:]).all():
    sig_list = sigmas_tensor.tolist()
    sig_repr = str(sig_list) if len(sig_list) <= 8 else f"{sig_list[:4]} ... {sig_list[-4:]} (len={len(sig_list)})"
    raise ValueError(
        f"`sigmas` must be monotonically non-increasing (each entry >= the next), got {sig_repr}"
    )
```

A companion range check is also needed — the CONST parametrization requires σ ∈ `[0, 1]`, which `set_timesteps` does not currently enforce.

**Note:** `FlowMatchEulerDiscreteScheduler` is not affected — it generates its sigma schedule internally and does not expose a raw `sigmas=` argument. The blast radius is asymmetric.

## Additional context

There is currently no test file for `LTXEulerAncestralRFScheduler` in `tests/schedulers/`, and the scheduler is absent from the API reference `_toctree.yml` despite being publicly exported.

I have a PR ready that addresses all of the above: monotonicity and range checks, `eta` boundary validation, 16 unit tests, an API docs page, and the toctree entry. Happy to open it once this issue is acknowledged.

/cc @yiyixuxu

VittoriaLanzo:
## What does this PR do?

Fixes #13411.

`LTXEulerAncestralRFScheduler.set_timesteps(sigmas=...)` silently accepted non-monotone sigma schedules. In `step()`, a reversed schedule computes:

```
downstep_ratio = 1 + (sigma_next / sigma - 1) * eta = 4.0   # should be in (0, 1]
sigma_down     = sigma_next * downstep_ratio         = 3.2   # outside [0, 1]
sigma_ratio    = sigma_down / sigma                  = 16.0
```

The Euler update becomes `16 × sample − 15 × denoised` — the latent is amplified, not denoised. `alpha_down = 1 − 3.2 = −2.2` (negative), violating the CONST parametrization invariant. No error, no NaN, no warning. Silent corruption.

This is a realistic failure mode: the `sigmas=` path is the ComfyUI interop surface, where raw user-defined sigma lists are passed directly.

Per [`PHILOSOPHY.md`](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md), line 32: *"Raising concise error messages is preferred to silently correct erroneous input."*

---

### `src/diffusers/schedulers/scheduling_ltx_euler_ancestral_rf.py`

- **Monotonicity check**: raises `ValueError` when `sigmas` is not non-increasing. Uses `>=` (non-strict) so plateau steps remain valid for `set_begin_index` img2img workflows.
- **Range check**: raises `ValueError` when `sigmas[0] > 1.0` — the CONST parametrization requires σ ∈ `[0, 1]`.
- **`eta` validation**: `eta < 0` raises in `__init__`. `eta > 1` on a coarse schedule clamps `sigma_down` to `[0, sigma_next]` with a one-time `logger.warning` (deduplication via `_sigma_down_warned` flag).

### `tests/schedulers/test_scheduler_ltx_euler_ancestral_rf.py` *(new)*

16 unit tests (no GPU required):

| Group | Tests |
|---|---|
| `set_timesteps` validation | monotone accepted, non-monotone raises, fully ascending raises, plateau valid, auto-generated length+monotonicity, sigma > 1 raises |
| `step()` invariants | fp16/fp32 dtype preserved, shape preserved, `return_dict=False` tuple, `eta=0` deterministic, final step formula, `eta < 0` raises at construction, `eta > 1` clamps+warns+stays finite, `_step_index` advances, past-end step returns finite output |

### `docs/source/en/api/schedulers/ltx_euler_ancestral_rf.md` *(new)*

API reference page. The scheduler was publicly exported but absent from the docs. Includes `[[autodoc]]` directives for `LTXEulerAncestralRFScheduler` and `LTXEulerAncestralRFSchedulerOutput`, `eta` valid range documentation, and a See Also pointer to `FlowMatchEulerDiscreteScheduler`.

### `docs/source/en/_toctree.yml`

Alphabetical entry between `LMSDiscreteScheduler` and `PNDMScheduler`.

---

## Design decisions

**Why `raise` and not `logger.warning`?**  
A non-monotone schedule has no valid use case — `sigma_down` exits `[0, 1]` regardless of `eta` or `s_noise`. Terminal-sigma warnings cover bounded, recoverable edge cases. This does not. `PHILOSOPHY.md` is explicit.

**Why `>=` not `>`?**  
Plateau steps (`[1.0, 1.0, 0.5, 0.0]`) are intentional in `set_begin_index` img2img workflows. `test_set_timesteps_plateau_is_valid` covers this.

**Why not `FlowMatchEulerDiscreteScheduler`?**  
That scheduler generates sigmas internally via `linspace`-derived transforms and cannot receive a non-monotone external list. The blast radius is asymmetric.

**Why clamp `sigma_down` instead of raising for `eta > 1`?**  
`eta > 1` is an experimental but not inherently wrong configuration. Raising would be a breaking change. Clamping + a one-time warning keeps the computation well-defined while signalling the issue to the user.

---

## Before submitting
- [x] Read contributor guideline
- [x] Read `PHILOSOPHY.md` — fix is a direct application of the stated principle
- [x] No modifications outside the 4 listed files
- [x] Documentation updated (new docs page + toctree entry)
- [x] Tests written (16 new tests, no GPU required)

## Who can review?
@yiyixuxu

### DIFF
--- a/docs/source/en/_toctree.yml
+++ b/docs/source/en/_toctree.yml
@@ -737,6 +737,8 @@
       title: LCMScheduler
     - local: api/schedulers/lms_discrete
       title: LMSDiscreteScheduler
+    - local: api/schedulers/ltx_euler_ancestral_rf
+      title: LTXEulerAncestralRFScheduler
     - local: api/schedulers/pndm
       title: PNDMScheduler
     - local: api/schedulers/repaint
--- a/docs/source/en/api/schedulers/ltx_euler_ancestral_rf.md
+++ b/docs/source/en/api/schedulers/ltx_euler_ancestral_rf.md
@@ -1,0 +1,45 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# LTXEulerAncestralRFScheduler
+
+The `LTXEulerAncestralRFScheduler` implements a K-diffusion-style Euler-Ancestral sampler
+for flow / CONST parameterization, closely mirroring ComfyUI's `sample_euler_ancestral_RF`
+implementation used for [LTX-Video](https://huggingface.co/docs/diffusers/api/pipelines/ltx_video).
+
+The scheduler operates on a normalized sigma schedule σ ∈ [0, 1] and reconstructs the clean
+estimate as `x0 = x_t − σ_t · v_t` (CONST parametrization). Stochastic noise reinjection is
+controlled by `eta` (`eta=0` gives a deterministic Euler step; `eta=1` matches ComfyUI's
+default RF behavior).
+
+This scheduler is used by [`LTXPipeline`], [`LTXImageToVideoPipeline`], and
+[`LTXConditionPipeline`].
+
+The `eta` parameter must be >= 0. `eta=0` gives a deterministic (DDIM-like) Euler step;
+`eta=1` matches ComfyUI's default RF behavior. Values above 1 are accepted but trigger a
+one-time warning when the schedule step is too coarse to keep `sigma_down` non-negative.
+
+<Tip>
+
+See also [`FlowMatchEulerDiscreteScheduler`], which this scheduler delegates to for
+auto-generated sigma schedules and shares config compatibility with via `_compatibles`.
+
+</Tip>
+
+## LTXEulerAncestralRFScheduler
+[[autodoc]] LTXEulerAncestralRFScheduler
+
+## LTXEulerAncestralRFSchedulerOutput
+[[autodoc]] schedulers.scheduling_ltx_euler_ancestral_rf.LTXEulerAncestralRFSchedulerOutput
--- a/src/diffusers/schedulers/scheduling_ltx_euler_ancestral_rf.py
+++ b/src/diffusers/schedulers/scheduling_ltx_euler_ancestral_rf.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Lightricks and The HuggingFace Team. All rights reserved.
+# Copyright 2025 Lightricks, Vittoria Lanzo and The HuggingFace Team. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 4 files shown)


VittoriaLanzo:
Hi @yiyixuxu, would you have bandwidth to take a look  when convenient? Happy to address any feedback. Thanks!

yiyixuxu:
hi @VittoriaLanzo thanks for the PR but we will not accept it
The point of this parameter is for user to supply a custom schedule, i.e. any schedule they'd like to try, it is not meaningful to add any checks here

VittoriaLanzo:
Thanks for the clarification @yiyixuxu . I see your point about user flexibility for custom schedules. I'll keep this in mind for future contributions.

PR #13427, opened 2026-04-06
[torchao] Fix `record_stream` with torchao and group offloading

asomoza:
# What does this PR do?

Fixes the use of `record_stream` with `torchao` and `group offload`. Without this, the resulting image is wrong.

|Without this PR|With this PR|
|---|---|
|flux_benchmark_dynamic_eager_bs1_goff_leaf_level_output_bad|flux_benchmark_dynamic_eager_bs1_output|

## Who can review?

@sayakpaul

### DIFF
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -166,6 +166,13 @@ class ModuleGroup:
         else:
             self.cpu_param_dict = self._init_cpu_param_dict()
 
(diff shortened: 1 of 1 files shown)


asomoza:
I did a double check and this is not a correct fix, the VRAM usage jumps from 5.67 GB to 13.71GB

asomoza:
actually, with the torchao PR this works ok, so closing.

PR #13428, opened 2026-04-06
FLUX.2 small decoder

huemin-art:
# What does this PR do?

Adds an optional `decoder_block_out_channels` parameter to `AutoencoderKLFlux2` so the decoder can use different channel widths than the encoder. This is needed to support the FLUX.2 small decoder VAE (`black-forest-labs/FLUX.2-small-decoder`). When not set, falls back to `block_out_channels` (no behavior change).

## Who can review?

@yiyixuxu


yiyixuxu (approved):
thanks!

PR #11311, opened 2025-04-14
Modular Diffusers Guiders

a-r-r-o-w:
The following methods are currently supported:
- [x] Classifier-free Guidance (https://huggingface.co/papers/2207.12598)
- [x] Skip-layer Guidance (https://github.com/Stability-AI/sd3.5)
- [x] Adaptive Projected Guidance (https://huggingface.co/papers/2410.02416)
- [x] Classifier-free Guidance Zero* (https://huggingface.co/papers/2503.18886)
- [x] Perturbed Attention Guidance (https://huggingface.co/papers/2403.17377)
- [x] Spatio-temporal Guidance (https://huggingface.co/papers/2411.18664)
- [x] AutoGuidance (https://huggingface.co/papers/2406.02507)
- [x] CFG++ (requires scheduler level changes) (https://huggingface.co/papers/2406.08070)
- [x] Smoothed Energy Guidance (requires new attention processor per model) (https://huggingface.co/papers/2408.00760)
- [ ] Self-attention Guidance (requires new attention processor per model) (https://huggingface.co/papers/2210.00939)
- [ ] Attend and Excite (requires new attention processor per model) (https://huggingface.co/papers/2301.13826)
- [x] Tangential CFG (https://huggingface.co/papers/2503.18137)

Note: PAG is implemented as Skip Layer Guidance and does not have its own guider implementation. The equivalent SLG initialization is:

```python
from diffusers import SkipLayerGuidance, LayerSkipConfig

config = LayerSkipConfig(indices=[2, 9], fqn="mid_block.attentions.0.transformer_blocks", skip_attention=False, skip_ff=False, skip_attention_scores=True)
slg = SkipLayerGuidance(guidance_scale=5.0, skip_layer_guidance_scale=2.5, skip_layer_config=config)
```

Note: STG is also implemented as Skip Layer Guidance:
- STG-r: `skip_attention=False, skip_ff=True, skip_attention_scores=False`
- STG-a: `skip_attention=True, skip_ff=False, skip_attention_scores=False`
- STG-t: `skip_attention=True, skip_ff=True, skip_attention_scores=False`
- STG-v:`skip_attention=False, skip_ff=False, skip_attention_scores=True` (essentially PAG)

Note: You can use different SLG configs for different parts of the model. Create multiple configs and pass as a list to `skip_layer_config`

<table align=center>
  <thead>
    <tr>
      <th>APG</th>
      <th>CFG</th>
      <th>CFGZ</th>
[...]

PR #13434, opened 2026-04-08
fix: correct AttributeError in TangentialCFG and NameError in FrequencyDecoupledGuidance

Ricardo-M-L:
## Summary

Fixes two functional bugs in the experimental guiders module, both introduced in #11311 (Modular Diffusers Guiders):

### Bug 1: `TangentialClassifierFreeGuidance.is_conditional` — AttributeError

The `is_conditional` property references `self._num_outputs_prepared`, which does not exist on the class or its base class `BaseGuidance`. Every other guider uses `self._count_prepared` (defined and managed by `BaseGuidance`). Accessing `is_conditional` on a `TangentialClassifierFreeGuidance` instance raises:

```
AttributeError: 'TangentialClassifierFreeGuidance' object has no attribute '_num_outputs_prepared'
```

**Fix:** Replace `self._num_outputs_prepared` with `self._count_prepared`.

### Bug 2: `FrequencyDecoupledGuidance.forward` — NameError / wrong pyramid level

In the `forward` method's loop over pyramid levels, the `else` branch (when FDG is disabled for a level) appends `pred_cond_freq` to the guided pyramid. However, `pred_cond_freq` is only assigned inside the `if` branch. This causes:

- **NameError** if the very first pyramid level has FDG disabled (variable never assigned)
- **Silent wrong results** if a prior level assigned it — the wrong level's frequency data gets used

**Fix:** Replace `pred_cond_freq` with `pred_cond_pyramid[level]` to correctly reference the current level's conditional prediction.

## Test plan

- [x] Verified `_num_outputs_prepared` is not defined anywhere in the codebase; `_count_prepared` is the correct attribute used by all other guiders
- [x] Verified `pred_cond_freq` is only assigned inside the `if` branch, confirming the `else` branch will fail or use stale data
- [x] Checked no existing open PRs address these issues

### DIFF
--- a/src/diffusers/guiders/frequency_decoupled_guidance.py
+++ b/src/diffusers/guiders/frequency_decoupled_guidance.py
@@ -275,7 +275,7 @@ class FrequencyDecoupledGuidance(BaseGuidance):
                     pred_guided_pyramid.append(pred)
                 else:
                     # Add the current pred_cond_pyramid level as the "non-FDG" prediction
-                    pred_guided_pyramid.append(pred_cond_freq)
+                    pred_guided_pyramid.append(pred_cond_pyramid[level])
 
             # Convert from frequency space back to data (e.g. pixel) space by applying inverse freq transform
             pred = build_image_from_pyramid(pred_guided_pyramid)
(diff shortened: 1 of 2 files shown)


Ricardo-M-L:
Friendly bump — this is a tiny (+2/-2 across 2 files) crash fix: `TangentialCFG` hits `AttributeError` and `FrequencyDecoupledGuidance` hits `NameError` due to a referenced-but-undefined symbol. Both are outright crashes when the classes are instantiated. Happy to rebase if needed — thanks for any review time you can spare.

PR #13448, opened 2026-04-12
Improve `trust_remote_code`

### LINKED ISSUE #13446: Insufficient trust check for custom_pipeline parameter of DiffusionPipeline.from_pretrained method
### Describe the bug

The `DiffusionPipeline.from_pretrained` API in the Diffusers library allows users to specify a `custom_pipeline`, which can point to an external Hugging Face repository containing a `pipeline.py`. This feature enables flexible extension of pipelines but may raise critical security concerns.

The current implementation enforces the `trust_remote_code` check **only on the main model repository** (`pretrained_model_name`), but **fails to validate the external repository specified via `custom_pipeline`**.

As a result, even when `trust_remote_code=False`, Diffusers may still download and execute arbitrary Python code from an attacker-controlled repository referenced by `custom_pipeline`. This creates a bypass of the intended security mechanism, leading to silent remote code execution.

### Root Cause

https://github.com/huggingface/diffusers/blob/dc8d9032171c83741fd37ed2b12bc9d8274464f3/src/diffusers/pipelines/pipeline_utils.py#L1676-L1704

The check does not inspect the external repository referenced by `custom_pipeline`. Therefore, If `custom_pipeline` points to another repository (e.g., `XManFromXlab/diffuser-custom-pipeline`), and the main repo does not contain the corresponding `.py` file, then `load_pipe_from_hub` evaluates to `False`. As a result, the security check is skipped entirely, and diffusers will proceed to fetch the external repository, load and execute its `pipeline.py`. This results in a trust boundary violation, where remote code execution occurs without explicit user consent.

### Reproduction

I created an example model repository on HuggingFace Hub for demonstration: `XManFromXlab/diffuser-custom-pipeline`. Note that the `trust_remote_code` flag is set to `False`. 
```python

benign_repo = "google/ddpm-cifar10-32"
evil_repo = "XManFromXlab/diffuser-custom-pipeline"
DiffusionPipeline.from_pretrained(
    benign_repo, custom_pipeline=evil_repo, trust_remote_code=False
)
```

In this example, it will print the following warning messages: 

```bash
!!!!!!! Execute Malicious Payload !!!!!!!
!!!!!!! Execute Malicious Payload !!!!!!!
!!!!!!! Execute Malicious Payload !!!!!!!
!!!!!!! Execute Malicious Payload !!!!!!!
!!!!!!! Execute Malicious Payload !!!!!!!
```

### Logs

```shell

```

### System Info

- 🤗 Diffusers version: 0.38.0.dev0
- Platform: Linux-6.8.0-88-generic-x86_64-with-glibc2.39
- Python version: 3.12.3
- PyTorch version (GPU?): 2.11.0+cu130 (False)
- Jax version: not installed
- Huggingface_hub version: 1.10.1
- Transformers version: not installed
- Accelerate version: not installed
- PEFT version: not installed
- Safetensors version: 0.7.0
- Accelerator: NVIDIA H200 NVL, 143771 MiB
NVIDIA H200 NVL, 143771 MiB
NVIDIA H200 NVL, 143771 MiB

### Who can help?

@sayakpaul @DN6

hlky:
# What does this PR do?

As per #13446 `trust_remote_code` fails under several circumstances:

- `pretrained_model_name_or_path` as Hub repo A and `custom_pipeline` as Hub repo B, `trust_remote_code` is bypassed and remote code runs from repo B
```python

DiffusionPipeline.from_pretrained(
    "google/ddpm-cifar10-32", custom_pipeline="XManFromXlab/diffuser-custom-pipeline", trust_remote_code=False
)

```
- `pretrained_model_name_or_path` as local directory and `custom_pipeline` as Hub repo B, `trust_remote_code` is **never checked** and remote code runs from repo B
```python

from huggingface_hub import snapshot_download

snapshot_path = snapshot_download(repo_id="google/ddpm-cifar10-32")
DiffusionPipeline.from_pretrained(
    snapshot_path, custom_pipeline="XManFromXlab/diffuser-custom-pipeline", trust_remote_code=False
)

```
- `pretrained_model_name_or_path` as local directory with custom components, `trust_remote_code` is never checked
```python

from huggingface_hub import snapshot_download

snapshot_path = snapshot_download(repo_id="hf-internal-testing/tiny-sdxl-custom-components")
    snapshot_path, trust_remote_code=False
)
assert pipeline.config.unet == ("diffusers_modules.local.my_unet_model", "MyUNetModel")
assert pipeline.config.scheduler == ("diffusers_modules.local.my_scheduler", "MyScheduler")

```

This moves `trust_remote_code` checks into `get_cached_module_file` where the actual custom module loading takes place. `trust_remote_code` is passed from several code paths for complete coverage.

I've added 3 separate `ValueError` in `get_cached_module_file` to account for the different sources of remote code: `local`, `git` and `hub`. `git` path could be considered trusted as these are from `https://huggingface.co/datasets/diffusers/community-pipelines-mirror`, `hub` path covers "custom pipelines" and `local` path is reached for "custom components" (these are added to the `allowed files` in `download` so become local files).

With PR the above 3 cases are resolved:
```
The repository for XManFromXlab/diffuser-custom-pipeline contains custom code in pipeline.py which must be executed to correctly load the model. You can inspect the repository content at https://hf.co/XManFromXlab/diffuser-custom-pipeline/blob/main/pipeline.py.
Please pass the argument `trust_remote_code=True` to allow custom code to be run.
```
```
The repository for XManFromXlab/diffuser-custom-pipeline contains custom code in pipeline.py which must be executed to correctly load the model. You can inspect the repository content at https://hf.co/XManFromXlab/diffuser-custom-pipeline/blob/main/pipeline.py.
Please pass the argument `trust_remote_code=True` to allow custom code to be run.
```
```
ValueError: The directory C:\Users\user\.cache\huggingface\hub\models--hf-internal-testing--tiny-sdxl-custom-components\snapshots\ce2b9d3f819e7791f53053646ebe37d7e87d73d3\unet contains custom code in my_unet_model.py which must be executed to correctly load the model. You can inspect the file content at C:\Users\user\.cache\huggingface\hub\models--hf-internal-testing--tiny-sdxl-custom-components\snapshots\ce2b9d3f819e7791f53053646ebe37d7e87d73d3\unet\my_unet_model.py.
Please pass the argument `trust_remote_code=True` to allow custom code to be run.
```

Fixes #13446

## Who can review?


src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1674,21 +1678,6 @@ def download(cls, pretrained_model_name, **kwargs) -> str | os.PathLike:
                 custom_class_name = config_dict["_class_name"][1]
 
             load_pipe_from_hub = custom_pipeline is not None and f"{custom_pipeline}.py" in filenames
```
hlky, line 1690: Just to note this remains to control `repo_id`:

https://github.com/huggingface/diffusers/blob/dc8d9032171c83741fd37ed2b12bc9d8274464f3/src/diffusers/pipelines/pipeline_utils.py#L1699

https://github.com/huggingface/diffusers/blob/dc8d9032171c83741fd37ed2b12bc9d8274464f3/src/diffusers/pipelines/pipeline_loading_utils.py#L441-L458

It helps distinguish between:
a) `custom_pipeline` is e.g. `my_pipeline` and that filename exists in `pretrained_model_name`'s files
b) `custom_pipeline` is Hub repo (and `pipeline.py` is used)

Maybe could be renamed `load_pipe_from_hub` -> `hub_contains_custom_pipeline`?

tests/pipelines/test_pipelines.py
```diff
@@ -1011,8 +1011,15 @@ def test_get_pipeline_class_from_flax(self):
 
 class CustomPipelineTests(unittest.TestCase):
     def test_load_custom_pipeline(self):
+        with self.assertRaises(ValueError):
```
sayakpaul, line 1014: Should we investigate the `ValueError` messaging as well (it should have something related to the use of `trust_remote_code` or not something else)?

sayakpaul: And on `main`, this should not have yielded a `ValueError`, right? That is how we know, for one instance, that it's broken.

src/diffusers/utils/dynamic_modules_utils.py
```diff
@@ -254,6 +254,7 @@ def get_cached_module_file(
     revision: str | None = None,
     local_files_only: bool = False,
     local_dir: str | None = None,
+    trust_remote_code: bool = False,
```
sayakpaul, line 257: Would it make sense to add the `ValueError` to the caller sites of `get_cached_module_file` instead? Because the function itself isn't specifically tied to custom pipelines, I think.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1674,21 +1678,6 @@ def download(cls, pretrained_model_name, **kwargs) -> str | os.PathLike:
...
-            load_components_from_hub = len(custom_components) > 0
```
sayakpaul, line 1687: Why is this going?

hlky: See case 3 - this code is in `download` which is only reached when we use a Hub path with `from_pretrained`. It is replaced by the check in `get_cached_module_file`, specifically the `local` code path.

Consider this scenario:

```sh
hf download rotcasuoicilam/SuperCoolNewModel --local-dir rotcasuoicilam/SuperCoolNewModel
```
```python
pipeline = DiffusionPipeline.from_pretrained("rotcasuoicilam/SuperCoolNewModel")
```

`rotcasuoicilam/SuperCoolNewModel` contains malicious custom components, user downloads the Hub repo assuming it is safe, Diffusers loads the custom components without the user's consent.

sayakpaul: I didn't get it. 

```bash
hf download rotcasuoicilam/SuperCoolNewModel --local-dir rotcasuoicilam/SuperCoolNewModel
```

is agnostic to `DiffusionPipeline.from_pretrained(...)`.

hlky: 1. A malicious actor uploads a model with malicious custom components
2. Either:
   2a. The user follows instructions that say to download the model first then run from the local path
   2b. The user chooses to download the model first out of personal preference
3. `DiffusionPipeline.from_pretrained(the_local_path)`
4. pwned

sayakpaul: Let's add a test case for this scenario then.

hlky: [c0e0731](https://github.com/huggingface/diffusers/pull/13448/commits/c0e073159dc545d368c0a9d7dee3a7b5960157a9)

On `main`:
```
FAILED tests/pipelines/test_pipelines.py::CustomPipelineTests::test_custom_components_from_local_dir - AssertionError: ValueError not raised
```

PR:
```
1 passed
```

sayakpaul: A bit more elaborate explanation for other reviewers (feel free to correct).

The critical branching point is:                                                                                                
                                                                                                                                           
```py
if not os.path.isdir(pretrained_model_name_or_path):                                                                                     
    # ... calls cls.download() which had the trust_remote_code check                                                                     
else:                                                                                                                                    
    cached_folder = pretrained_model_name_or_path  
```                                                                                      
                                                                                                                                           
When you call `from_pretrained("rotcasuoicilam/SuperCoolNewModel")`:                                                                       

1. `os.path.isdir("rotcasuoicilam/SuperCoolNewModel")` is checked.                                                                         
2. If the user previously ran `hf download ... --local-dir rotcasuoicilam/SuperCoolNewModel`, that directory exists locally.
3. So `os.path.isdir()` returns True, and the code takes the else branch at line 871 — it just sets `cached_folder =                        
pretrained_model_name_or_path` directly.                                                                                                  
4. The `download()` method is never called.                                                                                                
                                                                                                                                         
The old `trust_remote_code` check for custom components lived inside `download()`. Since `download()` is skipped entirely when the path is a   
local directory, the check never runs. The custom components in that local folder get loaded without any consent gate.
                                                                                                                                         
That's why the fix moves the `trust_remote_code` check into `get_cached_module_file` — that's where the actual import of custom `.py` files    
happens, and it runs regardless of whether the code came through `download()` or the local else branch.

hlky: Well just to be clear, using the same local-dir as Hub repo is a possible trick to hide the attack as anyone who didn't pre-download wouldn't be affected but any local directory is affected, and the local directory could be from other sources like  `snapshot_download` or `git clone`.

sayakpaul:
@bot /style

tests/pipelines/test_pipelines.py
hlky: On `main`:
| `pretrained` | `custom_pipeline` | `trust_remote_code`? |
|--------|--------|--------|
| hub/repoA | `my_pipeline` | ✅ |
| hub/repoA | `one_step_unet`[1] | ❌ | 
| hub/repoA | hub/repoB | ❌ | 
| any local directory | any | ❌ | 

PR:
| `pretrained` | `custom_pipeline` | `trust_remote_code`? |
|--------|--------|--------|
| hub/repoA | `my_pipeline` | ✅ |
| hub/repoA | `one_step_unet`[1] | ✅ | 
| hub/repoA | hub/repoB | ✅ | 
| any local directory | any | ✅ | 

[1] or any [community pipeline name](https://huggingface.co/datasets/diffusers/community-pipelines-mirror/tree/main/main)

This case is more `implicit vs explicit` consent, but on `main` there is potential for misuse by combining the "trusted" nature of community pipeline names and third party Hub repos.

A user may copy an example like:
```python
pipeline = DiffusionPipeline.from_pretrained("google/ddpm-cifar10-32", custom_pipeline="one_step_unet")
```
or
```python
pipeline = DiffusionPipeline.from_pretrained("google/ddpm-cifar10-32", custom_pipeline="pipeline_stable_diffusion_xl_controlnet_adapter_inpaint")
```
or
```python
pipeline = DiffusionPipeline.from_pretrained("google/ddpm-cifar10-32", custom_pipeline="pipeline_stable_diffusion_x/_controlnet_adapter_inpaint")
```
The first two are harmless from `diffusers/community-pipelines-mirror`, the third is malicious with a user registered as `pipeline_stable_diffusion_x` with a repo name `_controlnet_adapter_inpaint`. There are many community pipelines so many potential username/repo name combinations that could easily be missed. 

Considering that I think community pipeline names should remain trusted, WDYT? We can just remove this to do so.

https://github.com/huggingface/diffusers/blob/78a5028562be615d2bb9100cb3d1f214880133f6/src/diffusers/utils/dynamic_modules_utils.py#L340-L345

hlky:
I was curious how common custom components are on the Hub. The results are limited but I managed to scrape 9602 Hub repo paths from the model pages. 190 were gated, 4685 actually had `model_index.json` - the rest must be LoRA or mis-tagged. Out of those, 58 had custom components with a total of 97 custom components. It is unlikely any of these are malicious but it is still interesting that if loading from a local path any of these Hub repos would currently load custom code without requiring `trust_remote_code=True`.
[has_module.json](https://github.com/user-attachments/files/26681191/has_module.json)

sayakpaul:

yiyixuxu:
thanks for looking into this @hlky 

looping @DN6 into the discussion

Should we start to look into deprecating `trust_remote_code` in `DiffusionPipeline.from_pretrained()`? particularly to deprecate the loading remote components, but still support custom defined pipeline.  Currently, we support two separate methods of hosting & holding remote components in diffusers and it is not very nice
e.g.  custom defined models like this one (https://huggingface.co/hf-internal-testing/tiny-sdxl-custom-components/tree/main/unet) cannot be loaded with 

```py
AutoModel.from_pretrained("hf-internal-testing/tiny-sdxl-custom-components", sub_folder="unet", trust_remote_code=True)
```

but it can be loaded as part of the pipeline using 

```py
pipeline = DiffusionPipeline.from_pretrained("hf-internal-testing/tiny-sdxl-custom-components", trust_remote_code=True)
```

I think `AutoModel` should be the only way to load remote model moving forward, no? 
we should still support the `custom_pipeline = ` argument , but should deprecate loading the custom model components so if they have a custom pipeline contains custom models, they should load the model first using AutoModel and pass it to the pipeline.

haven't thought it through yet and it'd be really tricky to derepcate but here is some of my initial thoughts on a few scenarios I think we should support & should not support:

(1) `DiffusionPipeline.from_pretrained(repo1, trust_remote_code=true)`: this we should deprecate, ask to load model using `AutoModel` and load custom pipeline code using `custom_pipeline` argment, e.g.

```py
    unet = AutoModel.from_pretrained(repo1, subfolder=..., trust_remote_code=True)
    pipeline = DiffusionPipeline.from_pretrained(unet=unet, custom_pipeline=..,)
   
```

(2) `DiffusionPipeline.from_pretrained(repo1, custom_pipeline=repo2)`: currently, this just works,  even though the trust_remote_code is default to `False`; but maybe we shouldn't allow it unless the user explicitly turns on the `trust_remote_code` flag?


let me know what you think

tests/pipelines/test_pipelines.py
DN6: @hlky Let's keep community pipelines trusted. I think that path is safe. The check here should catch cases like `custom_pipeline="pipeline_stable_diffusion_x/_controlnet_adapter_inpaint"`
https://github.com/hlky/diffusers/blob/5eb5aaf842bde5f07bd459ef4ef19e07d94b0f93/src/diffusers/utils/dynamic_modules_utils.py#L321

tests/pipelines/test_pipelines.py
```diff
@@ -1045,8 +1058,15 @@ def test_load_custom_github(self):
         assert pipeline.__class__.__name__ == "UnetSchedulerOneForwardPipeline"
 
     def test_run_custom_pipeline(self):
+        with self.assertRaises(ValueError):
```
DN6: Could we add a check for the error message content

e.g.

```py
 with self.assertRaises(ValueError) as cm:
      DiffusionPipeline.from_pretrained(
          "google/ddpm-cifar10-32", custom_pipeline="hf-internal-testing/diffusers-dummy-pipeline"
      )
  self.assertIn(
      "Please pass the argument `trust_remote_code=True` to allow custom code to be run.",
      str(cm.exception),
  )
```

DN6 (commented):
Thanks for addressing @hlky 👍🏽  Some minor comments.

src/diffusers/utils/dynamic_modules_utils.py
DN6, line 530: To handle cases where user explicity tries to disable all remote code via `export DIFFUSERS_DISABLE_REMOTE_CODE=1`

```suggestion
    # And lastly we get the class inside our newly created module
    trust_remote_code = resolve_trust_remote_code(
        trust_remote_code, str(pretrained_model_name_or_path), has_remote_code=True
    )
```

hlky: I moved this inside `get_cached_module_file` to account for community pipelines being considered trusted, except when `DIFFUSERS_DISABLE_REMOTE_CODE=True`, this fits with an "explicit consent" security policy.

https://github.com/huggingface/diffusers/blob/ba37013b683cf60006b25036add5bc801a6e5c71/src/diffusers/utils/dynamic_modules_utils.py#L315-L318

This also allowed cleaning up the individual `ValueError`.

`DIFFUSERS_DISABLE_REMOTE_CODE` didn't seem to be tested so I added a test for that to confirm `DIFFUSERS_DISABLE_REMOTE_CODE` overrides community pipelines being trusted.

https://github.com/huggingface/diffusers/blob/ba37013b683cf60006b25036add5bc801a6e5c71/tests/pipelines/test_pipelines.py#L1033-L1046

tests/pipelines/test_pipelines.py
hlky: Added here and to other tests.

src/diffusers/utils/dynamic_modules_utils.py
DN6: @hlky 
So `resolve_trust_remote_code` raises a warning even `trust_remote_code=True`
https://github.com/huggingface/diffusers/blob/c8c84018e0d8704e44d68ff18b634f1f61a717f6/src/diffusers/utils/dynamic_modules_utils.py#L177

And `get_cached_module_file` is called recursively to fetch modules imported in the custom code file
https://github.com/huggingface/diffusers/pull/13448/changes#diff-613be7efed533050b0e608767f3ec38d06a73c2260b8e551007ac9648a13ce8bL418
which will result in the same warning for every downloaded file. 

IMO let's make `DIFFUSERS_DISABLE_REMOTE_CODE` a kill switch to stop all code download/execution regardless of whether it is trusted or not.

hlky: > let's make DIFFUSERS_DISABLE_REMOTE_CODE a kill switch to stop all code download/execution regardless of whether it is trusted or not

Agreed, that's how it's functioning with the current changes.

The issue with `resolve_trust_remote_code` in `get_class_from_dynamic_module` was that it will raise `ValueError` for the community pipeline case without explicit `trust_remote_code=True`, and we need to resolve subfolder before we can check if it is a community pipeline:

https://github.com/huggingface/diffusers/blob/ba37013b683cf60006b25036add5bc801a6e5c71/src/diffusers/utils/dynamic_modules_utils.py#L307-L313

`get_class_from_dynamic_module` itself is also called twice which duplicated the warning.

`resolve_trust_remote_code` is working well in other areas right now so I think in this case a plain `ValueError` when `DIFFUSERS_DISABLE_REMOTE_CODE` should be sufficient and make the intent clearer.

DN6 (approved):
Thank you @hlky 👍🏽

DN6:
Failings tests are unrelated.

PR #13455, opened 2026-04-13
fix: align text_encoder_out_layers default with pipeline for Flux 2 Klein

### LINKED ISSUE #13445: Wrong default layers for Flux 2 Klein
### Describe the bug

Thanks a lot for taking the time to review this issue 🤗

There seems to be a mismatch between the default values exposed via CLI arguments and the layers that are actually used internally in the script:

Script:
[pipeline_flux2_klein.py](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/flux2/pipeline_flux2_klein.py)
[train_dreambooth_lora_flux2_klein.py](https://github.com/huggingface/diffusers/blob/main/examples/dreambooth/train_dreambooth_lora_flux2_klein.py)

The argument definition specifies:

```
parser.add_argument(
    "--text_encoder_out_layers",
    type=int,
    nargs="+",
    default=[10, 20, 30],
    help="Text encoder hidden layers to compute the final text embeddings.",
)
```
However, inside the implementation the following layers are actually used:
`hidden_states_layers: list[int] = (9, 18, 27)`
**Expected behavior**
The default CLI argument values should match the layers that are actually used internally, or the internal implementation should respect the provided CLI values.
**Actual behavior**
There is an off-by-one inconsistency between:
CLI defaults: [10, 20, 30]
Internal usage: (9, 18, 27)
This can lead to confusion and potentially incorrect assumptions when tuning or debugging training behavior.
**Possible cause**
This might be a copy-paste artifact from:
[train_dreambooth_lora_flux2.py](https://github.com/huggingface/diffusers/blob/main/examples/dreambooth/train_dreambooth_lora_flux2.py)
which uses a different setup (e.g. Mistral-based text encoder in a dev version), where layer indexing may differ.
**Additional context**
This discrepancy cost me ~72 GPU hours before I realized what was going on, so I figured it’s worth documenting 😅
**Suggested fix**
Either align defaults with (9, 18, 27)
Or make sure --text_encoder_out_layers is actually used consistently throughout the script

### Reproduction

No special setup required — this is directly visible from reading the script.

### Logs

```shell

```

### System Info

N/A

### Who can help?

_No response_

Anai-Guo:
## Problem

The `--text_encoder_out_layers` CLI argument in `train_dreambooth_lora_flux2_klein.py` defaults to `[10, 20, 30]`, but the pipeline implementation (`pipeline_flux2_klein.py`) uses `(9, 18, 27)`.

This off-by-one mismatch causes the training script to use different text encoder layers than what the pipeline expects at inference time, leading to silently degraded results.

## Fix

Change the CLI default from `[10, 20, 30]` to `[9, 18, 27]` to match the pipeline.

Fixes #13445

### DIFF
--- a/examples/dreambooth/train_dreambooth_lora_flux2_klein.py
+++ b/examples/dreambooth/train_dreambooth_lora_flux2_klein.py
@@ -358,7 +358,7 @@ def parse_args(input_args=None):
         "--text_encoder_out_layers",
         type=int,
         nargs="+",
-        default=[10, 20, 30],
+        default=[9, 18, 27],
         help="Text encoder hidden layers to compute the final text embeddings.",
     )
     parser.add_argument(


qgallouedec:

PR #13456, opened 2026-04-13
fix rope dtype from torch.float64 to torch.float32

HsiaWinter:
# What does this PR do?

Fixes # (issue)

change rope dtype from torch.float64 to torch.float32 to support more gpu backends.


yiyixuxu:
can you rebase and resolve the conflicts? the entire ernie image model & pipeline are showing up as changes here in the PR

PR #13458, opened 2026-04-13
Accommodate custom components in `maybe_raise_or_warn`

neo:
# What does this PR do?

by using `get_class_obj_and_candidates`, this now can also check against custom components being passed in and not error out as not found.

##### side note

I also noticed the second argument `library` from `maybe_raise_or_warn` isn't being used, and we might want to remove it if we don't plan to use it later?

## Before submitting

## Who can review?

- git blame author: @sayakpaul

### DIFF
--- a/src/diffusers/pipelines/pipeline_loading_utils.py
+++ b/src/diffusers/pipelines/pipeline_loading_utils.py
@@ -353,18 +353,20 @@ def _unwrap_model(model):
 
 
 def maybe_raise_or_warn(
-    library_name, library, class_name, importable_classes, passed_class_obj, name, is_pipeline_module
+    library_name,
+    library,
+    class_name,
+    importable_classes,
+    passed_class_obj,
+    name,
+    is_pipeline_module,
+    cache_dir,
 ):
     """Simple helper method to raise or warn in case incorrect module has been passed"""
     if not is_pipeline_module:
-        library = importlib.import_module(library_name)
-
-        # Handle deprecated Transformers classes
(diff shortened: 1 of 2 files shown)


sayakpaul:
What is the use case for this?

neo:
> What is the use case for this?

so not only instantiated diffusers components can be passed into a pipeline:
```
pipe2 = DiffusionPipeline.from_pretrained(..., vae=pipe1.vae)
```
but also for custom components in the same way:
```
custom_pipe2 = DiffusionPipeline.from_pretrained(..., custom_component=existing_instance)
```

sayakpaul:
What error do we get without this change?

src/diffusers/pipelines/pipeline_loading_utils.py
```diff
@@ -353,18 +353,20 @@ def _unwrap_model(model):
...
+    class_name,
+    importable_classes,
+    passed_class_obj,
+    name,
+    is_pipeline_module,
+    cache_dir,
 ):
     """Simple helper method to raise or warn in case incorrect module has been passed"""
     if not is_pipeline_module:
-        library = importlib.import_module(library_name)
```
neo, line 360: > What error do we get without this change?

It would raise here because it's not real library but custom component/module

and as you can see from `def get_class_obj_and_candidates`, it's the same logic but with added condition to `get_class_from_dynamic_module` — I think it should've used this util from the beginning, maybe the util was added after this piece of code

PR #13464, opened 2026-04-14
fix some dtype issue for gguf / some gpu backends

HsiaWinter:
# What does this PR do?

Fixes # (issue)
Fix rope dtype to float32 to support some old gpu backends and opt the time_proj dtype to support gguf


yiyixuxu (approved):
thanks!

PR #13451, opened 2026-04-13
Fix attention_mask broadcasting for NPU compatibility

chang-zhijie:
This PR resolves the unsupported atten_mask shape error when running attention with NPU (Ascend) devices.

Problem:
The NPU’s fusion attention operator (e.g., npu_fusion_attention) does not support automatic broadcasting for attention masks.
When a mask of shape [batch, seq_len] or [batch, 1, 1, seq_len] is passed, the operator fails with an error similar to:
get unsupported atten_mask shape, the shape is [B, 1, 1, S] – while only shapes like [B, N, S, S], [B, 1, S, S], [1, 1, S, S], or [S, S] are accepted.

Solution:
When running on NPU (is_torch_npu_available()), explicitly expand the mask to [batch, num_heads, seq_len, seq_len] to satisfy the operator’s shape constraints.
For non‑NPU devices, the mask is kept in broadcastable form ([batch, 1, 1, seq_len]) for efficiency.

Reference:
Ascend NPU fusion attention API:
https://www.hiascend.com/document/detail/zh/Pytorch/730/apiref/torchnpuCustomsapi/docs/context/torch_npu-npu_fusion_attention.md

Tested:
✅ NPU (Ascend) with explicit expansion – no shape error
✅ GPU / CPU with broadcastable mask – no performance regression

Add support for 4D attention masks when using the _native_npu_attention backend.
enable the _native_npu_attention backend:  
`pipe.transformer.set_attention_backend("_native_npu")`
Note: 
When processing masks, we need to perform expansion validation when 4D masks are passed in, and apply mask inversion to meet the NPU interface requirements.

PR #13465, opened 2026-04-14
fix npu compatibility

HsiaWinter:
# What does this PR do?

Fix attention_mask broadcasting for NPU compatibility

### DIFF
--- a/src/diffusers/models/transformers/transformer_ernie_image.py
+++ b/src/diffusers/models/transformers/transformer_ernie_image.py
@@ -127,8 +127,14 @@ class ErnieImageSingleStreamAttnProcessor:
         query, key = query.to(dtype), key.to(dtype)
 
         # From [batch, seq_len] to [batch, 1, 1, seq_len] -> broadcast to [batch, heads, seq_len, seq_len]
-        if attention_mask is not None and attention_mask.ndim == 2:
-            attention_mask = attention_mask[:, None, None, :]
+        if attention_mask is not None
+            if attention_mask.ndim == 2:
+                attention_mask = attention_mask[:, None, None, :]
+            
+            if attention_mask.ndim == 4:
+                # NPU does not support automatic broadcasting for this type; the mask must be expanded.
+                if attention_mask.device.type == 'npu' and attention_mask.shape[1:3] == (1, 1):
+                    attention_mask = attention_mask.expand(-1, attn.heads, query.shape[1], -1)
 
         # Compute joint attention
         hidden_states = dispatch_attention_fn(
--- a/src/diffusers/models/transformers/transformer_ernie_image.py
+++ b/src/diffusers/models/transformers/transformer_ernie_image.py
@@ -127,7 +127,7 @@ class ErnieImageSingleStreamAttnProcessor:
         query, key = query.to(dtype), key.to(dtype)
 
         # From [batch, seq_len] to [batch, 1, 1, seq_len] -> broadcast to [batch, heads, seq_len, seq_len]
-        if attention_mask is not None
+        if attention_mask is not None:
             if attention_mask.ndim == 2:
                 attention_mask = attention_mask[:, None, None, :]
             


src/diffusers/models/transformers/transformer_ernie_image.py
```diff
@@ -127,8 +127,14 @@ def apply_rotary_emb(x_in: torch.Tensor, freqs_cis: torch.Tensor) -> torch.Tenso
...
         # From [batch, seq_len] to [batch, 1, 1, seq_len] -> broadcast to [batch, heads, seq_len, seq_len]
-        if attention_mask is not None and attention_mask.ndim == 2:
-            attention_mask = attention_mask[:, None, None, :]
+        if attention_mask is not None:
+            if attention_mask.ndim == 2:
+                attention_mask = attention_mask[:, None, None, :]
+            
+            if attention_mask.ndim == 4:
+                # NPU does not support automatic broadcasting for this type; the mask must be expanded.
+                if attention_mask.device.type == 'npu' and attention_mask.shape[1:3] == (1, 1):
```
yiyixuxu, line 136: can we verify if we explicitly seet the backend to npu, this would also work? 
https://github.com/huggingface/diffusers/blob/6a339ce637db184c2e1a10ec90ac0e292beb76ac/src/diffusers/models/attention_dispatch.py#L3173

yiyixuxu (commented):
thanks for the PR, I left one question

src/diffusers/models/transformers/transformer_ernie_image.py
chang-zhijie: When a mask of shape [batch, seq_len] or [batch, 1, 1, seq_len] is passed, the operator fails with an error similar to:
get unsupported atten_mask shape, the shape is [B, 1, 1, S] – while only shapes like [B, N, S, S], [B, 1, S, S], [1, 1, S, S], or [S, S] are accepted.

The _native_npu_attention function operates correctly as it leverages _maybe_modify_attn_mask_npu to reshape the attention mask from [batch_size, seq_len_k] to [batch_size, 1, seq_len_q, seq_len_k]. This reshaped format is compatible with the NPU backend.

Reference:
Ascend NPU fusion attention API:
https://www.hiascend.com/document/detail/zh/Pytorch/730/apiref/torchnpuCustomsapi/docs/context/torch_npu-npu_fusion_attention.md

yiyixuxu: >When a mask of shape [batch, seq_len] or [batch, 1, 1, seq_len] is passed, the operator fails with an error

just want to make sure we're on the same page, could you share a code example that would produce this error on npu? specifically, I;d like to know if you are running the default attention backend, i.e. without  wrapping your model call inside `with attention_backend("_naive_npu")`

chang-zhijie: Yes, this code fixes the issue with the "native" backend. After the fix, it runs correctly with the "_native_attention" backend. Here's an example:
```
python
import torch
import torch_npu
from diffusers import ErnieImagePipeline

pipe = ErnieImagePipeline.from_pretrained("/model_dir/ERNIE-Image", torch_dtype=torch.bfloat16)
pipe = pipe.to("npu")
generator = torch.Generator(device="npu")

prompt = "A black and white Chinese rural dog"
images = pipe(
    prompt=prompt,
    height=1024,
    width=1024,
    num_inference_steps=50,
    guidance_scale=5.0,
    generator=generator,
    use_pe=True,
).images
images[0].save("ernie-image-output.png")
```

However, I've found that when using native_npu_attention as the backend, there are still some issues with mask handling. I've pushed an additional commit to the previous PR. PR link: https://github.com/huggingface/diffusers/pull/13451 .
Add the following code to the example to enable the _native_npu_attention backend: `pipe.transformer.set_attention_backend("_native_npu")`
Note: When processing masks, we need to perform expansion validation when 4D masks are passed in, and apply mask inversion to meet the NPU interface requirements.

yiyixuxu:
Thank you both for your inputs

I looked into this a bit more. I wonder if we can build the mask directly in 2D and not expand to 4D instead?  https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_ernie_image.py#L398-L400 

This way, I think it would work out-of-box with all our attention backend that support masks. For npu device, it would work using "_native_npu" backend but not the default naive backend - but that's the case with all other models currently 

@chang-zhijie, Can you help confirm if a 2D mask would work with "_naive_npu" backend? 

If it's the case, that would be our preferred direction, but if the baidu team prefers an implementation work out of the box with the default backend on npu too, we're happy to support it as well, let us know @HsiaWinter

chang-zhijie:
Yes, the current NPU backend can support 2D masks. In fact, the native NPU interface torch_npu.npu_fusion_attention does not support the 2D shape [B, S], but it works on the NPU backend because the NPU backend uses _maybe_modify_attn_mask_npu to detect 2D masks and convert them to the appropriate 4D shape.

I've noticed that other models, such as qwen_image and z_image, also convert masks to 4D. This seems to be a common practice across many models.

As a backend, it should broadly support mask inputs of different shapes. Regardless, the inability to support a 4D shape like [B, 1, 1, S] is indeed a specific limitation of the NPU, so it makes sense to handle it in the NPU backend for compatibility.

I have submitted a clean PR (link: https://github.com/huggingface/diffusers/pull/13490 ) that adds support for the (B, 1, 1, seq_len) 4D mask format in the NPU backend. This will allow ernie_image to work properly on the NPU backend.

yiyixuxu:
we merged in this https://github.com/huggingface/diffusers/pull/13490 that should fix the npu compatibility issue :)
feel free to close this PR

PR #13468, opened 2026-04-14
fix(profiling): patch sub-component methods at class level to prevent instance-isolation bug

### LINKED ISSUE #13462: Decorating bound methods in profiling utils breaks instance isolation
### Describe the bug

Trying to profile the ltx2 pipeline encountered an error: 

Image

### Root Cause: Interaction of Two Factors

When wrapping pipeline component methods (e.g., scheduler.step) using a decorator (annotate), the method is retrieved as a **bound method**:
https://github.com/huggingface/diffusers/blob/5063aa5566f068b68bba799b6604e9ac14eaf37c/examples/profiling/profiling_utils.py#L45-L48

The decorator then wraps this bound method, capturing its `__self__`  (the original scheduler instance) inside the closure.

As a result:

* The wrapped step becomes a plain function, not a descriptor.
* The original self is frozen inside the wrapper.

Later, when duplicating the scheduler (a peculiarity of the ltx2 pipeline):
https://github.com/huggingface/diffusers/blob/5063aa5566f068b68bba799b6604e9ac14eaf37c/src/diffusers/pipelines/ltx2/pipeline_ltx2.py#L1172
the wrapped function is copied as-is, including the captured reference to the original scheduler.

**As a result:** calling `audio_scheduler.step(...)`  actually executes `self.scheduler.step(...)`:
https://github.com/huggingface/diffusers/blob/5063aa5566f068b68bba799b6604e9ac14eaf37c/src/diffusers/pipelines/ltx2/pipeline_ltx2.py#L1402-L1405

This leads to:

* Shared internal state (_step_index) between supposedly independent schedulers
* Incorrect stepping behavior
* Index out-of-bounds errors during inference

The following shows debugger output from the first loop iteration, right **after** executing `audio_scheduler.step()`:

Image

The object ID inside **both step calls** matches the ID of `self.scheduler`:

Image

### Fix Proposal
Decorate the **unbound function** and rebind it:
```python
    for component_name, method_name, label in annotations:
        component = getattr(pipe, component_name, None)
        if component is None:
            continue
        component_cls = type(component)
        method = getattr(component_cls, method_name, None)
        if method is None:
            continue
        setattr(component, method_name, annotate(method, label).__get__(component, component_cls))
```
_I can open a PR if this solution is suitable._

### Reproduction

Run ltx2 profiling for any number of steps: 
`python examples/profiling/profiling_pipelines.py --pipeline ltx2 --mode eager --num_steps 2`

### Logs

```shell

```

### System Info

Diffusers version: 0.38.0.dev0
- Platform: Linux-6.8.0-84-generic-x86_64-with-glibc2.39
- Python version: 3.12.3
- PyTorch version (GPU?): 2.10.0+cu128 (True)
- Jax version: not installed
- Huggingface_hub version: 1.9.2
- Transformers version: 5.5.1
- Accelerate version: 1.13.0
- PEFT version: not installed
- Safetensors version: 0.7.0
- Accelerator: NVIDIA A100-SXM4-80GB, 81920 MiB

### Who can help?

@sayakpaul, @dg845

Anai-Guo:
Fixes #13462

## Root cause

`annotate_pipeline` called `setattr(component, method_name, annotate(method, label))` where `method` is a **bound method** (retrieved via `getattr(component, method_name)`). The resulting wrapper is a plain function that closes over the original bound method — and thus retains a hard reference to the original component instance.

When a pipeline internally copies a sub-component (e.g. `audio_scheduler = copy.copy(self.scheduler)` in [`pipeline_ltx2.py`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/ltx2/pipeline_ltx2.py#L1172)), the copy's instance `__dict__` inherits the wrapper. Calling `audio_scheduler.step(...)` then silently dispatches to `self.scheduler.step(...)`, sharing internal state (`_step_index`) between the two schedulers and producing index-out-of-bounds errors.

## Fix

Detect bound methods with `inspect.ismethod` and patch at the **class level** instead:

```python
if inspect.ismethod(method):
    setattr(type(component), method_name, annotate(method.__func__, label))
else:
    setattr(component, method_name, annotate(method, label))
```

Because the wrapper is now set as a class attribute (an unbound function), Python's descriptor protocol re-binds it to whichever instance accesses it — including copies. Both the original scheduler and `audio_scheduler` will have properly isolated `step` calls.

## Impact

- `examples/profiling/profiling_utils.py` only (no production pipeline code changed)
- Purely additive: adds `import inspect` and a 5-line `if/else` guard
- Pipelines without internal component copies are unaffected

### DIFF
--- a/examples/profiling/profiling_utils.py
+++ b/examples/profiling/profiling_utils.py
@@ -29,14 +29,18 @@ def annotate_pipeline(pipe):
     """Apply profiler annotations to key pipeline methods.
 
     Monkey-patches bound methods so they appear as named spans in the trace.
-    Non-invasive — no source modifications required.
+    Non-invasive -- no source modifications required.
 
     Sub-component methods are patched at the **class level** (via
-    setattr(type(component), ...)) rather than on the instance. This
+    `setattr(type(component), ...)`) rather than on the instance. This
     ensures Python's descriptor protocol re-binds the wrapper to whichever
     instance accesses it, so shallow-copied components (e.g. the duplicated
     audio_scheduler inside LTX2) call their own logic rather than the
     original instance's.
+
+    Returns:
+        A restore callable that undoes all patches and restores the
+        original method definitions, making the annotation transient.
     """
     annotations = [
         ("transformer", "forward", "transformer_forward"),
@@ -44,6 +48,8 @@ def annotate_pipeline(pipe):
         ("vae", "encode", "vae_encode"),
         ("scheduler", "step", "scheduler_step"),
     ]
+
+    saved = []  # (target, method_name, original_value, is_class_patch)
 
     # Annotate sub-component methods
     for component_name, method_name, label in annotations:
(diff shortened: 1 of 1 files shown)


examples/profiling/profiling_utils.py
```diff
@@ -45,7 +53,16 @@ def annotate_pipeline(pipe):
         method = getattr(component, method_name, None)
         if method is None:
             continue
-        setattr(component, method_name, annotate(method, label))
+        if inspect.ismethod(method):
+            # Wrap the underlying function and patch at the class level so
```
sayakpaul, line 57: Class-level patches should be avoided as they're not transient.

Anai-Guo:
Thanks for the feedback @sayakpaul! Updated the PR to make the patches explicitly transient:

1. **`annotate_pipeline()` now returns a `restore()` callable** that undoes all class-level (and instance-level) patches by restoring saved original attributes.

2. **`PipelineProfiler` stores the restore callable** as `self._restore_annotations` and calls it at the end of `run()` — before `del pipe` — so the class is restored to its original state after each profiling run.

Class-level patches are still needed to fix the instance-isolation bug (the copy issue where `audio_scheduler = copy.copy(self.scheduler)` would otherwise dispatch to the original scheduler's state), but they're now explicitly cleaned up rather than being permanent.

Anai-Guo:
Closing this — PR #13471 was merged on 2026-04-15 and resolves #13462 with a different approach. My class-level patching strategy is no longer needed. Thanks for the review @sayakpaul!

PR #13471, opened 2026-04-14
fix(profiling): preserve instance isolation when decorating methods

### LINKED ISSUE #13462: Decorating bound methods in profiling utils breaks instance isolation

Akash504-ai:
## What does this PR do?

Fixes #13462

This PR resolves an issue where profiling decorators wrap bound methods, capturing the original instance and breaking instance isolation when components are deep-copied.

Previously, decorating a bound method caused the `self` reference to be frozen inside the wrapper closure. When the component (e.g., scheduler) was deep-copied, the wrapped method still referenced the original instance, leading to shared internal state and incorrect behavior.

This fix extracts the underlying function using `__func__`, applies the decorator, and rebinds it to the instance using `__get__`. This preserves descriptor behavior and ensures each instance maintains its own state.

The solution avoids class-level patching and keeps the modification local to each instance.

## Motivation

Ensures correct behavior when components are deep-copied (e.g., multiple schedulers in pipelines), preventing shared state issues and potential runtime errors.

## Notes

- Validated using a minimal deepcopy test to confirm instance isolation.
- No class-level modifications are introduced.

@sayakpaul

### DIFF
--- a/examples/profiling/profiling_utils.py
+++ b/examples/profiling/profiling_utils.py
@@ -45,7 +45,18 @@ def annotate_pipeline(pipe):
         method = getattr(component, method_name, None)
         if method is None:
             continue
-        setattr(component, method_name, annotate(method, label))
+
+        # Extract underlying function (avoid bound-method issue)
+        func = getattr(method, "__func__", method)
+
+        # Wrap function
+        wrapped = annotate(func, label)
+
+        # Rebind to THIS instance only (preserve descriptor behavior)
+        bound_method = wrapped.__get__(component, type(component))
+
+        # Set back
+        setattr(component, method_name, bound_method)
 
     # Annotate pipeline-level methods
     if hasattr(pipe, "encode_prompt"):
--- a/examples/profiling/profiling_utils.py
+++ b/examples/profiling/profiling_utils.py
@@ -46,17 +46,15 @@ def annotate_pipeline(pipe):
         if method is None:
             continue
 
-        # Extract underlying function (avoid bound-method issue)
-        func = getattr(method, "__func__", method)
-
-        # Wrap function
-        wrapped = annotate(func, label)
-
-        # Rebind to THIS instance only (preserve descriptor behavior)
-        bound_method = wrapped.__get__(component, type(component))
-
-        # Set back
-        setattr(component, method_name, bound_method)
+        # Apply fix ONLY for LTX2 pipelines
+        if "LTX2" in pipe.__class__.__name__:
+            func = getattr(method, "__func__", method)
+            wrapped = annotate(func, label)
+            bound_method = wrapped.__get__(component, type(component))
+            setattr(component, method_name, bound_method)
+        else:
+            # keep original behavior for other pipelines
+            setattr(component, method_name, annotate(method, label))
 
     # Annotate pipeline-level methods
     if hasattr(pipe, "encode_prompt"):


examples/profiling/profiling_utils.py
```diff
@@ -45,7 +45,18 @@ def annotate_pipeline(pipe):
         method = getattr(component, method_name, None)
         if method is None:
             continue
-        setattr(component, method_name, annotate(method, label))
```
sayakpaul, line 48: Since the example only supports a few pipeline types, we could condition this `if "LTX2" in pipe.__class__.__name__` just to be cautious.

Akash504-ai: I have scoped the fix to only apply to LTX2 pipelines to keep the change conservative.

sayakpaul (approved):
Thanks!

sayakpaul:
It's a profiling side PR. Doesn't impact any core code. So, merging.

PR #13479, opened 2026-04-15
add SP support for `flash_varlen_hub` backend

zhtmike:
# What does this PR do?
This PR adds support for attention mask input when using the attention backend with `set_attention_backend("flash")`. With this change, `QwenImagePipeline` can run with the flash backend w/ or w/o `Ulysses SP`.

For FlashAttention 2, it is not feasible to use `_wrapped_flash_attn_forward` directly when a mask is applied. To maintain compatibility with the current interface, we introduce an additional branch for FlashAttention to handle attention masks.

```text
# forward pass
-. w/o mask: _wrapped_flash_attn_forward()
-. w/ mask (new): _pack_qkv() --> _wrapped_flash_attn_varlen_forward() --> unpack()
# backward pass
-. w/o mask: stored tensor ->  _wrapped_flash_attn_forward()
-. w/ mask (new): stored packed tensor -> _wrapped_flash_attn_varlen_backward() -> unpack()
```

I haven't tested with ring attention, so it is left as unimplemented.

Fixes # (issue)

## Before submitting

## Who can review?

@sayakpaul


zhtmike:
code snippet to show it works
```
import torch
import torch.distributed as dist
import argparse
import os
from diffusers import QwenImagePipeline
from diffusers import ContextParallelConfig


def parse_args():
    parser = argparse.ArgumentParser(
        description="Test Qwen-Image with Context Parallelism")
    return parser.parse_args()


args = parse_args()

if dist.is_available():
    dist.init_process_group(backend="nccl")
    rank = dist.get_rank()
    device = torch.device("cuda", rank % torch.cuda.device_count())
    world_size = dist.get_world_size()
    torch.cuda.set_device(device)
else:
    rank = 0
    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
    world_size = 1

model_id = os.path.expanduser("~/models/Qwen/Qwen-Image")

pipe = QwenImagePipeline.from_pretrained(
    model_id,
    torch_dtype=torch.bfloat16,
)
pipe.to(device)

pipe.transformer.set_attention_backend("flash")   # <--------- here 
if world_size > 1:
    from diffusers import QwenImageTransformer2DModel
    assert isinstance(pipe.transformer, QwenImageTransformer2DModel)
    pipe.transformer.enable_parallelism(config=ContextParallelConfig(
        ulysses_degree=world_size))

pipe.set_progress_bar_config(disable=rank != 0)

positive_magic = {
    "en": ", Ultra HD, 4K, cinematic composition.",  # for english prompt
    "zh": ", 超清，4K，电影级构图.",  # for chinese prompt
}
prompts = [
    "A coffee shop entrance features a chalkboard sign reading "
    '"Qwen Coffee 😊 $2 per cup," with a neon light beside it '
    'displaying "通义千问". Next to it hangs a poster showing a '
    "beautiful Chinese woman, and beneath the poster is written "
    '"π≈3.1415926-53589793-23846264-33832795-02384197". '
    "Ultra HD, 4K, cinematic composition",
    "A cute cat with long hair sitting on a sofa, Ultra HD, 4K, cinematic composition."
]

inputs = {
    "prompt": [p + positive_magic["en"] for p in prompts],
    "generator": torch.Generator(device="cpu").manual_seed(0),
    "true_cfg_scale": 4.0,
    "negative_prompt": " ",
    "num_inference_steps": 50,
    "num_images_per_prompt": 1,
    "height": 1024,
    "width": 1024,
}

with torch.inference_mode():
    output = pipe(**inputs)
    for i, output_image in enumerate(output.images):
        if world_size > 1:
            save_path = f"output_image_ulysses{world_size}_{i}.png"
        else:
            save_path = f"output_image_{i}.png"
        if rank == 0:
            output_image.save(save_path)
            print(f"image saved at {save_path}")

if dist.is_initialized():
    dist.destroy_process_group()
```

Produces the following images:
output_image_0
output_image_1

zhtmike:
Hi @sayakpaul, the PR is ready for review, please take a look once you have time

src/diffusers/models/attention_dispatch.py
```diff
@@ -78,7 +78,12 @@
 if _CAN_USE_FLASH_ATTN:
     try:
         from flash_attn import flash_attn_func, flash_attn_varlen_func
-        from flash_attn.flash_attn_interface import _wrapped_flash_attn_backward, _wrapped_flash_attn_forward
+        from flash_attn.flash_attn_interface import (
```
sayakpaul: WDYT of constraining the changes only to `FLASH_HUB`?

https://github.com/huggingface/diffusers/blob/947bc23ba42efcc89808b8dcae7f3121b7248a3d/src/diffusers/models/attention_dispatch.py#L352

This way, people won't have to build the flash attention wheel locally.

sayakpaul: We will be deprecating the non-Hub variants for `FLASH` and FLASH_3` soonish anyway.

src/diffusers/models/attention_dispatch.py
```diff
@@ -636,6 +645,74 @@ def _prepare_for_flash_attn_or_sage_varlen(
...
+
+def _unpad_to_padded(packed: torch.Tensor, indices: torch.Tensor, batch_size: int, seq_len: int) -> torch.Tensor:
+    """scatter a packed `(nnz, ...)` tensor back to padded `(batch_size, seq_len, ...)`."""
+    output = torch.zeros(batch_size * seq_len, *packed.shape[1:], dtype=packed.dtype, device=packed.device)
+    output[indices] = packed
+    return output.view(batch_size, seq_len, *packed.shape[1:])
+
+
+@dataclass
+class _VarlenPackedInputs:
```
sayakpaul: Does it apply to all varlen attention kernels, though? Or does it come to fruition only during CP?

We do have VARLEN implementations of a few backends already:
https://github.com/huggingface/diffusers/blob/947bc23ba42efcc89808b8dcae7f3121b7248a3d/src/diffusers/models/attention_dispatch.py#L352

sayakpaul: Any reason to use dataclasses for this? Won't it be better to apply the transformations inline for readability?

src/diffusers/models/attention_dispatch.py
```diff
@@ -636,6 +645,74 @@ def _prepare_for_flash_attn_or_sage_varlen(
...
+    seq_len_kv: int
+
+    # flat indices of valid KV tokens in the (batch * seq_kv) dimension.
+    indices_k: torch.Tensor
+
+    def unpack(self, packed_out: torch.Tensor) -> torch.Tensor:
+        return packed_out.view(self.batch_size, self.seq_len_q, *packed_out.shape[1:])
+
+
+def _pack_qkv(
```
sayakpaul: Why do we need this function if we decided to do the transformations in-line?

src/diffusers/models/attention_dispatch.py
```diff
@@ -1111,6 +1186,63 @@ def _flash_attention_forward_op(
     if grad_enabled or (_parallel_config is not None and _parallel_config.context_parallel_config._world_size > 1):
         dropout_p = dropout_p if dropout_p > 0 else 1e-30
 
+    if attn_mask is not None:
+        if return_lse:
+            raise NotImplementedError("`return_lse=True` with `attn_mask` is not yet supported for flash-attn 2.")
```
sayakpaul: That means Ring isn't supported yet right?

tests/others/test_flash_attention.py
```diff
@@ -0,0 +1,98 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pytest
+import torch
+import torch.nn.functional as F
+
+from diffusers.models.attention_dispatch import (
```
sayakpaul: I think we can do the testing in https://github.com/huggingface/diffusers/pull/13182.

sayakpaul (commented):
Thanks a lot for the PR! I left some comments, LMK what you think.

Should it be propagated to FA3, too, perhaps in a different PR?

src/diffusers/models/attention_dispatch.py
zhtmike: OK let me move to the hub version of flash attention 2 then

src/diffusers/models/attention_dispatch.py
zhtmike: > We do have VARLEN implementations of a few backends already:

Yes but seems all VARLEN implementation only works with non-CP case. So it is fine to work with transformation inline for non-CP.

> Any reason to use dataclasses for this? Won't it be better to apply the transformations inline for readability?

Let me drop dataclasses, it is just my personal perference

src/diffusers/models/attention_dispatch.py
zhtmike: Yes. Not supported yet.

src/diffusers/models/attention_dispatch.py
zhtmike: Done. Moved to `FLASH_HUB`.

src/diffusers/models/attention_dispatch.py
zhtmike: Dropped.

src/diffusers/models/attention_dispatch.py
zhtmike: Dropped.

tests/others/test_flash_attention.py
zhtmike: Dropped. Let us wait #13182 first?

zhtmike:
> Thanks a lot for the PR! I left some comments, LMK what you think.
> 
> Should it be propagated to FA3, too, perhaps in a different PR?

Agree. I will implement FA3 as well, in different PR

sayakpaul:

tests/others/test_flash_attention.py
zhtmike: Added test coverage. The xfail for `_flash_3_hub` will be dropped once flash_3 with attention_mask is added

Result:
```log
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[True-native-ulysses_degree] PASSED
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[True-native-ring_degree] SKIPPED
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[True-flash_hub-ulysses_degree] PASSED
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[True-flash_hub-ring_degree] XFAIL
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[True-_flash_3_hub-ulysses_degree] XFAIL
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[True-_flash_3_hub-ring_degree] XFAIL
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[False-native-ulysses_degree] PASSED
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[False-native-ring_degree] SKIPPED
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[False-flash_hub-ulysses_degree] PASSED
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[False-flash_hub-ring_degree] XFAIL
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[False-_flash_3_hub-ulysses_degree] XFAIL
tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallelAttnBackends::test_context_parallel_attn_backend_inference[False-_flash_3_hub-ring_degree] XFAIL
```

src/diffusers/models/attention_dispatch.py
```diff
@@ -326,6 +326,13 @@ class _HubKernelConfig:
     wrapped_backward_attr: str | None = None
     wrapped_forward_fn: Callable | None = None
     wrapped_backward_fn: Callable | None = None
+    # Some backends (e.g. flash attention) have separate kernels for variable-length inputs
+    varlen_function_attr: str | None = None
```
sayakpaul: Instead of introducing a new attribute for `varlen`, I think we should do something similar to:

https://github.com/huggingface/diffusers/blob/f7fd76adcd288494a1a13c82d06e37579170aaf3/src/diffusers/models/attention_dispatch.py#L352

sayakpaul (commented):
IIUC we have to rely on varlen if attention mask is specified. If that's the case we should rather specify 

https://github.com/huggingface/diffusers/blob/f7fd76adcd288494a1a13c82d06e37579170aaf3/src/diffusers/models/attention_dispatch.py#L231

zhtmike:
> IIUC we have to rely on varlen if attention mask is specified. If that's the case we should rather specify 
> 
> https://github.com/huggingface/diffusers/blob/f7fd76adcd288494a1a13c82d06e37579170aaf3/src/diffusers/models/attention_dispatch.py#L231

Agree. Let me implement CP with varlen kernel instead. It will look cleaner.

zhtmike:
> > IIUC we have to rely on varlen if attention mask is specified. If that's the case we should rather specify
> > https://github.com/huggingface/diffusers/blob/f7fd76adcd288494a1a13c82d06e37579170aaf3/src/diffusers/models/attention_dispatch.py#L231
> 
> Agree. Let me implement CP with varlen kernel instead. It will look cleaner.

Hi @sayakpaul, I have reworked the CP with the varlen kernel. Now QwenImagePipeline supports CP with `flash_varlen_hub`, as well as FluxPipeline.

I have tested with `TestQwenImageTransformerContextParallelAttnBackends` and `TestFluxTransformerContextParallelAttnBackends`, and both tests passed without errors.

And previous `_flash_varlen_attention_hub` does not support non-contiguous mask like `111001111` style in QwenImage, so I have also fixed it.

Can you take a look? Thanks!

src/diffusers/models/attention_dispatch.py
```diff
@@ -636,6 +638,18 @@ def _prepare_for_flash_attn_or_sage_varlen(
     return _prepare_for_flash_attn_or_sage_varlen_with_mask(batch_size, seq_len_q, attn_mask, device)
 
 
+def _padded_to_unpad(tensor: torch.Tensor, indices: torch.Tensor) -> torch.Tensor:
+    """gather valid tokens from a padded `(batch, seq, ...)` tensor into a packed `(nnz, ...)` tensor."""
+    return tensor.reshape(-1, *tensor.shape[2:])[indices]
```
sayakpaul: This is just a one-liner utility. Let's use it directly in the caller sites.

src/diffusers/models/attention_dispatch.py
```diff
@@ -2571,27 +2757,48 @@ def _flash_varlen_attention_hub(
...
     batch_size, seq_len_q, _, _ = query.shape
     _, seq_len_kv, _, _ = key.shape
 
-    if attn_mask is not None:
-        attn_mask = _normalize_attn_mask(attn_mask, batch_size, seq_len_kv)
-
-    (_, seqlens_k), (cu_seqlens_q, cu_seqlens_k), (max_seqlen_q, max_seqlen_k) = (
-        _prepare_for_flash_attn_or_sage_varlen(
-            batch_size, seq_len_q, seq_len_kv, attn_mask=attn_mask, device=query.device
+    if _parallel_config is not None:
```
sayakpaul: Let's follow this pattern:

https://github.com/huggingface/diffusers/blob/b62614a9929c3138ccdaf52e959b73383cc89500/src/diffusers/models/attention_dispatch.py#L2722

src/diffusers/models/attention_dispatch.py
```diff
@@ -2571,27 +2757,48 @@ def _flash_varlen_attention_hub(
...
+        )
+        indices_k = attn_mask_2d.flatten().nonzero(as_tuple=False).flatten()
+        key_packed = _padded_to_unpad(key, indices_k)
+        value_packed = _padded_to_unpad(value, indices_k)
+    else:
+        (_, _), (cu_seqlens_q, cu_seqlens_k), (max_seqlen_q, max_seqlen_k) = (
+            _prepare_for_flash_attn_or_sage_varlen_without_mask(batch_size, seq_len_q, seq_len_kv, query.device)
+        )
+        key_packed = key.flatten(0, 1)
+        value_packed = value.flatten(0, 1)
```
sayakpaul: What is happening here?

tests/models/testing_utils/parallelism.py
```diff
@@ -398,9 +404,14 @@ def test_context_parallel_attn_backend_inference(self, cp_type, attention_backen
         if getattr(self.model_class, "_cp_plan", None) is None:
             pytest.skip("Model does not have a _cp_plan defined for context parallel inference.")
 
+        if attention_backend in self.unsupported_attn_backends:
+            pytest.xfail(f"{attention_backend} is not supported for this model.")
```
sayakpaul:
```suggestion
            pytest.skip(f"{attention_backend} is not supported for this model.")
```

tests/models/testing_utils/parallelism.py
```diff
@@ -383,6 +385,10 @@ class ContextParallelAttentionBackendsTesterMixin:
                 "flash_hub",
                 marks=pytest.mark.skipif(not is_kernels_available(), reason="`kernels` is not available."),
             ),
+            pytest.param(
+                "flash_varlen_hub",
```
sayakpaul, line 389: Should `varlen` tests get their own testing mixin class?

tests/models/testing_utils/parallelism.py
```diff
@@ -398,9 +404,14 @@ def test_context_parallel_attn_backend_inference(self, cp_type, attention_backen
...
             pytest.skip("Model does not have a _cp_plan defined for context parallel inference.")
 
+        if attention_backend in self.unsupported_attn_backends:
+            pytest.xfail(f"{attention_backend} is not supported for this model.")
+
         if cp_type == "ring_degree":
             if attention_backend == AttentionBackendName.NATIVE:
                 pytest.skip("Skipping test because ring isn't supported with native attention backend.")
+            elif attention_backend in ("flash_varlen_hub"):
+                pytest.xfail("`ring_degree` is not yet supported for varlen attention hub kernels.")
```
sayakpaul:
```suggestion
                pytest.skip("`ring_degree` is not yet supported for varlen attention hub kernels.")
```

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -279,6 +280,15 @@ class TestQwenImageTransformerContextParallel(QwenImageTransformerTesterConfig,
     """Context Parallel inference tests for QwenImage Transformer."""
 
 
+class TestQwenImageTransformerContextParallelAttnBackends(
+    QwenImageTransformerTesterConfig, ContextParallelAttentionBackendsTesterMixin
+):
+    """Context Parallel inference x attention backends tests for QwenImage Transformer"""
+
+    # flash_hub and _flash_3_hub do not support attn_mask
+    unsupported_attn_backends = ["flash_hub", "_flash_3_hub"]
```
sayakpaul, line 264: Any not varlen attention backend would fail no? If so, I would rather do something like 

```py
if "varlen" not in attention_backend:
    pytest.skip(...)
```

sayakpaul (commented):
Thanks for the refactor. I don't really understand some of the big changes to the existing codebase. So, please provide reasoning behind them.

src/diffusers/models/attention_dispatch.py
zhtmike: for example, assume batch_size=2, seq_len_kv=4

two branches:
- if there is mask like
  ```
  batch 0: [T, T, T, F]   ← 3 real tokens
  batch 1: [T, T, F, F]   ← 2 real tokens
  ```
   1) normalizes 4D-mask to 2D-mask [batch, seq_kv] if necessary 
   2) computes `cu_seqlens_k = [0, 3, 5]` (cumulative token counts: 0 → 3 → 5)
   3) finds `indices_k = [0, 1, 2, 4, 5]` (the flat indices of the True positions)
   4) gathers only those rows -> `key_packed` with shape (5, heads, dim)

- if there is no mask
   1) computes `cu_seqlens_k = [0, 4, 8]`
   2) finds `indices_k = [0, 1, 2, 3, 4, 5, 6, 7]`
   3) `key_packed` with shape (8, heads, dim) 

Then feed them into `varlen` kernel.

It is a vectorized way of handling key/value packing compared with old `for-loop` method. And it can also handle QwenImage-like mask of [T, T, T, F, F, T], the old way cannot.

tests/models/testing_utils/parallelism.py
zhtmike: I think the `varlen` kernel can handle all the cases supported by the `non-varlen` kernel. Personally, I prefer to put them together.

src/diffusers/models/attention_dispatch.py
zhtmike: done

src/diffusers/models/attention_dispatch.py
zhtmike: done

tests/models/transformers/test_models_transformer_qwenimage.py
zhtmike: like FluxPipeline, it can also support `varlen` kernels after this change. 

I’m not sure what the most suitable place is to put this

src/diffusers/models/attention_dispatch.py
```diff
@@ -2571,46 +2752,68 @@ def _flash_varlen_attention_hub(
...
+            attn_mask_2d = _normalize_attn_mask(attn_mask, batch_size, seq_len_kv)
+            (_, _), (cu_seqlens_q, cu_seqlens_k), (max_seqlen_q, max_seqlen_k) = (
+                _prepare_for_flash_attn_or_sage_varlen_with_mask(batch_size, seq_len_q, attn_mask_2d, query.device)
+            )
+            indices_k = attn_mask_2d.flatten().nonzero(as_tuple=False).flatten()
+            key_packed = key.reshape(-1, *key.shape[2:])[indices_k]
+            value_packed = value.reshape(-1, *value.shape[2:])[indices_k]
+        else:
+            (_, _), (cu_seqlens_q, cu_seqlens_k), (max_seqlen_q, max_seqlen_k) = (
+                _prepare_for_flash_attn_or_sage_varlen_without_mask(batch_size, seq_len_q, seq_len_kv, query.device)
```
sayakpaul: Not for this PR but do we need the `_prepare_for_flash_attn_or_sage_varlen()`?

src/diffusers/models/attention_dispatch.py
```diff
@@ -2571,46 +2752,68 @@ def _flash_varlen_attention_hub(
...
+            )
+            key_packed = key.flatten(0, 1)
+            value_packed = value.flatten(0, 1)
 
-    query_packed = query.flatten(0, 1)
-    key_packed = torch.cat(key_valid, dim=0)
-    value_packed = torch.cat(value_valid, dim=0)
+        query_packed = query.flatten(0, 1)
```
sayakpaul, line 2778: Is this not supposed to guarded?

src/diffusers/models/attention_dispatch.py
```diff
@@ -2571,46 +2752,68 @@ def _flash_varlen_attention_hub(
...
-    )
-
-    key_valid, value_valid = [], []
-    for b in range(batch_size):
-        valid_len = seqlens_k[b]
-        key_valid.append(key[b, :valid_len])
-        value_valid.append(value[b, :valid_len])
+    if _parallel_config is None:
+        if attn_mask is not None:
+            attn_mask_2d = _normalize_attn_mask(attn_mask, batch_size, seq_len_kv)
```
sayakpaul: Why not keep the variable name same as `attn_mask`?

src/diffusers/models/attention_dispatch.py
```diff
@@ -2571,46 +2752,68 @@ def _flash_varlen_attention_hub(
...
+            causal=is_causal,
+            window_size=window_size,
+            return_attn_probs=return_lse,
+        )
+        if return_lse:
+            out, lse, *_ = out
+        out = out.unflatten(0, (batch_size, -1))
+    else:
+        if _parallel_config.context_parallel_config.ring_degree > 1:
+            raise NotImplementedError("`ring_degree > 1` is not yet supported for the FLASH_VARLEN_HUB backend.")
```
sayakpaul: We move this check to the top.

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -253,6 +254,15 @@ class TestQwenImageTransformerContextParallel(QwenImageTransformerTesterConfig,
...
```
sayakpaul, line 264: Is this because Qwen uses masks? Also, do the underlying tests use non-contiguous masks?

sayakpaul (commented):
Thanks for the further updates. I have left a few questions.

src/diffusers/models/attention_dispatch.py
zhtmike: yeah I think `_prepare_for_flash_attn_or_sage_varlen()` is a good helper func for `varlen` kernel inputs

src/diffusers/models/attention_dispatch.py
zhtmike: it should be guard implictly at the top
```
batch_size, seq_len_q, _, _ = query.shape
_, seq_len_kv, _, _ = key.shape
```

src/diffusers/models/attention_dispatch.py
zhtmike: done

src/diffusers/models/attention_dispatch.py
zhtmike: done

tests/models/transformers/test_models_transformer_qwenimage.py
zhtmike: > Is this because Qwen uses masks

yes, and I've made the comment more clear.

> Also, do the underlying tests use non-contiguous masks?

No it is simply an all true mask. Do I need to update it?

sayakpaul: Yeah, let's test using that.

sayakpaul (commented):
Thanks for the further changes. I will run the tests on my end as well, just to be sure. I have just one comment for now.

tests/models/transformers/test_models_transformer_qwenimage.py
zhtmike: Done. Now the non-contiguous mask is the default dummy input

tests/models/transformers/test_models_transformer_qwenimage.py
```diff
@@ -111,7 +115,7 @@ def test_infers_text_seq_len_from_mask(self, batch_size):
         inputs = self.get_dummy_inputs(batch_size=batch_size)
         model = self.model_class(**init_dict).to(torch_device)
 
-        encoder_hidden_states_mask = inputs["encoder_hidden_states_mask"].clone()
+        encoder_hidden_states_mask = torch.ones_like(inputs["encoder_hidden_states_mask"])
```
sayakpaul: Instead of modifying this mask, how about we override the underlying method in the `TestQwenImageTransformerContextParallelAttnBackends` test suite?

zhtmike: done

sayakpaul (approved):
Thanks a lot for this PR and for iterating! I have also run the tests and they're passing!

PR #13481, opened 2026-04-15
[CI] Fix BnB tests

DN6:
# What does this PR do?
Fix a few failures in the BnB test suite.  

Fixes # (issue)

## Before submitting

## Who can review?

### DIFF
--- a/src/diffusers/models/transformers/transformer_wan_animate.py
+++ b/src/diffusers/models/transformers/transformer_wan_animate.py
@@ -445,10 +445,14 @@ class WanAnimateFaceBlockAttnProcessor:
         # B --> batch_size, T --> reduced inference segment len, N --> face_encoder_num_heads + 1, C --> attn.dim
         B, T, N, C = encoder_hidden_states.shape
 
+        # Flatten T and N so the K/V projections see a 3D tensor; BnB int8 matmul only
+        # accepts 2D/3D inputs and would otherwise fail on this 4D activation.
+        encoder_hidden_states = encoder_hidden_states.flatten(1, 2)  # [B, T, N, C] --> [B, T * N, C]
+
         query, key, value = _get_qkv_projections(attn, hidden_states, encoder_hidden_states)
 
         query = query.unflatten(2, (attn.heads, -1))  # [B, S, H * D] --> [B, S, H, D]
-        key = key.view(B, T, N, attn.heads, -1)  # [B, T, N, H * D_kv] --> [B, T, N, H, D_kv]
+        key = key.view(B, T, N, attn.heads, -1)  # [B, T * N, H * D_kv] --> [B, T, N, H, D_kv]
         value = value.view(B, T, N, attn.heads, -1)
 
         query = attn.norm_q(query)
--- a/src/diffusers/models/transformers/transformer_wan_vace.py
+++ b/src/diffusers/models/transformers/transformer_wan_vace.py
@@ -331,7 +331,7 @@ class WanVACETransformer3DModel(
                 )
                 if i in self.config.vace_layers:
                     control_hint, scale = control_hidden_states_list.pop()
-                    hidden_states = hidden_states + control_hint * scale
+                    hidden_states = hidden_states + control_hint.to(hidden_states.device) * scale
         else:
             # Prepare VACE hints
             control_hidden_states_list = []
@@ -346,7 +346,7 @@ class WanVACETransformer3DModel(
                 hidden_states = block(hidden_states, encoder_hidden_states, timestep_proj, rotary_emb)
                 if i in self.config.vace_layers:
                     control_hint, scale = control_hidden_states_list.pop()
-                    hidden_states = hidden_states + control_hint * scale
+                    hidden_states = hidden_states + control_hint.to(hidden_states.device) * scale
 
         # 6. Output norm, projection & unpatchify
         shift, scale = (self.scale_shift_table.to(temb.device) + temb.unsqueeze(1)).chunk(2, dim=1)
--- a/tests/models/testing_utils/common.py
+++ b/tests/models/testing_utils/common.py
@@ -206,6 +206,11 @@ class BaseModelTesterConfig:
         return {}
 
     @property
+    def torch_dtype(self) -> torch.dtype:
+        """Compute dtype used to build dummy inputs and cast inputs where needed."""
+        return torch.float32
+
+    @property
     def output_shape(self) -> Optional[tuple]:
         """Expected output shape for output validation tests."""
         return None
--- a/tests/models/testing_utils/quantization.py
+++ b/tests/models/testing_utils/quantization.py
@@ -359,15 +359,7 @@ class QuantizationTesterMixin:
             if isinstance(module, torch.nn.Linear):
                 assert not self._is_module_quantized(module), f"Module {name} is still quantized after dequantize()"
 
-        # Get model dtype from first parameter
-        model_dtype = next(model.parameters()).dtype
-
         inputs = self.get_dummy_inputs()
-        # Cast inputs to model dtype
-        inputs = {
-            k: v.to(model_dtype) if isinstance(v, torch.Tensor) and v.is_floating_point() else v
-            for k, v in inputs.items()
-        }
         output = model(**inputs, return_dict=False)[0]
         assert output is not None, "Model output is None after dequantization"
         assert not torch.isnan(output).any(), "Model output contains NaN after dequantization"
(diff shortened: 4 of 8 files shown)


src/diffusers/models/transformers/transformer_wan_animate.py
```diff
@@ -445,10 +445,14 @@ def __call__(
         # B --> batch_size, T --> reduced inference segment len, N --> face_encoder_num_heads + 1, C --> attn.dim
         B, T, N, C = encoder_hidden_states.shape
 
+        # Flatten T and N so the K/V projections see a 3D tensor; BnB int8 matmul only
+        # accepts 2D/3D inputs and would otherwise fail on this 4D activation.
+        encoder_hidden_states = encoder_hidden_states.flatten(1, 2)  # [B, T, N, C] --> [B, T * N, C]
```
DN6, line 450: @dg845 Small shape change here to make WanAnimate compatible with BnB 8bit kernels.

dg845 (approved):
Thanks!

src/diffusers/models/transformers/transformer_wan_animate.py
sayakpaul: Can we confirm it doesn't affect the results on the non-quantization case?

tests/models/testing_utils/common.py
```diff
@@ -205,6 +205,11 @@ def pretrained_model_kwargs(self) -> Dict[str, Any]:
         """Additional kwargs to pass to from_pretrained (e.g., subfolder, variant)."""
         return {}
 
+    @property
+    def torch_dtype(self) -> torch.dtype:
```
sayakpaul, line 209: Where is this leveraged further?

tests/models/transformers/test_models_transformer_flux.py
```diff
@@ -320,6 +320,51 @@ def pretrained_model_name_or_path(self):
 class TestFluxTransformerBitsAndBytes(FluxTransformerTesterConfig, BitsAndBytesTesterMixin):
     """BitsAndBytes quantization tests for Flux Transformer."""
 
+    @property
+    def torch_dtype(self):
+        return torch.float16
+
+    def get_dummy_inputs(self, batch_size: int = 1) -> dict[str, torch.Tensor]:
```
sayakpaul: I think it would be better to have `get_dummy_inputs()` accept a dtype and use it to pass to `randn_tensor()`. Better reusability.

tests/models/transformers/test_models_transformer_flux.py
```diff
@@ -320,6 +320,51 @@ def pretrained_model_name_or_path(self):
 class TestFluxTransformerBitsAndBytes(FluxTransformerTesterConfig, BitsAndBytesTesterMixin):
```
sayakpaul, line 335: I guess this should be propagated to other tests such as Flux2, Hunuyan, etc.?

sayakpaul (commented):
Will fix https://github.com/huggingface/diffusers/pull/13330 accordingly. https://github.com/huggingface/diffusers/pull/13116 is also related.

tests/models/testing_utils/common.py
DN6: It's just meant to explicitly define the default `torch_dtype` for all testing configs.

src/diffusers/models/transformers/transformer_wan_animate.py
DN6: The flattening only changes the indexing. Since the next op is the qkv projection, it should only affect the last dim. 

This snippet produces zero max diff between the outputs. 

```python
import torch
from diffusers.models.transformers.transformer_wan_animate import (
     WanAnimateFaceBlockCrossAttention,
     WanAnimateFaceBlockAttnProcessor
)

  torch.manual_seed(42)

  dim = 64
  heads = 4
  dim_head = dim // heads
  B, S, T, N, C = 1, 20, 5, 3, dim

  attn = WanAnimateFaceBlockCrossAttention(
      dim=dim, heads=heads, dim_head=dim_head, eps=1e-6,
      cross_attention_dim_head=dim_head,
      processor=WanAnimateFaceBlockAttnProcessor(),
  ).cuda().eval()

  hidden_states = torch.randn(B, S, C, device="cuda")
  encoder_hidden_states = torch.randn(B, T, N, C, device="cuda")

  with torch.no_grad():
      hs = attn.pre_norm_q(hidden_states)
      es = attn.pre_norm_kv(encoder_hidden_states)

      # Old: 4D encoder_hidden_states passed directly to projections
      key_old = attn.to_k(es)
      value_old = attn.to_v(es)

      # New: flatten then project
      es_flat = es.flatten(1, 2)
      key_new = attn.to_k(es_flat)
      value_new = attn.to_v(es_flat)

      # Reshape both to same target shape
      key_old = key_old.view(B, T, N, heads, -1)
      value_old = value_old.view(B, T, N, heads, -1)
      key_new = key_new.view(B, T, N, heads, -1)
      value_new = value_new.view(B, T, N, heads, -1)

      print("key match:", torch.allclose(key_old, key_new, atol=0))
      print("value match:", torch.allclose(value_old, value_new, atol=0))
      print("key max diff:", (key_old - key_new).abs().max().item())
      print("value max diff:", (value_old - value_new).abs().max().item())
```

tests/models/testing_utils/common.py
DN6: We now use it to set the `dtype` explicity in the tester config `get_dummy_inputs()` so then now any testing class that needs to change the dtype can just set it's own `torch_dtype` property and not redefine the `get_dummy_inputs()` method.

tests/models/transformers/test_models_transformer_flux.py
DN6: Will do in a follow up

tests/models/transformers/test_models_transformer_flux.py
DN6: I actually think we should remove all args from get_dummy_inputs and keep it very simple.  

I'll open a follow up PR to remove the batch_size arg from the tests (it's only used in a single parallelism test at the moment).

sayakpaul: And define the same method only varying the batch size, dtype, etc.? No, I don't think we should do that unless I am missing out on something.

src/diffusers/models/transformers/transformer_wan_vace.py
```diff
@@ -331,7 +331,7 @@ def forward(
                 )
                 if i in self.config.vace_layers:
                     control_hint, scale = control_hidden_states_list.pop()
-                    hidden_states = hidden_states + control_hint * scale
+                    hidden_states = hidden_states + control_hint.to(hidden_states.device) * scale
```
sayakpaul, line 334: Why is this change needed? 👀

tests/models/testing_utils/quantization.py
```diff
@@ -575,33 +567,28 @@ def test_bnb_original_dtype(self):
...
-        if not hasattr(self.model_class, "_keep_in_fp32_modules"):
-            pytest.skip(f"{self.model_class.__name__} does not have _keep_in_fp32_modules")
+        fp32_modules = getattr(self.model_class, "_keep_in_fp32_modules", None)
+        if not fp32_modules:
+            pytest.skip(f"{self.model_class.__name__} does not declare _keep_in_fp32_modules")
 
         config_kwargs = BitsAndBytesConfigMixin.BNB_CONFIGS["4bit_nf4"]
 
-        original_fp32_modules = getattr(self.model_class, "_keep_in_fp32_modules", None)
-        self.model_class._keep_in_fp32_modules = ["proj_out"]
```
sayakpaul, line 584: How is this implemented otherwise? It seems to be implemented for models we don't test bitsandbytes for:

https://github.com/search?q=repo%3Ahuggingface%2Fdiffusers%20_keep_in_fp32_modules&type=code

IMO, we could probably make it a property for easier testing or something similar.

tests/models/transformers/test_models_transformer_flux.py
sayakpaul: Oh I guess `dummy_input()` is now deriving the `self.torch_dtype` property. Nice. But how to extend to batch sizes? 👀 Okay to do in follow-up.

sayakpaul (commented):
I have this comment: https://github.com/huggingface/diffusers/pull/13481/changes#r3092426590

src/diffusers/models/transformers/transformer_wan_vace.py
DN6: When running `device_map` tests across multiple devices the vace blocks end up on a different device, so the control_hidden_states tensor needs to be moved to the hidden_states tensor in order for inference to work.

tests/models/testing_utils/quantization.py
DN6: It exists in the original BnB tests
https://github.com/huggingface/diffusers/blob/c8c84018e0d8704e44d68ff18b634f1f61a717f6/tests/quantization/bnb/test_4bit.py#L235

Here instead of manually adding a dummy keep in fp32 module with `self.model_class._keep_in_fp32_modules = ["proj_out"]` we just use the one that has been defined for the model class if it exists. 

If we use fp16 for inference it will likely fail because the it's artificially injected fp32 module might not handle  the casting properly. The original test seems to place the inference part under an autocast, but that jsut means the keep in FP32 module runs in FP16.

sayakpaul (approved):

PR #13485, opened 2026-04-15
[Neuron] Enable `torch.compile` compatibility with Neuron device

JingyaHuang:
# What does this PR do?

Following #13289

Adding torch.compile support for Neuron to diffusers.

Neuron supports `torch.compile` with the following mechanism to improve performance:

```
torch.compile(model, backend="neuron", fullgraph=True)                                                                                                 
           │                                                
           ▼                                                                                                                                             
     TorchDynamo                   traces Python → FX graph (aten ops)
           │                                                                                                                                             
           ▼                                                
     AOT Autograd                  lifts params as inputs, decomposes ops                                                                                
           │                                                                                                                                             
           ▼
     Neuron backend                Neuron-specific decompositions                                                                                        
     (torch_neuronx)               (int64→int32, SDPA lowering, NKI passes)
           │  StableHLO                                                                                                                                  
           ▼                                                                                                                                             
     neuronx-cc                    compiles HLO → NEFF binary                                                                                            
                                   allocates on-chip SRAM (SB, 28 MB/core)                                                                               
           │  NEFF                                                                                                                                       
           ▼                                                                                                                                             
     NEFF cache                    keyed by StableHLO hash                                                                                               
     (TORCH_NEURONX_NEFF_CACHE_DIR)  miss → recompile; hit → load instantly                                                                              
           │                                                                                                                                             
           ▼                                                                                                                                             
     Executor                      preprocesses inputs (contiguous, int64→int32)                                                                         
     (executor.py)                 calls _C.execute_compiled_graph() 
```

This PR integrates `torch.compile` on Neuron seamlessly into Diffusers via the following changes:

* In `src/diffusers/models/transformers/transformer_flux2.py`

   - `Flux2PosEmbed.forward()` — add Neuron to the `float32` RoPE guard (Neuron doesn't support float64, same pattern as MPS/NPU)
   - `Flux2SwiGLU.forward()` — replace x.chunk(2, dim=-1) with explicit x[..., :half] / x[..., half:] slices (equivalent on all backends; avoids a gather op in the Neuron IR that causes NCC_ITEN404 DataLocalityOpt compilation failure under `torch.compile`)
* In `src/diffusers/models/unets/unet_2d_condition.py`
   - `get_time_embed()` — use int32 timestep dtype on Neuron (Neuron doesn't support int64)                                                                 
                                                                                                                                                      
* In `src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py`                                                                      
  - Keep timesteps on CPU for Neuron (same pattern as XLA); 
  - cast to float32 before moving to device; 
  - run scheduler step on CPU output                    
                                                                                                                                                         
* In `src/diffusers/utils/import_utils.py + torch_utils.py`                                                                                                  
  - Register the neuron backend: BACKEND_SYNCHRONIZE, BACKEND_SUPPORTS_TRAINING, BACKEND_DEVICE_COUNT, randn_tensor forces CPU random generation then `.to(device)` (Neuron has no native RNG) 

* a test for pixart

## Validation

So far we validated the following models, the idea is to manually validated a part of representative models, and ensure the model coverage with pur CIs afterward.

* pixart
* sdxl
* flux2-klein-4B (able to run within one NeuronCore w/o. tp)

## Who can review?

## Caveats

Somes fixes need to be upsteamed to TorchNeuron, this PR won't be stable with potential changes in the SDK


src/diffusers/models/transformers/transformer_flux2.py
```diff
@@ -291,8 +291,8 @@ def __init__(self):
         self.gate_fn = nn.SiLU()
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        x1, x2 = x.chunk(2, dim=-1)
-        x = self.gate_fn(x1) * x2
+        half = x.shape[-1] // 2
+        x = self.gate_fn(x[..., :half]) * x[..., half:]
```
sayakpaul, line 296: Does it have any performance limitations?

src/diffusers/models/_modeling_parallel.py
```diff
@@ -64,9 +64,6 @@ class ContextParallelConfig:
             Whether to enable "Ulysses Anything" mode, which supports arbitrary sequence lengths and head counts that
             are not evenly divisible by `ulysses_degree`. When enabled, `ulysses_degree` must be greater than 1 and
             `ring_degree` must be 1.
-        ring_anything (`bool`, *optional*, defaults to `False`):
-            Whether to enable "Ring Anything" mode, which supports arbitrary sequence lengths. When enabled,
-            `ring_degree` must be greater than 1 and `ulysses_degree` must be 1.
```
sayakpaul, line 69: Seems unrelated to this PR.

src/diffusers/models/_modeling_parallel.py
```diff
@@ -119,13 +114,6 @@ def __post_init__(self):
                 raise ValueError("ulysses_degree must be greater than 1 for ulysses_anything to be enabled.")
             if self.ring_degree > 1:
                 raise ValueError("ulysses_anything cannot be enabled when ring_degree > 1.")
-        if self.ring_anything:
-            if self.ring_degree == 1:
-                raise ValueError("ring_degree must be greater than 1 for ring_anything to be enabled.")
-            if self.ulysses_degree > 1:
-                raise ValueError("ring_anything cannot be enabled when ulysses_degree > 1.")
-        if self.ulysses_anything and self.ring_anything:
-            raise ValueError("ulysses_anything and ring_anything cannot both be enabled.")
```
sayakpaul, line 128: Seems unrelated to this PR.

sayakpaul (approved):
I think there's some overlap with https://github.com/huggingface/diffusers/pull/13289?

But rest looks good to me.

sayakpaul:
@claude-2-serge could you do a review?

src/diffusers/models/transformers/transformer_flux2.py
JingyaHuang: both of them are zero-copy tensor views so should be the same performance wise, the change just avoid the graph break for dynamo

sayakpaul: I think you meant the Neuron dynamo? Because it passes regular CUDA during compilation when `fullgraph=True` during compilation.

src/diffusers/pipelines/flux2/pipeline_flux2_klein.py
```diff
@@ -826,7 +827,8 @@ def __call__(
         # 7. Denoising loop
         # We set the index here to remove DtoH sync, helpful especially during compilation.
         # Check out more details here: https://github.com/huggingface/diffusers/pull/11696
-        self.scheduler.set_begin_index(0)
+        if hasattr(self.scheduler, "set_begin_index"):
```
sayakpaul, line 830: This seems like a bigger change that should be propagated to other pipelines that use `self.scheduler.set_begin_index(0)`?

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
```diff
@@ -903,7 +907,8 @@ def __call__(
 
         # 7. Denoising loop
         num_warmup_steps = max(len(timesteps) - num_inference_steps * self.scheduler.order, 0)
-
+        if hasattr(self.scheduler, "set_begin_index"):
+            self.scheduler.set_begin_index(0)
```
sayakpaul, line 911: If the pipeline initially didn't have it, maybe we don't need it?

tests/pipelines/flux2/test_pipeline_flux2_klein.py
```diff
@@ -181,3 +189,96 @@ def test_image_input(self):
...
+        self.assertGreater(image_slice.std(), 0.01, "Output image should have meaningful variance")
+
+        atol = 1e-2 if is_torch_neuronx_available() else 1e-4
+        _ = atol
+
+    @unittest.skipUnless(is_torch_neuronx_available(), "torch_neuronx not available")
+    def test_flux2_klein_neuron_compile_128(self):
+        import torch_neuronx  # noqa: F401 — registers torch.neuron
+        from torch_neuronx.neuron_dynamo_backend import set_model_name
+        from transformers.utils.output_capturing import install_all_output_capturing_hooks
```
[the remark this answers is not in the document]
sayakpaul: Why do we need `install_all_output_capturing_hooks`? `test_pixart_512_neuron_compile` doesn't need it, it seems.

sayakpaul (commented):
Left some further comments.

src/diffusers/models/transformers/transformer_flux2.py
JingyaHuang: yeah, here is totally a fix for neuron compilation. neuron compiler lower `chunk` to split/gather and previously introduced a compilation failure, now it's fixed by neuron's new beta drop. But the explicit static slices `x[..., :half] / x[..., half:]` lower to clean slice ops, and the neuron compiler compiles it faster.

src/diffusers/pipelines/pixart_alpha/pipeline_pixart_alpha.py
JingyaHuang: The change was suggested in the initial PR for supporting neuron: https://github.com/huggingface/diffusers/pull/13289#discussion_r3264338156. 

the line is needed for the compiled path otherwise the first `scheduler.step` calls `index_for_timestep`, which does a comparison/nonzero and forces a DtoH sync inside the compiled denoising loop (same motivation as #11696).

src/diffusers/pipelines/flux2/pipeline_flux2_klein.py
JingyaHuang: yes, only a part of self.scheduler have `set_begin_index`, so it will be good to have the check, I only added for the pipelines related to the PR, do you want me to propagate to others? Or I could open a separate PR for this.

tests/pipelines/flux2/test_pipeline_flux2_klein.py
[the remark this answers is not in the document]
JingyaHuang: It was added for the compilation of the text encoder, which is not a component we compile in general, removed it!

JingyaHuang (commented):
Thanks @sayakpaul for the review! I addressed the comments you made.

sayakpaul (approved):

yiyixuxu:
thanks a ton for working on this @JingyaHuang !

PR #13486, opened 2026-04-15
Fix Flux2 non-diffusers guidance LoRA conversion

### LINKED ISSUE #13484: Lora loading fail on Flux 2 dev for FAL loras
I can't load some Fal loras on the Flux2.dev pipe:

First example: "lovis93/Flux-2-Multi-Angles-LoRA-v2"

### Reproduction

To reproduce the bug:

```python
import torch

pipe.load_lora_weights("lovis93/Flux-2-Multi-Angles-LoRA-v2")

```

### Logs

```shell
ValueError: `original_state_dict` should be empty at this point but has original_state_dict.keys()=dict_keys(['guidance_in.in_layer.lora_A.weight', 'guidance_in.in_layer.lora_B.weight', 'guidance_in.out_layer.lora_A.weight', 'guidance_in.out_layer.lora_B.weight']).
```

### System Info

Diffusers commit 71a6fd9f0df04d3764dfa999268a05d87903a85a

### Who can help?

@yiyixuxu @DN6 @sayakpaul

yadferhad:
## Summary
- map Flux2 `guidance_in.*` non-diffusers LoRA keys to the diffusers guidance embedder
- raise a clear error when Flux2 LoRAs are loaded into the text-only modular sub-pipeline
- add regression tests covering both behaviors

## Root cause
The Flux2 non-diffusers conversion table handled `time_in.*` weights but omitted `guidance_in.*`, so those keys remained in the original checkpoint and caused conversion to fail. The text-only modular sub-pipeline also has no `transformer`, so the old loading path fell through to a misleading error instead of rejecting the call directly.

## Validation
- `python -m ruff check src/diffusers/loaders/lora_conversion_utils.py src/diffusers/loaders/lora_pipeline.py tests/lora/test_lora_conversion_utils_flux2.py`
- `python -m pytest tests/lora/test_lora_conversion_utils_flux2.py -q`
- `Flux2LoraLoaderMixin.lora_state_dict("lovis93/Flux-2-Multi-Angles-LoRA-v2")` now returns 340 converted weights and includes `guidance_embedder` keys

Fixes #13484


yadferhad:
I checked the extra issue context that was added after this PR went up and widened the fix accordingly.

Originally this PR only covered the missing `guidance_in.*` mapping plus the text-only sub-pipeline guard. After reviewing the added example, I updated the Flux2 non-diffusers conversion path to also accept expanded Flux2 block names directly, including the legacy `sformer_blocks.*` alias for double-stream blocks.

Validation on this branch:
- `python -m pytest tests/lora/test_lora_conversion_utils_flux2.py -q`
- `Flux2LoraLoaderMixin.lora_state_dict("fal/FLUX.2-dev-Turbo")`
- explicit checks with both `weight_name="flux.2-turbo-lora.safetensors"` and `weight_name="comfy/Flux_2-Turbo-LoRA_comfyui.safetensors"`

One other note: the current Hub snapshot for `fal/FLUX.2-dev-Turbo` now appears to resolve both advertised weight paths to the same blob, so the `sformer_blocks.*` example looks like an older export/checkpoint variant. The updated normalization still covers that variant if it shows up again.

tests/lora/test_lora_conversion_utils_flux2.py
```diff
@@ -0,0 +1,103 @@
+# coding=utf-8
```
sayakpaul: Not needed. These are community checkpoints which we cannot exhaustively check.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -5703,6 +5703,13 @@ def load_lora_weights(
         """
         See [`~loaders.StableDiffusionLoraLoaderMixin.load_lora_weights`] for more details.
         """
+        transformer = getattr(self, self.transformer_name, None)
+        if transformer is None:
+            raise ValueError(
+                "Flux2 LoRA weights can only be loaded into a pipeline that defines a `transformer` component. "
+                "This modular sub-pipeline only exposes a subset of components."
+            )
```
sayakpaul: No need.

src/diffusers/loaders/lora_pipeline.py
```diff
@@ -5726,7 +5733,7 @@ def load_lora_weights(
 
         self.load_lora_into_transformer(
             state_dict,
-            transformer=getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer,
+            transformer=transformer,
```
sayakpaul: No need.

src/diffusers/loaders/lora_conversion_utils.py
```diff
@@ -2331,6 +2331,20 @@ def _convert_non_diffusers_flux2_lora_to_diffusers(state_dict):
             temp_state_dict[new_key] = v
         original_state_dict = temp_state_dict
 
+    # Some Flux2 checkpoints skip the ai-toolkit `single_blocks` / `double_blocks`
+    # layout and already store expanded diffusers block names. Accept those
+    # directly, and normalize the legacy `sformer_blocks` alias used by some exports.
+    expanded_block_prefixes = {
```
sayakpaul:
```suggestion
    possible_expanded_block_prefixes = {
```

sayakpaul (commented):
Thanks for your PR!

src/diffusers/loaders/lora_conversion_utils.py
yadferhad: Applied this suggestion. I kept the conversion-path fix for the expanded Flux2 block names and the `guidance_in.*` mapping.

tests/lora/test_lora_conversion_utils_flux2.py
yadferhad: Removed the added Flux2-specific test file.

src/diffusers/loaders/lora_pipeline.py
yadferhad: Reverted the temporary `transformer` variable here as well.

src/diffusers/loaders/lora_pipeline.py
yadferhad: Reverted this extra `transformer` presence guard.

sayakpaul (approved):

sayakpaul:
@claude could you do a review as well?

sayakpaul:
Thanks for the PR!

PR #13495, opened 2026-04-17
Fix load_lora_weights crash on modular sub-pipelines without transformer

### LINKED ISSUE #13487: Loading loras on "text_pipe" of a modular pipeline fails
### Describe the bug

The following code works:

```python
import torch

pipe.load_lora_weights("flux.2-turbo-lora.safetensors")
```

but on modular it doesn't:

```python
import torch

pipe = ModularPipeline.from_pretrained("black-forest-labs/FLUX.2-dev")
blocks = pipe.blocks
text_blocks = blocks.sub_blocks.pop("text_encoder")
text_pipe = text_blocks.init_pipeline("black-forest-labs/FLUX.2-dev")
text_pipe.load_components(torch_dtype=torch.bfloat16)
text_pipe.load_lora_weights("flux.2-turbo-lora.safetensors")
text_pipe.set_adapters(["default"], adapter_weights=[1])
```

The lora is found here: https://huggingface.co/fal/FLUX.2-dev-Turbo/blob/main/flux.2-turbo-lora.safetensors

### Logs

```shell
src/diffusers/configuration_utils.py", line 176, in __getattr__
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'Flux2ModularPipeline' object has no attribute 'transformer'. Did you mean: 'transformer_name'?
```

### System Info

Latest commit of Diffusers 71a6fd9f0df04d3764dfa999268a05d87903a85a

### Who can help?

 @yiyixuxu @DN6 @sayakpaul

ParamChordiya:
## Summary

Fixes #13487

- `load_lora_weights` crashes with `AttributeError: 'Flux2ModularPipeline' object has no attribute 'transformer'` when called on a modular sub-pipeline that doesn't have the transformer component (e.g., a text-encoder-only sub-pipeline)
- Root cause: `load_lora_weights` unconditionally accesses `self.transformer` via unsafe `getattr`/`hasattr` patterns
- Fix: use safe attribute access (`getattr` with default `None`) and skip loading with a warning when the target component is not available

### Changes

**`Flux2LoraLoaderMixin.load_lora_weights`**: checks if transformer exists; warns and returns early if missing.

**`FluxLoraLoaderMixin.load_lora_weights`**: independently guards transformer and text_encoder loading, so a sub-pipeline with only one of them still works. Only returns early if both are absent.

**Test**: added `test_load_lora_weights_warns_when_transformer_missing` to verify the fix.

### Before
```python
text_pipe.load_lora_weights("flux.2-turbo-lora.safetensors")
# AttributeError: 'Flux2ModularPipeline' object has no attribute 'transformer'
```

### After
```python
text_pipe.load_lora_weights("flux.2-turbo-lora.safetensors")
# WARNING: The `transformer` component is not available in this pipeline. Skipping LoRA weight loading.
```

## Test plan
- [ ] Verify `load_lora_weights` on a full `Flux2Pipeline` still works normally
- [ ] Verify `load_lora_weights` on a text-encoder-only modular sub-pipeline warns and skips without crashing
- [ ] Verify `load_lora_weights` on a denoise-only modular sub-pipeline loads transformer LoRA correctly
- [ ] Run existing LoRA tests: `pytest tests/lora/test_lora_layers_flux2.py`


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -1674,9 +1674,19 @@ def load_lora_weights(
             and any(norm_key in k for norm_key in self._control_lora_supported_norm_keys)
         }
 
-        transformer = getattr(self, self.transformer_name) if not hasattr(self, "transformer") else self.transformer
+        transformer = getattr(self, self.transformer_name, None)
```
yiyixuxu, line 1677: so we don't actually have flux lora that touches text_encoder, and the current implementation is based on that assumption. I think we don't need to change that. instead, we can just skip and output a warning if the transformer is None, what do you think?

yiyixuxu (commented):
thanks for the PR!
i left one comment

src/diffusers/loaders/lora_pipeline.py
ParamChordiya: Updated the fix to only guard the transformer and return early with a warning if it's None. Thanks for the review!

PR #13496, opened 2026-04-17
add _repeated_blocks for ErnieImageTransformer2DModel

kaixuanliu:
This PR fixes failed test case:
`tests/models/transformers/test_models_transformer_ernie_image.py::TestErnieImageTransformerCompile::test_torch_compile_repeated_blocks`
@yiyixuxu pls help review, thx!


dg845 (approved):
Thanks for the PR! I think we can remove the `pytest.mark.skip` decorators from the skipped tests in `TestErnieImageTransformerCompile` since they all pass after this fix.

kaixuanliu:
@dg845 Thx for your review, but there is a comment: "The repeated block in this model is ErnieImageSharedAdaLNBlock. As a consequence of this, the inputs recorded for the block would vary during compilation and full compilation with fullgraph=True would trigger recompilation." in [L119](https://github.com/huggingface/diffusers/blob/main/tests/models/transformers/test_models_transformer_ernie_image.py#L119). Hence I am not sure if it is suitable to remove the `pytest.mark.skip` decorators here. Maybe @HsiaWinter can give some advice on this?

dg845:
Merging this PR as I think we can handle the other skipped `TestErnieImageTransformerCompile` tests separately in a follow-up PR (and the CI failures are unrelated). @HsiaWinter would still be interested in your input as to whether the `_repeated_blocks` fix in this PR solves the model compile issues (the currently skipped tests pass after the fix).

PR #13500, opened 2026-04-18
Fix: use resample-aware bilinear+antialias interpolation for tensor/numpy resize in VaeImageProcessor

GitGlimpse895:
# What does this PR do?

`VaeImageProcessor` exposes a `resample` config parameter (defaulting to `"lanczos"`) 
and correctly applies it when resizing **PIL images** via `PIL_INTERPOLATION`. However, 
the two `torch.nn.functional.interpolate` calls handling **`torch.Tensor`** and 
**`np.ndarray`** inputs passed no `mode` argument — causing PyTorch to silently default 
to `"nearest"` neighbor interpolation, regardless of the configured resample filter. 
No `antialias=True` was set either, causing aliasing artifacts on downsampling.

This fix:
- Adds a `TORCH_INTERPOLATION` dict in `pil_utils.py` mapping the same resample-string 
  keys as `PIL_INTERPOLATION` to their `torch.nn.functional.interpolate` equivalents 
  (with `antialias` eligibility). `"lanczos"` maps to `bilinear+antialias`, the closest 
  high-quality torch substitute.
- Updates both tensor branches of `VaeImageProcessor.resize()` to use the mapped mode 
  and antialias flag, making tensor/numpy resize quality consistent with the PIL path.

This silently affected every pipeline that passes `torch.Tensor` inputs to 
`VaeImageProcessor` (ControlNet conditioning, IP-Adapter, img2img, etc.).

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if 
      that's the case).
- [ ] Did you read our [philosophy doc](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md) 
      (important for complex PRs)?
- [ ] Was this discussed/approved via a GitHub issue or the 
      [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? 
      Please add a link to it if that's the case.

## Who can review?
@yiyixuxu @sayakpaul @DN6

### DIFF
--- a/src/diffusers/utils/pil_utils.py
+++ b/src/diffusers/utils/pil_utils.py
@@ -21,6 +21,17 @@ else:
         "nearest": PIL.Image.NEAREST,
     }
 
+# Maps VaeImageProcessor `resample` config strings to torch.nn.functional.interpolate modes.
+# "lanczos" has no native torch equivalent; bilinear+antialias is the closest high-quality substitute.
+# The bool indicates whether antialias=True should be passed (not supported for "nearest").
+TORCH_INTERPOLATION = {
+    "linear": ("bilinear", True),
+    "bilinear": ("bilinear", True),
+    "bicubic": ("bicubic", True),
+    "lanczos": ("bilinear", True),
+    "nearest": ("nearest", False),
+}
+
 
 def pt_to_pil(images):
     """
--- a/src/diffusers/utils/pil_utils.py
+++ b/src/diffusers/utils/pil_utils.py
@@ -21,14 +21,13 @@ else:
         "nearest": PIL.Image.NEAREST,
     }
 
-# Maps VaeImageProcessor `resample` config strings to torch.nn.functional.interpolate modes.
-# "lanczos" has no native torch equivalent; bilinear+antialias is the closest high-quality substitute.
-# The bool indicates whether antialias=True should be passed (not supported for "nearest").
+# Resample modes natively supported by torch.nn.functional.interpolate.
+# Modes absent from this dict (e.g. "lanczos") are not supported; callers
+# should warn and fall back to default "nearest" behavior.
 TORCH_INTERPOLATION = {
     "linear": ("bilinear", True),
     "bilinear": ("bilinear", True),
     "bicubic": ("bicubic", True),
-    "lanczos": ("bilinear", True),
     "nearest": ("nearest", False),
 }
 
--- a/src/diffusers/image_processor.py
+++ b/src/diffusers/image_processor.py
@@ -506,24 +506,25 @@ class VaeImageProcessor(ConfigMixin):
             else:
                 raise ValueError(f"resize_mode {resize_mode} is not supported")
 
-        elif isinstance(image, torch.Tensor):
-            torch_mode, use_antialias = TORCH_INTERPOLATION[self.config.resample]
+        elif isinstance(image, (torch.Tensor, np.ndarray)):
+            resample = self.config.resample
+            if resample not in TORCH_INTERPOLATION:
+                logger.warning(
+                    f"The resample mode '{resample}' is not supported for torch.Tensor/np.ndarray inputs "
+                    f"and will be ignored. Supported modes are: {list(TORCH_INTERPOLATION.keys())}. "
+                    "Falling back to default 'nearest' interpolation."
+                )
+                torch_mode, use_antialias = "nearest", False
+            else:
+                torch_mode, use_antialias = TORCH_INTERPOLATION[resample]
+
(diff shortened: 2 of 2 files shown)


ParamChordiya:
## Code Review: Fix resample-aware interpolation in VaeImageProcessor

### Summary

This fixes a legitimate quality bug: `VaeImageProcessor.resize()` correctly applies the configured `resample` filter for PIL inputs, but the `torch.Tensor` and `np.ndarray` branches called `F.interpolate` with no `mode` argument, defaulting to `"nearest"` neighbor — producing blocky/aliased results. The `TORCH_INTERPOLATION` mapping dict and `(mode, antialias)` tuple pattern is clean.

### Issues

1. **`"lanczos"` maps to `"bilinear"` — should it be `"bicubic"`?** — `"bicubic"` is a closer approximation to Lanczos in frequency response and sharpness. Since `"lanczos"` is the **default** resample value, this mapping affects the vast majority of users. Worth discussing.

2. **No new tests** — Existing tests only check output shapes, not interpolation quality or mode. Needs at minimum:
   - A test verifying interpolation is not `"nearest"` (e.g. checkerboard pattern resize)
   - A test that each key in `TORCH_INTERPOLATION` works without error
   - A test comparing tensor vs PIL resize output similarity (PSNR/MSE threshold)

3. **Import path inconsistency** — `PIL_INTERPOLATION` is exported via `utils/__init__.py`, but `TORCH_INTERPOLATION` is imported directly from `utils.pil_utils`. Should be consistent.

4. **Silent behavior change** — This changes outputs for all existing pipelines passing tensor/numpy inputs. Deserves a changelog entry flagging it as a fix that changes output values.

5. **No `KeyError` guard on mapping lookup** — Invalid resample strings will produce an unhelpful `KeyError`. A `ValueError` with supported options listed would be better.

6. **Duplicate code** — The `TORCH_INTERPOLATION` lookup is identical in both the tensor and numpy branches. Could factor it out above with `isinstance(image, (torch.Tensor, np.ndarray))`.

7. **`resize_and_crop_tensor` not updated** — This static method also calls `F.interpolate` with hardcoded `mode="bilinear"` and no `antialias`. Worth a follow-up.

### Verdict

**Request changes** — The core fix is correct, but needs tests, the lanczos mapping deserves discussion, and the import path should be consistent.

src/diffusers/utils/pil_utils.py
```diff
@@ -21,6 +21,17 @@
...
     }
 
+# Maps VaeImageProcessor `resample` config strings to torch.nn.functional.interpolate modes.
+# "lanczos" has no native torch equivalent; bilinear+antialias is the closest high-quality substitute.
+# The bool indicates whether antialias=True should be passed (not supported for "nearest").
+TORCH_INTERPOLATION = {
+    "linear": ("bilinear", True),
+    "bilinear": ("bilinear", True),
+    "bicubic": ("bicubic", True),
+    "lanczos": ("bilinear", True),
```
yiyixuxu: ohh, so if this option is not supported in torch, let's not map it to anything
just send a warning that says this `resample` mode is not supported for tensor/ndarray so it will be ignored (the default nearest is used intead). this way we don't change the default behavior for resize 
what do you think?

yiyixuxu (commented):
thanks for the PR!
i left one comment

GitGlimpse895:
Thanks @yiyixuxu — updated! Revised approach:

- `TORCH_INTERPOLATION` now only maps modes torch natively supports
  (`bilinear`, `bicubic`, `nearest`).
- For unsupported modes like `"lanczos"`, the code now emits a
  `logger.warning` and falls back to `"nearest"`, preserving existing
  default behavior with no silent output change.
- Also factored the duplicate lookup out of both tensor and numpy
  branches into a single shared `isinstance(image, (torch.Tensor, np.ndarray))`
  branch, eliminating the code duplication @ParamChordiya flagged.

GitGlimpse895:
Hi @yiyixuxu — just a gentle ping to let you know the PR has been updated per your feedback. The unsupported resample mode (e.g. `lanczos`) now emits a warning and falls back to `nearest` rather than silently remapping, which should keep the default behavior intact. Would love to get your eyes on it when you have a moment. Thanks!

GitGlimpse895:
Hi @yiyixuxu — just a gentle ping to let you know the PR has been updated per your feedback. The unsupported resample mode (e.g. lanczos) now emits a warning and falls back to nearest rather than silently remapping, which should keep the default behavior intact. Would love to get your eyes on it when you have a moment. Thanks!

PR #1902, opened 2023-01-03
[Deterministic torch randn] Allow tensors to be generated on CPU

patrickvonplaten:
This PR adds a new helper function `torch_randn` that has three purposes:
- Allow fully deterministic generation by creating the tensors on CPU as also demanded here: https://github.com/huggingface/diffusers/issues/1514
- Avoids lots of duplicated code, as the same 5,6 lines are implemented in 10,20+ pipeline files.
- Helps us to better tests models by always creating tensors on CPU

This is the first PR of a two PR series to see if this helps to make UnCLIP deterministic across different CUDA, GPUs. In a follow-up PR, I'd then like to replace all existing `torch.randn` functions with this function as well as add a `reproducibility.md` guide.

PR #13508, opened 2026-04-20
fix(randn_tensor): compare device.type, not torch.device, when suppressing MPS info log

Ricardo-M-L:
## What does this PR do?

Fixes a latent bug in `randn_tensor` where the MPS-specific info-log suppression has been broken since PR #1902 (Jan 2023).

### The bug

```python
# src/diffusers/utils/torch_utils.py
if isinstance(device, str):
    device = torch.device(device)             # ← device is now a torch.device object
...
if generator is not None:
    gen_device_type = ...
    if gen_device_type != device.type and gen_device_type == "cpu":
        rand_device = "cpu"
        if device != "mps":                   # ← always True
            logger.info("The passed generator was created on 'cpu' ...")
```

The `if device != "mps":` guard was meant to silence the info log on MPS, because MPS doesn't support device-side generators — suggesting the user "create a generator on the mps device" would be misleading.

But by this point `device` is a `torch.device` object (coerced a few lines earlier). `torch.device("mps") == "mps"` is `False` — `torch.device.__eq__` returns `NotImplemented` when compared with a string, Python falls back to identity, and the two are different types. So the guard is effectively always true, and MPS users get the spurious log on every call where a CPU generator is passed — the opposite of the documented intent.

Sanity check:

```python
>>> import torch
>>> torch.device("mps") == "mps"
False
>>> torch.device("mps") != "mps"
True
>>> torch.device("mps").type == "mps"
True
```

### The fix

One-character change: compare `device.type` (a `str`) to `"mps"`.

```diff
-            if device != "mps":
+            if device.type != "mps":
```

### Test

Added `RandnTensorTester.test_mps_suppresses_cpu_generator_info_log` in `tests/others/test_utils.py`, which:

- Asserts **no** CPU-fallback info log is emitted when a CPU generator targets MPS.
- Asserts the log **is** emitted when a CPU generator targets a non-MPS accelerator.

Verified that the test fails on `main` and passes with this fix. All existing tests in `tests/others/test_utils.py` still pass (14/14).

## Before submitting
- [x] This PR fixes a bug that affects MPS users.
- [x] Did you write any new necessary tests? — Yes, regression test added.

## Who can review?

@sayakpaul @DN6

### DIFF
--- a/tests/others/test_utils.py
+++ b/tests/others/test_utils.py
@@ -247,6 +247,54 @@ class FourierFilterTester(unittest.TestCase):
             assert out.shape == x.shape
 
 
+class RandnTensorTester(unittest.TestCase):
+    """Tests for :func:`diffusers.utils.torch_utils.randn_tensor`."""
+
+    def test_mps_suppresses_cpu_generator_info_log(self):
+        """
+        When a CPU generator targets MPS, the informational log about falling back to CPU
+        should be suppressed — MPS does not support device-side generators, so the message
+        would be misleading. Prior to fixing the ``device != "mps"`` string/``torch.device``
+        comparison this log fired on every device, including MPS.
+        """
+        import logging as py_logging
+
+        import torch
+
+        from diffusers.utils import logging as diffusers_logging
+        from diffusers.utils import torch_utils
+
+        gen = torch.Generator(device="cpu")
+
+        diffusers_logging.set_verbosity_info()
+        prev_level = torch_utils.logger.level
+        torch_utils.logger.setLevel(py_logging.INFO)
+
+        def _capture(target_device):
+            with self.assertLogs(torch_utils.logger, level="INFO") as cm:
+                torch_utils.logger.info("sentinel")
+                try:
+                    torch_utils.randn_tensor((1, 2), generator=gen, device=target_device, dtype=torch.float32)
+                except (AssertionError, RuntimeError):
+                    pass
+            return [m for m in cm.output if "sentinel" not in m]
+
+        try:
+            mps_logs = _capture("mps")
+            self.assertFalse(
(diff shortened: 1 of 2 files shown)


tests/others/test_utils.py
```diff
@@ -204,6 +204,56 @@ def test_deprecate_testing_utils_module(self):
...
+        from diffusers.utils import logging as diffusers_logging
+        from diffusers.utils import torch_utils
+
+        gen = torch.Generator(device="cpu")
+
+        diffusers_logging.set_verbosity_info()
+        prev_level = torch_utils.logger.level
+        torch_utils.logger.setLevel(py_logging.INFO)
+
+        def _capture(target_device):
```
yiyixuxu, line 230: we can use this? https://github.com/huggingface/diffusers/blob/main/tests/testing_utils.py#L1320

yiyixuxu (approved):
thanks
I left a comment

tests/others/test_utils.py
Ricardo-M-L: Good call! Refactored the test to use `CaptureLogger` — much cleaner. Also rebased on latest main to resolve the conflict.

Ricardo-M-L:
Gentle nudge — this has been APPROVED for ~2 weeks now and all checks are green. @yiyixuxu would you mind merging when you get a moment? Happy to rebase or address anything if needed; thanks!

Ricardo-M-L:
Hi @sayakpaul @DN6, gentle bump — one-line fix for an MPS info-log suppression that's been silently broken since #1902 (compares `torch.device` to `"mps"`, which is always non-equal). A regression test is included that fails on `main` and passes with the fix. Happy to rebase or address any feedback whenever there's time 🙏

PR #13526, opened 2026-04-21
Fix TypeError in LongCatImageEditPipeline truncation warning

Ricardo-M-L:
## What this PR does

`LongCatImageEditPipeline._encode_prompt` calls `len()` twice on an `int` when building its truncation warning message:

https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/longcat_image/pipeline_longcat_image_edit.py#L284-L289

```python
if len(all_tokens) > self.tokenizer_max_length:
    logger.warning(
        "Your input was truncated because \`max_sequence_length\` is set to "
        f" {self.tokenizer_max_length} input token nums : {len(len(all_tokens))}"
    )
    all_tokens = all_tokens[: self.tokenizer_max_length]
```

`len(all_tokens)` already returns an `int`, so the outer `len()` raises `TypeError: object of type 'int' has no len()` — inside the f-string formatting — every time the branch fires.

## Why this is a real bug

The warning only executes when `len(all_tokens) > self.tokenizer_max_length` (default 512). That is the exact scenario the warning is supposed to inform the user about. Instead of informing them, the pipeline crashes with a `TypeError` during prompt encoding.

Minimal repro:

```python
pipe = LongCatImageEditPipeline.from_pretrained(...)
pipe(prompt="a very long prompt ..." * 200, image=image)
# TypeError: object of type 'int' has no len()
```

The sibling `LongCatImagePipeline._encode_prompt` has the correct form at [line 291](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/longcat_image/pipeline_longcat_image.py#L291) — this is a local typo in the edit pipeline only.

## Fix

Drop the extraneous `len()` call to match the sibling pipeline:

```diff
-                f" {self.tokenizer_max_length} input token nums : {len(len(all_tokens))}"
+                f" {self.tokenizer_max_length} input token nums : {len(all_tokens)}"
```

## Before submitting
- [x] Was this discussed/approved via a Github issue or the forum? N/A — single-character typo fix.
- [x] Did you make sure to update the documentation with your changes? N/A.
- [x] Did you write any new necessary tests? N/A — fixes a log-message crash; no behavior change.

## Who can review?
@yiyixuxu @sayakpaul

### DIFF
--- a/src/diffusers/pipelines/longcat_image/pipeline_longcat_image_edit.py
+++ b/src/diffusers/pipelines/longcat_image/pipeline_longcat_image_edit.py
@@ -284,7 +284,7 @@ class LongCatImageEditPipeline(DiffusionPipeline, FromSingleFileMixin):
         if len(all_tokens) > self.tokenizer_max_length:
             logger.warning(
                 "Your input was truncated because `max_sequence_length` is set to "
-                f" {self.tokenizer_max_length} input token nums : {len(len(all_tokens))}"
+                f" {self.tokenizer_max_length} input token nums : {len(all_tokens)}"
             )
             all_tokens = all_tokens[: self.tokenizer_max_length]
 


Ricardo-M-L:
Polite bump on this one — it's a 1-character TypeError fix (\`len(len(all_tokens))\` would raise \`TypeError: object of type 'int' has no len()\` the first time the truncation path actually fires). cc @yiyixuxu in case you have a moment.

CI is green, sitting since 2026-04-21. Happy to address any feedback. 🙏

Ricardo-M-L:
Hi @yiyixuxu @sayakpaul, friendly ping — this is a one-character typo fix (drops a duplicate `len()` inside an f-string) that turns the existing truncation warning into a `TypeError` whenever the user actually hits the truncation path in `LongCatImageEditPipeline`. The sibling `LongCatImagePipeline` already has the correct form. Happy to address any feedback 🙏

Ricardo-M-L:
Hi team, just following up on this PR. Happy to make any changes or address feedback if needed. Thanks for considering it!

Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

Ricardo-M-L:
Friendly ping - this PR has been open for a while. Would appreciate a review when you get a chance. Thanks!

yiyixuxu:
hi @Ricardo-M-L

I'm closing this per the AI-assisted contribution guidelines: https://huggingface.co/docs/diffusers/main/en/conceptual/contribution#ai-assisted-and-agentic-contributions

We're seeing an increasing volume of contributions and review capacity is currently our bottleneck, so the guidelines ask for maintainer acknowledgment on an issue before a PR is opened — that's what lets us prioritize the bugs users are actually hitting.

Normally we'd leave this as a suggestion and keep working with a contributor on the PR, but the volume here has made that impractical — these have come in faster than we can review them, so I'm closing all of them that's opened bu you rather than reviewing them individually.

PR #13535, opened 2026-04-21
fix the test_hotswapping_compiled_model_xxx case failure in xpu

sywangyi:
following case fail
tests/models/transformers/test_models_transformer_bria.py::BriaTransformerLoRAHotSwapTests::test_hotswapping_compiled_model_linear_2
tests/models/transformers/test_models_transformer_bria.py::BriaTransformerLoRAHotSwapTests::test_hotswapping_compiled_model_both_linear_and_other_2
tests/models/transformers/test_models_transformer_chroma.py::ChromaTransformerLoRAHotSwapTests::test_hotswapping_compiled_model_linear_2
tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelLoRAHotSwapTests::test_hotswapping_compiled_model_conv2d_2
tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelLoRAHotSwapTests::test_hotswapping_compiled_model_both_linear_and_conv2d_2
tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelLoRAHotSwapTests::test_hotswapping_compiled_model_both_linear_and_other_2
tests/models/unets/test_models_unet_2d_condition.py::UNet2DConditionModelLoRAHotSwapTests::test_hotswapping_compiled_model_linear_2
tests/models/transformers/test_models_transformer_chroma.py::ChromaTransformerLoRAHotSwapTests::test_hotswapping_compiled_model_both_linear_and_other_2

### DIFF
--- a/src/diffusers/loaders/peft.py
+++ b/src/diffusers/loaders/peft.py
@@ -311,12 +311,28 @@ class PeftAdapterMixin:
                     state_dict = map_state_dict_for_hotswap(state_dict)
                     check_hotswap_configs_compatible(self.peft_config[adapter_name], lora_config)
                     try:
-                        hotswap_adapter_from_state_dict(
-                            model=self,
-                            state_dict=state_dict,
-                            adapter_name=adapter_name,
-                            config=lora_config,
-                        )
+                        # When enable_lora_hotswap was called, weights were padded to target_rank so that
+                        # torch.compile does not recompile after hotswapping. peft's hotswap_adapter_from_state_dict
+                        # uses swap_tensors for un-compiled models, which replaces the parameter tensor object.
+                        # On backends without CUDA Graphs (e.g. XPU), the compiled inductor code re-fetches
+                        # the real model parameters on every call and its assert_size_stride checks fail because
+                        # the swapped-in tensor has the smaller (un-padded) rank shape.
+                        # Fix: temporarily signal to peft that this model should be treated as compiled so
+                        # it uses the in-place data copy path (which preserves both the tensor object and the
+                        # padded shape).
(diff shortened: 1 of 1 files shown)


sywangyi:
@sayakpaul please help review

sayakpaul:
@claude could you review this?

src/diffusers/loaders/peft.py
```diff
@@ -311,12 +311,28 @@ def map_state_dict_for_hotswap(sd):
                     state_dict = map_state_dict_for_hotswap(state_dict)
                     check_hotswap_configs_compatible(self.peft_config[adapter_name], lora_config)
                     try:
-                        hotswap_adapter_from_state_dict(
-                            model=self,
-                            state_dict=state_dict,
-                            adapter_name=adapter_name,
-                            config=lora_config,
-                        )
+                        # When enable_lora_hotswap was called, weights were padded to target_rank so that
```
BenjaminBossan, line 314: If we keep things like this, how about turning this into a context manager to keep the code inside of `load_lora_adapter` more readable? It would also allow to re-use the context manager if it's ever needed.

BenjaminBossan (commented):
I'm agreeing with Claude's 3rd point, namely that I'd rather like to see this fixed in PEFT. Is there a technical reason why it's not possible? Otherwise, a similar patch would need to be rolled out to transformers and to every user of PEFT hotswapping with a similar use case and XPU.

sywangyi:
@BenjaminBossan thanks for the review, I put the fix in peft. https://github.com/huggingface/peft/pull/3183

sywangyi:
this WA fix could be close and put the fix in peft

PR #13543, opened 2026-04-22
Fix AdaLayerNormZeroSingle.forward return type annotation

Ricardo-M-L:
## What does this PR do?

`AdaLayerNormZeroSingle.forward()` has a return type annotation declaring a 5-element tuple:

```python
def forward(self, x, emb=None) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
    ...
    return x, gate_msa  # actually returns 2 elements
```

The 5-element annotation was copied from `AdaLayerNormZero.forward()` which does return 5 elements (`x, gate_msa, shift_mlp, scale_mlp, gate_mlp`), but the "Single" variant only returns 2 (`x, gate_msa`).

All callers correctly unpack 2 elements (e.g., `norm_hidden_states, gate = self.norm(hidden_states, emb=temb)` in FluxSingleTransformerBlock), so this is just a type annotation fix.

## Before submitting

## Who can review?
@yiyixuxu @sayakpaul

### DIFF
--- a/src/diffusers/models/normalization.py
+++ b/src/diffusers/models/normalization.py
@@ -195,7 +195,7 @@ class AdaLayerNormZeroSingle(nn.Module):
         self,
         x: torch.Tensor,
         emb: torch.Tensor | None = None,
-    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+    ) -> tuple[torch.Tensor, torch.Tensor]:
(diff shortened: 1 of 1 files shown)


Ricardo-M-L:
Closing this in favor of the substantive bug fixes I have open in the queue. A return-type annotation fix is low-priority and shouldn't compete for review attention with the actual crash/AttributeError fixes.

PR #13564, opened 2026-04-26
Eliminate GPU sync overhead and CPU→GPU transfers across LTX2 pipeline

ViktoriiaRomanova:
Fixes performance issues identified by profiling LTX2 with torch.profiler as part of #13401.

Optimises LTX2 by removing unnecessary GPU synchronisation points and replacing CPU tensor creation with on-device tensor operations across the decoding pipeline, transformer RoPE computations, scheduler, and connector padding logic.

## Pipeline Denoising Optimisation

1. Added explicit set_begin_index(0) calls to both video and audio schedulers. This avoids the DtoH sync in _init_step_index. Uses the same pattern as the issue fixed in PR #11696.
Before (eager mode):
image
After (eager mode, no sync gap):
image

Before (compile mode):
image
After (compile mode, no sync gap):
image

2. Replaced torch.tensor(..., device=device) with on-device torch.stack([torch.ones(...)*s for s in decode_noise_scale]). Avoids CPU tensor allocation and GPU transfers for decode noise scaling.

## Transformer Model Optimisation

Replaced CPU tensor creation for patch sizes with on-device tensor construction.
Eliminates unnecessary CPU-to-GPU memcpy operations during RoPE coordinate preparation.

## Connector Refactoring

Replaced list-comprehension-based padding logic with vectorised masking.  This simplifies left-padding layout logic and eliminates unnecessary cudaStreamSynchronize calls.

## Performance Results
|Metric	| Before | After |
| -------- | -------- | -------- |
|cudaStreamSynchronize calls (total)	| 18	| 6 |
| Scheduler sync (eager mode)	 |233ms |	eliminated |
|Scheduler sync (compiled mode) |	573ms |	eliminated |
| Other syncs total (eager mode) |	88ms |	25ms |
| Other syncs total (compiled mode)|	93ms |	25ms |

### Profiler trace
https://drive.google.com/drive/folders/1cZn1xw-8Eon22mA2zP1uoF1nE4YCC3Wo?usp=drive_link

## Before submitting
- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? Please add a link to it if that's the case. #13401

## Who can review?
@sayakpaul @dg845


src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -894,9 +894,14 @@ def prepare_video_coords(
...
-        patch_size_delta = torch.tensor(patch_size, dtype=grid.dtype, device=grid.device)
-        patch_ends = grid + patch_size_delta.view(3, 1, 1, 1)
+        patch_size_delta = torch.stack(
+            [
+                grid.new_ones(1) * self.patch_size_t,
+                grid.new_ones(1) * self.patch_size,
+                grid.new_ones(1) * self.patch_size,
+            ]
+        ).reshape(3, 1, 1, 1)
+        patch_ends = grid + patch_size_delta
```
sayakpaul: This refactor seems unnecessary.

src/diffusers/models/transformers/transformer_ltx2.py
```diff
@@ -905,7 +910,7 @@ def prepare_video_coords(
         latent_coords = latent_coords.unsqueeze(0).repeat(batch_size, 1, 1, 1)
 
         # 3. Calculate the pixel space patch boundaries from the latent boundaries.
-        scale_tensor = torch.tensor(self.scale_factors, device=latent_coords.device)
+        scale_tensor = torch.stack([latent_coords.new_ones(1) * factor for factor in self.scale_factors])
```
sayakpaul: Same.

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
```diff
@@ -362,11 +362,13 @@ def set_timesteps(
             sigmas = self._convert_to_beta(in_sigmas=sigmas, num_inference_steps=num_inference_steps)
 
         # 5. Convert sigmas and timesteps to tensors and move to specified device
-        sigmas = torch.from_numpy(sigmas).to(dtype=torch.float32, device=device)
+        sigmas = torch.from_numpy(sigmas).pin_memory().to(dtype=torch.float32, device=device, non_blocking=True)
```
sayakpaul: This is a hard no. We cannot be pinning memory and running it async within the scheduler.

src/diffusers/pipelines/ltx2/pipeline_ltx2.py
```diff
@@ -1189,6 +1189,11 @@ def __call__(
         num_warmup_steps = max(len(timesteps) - num_inference_steps * self.scheduler.order, 0)
         self._num_timesteps = len(timesteps)
 
+        # Set begin index to skip nonzero().item() call in scheduler initialization, which triggers GPU sync
+        if hasattr(self.scheduler, "set_begin_index"):
+            self.scheduler.set_begin_index(0)
+            audio_scheduler.set_begin_index(0)
```
sayakpaul: Move it out of the `set_begin_index` hasattr check.

sayakpaul (commented):
Thanks for the PR. Please provide comments inline to the changes explaining how they eliminate the syncs.

src/diffusers/models/transformers/transformer_ltx2.py
ViktoriiaRomanova: This replaces host-side tensor construction with device-native ops to eliminate an implicit cudaStreamSynchronize (~60ms).

src/diffusers/models/transformers/transformer_ltx2.py
ViktoriiaRomanova: Also avoids implicit cudaStreamSynchronize (~60ms) by replacing torch.tensor(...) with device-native tensor construction.

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
ViktoriiaRomanova: @sayakpaul  Should I provide an alternative implementation to avoid the cudaStreamSynchronize, or is this sync considered acceptable in this case?

src/diffusers/models/transformers/transformer_ltx2.py
dg845: This refactor eliminates a CPU --> GPU sync (since the `patch_size` tuple lives on the CPU host, `torch.tensor` needs to copy it to GPU, which the refactor avoids by doing the operations on the GPU), but because the tuple is really small, it looks like the corresponding `cudaStreamSynchronize` block (and that of the similar `scale_tensor` refactor below) takes about 2ms. I think most of the removed non-scheduler sync time is in the `connectors.py` refactor.

dg845:
@bot /style

dg845 (commented):
Thanks for the PR! I agree with https://github.com/huggingface/diffusers/pull/13564#pullrequestreview-4210213892 that having comments for the changes would be helpful.

src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
dg845: My current thoughts are that an alternative implementation which avoids the `cudaStreamSynchronize` would be good, but accepting the sync would be preferable to the `pin_memory` implementation. (As far as I can tell, this particular sync point overlaps with GPU kernels, so it's probably not in the critical path.)

dg845 (approved):
Thanks for the follow-up changes!

ViktoriiaRomanova:
> Thanks for the follow-up changes!

Happy to contribute. Please let me know if there are any other tasks I could take part in.

P.S. The test failures do not appear to be related to this PR.

dg845:

Issue #13641, opened 2026-04-29
hunyuandit model/pipeline review


hlky:
# `hunyuandit` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

Reviewed: target model/pipeline files, lazy exports, config/loading paths, dtype/device/offload behavior, attention processors, docs, examples, and tests. Fast and slow tests exist for both base and ControlNet pipelines, but ControlNet has several no-op fast tests noted below. Local `.venv` snippets were run; full pytest collection was blocked because this `.venv` torch build is missing `torch._C._distributed_c10d`.

Duplicate search: searched GitHub Issues and PRs for `hunyuandit`, target class/file names, `from_transformer transformer_num_layers`, `prompt_embeds num_images_per_prompt attention_mask`, `MultiControlNet controlnet_conditioning_scale`, `_no_split_modules`, and `HunyuanAttnProcessor2_0 set_attention_backend`. No exact duplicates found. Related closed issue: https://github.com/huggingface/diffusers/issues/9142 covered older multi-ControlNet loading, not the runtime scale/validation failures below.

## Issue 1: `from_transformer()` cannot build a matching HunyuanDiT ControlNet

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/models/controlnets/controlnet_hunyuan.py#L177-L209

Problem:
`HunyuanDiT2DControlNetModel.from_transformer()` reads `config.transformer_num_layers`, but `HunyuanDiT2DModel` stores `num_layers`. Even when `transformer_num_layers` is passed manually, the method does not copy `pooled_projection_dim` or `use_style_cond_and_image_meta_size`, so weight loading can hit size mismatches.

Impact:
The public constructor for deriving a ControlNet from a transformer is broken for normal HunyuanDiT transformers and can also create a ControlNet with incompatible conditioning embeddings.

Reproduction:
```python
from diffusers import HunyuanDiT2DControlNetModel, HunyuanDiT2DModel

transformer = HunyuanDiT2DModel(
    sample_size=8, num_layers=4, patch_size=2,
    attention_head_dim=4, num_attention_heads=2, in_channels=4,
    cross_attention_dim=8, cross_attention_dim_t5=8,
    pooled_projection_dim=4, hidden_size=8, text_len=4, text_len_t5=4,
[...]

PR #13658, opened 2026-04-30
fix(controlnet_hunyuan): wire num_layers, pooled_projection_dim, use_style_cond into from_transformer (#13641)

Anai-Guo:
## Summary

Fixes Issue 1 in #13641.

`HunyuanDiT2DControlNetModel.from_transformer()` was reading `config.transformer_num_layers`, but `HunyuanDiT2DModel` stores the layer count under `config.num_layers`. The default call path therefore raised `AttributeError`. Even when `transformer_num_layers` was passed manually, the method also dropped `pooled_projection_dim` and `use_style_cond_and_image_meta_size`, so a ControlNet built from a non-default transformer silently fell back to defaults (`pooled_projection_dim=1024`, `use_style_cond_and_image_meta_size=True`) and could fail weight loading with size mismatches.

## Fix

- Read `config.num_layers` (the actual `HunyuanDiT2DModel` config attribute) rather than the non-existent `transformer_num_layers`.
- Forward `pooled_projection_dim` and `use_style_cond_and_image_meta_size` from the source transformer config, so the ControlNet is constructed with the same conditioning embedding shapes and style-cond setting as the transformer it derives from.

## Verification

The reproduction snippet from #13641 now works:

```python
from diffusers import HunyuanDiT2DControlNetModel, HunyuanDiT2DModel

transformer = HunyuanDiT2DModel(
    sample_size=8, num_layers=4, patch_size=2,
    attention_head_dim=4, num_attention_heads=2, in_channels=4,
    cross_attention_dim=8, cross_attention_dim_t5=8,
    pooled_projection_dim=4, hidden_size=8, text_len=4, text_len_t5=4,
    use_style_cond_and_image_meta_size=False,
)

cn = HunyuanDiT2DControlNetModel.from_transformer(transformer)  # no AttributeError
assert cn.config.transformer_num_layers == 4
assert cn.config.pooled_projection_dim == 4
assert cn.config.use_style_cond_and_image_meta_size is False
```

Existing default-config call sites (e.g. with the public `Tencent-Hunyuan/HunyuanDiT-Diffusers` checkpoint) are unchanged: the default `pooled_projection_dim=1024` and `use_style_cond_and_image_meta_size=True` match the public config exactly, so they round-trip to the same values that were previously hardcoded by omission.

### DIFF
--- a/src/diffusers/models/controlnets/controlnet_hunyuan.py
+++ b/src/diffusers/models/controlnets/controlnet_hunyuan.py
@@ -189,7 +189,9 @@ class HunyuanDiT2DControlNetModel(ModelMixin, ConfigMixin):
         text_len_t5 = config.text_len_t5
 
         conditioning_channels = conditioning_channels
-        transformer_num_layers = transformer_num_layers or config.transformer_num_layers
+        transformer_num_layers = transformer_num_layers or config.num_layers
+        pooled_projection_dim = config.pooled_projection_dim
+        use_style_cond_and_image_meta_size = config.use_style_cond_and_image_meta_size
 
         controlnet = cls(
             conditioning_channels=conditioning_channels,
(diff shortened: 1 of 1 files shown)


qgallouedec:

Issue #13629, opened 2026-04-29
stable_audio model/pipeline review


hlky:
# `stable_audio` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

Files/categories reviewed: target pipeline/model files, public lazy imports, top-level exports, config/loading/device-map behavior, dtype/device handling, offload-related tests, attention processor behavior, docs, examples, fast/nightly/slow test coverage.

Verification note: attempted `.venv\Scripts\python.exe -m pytest tests/pipelines/stable_audio/test_stable_audio.py -q`, but local test collection fails before Stable Audio tests run because this Windows torch build lacks `torch._C._distributed_c10d` while importing FSDP. Narrow reproduction snippets below were checked with `.venv`.

Duplicate-search status: searched GitHub Issues/PRs for `stable_audio`, `StableAudioPipeline`, `StableAudioDiTModel device_map`, `StableAudioAttnProcessor2_0 set_attention_backend`, and `initial_audio_waveforms num_waveforms_per_prompt`. Found related but not exact duplicates: #10861 for initial-audio scaling and #8989 for sequential offload testing. No exact duplicate found for the batch-order, `_no_split_modules`, attention-backend, dtype, or docs findings.

## Issue 1: Batched initial audio is paired with the wrong prompt when generating multiple waveforms

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/stable_audio/pipeline_stable_audio.py#L485-L487
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/stable_audio/pipeline_stable_audio.py#L670-L680

Problem:
`text_audio_duration_embeds` is expanded per prompt as `[prompt0, prompt0, prompt1, prompt1]`, but encoded initial audio is expanded with `encoded_audio.repeat((num_waveforms_per_prompt, 1, 1))`, producing `[audio0, audio1, audio0, audio1]`. For batched audio-to-audio with `num_waveforms_per_prompt > 1`, prompts and initial audio become misaligned.

Impact:
Users requesting multiple variations per prompt with batched `initial_audio_waveforms` condition some generations on another prompt's audio. Existing tests only assert output shape, so this does not get caught.

Reproduction:
```python
from types import SimpleNamespace
import torch
[...]

PR #13659, opened 2026-04-30
fix(stable_audio): align batched initial audio with prompts (#13629)

Anai-Guo:
## Summary

Fixes Issue 1 in #13629.

`StableAudioPipeline.prepare_latents()` was expanding batched initial audio with `encoded_audio.repeat((num_waveforms_per_prompt, 1, 1))`, which produces an interleaved layout `[audio0, audio1, audio0, audio1]`. The corresponding text+duration embeds are expanded per prompt by `encode_prompt()` to `[prompt0, prompt0, prompt1, prompt1]`. For batched audio-to-audio with `num_waveforms_per_prompt > 1`, prompts and initial audio became misaligned, so each generation could be conditioned on another prompt’s initial audio. Existing tests only assert output shape and missed it.

## Fix

Switch `repeat` to `repeat_interleave(..., dim=0)` so batched initial audio expands as `[audio0, audio0, audio1, audio1]`, matching the prompt expansion.

## Verification

The reproduction snippet from #13629 now prints the expected order:

```python
import torch
from diffusers import StableAudioPipeline
# (same DummyVAE / SimpleNamespace scaffolding as in the issue)
print(latents[:, 0, 0].tolist())  # [10.0, 10.0, 20.0, 20.0]
```

`num_waveforms_per_prompt == 1` is unchanged because `repeat_interleave` and `repeat` are equivalent in that case.

### DIFF
--- a/src/diffusers/pipelines/stable_audio/pipeline_stable_audio.py
+++ b/src/diffusers/pipelines/stable_audio/pipeline_stable_audio.py
@@ -483,7 +483,7 @@ class StableAudioPipeline(DiffusionPipeline):
             audio[:, :, : min(audio_length, audio_vae_length)] = initial_audio_waveforms[:, :, :audio_vae_length]
 
             encoded_audio = self.vae.encode(audio).latent_dist.sample(generator)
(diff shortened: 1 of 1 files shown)


qgallouedec:

PR #13665, opened 2026-05-01
[agents docs] update models.md with class attributes and attention mask

yiyixuxu:


.ai/models.md
```diff
@@ -61,23 +61,109 @@ class MyModelAttention(nn.Module, AttentionModuleMixin):
...
+- **Padding mask → bool `(B, L)` or `(B, 1, 1, L)`.** Only pass when the batch actually contains different-length sequences (i.e. there is real padding). If all sequences are the same length, set the mask to `None` — many backends (flash, sage, aiter) raise `ValueError` on any non-None mask, and even SDPA-based backends pay unnecessary overhead processing a no-op mask. See `pipeline_qwenimage.py` `encode_prompt` for the pattern: `if mask.all(): mask = None`. When a mask is needed, use bool format — it stays compatible with the `*_varlen` kernels via `_normalize_attn_mask` (`attention_dispatch.py:639`), which reduces bool masks to `cu_seqlens`. Dense additive-float masks *cannot* be reduced this way and so lose the varlen path.
+- **Other mask types (structural, BlockMask, etc.)** — if the model requires a different mask pattern, figure out how to support as many backends as possible (e.g. use `window_size` kwarg for sliding window on flash, `BlockMask` for Flex) and document which backends are supported for that model.
```
sayakpaul, line 65: I think we might want to keep the kandinsky reference in case it ends up discovering that Flex would be the way to go.

.ai/models.md
```diff
@@ -61,23 +61,109 @@ class MyModelAttention(nn.Module, AttentionModuleMixin):
...
+
+Lists which `nn.Module` subclasses must stay on a single device (i.e. never have their children placed on different devices).
+
+- **`None` (default)** — `from_pretrained(..., device_map="auto")` raises `ValueError` (`modeling_utils.py:1863`).
+- **`[]`** — split anywhere you like.
+- **`["MyBlock"]`** — keep all `MyBlock` instances intact on one device.
+
+**Why it's needed.** When `accelerate` splits a model across devices, it installs hooks on leaf modules that move inputs to the module's device before `forward` runs. Any inline operation (`+`, `*`, `torch.cat`) that combines tensors from different submodules has no hook — if those submodules landed on different devices, it crashes with "tensors on different devices". The fix is either: (a) list the parent module in `_no_split_modules` so all its children stay co-located, or (b) pack the operation into its own `nn.Module`. Inline ops on outputs from the **same** submodule call are fine since they're already on the same device.
```
sayakpaul, line 82: A good rule of thumb is any block where the forward has a residual connection.

sayakpaul (commented):
I like the dead code path addition. It's quite nice!

.ai/models.md
yiyixuxu: let's see if it can figure it out by itself next time when it reviews a model that only works with flex
I think it's pretty straightforward, but we can always add a reference if it struggles

Issue #13653, opened 2026-04-29
`pipeline_infrastructure` model/pipeline review


hlky:
# `pipeline_infrastructure` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

Files reviewed: `src/diffusers/pipelines/__init__.py`, `src/diffusers/pipelines/auto_pipeline.py`, `src/diffusers/pipelines/free_init_utils.py`, `src/diffusers/pipelines/free_noise_utils.py`, `src/diffusers/pipelines/onnx_utils.py`, `src/diffusers/pipelines/pipeline_flax_utils.py`, `src/diffusers/pipelines/pipeline_loading_utils.py`, `src/diffusers/pipelines/pipeline_utils.py`, `src/diffusers/pipelines/transformers_loading_utils.py`.

Duplicate search: checked GitHub Issues and PRs for `pipeline_infrastructure`, FreeNoise failures, `OnnxRuntimeModel provider_options`, AutoPipeline ControlNetUnion, and DDUF tokenizer loading. No duplicates found. Related but not duplicate: https://github.com/huggingface/diffusers/pull/10661 added `provider_options`.

## Issue 1: Direct ONNX model loading requires an optional kwarg

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/onnx_utils.py#L179-L205

Problem:
`OnnxRuntimeModel._from_pretrained()` calls `kwargs.pop("provider_options")` in both local and Hub paths. Pipeline loading passes `provider_options=None`, but direct public `OnnxRuntimeModel.from_pretrained(...)` does not, so it raises `KeyError` before loading the model.

Impact:
The exported ONNX helper cannot be used directly with default arguments.

Reproduction:
```python
import tempfile
from diffusers.pipelines.onnx_utils import OnnxRuntimeModel

with tempfile.TemporaryDirectory() as d:
    OnnxRuntimeModel.from_pretrained(d)
# KeyError: 'provider_options'
```

Relevant precedent:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/onnx_utils.py#L78-L81

Suggested fix:
```python
provider_options = kwargs.pop("provider_options", None)
[...]

PR #13667, opened 2026-05-01
fix(onnx): make provider_options optional in OnnxRuntimeModel._from_pretrained (#13653)

Anai-Guo:
## Problem

Fixes Issue 1 of #13653.

`OnnxRuntimeModel._from_pretrained()` calls `kwargs.pop("provider_options")` without a default in both the local-directory and Hub branches. Pipeline loading happens to pass `provider_options=None`, so the failure is hidden in that path. Direct use of the public helper, however, raises `KeyError` before any loading happens.

## Reproduction

```python
import tempfile
from diffusers.pipelines.onnx_utils import OnnxRuntimeModel

with tempfile.TemporaryDirectory() as d:
    OnnxRuntimeModel.from_pretrained(d)
# KeyError: 'provider_options'
```

## Fix

Pass a `None` default to both `kwargs.pop("provider_options", None)` calls. The downstream `OnnxRuntimeModel.load_model()` already normalises `None → []` (lines 86-89), so the new default keeps existing pipeline-loading behaviour identical and unblocks direct `from_pretrained()` calls.

## Notes

- Two-line change, scoped strictly to lines 185 and 205 of `onnx_utils.py`.
- No new tests added — the maintainer's review note already flags missing test coverage for direct `OnnxRuntimeModel.from_pretrained()` and lists this in the meta-issue test-coverage section, which felt like an explicit follow-up rather than something to slip into this surgical fix. Happy to add a fast unit test if preferred.

### DIFF
--- a/src/diffusers/pipelines/onnx_utils.py
+++ b/src/diffusers/pipelines/onnx_utils.py
@@ -182,7 +182,7 @@ class OnnxRuntimeModel:
                 Path(model_id, model_file_name).as_posix(),
                 provider=provider,
                 sess_options=sess_options,
-                provider_options=kwargs.pop("provider_options"),
+                provider_options=kwargs.pop("provider_options", None),
             )
             kwargs["model_save_dir"] = Path(model_id)
         # load model from hub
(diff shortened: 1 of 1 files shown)


qgallouedec:

Issue #13649, opened 2026-04-29
ddpm model/pipeline review


hlky:
# `ddpm` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

Coverage status: fast DDPM tests exist, and a slow accelerator integration test exists. Local run: `2 passed, 1 skipped` for `tests/pipelines/ddpm/test_ddpm.py`; the skipped test is marked `@slow`.

## Issue 1: `DDPMPipeline` does not run latents on the offload execution device

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/ddpm/pipeline_ddpm.py#L111-L139

Problem:
`DDPMPipeline` declares `model_cpu_offload_seq = "unet"` but initializes `image` on `self.device`, not `self._execution_device`, and never calls `self.maybe_free_model_hooks()` before returning. Under `enable_model_cpu_offload()`, `self.device` remains CPU while the UNet forward is executed on the accelerator by accelerate hooks. That leaves the scheduler step operating on mismatched CPU/GPU tensors.

Duplicate search:
No duplicate found for `DDPMPipeline enable_model_cpu_offload`, `DDPMPipeline maybe_free_model_hooks`, or `DDPMPipeline _execution_device`.

Impact:
Users enabling CPU offload can hit device mismatch failures or leave the UNet resident on the accelerator after the call, defeating the offload contract.

Reproduction:
```python
import torch
from diffusers import DDPMPipeline, DDPMScheduler, UNet2DModel

if not torch.cuda.is_available():
    raise SystemExit("CUDA/accelerator required")

unet = UNet2DModel(
    block_out_channels=(4, 8),
    layers_per_block=1,
    norm_num_groups=4,
    sample_size=8,
    in_channels=3,
    out_channels=3,
    down_block_types=("DownBlock2D", "AttnDownBlock2D"),
    up_block_types=("AttnUpBlock2D", "UpBlock2D"),
)
pipe = DDPMPipeline(unet=unet, scheduler=DDPMScheduler(num_train_timesteps=2))
pipe.set_progress_bar_config(disable=True)
pipe.enable_model_cpu_offload(device="cuda")

pipe(num_inference_steps=1, output_type="np")
```

Relevant precedent:
[...]

PR #13663, opened 2026-04-30
Address ernie-image review findings #13577

akshan-main:
# What does this PR do?

Partial fix for #13577. Addresses 1, 2, 5 per @yiyixuxu's scope

- **(1)** Switch `ErnieImageAutoPromptEnhancerStep` to `ConditionalPipelineBlocks` so `use_pe=False` actually skips the prompt enhancer (`AutoPipelineBlocks` selected on presence, not truthiness).
- **(2)** Align modular VAE BN epsilon to the standard pipeline's hardcoded `1e-5` (matches training; the hub config currently reports `1e-4`).
- **(5)** Restructure `output_type=\"latent\"` so it runs `maybe_free_model_hooks()` and honors `return_dict`, matching the QwenImage/Flux2 pattern.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/diffusers/blob/main/CONTRIBUTING.md)?
- [x] Did you read our [philosophy doc](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md) (important for complex PRs)?
- [x] Was this discussed/approved via a GitHub issue? https://github.com/huggingface/diffusers/issues/13577
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

@yiyixuxu

Issue #13656, opened 2026-04-29
[meta issue] Systematic model/pipeline review findings / tracking


hlky:
# [meta issue] Systematic model/pipeline review findings / tracking

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

## Summary

- Reviewed 76 model/pipeline/shared/infrastructure targets
- Aggregated 498 issue-level findings into recurring cross-family patterns
- Findings suggest systemic inconsistencies rather than isolated bugs

These patterns are already generating duplicate low-effort PRs (often agent-generated) for the same underlying issues, increasing maintainer review load without addressing root causes.

## Duplicate Check

Searches for broad/meta tracking issues or PRs did not find an existing systematic tracker. Some individual patterns are partially known through targeted issues/PRs, for example [#11762](https://github.com/huggingface/diffusers/issues/11762), [#9371](https://github.com/huggingface/diffusers/issues/9371), [#8989](https://github.com/huggingface/diffusers/issues/8989), [#12533](https://github.com/huggingface/diffusers/issues/12533), and [PR #13532](https://github.com/huggingface/diffusers/pull/13532), but those do not address the recurring root causes across families.

### Pattern 1: Batch and Conditioning Expansion Drift

Description:
Many pipelines accept batched prompts, images, masks, latents, or `num_images_per_prompt` / `num_videos_per_prompt`, but only expand part of the conditioning state.

Root cause:
Batch construction is duplicated per pipeline instead of enforced by a shared invariant after prompt/image/control/mask preparation.

Impact:
Incorrect conditioning, crashes, silently ignored extra outputs, and non-reproducible batched generation.

Representative examples:
- [wan review](https://github.com/huggingface/diffusers/issues/13578)
- [hunyuan_video1_5 review](https://github.com/huggingface/diffusers/issues/13582)
- [flux review](https://github.com/huggingface/diffusers/issues/13584)
- [consisid review](https://github.com/huggingface/diffusers/issues/13586)
- [cogvideo review](https://github.com/huggingface/diffusers/issues/13622)

### Pattern 2: Public Arguments Accepted but Ignored

Description:
Several public APIs validate or document arguments such as `latents`, `attention_kwargs`, `cross_attention_kwargs`, `max_sequence_length`, `timesteps`, `num_frames`, masks, or callbacks, but do not actually consume them.

Root cause:
[...]

PR #13671, opened 2026-05-01
fix(ddpm): use _execution_device, validate inputs, free hooks (#13649)

Anai-Guo:
## What does this PR do?

Fixes the three issues called out in the [`ddpm` model/pipeline review](https://github.com/huggingface/diffusers/issues/13649) (cc @hlky). All three changes live in `pipelines/ddpm/pipeline_ddpm.py` and follow the suggested fixes from the issue, with precedents from `DDIMPipeline` and `ConsistencyModelPipeline`.

### Issue 1 — `DDPMPipeline` does not run latents on the offload execution device

`DDPMPipeline` declares `model_cpu_offload_seq = "unet"` but initializes the latent on `self.device` (which stays CPU under `enable_model_cpu_offload()`) and never calls `self.maybe_free_model_hooks()` before returning. Switched to `self._execution_device`, threaded that into the `randn_tensor` call (and the `mps` branch), and added the `maybe_free_model_hooks()` call before the return so the offload contract is honored and the UNet doesn't stay resident on the accelerator.

### Issue 2 — Generator lists are not validated against `batch_size`

A short generator list either gets silently treated like a single generator or raises a raw `IndexError`. Added the same `ValueError` guard `DDIMPipeline` and `ConsistencyModelPipeline` use, so users get a clear message when `len(generator) != batch_size`.

### Issue 3 — Invalid `output_type` values silently return NumPy

Any `output_type` other than `"pil"` previously fell through to NumPy, including typos and `"pt"`. Added an explicit `output_type in {"pt", "np", "pil"}` check at the top of `__call__`, made `"pt"` actually return the tensor (skipping the `.cpu().numpy()` round-trip), and updated the docstring to mention `torch.Tensor`. Behavior for the documented values (`"pil"`, default; `"np"`) is unchanged.

## Reproductions

The repros from #13649 are short and self-contained — see the issue body for runnable scripts that demonstrate each failure mode.

## Before submitting
- [x] Did you read the [contributor guideline](https://github.com/huggingface/diffusers/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr)?
- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? Please add a link to it if that's the case. Yes — #13649.
- [ ] Did you make sure to update the documentation with your changes? Doc string updated in this file; no separate docs page touched.

## Who can review?

@hlky — this addresses all three issues from the `ddpm` review.

### DIFF
--- a/src/diffusers/pipelines/ddpm/pipeline_ddpm.py
+++ b/src/diffusers/pipelines/ddpm/pipeline_ddpm.py
@@ -73,7 +73,8 @@ class DDPMPipeline(DiffusionPipeline):
                 The number of denoising steps. More denoising steps usually lead to a higher quality image at the
                 expense of slower inference.
             output_type (`str`, *optional*, defaults to `"pil"`):
-                The output format of the generated image. Choose between `PIL.Image` or `np.array`.
+                The output format of the generated image. Choose between `PIL.Image`, `np.array` or
+                `torch.Tensor`.
             return_dict (`bool`, *optional*, defaults to `True`):
                 Whether or not to return a [`~pipelines.ImagePipelineOutput`] instead of a plain tuple.
 
@@ -97,6 +98,17 @@ class DDPMPipeline(DiffusionPipeline):
                 If `return_dict` is `True`, [`~pipelines.ImagePipelineOutput`] is returned, otherwise a `tuple` is
                 returned where the first element is a list with the generated images
         """
+        if output_type not in ["pt", "np", "pil"]:
+            raise ValueError(f"output_type must be one of ['pt', 'np', 'pil'], got '{output_type}'.")
+
+        if isinstance(generator, list) and len(generator) != batch_size:
+            raise ValueError(
+                f"You have passed a list of generators of length {len(generator)}, but requested an effective batch"
+                f" size of {batch_size}. Make sure the batch size matches the length of the generators."
+            )
+
+        device = self._execution_device
+
         # Sample gaussian noise to begin loop
         if isinstance(self.unet.config.sample_size, int):
             image_shape = (
@@ -108,12 +120,12 @@ class DDPMPipeline(DiffusionPipeline):
         else:
             image_shape = (batch_size, self.unet.config.in_channels, *self.unet.config.sample_size)
 
-        if self.device.type == "mps":
+        if device.type == "mps":
             # randn does not work reproducibly on mps
             image = randn_tensor(image_shape, generator=generator, dtype=self.unet.dtype)
-            image = image.to(self.device)
+            image = image.to(device)
         else:
-            image = randn_tensor(image_shape, generator=generator, device=self.device, dtype=self.unet.dtype)
+            image = randn_tensor(image_shape, generator=generator, device=device, dtype=self.unet.dtype)
 
         # set step values
         self.scheduler.set_timesteps(num_inference_steps)
@@ -129,9 +141,12 @@ class DDPMPipeline(DiffusionPipeline):
                 xm.mark_step()
 
         image = (image / 2 + 0.5).clamp(0, 1)
-        image = image.cpu().permute(0, 2, 3, 1).numpy()
-        if output_type == "pil":
-            image = self.numpy_to_pil(image)
+        if output_type != "pt":
+            image = image.cpu().permute(0, 2, 3, 1).numpy()
+            if output_type == "pil":
+                image = self.numpy_to_pil(image)
+
+        self.maybe_free_model_hooks()
 
         if not return_dict:
             return (image,)


src/diffusers/pipelines/ddpm/pipeline_ddpm.py
```diff
@@ -97,6 +98,17 @@ def __call__(
                 If `return_dict` is `True`, [`~pipelines.ImagePipelineOutput`] is returned, otherwise a `tuple` is
                 returned where the first element is a list with the generated images
         """
+        if output_type not in ["pt", "np", "pil"]:
+            raise ValueError(f"output_type must be one of ['pt', 'np', 'pil'], got '{output_type}'.")
```
hlky, line 102: cc @yiyixuxu Should output_type validation be a review rule?

src/diffusers/pipelines/ddpm/pipeline_ddpm.py
```diff
@@ -97,6 +98,17 @@ def __call__(
...
+
+        if isinstance(generator, list) and len(generator) != batch_size:
+            raise ValueError(
+                f"You have passed a list of generators of length {len(generator)}, but requested an effective batch"
+                f" size of {batch_size}. Make sure the batch size matches the length of the generators."
+            )
```
hlky, line 108: Great, this resolves issue 2

src/diffusers/pipelines/ddpm/pipeline_ddpm.py
```diff
@@ -108,12 +120,12 @@ def __call__(
...
 
-        if self.device.type == "mps":
+        if device.type == "mps":
             # randn does not work reproducibly on mps
             image = randn_tensor(image_shape, generator=generator, dtype=self.unet.dtype)
-            image = image.to(self.device)
+            image = image.to(device)
         else:
-            image = randn_tensor(image_shape, generator=generator, device=self.device, dtype=self.unet.dtype)
+            image = randn_tensor(image_shape, generator=generator, device=device, dtype=self.unet.dtype)
```
hlky, line 128: Great, this + [a later line](https://github.com/huggingface/diffusers/pull/13671/changes#diff-96f14232cbdd535d6eb21ba9146ff3c884742c8f6ed0c9edb032785d43883564R149) resolves issue 1

src/diffusers/pipelines/ddpm/pipeline_ddpm.py
```diff
@@ -129,9 +141,12 @@ def __call__(
                 xm.mark_step()
 
         image = (image / 2 + 0.5).clamp(0, 1)
-        image = image.cpu().permute(0, 2, 3, 1).numpy()
-        if output_type == "pil":
-            image = self.numpy_to_pil(image)
+        if output_type != "pt":
+            image = image.cpu().permute(0, 2, 3, 1).numpy()
+            if output_type == "pil":
+                image = self.numpy_to_pil(image)
```
hlky, line 147: As per additional review in https://github.com/huggingface/diffusers/pull/13663#discussion_r3170192282 this could use `VaeImageProcessor`, this would fix all output_type's without the `if` statements. cc @yiyixuxu For awareness, this is another case that supports introducing `VaeImageProcessor` usage as a review rule.

hlky (commented):
Thanks @Anai-Guo.

I left a few comments. Some are just confirming that the PR resolves the reported issues, and a couple are broader questions for maintainers about whether these patterns should become review/agent rules:

* validating `output_type`
* preferring `VaeImageProcessor.postprocess(...)` over manual clamp/permute/NumPy/PIL handling

Overall, this looks like it addresses the `ddpm` findings from #13649.

Also, feel free to tag me on PRs that come from the review issues / meta issue since I can usually confirm whether they address the reported finding. For approval/merge, please also tag an appropriate maintainer since I can review the issue context but cannot approve. In general, I’m happy to help clarify the reports or point out related patterns from #13656 if useful.

Anai-Guo:
Thanks @hlky for the review! Glad the fix resolves both reported issues.

For the line 147 suggestion — happy to refactor to `VaeImageProcessor.postprocess(...)` in this PR if @yiyixuxu agrees that's the preferred direction; otherwise I can keep this PR scoped to the original two bugs and leave the broader `VaeImageProcessor` migration for a follow-up. Just let me know which you prefer.

Anai-Guo:
Friendly ping @yiyixuxu — this has been waiting since early May. @hlky confirmed it resolves both reported issues from #13649; the only open item is the line-147 question of whether to refactor to `VaeImageProcessor.postprocess(...)` here vs. leaving the broader migration for a follow-up.

Happy to do the `VaeImageProcessor` refactor in this PR if that's the preferred direction — otherwise it's ready to merge scoped to the two original bugs. Just let me know which you'd prefer.

qgallouedec:

PR #13681, opened 2026-05-05
Add a (rolling) KV cache for Wan models to enable autoregressive rollouts ("SelfForcing-style").

gueraf:
# What does this PR do?

- Implements a simple (rolling) KV cache for Wan models to enable autoregressive generation.
- Tries to mirror the KV cache pattern in transformer_flux2.py as well as transformer's DynamicCache as much as possible.
- Videos and byte-level equivalence against upstream Self Forcing tested in https://github.com/gueraf/self-forcing-diffusers/ (see videos attached to release, and inference script [here](https://github.com/gueraf/self-forcing-diffusers/blob/8a58097e7e899f1361adbe6dfef86a60222c4115/scripts/validate_self_forcing_against_upstream.py#L430-L434)).
- This initial PR does not yet implement sink-frame pinning yet, lacks some model-level adjustments (Self Forcing has cross-attention QK norms and per-frame timestep modulation), and does not implement cross attention caching (easy to add, but in reality this is negligible GPU time and often a small regression).
- Add tests for cache append/overwrite, and window eviction behavior.

## Motivation

This is a tightly scoped follow-up to #12773 and a first step toward #12600. The previous draft explored similar functionality but also included Krea-specific experiments and broader integration work.

As for practical use, we (https://odyssey.ml/) would like to rely on the Hugging Face Diffusers ecosystem to ship Self-Forcing-like models without having to ship many custom modules, ideally none.

Progresses #12600

## Before submitting

## Who can review?

@sayakpaul thanks for offering help with this :)


rootonchair:
Hi, I'm very interested in this line of work. It would be great if there is an example inference script for running Wan in Self-Forcing style?

gueraf:
```gh repo clone gueraf/self-forcing-diffusers && \
uv run python scripts/autoregressive_video_generation.py \
  --prompt "A cat walks on the grass, realistic style, high quality" \
  --output ./autoregressive.mp4```

is a starting point if you have a GPU with enough HBM (i tested it on an rtx 6000). 

Caveats: 
- It runs with window_size=-1, i.e. the KV cache grows as we generate. 
- There are some minor patches needed for some Wan model layers to create full parity with the self forcing codebase (norm precisions, norm order, etc.). I'll submit them separately.

sayakpaul:
@zucchini-nlp FYI.

yiyixuxu:
hi @gueraf thanks for the PR

We've been supporting similar use cases (self-forcing + autoregresssive wan based) via Modular diffusers + remote code on hub, see two examples:
* krea: https://huggingface.co/krea/krea-realtime-video
* waypoint: https://huggingface.co/Overworld/Waypoint-1-Small

do you think we support Odyssey in the same way? 

Regarding #12773 and #12600 — we're still keeping an eye on these. At the moment, we don't want to introduce a general cache abstraction in core; we'd rather keep caches model-specific, but follow a consistent pattern across them

gueraf:
Hi @yiyixuxu,

yeah, I looked at both. The thing I'd flag: to ship those, Krea forks ~1400 lines of `WanTransformer3DModel` into `causal_model.py` and Waypoint does roughly the same (shipping Wan instead of diffusers). That's two independent teams each redistributing a near-identical Wan fork via `trust_remote_code` just to thread a per-block rolling KV cache and `frame_offset` through RoPE. The cache classes themselves are near-duplicates.

We could ship Odyssey the same way and probably will to move faster, so I'm not pushing hard on this PR. But it does seem worth a broader conversation: is the long-term plan that anyone doing Self-Forcing-style Wan ships their own forked transformer via remote code? That works, but it's a lot of duplicated surface area, and `trust_remote_code` isn't free (security, composability with quantization / offloading / attention backends).

What's the (ideally measurable) criterion you have in mind for taking action on the framework level? Are there past conversations for other "building blocks" that we could align with?

Cheers,
Fabian

PR #13687, opened 2026-05-07
[ernie-image] use concrete Mistral3Model / Ministral3ForCausalLM types

akshan-main:
# What does this PR do?

Follow-up to #13663 (per @asomoza's review there): replace the `AutoModel` / `AutoModelForCausalLM` references for `text_encoder` and `pe` with the concrete `Mistral3Model` / `Ministral3ForCausalLM` so the loaded checkpoint matches the declared types.

To avoid breaking installs with older `transformers` (`Ministral3ForCausalLM` is only in 5.0+, `Mistral3Model` is in 4.50+):

- Standard: `TYPE_CHECKING` + string annotations on `__init__` so the imports never run at module load. `_get_signature_types` uses `get_type_hints` and falls back gracefully when a name can't be resolved.
- Modular `encoders.py`: `try/except` import that aliases the concrete class when available and falls back to `AutoModel` / `AutoModelForCausalLM` otherwise. The `ComponentSpec` then references the alias.

So users on recent transformers get the silenced warning; users on older transformers keep working as before.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? https://github.com/huggingface/diffusers/pull/13663

## Who can review?

@yiyixuxu @asomoza


asomoza:
thanks @akshan-main for what understand, this is not the best fix, since with the fallback (as it was before) it will still not work because the models are not present in `transformers`. So probably instead of the fallback we should raise an error but the version depends if they use or not the PE @yiyixuxu WDYT?

yiyixuxu:
ohh I just thought it was an issue with CI on our end, not something you need to fix
can you just keep the change (not the fix), so that we can see if the CI fails here and look into it

akshan-main:
@yiyixuxu kept just the type change

yiyixuxu:
can you try update the minimum heree https://github.com/huggingface/diffusers/blob/main/setup.py#L143
not sure if it'd fix this

cc @sayakpaul @DN6 
we also run into this in https://github.com/huggingface/diffusers/pull/13551

akshan-main:
@yiyixuxu done

sayakpaul:
Hmm, I think `Ministral3ForCausalLM` should be guarded with something like:

```py
if is_transformers_version("<", "5.0.0"):
   raise
```

This way, I don't think we have to update the `setup.py` (I don't mind the change, either as we made Diffusers compatible with Transformers v5 a while back). In our CI, we install `transformers` from `main`.

akshan-main:
@sayakpaul @yiyixuxu applied it, lmk if I need to revert

yiyixuxu:
@sayakpaul 
Do you want to look into this? Maybe open a separate PR to test it out
Guarding against 5.0 is fine, but our CI still errors out

yiyixuxu:
@akshan-main 
do you want to send your email to yiyi@huggingface.co so we can connect on slacks?

akshan-main:
Sent! Really appreciate the chance to contribute more!

sayakpaul:
Yeah I will look into it.

sayakpaul:
Checking if https://github.com/huggingface/diffusers/pull/13723 works 👀

PR #13693, opened 2026-05-07
Fix Wan batch conditioning and config handling

### LINKED ISSUE #13578: `wan` model/pipeline review
# `wan` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

Duplicate search status: searched existing `huggingface/diffusers` Issues and PRs for `Wan`, affected class names, and specific failure modes. I found related but non-duplicate Wan issues/PRs, including `#12348`, `#12574`, `#12496`, and `#11582`; no likely duplicates for the actionable findings below.

## Issue 1: `WanImageToVideoPipeline` does not expand image conditioning for `num_videos_per_prompt`

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_i2v.py#L691-L699
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_i2v.py#L770-L786

Problem:
`image_embeds` is repeated by `batch_size`, but not by `num_videos_per_prompt`. With `num_videos_per_prompt > 1`, latent inputs are expanded to the final effective batch while image embeddings remain smaller, causing a transformer concat failure.

Impact:
Users cannot generate multiple videos per prompt in image-to-video mode. This also leaves prompt/image batching semantics inconsistent with the rest of the pipeline.

Reproduction:
```python
import torch
from PIL import Image
from transformers import AutoConfig, AutoTokenizer, T5EncoderModel, CLIPImageProcessor, CLIPVisionConfig, CLIPVisionModelWithProjection
from diffusers import AutoencoderKLWan, FlowMatchEulerDiscreteScheduler, WanImageToVideoPipeline, WanTransformer3DModel

vae = AutoencoderKLWan(base_dim=3, z_dim=16, dim_mult=[1, 1, 1, 1], num_res_blocks=1, temperal_downsample=[False, True, True])
scheduler = FlowMatchEulerDiscreteScheduler(shift=7.0)
config = AutoConfig.from_pretrained("hf-internal-testing/tiny-random-t5")
text_encoder = T5EncoderModel(config)
tokenizer = AutoTokenizer.from_pretrained("hf-internal-testing/tiny-random-t5")
transformer = WanTransformer3DModel(
    patch_size=(1, 2, 2), num_attention_heads=2, attention_head_dim=12,
    in_channels=36, out_channels=16, text_dim=32, freq_dim=256,
    ffn_dim=32, num_layers=1, cross_attn_norm=True,
    qk_norm="rms_norm_across_heads", rope_max_seq_len=32, image_dim=4,
)
image_encoder_config = CLIPVisionConfig(
    hidden_size=4, projection_dim=4, num_hidden_layers=1,
    num_attention_heads=2, image_size=32, intermediate_size=16, patch_size=1,
)
pipe = WanImageToVideoPipeline(
    tokenizer, text_encoder, vae, scheduler,
    CLIPImageProcessor(crop_size=32, size=32),
    CLIPVisionModelWithProjection(image_encoder_config),
    transformer=transformer,
).to("cpu")

pipe(
    image=Image.new("RGB", (16, 16)),
    prompt="test prompt",
    negative_prompt="negative",
    height=16,
    width=16,
    num_frames=9,
    num_inference_steps=1,
    guidance_scale=1.0,
    num_videos_per_prompt=2,
    max_sequence_length=16,
    output_type="latent",
    generator=torch.Generator(device="cpu").manual_seed(0),
)
```

Relevant precedent:
Standard text/video batch expansion in Wan uses `batch_size * num_videos_per_prompt` before denoising.

Suggested fix:
```python
target_batch_size = batch_size * num_videos_per_prompt

if image_embeds.shape[0] == 1:
    image_embeds = image_embeds.repeat_interleave(target_batch_size, dim=0)
elif image_embeds.shape[0] == batch_size:
    image_embeds = image_embeds.repeat_interleave(num_videos_per_prompt, dim=0)
elif image_embeds.shape[0] != target_batch_size:
    raise ValueError(
        f"`image_embeds` batch size must be 1, {batch_size}, or {target_batch_size}, "
        f"but got {image_embeds.shape[0]}."
    )
```

## Issue 2: `WanImageToVideoPipeline` rejects documented list image inputs

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_i2v.py#L350-L355
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_i2v.py#L509-L527

Problem:
The public call signature accepts `PipelineImageInput`, but `check_inputs` only accepts a single `PIL.Image.Image` or `torch.Tensor`. A list of PIL images is rejected before preprocessing.

Impact:
Prompt/image batching through list inputs does not work, despite being part of the public type contract.

Reproduction:
```python
from types import SimpleNamespace
from PIL import Image
from diffusers import WanImageToVideoPipeline

pipe_like = SimpleNamespace(
    _callback_tensor_inputs=["latents", "prompt_embeds", "negative_prompt_embeds"],
    config=SimpleNamespace(boundary_ratio=None),
)

WanImageToVideoPipeline.check_inputs(
    pipe_like,
    prompt="test",
    negative_prompt=None,
    image=[Image.new("RGB", (16, 16))],
    height=16,
    width=16,
    prompt_embeds=None,
    negative_prompt_embeds=None,
    image_embeds=None,
    callback_on_step_end_tensor_inputs=["latents"],
    guidance_scale_2=None,
)
```

Relevant precedent:
Other image-conditioned pipelines validate `PipelineImageInput` through shared image/video processor semantics instead of rejecting lists at `check_inputs`.

Suggested fix:
Allow `list`/`tuple` image inputs in `check_inputs`, validate their members, and align the resulting image latents and image embeddings to `batch_size * num_videos_per_prompt`.

## Issue 3: `WanVACEPipeline` returns reference latents in `output_type="latent"`

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_vace.py#L930-L935
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_vace.py#L1020-L1036

Problem:
When `reference_images` are provided, VACE prepends reference latents. The decode path trims those latents, but the latent-output path returns them untrimmed.

Impact:
`output_type="latent"` returns a latent video with extra frames. Callers expecting returned latents to correspond to the requested video length get an off-by-reference count shape.

Reproduction:
```python
import torch
from PIL import Image
from transformers import AutoConfig, AutoTokenizer, T5EncoderModel
from diffusers import AutoencoderKLWan, FlowMatchEulerDiscreteScheduler, WanVACEPipeline, WanVACETransformer3DModel

vae = AutoencoderKLWan(base_dim=3, z_dim=16, dim_mult=[1, 1, 1, 1], num_res_blocks=1, temperal_downsample=[False, True, True])
scheduler = FlowMatchEulerDiscreteScheduler(shift=7.0)
config = AutoConfig.from_pretrained("hf-internal-testing/tiny-random-t5")
text_encoder = T5EncoderModel(config)
tokenizer = AutoTokenizer.from_pretrained("hf-internal-testing/tiny-random-t5")
transformer = WanVACETransformer3DModel(
    patch_size=(1, 2, 2), num_attention_heads=2, attention_head_dim=12,
    in_channels=16, out_channels=16, text_dim=32, freq_dim=256,
    ffn_dim=32, num_layers=3, cross_attn_norm=True,
    qk_norm="rms_norm_across_heads", rope_max_seq_len=32,
    vace_layers=[0, 2], vace_in_channels=96,
)
pipe = WanVACEPipeline(tokenizer, text_encoder, vae, scheduler, transformer=transformer).to("cpu")

num_frames = 17
video = [Image.new("RGB", (16, 16))] * num_frames
mask = [Image.new("L", (16, 16), 0)] * num_frames

latents = pipe(
    video=video,
    mask=mask,
    reference_images=Image.new("RGB", (16, 16)),
    prompt="test prompt",
    negative_prompt="negative",
    num_inference_steps=1,
    guidance_scale=1.0,
    height=16,
    width=16,
    num_frames=num_frames,
    max_sequence_length=16,
    output_type="latent",
    generator=torch.Generator(device="cpu").manual_seed(0),
).frames

print(tuple(latents.shape))
print("expected latent frames:", (num_frames - 1) // vae.config.scale_factor_temporal + 1)
```

Relevant precedent:
The same pipeline already trims reference latents before VAE decode.

Suggested fix:
```python
if num_reference_images:
    latents = latents[:, :, num_reference_images:]

if output_type != "latent":
    latents = latents.to(vae_dtype)
    video = self.vae.decode(latents, return_dict=False)[0]
    video = self.video_processor.postprocess_video(video, output_type=output_type)
else:
    video = latents
```

## Issue 4: VACE, video-to-video, and modular Wan ignore VAE scale-factor config

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_vace.py#L199-L200
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_video2video.py#L217-L218
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/modular_pipelines/wan/modular_pipeline.py#L69-L80

Problem:
These paths derive scale factors from `len(vae.temperal_downsample)` and hard-code spatial scale as `8`, instead of using `vae.config.scale_factor_temporal` and `vae.config.scale_factor_spatial`.

Impact:
Custom or future Wan VAE configs serialize valid scale factors but the affected pipelines preprocess sizes and frame counts incorrectly.

Reproduction:
```python
from diffusers import AutoencoderKLWan, WanVACEPipeline, WanVideoToVideoPipeline

vae = AutoencoderKLWan(
    base_dim=3,
    z_dim=16,
    dim_mult=[1, 1, 1, 1],
    num_res_blocks=1,
    temperal_downsample=[False, True, True],
    scale_factor_spatial=4,
)

v2v = WanVideoToVideoPipeline(tokenizer=None, text_encoder=None, transformer=None, vae=vae, scheduler=None)
vace = WanVACEPipeline(tokenizer=None, text_encoder=None, transformer=None, transformer_2=None, vae=vae, scheduler=None)

print("vae config:", vae.config.scale_factor_spatial)
print("video2video:", v2v.vae_scale_factor_spatial)
print("vace:", vace.vae_scale_factor_spatial)
```

Relevant precedent:
`WanPipeline`, `WanImageToVideoPipeline`, and `WanAnimatePipeline` use the VAE config scale factors.

Suggested fix:
```python
self.vae_scale_factor_temporal = self.vae.config.scale_factor_temporal if getattr(self, "vae", None) else 4
self.vae_scale_factor_spatial = self.vae.config.scale_factor_spatial if getattr(self, "vae", None) else 8
```

Apply the same config-based source in the modular Wan pipeline state initialization.

## Issue 5: `WanAnimateImageProcessor` drops constructor config values

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/image_processor.py#L57-L71
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/wan/pipeline_wan_animate.py#L227-L233

Problem:
`WanAnimateImageProcessor.__init__` accepts `vae_scale_factor`, `vae_latent_channels`, and `resample`, but then calls `super().__init__()` without forwarding them. The base class re-registers default values and overwrites the subclass config.

Impact:
The animate pipeline constructs its processor with Wan-specific values, but the stored config reports base defaults. This can affect serialization, reload behavior, and downstream code that reads processor config.

Reproduction:
```python
from diffusers.pipelines.wan.image_processor import WanAnimateImageProcessor

processor = WanAnimateImageProcessor(
    vae_scale_factor=16,
    vae_latent_channels=16,
    spatial_patch_size=(3, 5),
    resample="bilinear",
    fill_color=123,
)

print(processor.config.vae_scale_factor)
print(processor.config.vae_latent_channels)
print(processor.config.resample)
print(processor.config.spatial_patch_size)
```

Relevant precedent:
`VaeImageProcessor.__init__` accepts these fields directly and should be initialized with the subclass values.

Suggested fix:
```python
super().__init__(
    do_resize=do_resize,
    vae_scale_factor=vae_scale_factor,
    vae_latent_channels=vae_latent_channels,
    resample=resample,
    reducing_gap=reducing_gap,
    do_normalize=do_normalize,
    do_binarize=do_binarize,
    do_convert_rgb=do_convert_rgb,
    do_convert_grayscale=do_convert_grayscale,
)
self.register_to_config(spatial_patch_size=spatial_patch_size, fill_color=fill_color)
```

## Issue 6: Modular Wan casts scheduler timesteps to model dtype

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/modular_pipelines/wan/before_denoise.py#L252-L265
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/modular_pipelines/wan/denoise.py#L205-L211
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/modular_pipelines/wan/denoise.py#L340-L346

Problem:
`WanTextInputStep` derives `block_state.dtype` from `prompt_embeds.dtype`, despite the comment saying transformer dtype should be used. The denoise blocks then cast scheduler timesteps to that dtype. With bf16/fp16 model paths, timestep values can be rounded before reaching the model.

Impact:
The modular pipeline can diverge from standard Wan pipeline behavior and lose scheduler precision.

Reproduction:
```python
import torch

for value in [999.0, 998.0, 995.0, 750.3]:
    timestep = torch.tensor([value], dtype=torch.float32)
    print(value, "->", timestep.to(torch.bfloat16).to(torch.float32).item())
```

Relevant precedent:
The standard Wan pipelines cast prompt embeddings and latents to transformer dtype, but pass the scheduler timestep tensor without pre-casting it to bf16/fp16.

Suggested fix:
```python
# before denoise
block_state.dtype = components.transformer.dtype
block_state.prompt_embeds = block_state.prompt_embeds.to(block_state.dtype)
if block_state.negative_prompt_embeds is not None:
    block_state.negative_prompt_embeds = block_state.negative_prompt_embeds.to(block_state.dtype)

# denoise
timestep = t.expand(block_state.latent_model_input.shape[0])
```

For Wan 2.2 modular denoising, choose the latent/model dtype after selecting the current transformer, but keep scheduler timestep precision intact.

## Issue 7: Slow test coverage is missing for several public Wan workflows

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/tests/pipelines/wan/test_wan.py#L185-L201
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/tests/pipelines/wan/test_wan_animate.py#L224-L240
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/tests/pipelines/wan/test_wan_image_to_video.py#L40-L173
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/tests/pipelines/wan/test_wan_vace.py#L39-L285
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/tests/pipelines/wan/test_wan_video_to_video.py#L35-L149
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/tests/modular_pipelines/wan/test_modular_pipeline_wan.py#L23-L48

Problem:
`WanPipeline` and `WanAnimatePipeline` have slow integration coverage. I did not find slow tests for Wan image-to-video, Wan 2.2 text-to-video, Wan 2.2 image-to-video, VACE, video-to-video, or modular Wan parity.

Impact:
Large public workflows can regress without any real-checkpoint coverage. The issues above are examples that fast synthetic tests did not catch.

Reproduction:
```python
from pathlib import Path

files = [
    "tests/pipelines/wan/test_wan.py",
    "tests/pipelines/wan/test_wan_22.py",
    "tests/pipelines/wan/test_wan_image_to_video.py",
    "tests/pipelines/wan/test_wan_22_image_to_video.py",
    "tests/pipelines/wan/test_wan_vace.py",
    "tests/pipelines/wan/test_wan_video_to_video.py",
    "tests/pipelines/wan/test_wan_animate.py",
    "tests/modular_pipelines/wan/test_modular_pipeline_wan.py",
]

for file in files:
    text = Path(file).read_text(encoding="utf-8")
    print(file, "@slow" in text)
```

Relevant precedent:
The existing slow tests for `WanPipeline` and `WanAnimatePipeline` provide the pattern.

Suggested fix:
Add at least one `@slow` integration test for each missing public workflow, using fixed seeds and slice assertions. Include VACE coverage with `reference_images` and `output_type="latent"`, and modular-vs-standard parity coverage for a representative Wan pipeline.

hlky:
Fixes #13578.

## What does this PR do?

This PR addresses the Wan systemic review findings:

- Expands Wan i2v and Animate image embeddings, masks, latents, and conditioning tensors to `batch_size * num_videos_per_prompt` in per-prompt order.
- Keeps `image` required where VAE conditioning still needs it, while allowing precomputed CLIP `image_embeds`.
- Accepts documented Wan i2v image list/tuple inputs and validates precomputed image-embed batch sizes.
- Uses Wan VAE config scale factors in VACE, video-to-video, and modular Wan paths.
- Trims VACE reference latents for `output_type="latent"`.
- Preserves Wan Animate image processor config values.
- Stops silently ignoring unsupported Wan video-to-video `num_videos_per_prompt != 1`.
- Keeps modular Wan timesteps in scheduler precision while casting model inputs to the selected transformer dtype.
- Adds focused fast regression tests and a `.ai/pipelines.md` checklist note for `num_*_per_prompt` expansion drift.

## Tests

- `ruff check ...` on touched Python files
- `ruff format --check ...` on touched Python files
- `git diff --check`
- `python -m compileall -q src/diffusers/pipelines/wan src/diffusers/modular_pipelines/wan tests/pipelines/wan tests/modular_pipelines/wan`
- `python -m pytest tests/pipelines/wan/test_wan_image_to_video.py tests/pipelines/wan/test_wan_vace.py tests/pipelines/wan/test_wan_video_to_video.py tests/pipelines/wan/test_wan_animate.py tests/modular_pipelines/wan/test_modular_pipeline_wan.py -q -k "not test_save_load_float16"`

Result: `180 passed, 21 skipped, 5 deselected, 61 warnings, 6 subtests passed`.

Note: `WanFLFToVideoPipelineFastTests::test_save_load_float16` fails the same way on clean `upstream/main` (`max diff 0.04175 > 0.01`), so it was excluded from the clean touched-file verification.


src/diffusers/modular_pipelines/wan/before_denoise.py
```diff
@@ -252,22 +252,16 @@ def __call__(self, components: WanModularPipeline, state: PipelineState) -> Pipe
...
-            _, seq_len, _ = block_state.negative_prompt_embeds.shape
-            block_state.negative_prompt_embeds = block_state.negative_prompt_embeds.repeat(
-                1, block_state.num_videos_per_prompt, 1
-            )
-            block_state.negative_prompt_embeds = block_state.negative_prompt_embeds.view(
-                block_state.batch_size * block_state.num_videos_per_prompt, seq_len, -1
-            )
+            block_state.negative_prompt_embeds = block_state.negative_prompt_embeds.repeat_interleave(
+                block_state.num_videos_per_prompt, dim=0
+            ).to(block_state.dtype)
```
yiyixuxu: I think we already cast dtype later in denoise step no?

src/diffusers/modular_pipelines/wan/denoise.py
```diff
@@ -165,6 +171,18 @@ def inputs(self) -> list[tuple[str, Any]]:
                 type_hint=int,
                 description="The number of inference steps to use for the denoising process. Can be generated in set_timesteps step.",
             ),
+            InputParam(
```
yiyixuxu: this is created in a previous loop step, I think we don't need to declare it

src/diffusers/modular_pipelines/wan/denoise.py
```diff
@@ -54,17 +54,21 @@ def inputs(self) -> list[InputParam]:
...
-            InputParam(
-                "dtype",
-                required=True,
-                type_hint=torch.dtype,
-                description="The dtype of the model inputs. Can be generated in input step.",
-            ),
+        ]
+
+    @property
+    def intermediate_outputs(self) -> list[OutputParam]:
```
yiyixuxu: I think you don't need to declare the output here https://huggingface.co/docs/diffusers/modular_diffusers/loop_sequential_pipeline_blocks#loop-blocks

src/diffusers/modular_pipelines/wan/denoise.py
```diff
@@ -94,19 +98,21 @@ def inputs(self) -> list[InputParam]:
...
```
yiyixuxu: I think we don't need to declare it here > this does not go beyond the loop step 
https://huggingface.co/docs/diffusers/modular_diffusers/loop_sequential_pipeline_blocks#loop-blocks

src/diffusers/modular_pipelines/wan/denoise.py
```diff
@@ -198,16 +227,16 @@ def __call__(
...
                 if k in self._guider_input_fields.keys()
             }
 
             # Predict the noise residual
             # store the noise_pred in guider_state_batch so that we can apply guidance across all batches
             guider_state_batch.noise_pred = components.transformer(
-                hidden_states=block_state.latent_model_input.to(block_state.dtype),
-                timestep=t.expand(block_state.latent_model_input.shape[0]).to(block_state.dtype),
+                hidden_states=block_state.latent_model_input.to(dtype),
+                timestep=t.expand(block_state.latent_model_input.shape[0]),
```
yiyixuxu, line 203: the `dtype` update here makes sense,

>Keeps modular Wan timesteps in scheduler precision while casting model inputs to the selected transformer dtype.

I also liked we use `transformer.dtype` instead of `prompt_embeds.dtype` -> should be the stanrdard preferred way of infer dtype moving forward 

But most of the rest of the change on inputs/outputs does not seem valid to me, loop step works a bit differently https://huggingface.co/docs/diffusers/modular_diffusers/loop_sequential_pipeline_blocks#loop-blocks; let me know if I missed anything

src/diffusers/pipelines/wan/pipeline_wan_animate.py
```diff
@@ -485,6 +480,30 @@ def get_i2v_mask(
 
         return mask_lat_size
 
+    def _expand_tensor_to_effective_batch(
```
yiyixuxu: can we make it a regular function?

src/diffusers/pipelines/wan/pipeline_wan_i2v.py
```diff
@@ -342,17 +406,16 @@ def check_inputs(
         callback_on_step_end_tensor_inputs=None,
         guidance_scale_2=None,
     ):
-        if image is not None and image_embeds is not None:
-            raise ValueError(
-                f"Cannot forward both `image`: {image} and `image_embeds`: {image_embeds}. Please make sure to"
-                " only forward one of the two."
-            )
-        if image is None and image_embeds is None:
+        if image is None:
```
yiyixuxu, line 411: i don't think the list of image make sense here for I2V pipeline, no? 
Maybe we should just fix the docstring instead?

yiyixuxu (commented):
I left some comments

src/diffusers/modular_pipelines/wan/before_denoise.py
hlky: https://github.com/hlky/diffusers/blob/7f2922d373e9f5abc0404fe4ea2146c588557bd7/src/diffusers/modular_pipelines/wan/denoise.py#L225-L233

Yes I think it would be cast here.

.ai/pipelines.md
```diff
@@ -60,3 +60,5 @@ When adding a new pipeline (or reviewing one), skim `pipeline_flux.py`, `pipelin
...
+
+6. **Partial batch expansion with `num_*_per_prompt`.** When a pipeline accepts `num_images_per_prompt`, `num_videos_per_prompt`, or precomputed conditioning tensors, every per-prompt input that reaches the denoising loop must be expanded to the same effective batch size in the same prompt order. Check prompt embeds, pooled embeds, image/video embeds, masks, control latents, image latents, guidance tensors, and any added conditioning. Prefer `torch.repeat_interleave(tensor, repeats=num_per_prompt, dim=0, output_size=tensor.shape[0] * num_per_prompt)` for contiguous per-prompt grouping, and validate precomputed tensors that are already batched instead of letting a later concat fail. Avoid `tensor.repeat((num_per_prompt, ...))` for batch expansion because it interleaves prompts (`[A, B] -> [A, B, A, B]`) rather than grouping them (`[A, A, B, B]`).
```
sayakpaul, line 64: > every per-prompt input that reaches the denoising loop must be expanded to the same effective batch size in the same prompt order.

Should we try to include an example here?

> Check prompt embeds, pooled embeds, image/video embeds, masks, control latents, image latents, guidance tensors, and any added conditioning.

Should we be specific in terms of what we want it to check?

src/diffusers/pipelines/wan/image_processor.py
```diff
@@ -68,13 +66,18 @@ def __init__(
         do_convert_grayscale: bool = False,
         fill_color: str | float | tuple[float, ...] | None = 0,
     ):
-        super().__init__()
-        if do_convert_rgb and do_convert_grayscale:
-            raise ValueError(
-                "`do_convert_rgb` and `do_convert_grayscale` can not both be set to `True`,"
-                " if you intended to convert the image into RGB format, please set `do_convert_grayscale = False`.",
-                " if you intended to convert the image into grayscale format, please set `do_convert_rgb = False`",
-            )
```
sayakpaul, line 77: Should we still keep the `ValueError`, though?

src/diffusers/pipelines/wan/pipeline_wan_animate.py
```diff
@@ -384,17 +384,12 @@ def check_inputs(
...
     ):
-        if image is not None and image_embeds is not None:
-            raise ValueError(
-                f"Cannot forward both `image`: {image} and `image_embeds`: {image_embeds}. Please make sure to"
-                " only forward one of the two."
-            )
-        if image is None and image_embeds is None:
-            raise ValueError(
-                "Provide either `image` or `prompt_embeds`. Cannot leave both `image` and `image_embeds` undefined."
-            )
```
sayakpaul, line 395: Why do we want to remove this check?

Update: nevermind we need `image` for VAE encodings. In that case, should we tell the users in case `image is None and image_embeds is not None`?

src/diffusers/pipelines/wan/pipeline_wan_animate.py
```diff
@@ -776,13 +795,13 @@ def __call__(
...
         guidance_scale: float = 1.0,
-        num_videos_per_prompt: int | None = 1,
+        num_videos_per_prompt: int = 1,
         generator: torch.Generator | list[torch.Generator] | None = None,
         latents: torch.Tensor | None = None,
         prompt_embeds: torch.Tensor | None = None,
         negative_prompt_embeds: torch.Tensor | None = None,
         image_embeds: torch.Tensor | None = None,
-        output_type: str | None = "np",
+        output_type: str = "np",
```
sayakpaul: (Optional): while we're at it, we could also do a `Literal` and help make it easier for users to understand what values are allowed here.

src/diffusers/pipelines/wan/pipeline_wan_i2v.py
```diff
@@ -329,6 +328,71 @@ def encode_prompt(
 
         return prompt_embeds, negative_prompt_embeds
 
+    def _expand_tensor_to_effective_batch(
+        self,
+        tensor: torch.Tensor,
+        batch_size: int,
+        num_videos_per_prompt: int,
+        tensor_name: str,
```
sayakpaul: I think `tensor_name` is a nice-to-have rather than must-have. We could make it optional or completely remove it.

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -161,6 +161,125 @@ def test_inference(self):
         generated_slice = torch.cat([generated_slice[:8], generated_slice[-8:]])
         self.assertTrue(torch.allclose(generated_slice, expected_slice, atol=1e-3))
 
+    def test_num_videos_per_prompt_with_image_embeds(self):
+        device = "cpu"
```
sayakpaul: Let's use `torch_device` as we're not doing any value-based assertions here.

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -161,6 +161,125 @@ def test_inference(self):
...
+                "output_type": "latent",
+            }
+        )
+        with torch.no_grad():
+            image_embeds = pipe.encode_image(inputs["image"], device)
+        inputs["image_embeds"] = torch.cat([image_embeds, image_embeds + 1.0], dim=0)
+
+        video = pipe(**inputs).frames
+
+        self.assertEqual(video.shape, (6, 16, 1, 2, 2))
```
sayakpaul: Let's derive `6` as the effective batch size programmatically.

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -161,6 +161,125 @@ def test_inference(self):
...
+                "guidance_scale": 1.0,
+                "num_videos_per_prompt": 3,
+                "num_inference_steps": 1,
+                "num_frames": 1,
+                "output_type": "latent",
+            }
+        )
+        with torch.no_grad():
+            image_embeds = pipe.encode_image(inputs["image"], device)
+        inputs["image_embeds"] = torch.cat([image_embeds, image_embeds + 1.0], dim=0)
```
sayakpaul: Should this not be done within the pipeline itself? If so, I would just do `video = pipe(**inputs).frames` and test `image_embeds` separately.

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -161,6 +161,125 @@ def test_inference(self):
...
+
+        video = pipe(**inputs).frames
+
+        self.assertEqual(video.shape, (6, 16, 1, 2, 2))
+
+    def test_image_embeds_invalid_batch_size_raises(self):
+        device = "cpu"
```
sayakpaul: Same as above.

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -161,6 +161,125 @@ def test_inference(self):
...
+        )
+
+        video = pipe(**inputs).frames
+
+        self.assertEqual(video.shape, (2, 16, 1, 2, 2))
+
+    def test_last_image_embeds_expand_per_prompt_order(self):
+        components = self.get_dummy_components()
+        pipe = self.pipeline_class(**components)
+        pipe.transformer.condition_embedder.image_embedder.pos_embed = torch.nn.Parameter(torch.zeros(1, 2, 4))
```
sayakpaul, line 276: Why is this needed?

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -161,6 +161,125 @@ def test_inference(self):
...
+
+        image_embeds = torch.tensor([0.0, 1.0, 10.0, 11.0]).reshape(4, 1, 1)
```
sayakpaul: Let's give `4`, `1`, and `1` variable names for easier parsing of info.

tests/pipelines/wan/test_wan_image_to_video.py
```diff
@@ -161,6 +161,125 @@ def test_inference(self):
...
+            device=torch.device("cpu"),
+            batch_size=2,
+            num_videos_per_prompt=2,
+            transformer_dtype=torch.float32,
+            image_embeds=image_embeds,
+            last_image=[Image.new("RGB", (16, 16)), Image.new("RGB", (16, 16))],
+        )
+
+        expected = torch.tensor([0.0, 1.0, 0.0, 1.0, 10.0, 11.0, 10.0, 11.0])
+        torch.testing.assert_close(image_embeds[:, 0, 0], expected)
```
sayakpaul, line 320: It would be helpful have to have some comments around how this is the expected order.

sayakpaul (commented):
Thanks! I reviewed the bits around batch expansion. LMK if the comments make sense.

.ai/pipelines.md
hlky: Agree this should be expanded, adjustments to `.ai/` docs can be done in a separate PR if you prefer.

src/diffusers/modular_pipelines/wan/denoise.py
hlky: Thanks, will check this. I think any modular changes will need more in-depth review - it's a new/experimental feature so agents can get it wrong.

src/diffusers/modular_pipelines/wan/denoise.py
hlky: Agree `transformer` dtype should be the target - all inputs go to `transformer`/`unet` in the end so it makes sense.

src/diffusers/pipelines/wan/image_processor.py
hlky: Same ValueError is in `VaeImageProcessor` so it should be covered via `super().__init__(...)`.

src/diffusers/pipelines/wan/pipeline_wan_animate.py
hlky: If `image is None and image_embeds is not None` it would be covered by `if image is None` check.

src/diffusers/pipelines/wan/pipeline_wan_animate.py
hlky: Yes, we can also just inline it for now and standardize all repeat variants at a later point.

src/diffusers/pipelines/wan/pipeline_wan_animate.py
hlky: Agree, standardizing this across pipelines would be a good move.

tests/pipelines/wan/test_wan_image_to_video.py
hlky: The tiny WanTransformer3DModel config doesn't include `pos_embed_seq_len` so `WanImageEmbedding` ends up without `pos_embed`. Changing the actual config could change expected slices for other tests.

hlky (commented):
Thanks for the review @yiyixuxu @sayakpaul. Will address the changes soon.

PR #2643, opened 2023-03-11
[MPS] CrossAttentionProcessor: (partial) fix for NaN attention_scores on PyTorch nightly

Birch-san:
when `attention_mask is None`, we set [`baddbmm`](https://pytorch.org/docs/stable/generated/torch.baddbmm.html#torch.baddbmm)'s `beta = 0`.

> If `beta` is 0, then `input` will be ignored, and _nan_ and _inf_ in it will not be propagated.

in other words, `baddbmm_input` is an unused tensor that we supply just because the API requires it. `beta = 0`, so the tensor isn't even used. this is why when I originally introduced it: I knew there was no need to even initialize the memory (hence `empty` instead of `zero`).

on MPS backend: this has been fine in some PyTorch versions, but on the recent nightly I tried (2.1.0.dev20230310): allocating such a large tensor (`40, 4096, 4096`) causes `baddbmm()` to return NaN attention_scores (and consequently black images).

so, I've changed it to do a smaller allocation, and expand the tensor to the expected size.  
at this allocation size (one element): I'm unconcerned about whether initializing the memory costs anything, so I've changed it to use zeroes instead of uninitialized memory.

a (`1, 1, 1`) tensor would work too (`baddbmm` supports broadcasting), but I think the explicit shape is helpful documentation, and doesn't cost much with `expand()`. it might also help model conversion (such as coremltools) with expressing the baddbmm in terms of unfused operations (GEMM * scale + bias).

=====

the reason I call this a _partial_ fix for black images (in this PyTorch nightly, on MPS) is because I found one more fix was necessary:

`CrossAttention#get_attention_scores`:  
```diff
  attention_scores = torch.baddbmm(
      attention_bias,
      query,
      key.transpose(-1, -2),
      beta=beta,
      alpha=self.scale,
  )
+ # accessing the tensor seems to help prevent black images on MPS on PyTorch nightly
+ assert not attention_scores.isnan().any().item()
```

`AttnProcessor2_0#__call__`:  
```diff
  # the output of sdp = (batch, num_heads, seq_len, head_dim)
  hidden_states = F.scaled_dot_product_attention(
      query, key, value, attn_mask=attention_mask, dropout_p=0.0, is_causal=False
  )
+ # accessing the tensor seems to help prevent black images on MPS on PyTorch nightly
+ assert not hidden_states.isnan().any().item()
```

[...]

PR #13701, opened 2026-05-08
[mps] Fix NaN in Attention.get_attention_scores when attention_mask is None

### LINKED ISSUE #11229: enable_attention_slicing give NaN results for SDXL on MPS
### Describe the bug

If I call pipe.enable_attention_slicing I get NaN's returned when output type is 'latent' and a value error for image output.

The error is....
```py
/Volumes/SSD2TB/AI/Diffusers/lib/python3.11/site-packages/diffusers/image_processor.py:147: RuntimeWarning: invalid value encountered in cast
  images = (images * 255).round().astype("uint8")
```

printing the latent gives

```py
$ python tas.py 
Loading pipeline components...: 100%|█████████████████████████████████████████████████████| 6/6 [00:00<00:00,  8.89it/s]
100%|█████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:10<00:00,  5.26s/it]
tensor([[[[nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          ...,
...
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan]]]], device='mps:0',
       dtype=torch.float16)
```

commenting out the call to enable_attention_slicing gives non NaN latest and a proper image.

I've tested this on the current release version and the current HEAD version

### Reproduction

```py
import torch

model_id = "stabilityai/stable-diffusion-xl-base-1.0"

pipe = DiffusionPipeline.from_pretrained(model_id, variant="fp16",
                                         torch_dtype=torch.float16)


pipe.to(device="mps", torch_dtype=torch.float16)

pipe.enable_attention_slicing()


prompt = "analog film photo Butterflies in a jungle, cold color palette, vivid colors, detailed, 8k, 35mm photo, Kodachrome, Lomography, highly detailed"
negative_prompt = "painting, drawing, illustration, glitch, deformed, mutated, cross-eyed, ugly, disfigured"


images = pipe(
      prompt=prompt,
      num_inference_steps=2,
      guidance_scale=7,
      output_type='latent'
  ).images

print(images)
```

comment out or delete `pipe.enable_attention_slicing()` and the script works as expected. 

### Logs

```shell
(Diffusers) $ python tas.py 
Loading pipeline components...: 100%|█████████████████████████████████████████████████████| 7/7 [00:00<00:00, 10.57it/s]
100%|█████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:10<00:00,  5.34s/it]
tensor([[[[nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          ...,
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan]],

         [[nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          ...,
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan]],

         [[nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          ...,
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan]],

         [[nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          ...,
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan],
          [nan, nan, nan,  ..., nan, nan, nan]]]], device='mps:0',
       dtype=torch.float16)
(Diffusers) $
```

### System Info

- 🤗 Diffusers version: 0.33.0.dev0
- Platform: macOS-15.3.2-arm64-arm-64bit
- Python version: 3.11.10
- PyTorch version (GPU?): 2.6.0 (False)
- Jax version: not installed
- Huggingface_hub version: 0.27.1
- Transformers version: 4.50.3
- Accelerate version: 0.34.2
- PEFT version: not installed
- Safetensors version: 0.4.5
- Accelerator: Apple M3
- Using GPU in script?: Yes

### Who can help?

 @pcuenca 

I'm assuming its more MPS related than SDXL.

Shreyas-jk:
# What does this PR do?

Fixes #11229

`Attention.get_attention_scores` allocates `baddbmm_input` with `torch.empty()` and uses `beta=0`, relying on `baddbmm` to ignore the uninitialized input. The MPS `baddbmm` kernel does not short-circuit on `beta=0`, so any NaN/Inf in the uninitialized memory propagates through `0 * NaN = NaN` and poisons the attention output. CUDA happens to mask this because its allocator typically returns zero-initialized memory.

This change uses `torch.zeros` instead of `torch.empty` **only on MPS**, leaving the CUDA / CPU / XPU paths unchanged so they don't pay the extra fill cost.

In real workloads this surfaces as black/NaN images from `StableDiffusionXLPipeline` with `enable_attention_slicing()` on Apple Silicon + fp16, which is the standard memory-saving path on Macs with limited unified memory.

## Reproduction

Minimal repro on M-series MPS (without the fix): 30/30 trials produce NaN. With the fix: 0/30. CPU baseline: 0/5. Verified on M4 MacBook Pro, torch 2.11.0, fp16 and fp32.

The added test `GetAttentionScoresMPSTests.test_no_nan_when_attention_mask_is_none_on_mps` reproduces the bug deterministically (fails 20/20 without the fix, passes 20/20 with it) by polluting the MPS allocator pool with NaN-filled tensors before each call.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue or the forum?  --> Fixes existing issue #11229.

## Who can review?

@pcuenca (MPS / Apple Silicon maintainer per the PR template)

### DIFF
--- a/src/diffusers/models/attention_processor.py
+++ b/src/diffusers/models/attention_processor.py
@@ -676,9 +676,10 @@ class Attention(nn.Module):
             key = key.float()
 
         if attention_mask is None:
-            if query.device.type == "mps":
-                # MPS' baddbmm does not short-circuit on beta=0, so an
-                # uninitialized input from torch.empty() can propagate NaN.
+            if query.device.type == "mps" and is_torch_version("<", "2.14.0"):
+                # Before torch 2.14 (pytorch#187522), MPS' baddbmm did not
+                # short-circuit on beta=0, so an uninitialized input from
+                # torch.empty() could propagate NaN. Fixed upstream from 2.14.
                 baddbmm_input = torch.zeros(
                     query.shape[0], query.shape[1], key.shape[1], dtype=query.dtype, device=query.device
                 )
--- a/src/diffusers/models/attention_processor.py
+++ b/src/diffusers/models/attention_processor.py
@@ -676,9 +676,16 @@ class Attention(nn.Module):
             key = key.float()
 
         if attention_mask is None:
-            baddbmm_input = torch.empty(
-                query.shape[0], query.shape[1], key.shape[1], dtype=query.dtype, device=query.device
-            )
+            if query.device.type == "mps":
+                # MPS' baddbmm does not short-circuit on beta=0, so an
+                # uninitialized input from torch.empty() can propagate NaN.
+                baddbmm_input = torch.zeros(
+                    query.shape[0], query.shape[1], key.shape[1], dtype=query.dtype, device=query.device
+                )
+            else:
+                baddbmm_input = torch.empty(
+                    query.shape[0], query.shape[1], key.shape[1], dtype=query.dtype, device=query.device
+                )
             beta = 0
         else:
             baddbmm_input = attention_mask
--- a/tests/models/test_attention_processor.py
+++ b/tests/models/test_attention_processor.py
@@ -10,7 +10,7 @@ from packaging import version
 from diffusers import DiffusionPipeline
 from diffusers.models.attention_processor import Attention, AttnAddedKVProcessor
 
-from ..testing_utils import torch_device
+from ..testing_utils import is_torch_version, torch_device
 
 
 class AttnAddedKVProcessorTests(unittest.TestCase):
@@ -135,30 +135,33 @@ class DeprecatedAttentionBlockTests(unittest.TestCase):
         self.assertTrue(np.allclose(conversion, after_conversion, atol=1e-3))
 
 
-class GetAttentionScoresMPSTests(unittest.TestCase):
-    @pytest.mark.skipif(torch_device != "mps", reason="test exercises an MPS-specific code path")
-    def test_no_nan_when_attention_mask_is_none_on_mps(self):
-        # Regression test: torch.empty() on MPS can return non-finite values,
-        # and MPS' baddbmm does not short-circuit on beta=0, so an unmasked
-        # call to get_attention_scores used to propagate NaN into the output.
-        torch.manual_seed(0)
-        heads, dim_head, seq_len = 4, 32, 256
-        attn = Attention(
-            query_dim=heads * dim_head,
-            heads=heads,
-            dim_head=dim_head,
-            bias=False,
-        ).to(torch_device, torch.float16)
+@pytest.mark.skipif(torch_device != "mps", reason="test exercises an MPS-specific code path")
+@pytest.mark.skipif(
+    is_torch_version(">=", "2.14.0"),
+    reason="baddbmm beta=0 NaN fixed upstream in pytorch#187522 (torch>=2.14); MPS workaround no longer applied",
+)
+def test_no_nan_when_attention_mask_is_none_on_mps():
+    # Regression test: torch.empty() on MPS can return non-finite values,
+    # and MPS' baddbmm does not short-circuit on beta=0, so an unmasked
+    # call to get_attention_scores used to propagate NaN into the output.
+    torch.manual_seed(0)
+    heads, dim_head, seq_len = 4, 32, 256
+    attn = Attention(
+        query_dim=heads * dim_head,
+        heads=heads,
+        dim_head=dim_head,
+        bias=False,
+    ).to(torch_device, torch.float16)
 
-        for _ in range(20):
-            # Pollute the MPS allocator pool with non-finite values so that a
-            # subsequent torch.empty() is likely to return NaN-filled memory.
-            polluter = torch.full((heads, seq_len, seq_len), float("nan"), device=torch_device, dtype=torch.float16)
-            del polluter
+    for _ in range(20):
+        # Pollute the MPS allocator pool with non-finite values so that a
+        # subsequent torch.empty() is likely to return NaN-filled memory.
+        polluter = torch.full((heads, seq_len, seq_len), float("nan"), device=torch_device, dtype=torch.float16)
+        del polluter
 
-            query = torch.randn(1, seq_len, heads * dim_head, device=torch_device, dtype=torch.float16)
-            key = torch.randn(1, seq_len, heads * dim_head, device=torch_device, dtype=torch.float16)
-            scores = attn.get_attention_scores(
-                attn.head_to_batch_dim(query), attn.head_to_batch_dim(key), attention_mask=None
-            )
-            self.assertFalse(torch.isnan(scores).any().item(), "attention scores contain NaN on MPS")
+        query = torch.randn(1, seq_len, heads * dim_head, device=torch_device, dtype=torch.float16)
+        key = torch.randn(1, seq_len, heads * dim_head, device=torch_device, dtype=torch.float16)
+        scores = attn.get_attention_scores(
+            attn.head_to_batch_dim(query), attn.head_to_batch_dim(key), attention_mask=None
+        )
+        assert not torch.isnan(scores).any().item(), "attention scores contain NaN on MPS"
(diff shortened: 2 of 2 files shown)


Shreyas-jk:
Ping gently, this is open since May 8. Tiny MPS-based fix for the baddbmm beta=0 NaN in get_attention_scores (similar to #2643), including a regression test that does not pass without the fix, and passes with it. This is simply rebased onto main as-is, and is ready to go. Could a maintainer trigger CI or review when time permits? @yiyixuxu  @sayakpaul

pcuenca (approved):
looks good to me, will defer to the team for additional opinions

pcuenca:
also cc @Isalia20 for info

Isalia20:
Thanks for tagging @pcuenca It is already fixed on nightly https://github.com/pytorch/pytorch/pull/187522

src/diffusers/models/attention_processor.py
```diff
@@ -676,9 +676,16 @@ def get_attention_scores(
...
+            if query.device.type == "mps":
+                # MPS' baddbmm does not short-circuit on beta=0, so an
+                # uninitialized input from torch.empty() can propagate NaN.
+                baddbmm_input = torch.zeros(
+                    query.shape[0], query.shape[1], key.shape[1], dtype=query.dtype, device=query.device
+                )
+            else:
+                baddbmm_input = torch.empty(
+                    query.shape[0], query.shape[1], key.shape[1], dtype=query.dtype, device=query.device
+                )
```
sayakpaul: Should this be gated on a future PyTorch version?
https://github.com/huggingface/diffusers/pull/13701#issuecomment-4808403576

Because when the nightly is eventually in a stable version, then we shouldn't have to do `zeros()` right?

tests/models/test_attention_processor.py
```diff
@@ -133,3 +133,32 @@ def test_conversion_when_using_device_map(self):
 
         self.assertTrue(np.allclose(pre_conversion, conversion, atol=1e-3))
         self.assertTrue(np.allclose(conversion, after_conversion, atol=1e-3))
+
+
+class GetAttentionScoresMPSTests(unittest.TestCase):
```
sayakpaul: 1. Should be implemented using `pytest`.
2. Should have a note or some `xfail` decorator indicating that we won't need it after the nightly is in a stable.

Shreyas-jk: Thank you @sayakpaul! I agree in principle, but since the MPS branch in question allocates and zeros out a full (batch, q_len, k_len) tensor each time, once pytorch#187522 is included in a stable release then it will be clear that using an empty() tensor is better.

Now the issue with this approach is that the fix is in nightly right now (@Isalia20), meaning there is no stable release to guard against yet; users with the current stable PyTorch will trigger the NaN issue and so the guard is necessary. There are two possible ways to do this:

1. Guard the MPS branch on `is_torch_version("<", "<target>")`, so that when the fix is released it stops doing anything. Do you know which release pytorch#187522 is targeted to? If it's tagged, then I'll gate on that specific version.
2. If there is not yet a target version for the release, leave the guard unconditional for now and add a TODO reference to pytorch#187522, and add the guard when we know the release version.

In either case I'll move the test over to pytest. As for the idea about making the test an xfail, it seems that with the fix the test passes, so it self-deactivates.

Does that match what you had in mind?

sayakpaul: > Guard the MPS branch on is_torch_version("<", "<target>"), so that when the fix is released it stops doing anything. Do you know which release pytorch#187522 is targeted to? If it's tagged, then I'll gate on that specific version.

We could do `pytest.mark.xfails(is_torch_version(">", current_stable_version)`?

Shreyas-jk: Confirmed pytorch#187522 is merged to main (by malfet), but PyTorch didn't set a milestone on it. Since 2.12 has already shipped, it lands in 2.13 so I'll gate on is_torch_version("<", "2.13.0"). Shout if you'd rather wait for the tag, but the fix is on main so 2.13 is where it surfaces.

Isalia20: I doubt it will land on 2.13. PR submissions for 2.13 was closed on:
`M2: All PRs landed in PyTorch repo / Feature Submission Closed (5/6/26)`
and that PR was merged after it, so please guard it on 2.14

Shreyas-jk: Good catch, thanks. Didn't realize it missed the 2.13 feature window. Guarding on
is_torch_version("<", "2.14.0") and skipping the test at >=2.14.0 to match. Pushing
the pytest port + gate shortly.

Shreyas-jk: Done, gated the MPS workaround for is_torch_version("<", "2.14.0") and converted the test to pytest with a skipif condition at >=2.14.0 (an xfail condition will XPASS when the upstream patch is applied, thus skipif works better). A new commit on top;, thank you @Isalia20 for the 2.14 fix.

Shreyas-jk:
Gentle follow-up, it's been about a month since I pushed the requested changes
(2.14 version gate + pytest port + skipif). Happy to adjust anything further if
there are more comments. @yiyixuxu, tagging you in case this fell off the radar and
would appreciate a look when you have a chance. Thanks!

Shreyas-jk:
Rebased onto main; approved since Jun 26, and the review comments were addressed in `7eb4c81` (now `dd10d2d5` after the rebase — the test moved to pytest style to match the migration that landed on `main` since).

One thing that's changed since and is worth weighing: the workaround is gated on `is_torch_version("<", "2.14.0")`, because the underlying fix landed upstream in pytorch#187522. So this PR has an expiry date — once 2.14 ships it's inert, and its entire value is for users on current stable. If it isn't going to land before then I'd rather close it than have it merge as dead code.

@yiyixuxu @sayakpaul — anything else needed from me?

RudraMantri123:
@Shreyas-jk heads-up: my #14459 overlaps with this (I found it only after opening — apologies, this PR is three months earlier and has been approved since June). I've posted a consolidation map on my PR proposing either path: yours merges now and I rescope mine to the `attention.py` duplicate implementation + tests that yours doesn't touch, or the reverse. Your call and the maintainers' — no preference on credit here, just want one of them to land; the underlying bug has open reports going back 16 months (#11229).

Shreyas-jk:
@RudraMantri123 No apology needed, and thanks for writing up the map rather than just opening in parallel. Agreed on the first path: this one is approved and rebased, so merging it and rescoping yours to the attention.py duplicate plus the tests this doesn't cover is the cleaner split. I'll stay out of that file.

For maintainers, one thing worth weighing on timing: the workaround is gated on is_torch_version("<", "2.14.0") because the underlying fix landed upstream in pytorch#187522, so its entire value is for users on current stable. Two independent PRs opened three months apart, plus reports going back to #11229, suggests people are still hitting this on 2.13.x today.

@yiyixuxu @sayakpaul approved since Jun 26 and rebased onto main. Happy to rebase again if it has gone stale.

PR #13712, opened 2026-05-10
Allow bucket reshuffling with DreamBooth caches

azolotenkov:
# What does this PR do?

Allow DreamBooth bucket batches to reshuffle each epoch while keeping cached latents and custom-caption prompt embeddings aligned.

After #13353, bucket batches with cached latents/custom captions were kept in stable step order because caches were indexed by dataloader step. This fixes the underlying limitation by indexing cached latents and prompt embeddings by dataset sample index instead. The training dataloader can then reshuffle bucket batches each epoch without reading the wrong cached tensors.

The cache precompute pass now uses a non-dropping cache dataloader, so every sample that can appear in a later reshuffled training epoch has a cache entry.

This also avoids mutating static prompt embeddings inside the training loop. Each step now derives repeated prompt/text embeddings from the original static tensors, which keeps prior-preservation runs with multiple steps stable.

### Tested:

`Klein` smoke tests with `hf-internal-testing/tiny-flux2-klein`:

- static prompt, no prior, no cache
- static prompt, no prior, `--cache_latents`
- custom captions, no prior, no latent cache
- custom captions, no prior, `--cache_latents`
- static prompt + prior preservation, no cache
- static prompt + prior preservation, `--cache_latents`
- custom captions + prior preservation, no latent cache
- custom captions + prior preservation, `--cache_latents`
- custom captions + prior preservation + `--cache_latents`, crossing an epoch boundary with `max_train_steps=7`

`Flux2` smoke tests with `hf-internal-testing/tiny-flux2` using the standard tiny-model settings:

- no prior preservation, no cache, `train_batch_size=1`, `max_train_steps=2`
- prior preservation, `train_batch_size=2`, `max_train_steps=2`
- prior preservation + `--cache_latents`, `train_batch_size=2`, `max_train_steps=3`

## Before submitting

## Who can review?

@sayakpaul


examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -1748,25 +1766,35 @@ def get_sigmas(timesteps, n_dim=4, dtype=torch.float32):
     for epoch in range(first_epoch, args.num_train_epochs):
         transformer.train()
 
-        for step, batch in enumerate(train_dataloader):
+        for batch in train_dataloader:
             models_to_accumulate = [transformer]
+            indices = batch["indices"]
             prompts = batch["prompts"]
```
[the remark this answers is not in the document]
azolotenkov: I renamed the dataset/cache indices to `sample_indices` so they are distinct from the later timestep sampling `indices`.

sayakpaul:
@claude can you do a review here?

azolotenkov:
@sayakpaul Thanks for the review. 
I addressed Claude’s suggestions in the latest push by adding cache-fill sanity checks after precompute and clarifying why `instance_prompts` is kept separate from `prompts` in `collate_fn`.

examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -969,18 +973,17 @@ def collate_fn(examples, with_prior_preservation=False):
     pixel_values = torch.stack(pixel_values)
     pixel_values = pixel_values.to(memory_format=torch.contiguous_format).float()
 
-    batch = {"pixel_values": pixel_values, "prompts": prompts}
+    batch = {
+        "indices": indices,
+        "pixel_values": pixel_values,
+        "instance_prompts": instance_prompts,
```
sayakpaul, line 979: Why do we need the instance prompts?

examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -969,18 +973,17 @@ def collate_fn(examples, with_prior_preservation=False):
...
     return batch
 
 
 class BucketBatchSampler(BatchSampler):
-    def __init__(
-        self,
-        dataset: DreamBoothDataset,
-        batch_size: int,
-        drop_last: bool = False,
-        shuffle_batches_each_epoch: bool = True,
```
sayakpaul, line 982: Why is `shuffle_batches_each_epoch` going away? Because we're storing indices?

examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -989,37 +992,32 @@ def __init__(
...
+    def __iter__(self):
+        batches = []
         for indices_in_bucket in self.bucket_indices:
-            # Shuffle indices within the bucket
-            random.shuffle(indices_in_bucket)
-            # Create batches
-            for i in range(0, len(indices_in_bucket), self.batch_size):
-                batch = indices_in_bucket[i : i + self.batch_size]
+            shuffled_indices = indices_in_bucket.copy()
+            random.shuffle(shuffled_indices)
```
sayakpaul: We should be able to seed it too, for reproducibility?

examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -1599,32 +1592,70 @@ def _encode_single(prompt: str):
...
+        for batch in tqdm(cache_dataloader, desc="Caching latents"):
             with torch.no_grad():
+                sample_indices = batch["indices"]
                 if args.cache_latents:
                     with offload_models(vae, device=accelerator.device, offload=args.offload):
                         batch["pixel_values"] = batch["pixel_values"].to(
                             accelerator.device, non_blocking=True, dtype=vae.dtype
                         )
-                        latents_cache.append(vae.encode(batch["pixel_values"]).latent_dist)
+                        latents = vae.encode(batch["pixel_values"]).latent_dist.mode()
```
sayakpaul, line 1626: Why are we switching to `.mode()` here?

sayakpaul (commented):
The diff looks very complicated to me, to the point where it's getting increasingly difficult for me to wrap my head around. Can you please explain the major changes in line?

examples/dreambooth/train_dreambooth_lora_flux2.py
azolotenkov: With `train_batch_size=2`, custom captions, and prior preservation, `collate_fn` receives two examples:

```python
examples = [
    {
        "index": 0,
        "instance_prompt": "a sks dog on grass",
        "instance_images": inst_img_0,
        "class_prompt": "a dog",
        "class_images": class_img_0,
    },
    {
        "index": 1,
        "instance_prompt": "a sks dog near water",
        "instance_images": inst_img_1,
        "class_prompt": "a dog",
        "class_images": class_img_1,
    },
]
```

First, it stores sample indices and instance image tensors:

```python
indices = [example["index"] for example in examples]
pixel_values = [example["instance_images"] for example in examples]
```

Result:

```python
indices = [0, 1]
pixel_values = [inst_img_0, inst_img_1]
```

Then it builds two prompt lists from the same source:

```python
instance_prompts = [example["instance_prompt"] for example in examples]
prompts = [example["instance_prompt"] for example in examples]
```

Result:

```python
instance_prompts = [
    "a sks dog on grass",
    "a sks dog near water",
]

prompts = [
    "a sks dog on grass",
    "a sks dog near water",
]
```

If prior preservation is enabled, `collate_fn` appends class images and class prompts only to the training batch lists:

```python
if with_prior_preservation:
    pixel_values += [example["class_images"] for example in examples]
    prompts += [example["class_prompt"] for example in examples]
```

Result:

```python
pixel_values = [
    inst_img_0,
    inst_img_1,
    class_img_0,
    class_img_1,
]

prompts = [
    "a sks dog on grass",
    "a sks dog near water",
    "a dog",
    "a dog",
]
```

`instance_prompts` does not change:

```python
instance_prompts = [
    "a sks dog on grass",
    "a sks dog near water",
]
```

Batch:

```python
batch = {
    "indices": [0, 1],
    "pixel_values": tensor([inst_img_0, inst_img_1, class_img_0, class_img_1]),
    "instance_prompts": [
        "a sks dog on grass",
        "a sks dog near water",
    ],
    "prompts": [
        "a sks dog on grass",
        "a sks dog near water",
        "a dog",
        "a dog",
    ],
}
```

Why is the instance-only prompt list needed?

During custom-caption precompute, we cache prompt embeddings per dataset sample.

See lines 1604-1606:

```python
if train_dataset.custom_instance_prompts:
    prompt_embeds_cache = [None] * train_dataset.num_instance_images
    text_ids_cache = [None] * train_dataset.num_instance_images
```

Then, inside the cache loop, we need to encode only the instance captions, not the class prompts.

See lines 1632-1646:

```python
if train_dataset.custom_instance_prompts:
    prompt_embeds, text_ids = compute_text_embeddings(
        batch["instance_prompts"], text_encoding_pipeline
    )
    for i, idx in enumerate(sample_indices):
        prompt_embeds_cache[idx] = prompt_embeds[i : i + 1]
        text_ids_cache[idx] = text_ids[i : i + 1]
```

Using the example:

```python
batch["instance_prompts"] = ["a sks dog on grass", "a sks dog near water"]
```

So we cache:

```python
prompt_embeds_cache[0] = embed("a sks dog on grass")
prompt_embeds_cache[1] = embed("a sks dog near water")
```

Custom captions are per instance image.

Class prompts are not cached the same way. The class prompt is shared. Usually it is one fixed string:

```bash
--class_prompt "a dog"
```

So every class image uses the same prompt.

The script already encodes this fixed class prompt separately before training:

```python
class_prompt_hidden_states, class_text_ids = compute_text_embeddings(
    args.class_prompt, text_encoding_pipeline
)
```

Then the training loop repeats the class embedding for the current batch size.

See lines 1791-1795:

```python
if args.with_prior_preservation:
    prompt_embeds = torch.cat(
        [prompt_embeds, class_prompt_hidden_states.repeat(len(sample_indices), 1, 1)], dim=0
    )
    text_ids = torch.cat([text_ids, class_text_ids.repeat(len(sample_indices), 1, 1)], dim=0)
```

Using the example:

```python
prompt_embeds = [
    cached_embed_for_sample_0,
    cached_embed_for_sample_1,
    embed("a dog"),
    embed("a dog"),
]
```

This matches the image/latent order. 

See lines 1805-1810:

```python
model_input = torch.cat([instance_latents_cache[idx] for idx in sample_indices], dim=0)
if args.with_prior_preservation:
    model_input = torch.cat(
        [model_input, torch.cat([class_latents_cache[idx] for idx in sample_indices], dim=0)],
        dim=0,
    )
```

So the latent batch is:

```python
model_input = [
    instance_latent_0,
    instance_latent_1,
    class_latent_0,
    class_latent_1,
]
```

And the prompt batch is:

```python
prompt_embeds = [
    embed("a sks dog on grass"),
    embed("a sks dog near water"),
    embed("a dog"),
    embed("a dog"),
]
```

The ordering matches.

Also, we need an instance-only prompt list.

I could remove `instance_prompts` from `collate_fn` because after `collate_fn`, the first `len(batch["indices"])` prompts are always the instance prompts.

Then we have:
```python
batch = {
    "indices": indices,
    "pixel_values": pixel_values,
    "prompts": prompts,
}
```
Result:

```python
batch["prompts"] = [
    "a sks dog on grass",
    "a sks dog near water",
    "a dog",
    "a dog",
]

batch["indices"] = [0, 1]
```

So we can derive:

```python
sample_indices = batch["indices"]
instance_prompts = batch["prompts"][: len(sample_indices)]
```

Result:

```python
instance_prompts = [
    "a sks dog on grass",
    "a sks dog near water",
]
```

Then cache precompute can use the local variable:

```python
prompt_embeds, text_ids = compute_text_embeddings(instance_prompts, text_encoding_pipeline)
```

This gives the same behavior with less batch state.

examples/dreambooth/train_dreambooth_lora_flux2.py
azolotenkov: Yes. `shuffle_batches_each_epoch=False` was needed because caches were step-indexed: `cache[step]` had to match the exact batch seen at that dataloader step during precompute.

This PR makes caches sample-indexed instead:

`dataset index -> batch["indices"] -> cache[index] -> training retrieves cache[index]`

So cache correctness no longer depends on stable dataloader step order, and epoch-wise reshuffling is safe.

examples/dreambooth/train_dreambooth_lora_flux2.py
azolotenkov: I thought it was already seeded indirectly because `accelerate.utils.set_seed` seeds Python `random`, and `BucketBatchSampler` uses `random.shuffle`, so reshuffling should be reproducible when `--seed` is set.

But you’re right the sampler sequence can be affected by any other code that consumes Python random state. I can make this explicit by passing `args.seed` into `BucketBatchSampler` and using a sampler-local `random.Random(seed)` instance.

examples/dreambooth/train_dreambooth_lora_flux2.py
azolotenkov: `latents` contains the encoded latents for the whole batch. `vae.encode(image).latent_dist` is not one latent tensor. It is a probability distribution, right? To turn that distribution into a latent tensor we use `.mode()`:

```python
def mode(self):
    return self.mean
```

No random noise. 

The script already uses `.mode()` for the non-cached path:

See line:1815

```
model_input = vae.encode(pixel_values).latent_dist.mode()
```

So the cached path should also use `.mode()` to match that behavior.

azolotenkov:
@sayakpaul Have you had a chance to look at it?

examples/dreambooth/train_dreambooth_lora_flux2.py
sayakpaul: Yeah let's do that.

sayakpaul (commented):
Just one comment. Let's try to avoid this kind of heavy PRs in the future. There's lot going on which are very easy to miss.

examples/dreambooth/train_dreambooth_lora_flux2.py
azolotenkov: I updated both scripts to pass `args.seed` into `BucketBatchSampler`.

azolotenkov:
@sayakpaul Thanks. And agreed on PR size. This one became larger because it changes cache indexing, prompt-cache alignment, and epoch reshuffling together, but I’ll keep future PRs smaller where possible.

azolotenkov:
@sayakpaul Have you had a chance to look at seeding (reproducibility)?

sayakpaul (approved):

sayakpaul:
Failing tests are unrelated.

PR #13716, opened 2026-05-11
better usage of UV_PRERELEASE=allow

sayakpaul:
# What does this PR do?

The huggingface_hub RC-testing bot (which produces branches like `ci-test-huggingface-hub-1.14.0.rc0-release`) rewrites every `uv pip install` in our workflows to `uv pip install --prerelease=allow` via a text substitution. On lines that already carried `--prerelease allow` (added in #12395 so non-RC CI can still install transformers-from-main when it depends on a hub RC), this produced:

```bash
uv pip install --prerelease=allow --prerelease allow -U transformers@git+...
```

`uv` rejects this:

```bash
error: the argument '--prerelease <PRERELEASE>' cannot be used multiple times
```

Example: https://github.com/huggingface/diffusers/actions/runs/25439680072/job/74627420559

Switching to the `UV_PRERELEASE=allow` env-var prefix is equivalent and fixes the problem.

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -75,7 +75,7 @@ jobs:
       - name: Install dependencies
         run: |
           uv pip install -e ".[quality]"
-          uv pip uninstall transformers huggingface_hub && uv pip install --prerelease allow -U transformers@git+https://github.com/huggingface/transformers.git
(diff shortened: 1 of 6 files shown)


DN6 (approved):

Wauplin:
Thank you!

PR #13719, opened 2026-05-11
Incorporate safetensors support to TorchAO

### LINKED ISSUE #13713: Incorporate safetensors support to TorchAO
`transformers` supports safetensors when dealing with TorchAO quantization. We should do the same for `diffusers`. This will also enable group offloading with disk support for TorchAO.

If anyone's interested in working on this, that would be huge!

hlky:
# What does this PR do?

Fixes #13713

### Preexisting test failures

Some preexisting tests failed in my environment on `main`, the same tests fail on this PR, no new test failures are introduced.

<details><summary>Failed tests</summary>
<p>

```
FAILED tests/quantization/torchao/test_torchao.py::TorchAoTest::test_device_map - RuntimeError: cutlass cannot initialize
FAILED tests/quantization/torchao/test_torchao.py::TorchAoTest::test_memory_footprint - AssertionError: False is not true
FAILED tests/quantization/torchao/test_torchao.py::TorchAoTest::test_model_memory_usage - assert (34840064 / 34806272) >= 1.02
FAILED tests/quantization/torchao/test_torchao.py::TorchAoTest::test_modules_to_not_convert - AssertionError: False is not true
FAILED tests/quantization/torchao/test_torchao.py::TorchAoTest::test_quantization - TypeError: IntxWeightOnlyConfig.__init__() got an unexpected keyword argument 'dtype'
FAILED tests/quantization/torchao/test_torchao.py::TorchAoTest::test_training - RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

</p>
</details> 

### Test coverage

More test coverage would be useful.

## Who can review?

@sayakpaul @DN6


wadeKeith (commented):
Withdrawn; this review was too cursory.

sayakpaul:
@hlky thanks.

I ran some of those tests on an H100 with the following env:

```bash
- 🤗 Diffusers version: 0.39.0.dev0
- Platform: Linux-5.15.0-1048-aws-x86_64-with-glibc2.31
- Python version: 3.12.12
- PyTorch version (GPU?): 2.11.0+cu129 (True)
- Jax version: not installed
- Huggingface_hub version: 1.11.0
- Transformers version: 5.6.0.dev0
- Accelerate version: 1.12.0
- PEFT version: 0.18.2.dev0
- Bitsandbytes version: 0.49.0
- Safetensors version: 0.7.0
- Accelerator: NVIDIA H100 80GB HBM3, 81559 MiB
```

Some passed and some failed not because all of them are functionally incorrect. They failed because of the dependency on the hardware. Stuff like assertion errors on slices, memory ratio (expectation-based) are a bit flaky, which I think are safe to ignore. I have fixed some of them in https://github.com/huggingface/diffusers/pull/13330

src/diffusers/hooks/group_offloading.py
```diff
@@ -146,26 +157,28 @@ def __init__(
...
+        all_tensors = []
+        for module in self.modules:
+            all_tensors.extend(list(module.parameters()))
+            all_tensors.extend(list(module.buffers()))
+        all_tensors.extend(self.parameters)
+        all_tensors.extend(self.buffers)
+        all_tensors = list(dict.fromkeys(all_tensors))  # Remove duplicates
+
+        self.tensor_to_key = {tensor: f"tensor_{i}" for i, tensor in enumerate(all_tensors)}
+        self._torchao_disk_key_remap: dict[str, str] = {}
```
sayakpaul: We don't need to hold it outside of the `self.offload_to_disk_path` no?

src/diffusers/hooks/group_offloading.py
```diff
@@ -238,19 +271,79 @@ def _process_tensors_from_modules(self, pinned_memory=None, default_stream=None)
...
 
-    def _check_disk_offload_torchao(self):
-        all_tensors = list(self.tensor_to_key.keys())
-        has_torchao = any(_is_torchao_tensor(t) for t in all_tensors)
-        if has_torchao:
-            raise ValueError(
-                "Disk offloading is not supported for TorchAO quantized tensors because safetensors "
-                "cannot serialize TorchAO subclass tensors. Use memory offloading instead by not "
-                "setting `offload_to_disk_path`."
+    def _get_disk_state_dict(self):
```
sayakpaul: This diff seems complicated to me. I think the code would simplify a bit to try to aim for a pattern like:

```py
if has_torchao:
    handle_for_torchao_safetensors_group_offloading_with_disk()
else:
    keep_the_existing_code
```

Once this pattern is established, we could look into what needs to be factored out in utilities.

If it helps, I wouldn't mind doing it in a separate PR.

src/diffusers/models/modeling_utils.py
```diff
@@ -759,6 +759,17 @@ def save_pretrained(
 
         # Save the model
         state_dict = model_to_save.state_dict()
+        quantization_metadata = {}
+        if safe_serialization and hf_quantizer is not None:
```
sayakpaul: Should it also include a check if the quantizer is of TorchAO type just to be explicit?

src/diffusers/models/modeling_utils.py
```diff
@@ -759,6 +759,17 @@ def save_pretrained(
...
+            get_state_dict_and_metadata = getattr(hf_quantizer, "get_state_dict_and_metadata", None)
+            if callable(get_state_dict_and_metadata):
+                state_dict_and_metadata = get_state_dict_and_metadata(model_to_save)
+            else:
+                state_dict_and_metadata = model_to_save.state_dict()
+            if isinstance(state_dict_and_metadata, tuple):
+                state_dict, quantization_metadata = state_dict_and_metadata
+            else:
+                state_dict = state_dict_and_metadata
```
sayakpaul: Where do we get a `model_to_save.state_dict()` that is of type `tuple`?

src/diffusers/models/modeling_utils.py
```diff
@@ -803,15 +814,21 @@ def save_pretrained(
                 shard = {tensor: state_dict[tensor].contiguous() for tensor in tensors}
                 filepath = os.path.join(save_directory, filename)
                 if safe_serialization:
+                    metadata = dict(state_dict_split.metadata)
+                    metadata.update(quantization_metadata)
+                    metadata = {k: str(v) if not isinstance(v, str) else v for k, v in metadata.items()}
                     # At some point we will need to deal better with save_function (used for TPU and other distributed
                     # joyfulness), but for now this enough.
-                    safetensors.torch.save_file(shard, filepath, metadata={"format": "pt"})
+                    safetensors.torch.save_file(shard, filepath, metadata=metadata)
```
sayakpaul, line 821: If we're saving all metadata, then I'd restrict it to torchao only.

src/diffusers/models/modeling_utils.py
```diff
@@ -1367,11 +1384,27 @@ def from_pretrained(cls, pretrained_model_name_or_path: str | os.PathLike | None
...
             hf_quantizer.preprocess_model(
-                model=model, device_map=device_map, keep_in_fp32_modules=keep_in_fp32_modules
+                model=model,
+                device_map=device_map,
+                keep_in_fp32_modules=keep_in_fp32_modules,
+                checkpoint_files=checkpoint_files,
             )
 
+        if hf_quantizer is not None and hf_quantizer.quantization_config.quant_method == QuantizationMethod.TORCHAO:
+            is_parallel_loading_enabled = False
```
sayakpaul, line 1400: Why is that?

tests/quantization/torchao/test_torchao.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import gc
+import os
```
sayakpaul: We could move this test to https://github.com/huggingface/diffusers/blob/303f3a7061f53054287fb847a2c59078ea2b0218/tests/models/testing_utils/quantization.py#L835

Also, I don't think we have to be exhaustive across the different quant formats offered by TorchAO. Just 2/3 (int8, fp8, etc.) are enough IMO.

sayakpaul (commented):
Left some comments. LMK what you think.

hlky:
@sayakpaul Thanks for the review. I rebased this onto the latest `upstream/main` and pushed the updated branch.

Changes made:
- Simplified group-offload disk handling so the existing non-TorchAO path stays separate from the TorchAO safetensors path.
- Restricted safetensors metadata save/load handling to TorchAO explicitly.
- Removed the ambiguous `state_dict()` tuple fallback.
- Disabled parallel shard loading only when TorchAO safetensors metadata is present, since reconstruction can carry pending tensor pieces across shards.
- Moved the broader TorchAO safetensors serialization coverage into `TorchAoTesterMixin` and kept the Flux-specific disk group-offload test in `test_torchao.py`.
- Gated TorchAO safetensors support on `torchao >= 0.16.0`; `0.15.0` exposes the helper APIs but does not support `Int8Tensor` in the safetensors allowlist.

sayakpaul:
@DN6 could you do a review?

sayakpaul:
@askserge could you do a review as well?

hlky:
@sayakpaul Pushed a follow-up commit: `8c1be50fc`.

Addressed the latest review points:

- TorchAO disk-offload reconstruction now raises if safetensors metadata reconstruction fails instead of falling back to raw flattened tensors.
- The TorchAO disk key remap is now derived during load as well as save, so a fresh `ModuleGroup` can reuse an existing offload file safely.
- Cleaned up the TorchAO quantizer metadata state:
  - removed unused reconstruction bookkeeping
  - removed the unused metadata setter
  - `get_weight_names()` now derives directly from safetensors metadata
- Strengthened the safetensors serialization tests to compare original vs reloaded model outputs, rather than only checking for non-NaN outputs.
- Added coverage for reusing an existing disk offload directory with a fresh quantized model.

Validation was run on a Runpod RTX 3090 with `torch==2.9.1+cu128`, `torchao==0.17.0`.

Focused PR coverage:

```bash
pytest -q \
  tests/quantization/torchao/test_torchao.py::TorchAoSerializationTest::test_group_offload_to_disk \
  tests/models/transformers/test_models_transformer_flux.py::TestFluxTransformerTorchAo::test_torchao_quantization_serialization[int8wo] \
  tests/models/transformers/test_models_transformer_flux.py::TestFluxTransformerTorchAo::test_torchao_quantization_serialization[int8dq] \
  tests/models/transformers/test_models_transformer_flux.py::TestFluxTransformerTorchAo::test_torchao_quantization_sharded_serialization[int8dq]
```

Result: `4 passed`.

Full TorchAO test file:

```bash
pytest -q tests/quantization/torchao/test_torchao.py
```

Result: `13 passed, 10 skipped, 6 failed`.

The 6 failures are the same known TorchAO/env failures unrelated to this PR:
- int4 tests requiring `mslk >= 1.0.0`
- `IntxWeightOnlyConfig(dtype=...)` API mismatch
- existing `modules_to_not_convert` assertion drift
- existing training grad behavior failure

src/diffusers/hooks/group_offloading.py
```diff
@@ -146,6 +161,11 @@ def __init__(
         self.offload_to_disk_path = offload_to_disk_path
         self._is_offloaded_to_disk = False
 
+        self.tensor_to_key = {}
+        self.key_to_tensor = {}
```
sayakpaul: Does it need to be initialized unconditionally? Not a big deal but just a nit. I think we will only ever need them when `offload_to_disk_path is not None`?

src/diffusers/hooks/group_offloading.py
```diff
@@ -179,6 +200,26 @@ def _to_cpu(tensor, low_cpu_mem_usage):
...
+    @staticmethod
+    def _get_torchao_subset_metadata_for_unflatten(metadata):
+        tensor_names = metadata.get("tensor_names")
+        if tensor_names is None:
+            return None
+
+        try:
+            tensor_names = json.loads(tensor_names)
+        except (TypeError, json.JSONDecodeError):
+            return None
```
sayakpaul: Should it be logged as well? Provides more clarity to the user.

src/diffusers/hooks/group_offloading.py
```diff
@@ -238,18 +279,106 @@ def _process_tensors_from_modules(self, pinned_memory=None, default_stream=None)
...
             raise ValueError(
-                "Disk offloading is not supported for TorchAO quantized tensors because safetensors "
-                "cannot serialize TorchAO subclass tensors. Use memory offloading instead by not "
-                "setting `offload_to_disk_path`."
+                "Disk offloading TorchAO quantized tensors requires torchao >= 0.16.0 because older torchao "
+                "versions cannot serialize tensor subclasses with safetensors. Use memory offloading instead by "
+                "not setting `offload_to_disk_path`."
             )
 
+    def _get_torchao_disk_state_dict(self):
```
sayakpaul: It would be nice to have a short explainer for future reference as why we have to perform this.

src/diffusers/hooks/group_offloading.py
```diff
@@ -238,18 +279,106 @@ def _process_tensors_from_modules(self, pinned_memory=None, default_stream=None)
...
+
+    def _release_torchao_onload_tensors(self):
+        for tensor_obj in self.tensor_to_key.keys():
+            if _is_torchao_tensor(tensor_obj):
+                placeholder = tensor_obj.to(self.offload_device)
+                _swap_torchao_tensor(tensor_obj, placeholder)
+            else:
+                tensor_obj.data = torch.empty_like(tensor_obj.data, device=self.offload_device)
+
+    def _onload_torchao_from_disk(self):
```
sayakpaul: I believe we could simplify things a little bit here:

```diff
diff --git a/src/diffusers/hooks/group_offloading.py b/src/diffusers/hooks/group_offloading.py
index dbcb135e8..084a014e1 100644
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -347,7 +347,9 @@ class ModuleGroup:
             else:
                 tensor_obj.data = torch.empty_like(tensor_obj.data, device=self.offload_device)
 
-    def _onload_torchao_from_disk(self):
+    def _onload_from_disk(self):
+        self._check_disk_offload_torchao_support()
+
         if self.stream is not None:
             # Wait for previous Host->Device transfer to complete
             self.stream.synchronize()
@@ -357,14 +359,18 @@ class ModuleGroup:
 
         with context:
             device = str(self.onload_device) if self.stream is None else "cpu"
-            loaded_tensors = self._load_torchao_disk_state_dict(device=device)
+            if self._has_torchao_tensors:
+                loaded_tensors = self._load_torchao_disk_state_dict(device=device)
+            else:
+                loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device=device)
 
             if self.stream is not None:
                 pinned_memory = {
                     tensor_obj: loaded_tensors[self.tensor_to_key[tensor_obj]].pin_memory()
                     for tensor_obj in self.tensor_to_key
                 }
-                self._process_tensors_from_modules(pinned_memory, default_stream=current_stream)
+                for tensor_obj in self.tensor_to_key:
+                    self._transfer_tensor_to_device(tensor_obj, pinned_memory[tensor_obj], current_stream)
             else:
                 for tensor_obj in self.tensor_to_key:
                     self._transfer_tensor_to_device(
@@ -373,39 +379,6 @@ class ModuleGroup:
                         default_stream=None,
                     )
 
-    def _onload_from_disk(self):
-        self._check_disk_offload_torchao_support()
-
-        if self._has_torchao_tensors:
-            self._onload_torchao_from_disk()
-            return
-
-        if self.stream is not None:
-            # Wait for previous Host->Device transfer to complete
-            self.stream.synchronize()
-
-        context = nullcontext() if self.stream is None else self._torch_accelerator_module.stream(self.stream)
-        current_stream = self._torch_accelerator_module.current_stream() if self.record_stream else None
-
-        with context:
-            # Load to CPU (if using streams) or directly to target device, pin, and async copy to device
-            device = str(self.onload_device) if self.stream is None else "cpu"
-            loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device=device)
-
-            if self.stream is not None:
-                for key, tensor_obj in self.key_to_tensor.items():
-                    pinned_tensor = loaded_tensors[key].pin_memory()
-                    tensor_obj.data = pinned_tensor.to(self.onload_device, non_blocking=self.non_blocking)
-                    if self.record_stream:
-                        tensor_obj.data.record_stream(current_stream)
-            else:
-                onload_device = (
-                    self.onload_device.type if isinstance(self.onload_device, torch.device) else self.onload_device
-                )
-                loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device=onload_device)
-                for key, tensor_obj in self.key_to_tensor.items():
-                    tensor_obj.data = loaded_tensors[key]
-
     def _onload_from_memory(self):
         if self.stream is not None:
             # Wait for previous Host->Device transfer to complete

```

src/diffusers/hooks/group_offloading.py
```diff
@@ -292,8 +421,29 @@ def _onload_from_memory(self):
             else:
                 self._process_tensors_from_modules(None)
 
+    def _offload_torchao_to_disk(self):
```
sayakpaul: Same as above I believe?

src/diffusers/models/modeling_utils.py
```diff
@@ -716,12 +716,17 @@ def save_pretrained(
...
+            hf_quantizer is not None and hf_quantizer.quantization_config.quant_method == QuantizationMethod.TORCHAO
+        )
         if hf_quantizer is not None:
             quantization_serializable = (
                 hf_quantizer is not None
                 and isinstance(hf_quantizer, DiffusersQuantizer)
                 and hf_quantizer.is_serializable
             )
+            if safe_serialization and is_torchao_quantized:
+                quantization_serializable = quantization_serializable and hf_quantizer.is_safetensors_serializable
```
sayakpaul: I am not following. Why does it have to only apply to TorchAO, though?

src/diffusers/models/modeling_utils.py
```diff
@@ -1367,11 +1382,34 @@ def from_pretrained(cls, pretrained_model_name_or_path: str | os.PathLike | None
...
+        checkpoint_files = resolved_model_file
+        if hf_quantizer is not None:
+            if is_torchao_quantized and hasattr(hf_quantizer, "set_metadata"):
+                hf_quantizer.set_metadata(checkpoint_files)
+                has_torchao_safetensors_metadata = bool(getattr(hf_quantizer, "metadata", None))
+            quantized_weight_names = []
+            if is_torchao_quantized and hasattr(hf_quantizer, "get_weight_names"):
+                quantized_weight_names = hf_quantizer.get_weight_names()
+            if quantized_weight_names:
+                loaded_keys = list(quantized_weight_names)
```
sayakpaul: This path feels like it should be restricted to torchao only.

tests/quantization/torchao/test_torchao.py
```diff
@@ -595,7 +595,9 @@ def _check_serialization_expected_slice(self, quant_type, expected_slice, device
         with tempfile.TemporaryDirectory() as tmp_dir:
             quantized_model.save_pretrained(tmp_dir, safe_serialization=False)
             loaded_quantized_model = FluxTransformer2DModel.from_pretrained(
-                tmp_dir, torch_dtype=torch.bfloat16, use_safetensors=False
+                tmp_dir,
+                torch_dtype=torch.bfloat16,
+                use_safetensors=False,
```
sayakpaul: Seems unrelated.

sayakpaul (commented):
Thanks! Left some further comments.

src/diffusers/models/modeling_utils.py
hlky: `is_serializable` is intentionally format-agnostic here: it answers whether the quantized model can be saved at all, covering the existing TorchAO constraints like `huggingface_hub >= 0.25.0` and offloaded modules.

`safetensors` needs one extra capability check for TorchAO because tensor subclasses only became safely flattenable/reconstructable through TorchAO’s safetensors support in `torchao >= 0.16.0`. So TorchAO can be serializable in the general sense while still not being serializable with `safe_serialization=True`.

That is why this is a separate `is_safetensors_serializable` capability instead of folding it into `is_serializable`: otherwise we would either reject valid non-safetensors saves or incorrectly allow safetensors saves that cannot round-trip.

hlky:
@sayakpaul Thanks for the follow-up review. I addressed the requested cleanup:

- Simplified the TorchAO metadata/load-key path so it is guarded directly by `is_torchao_quantized`. Since that already implies `hf_quantizer` is present and is the TorchAO quantizer, the extra `hf_quantizer is not None` / `hasattr(...)` checks were unnecessary.
- Kept `has_torchao_safetensors_metadata` tied to actual metadata presence rather than TorchAO quantization in general. TorchAO can still be used for normal on-load quantization from checkpoints without TorchAO safetensors metadata, so we should only override `loaded_keys` and disable parallel shard loading when metadata was actually found.
- Merged the duplicated TorchAO/non-TorchAO disk offload/onload paths in group offloading, leaving only the TorchAO-specific save/load/release pieces branched internally.
- Restricted `checkpoint_files` plumbing to TorchAO preprocessing only.
- Moved disk-offload bookkeeping initialization into the disk-offload branch.
- Added a warning for malformed TorchAO safetensors metadata fallback and a short comment explaining why the TorchAO disk path goes through flatten/unflatten.

Validation:
- Local `ruff`, `compileall`, and `git diff --check` passed.
- Focused CUDA tests on an A40 passed: `4 passed`.
- Targeted regular TorchAO load path without TorchAO safetensors metadata passed.
- Full `tests/quantization/torchao/test_torchao.py` keeps the same known profile: `13 passed`, `10 skipped`, `6 failed`. The failures match the known TorchAO/upstream/environment issues already discussed, including missing `mslk` for int4 and the pending TorchAO config/API mismatch.

src/diffusers/models/modeling_utils.py
```diff
@@ -1367,10 +1382,31 @@ def from_pretrained(cls, pretrained_model_name_or_path: str | os.PathLike | None
...
-                model=model, device_map=device_map, keep_in_fp32_modules=keep_in_fp32_modules
-            )
+            preprocess_kwargs = {
+                "model": model,
+                "device_map": device_map,
+                "keep_in_fp32_modules": keep_in_fp32_modules,
+            }
+            if is_torchao_quantized:
+                preprocess_kwargs["checkpoint_files"] = checkpoint_files
+            hf_quantizer.preprocess_model(**preprocess_kwargs)
```
sayakpaul: Why do we need to pass `checkpoint_files` to `preprocess_model`?

tests/models/testing_utils/quantization.py
```diff
@@ -805,16 +805,16 @@ class TorchAoConfigMixin:
     }
 
     @staticmethod
-    def _get_quant_config(config_name):
+    def _get_quant_config(config_name, **config_kwargs):
```
sayakpaul: Is this just for `version`? If so, we could always use `version=2`?

tests/models/testing_utils/quantization.py
```diff
@@ -905,19 +905,59 @@ def test_torchao_quantized_layers(self, quant_type):
     def test_torchao_quantization_lora_inference(self, quant_type):
         self._test_quantization_lora_inference(TorchAoConfigMixin.TORCHAO_QUANT_TYPES[quant_type])
 
-    @pytest.mark.parametrize("quant_type", ["int8wo"], ids=["int8wo"])
+    @pytest.mark.parametrize("quant_type", ["int8wo", "int8dq"], ids=["int8wo", "int8dq"])
```
sayakpaul: Any reason to increase it?

sayakpaul (commented):
Apologies for the delay! 

Thanks! I left some minor comments and I think we're quite close to merging. @DN6 over to you. 

Currently, running the tests (edit: the new tests introduced in this PR are passing).

src/diffusers/models/modeling_utils.py
hlky: Good catch. Seems to be bad assumption from agent like "`checkpoint_files` used in `set_metadata` only for torchao" -> "`preprocess_model` has `kwargs`" -> "pass `checkpoint_files` to `preprocess_model` for torchao". Negative result like this is just as good as positive result though, it can be codified into a rule like "You must validate anything passed as `kwargs` is really consumed by the function".

tests/models/testing_utils/quantization.py
hlky: Yes looks like we can always use `version=2`. For context:

- TorchAO docs show Int8WeightOnlyConfig still defaults to version=1, but its implementation emits a deprecation warning and tells users to use version=2.
- Version 2 moves int8 configs toward the newer Int8Tensor/granularity path; version 1 is the older path using AffineQuantizedTensor-style internals.
- Int4WeightOnlyConfig already defaults to version=2, so setting version=2 uniformly is mostly about avoiding int8 version-1 deprecation/noise and aligning test serialization with the newer tensor subclass path.

tests/models/testing_utils/quantization.py
hlky: Likely expanded while moving tests into `TorchAoTesterMixin` after the [earlier comment](https://github.com/huggingface/diffusers/pull/13719#discussion_r3225401830). Sharded serialization already covers `int8dq` so should be fine to leave this as only `int8wo`.

src/diffusers/models/modeling_utils.py
```diff
@@ -803,15 +811,22 @@ def save_pretrained(
...
                     # At some point we will need to deal better with save_function (used for TPU and other distributed
                     # joyfulness), but for now this enough.
-                    safetensors.torch.save_file(shard, filepath, metadata={"format": "pt"})
+                    safetensors.torch.save_file(shard, filepath, metadata=metadata)
                 else:
                     torch.save(shard, filepath)
 
             if state_dict_split.is_sharded:
+                metadata = dict(state_dict_split.metadata)
+                if is_torchao_quantized:
```
DN6: Since `quantization_metdata` is always declared, we can just add it as long as a quantizer is present. 
```python
                if hf_quantizer is not None:
                      metadata.update(quantization_metdata)
```

DN6:
@hlky It's looking good. My main concern is adding a lot of `torchao` specific checks into the core model loading/saving logic. I think it would be better to achieve the same functionality via methods in the `hf_quantizer`.

e.g

```python
class DiffusersQuantizer(ABC): 

    def maybe_update_loaded_keys(self, loaded_keys: list[str], checkpoint_files: list[str]) -> list[str]:
        # Instead of: the `is_torchao_quantized` block in from_pretrained that calls
        #   `set_metadata(checkpoint_files)` + `get_weight_names()` and the
        #   `has_torchao_safetensors_metadata` bookkeeping — fold all of it into this method.
        return loaded_keys

    def maybe_update_state_dict(self, state_dict: dict[str, Any]) -> dict[str, Any]:
        # Instead of: `hf_quantizer.get_reconstructed_state_dict(state_dict)` in _load_shard_file,
        #   including the `hasattr(hf_quantizer, "get_reconstructed_state_dict")` probe.
        #   Just run it over each shard (or the single state dict) when a quantizer is present.
        return state_dict

    @property
    def supports_parallel_loading(self) -> bool:
        return True

    def get_state_dict_and_metadata(self, state_dict: dict[str, Any], safe_serialization: bool = False):
        return state_dict, {}

    @property
    def supports_safetensors_serialization(self) -> bool:
        return True
``` 

Also, would it be possible to move the group offloading support into a separate PR. I think the saving/loading component is much closer to merge, and I'd like to spend a bit more time to review the offloading component.

hlky:
Thanks @DN6, applied the requested scope split and refactor.

What changed on this PR:
- Removed the TorchAO disk group-offload changes from this branch, including the group-offload-specific test coverage.
- Kept this PR focused on TorchAO safetensors save/load.
- Added generic quantizer hooks on `DiffusersQuantizer` for loaded-key updates, state-dict updates, save metadata, parallel-loading support, and safetensors serialization support.
- Moved the TorchAO-specific metadata/key/state-dict reconstruction behavior into `TorchAoHfQuantizer`.
- Removed the explicit TorchAO branches from core model loading/saving code and replaced them with the generic quantizer hooks.
- Kept `checkpoint_files` out of `preprocess_model`; it is only consumed through the TorchAO metadata hook.

I split the group-offload work into a separate draft follow-up PR based on `upstream/main`: https://github.com/huggingface/diffusers/pull/13875

Validation:
- Local `ruff`, `compileall`, and `git diff --check` passed.
- CUDA A40 focused serialization tests passed: `2 passed`.
- Full `tests/quantization/torchao/test_torchao.py` result after removing the group-offload test: `12 passed`, `10 skipped`, `6 known failures`. The failures match the previously discussed TorchAO/upstream/environment issues.

tests/models/testing_utils/quantization.py
```diff
@@ -906,18 +907,58 @@ def test_torchao_quantization_lora_inference(self, quant_type):
...
+        model_loaded = self.model_class.from_pretrained(
+            str(tmp_path), device_map=str(torch_device), use_safetensors=True
+        )
 
-        model.save_pretrained(str(tmp_path), safe_serialization=False)
+        with torch.no_grad():
+            output = model_loaded(**inputs, return_dict=False)[0].detach().cpu()
 
-        model_loaded = self.model_class.from_pretrained(str(tmp_path), device_map=str(torch_device))
+        torch.testing.assert_close(output, expected_output, rtol=1e-3, atol=1e-3)
```
DN6: Can we use: 
https://github.com/huggingface/diffusers/blob/7e87c9b8df639001c1a101e361a83df16af32a99/tests/testing_utils.py#L138

tests/models/testing_utils/quantization.py
```diff
@@ -906,18 +907,58 @@ def test_torchao_quantization_lora_inference(self, quant_type):
...
+        )
+
+        model_loaded = self.model_class.from_pretrained(
+            str(tmp_path), device_map=str(torch_device), use_safetensors=True
+        )
+
+        with torch.no_grad():
+            output = model_loaded(**inputs, return_dict=False)[0].detach().cpu()
+
+        torch.testing.assert_close(output, expected_output, rtol=1e-3, atol=1e-3)
```
DN6: Can we use: 
https://github.com/huggingface/diffusers/blob/7e87c9b8df639001c1a101e361a83df16af32a99/tests/testing_utils.py#L138

DN6 (approved):
LGTM. One final request to reuse the existing `assert_tensors_close` utility in the tests and we're good to merge 👍🏽 

Thank you @hlky

sayakpaul:
Failing tests are unrelated. Thanks, @hlky!

PR #13724, opened 2026-05-12
schedulers/flow_match_euler_discrete: fix double-shift bug by storing sigma_min/sigma_max before shift is applied

### LINKED ISSUE #13243: [BUG] FlowMatchEulerDiscreteScheduler.__init__ computes sigma_min/sigma_max after shift, causing duplicate shift in set_timesteps
### Describe the bug

### Reproduction

        )
print(scheduler.sigmas)
scheduler.set_timesteps(1000)
print(scheduler.sigmas)

### Logs

```shell

```

### System Info

Environment
diffusers version
Python 3.13

### Who can help?

_No response_

Dev-X25874:
# What does this PR do?

Fixes #13243

`FlowMatchEulerDiscreteScheduler.__init__` was storing `sigma_min` and `sigma_max` **after** the shift formula had already been applied to `sigmas`:

```python
sigmas = shift * sigmas / (1 + (shift - 1) * sigmas)  # shift applied here
...
self.sigma_min = self.sigmas[-1].item()  # already-shifted value stored
self.sigma_max = self.sigmas[0].item()  # already-shifted value stored
```

Then in `set_timesteps`, when no custom sigmas/timesteps are provided, the scheduler reconstructs a linspace from `sigma_max` → `sigma_min` (the shifted values), divides back by `num_train_timesteps`, and **applies the shift formula a second time** — resulting in a doubly-shifted sigma schedule that diverges from what the model was trained with.

This PR moves the `sigma_min`/`sigma_max` assignment to **before** the shift is applied, so `set_timesteps` always starts from the correct unshifted linear range and applies the shift exactly once.

The fix is surgical: `self.sigmas` and `self.timesteps` stored in `__init__` are unchanged. Only the two boundary reference scalars are corrected. Users passing custom `sigmas` or `timesteps` directly to `set_timesteps` are entirely unaffected.

## Before submitting

## Who can review?
@yiyixuxu

### DIFF
--- a/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
+++ b/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
@@ -128,6 +128,11 @@ class FlowMatchEulerDiscreteScheduler(SchedulerMixin, ConfigMixin):
         timesteps = torch.from_numpy(timesteps).to(dtype=torch.float32)
 
         sigmas = timesteps / num_train_timesteps
+        # Store sigma_min/sigma_max from the unshifted linear sigmas so that set_timesteps
+        # can reconstruct the correct input range and apply the shift exactly once.
+        self.sigma_min = sigmas[-1].item()
+        self.sigma_max = sigmas[0].item()
+
         if not use_dynamic_shifting:
             # when use_dynamic_shifting is True, we apply the timestep shifting on the fly based on the image resolution
             sigmas = shift * sigmas / (1 + (shift - 1) * sigmas)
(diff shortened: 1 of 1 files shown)


Dev-X25874:
Hey @yiyixuxu — could you please take a look at this PR when you get a chance? It fixes a double-shift bug in `FlowMatchEulerDiscreteScheduler` where `sigma_min`/`sigma_max` were being stored after the shift formula was applied, causing the shift to be applied a second time in `set_timesteps`. The fix is surgical — just moves two assignments before the shift, no behaviour change for users passing custom sigmas. Thanks!

PR #13736, opened 2026-05-13
[tests] Fix controlnet tests

sayakpaul:
# What does this PR do?

Fix https://github.com/huggingface/diffusers/actions/runs/25780617099/job/75722330514?pr=13733


sayakpaul:
@claude-2-serge could you perform a review?

dg845 (approved):
Thanks for the PR! The change from FP16 to FP32 looks good to me, and the tests look fixed as well.

But why does switching to FP32 fix the underlying issue(s) that caused the tests to fail in the first place? I see that full inference in the slow tests uses BF16 (Flux) or FP16 (Hunyuan, SD3); if there really is an underlying issue affecting FP16/BF16 ControlNet inference specifically, it seems like the (fast) tests won't catch it.

sayakpaul:
They are not fixing the assertion problem (the assertion slices are being changed). It's just a safeguard to avoid numerical problems in CPUs where our fast tests run.

PR #13739, opened 2026-05-13
[tests] refactor ltx2 autoencoder tests to use latest mixins

sayakpaul:
As the PR title says.

### DIFF
--- a/tests/models/autoencoders/test_models_autoencoder_kl_ltx2_audio.py
+++ b/tests/models/autoencoders/test_models_autoencoder_kl_ltx2_audio.py
@@ -14,10 +14,12 @@
 # limitations under the License.
 
 import pytest
+import torch
 
 from diffusers import AutoencoderKLLTX2Audio
+from diffusers.utils.torch_utils import randn_tensor
 
-from ...testing_utils import floats_tensor, torch_device
+from ...testing_utils import torch_device
 from ..testing_utils import BaseModelTesterConfig, MemoryTesterMixin, ModelTesterMixin, TrainingTesterMixin
 from .testing_utils import NewAutoencoderTesterMixin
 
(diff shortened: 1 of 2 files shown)


dg845 (commented):
Thanks for the PR! If I run the tests locally, the following tests fail for both the LTX-2.X video and audio VAEs (and [on the CI](https://github.com/huggingface/diffusers/actions/runs/25786984624/job/75742881416?pr=13739)):

1. `test_from_save_pretrained`
2. `test_from_save_pretrained_variant`
3. `test_determinism`
4. `test_sharded_checkpoints_with_parallel_loading`

Since `test_determinism` is one of the tests that fail, I wonder if the underlying non-determinism is also affecting the save-then-load tests as well.

I also see that the following tests fail locally (but not in the above CI run):

1. `test_sharded_checkpoints`
2. `test_sharded_checkpoints_with_variant`

Finally, the `test_group_offloading_with_disk[leaf_level-False]` test also fails locally for the LTX-2.X audio VAE, so it seems like it might have some offloading issues. Could you look into these test failures?

sayakpaul:
@dg845 thanks for testing, requested for a review a bit sooner. I think it's good now.

dg845 (approved):
Thanks! I see that the `TestAutoencoderKLLTX2AudioMemory::test_group_offloading_with_disk[leaf_level-False]` test still fails locally for the LTX-2 audio VAE, do you think we should skip this particular test?

dg845:
Actually, it seems like the LTX-2 audio VAE `test_group_offloading_with_disk[leaf_level-False]` is flaky, since sometimes it passes and sometimes it fails. Not sure why.

sayakpaul:
We can add `is_flaky` to the test to reflect that it's flaky?

PR #13744, opened 2026-05-14
[CI] implement a bot to remind prs to link issues if not.

sayakpaul:
# What does this PR do?

We implemented a system in #13380 to better tag issues and PRs for managing the maintenance load better ourselves. This was also to better prioritize PRs that are genuine and fix issues users are facing. 

To build on top of that, this PR implements a system to periodically remind PR authors if their PR isn't linked to an existing issue(s). We post three reminders (each 7 days apart) and ~close the PR automatically after that with an explanation~.

Testing here: https://github.com/sayakpaul/qwenimage-ptxla/pull/3


utils/remind_link_issue.py
```diff
@@ -0,0 +1,139 @@
...
+    lines = [
+        REMINDER_MARKER,
+        f"Hi @{author}, this PR does not appear to link an issue it fixes. "
+        "If this PR addresses an existing issue, please add a closing keyword "
+        "(e.g. `Fixes #1234`) to the PR description so the issue is linked. "
+        f"See the [contribution guide]({CONTRIBUTION_GUIDE_URL}) for more details.",
+        "",
+        f"Reminder **{count}/{MAX_REMINDERS}**.",
+    ]
+    if remaining > 0:
```
yiyixuxu: I think the 3 reminder schedule works well if we are to do auto-close, we may do that soon, 
But for now, let's keep it simple and just post one friendly reminder to see how it goes?

yiyixuxu (commented):
thanks, i left one feedback

utils/remind_link_issue.py
sayakpaul: Done!

utils/remind_link_issue.py
```diff
@@ -0,0 +1,104 @@
...
+def main():
+    token = os.environ["GITHUB_TOKEN"]
+    g = Github(token)
+    repo = g.get_repo(REPO)
+    owner, name = REPO.split("/", 1)
+
+    for pr in repo.get_pulls(state="open"):
+        if pr.draft:
+            continue
+        if pr.user is None:
```
yiyixuxu: I think maybe only look at recent PRs? (like from past 2 days or so)?
because we only post a reminder once, I think we pretty much only need to look at new PRs each day

yiyixuxu (commented):
thanks, I left one comment
in addition to serge's

utils/remind_link_issue.py
sayakpaul: Yeah good idea. Done.

yiyixuxu (approved):
thanks a lot!

PR #13749, opened 2026-05-14
feat: Add HiDream-O1 transformer and image generation pipeline

chinoll:
# What does this PR do?

This PR adds Diffusers support for HiDream-O1 image generation.

HiDream-O1 is a Qwen3-VL based image generation model that denoises raw RGB image patches directly.
Unlike HiDream-I1 and most image diffusion pipelines, it does not use a VAE component.

This PR adds:

- `HiDreamO1Transformer2DModel`, a `ModelMixin` / `ConfigMixin` wrapper for HiDream-O1 checkpoints.
- `HiDreamO1AttnProcessor`, a dedicated attention processor for the HiDream-O1 two-pass attention path.
- `HiDreamO1ImagePipeline`, a text-to-image pipeline for raw RGB patch denoising.
- Loading support for official Transformers-style HiDream-O1 checkpoints.
- A generation script at `scripts/generate_hidream_o1_image.py`.
- API documentation for the model and pipeline.
- Tests for model loading, serialization, attention processor behavior, official implementation parity, and pipeline smoke generation.

Original implementation and checkpoints:

- https://github.com/HiDream-ai/HiDream-O1-Image
- https://huggingface.co/HiDream-ai/HiDream-O1-Image
- https://huggingface.co/HiDream-ai/HiDream-O1-Image-Dev

## Notes

HiDream-O1 does not use a VAE. The pipeline prepares Qwen3-VL chat inputs, constructs O1 multimodal RoPE positions, denoises patchified RGB noise, and unpatchifies the final tensor into image space.

The transformer can also be loaded independently:

```python
import torch
from diffusers import HiDreamO1Transformer2DModel

transformer = HiDreamO1Transformer2DModel.from_pretrained(
    "HiDream-ai/HiDream-O1-Image",
    torch_dtype=torch.bfloat16,
)
```

The pipeline can be loaded with:

```python
import torch
from diffusers import HiDreamO1ImagePipeline

pipe = HiDreamO1ImagePipeline.from_pretrained(
    "HiDream-ai/HiDream-O1-Image",
    torch_dtype=torch.bfloat16,
)
pipe.to("cuda")

image = pipe(
    "A cinematic portrait of a glass astronaut standing in a neon-lit botanical garden.",
    generator=torch.Generator("cuda").manual_seed(32),
).images[0]
```

For the dev checkpoint:

```python
import torch
from diffusers import HiDreamO1ImagePipeline

pipe = HiDreamO1ImagePipeline.from_pretrained(
    "HiDream-ai/HiDream-O1-Image-Dev",
    torch_dtype=torch.bfloat16,
)
pipe.to("cuda")

image = pipe(
    "A cinematic portrait of a glass astronaut standing in a neon-lit botanical garden.",
    model_type="dev",
    generator=torch.Generator("cuda").manual_seed(32),
).images[0]
```

## Tests

```bash
PYTHONPATH=src python -m pytest -q tests/models/transformers/test_models_transformer_hidream_o1.py tests/pipelines/hidream_o1/test_pipeline_hidream_o1.py
```

Result:

```text
5 passed, 6 warnings
```

I also ran real image generation tests with the full and dev checkpoints in bfloat16, including multiple aspect ratios.

## Before submitting

- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the
case).
- [x] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/diffusers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/diffusers/tree/main/docs#writing-source-documentation).

## Who can review?
@yiyixuxu ,@asomoza , @sayakpaul

## Generate Image
hidream_o1_processor_full_bf16_exact_2048x1024_50_seed32


sayakpaul:

chinoll:
Thanks for the review. I pushed follow-up changes up to `f9e374a59` on `chinoll:hidream-o1-transformer-model`.

Summary by review item:

1. Attention backend: addressed. `HiDreamO1AttnProcessor` now uses `dispatch_attention_fn`, and the attention module follows the diffusers processor pattern with `AttentionModuleMixin`, `_default_processor_cls`, and `_available_processors`.
2. Runtime env vars: addressed. The flash-attn / RoPE environment-variable switches were removed.
3. Model class attributes: addressed for the required attributes. Added `_repeated_blocks`, `_skip_layerwise_casting_patterns`, and `_no_split_modules`. I did not add `_keep_in_fp32_modules` yet because it was phrased as optional; I can add it if maintainers prefer a specific list.
4. Pipeline device/dtype helpers: addressed. The pipeline now uses `self._execution_device` and `self.transformer.dtype`.
5. Pipeline `from_pretrained`: addressed. The custom pipeline override was removed; model loading compatibility stays in `HiDreamO1Transformer2DModel.from_pretrained`.
6. Scheduler / denoising: addressed. The pipeline no longer mutates `scheduler.timesteps` / `scheduler.sigmas` directly, and now routes custom schedules through `scheduler.set_timesteps(...)`. The default scheduler uses `prediction_type="sample"`, so the model's x0 prediction is passed to `scheduler.step()` directly instead of converting to a hand-written velocity prediction.
7. `model_type`: addressed. The magic `model_type` preset switch was removed from the pipeline call; full/dev settings are passed explicitly by users or by the helper script.
8. Training-time branch: intentionally kept for now. The public HiDream-O1 checkpoints are Qwen3-VL-style checkpoints, not a pruned image-only checkpoint, and keeping this code path helps preserve compatibility with the official module structure while loading the official weights directly.
9. Output fields / LM path / `lm_head`: intentionally kept for now. The official checkpoints include the CausalLM head and related Qwen3-VL weights. Removing `lm_head` / CausalLM-compatible structures would require key filtering or a conversion-only loading path, and would break direct loading from the official checkpoint format. The text-to-image pipeline does not use these fields, but they are retained to keep checkpoint compatibility and avoid silently dropping official weights.
10. Module constants: partially addressed. Removed obsolete scheduler constants from the pipeline. `PATCH_SIZE` and the official resolution buckets remain because HiDream-O1 is a raw-RGB patch model with fixed official generation buckets; I can move `PATCH_SIZE` reads to `self.transformer.config.patch_size` in a follow-up if preferred.
11. Missing docstrings: addressed. Added docs for `noise_scale_start`, `noise_scale_end`, and `noise_clip_std`.
12. Scheduler shift helper: addressed in `f9e374a59`. The helper is now explicit: it supports schedulers with `flow_shift` config or `set_shift()`, and raises a `ValueError` otherwise instead of silently probing/falling through.
13. Init-time weight initialization: addressed. Removed the extra `initialize_weights()` / `_init_weights()` calls from the O1 patch and final layers.
14. Generation script official-repo dependency: addressed. The script now uses diffusers components and no longer imports schedulers or helpers from the official repository.
15. Gradient checkpointing: addressed. The two-pass decoder path now uses the model's `_gradient_checkpointing_func`. I also verified locally that `HiDreamO1Transformer2DModel.enable_gradient_checkpointing()` propagates to the inner Qwen3-VL language model used by this path.
16. `/tmp` in tests: addressed. The parity test now requires `HIDREAM_O1_OFFICIAL_REPO` explicitly and no longer defaults to `/tmp/HiDream-O1-Image`.

Validation run locally:

- `python -m py_compile ...`
- `PYTHONPATH=src python -m pytest -q tests/pipelines/hidream_o1/test_pipeline_hidream_o1.py tests/models/transformers/test_models_transformer_hidream_o1.py` -> `5 passed, 1 skipped`
- `HIDREAM_O1_OFFICIAL_REPO=/tmp/HiDream-O1-Image PYTHONPATH=src python -m pytest -q tests/models/transformers/test_models_transformer_hidream_o1.py tests/pipelines/hidream_o1/test_pipeline_hidream_o1.py` -> `5 passed`

I also smoke-tested Dev bf16 generation on a CUDA machine at 1024x1024 with both the official Dev timestep schedule and the default 28-step schedule; both generated images successfully.

chinoll:
@claude review

yiyixuxu (commented):
thanks for open the PR
i have two feedbacks on high-level:
1. can you host the transformer in transformers  library using remote code?
2. can you host the pipeline in modular diffusers only? documentation here https://huggingface.co/docs/diffusers/main/en/modular_diffusers/overview we also have pretty good doc for AI agents on this so they should have pretty good idea how to turn it into modular one

chinoll:
@yiyixuxu thanks, pushed `4b5a9b9fe`.

This update moves HiDream-O1 to modular diffusers only: the classic pipeline/docs/tests were removed, and `HiDreamO1ModularPipeline` / `HiDreamO1AutoBlocks` were added with modular registration, docs, and tests. The O1-specific no-VAE raw RGB patchify/unpatchify path and scheduler support are kept in the modular blocks.

I did not move the transformer to Transformers remote code in this PR. O1 is Qwen3-VL-style, but its image-generation path is not compatible with plain `Qwen3VLForConditionalGeneration` because it uses a custom raw-RGB patch path, two-pass attention flow, and image output head. That likely needs a separate remote-code/checkpoint packaging change on the model repo side.

Validation: modular pipeline + transformer tests pass (`5 passed, 1 skipped`), and bf16 CUDA generation works for official Full O1 at 2048x2048 / 50 steps.

src/diffusers/models/transformers/transformer_hidream_o1.py
```diff
@@ -0,0 +1,858 @@
...
+            pixel_values=pixel_values,
+            pixel_values_videos=pixel_values_videos,
+            image_grid_thw=image_grid_thw,
+            video_grid_thw=video_grid_thw,
+            cache_position=cache_position,
+            **kwargs,
+        )
+
+
+class HiDreamO1ForConditionalGeneration(Qwen3VLPreTrainedModel, GenerationMixin):
```
yiyixuxu, line 584: Here you subclass from the transformer Qwen3VL and then monkey-patch it. Diffusers models always inherit from `ModelMixin', and most of the layers are defined in the same file, with the exception of some small, common norm/embedding layers we import from the common place. 

we really do not want a part-diffusers, part-transformers model in diffusers

I'm not asking you to upstreaem it into transformers, my ask was to package the custom modeling code as remote code on hub so it can be picked up using `AutoModel.from_pretrained(..., trust_remote_code=True)`

you can host the remote checkpoint on your personal hub account for the PR - we could either PR into the Hidream hub repo or host under a diffusers org once we merge the PR

src/diffusers/modular_pipelines/hidream_o1/modular_blocks_hidream_o1.py
```diff
@@ -0,0 +1,560 @@
...
+        position_ids=sample["position_ids"],
+        vinputs=patches,
+        timestep=timestep.reshape(-1),
+        token_types=sample["token_types"],
+        attention_kwargs=attention_kwargs,
+    )
+    return outputs.sample[0, sample["vinput_mask"][0]].unsqueeze(0)
+
+
+class HiDreamO1PromptSampleStep(ModularPipelineBlocks):
```
dg845, line 123: I think we should refactor these blocks into the standard files such as `before_denoise.py`, `denoise.py`, etc. as follows:

- `before_denoise.py`
    - `HiDreamO1SetTimestepsStep` 
    - `HiDreamO1PromptSampleStep`
    - `HiDreamO1PrepareImageNoiseStep`
- `denoise.py`
    - `HiDreamO1DenoiseStep`
- `decoders.py`
    - `HiDreamO1DecodeStep`

This is the standard modular design and makes the code easier to follow, especially if we want to add more blocks in the future. See the [Flux 2 modular pipeline](https://github.com/huggingface/diffusers/tree/main/src/diffusers/modular_pipelines/flux2) for a concrete example.

src/diffusers/modular_pipelines/hidream_o1/modular_blocks_hidream_o1.py
```diff
@@ -0,0 +1,560 @@
...
+                for step in range(len(block_state.timesteps))
+            ]
+        else:
+            block_state.noise_scale_schedule = [block_state.noise_scale_start]
+
+        self.set_block_state(state, block_state)
+        return components, state
+
+
+class HiDreamO1DenoiseStep(ModularPipelineBlocks):
```
dg845, line 350: I think we should refactor this into several blocks, with the denoising loop part being implemented as a `LoopSequentialPipelineBlocks` subclass, following the standard denoise step modular design. Concretely, this could look something like:

1. `HiDreamO1DenoiseStep`: defines how the denoising blocks compose together. Subclasses the loop wrapper class below. See `Flux2DenoiseStep` for an example: https://github.com/huggingface/diffusers/blob/7aa746c4b07f07addcedeb59eaf43d8ee4174dd0/src/diffusers/modular_pipelines/flux2/denoise.py#L464-L466
2. `HiDreamO1DenoiseLoopWrapper`: inherits from `LoopSequentialPipelineBlocks`, handles the denoising loop and progress bar. See `Flux2DenoiseLoopWrapper` for an example: https://github.com/huggingface/diffusers/blob/7aa746c4b07f07addcedeb59eaf43d8ee4174dd0/src/diffusers/modular_pipelines/flux2/denoise.py#L406-L407
3. `HiDreamO1LoopDenoiser`: handles the loop step logic, including guidance (using a guider). See `Flux2LoopDenoiser` for an example: https://github.com/huggingface/diffusers/blob/7aa746c4b07f07addcedeb59eaf43d8ee4174dd0/src/diffusers/modular_pipelines/flux2/denoise.py#L45-L46
4. `HiDreamO1AfterDenoiser`: handles the scheduler `step` call. See `Flux2LoopAfterDenoiser` for an example: https://github.com/huggingface/diffusers/blob/7aa746c4b07f07addcedeb59eaf43d8ee4174dd0/src/diffusers/modular_pipelines/flux2/denoise.py#L362-L363
5. `HiDreamO1BeforeDenoiser` (if necessary): handles any preparation needed before the denoising loop. See `WanLoopBeforeDenoiser` for an example: https://github.com/huggingface/diffusers/blob/7aa746c4b07f07addcedeb59eaf43d8ee4174dd0/src/diffusers/modular_pipelines/wan/denoise.py#L37-L38

src/diffusers/modular_pipelines/hidream_o1/modular_pipeline.py
```diff
@@ -0,0 +1,77 @@
...
+# limitations under the License.
+
+import torch
+
+from ...configuration_utils import ConfigMixin, register_to_config
+from ..modular_pipeline import ModularPipeline
+from .utils import PATCH_SIZE
+
+
+class HiDreamO1Patchifier(ConfigMixin):
```
dg845, line 22: I think it is better to implement the `pack_image`/`unpack_image` methods inline where they are used (`HiDreamO1PrepareImageNoiseStep` and `HiDreamO1DecodeStep`) rather than defining a separate patchifier component.

src/diffusers/modular_pipelines/hidream_o1/utils.py
```diff
@@ -0,0 +1,262 @@
...
+from typing import Any, Optional
+
+import numpy as np
+import torch
+
+
+TIMESTEP_TOKEN_NUM = 1
+PATCH_SIZE = 32
+FULL_NOISE_SCALE = 8.0
+T_EPS = 0.001
```
dg845, line 25: I think we don't need a `PATCH_SIZE` constant as we can read the patch size from the transformer component via `transformer.config.patch_size`. For the other constants, I think we put them in the files where they are used or inline them if they are used as an `InputParam` (e.g. `FULL_NOISE_SCALE`).

src/diffusers/modular_pipelines/hidream_o1/modular_blocks_hidream_o1.py
```diff
@@ -0,0 +1,560 @@
...
+    def inputs(self) -> list[InputParam]:
+        return [
+            InputParam.template("height", required=True),
+            InputParam.template("width", required=True),
+            InputParam.template("generator"),
+            InputParam.template("latents"),
+            InputParam(
+                "noise_scale_start",
+                type_hint=float,
+                default=FULL_NOISE_SCALE,
```
dg845, line 231:
```suggestion
                default=8.0,
```
Inline `FULL_NOISE_SCALE` as suggested in https://github.com/huggingface/diffusers/pull/13749#discussion_r3286875788

src/diffusers/modular_pipelines/hidream_o1/utils.py
```diff
@@ -0,0 +1,262 @@
...
+    (2496, 1664),
+    (1664, 2496),
+    (3104, 1312),
+    (1312, 3104),
+    (2304, 1792),
+    (1792, 2304),
+]
+
+
+def find_closest_resolution(width: int, height: int) -> tuple[int, int]:
```
dg845, line 42: I think we should inline the helper functions here which are only used once (such as `find_closest_resolution`) so it's easier to follow the code.

dg845 (commented):
Thanks for the PR! Left an initial design review for the modular pipeline :).

bghira:
note that the contributor for this isn't from the original hidream team. asking them for this many back and forth changes, at some point, would be better if the diffusers team themselves handled it.

dg845:
Hi @chinoll, a gentle ping on this. I'm happy to help out with anything :). If you would prefer, the `diffusers` team could also take over work on this PR.

Issue #13696, opened 2026-05-08
[bug] The mask is not correctly sharded for QwenImageTransformer + Ulysses SP


zhtmike:
### Describe the bug

The mask is not correctly sharded in `QwenImageTransformer` when Ulysses SP is enabled.
Since the `QwenImageTransformer`'s mask is non-contiguous, it should be expanded in an interleaved manner before being fed into each SP rank.

The current implementation shows no clear visual difference with or without correct mask sharding. However, for RL training, we find that it causes abnormal losses and metrics.

This issue can also be validated by the following SP accuracy tests.

### Reproduction

The test should be failed since the forward output for SP and non-SP is not same.

```python
"""
Run with 2 GPUs:
    torchrun --nproc_per_node=2 --local-ranks-filter=0 test.py
"""

import os
from datetime import timedelta

import torch
import torch.distributed
from torch.distributed.device_mesh import init_device_mesh
from diffusers import ContextParallelConfig, QwenImageTransformer2DModel

_SP_SIZE = 2
_MODEL_CONFIG = {
    "patch_size": 2,
    "in_channels": 16,
    "out_channels": 4,
    "num_layers": 2,
    "attention_head_dim": 16,
    "num_attention_heads": 4,
    "joint_attention_dim": 16,
    "guidance_embeds": False,
    "axes_dims_rope": (8, 4, 4),
}


def main():
    torch.distributed.init_process_group("nccl", timeout=timedelta(seconds=36000))
    torch.cuda.set_device(int(os.environ["LOCAL_RANK"]))

    rank = torch.distributed.get_rank()
    mesh = init_device_mesh("cuda", mesh_shape=(1, 1, _SP_SIZE), mesh_dim_names=("dp", "ring", "ulysses"))

    module_sp = QwenImageTransformer2DModel(**_MODEL_CONFIG)
    module_sp.enable_parallelism(config=ContextParallelConfig(ulysses_degree=_SP_SIZE, mesh=mesh))
    module_sp = module_sp.to("cuda", dtype=torch.bfloat16)
    for p in module_sp.parameters():
        torch.distributed.broadcast(p.data, src=0)

    module_no_sp = QwenImageTransformer2DModel(**_MODEL_CONFIG)
    module_no_sp = module_no_sp.to("cuda", dtype=torch.bfloat16)
    module_no_sp.load_state_dict({k: v.clone() for k, v in module_sp.state_dict().items()}, strict=False)

[...]

PR #13756, opened 2026-05-15
Fix the QwenImage Attention mask under Ulysses SP

zhtmike:
# What does this PR do?
This fixes the issue #13696 . The test should be passed after this PR.

This the problem I found: The mask does not have a one-to-one correspondence with the content.

For QwenImage Pipeline, use the following example

```text
Position:  0  1  2  3  4  5  6  7 | 8  9  10 11
Content:  T0 T1 T2 T3 T4 T5 T6 T7 | I0 I1 I2 I3
Mask:      1  1  0  0  0  0  0  0 | 1  1  1  1
```

After CP shard (assume 2 ranks)
```text
Rank 0: text=[T0 T1 T2 T3],  image=[I0 I1]  → joint=[T0 T1 T2 T3 I0 I1]
Rank 1: text=[T4 T5 T6 T7],  image=[I2 I3]  → joint=[T4 T5 T6 T7 I2 I3]
```

After All-to-all
```text
Position:  0  1  2  3 | 4  5  | 6  7  8  9  | 10 11
Content:  T0 T1 T2 T3 | I0 I1 | T4 T5 T6 T7 | I2 I3
           ← rank 0 →          ← rank 1 →
```

But the mask is not handled correctly
```text
Position:  0  1  2  3  4  5  6  7 | 8  9  10 11
Mask:      1  1  0  0  0  0  0  0 | 1  1  1  1
```

This PR makes mask correctly assigned
```text
Position:  0  1  2  3 | 4  5 | 6  7  8  9 | 10 11
Mask:      1  1  0  0 | 1  1 | 0  0  0  0 | 1  1
```

Fixes # (issue)

## Before submitting

## Who can review?
@sayakpaul


zhtmike:
@sayakpaul, I can make this fix more generic by addressing `TemplatedUlyssesAttention` directly, which should help prevent similar errors in the future.

Here’s my suggestion:
- Maintain separate local masks for image and text.
- During all-to-all operations, ensure the mask and the corresponding content always have a one-to-one relationship.
This way, we avoid numerical inconsistencies caused by mask-related issues.

However, this approach requires handling masks in attention layer locally (and with extra communication cost) and may revert some of the performance improvements introduced in #12702.

So do you have any suggestions?

sayakpaul:
Could you explain a bit why the actual mask is

```bash
Position:  0  1  2  3 | 4  5 | 6  7  8  9 | 10 11
Mask:      1  1  0  0 | 1  1 | 0  0  0  0 | 1  1
```

zhtmike:
> Could you explain a bit why the actual mask is
> 
> ```shell
> Position:  0  1  2  3 | 4  5 | 6  7  8  9 | 10 11
> Mask:      1  1  0  0 | 1  1 | 0  0  0  0 | 1  1
> ```

since the original input is (T: token for text; I: Token for image)
```text
Position:  0  1  2  3  4  5  6  7 | 8  9  10 11
Content:  T0 T1 T2 T3 T4 T5 T6 T7 | I0 I1 I2 I3
Mask:      1  1  0  0  0  0  0  0 | 1  1  1  1
```

after all-to-all
```text
Position:  0  1  2  3 | 4  5  | 6  7  8  9  | 10 11
Content:  T0 T1 T2 T3 | I0 I1 | T4 T5 T6 T7 | I2 I3
           ← rank 0 →          ← rank 1 →
```

so the mask should be in the correct position matching the content, avoid the wrong attention of KV

```text
Position:  0  1  2  3 | 4  5  | 6  7  8  9  | 10 11
Content:  T0 T1 T2 T3 | I0 I1 | T4 T5 T6 T7 | I2 I3
Mask:      1  1  0  0 | 1  1  | 0  0  0  0  | 1  1
```

sayakpaul:
I need @DN6's thoughts on this one. Dhruv, should we tackle it from the forward pass of the model or delegate to the Ulysses utilities we have? Personally, I would prefer doing it from the Ulysses utilities as it would help to keep the forward clean.

sayakpaul:
@claude-2-serge could you do a review?

DN6:
> I need @DN6's thoughts on this one. Dhruv, should we tackle it from the forward pass of the model or delegate to the Ulysses utilities we have? Personally, I would prefer doing it from the Ulysses utilities as it would help to keep the forward clean.

Yeah, it would be better to handle via the CP utilities and keep the forward as clean as possible.

zhtmike:
Hi @sayakpaul and @DN6 , I've updated the PR using local mask handling.

This is what happens now:

Inputs (T for text token, I for image token):
```text
Position:  0  1  2  3  4  5  6  7 | 8  9  10 11
Content:  T0 T1 T2 T3 T4 T5 T6 T7 | I0 I1 I2 I3
Mask:      1  1  0  0  0  0  0  0 | 1  1  1  1
```

After CP shard (assume 2 ranks)
```text
Rank 0: text=[T0 T1 T2 T3],  image=[I0 I1], encoder mask=[1,1,0,0], image mask=[1,1] 
→ joint=[T0 T1 T2 T3 I0 I1], mask=[1,1,0,0,1,1]

Rank 1: text=[T4 T5 T6 T7],  image=[I2 I3], encoder mask=[0,0,0,0], image mask=[1,1]  
→ joint=[T4 T5 T6 T7 I2 I3], mask=[0,0,0,0,1,1]
```

After QKV All-to-all and mask all-gather
```text
Position:  0  1  2  3 | 4  5  | 6  7  8  9  | 10 11
Content:  T0 T1 T2 T3 | I0 I1 | T4 T5 T6 T7 | I2 I3
Mask:      1  1  0  0 | 1  1  |  0  0  0  0 | 1  1
```

The KV and mask should have the one-to-one correspondence now.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -497,6 +497,12 @@ def __call__(
         if encoder_hidden_states is None:
             raise ValueError("QwenDoubleStreamAttnProcessor2_0 requires encoder_hidden_states (text stream)")
 
+        if attention_mask is None and encoder_hidden_states_mask is not None:
```
sayakpaul: But we're not passing `attention_mask` later no?

src/diffusers/models/attention_dispatch.py
```diff
@@ -2200,6 +2200,11 @@ def forward(
         query, key, value = (_all_to_all_single(x, group) for x in (query, key, value))
         query, key, value = (x.flatten(0, 1).permute(1, 0, 2, 3).contiguous() for x in (query, key, value))
 
+        if attn_mask is not None and attn_mask.shape[-1] == S_KV_LOCAL:
+            mask_list = [torch.empty_like(attn_mask) for _ in range(world_size)]
+            dist.all_gather(mask_list, attn_mask, group=group)
+            attn_mask = torch.cat(mask_list, dim=-1)
```
sayakpaul, line 2207: An explanatory comment would be nice.

src/diffusers/models/transformers/transformer_qwenimage.py
zhtmike: Done. Drop `attention_mask` with a type check instead

src/diffusers/models/attention_dispatch.py
zhtmike: Added comments. I am aware that Ulysses Anything have the same problem with QwenImage, so I fixed that too.

sayakpaul:
@claude-2-serge could you do a review?

sayakpaul (commented):
Should there be a test?

zhtmike:
> Should there be a test?

Done. Added an accuracy test under `ContextParallelTesterMixin`. This test should guard this (main branch should fail).
And I've tested with other two models flux and flux2. So I think the threshold should be fine.

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -909,38 +922,27 @@ def forward(
 
         image_rotary_emb = self.pos_embed(img_shapes, max_txt_seq_len=text_seq_len, device=hidden_states.device)
 
-        # Construct joint attention mask once to avoid reconstructing in every block
-        # This eliminates 60 GPU syncs during training while maintaining torch.compile compatibility
-        block_attention_kwargs = attention_kwargs.copy() if attention_kwargs is not None else {}
```
yiyixuxu, line 916: cc @kashif here since this would revert the optimization as part of this PR #12702 

I tried to go over #12702, but was not able to find much detail about this optimization. I would love to understand more about the cause of the sync and performance delta, because the pre-built joint mask does not shard correctly under CP

zhtmike:
Hi @sayakpaul  @yiyixuxu , may I know if there is any concern for this PR?

sayakpaul:
We're waiting for @kashif for https://github.com/huggingface/diffusers/pull/13756#discussion_r3276526065

src/diffusers/models/transformers/transformer_qwenimage.py
kashif: thanks @yiyixuxu! the #12702 bit was just me building the joint mask once instead of per-block. "eliminates 60 GPU syncs" was a bad comment on my part, i checked and it's actually 0 syncs, just a plain cat/ones. cost of dropping it is ~0.85ms/fwd eager and basically 0 with compile, so no real loss.

and yeah it has to go for CP anyway: the pre-built mask is over the full unsharded seq so it can't line up after the all-to-all. confirmed with the #13696 repro, main is off by 2.9e-2 and this PR is exactly 0.0. lgtm from me 👍

kashif:
@zhtmike one heads up: the new correctness test actually passes on main as-is, so it won't catch the regression. if you bump it to batch_size=2 with per-sample padding like in #13696 (e.g. `mask[0, :2]=1`, `mask[1, :6]=1`) it fails on main (2.9e-2) and passes here (0.0). not blocking, the fix itself is correct.

zhtmike:
Hi @kashif , thank you for your valuable comments!

In my local testing environment, I just used the latest main with two files cherry-picked from the current PR.

- `tests/models/testing_utils/parallelism.py` 
- `tests/models/transformers/test_models_transformer_qwenimage.py`, 

and ran the following test:

```bash
pytest -rxXs tests/models/transformers/test_models_transformer_qwenimage.py::TestQwenImageTransformerContextParallel
``` 

And I got the following error

```python

============================= test session starts ==============================
platform linux -- Python 3.12.12, pytest-9.0.2, pluggy-1.6.0
rootdir: /scratch/fq9hpsac/mikecheung/gitlocal/diffusers
configfile: pyproject.toml
plugins: timeout-2.4.0, xdist-3.8.0, anyio-4.12.1, hydra-core-1.3.2, requests-mock-1.10.0
collected 12 items

tests/models/transformers/test_models_transformer_qwenimage.py .s.s.s.ss [ 75%]
.Fs                                                                      [100%]

_ TestQwenImageTransformerContextParallel.test_context_parallel_output_correctness[ulysses] _

self = <tests.models.transformers.test_models_transformer_qwenimage.TestQwenImageTransformerContextParallel object at 0x1550c59ab710>
cp_type = 'ulysses_degree', batch_size = 1

    @pytest.mark.parametrize("cp_type", ["ulysses_degree", "ring_degree"], ids=["ulysses", "ring"])
    def test_context_parallel_output_correctness(self, cp_type, batch_size: int = 1):
        """Verify that CP output is numerically identical to a single-GPU reference forward pass."""
        if not torch.distributed.is_available():
            pytest.skip("torch.distributed is not available.")
    
        if not hasattr(self.model_class, "_cp_plan") or self.model_class._cp_plan is None:
    
        if cp_type == "ring_degree":
            active_backend, _ = _AttentionBackendRegistry.get_active_backend()
            if active_backend == AttentionBackendName.NATIVE:
                pytest.skip("Ring attention is not supported with the native attention backend.")
    
        world_size = 2
        init_dict = self.get_init_dict()
        inputs_dict = self.get_dummy_inputs(batch_size=batch_size)
    
        # Single-GPU reference
        model = self.model_class(**init_dict).eval().to(torch_device)
        state_dict = {k: v.cpu() for k, v in model.state_dict().items()}
        with torch.no_grad():
            ref_output = model(**inputs_dict, return_dict=False)[0].cpu()
    
        # Context-parallel run with the same weights
        inputs_cpu = {k: v.cpu() if isinstance(v, torch.Tensor) else v for k, v in inputs_dict.items()}
        cp_dict = {cp_type: world_size}
    
        master_port = _find_free_port()
        manager = mp.Manager()
        return_dict = manager.dict()
    
        mp.spawn(
            _context_parallel_correctness_worker,
            args=(world_size, master_port, self.model_class, init_dict, state_dict, cp_dict, inputs_cpu, return_dict),
            nprocs=world_size,
            join=True,
        )
    
        assert return_dict.get("status") == "success", (
            f"Context parallel correctness check failed: {return_dict.get('error', 'Unknown error')}"
        )
    
        cp_output = torch.tensor(return_dict["output"])
>       torch.testing.assert_close(ref_output, cp_output, atol=1e-4, rtol=1e-4)
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 251 / 256 (98.0%)
E       Greatest absolute difference: 0.0255483016371727 at index (0, 9, 3) (up to 0.0001 allowed)
E       Greatest relative difference: 1.582349419593811 at index (0, 11, 3) (up to 0.0001 allowed)

tests/models/testing_utils/parallelism.py:461: AssertionError
```

So, in my testing environment, the newly added test correctly guards against the error in #13696. And I think the error should be irrelevant to the batch size, so I prefer to keep it simple with batch_size = 1.

Can you please take a look to see if I have missed something?

kashif (approved):

kashif:
no you are right! @zhtmike I was not in the main branch 🙈

zhtmike:
> We're waiting for @kashif for [#13756 (comment)](https://github.com/huggingface/diffusers/pull/13756#discussion_r3276526065)

Kindly ping @sayakpaul

sayakpaul:
We should fix the tests
https://github.com/huggingface/diffusers/actions/runs/26993270116/job/79657864315?pr=13756

zhtmike:
The two failed tests are related to `tests.models.transformers.test_models_transformer_chronoedit`. Seems irrelevant to this PR, should we fix it here?

sayakpaul:
Sorry I meant to tag https://github.com/huggingface/diffusers/actions/runs/26993270116/job/79657864295?pr=13756

zhtmike:
Updated. The error is due to the input check of `QwenDoubleStreamAttnProcessor2_0` introduced in this PR.

The `QwenImageControlNetModel` uses a similar pre-built mask optimization as `QwenImageTransformer2DModel`. Although it is not SP-supported (so there is no such bug for `QwenControlNetPipeline` currently), I prefer to make `QwenImageControlNetModel` follow the same style as `QwenImageTransformer2DModel`. Please check if this change is suitable.

src/diffusers/models/controlnets/controlnet_qwenimage.py
```diff
@@ -205,36 +205,27 @@ def forward(
         encoder_hidden_states = self.txt_norm(encoder_hidden_states)
         encoder_hidden_states = self.txt_in(encoder_hidden_states)
 
-        # Construct joint attention mask once to avoid reconstructing in every block
```
sayakpaul, line 208: It will be automatically constructed right?

sayakpaul: Should there be something similar to https://github.com/huggingface/diffusers/blob/f3d42be118f9af7ed9697b686fba09a8bdcd71d1/src/diffusers/models/transformers/transformer_qwenimage.py#L899?

tests/models/testing_utils/parallelism.py
```diff
@@ -227,6 +227,51 @@ def _custom_mesh_worker(
             dist.destroy_process_group()
 
 
+def _context_parallel_correctness_worker(
```
sayakpaul: Should it not be folded into https://github.com/huggingface/diffusers/blob/f3d42be118f9af7ed9697b686fba09a8bdcd71d1/tests/models/testing_utils/parallelism.py#L53?

sayakpaul (commented):
Thanks for your patience and immense amount of hardwork!

src/diffusers/models/controlnets/controlnet_qwenimage.py
zhtmike: > It will be automatically constructed right?

Yes, it is now constructed under QwenDoubleStreamAttnProcessor2_0 instead.

```diff
--- a/src/diffusers/models/transformers/transformer_qwenimage.py
+++ b/src/diffusers/models/transformers/transformer_qwenimage.py
@@ -497,6 +497,18 @@ class QwenDoubleStreamAttnProcessor2_0:
         if encoder_hidden_states is None:
             raise ValueError("QwenDoubleStreamAttnProcessor2_0 requires encoder_hidden_states (text stream)")
 
+        if attention_mask is not None:
+            raise ValueError(
+                "QwenDoubleStreamAttnProcessor2_0 does not accept an external attention_mask. "
+                "Pass encoder_hidden_states_mask to let the processor build the joint mask."
+            )
+
+        if encoder_hidden_states_mask is not None:
+            seq_img = hidden_states.shape[1]
+            image_mask = torch.ones((hidden_states.shape[0], seq_img), dtype=torch.bool, device=hidden_states.device)
+            attention_mask = torch.cat([encoder_hidden_states_mask, image_mask], dim=1)
+            attention_mask = attention_mask[:, None, None, :]
```


> Should there be something similar to
> 
> https://github.com/huggingface/diffusers/blob/f3d42be118f9af7ed9697b686fba09a8bdcd71d1/src/diffusers/models/transformers/transformer_qwenimage.py#L899
> 
> ?

Yes, it is already there
https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/controlnets/controlnet_qwenimage.py#L198

tests/models/testing_utils/parallelism.py
zhtmike: Done. It indeed has a lot of duplicated code; refactored it into a single one.

src/diffusers/models/controlnets/controlnet_qwenimage.py
kashif: addressed in `fix controlnet` - the controlnet now passes `encoder_hidden_states_mask` through to the blocks like the transformer, so the build-joint-mask-once is gone. CP tests pass.

tests/models/testing_utils/parallelism.py
kashif: done in `refactor worker` - the correctness path now reuses `_context_parallel_worker` (it optionally returns the output) instead of a separate worker.

tests/models/testing_utils/parallelism.py
```diff
@@ -369,6 +383,62 @@ def test_context_parallel_custom_mesh(self, cp_type, mesh_shape, mesh_dim_names)
             f"Custom mesh context parallel inference failed: {return_dict.get('error', 'Unknown error')}"
         )
 
+    @pytest.mark.parametrize("cp_type", ["ulysses_degree", "ring_degree"], ids=["ulysses", "ring"])
+    def test_context_parallel_output_correctness(self, cp_type, batch_size: int = 1):
```
sayakpaul: Maybe I didn't make myself clear but why can't we just correctness validation to 
https://github.com/huggingface/diffusers/blob/f3d42be118f9af7ed9697b686fba09a8bdcd71d1/tests/models/testing_utils/parallelism.py#L234 ?

zhtmike: Done! Originally, I thought a two-level approach was more suitable, since an accurate test is a stricter guard than merely runnable. Now it has been merged into a single test.

kashif:
do we want to add the same all CP modes, including ring?

zhtmike:
> do we want to add the same all CP modes, including ring?

Hi @kashif and @sayakpaul, I think a general test case for the precision of CP is still necessary.

Currently, the change of `ContextParallelTesterMixin` in this PR covers the accuracy tests for Ulysses SP only (ring attention was skipped since `ContextParallelTesterMixin` only covers the `SDPA` backend by default, and the `SDPA` backend does not support ring attention).

So for now, the modifications in tests only cover all Ulysses‑SP‑supported (and test‑guarded) models, including `QwenImage`, `Flux`, and `Flux2`, as they have the `ContextParallelTesterMixin` test suites. I have run the tests locally, and the newly added tests passed for these three models. So I think the test should be fine.

For ring attention, I think we should add a similar test under `ContextParallelAttentionBackendsTesterMixin`, but that should be outside the scope of this PR.

May I know if there are any other concerns?

sayakpaul:
> For ring attention, I think we should add a similar test under ContextParallelAttentionBackendsTesterMixin, but that should be outside the scope of this PR.

> May I know if there are any other concerns?

Yeah good idea. Maybe a future PR?

zhtmike:
> > For ring attention, I think we should add a similar test under ContextParallelAttentionBackendsTesterMixin, but that should be outside the scope of this PR.
> 
> > May I know if there are any other concerns?
> 
> Yeah good idea. Maybe a future PR?

sure I can help to do this~

src/diffusers/models/transformers/transformer_qwenimage.py
```diff
@@ -497,6 +497,18 @@ def __call__(
...
+                "QwenDoubleStreamAttnProcessor2_0 does not accept an external attention_mask. "
+                "Pass encoder_hidden_states_mask to let the processor build the joint mask."
+            )
+
+        if encoder_hidden_states_mask is not None:
+            seq_img = hidden_states.shape[1]
+            image_mask = torch.ones((hidden_states.shape[0], seq_img), dtype=torch.bool, device=hidden_states.device)
+            attention_mask = torch.cat([encoder_hidden_states_mask, image_mask], dim=1)
+            attention_mask = attention_mask[:, None, None, :]
+
```
sayakpaul, line 511: Should the controlnet also have similar changes or not because it doesn't define `_cp_plan`?

sayakpaul (approved):
Great work @zhtmike! I will merge after running the tests locally myself :)

src/diffusers/models/transformers/transformer_qwenimage.py
zhtmike: the controlnet model is calling `QwenDoubleStreamAttnProcessor2_0` here

sayakpaul: Could you elaborate that?

zhtmike: The controlnet model itself does not have the CP plan. However, because we have modified `QwenDoubleStreamAttnProcessor2_0` in this PR with an extra guard:

```python
if attention_mask is not None:
    raise ValueError(
        "QwenDoubleStreamAttnProcessor2_0 does not accept an external attention_mask. "
        "Pass encoder_hidden_states_mask to let the processor build the joint mask."
    )
```

So we either need to:
- modify the controlnet following the QwenImage transformer’s change; or
- drop the guard and avoid touching the controlnet.

In the future, we may need to add support for the SP implementation for controlnet. In that case, option 1 may be a better solution, since it will avoid such mask bugs more easily and is also consistent with the style of the Qwen‑Image transformer. So personally, I prefer option 1.

sayakpaul:
@bot /style

sayakpaul:
I ran `pytest tests/models/transformers/test_models_transformer_qwenimage.py -k "context_parallel_inference"` on your branch and `main`. Both are passing. Is this not the right way to test?

zhtmike:
> I ran `pytest tests/models/transformers/test_models_transformer_qwenimage.py -k "context_parallel_inference"` on your branch and `main`. Both are passing. Is this not the right way to test?

you might need to pick `tests/models/transformers/test_models_transformer_qwenimage.py` in the branch. Since in the main branch the testing mask is all true, it will not reflect the real senerio.

PR #13758, opened 2026-05-15
[WIP] chore: add utilities to check if call/forward methods are documented.

sayakpaul:
# What does this PR do?

To have some much-needed consistency in our docstrings where it should matter.

The diff seems large because of the docstring stuff. Main stuff is under `.github/workflows/`, `utils/check_forward_call_docstrings.py`, and `Makefile`.


stevhliu (commented):
very nice!

should we run the checker locally and see what pops up and include the cleanups here before landing this pr?

sayakpaul:
> should we run the checker locally and see what pops up and include the cleanups here before landing this pr?

We can actually run `python utils/check_forward_call_docstrings.py` locally to get a list of what needs fixing. The CI is also reporting those. 

Do we want to fix them in this PR?

stevhliu:
oh i see it now, wow this is really handy! i'm ok with either fixing here or in separate pr :)

sayakpaul:
> i'm ok with either fixing here or in separate pr :)

Do you maybe want to branch off of this PR branch and open a PR to this one with the fixes?

yiyixuxu (approved):
thanks,
not merge-blocking, but I think, it could make a better developer experience if we also add a placeholder entry for missing arg & remove the ones not exist

yiyixuxu:
(testing , ignore us....)

@claude can you open a PR to add an option "--fix" that:
1. add a placeholder for missing argument
2. remove the non-existing argument

COMMIT THIS

PR #13767, opened 2026-05-19
ci: use uv overrides to make sure tokenizers install from <=0.23.0 under subs

sayakpaul:
Refer to https://huggingface.slack.com/archives/C02V5EA0A95/p1779187819850039?thread_ts=1778844656.429709&cid=C02V5EA0A95


sayakpaul:
Failing tests are unrelated. 

Cc: @hanouticelina @Wauplin could you review the PR and see if the `prerelease` substitution (typically done during CI testing for HFH releases) would comply?

Wauplin:
we can check on next release this week yes :+1: thanks for looking into it

sayakpaul:
I think I can check if the substitution of the HFH CI bot has any disruptive effect on this PR itself. If you could share what the installation would be to substitute, it would be helpful.

Wauplin (approved):
I've tested this PR for the 1.16.0rc0 prerelease and it seems to work fine!

- branch: https://github.com/huggingface/diffusers/compare/main...ci-test-huggingface-hub-1.16.0.rc0-release
- workflow: https://github.com/huggingface/diffusers/actions/runs/262246119743

Got 2 failures but the tests did ran which is the goal of this PR (before it was stopping before the tests, at the install step)

```
FAILED tests/pipelines/z_image/test_z_image_img2img.py::ZImageImg2ImgPipelineFastTests::test_pipeline_with_accelerator_device_map - AssertionError: np.float32(nan) not less than 0.0005
```

```
E               AssertionError: False is not true : No param should be on 'meta' device.
tests/lora/utils.py:420: AssertionError
cat: reports/tests_models_lora_peft_main_failures_short.txt: No such file or directory
```

I'm therefore approving this PR which should help seamlessly testing hfh prereleases on diffusers in the future! ✔️

sayakpaul:
Thanks a lot sir! Those failures are transient and unrelated.

PR #13768, opened 2026-05-20
schedulers/scheduling_dpmsolver_multistep: round timesteps for exponential and beta sigma schedules

Dev-X25874:
# What does this PR do?

In `DPMSolverMultistepScheduler.set_timesteps`, the `use_karras_sigmas` and
`use_lu_lambdas` branches both call `timesteps.round()` after computing
timesteps via `_sigma_to_t` (which returns floating-point values through
interpolation), before those timesteps are cast to `torch.int64`.

The `use_exponential_sigmas` and `use_beta_sigmas` branches were missing this
`.round()` call, causing raw float values to be **truncated** (not rounded) on
the `torch.int64` cast at the end of `set_timesteps`. This silently introduces
a systematic off-by-one error in the noise schedule index lookup whenever
either of these two sigma modes is used with this scheduler.

This PR adds `timesteps = timesteps.round()` to both branches, consistent with
the existing pattern already established by `use_karras_sigmas` and
`use_lu_lambdas`.

Fixes # (issue)

## Before submitting

## Who can review?

@DN6

### DIFF
--- a/src/diffusers/schedulers/scheduling_dpmsolver_multistep.py
+++ b/src/diffusers/schedulers/scheduling_dpmsolver_multistep.py
@@ -455,10 +455,12 @@ class DPMSolverMultistepScheduler(SchedulerMixin, ConfigMixin):
             sigmas = np.flip(sigmas).copy()
             sigmas = self._convert_to_exponential(in_sigmas=sigmas, num_inference_steps=num_inference_steps)
             timesteps = np.array([self._sigma_to_t(sigma, log_sigmas) for sigma in sigmas])
+            timesteps = timesteps.round()
         elif self.config.use_beta_sigmas:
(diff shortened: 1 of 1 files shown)


Dev-X25874:
Hey @DN6 — could you please take a look at this PR when you get a chance? It fixes a silent off-by-one error in `DPMSolverMultistepScheduler.set_timesteps` where the `use_exponential_sigmas` and `use_beta_sigmas` branches were missing a `.round()` call before the `torch.int64` cast — causing float truncation instead of rounding. The `use_karras_sigmas` and `use_lu_lambdas` branches already do this correctly; this PR just brings the other two in line. Two-line fix, no behaviour change for any other code path. Thanks!

PR #13771, opened 2026-05-20
[docs] remove pipeline examples section

stevhliu:
removes the pipeline examples section and migrates their content over to the model docs (diffedit and tcd-lora didn't make it because they're deprecated 🪦). this makes it easier to maintain (no duplication) and removes an arbitrary/inconsistent grouping (avoid signaling some pipelines are more than others).


sayakpaul (approved):
Super cleanup!

PR #13570, opened 2026-04-28
[agents docs] update pipelines.md: 

yiyixuxu:
adding a little bit tip on how we structure the pipeline

PR #13774, opened 2026-05-20
[CI] claude_review: target source PR's branch for follow-up PRs

yiyixuxu:
try to make it so claude CI can open PRs (against source branch ) https://github.com/huggingface/diffusers/pull/13570#issuecomment-4502923753

### DIFF
--- a/.github/workflows/claude_review.yml
+++ b/.github/workflows/claude_review.yml
@@ -156,7 +156,6 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ github.event.issue.number || github.event.pull_request.number }}
           COMMENT_USER: ${{ github.event.comment.user.login }}
-          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


yiyixuxu:
@sayakpaul 

I'm going to go ahead and merge this one now (and likely follow-up ones if it does not work) since there is no easier way to test out this workflow. All my changes are limited to the "claude open follow-up PR step" of this file (which is not currently working anyway https://github.com/huggingface/diffusers/pull/13570#issuecomment-4502923753 

I disabled "COMMIT THIS" on the forked branch for now, so we can quickly get it to work and test it out on our own branches. For forks, I think maybe the cleanest way is to open an issue from the review comments, and at some point it might be nice to have CLAUDE CI come work on issues too :)

feel free to push more changes on top of mine, i just want to quickly get a working prototype

sayakpaul:
Yes totally fine. I am going to work on:

> For forks, I think maybe the cleanest way is to open an issue from the review comments, and at some point it might be nice to have CLAUDE CI come work on issues too :)

I will also look after what you suggested in https://github.com/huggingface/diffusers/pull/13758/#pullrequestreview-4331961470

joangava (commented):
PR_INFO=$(gh pr view "$PR_NUMBER" --json headRefName,isCrossRepository)
PR_BRANCH=$(echo "$PR_INFO" | jq -r '.headRefName')
IS_FORK=$(echo "$PR_INFO" | jq -r '.isCrossRepository')

PR #13776, opened 2026-05-21
Reduce FLUX int8 test peak memory with sequential offload

jiqing-feng:
## Summary

Update the slow FLUX bitsandbytes int8 tests to use sequential CPU offload instead of model CPU offload.

`enable_model_cpu_offload()` can move an entire sub-model onto the GPU at once. For `black-forest-labs/FLUX.1-dev`, this can OOM on <=24 GB cards even when the T5 encoder and transformer are loaded from the pre-quantized int8 test checkpoint. Sequential CPU offload keeps peak memory lower by materializing one layer at a time, which lets the int8 FLUX tests run in more constrained environments.

The LoRA-loading assertion tolerance is also relaxed from `1e-3` to `2e-3` to account for small backend-specific numerical differences observed in the slow int8 path.

## Changes

- Switch `SlowBnb8bitFluxTests` setup from `enable_model_cpu_offload()` to `enable_sequential_cpu_offload()`.
- Document why sequential offload is needed for the FLUX int8 slow tests.
- Relax the `test_lora_loading` cosine-distance tolerance to `2e-3`.

## Validation

Run the affected slow tests:

```bash
RUN_SLOW=1 python -m pytest \
  tests/quantization/bnb/test_mixed_int8.py::SlowBnb8bitFluxTests::test_quality \
  tests/quantization/bnb/test_mixed_int8.py::SlowBnb8bitFluxTests::test_lora_loading \
  -x -s
```

Observed result:

```text
2 passed
```


jiqing-feng:
require change: https://github.com/huggingface/accelerate/pull/4044 merged.

jiqing-feng:
Hi @sayakpaul . Would you please review the PR? Thanks!

tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -663,7 +663,11 @@ def setUp(self) -> None:
             transformer=transformer_8bit,
             torch_dtype=torch.float16,
         )
-        self.pipeline_8bit.enable_model_cpu_offload()
+        # Use sequential CPU offload to keep peak GPU memory minimal (one layer at a time).
+        # enable_model_cpu_offload moves an entire sub-model to GPU at once, which OOMs on
+        # <=24 GB cards for FLUX.1-dev even with int8 quantization.
+        # This requires the bitsandbytes fix that preserves Int8Params.SCB across .to() calls.
+        self.pipeline_8bit.enable_sequential_cpu_offload()
```
sayakpaul: Why do we keep making the same kind of changes i.e., if something fails on your particular environment, it's always better to guard them accordingly rather than doing it in a straightforward way like this.

jiqing-feng: Thanks for the feedback! Updated to guard by device memory instead of unconditionally switching:

```python
_, total_mem = torch.accelerator.get_memory_info(0)
if total_mem <= 25 * (1024**3):
    self.pipeline_8bit.enable_sequential_cpu_offload()
else:
    self.pipeline_8bit.enable_model_cpu_offload()
```

This keeps the original `enable_model_cpu_offload` path on large-memory devices and only falls back to sequential offload on ≤24 GB cards. `torch.accelerator` works across CUDA/XPU/ROCm.

jiqing-feng:
Hi @sayakpaul . I have fixed your comment, please review the new change. Thanks!

jiqing-feng:
Hi @sayakpaul . Would you please review the PR? Thanks!

jiqing-feng:

jiqing-feng:
Hi @sayakpaul . In case you forgot this PR. Please review it when you have bandwidth. Thanks.

tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -663,7 +663,14 @@ def setUp(self) -> None:
             transformer=transformer_8bit,
             torch_dtype=torch.float16,
         )
-        self.pipeline_8bit.enable_model_cpu_offload()
+        # On devices with <= 24 GB VRAM, enable_model_cpu_offload can OOM because it
+        # moves an entire sub-model to the accelerator at once.  Fall back to
+        # sequential (per-layer) CPU offload in that case.
+        _, total_mem = torch.accelerator.get_memory_info(0)
```
sywangyi: torch.accelerator.get_memory_info() instead of  torch.accelerator.get_memory_info(0)？

tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -663,7 +663,14 @@ def setUp(self) -> None:
...
+        if total_mem <= 25 * (1024**3):
```
sywangyi: not align with comment，24GB in comment

tests/quantization/bnb/test_mixed_int8.py
jiqing-feng: fixed.

tests/quantization/bnb/test_mixed_int8.py
jiqing-feng: fixed.

tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -709,7 +716,7 @@ def test_lora_loading(self):
         expected_slice = np.array([0.3916, 0.3916, 0.3887, 0.4243, 0.4155, 0.4233, 0.4570, 0.4531, 0.4248])
 
         max_diff = numpy_cosine_similarity_distance(expected_slice, out_slice)
-        self.assertTrue(max_diff < 1e-3)
+        self.assertTrue(max_diff < 2e-3)
```
IlyasMoutawwakil: why did this increase ?

IlyasMoutawwakil:
please resolve conflicts

tests/quantization/bnb/test_mixed_int8.py
jiqing-feng: The expected slice was recorded on one backend, but the test runs on CUDA/XPU/ROCm too. bnb int8 matmul differs slightly per backend, so the distance goes just over 1e-3 while the image is basically the same. 2e-3 is still tight enough.

tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -605,7 +605,14 @@ def _setup_slow_flux(self):
...
+        # On devices with <= 24 GB VRAM, enable_model_cpu_offload can OOM because it
+        # moves an entire sub-model to the accelerator at once.  Fall back to
+        # sequential (per-layer) CPU offload in that case.
+        _, total_mem = torch.accelerator.get_memory_info()
+        if total_mem <= 24 * (1024**3):
+            self.pipeline_8bit.enable_sequential_cpu_offload()
+        else:
+            self.pipeline_8bit.enable_model_cpu_offload()
         yield
         del self.pipeline_8bit
```
sayakpaul: We could also use `require_big_accelerator` to denote that this test won't run on those devices. I think it's better that way.

tests/quantization/bnb/test_mixed_int8.py
IlyasMoutawwakil: would loosen it up only for the backend that requires loosening, and add a comment explaining the difference in numerics stability please, worth documenting.

IlyasMoutawwakil (approved):
LGTM

tests/quantization/bnb/test_mixed_int8.py
jiqing-feng: done.

sayakpaul (approved):

PR #13782, opened 2026-05-21
Fix `EMAModel.restore()` foreach path crashing with device mismatch when model is on GPU

Dev-X25874:
# What does this PR do?

Fixes a runtime crash in `EMAModel.restore()` when `foreach=True` and the model lives on a non-CPU device (e.g. CUDA).

`store()` always saves parameters to CPU (`param.detach().cpu().clone()`). The `foreach` path in `restore()` then passed those raw CPU tensors directly to `torch._foreach_copy_()`, which requires all tensors to be on the same device:

```python
# before (broken on GPU)
torch._foreach_copy_(
    [param.data for param in parameters],
    [c_param.data for c_param in self.temp_stored_params],  # always CPU
)
```

This raises `RuntimeError: Expected all tensors to be on same device` for any user who calls the standard EMA validation pattern (`store → copy_to → restore`) with `foreach=True` on a GPU machine.

The fix mirrors the pattern already used correctly in `copy_to()`'s foreach path (line 780), which moves each shadow param to the target device before the copy:

```python
# after (matches copy_to() pattern)
torch._foreach_copy_(
    [param.data for param in parameters],
    [c_param.to(param.device).data for c_param, param in zip(self.temp_stored_params, parameters)],
)
```

Also adds `test_store_restore` to both `EMAModelTests` and `EMAModelTestsForeach` — the store/restore round-trip was completely untested prior to this PR.

## Before submitting
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/diffusers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/diffusers/tree/main/docs#writing-source-documentation).

## Who can review?

@sayakpaul

### DIFF
--- a/tests/others/test_ema.py
+++ b/tests/others/test_ema.py
@@ -180,28 +180,6 @@ class EMAModelTests(unittest.TestCase):
         assert torch.allclose(output, output_loaded, atol=1e-4)
 
     def test_store_restore(self):
-        # The foreach restore() path passed raw CPU tensors from store() to
-        # torch._foreach_copy_(), which requires same-device tensors and therefore
-        # crashes on GPU.  Fix: mirror copy_to()'s pattern of moving each stored
-        # tensor to param.device before the foreach copy.
-        unet, ema_unet = self.get_models()
-        original_params = [p.data.clone() for p in unet.parameters()]
-
-        # Simulate one EMA step so shadow params differ from model params.
-        unet = self.simulate_backprop(unet)
-        ema_unet.step(unet.parameters())
-
-        # Standard EMA validation pattern: store → copy_to → restore.
-        ema_unet.store(unet.parameters())
-        ema_unet.copy_to(unet.parameters())
-        ema_unet.restore(unet.parameters())
-
-        # After restore(), model weights must equal the pre-copy_to values.
-        for restored, original in zip(unet.parameters(), original_params):
-            assert torch.allclose(restored.data, original.to(restored.device), atol=1e-6), (
-                "restore() foreach path did not correctly recover the stored parameters"
-            )
-    def test_store_restore(self):
         # store() saves params to CPU; restore() must move them back to the model's
         # device before copying.  The non-foreach path uses copy_() which is
         # cross-device safe; this test guards that restore() actually round-trips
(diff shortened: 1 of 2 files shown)


sayakpaul:
How can I minimally reproduce the bug?

Dev-X25874:
Hi @sayakpaul, here's a minimal repro (requires a CUDA GPU):

```python
import torch
import torch.nn as nn
from diffusers.training_utils import EMAModel

model = nn.Linear(4, 4).cuda()
ema = EMAModel(model.parameters(), foreach=True)

# Simulate a training step so shadow params differ from model params
with torch.no_grad():
    for p in model.parameters():
        p.add_(torch.randn_like(p))
ema.step(model.parameters())

# Standard EMA validation pattern
ema.store(model.parameters())     # saves to CPU
ema.copy_to(model.parameters())  # works fine
ema.restore(model.parameters())  # RuntimeError: Expected all tensors to be on same device
```

The crash happens because `store()` always clones params to CPU, but the `foreach` path in `restore()` feeds those raw CPU tensors into `torch._foreach_copy_()` alongside the GPU model params. The non-foreach path is unaffected since `copy_()` handles cross-device copies natively.

tests/others/test_ema.py
```diff
@@ -179,6 +179,51 @@ def test_serialization(self):
...
+        ema_unet.store(unet.parameters())
+        ema_unet.copy_to(unet.parameters())
+        ema_unet.restore(unet.parameters())
+
+        # After restore(), model weights must equal the pre-copy_to values.
+        for restored, original in zip(unet.parameters(), original_params):
+            assert torch.allclose(restored.data, original.to(restored.device), atol=1e-6), (
+                "restore() foreach path did not correctly recover the stored parameters"
+            )
+    def test_store_restore(self):
```
sayakpaul, line 182: This seems duplicated.

sayakpaul:
How come this works?

```py
import torch

gpu = [torch.zeros(3, device="cuda")]
cpu = [torch.arange(3, dtype=torch.float32)]
torch._foreach_copy_(gpu, cpu)
print(gpu[0])
```

Prints:

```bash
tensor([0., 1., 2.], device='cuda:0')
```

tests/others/test_ema.py
Dev-X25874: My bad, the foreach test_store_restore was mistakenly placed inside EMAModelTests instead of EMAModelTestsForeach. Fixed now.

Dev-X25874:
Sorry for the noise, converting this to draft while I investigate further.

Dev-X25874:
Verified on PyTorch 2.10.0+cu128 — no crash. You're right, `torch._foreach_copy_()` handles cross-device fine on modern PyTorch. Closing this PR. Sorry for the noise.

PR #13791, opened 2026-05-22
fix(modeling_utils): rename shadowed `tuple` loop var in get_parameter_dtype (#13789)

### LINKED ISSUE #13789: UNet2DModel dtype property fails under nn.DataParallel with UnboundLocalError in get_parameter_dtype
## Description

`UNet2DModel` fails under `torch.nn.DataParallel` during the forward pass when it accesses `self.dtype` inside `UNet2DModel.forward`.

The error appears to come from `diffusers.models.modeling_utils.get_parameter_dtype`, where the nested function annotation uses `tuple`:

```python
def find_tensor_attributes(module: nn.Module) -> list[tuple[str, Tensor]]:

### Reproduction

import torch
from diffusers import UNet2DModel

print("torch:", torch.__version__)

import diffusers
print("diffusers:", diffusers.__version__)

model = UNet2DModel(
    sample_size=128,
    in_channels=10,
    out_channels=5,
    layers_per_block=2,
    block_out_channels=(64, 128, 256, 256),
    down_block_types=(
        "DownBlock2D",
        "DownBlock2D",
        "AttnDownBlock2D",
        "DownBlock2D",
    ),
    up_block_types=(
        "UpBlock2D",
        "AttnUpBlock2D",
        "UpBlock2D",
        "UpBlock2D",
    ),
    norm_num_groups=8,
)

model = model.cuda()
model = torch.nn.DataParallel(model, device_ids=[0, 1])
model.eval()

x = torch.randn(8, 10, 128, 128, device="cuda")
t = torch.rand(8, device="cuda") * 1000

with torch.no_grad():
    y = model(x, t, return_dict=False)[0]

print(y.shape)

### Logs

```shell
File ".../diffusers/models/unets/unet_2d.py", line ..., in forward
    t_emb = t_emb.to(dtype=self.dtype)

File ".../diffusers/models/modeling_utils.py", line ..., in dtype
    return get_parameter_dtype(self)

File ".../diffusers/models/modeling_utils.py", line ..., in get_parameter_dtype
    def find_tensor_attributes(module: nn.Module) -> list[tuple[str, Tensor]]:

UnboundLocalError: cannot access local variable 'tuple' where it is not associated with a value
```

### System Info

Python 3.11.13, torch 2.9.1+cu128, diffusers 0.38.0, CUDA Version: 12.9

### Who can help?

_No response_

Anai-Guo:
## What

Fixes #13789.

Inside `get_parameter_dtype` (`src/diffusers/models/modeling_utils.py`), the
inner `find_tensor_attributes` fallback used for `nn.DataParallel`
compatibility is annotated with `list[tuple[str, Tensor]]`. The enclosing
function also did:

```python
for tuple in gen:
    last_tuple = tuple
    if tuple[1].is_floating_point():
        return tuple[1].dtype
```

Assigning to `tuple` anywhere in `get_parameter_dtype` makes it a function-
local for the entire scope. When execution reaches the DataParallel fallback
before the `for` loop has ever bound `tuple` — which is exactly what happens
under `nn.DataParallel` where the replicated module on the non-master device
has empty `named_parameters()` / `named_buffers()` — the inner function
annotation evaluates `tuple[str, Tensor]` and raises:

```
UnboundLocalError: cannot access local variable tuple where it is not associated with a value
```

This breaks `UNet2DModel(...).cuda(); torch.nn.DataParallel(model, ...)`
forward, because `unet_2d.UNet2DModel.forward` does `t_emb.to(dtype=self.dtype)`.

## Fix

Rename the loop variable to `named_tensor` so the builtin `tuple` is no
longer shadowed. `get_parameter_device` already uses a distinct name
(`first_tuple`) and is not affected.

Minimal diff in `get_parameter_dtype`:

```diff
     gen = parameter._named_members(get_members_fn=find_tensor_attributes)
     last_tuple = None
-    for tuple in gen:
-        last_tuple = tuple
-        if tuple[1].is_floating_point():
-            return tuple[1].dtype
+    for named_tensor in gen:
+        last_tuple = named_tensor
+        if named_tensor[1].is_floating_point():
+            return named_tensor[1].dtype
```

## Repro

The repro from #13789 (two-GPU `nn.DataParallel(UNet2DModel(...))`) raises
`UnboundLocalError` on `main` and runs to completion with this patch.

---

AI-assisted, human reviewed.

### DIFF
--- a/src/diffusers/models/modeling_utils.py
+++ b/src/diffusers/models/modeling_utils.py
@@ -199,10 +199,10 @@ def get_parameter_dtype(parameter: torch.nn.Module) -> torch.dtype:
 
     gen = parameter._named_members(get_members_fn=find_tensor_attributes)
     last_tuple = None
-    for tuple in gen:
-        last_tuple = tuple
-        if tuple[1].is_floating_point():
-            return tuple[1].dtype
+    for named_tensor in gen:
+        last_tuple = named_tensor
+        if named_tensor[1].is_floating_point():
+            return named_tensor[1].dtype
 
     if last_tuple is not None:
         # fallback to the last dtype


qgallouedec:

PR #13796, opened 2026-05-23
[SDXL] Fix clip_skip not applied to negative prompt embeddings in encode_prompt

Dev-X25874:
# What does this PR do?

In `StableDiffusionXLPipeline.encode_prompt`, the positive prompt correctly
respects `clip_skip` when selecting the hidden layer:

```python
if clip_skip is None:
    prompt_embeds = prompt_embeds.hidden_states[-2]
else:
    prompt_embeds = prompt_embeds.hidden_states[-(clip_skip + 2)]
```

But the negative prompt path (line 473) always hardcodes `hidden_states[-2]`
regardless of `clip_skip`:

```python
negative_prompt_embeds = negative_prompt_embeds.hidden_states[-2]  # clip_skip ignored
```

This means when a user passes `clip_skip=1` (or any non-None value), the
positive and negative embeddings are extracted from **different CLIP layers**,
which corrupts classifier-free guidance since CFG subtracts the two vectors.

This PR fixes the negative prompt path to mirror the positive prompt logic
exactly, and adds a test that calls `encode_prompt` with `clip_skip=1` and
asserts the negative embeddings change accordingly.

## Before submitting

## Who can review?
@yiyixuxu @asomoza

### DIFF
--- a/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
+++ b/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py
@@ -470,7 +470,11 @@ class StableDiffusionXLPipeline(
                 # We are only ALWAYS interested in the pooled output of the final text encoder
                 if negative_pooled_prompt_embeds is None and negative_prompt_embeds[0].ndim == 2:
                     negative_pooled_prompt_embeds = negative_prompt_embeds[0]
-                negative_prompt_embeds = negative_prompt_embeds.hidden_states[-2]
+                if clip_skip is None:
+                    negative_prompt_embeds = negative_prompt_embeds.hidden_states[-2]
+                else:
+                    # "2" because SDXL always indexes from the penultimate layer.
+                    negative_prompt_embeds = negative_prompt_embeds.hidden_states[-(clip_skip + 2)]
 
                 negative_prompt_embeds_list.append(negative_prompt_embeds)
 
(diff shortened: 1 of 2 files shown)


bghira:
you should AT LEAST provide **examples**

Dev-X25874:
@bghira Here's an example showing the bug and fix:
With clip_skip=1, before this fix the negative embeddings were always pulled from hidden_states[-2] while positive used hidden_states[-3], causing a layer mismatch during CFG. After the fix both sides use hidden_states[-3] consistently.
The added test test_encode_prompt_clip_skip_applied_to_negative directly verifies this — it asserts that neg_no_skip != neg_skip when clip_skip changes.

bghira:
please less AI slop comment replies, by example i mean generate images before and after with a clip skip specific model that will show the effect properly

Dev-X25874:
@bghira Here are the before/after outputs side by side, clip_skip=2, seed 42, SDXL base 1.0, same prompt both sides. Shadow artifact visible on the right is CFG corruption from the layer mismatch — gone on the left after the fix.

comparison

bghira:
but clip_skip 2 is already applied by the vanilla diffusers SDXL code. when you say 2, you mean 4?

bghira:
also, SDXL uses zeroes for the negative embed, unlike SD1x which uses an encoded empty string. for most use cases that go for the defaults (no negative prompt) it would not have been noticed. you should try empty string as well for comparison.

bghira:
just curious are you using OpenClaw for all of this? your responses all look directly copy-pasted from the language model, even with the em-dashes and LLM-speak, "cleaner demo".

bghira:
yeah...

image

bghira:
@sayakpaul what's the Diffusers' project policy on autonomous agents doing things like this?

PR #13797, opened 2026-05-23
[DreamBooth SDXL LoRA] Apply SNR min-weighting to prior preservation loss when snr_gamma is set

Dev-X25874:
# What does this PR do?

In `examples/dreambooth/train_dreambooth_lora_sdxl.py`, when training with
both `--snr_gamma` and `--with_prior_preservation`, the Min-SNR loss
weighting (Section 3.4 of https://huggingface.co/papers/2303.09556) is
applied to the instance loss but silently skipped for the prior preservation
loss. The prior loss is computed before the `snr_gamma` branch runs, so
`mse_loss_weights` is never multiplied into it — the prior loss always
falls back to plain MSE regardless of `--snr_gamma`.

This PR fixes the `snr_gamma` branch to also apply `mse_loss_weights` to
`prior_loss`, mirroring the same pattern already used for the instance loss.
A new test covering the `--snr_gamma` + `--with_prior_preservation` combination
is also added to `test_dreambooth_lora.py`.

Fixes # (issue)

## Before submitting

## Who can review?
@sayakpaul

### DIFF
--- a/examples/dreambooth/test_dreambooth_lora.py
+++ b/examples/dreambooth/test_dreambooth_lora.py
@@ -377,3 +377,29 @@ class DreamBoothLoRASDXL(ExamplesTestsAccelerate):
                 # checkpoint-2 should have been deleted
                 {"checkpoint-4", "checkpoint-6"},
             )
+
+    def test_dreambooth_lora_sdxl_snr_gamma_with_prior_preservation(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            test_args = f"""
+                examples/dreambooth/train_dreambooth_lora_sdxl.py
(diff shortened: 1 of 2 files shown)


Dev-X25874:
Hey @sayakpaul — would love to get your eyes on this when you have a moment! 🙏

This fixes a silent bug where `mse_loss_weights` from `--snr_gamma` was never applied to the prior preservation loss in `train_dreambooth_lora_sdxl.py` — so prior loss always fell back to plain MSE even when SNR weighting was explicitly requested. The fix mirrors the exact same pattern already used for instance loss, and I've added a test covering the `--snr_gamma` + `--with_prior_preservation` combination.

Happy to adjust anything based on your feedback!

PR #13802, opened 2026-05-24
fix(ernie-image): pass attn_mask=None when text is unpadded

### LINKED ISSUE #13801: Incompatibility between FlashAttention and ERNIE-Image
Hello,

When I enable FlashAttention backend on an `ErnieImagePipeline`, it crashes with following error:

> ValueError: attn_mask is not supported for flash-attn 2.

But same Python code (see below) doesn't crash with `ZImagePipeline` and `Flux2KleinPipeline`.

### Reproduction

````python
pipeline.transformer.set_attention_backend("flash")
````

### System Info

- OS: Windows
- Python: 3.13.11
- Diffusers: 9c4e201dd1cf7d16449c3f4e9affc556d7bc3404
- FlashAttention: 2.8.3+cu130torch2.10
- SDNQ: 0.1.6

### PR DESCRIPTION
**Pull request by Ace3Z**
## What does this PR do?

Fixes #13801.

If you call `pipeline.transformer.set_attention_backend("flash")` on an `ErnieImagePipeline`, it crashes:

```
ValueError: attn_mask is not supported for flash-attn 2.
```

The same code works fine on `ZImagePipeline` and `Flux2KleinPipeline`.

## Root cause

`ErnieImageTransformer2DModel.forward` always builds a boolean attention mask from `text_lens`, even when every sample already has the full text length. In the common single-prompt case (where `Tmax = lens.max()` in `_pad_text`), every position is valid and the mask is just all True. flash-attn 2 rejects any non-`None` `attn_mask`, so the call fails before any attention runs.

Z-Image already handles this. In `_prepare_for_attention` (transformer_z_image.py:792-797) it does:

```python
if all(seq == max_seqlen for seq in item_seqlens):
    attn_mask = None
else:
    attn_mask = ...  # build it
```

## Fix

Do the same thing in Ernie's forward. If at least one sample is padded, build the mask. Otherwise pass `None`.

The previous all-True bool mask was a no-op on sdpa, cudnn and native paths (no positions get masked, no `-inf` rows), so behavior on those backends doesn't change. I verified that numerically: with the same seed and inputs, baseline and fix produce bit-identical output in both the uniform (`text_lens=[16,16]`) and padded (`text_lens=[16,12]`) cases. Norm, mean, std, and per-element samples match to 7 decimal places.

flash-attn and its variants are now usable when the batch is uniform.

## Test

Wires the existing `AttentionBackendTesterMixin` into the Ernie test class (mirrors the Flux pattern). Native_cudnn backend tests pass; flash variants skip cleanly when `kernels` isn't installed, and will catch the regression when it is.

## Before submitting

- [ ] This PR fixes a typo or improves the docs.
- [x] Read the contributor guideline.
- [x] Read the philosophy doc.
- [x] Discussed via GitHub issue (#13801).
- [x] Documentation: mask construction is internal to the forward and not referenced in `docs/source/en/`.
- [x] New tests: wired in `AttentionBackendTesterMixin` for Ernie.

## Who can review?

@yiyixuxu @sayakpaul

### DIFF
--- a/src/diffusers/models/transformers/transformer_ernie_image.py
+++ b/src/diffusers/models/transformers/transformer_ernie_image.py
@@ -408,15 +408,15 @@ class ErnieImageTransformer2DModel(ModelMixin, ConfigMixin, PeftAdapterMixin):
         )
         rotary_pos_emb = self.pos_embed(torch.cat([image_ids, text_ids], dim=1))
 
-        # Attention mask: True = valid (attend), False = padding (mask out), matches sdpa bool convention
-        valid_text = (
-            torch.arange(Tmax, device=device).view(1, Tmax) < text_lens.view(B, 1)
-            if Tmax > 0
-            else torch.zeros((B, 0), device=device, dtype=torch.bool)
-        )
-        attention_mask = torch.cat([torch.ones((B, N_img), device=device, dtype=torch.bool), valid_text], dim=1)[
-            :, None, None, :
-        ]
+        # Attention mask: True = valid (attend), False = padding (mask out), matches sdpa bool convention.
+        # Skip building it when there's no padding so flash-attn (which rejects non-None masks) can run.
+        if Tmax > 0 and not bool((text_lens == Tmax).all()):
+            valid_text = torch.arange(Tmax, device=device).view(1, Tmax) < text_lens.view(B, 1)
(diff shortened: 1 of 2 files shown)


Ace3Z:
Friendly ping on this one. @yiyixuxu @sayakpaul, when you have a moment, would you mind taking a look? It's a small 10 line change to `transformer_ernie_image.py` plus a 5 line test wiring in the existing `AttentionBackendTesterMixin`, mirroring the Z Image pattern. Closes #13801.

PR #13804, opened 2026-05-25
fix(ernie_image): pass attn_mask=None when text is unpadded

### LINKED ISSUE #13801: Incompatibility between FlashAttention and ERNIE-Image

Ace3Z:
Closes #13801.

`pipeline.transformer.set_attention_backend("flash")` on an `ErnieImagePipeline` raises `ValueError: attn_mask is not yet supported for flash-attn 2.`, while the equivalent call on `ZImagePipeline` and `Flux2KleinPipeline` works fine.

`ErnieImageTransformer2DModel.forward` was unconditionally building an attention mask (image ones concatenated with the text validity vector) and passing it to `dispatch_attention_fn` as `attn_mask`. When every text in the batch is at max length the mask is all ones and isn't doing anything, but the flash-attn 2 backend rejects any non-None `attn_mask` (the check at `attention_dispatch.py:1106`).

`ZImageTransformer2DModel._build_inputs` already does the right thing: it sets `attn_mask = None` when every sequence is at `max_seqlen`. Applied the same convention here:

```python
if Tmax > 0 and bool((text_lens < Tmax).any()):
    valid_text = ...
    attention_mask = torch.cat([torch.ones(...), valid_text], dim=1)[:, None, None, :]
else:
    attention_mask = None
```

`ErnieImageSingleStreamAttnProcessor` and `ErnieImageSharedAdaLNBlock` already accept `attention_mask: torch.Tensor | None`, so this is a transformer level change with no downstream edits.

Added one regression test in the existing `test_models_transformer_ernie_image.py`. It uses the existing tester config's `get_init_dict` and `get_dummy_inputs` (which produce unpadded text by default), intercepts `dispatch_attention_fn`, and asserts the `attn_mask` it sees is None. Passes on this branch, fails on main.

For the padded case I ran the forward pass on the same model with the same seed and compared output hashes before vs after the change: byte identical (`4442ce01ccddd5bf` either way for `text_lens=[8,5]`). The all-True mask the unfixed code passed is mathematically equivalent to None for attention math, so output is unchanged.

Also stubbed the actual flash-attn 2 backend dispatch check (`if attn_mask is not None: raise ValueError`) and ran it against the model. On main it reproduces the issue's exact `ValueError`. On this branch the unpadded path runs clean.

### DIFF
--- a/src/diffusers/models/transformers/transformer_ernie_image.py
+++ b/src/diffusers/models/transformers/transformer_ernie_image.py
@@ -408,15 +408,15 @@ class ErnieImageTransformer2DModel(ModelMixin, ConfigMixin, PeftAdapterMixin):
         )
         rotary_pos_emb = self.pos_embed(torch.cat([image_ids, text_ids], dim=1))
 
-        # Attention mask: True = valid (attend), False = padding (mask out), matches sdpa bool convention
-        valid_text = (
-            torch.arange(Tmax, device=device).view(1, Tmax) < text_lens.view(B, 1)
-            if Tmax > 0
-            else torch.zeros((B, 0), device=device, dtype=torch.bool)
-        )
-        attention_mask = torch.cat([torch.ones((B, N_img), device=device, dtype=torch.bool), valid_text], dim=1)[
-            :, None, None, :
-        ]
+        # Only build the mask when there's real padding. flash-attn 2 rejects
+        # any non-None attn_mask, so we leave it None for unpadded inputs.
+        if Tmax > 0 and bool((text_lens < Tmax).any()):
+            valid_text = torch.arange(Tmax, device=device).view(1, Tmax) < text_lens.view(B, 1)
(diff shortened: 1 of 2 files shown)


Ace3Z:
Closing as duplicate of #13802, which fixes the same issue (#13801) with a smaller diff.

PR #13805, opened 2026-05-25
Initialize ZImage pad tokens deterministically

sywangyi:
This PR fixes a determinism issue in ZImageTransformer2DModel by initializing its pad tokens with zeros instead of torch.empty().

The model's pad tokens participate in the forward path when padded caption or image tokens are present. Because they were previously created with torch.empty(), direct model instantiation from config could pick up uninitialized values, which made outputs unstable across runs and could surface NaNs in layerwise casting tests.

pytest tests/models/transformers/test_models_transformer_z_image.py::ZImageTransformerTests::test_layerwise_casting_inference
this case sometimes pass, sometimes fail


sywangyi:
@sayakpaul please help review

sayakpaul (approved):

dg845 (approved):
Thanks for the PR!

PR #13807, opened 2026-05-26
note: torch.zeros -> torch.empty

sayakpaul:
# What does this PR do?

`torch.empty` has disrupted our CI quite a bit in the past. This PR adds a note about it in `.ai/models.md` and encourages the use of `torch.zeros` or `torch.ones`.


.ai/models.md
yiyixuxu, line 156: ohh let's add it here

sayakpaul: Done

yiyixuxu (approved):
thanks

PR #13810, opened 2026-05-26
This PR updates the Stable Diffusion IP-Adapter integration

sywangyi:
What changed

The tests now patch the pipeline-level randn_tensor calls for Stable Diffusion, Stable Diffusion Img2Img, and Stable Diffusion Inpaint with a fixed noise tensor. This removes the dependency on PyTorch's version-specific random sampling behavior and makes the assertions deterministic again. The expected output slices were updated accordingly.

Why

The failures are not caused by a regression in IP-Adapter itself. The root cause is that PyTorch 2.12 generates different random values than PyTorch 2.11 under the same seed when the tests reach the torch.randn path. Since these tests compare output slices against fixed reference values, the changed latents cause them to fail even though the underlying feature behavior is unchanged.

@sayakpaul


dg845 (approved):
Thanks for the PR!

dg845:

PR #13815, opened 2026-05-27
[Pipelines] Add DreamLite text-to-image and image-edit pipelines

Carlofkl:
## Context

This PR integrates **DreamLite** — ByteDance's text-to-image / image-edit diffusion model — into `diffusers`, following an invitation from @NielsRogge to release the model on the Hub in `diffusers` format.

Related issue: https://github.com/ByteVisionLab/DreamLite/issues/3#issuecomment-4289298178

Model cards (public, ungated):
- Base (3-branch dual CFG): https://huggingface.co/carlofkl/DreamLite-base
- Mobile (distilled, single forward): https://huggingface.co/carlofkl/DreamLite-mobile

Both repos use a `diffusers` branch (loaded via `revision="diffusers"`) to keep the original ByteDance-internal `main` branch intact for backward compatibility with existing users.

## What's added

```
src/diffusers/
├── models/unets/unet_dreamlite.py            # DreamLiteUNetModel
├── pipelines/dreamlite/
│   ├── __init__.py
│   ├── pipeline_dreamlite.py                  # DreamLitePipeline (3-branch dual CFG)
│   ├── pipeline_dreamlite_mobile.py           # DreamLiteMobilePipeline (distilled)
│   └── pipeline_output.py
└── (registered in src/diffusers/__init__.py, models/__init__.py,
    pipelines/__init__.py, utils/dummy_*.py)

docs/source/en/api/pipelines/dreamlite.md
tests/pipelines/dreamlite/
├── test_pipeline_dreamlite.py
└── test_pipeline_dreamlite_mobile.py
```

## Architecture highlights

- **`DreamLiteUNetModel`** — UNet-based denoiser conditioned on Qwen3-VL text/vision embeddings.
- **`DreamLitePipeline`** — runs 3 forward passes per step (text-cond / image-cond / uncond) and combines them with a dual-CFG schedule for high-fidelity text-to-image and image edit.
- **`DreamLiteMobilePipeline`** — distilled single-pass variant; no CFG; designed for on-device inference. Pairs with `AutoencoderTiny`.
- Both pipelines use `FlowMatchEulerDiscreteScheduler`.

## Testing

- Loading smoke test against `carlofkl/DreamLite-base` with `revision="diffusers"` — all 6 sub-modules resolve to the correct `diffusers.*` namespace.
- Inference smoke test — generates a 1024×1024 image in ~0.6s/step on a single A800; output stats sane (`std≈93`, no NaN/Inf).
- Standard pipeline tests in `tests/pipelines/dreamlite/`.

## Before submitting

- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussions-related-to-diffusers/63)? Please add a link to it if that's the case. → https://github.com/ByteVisionLab/DreamLite/issues/3
- [x] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/diffusers/tree/main/docs).

## Who can review?

cc @sayakpaul @yiyixuxu @DN6 — thanks in advance for the review!


Carlofkl:
Hi @sayakpaul @yiyixuxu — pushed a small follow-up commit (`032412c`) that fixes the two `check_repository_consistency` failures from the previous run:

1. Registered `DreamLiteAttnProcessor2_0` in `docs/source/en/api/attnprocessor.md` (was missing — `check_support_list.py`).
2. Split combined `height / width` and `guidance_scale / image_guidance_scale` docstring entries in both pipelines into separate lines, and added a complete `Args` block to `DreamLiteTransformer2DModel.forward` (was tripping `check_forward_call_docstrings.py`).

No behavioral change — docs/docstrings only. Verified both lints pass locally.

Whenever convenient, could you re-approve the workflows? Thanks!

Carlofkl:
Hi @yiyixuxu @DN6 @sayakpaul  — quick update: CI is now fully green
(all 15 checks passing) and I've just marked the PR as ready for review.
Whenever you have a moment to take a look, I'd really appreciate it. Thanks!

src/diffusers/models/unets/unet_2d_blocks_dreamlite.py
```diff
@@ -0,0 +1,705 @@
...
+                )[0]
+                hidden_states = resnet(hidden_states, temb)
+
+        return hidden_states
+
+
+# ---------------------------------------------------------------------------
+# Down blocks
+# ---------------------------------------------------------------------------
+def _make_down_block_class(class_name: str, *, remove_self_attn: bool):
```
dg845: Rather than having `_make_down_block_class`, I think we should define the down block classes directly:

```python
# Version with both self attention and cross attention
class DreamLiteAttnDownBlock2D(nn.Module):
    ...

# Version with only cross attention
class DreamLiteCrossAttnDownBlock2D(nn.Module):
    ...
```

since this would be more clear, at the cost of some duplicated code. (I also think the current names are hard to follow, especially `CrossAttnUpRemoveSelfAttnBlock2DV1DreamLite`.)

src/diffusers/models/unets/unet_2d_blocks_dreamlite.py
```diff
@@ -0,0 +1,705 @@
...
+CrossAttnDownBlock2DDreamLite = _make_down_block_class("CrossAttnDownBlock2DDreamLite", remove_self_attn=False)
+CrossAttnDownRemoveSelfAttnBlock2DDreamLite = _make_down_block_class(
+    "CrossAttnDownRemoveSelfAttnBlock2DDreamLite", remove_self_attn=True
+)
+
+
+# ---------------------------------------------------------------------------
+# Up blocks
+# ---------------------------------------------------------------------------
+def _make_up_block_class(class_name: str, *, remove_self_attn: bool):
```
dg845: Analogous comment to https://github.com/huggingface/diffusers/pull/13815#discussion_r3347011790: I think it would be better to define the two attention up block classes directly.

src/diffusers/models/unets/unet_2d_blocks_dreamlite.py
```diff
@@ -0,0 +1,705 @@
...
+CrossAttnUpBlock2DDreamLite = _make_up_block_class("CrossAttnUpBlock2DDreamLite", remove_self_attn=False)
+CrossAttnUpRemoveSelfAttnBlock2DV1DreamLite = _make_up_block_class(
+    "CrossAttnUpRemoveSelfAttnBlock2DV1DreamLite", remove_self_attn=True
+)
+
+
+# ---------------------------------------------------------------------------
+# Plain resnet-only blocks (no attention)
+# ---------------------------------------------------------------------------
+class DownBlock2DDreamLite(nn.Module):
```
dg845:
```suggestion
class DreamLiteDownBlock2D(nn.Module):
```
nit: I think the above suggestion better follows the current `diffusers` model naming patterns.

src/diffusers/models/unets/unet_2d_blocks_dreamlite.py
```diff
@@ -0,0 +1,705 @@
...
+
+        if self.downsamplers is not None:
+            for downsampler in self.downsamplers:
+                hidden_states = downsampler(hidden_states)
+            output_states = output_states + (hidden_states,)
+
+        return hidden_states, output_states
+
+
+class UpBlock2DDreamLite(nn.Module):
```
dg845:
```suggestion
class DreamLiteUpBlock2D(nn.Module):
```
Similar comment to https://github.com/huggingface/diffusers/pull/13815#discussion_r3347025481.

src/diffusers/models/unets/unet_dreamlite.py
```diff
@@ -0,0 +1,703 @@
...
+from ..normalization import RMSNorm
+from .unet_2d_blocks_dreamlite import (
+    CrossAttnDownBlock2DDreamLite,
+    CrossAttnDownRemoveSelfAttnBlock2DDreamLite,
+    CrossAttnUpBlock2DDreamLite,
+    CrossAttnUpRemoveSelfAttnBlock2DV1DreamLite,
+    DownBlock2DDreamLite,
+    UNetMidBlock2DCrossAttnDreamLite,
+    UpBlock2DDreamLite,
+)
```
dg845: I think implementing all of the DreamLite model blocks in a single file (like how recent transformer models are implemented) would better follow the current model design. CC @yiyixuxu

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,538 @@
...
+        return torch.split(selected, valid_lengths, dim=0)
+
+    def encode_prompt(
+        self,
+        mode: str,
+        prompts: List[str],
+        device: torch.device,
+        dtype: torch.dtype,
+        image: Optional[Image.Image] = None,
+        max_sequence_length: int = 500,
```
dg845, line 251: It looks like `max_sequence_length` is currently unused in `encode_prompt`, is this intentional?

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,538 @@
...
+        text_pad_embedding: Optional[torch.Tensor] = None,
+    ):
+        if mode == "edit":
+            drop_idx = 64
```
dg845: Can we document what `drop_idx` means here? Would it be possible to get this value from e.g. `self.processor` instead of hardcoding it?

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,538 @@
...
+            outputs = self.text_encoder(
+                input_ids=tk_out.input_ids,
+                attention_mask=tk_out.attention_mask,
+                pixel_values=tk_out.pixel_values,
+                image_grid_thw=tk_out.image_grid_thw,
+                output_hidden_states=True,
+            )
+
+        elif mode == "generate":
+            drop_idx = 34
```
dg845: Analogous comment to https://github.com/huggingface/diffusers/pull/13815#discussion_r3347061222.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,538 @@
...
+                mode="edit",
+                prompts=[negative_prompt, negative_prompt, prompt_str],
+                image=image,
+                device=device,
+                dtype=dtype,
+            )
+            if num_images_per_prompt > 1:
+                prompt_embeds = prompt_embeds.repeat_interleave(num_images_per_prompt, dim=0)
+                text_attention_mask = text_attention_mask.repeat_interleave(num_images_per_prompt, dim=0)
+            image_processed = self.image_processor.preprocess(image.resize((width, height), Image.Resampling.LANCZOS))
```
dg845:
```suggestion
            image_processed = self.image_processor.preprocess(image, height=height, width=width)
```
Would the above suggestion work? `VaeImageProcessor`'s default `resample` value is `"lanczos"`, so I think we should be able to call `preprocess` normally instead of manually resizing the image first.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,538 @@
...
+                    encoder_attention_mask=text_attention_mask,
+                    added_cond_kwargs={"time_ids": time_ids_in},
+                    return_dict=False,
+                )[0]
+
+                # Classifier-Free Guidance (single for T2I, dual for I2I)
+                noise_pred = noise_pred[..., : latents.shape[-1]]
+                if task == "generate":
+                    noise_pred_uncond, noise_pred_cond = noise_pred.chunk(2)
+                    noise_pred = noise_pred_uncond + self._guidance_scale * (noise_pred_cond - noise_pred_uncond)
```
dg845:
```suggestion
                    noise_pred = noise_pred_uncond + self.guidance_scale * (noise_pred_cond - noise_pred_uncond)
```
nit: I think we should use the `guidance_scale` property here.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,538 @@
...
+                else:  # edit
+                    noise_pred_uncond, noise_pred_image, noise_pred_text = noise_pred.chunk(3)
+                    noise_pred = (
+                        noise_pred_uncond
+                        + self._guidance_scale * (noise_pred_text - noise_pred_image)
+                        + self._image_guidance_scale * (noise_pred_image - noise_pred_uncond)
+                    )
+
```
dg845, line 563:
```suggestion
                    noise_pred = (
                        noise_pred_uncond
                        + self.guidance_scale * (noise_pred_text - noise_pred_image)
                        + self.image_guidance_scale * (noise_pred_image - noise_pred_uncond)
                    )
```
nit: similar comment to https://github.com/huggingface/diffusers/pull/13815#discussion_r3347120330.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite_mobile.py
```diff
@@ -0,0 +1,440 @@
...
+from ...image_processor import VaeImageProcessor
+from ...loaders import FromSingleFileMixin, TextualInversionLoaderMixin
+from ...models import AutoencoderTiny
+from ...models.unets.unet_dreamlite import DreamLiteUNetModel
+from ...schedulers import FlowMatchEulerDiscreteScheduler
+from ...utils import is_torch_xla_available, logging
+from ...utils.torch_utils import randn_tensor
+from ..pipeline_utils import DiffusionPipeline
+from ..stable_diffusion.pipeline_stable_diffusion_img2img import retrieve_latents
+from .pipeline_dreamlite import calculate_shift, retrieve_timesteps
```
dg845:
```suggestion
```
We prefer to copy helper functions like these and use the `# Copied from` mechanism to sync the implementations, similar to how `retrieve_timesteps` is implemented in `src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py`.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite_mobile.py
```diff
@@ -0,0 +1,440 @@
...
+            self.vae_scale_factor = 8
+
+        self.image_processor = VaeImageProcessor(vae_scale_factor=self.vae_scale_factor * 2)
+        self.default_sample_size = 128
+
+    # ---------------------------------------------------------------------
+    # Helpers (identical to DreamLitePipeline)
+    # ---------------------------------------------------------------------
+    @staticmethod
+    def _extract_masked_hidden(hidden_states: torch.Tensor, mask: torch.Tensor) -> List[torch.Tensor]:
```
dg845, line 241:
```suggestion
    @staticmethod
    # Copied from diffusers.pipelines.dreamlite.pipeline_dreamlite.DreamLitePipeline._extract_masked_hidden
    def _extract_masked_hidden(hidden_states: torch.Tensor, mask: torch.Tensor) -> List[torch.Tensor]:
```
We should use the `# Copied from` mechanism for all copied helper methods so that the implementations are synced.

dg845 (commented):

dg845:
Also, if I test out the example using the following script:

```python
import torch
from diffusers import DreamLitePipeline

model_id = "carlofkl/DreamLite-base"
device = "cuda"
dtype = torch.float16

pipe = DreamLitePipeline.from_pretrained(model_id, revision="diffusers", torch_dtype=dtype)
pipe.to(device=device)

# Text-to-image
image = pipe(
    prompt="A serene mountain lake at sunrise",
    generator=torch.Generator(device=device).manual_seed(42),
).images[0]

image.save("dreamlite_t2i.png")

# Image-to-image (instruction-based edit)
image_url = "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/astronaut.jpg"
init_image = load_image(image_url)
edited = pipe(
    prompt="make it snowy",
    image=init_image,
    generator=torch.Generator(device=device).manual_seed(42),
).images[0]

edited.save("dreamlite_i2i.png")
```

I get the following T2I sample:

dreamlite_t2i

and the following I2I sample:

dreamlite_i2i

Is the sample quality expected? The T2I image in particular has a weird block pattern.

src/diffusers/models/unets/unet_2d_blocks_dreamlite.py
Carlofkl: Done in 62a5db6ae. Inlined the factory and the down classes are now defined directly as `DreamLiteCrossAttnDownBlock2D` and                                 
  `DreamLiteCrossAttnNoSelfAttnDownBlock2D`. Much easier to follow now — thanks!

src/diffusers/models/unets/unet_2d_blocks_dreamlite.py
Carlofkl: Done in 62a5db6ae. Same treatment on the up side: `DreamLiteCrossAttnUpBlock2D` and `DreamLiteCrossAttnNoSelfAttnUpBlock2D` are now defined directly.

src/diffusers/models/unets/unet_2d_blocks_dreamlite.py
Carlofkl: Renamed in 62a5db6ae.

src/diffusers/models/unets/unet_2d_blocks_dreamlite.py
Carlofkl: Renamed in 62a5db6ae.

src/diffusers/models/unets/unet_dreamlite.py
Carlofkl: Done in 62a5db6ae. Deleted `unet_2d_blocks_dreamlite.py` and merged all DreamLite blocks (Down / Mid / Up + factory helpers) into `unet_dreamlite.py`, mirroring the structure of recent transformer model files. cc @yiyixuxu

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
Carlofkl: Good catch — it was supposed to bound the generate-mode tokenizer call. Wired in 62a5db6ae at `pipeline_dreamlite.py:268`: `max_length=max_sequence_length + drop_idx`. The edit branch uses the multimodal processor's native padding so this knob doesn't apply there.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
Carlofkl: Done in 62a5db6ae. Replaced the magic number with `self.prompt_template_encode_edit_start_idx`. There's a comment block at `pipeline_dreamlite.py:200-205` explaining how the value is derived: it's the prefix length (system + chat-template scaffolding) you get from running each template through the matching `Qwen3VLProcessor` with an empty user prompt. The values are pinned for reproducibility, mirroring the Qwen-Image pipelines.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
Carlofkl: Done in 62a5db6ae. Same treatment: now `self.prompt_template_encode_generate_start_idx`. See the comment block in the reply above for how the value is derived.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
Carlofkl: Done in 62a5db6ae. Switched to `self.image_processor.preprocess(image, height=height, width=width)` — `VaeImageProcessor`'s default LANCZOS resampling matches the previous behaviour exactly.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
Carlofkl: Done in 62a5db6ae.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
Carlofkl: Done in 62a5db6ae.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite_mobile.py
Carlofkl: Done in 62a5db6ae. `retrieve_latents`, `calculate_shift`, and `retrieve_timesteps` are now defined locally in `pipeline_dreamlite_mobile.py` with `# Copied from` markers (lines 46 / 60 / 74). The cross-pipeline imports have been removed.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite_mobile.py
Carlofkl: Done in 62a5db6ae.

Carlofkl:
@dg845 Thanks for testing! 

Confirmed: the artifact only appears with `transformers >= 5.0`, which changed the underlying tokenization/processor logic for Qwen3-VL in a way that breaks DreamLite. With `transformers==4.57.3` it reproduces cleanly on my end (A800, fp16, `carlofkl/DreamLite-base`) — see attached samples. I've pinned `transformers==4.57.3` in the DreamLite README accordingly.

dreamlite_t2i
dreamlite_i2i

Carlofkl:
@dg845 thanks for the initial review — pushed updates in 62a5db6ae addressing all the inline comments, please take another look when you have time. 

@yiyixuxu friendly ping — would love your eyes on this whenever you get a chance, especially the single-file unet decision in src/diffusers/models/unets/unet_dreamlite.py (per dg845's comment).

src/diffusers/models/resnet_dreamlite.py
```diff
@@ -0,0 +1,254 @@
...
+        )
+        self.pointwise = nn.Conv2d(in_channels, int(out_channels * expand_ratio), kernel_size=1, bias=bias)
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        hidden_states = self.depthwise(hidden_states)
+        hidden_states = self.pointwise(hidden_states)
+        return hidden_states
+
+
+class ResnetBlock2DDreamLite(nn.Module):
```
dg845: I think it fits the single-file design better to implement the resnet inside `unet_dreamlite.py`, especially since this is currently being implemented in its own file (unlike other resnets).

src/diffusers/models/attention_processor.py
```diff
@@ -2787,6 +2787,114 @@ def __call__(
         return hidden_states
 
 
+class DreamLiteAttnProcessor2_0:
```
dg845: Since we've already refactored most of the modeling to live in `unet_dreamlite.py`, I think it makes sense design-wise to also implement the attention processor there, following what recent transformer models do.

src/diffusers/models/attention_processor.py
```diff
@@ -2787,6 +2787,114 @@ def __call__(
...
+    attn.heads`` by reshaping K/V to ``kv_heads`` and then ``repeat_interleave``-ing them up to ``attn.heads``. This is
+    required by the DreamLite UNet, which combines GQA with ``qk_norm`` — a combination the default
+    :class:`AttnProcessor2_0` does not handle.
+    """
+
+    def __init__(self):
+        if not hasattr(F, "scaled_dot_product_attention"):
+            raise ImportError(
+                "DreamLiteAttnProcessor2_0 requires PyTorch 2.0, to use it, please upgrade PyTorch to 2.0."
+            )
```
dg845: I think we can remove this check as `diffusers` currently already requires `torch>=2.0`, and we've been deprecating these errors, for example in Wan:

https://github.com/huggingface/diffusers/blob/f3d42be118f9af7ed9697b686fba09a8bdcd71d1/src/diffusers/models/transformers/transformer_wan.py#L165-L172

src/diffusers/models/attention_processor.py
```diff
@@ -2787,6 +2787,114 @@ def __call__(
...
+        hidden_states: torch.Tensor,
+        encoder_hidden_states: torch.Tensor | None = None,
+        attention_mask: torch.Tensor | None = None,
+        temb: torch.Tensor | None = None,
+        *args,
+        **kwargs,
+    ) -> torch.Tensor:
+        if len(args) > 0 or kwargs.get("scale", None) is not None:
+            deprecation_message = "The `scale` argument is deprecated and will be ignored. Please remove it, as passing it will raise an error in the future. `scale` should directly be passed while calling the underlying pipeline component i.e., via `cross_attention_kwargs`."
+            deprecate("scale", "1.0.0", deprecation_message)
```
dg845:
```suggestion
    ) -> torch.Tensor:
```
Since this a new model, we don't need to support the deprecated `scale` argument.

src/diffusers/models/attention_processor.py
```diff
@@ -2787,6 +2787,114 @@ def __call__(
...
+            key = torch.repeat_interleave(key, heads_per_kv_head, dim=1, output_size=key.shape[1] * heads_per_kv_head)
+            value = torch.repeat_interleave(
+                value, heads_per_kv_head, dim=1, output_size=value.shape[1] * heads_per_kv_head
+            )
+        # ------------------------------------------------------------------------
+
+        # the output of sdp = (batch, num_heads, seq_len, head_dim)
+        hidden_states = F.scaled_dot_product_attention(
+            query, key, value, attn_mask=attention_mask, dropout_p=0.0, is_causal=False
+        )
```
dg845: We prefer that new attention processors use `dispatch_attention_fn` for the SDPA calculation, since it supports using different attention backends (including Flash Attention, the `torch` native SDPA currently used here, etc.). See Flux 2 for an example: 

https://github.com/huggingface/diffusers/blob/79e408a705d6aff9762c58e4601b11e7ea00ea86/src/diffusers/models/transformers/transformer_flux2.py#L369-L376

And for reference, the attention backend docs are [here](https://huggingface.co/docs/diffusers/v0.38.0/en/optimization/attention_backends).

src/diffusers/models/unets/unet_dreamlite.py
```diff
@@ -0,0 +1,1663 @@
...
+            use_sep_conv=use_sep_conv,
+        )
+    if down_block_type in ("CrossAttnDownBlock2D", "CrossAttnDownRemoveSelfAttnBlock2D"):
+        if cross_attention_dim is None:
+            raise ValueError(f"cross_attention_dim must be specified for {down_block_type}")
+        cls = (
+            DreamLiteCrossAttnDownBlock2D
+            if down_block_type == "CrossAttnDownBlock2D"
+            else DreamLiteCrossAttnNoSelfAttnDownBlock2D
+        )
```
dg845: nit: I think we should change the names used here to better reflect the module class names (for example, `"CrossAttnDownBlock2D"` --> `"CrossAttnDownBlock2D"` or `"DreamLiteCrossAttnDownBlock2D"`).

src/diffusers/models/unets/unet_dreamlite.py
```diff
@@ -0,0 +1,1663 @@
...
+            use_sep_conv=use_sep_conv,
+        )
+    if up_block_type in ("CrossAttnUpBlock2D", "CrossAttnUpRemoveSelfAttnBlock2DV1"):
+        if cross_attention_dim is None:
+            raise ValueError(f"cross_attention_dim must be specified for {up_block_type}")
+        cls = (
+            DreamLiteCrossAttnUpBlock2D
+            if up_block_type == "CrossAttnUpBlock2D"
+            else DreamLiteCrossAttnNoSelfAttnUpBlock2D
+        )
```
dg845: nit: analogous comment to https://github.com/huggingface/diffusers/pull/13815#discussion_r3360847690 for the up block names.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,566 @@
...
+from ...image_processor import VaeImageProcessor
+from ...loaders import FromSingleFileMixin, TextualInversionLoaderMixin
+from ...models import AutoencoderTiny
+from ...models.unets.unet_dreamlite import DreamLiteUNetModel
+from ...schedulers import FlowMatchEulerDiscreteScheduler
+from ...utils import is_torch_xla_available, logging
+from ...utils.torch_utils import randn_tensor
+from ..pipeline_utils import DiffusionPipeline
+from ..stable_diffusion.pipeline_stable_diffusion_img2img import retrieve_latents
```
dg845: nit: following `pipeline_dreamlite_mobile.py`, I think we should copy over `retrieve_latents` and use the `# Copied from` mechanism to sync the implementations.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,566 @@
...
+
+    def prepare_image_latents(
+        self,
+        image: Union[torch.Tensor, Image.Image, List[Image.Image]],
+        dtype: torch.dtype,
+        device: torch.device,
+        generator: Optional[torch.Generator] = None,
+    ) -> torch.Tensor:
+        if not isinstance(image, (torch.Tensor, Image.Image, list)):
+            raise ValueError(f"`image` must be of type `torch.Tensor`, `PIL.Image.Image` or `list`, got {type(image)}")
```
dg845: I think we should move this check into a `check_inputs` method, and also add some standard pipeline checks in that method as well. See the Flux 2 pipeline for an example:

https://github.com/huggingface/diffusers/blob/f3d42be118f9af7ed9697b686fba09a8bdcd71d1/src/diffusers/pipelines/flux2/pipeline_flux2.py#L688

We should also add a `check_inputs` method for `DreamLiteMobilePipeline` as well.

dg845 (commented):
Thanks for iterating! Left some follow up comments.

src/diffusers/models/attention_processor.py
Carlofkl: Done in 9fd711a2a. Moved `DreamLiteAttnProcessor2_0` into `unet_dreamlite.py`, removed the class from `attention_processor.py`, and repointed the autodoc entry in `attnprocessor.md` to the new location. Matches what recent transformer models do.

src/diffusers/models/attention_processor.py
Carlofkl: Done in 9fd711a2a. Removed the `hasattr(F, "scaled_dot_product_attention")` ImportError; we rely on the package-level torch>=2.0 requirement.

src/diffusers/models/resnet_dreamlite.py
Carlofkl: Done in 9fd711a2a. Inlined `DepthwiseSeparableConv` and `ResnetBlock2DDreamLite` into `unet_dreamlite.py` and deleted the standalone module — everything DreamLite-specific now lives in one file.

src/diffusers/models/attention_processor.py
Carlofkl: Done in 9fd711a2a. Dropped the `*args, **kwargs` slot and the `deprecate("scale", ...)` block; signature now matches the suggestion exactly.

src/diffusers/models/attention_processor.py
Carlofkl: Done in 9fd711a2a. Switched from `F.scaled_dot_product_attention(...)` to `dispatch_attention_fn(...)`, with `_attention_backend` / `_parallel_config` class attrs (mirroring `Flux2AttnProcessor`). Backend selection / FlashAttention etc. now work out of the box.

src/diffusers/models/unets/unet_dreamlite.py
Carlofkl: Done in 9fd711a2a. Renamed the dispatch keys to match the class names: `"DreamLiteCrossAttnDownBlock2D"`, `"DreamLiteCrossAttnNoSelfAttnDownBlock2D"`, and `"DreamLiteDownBlock2D"`. The `DreamLiteUNetModel.__init__` defaults and the test fixtures were updated in lock-step, and the `unet/config.json` on the `diffusers` branch of the `carlofkl/DreamLite-{base,mobile}` Hub repos has been re-uploaded with the new keys.

src/diffusers/models/unets/unet_dreamlite.py
Carlofkl: Done in 9fd711a2a together with the down-block rename. Same convention: `"DreamLiteCrossAttnUpBlock2D"`, `"DreamLiteCrossAttnNoSelfAttnUpBlock2D"`, `"DreamLiteUpBlock2D"`, plus `"DreamLiteUNetMidBlock2DCrossAttn"` for the mid block.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
Carlofkl: Done in 9fd711a2a. Removed the `from ..stable_diffusion.pipeline_stable_diffusion_img2img import retrieve_latents` import and inlined the function with a `# Copied from` marker, matching `pipeline_dreamlite_mobile.py`.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
Carlofkl: Done in 9fd711a2a. Added `check_inputs(prompt, image, height, width)` to `DreamLitePipeline`, called from `__call__`. The `image` type-check moved out of `prepare_image_latents` into `check_inputs`, and we now also validate that `prompt` is a `str` and warn if `height/width` are not multiples of `vae_scale_factor`. The mobile pipeline gets the same method via `# Copied from diffusers.pipelines.dreamlite.pipeline_dreamlite.DreamLitePipeline.check_inputs`.

Carlofkl:
@dg845 follow-up review fully addressed across two commits (`9fd711a2a` + `63304257e`):

-1. resnet (`DepthwiseSeparableConv` + `ResnetBlock2DDreamLite`) inlined into `unet_dreamlite.py`; `resnet_dreamlite.py` deleted.
-2.`DreamLiteAttnProcessor2_0` moved into `unet_dreamlite.py`; `attention_processor.py` cleaned up; the autodoc entry in `attnprocessor.md` repointed to the new path.
-3. Dropped the `hasattr(F, "scaled_dot_product_attention")` ImportError — `diffusers` already requires `torch>=2.0`.
-4. Dropped the deprecated `scale` handling (`*args, **kwargs` + `deprecate(...)`); the new signature matches your suggestion exactly. 
-5. Switched SDPA to `dispatch_attention_fn` with `_attention_backend` / `_parallel_config` class attrs (mirrors `Flux2AttnProcessor`). Follow-up fix in `63304257e`: kept Q/K/V in `(batch, seq, heads, head_dim)` layout — that's what `dispatch_attention_fn` expects, the previous `(batch, heads, seq, head_dim)` layout collided with the dispatch's internal transpose and broke inference.
-6+7. Renamed block dispatch keys to match the Python class names: `DreamLiteCrossAttn{Down,Up}Block2D`, `DreamLiteCrossAttnNoSelfAttn{Down,Up}Block2D`, `DreamLite{Down,Up}Block2D`, `DreamLiteUNetMidBlock2DCrossAttn`. The `DreamLiteUNetModel.__init__` defaults, the test fixtures, and the `unet/config.json` on the `diffusers` branch of `carlofkl/DreamLite-{base,mobile}` are all updated in lock-step.
-8. `retrieve_latents` inlined locally in `pipeline_dreamlite.py` with a `# Copied from` marker; the cross-pipeline import is gone — same shape as `pipeline_dreamlite_mobile.py`. 
-9. Added `check_inputs(prompt, image, height, width)` to `DreamLitePipeline` and to `DreamLiteMobilePipeline` (mobile uses `# Copied from diffusers.pipelines.dreamlite.pipeline_dreamlite.DreamLitePipeline.check_inputs`); both pipelines call it from `__call__`. The `image`-type validation moved out of `prepare_image_latents` into `check_inputs`, plus we now also check that `prompt` is a `str` and warn if `height`/`width` aren't multiples of `vae_scale_factor`.                                                                                                           

Re-tested end-to-end T2I + I2I locally on A800 / bf16 against `carlofkl/DreamLite-base` (`diffusers` branch) with `transformers==4.57.3` — both modes still produce clean outputs.

Note on CI: the 5 failing PR test jobs (Fast PyTorch Pipeline / Models CPU tests, PyTorch Example CPU tests, Hub tests, LoRA tests with PEFT main) appear to be failing on `main` itself as well — looks like an environment / dependency issue upstream, not introduced by this PR. Happy to investigate further if it persists after main goes green.

tests/pipelines/dreamlite/test_pipeline_dreamlite.py
```diff
@@ -0,0 +1,429 @@
...
+            num_encoder_blocks=(1, 1),
+            num_decoder_blocks=(1, 1),
+            latent_channels=4,
+        )
+
+        scheduler = FlowMatchEulerDiscreteScheduler(num_train_timesteps=1000)
+
+        # text_encoder must expose a real torch.dtype because pipeline does
+        # ``dtype = self.text_encoder.dtype``. Everything else is mocked.
+        text_encoder = MagicMock()
```
dg845: Would it be possible to use a small `Qwen3VLForConditionalGeneration` text encoder model for testing here instead of a `MagicMock` instance (and similarly for the `tokenizer` and `processor`)? For example, NucleusMoE-Image does the following:

https://github.com/huggingface/diffusers/blob/1c7f759d8427f8f0553609375d53932342213584/tests/pipelines/nucleusmoe_image/test_nucleusmoe_image.py#L116-L117

This should let us run the skipped tests below normally. (I'm still reviewing the rest of the PR but wanted to leave this comment first since the change would make reviewing the tests easier.)

Carlofkl: Done in `b39e67251`. Switched both `test_pipeline_dreamlite.py` and `test_pipeline_dreamlite_mobile.py` to a real tiny `Qwen3VLForConditionalGeneration` + `Qwen3VLProcessor`, mirroring `tests/pipelines/nucleusmoe_image/test_nucleusmoe_image.py`. `encode_prompt` is no longer mocked, and as a bonus 9 mixin tests (`save_load_*`, `to_dtype`, `torch_dtype_dict`, `dict_tuple_outputs_equivalent`, `pipeline_with_accelerator_device_map`, `loading_with_variants`, `from_pipe_consistent_config`) now run for real instead of being skipped — see the summary comment for the full before/after.

Carlofkl:
@dg845  thanks for the follow-up! Pushed `b39e67251` which addresses the "use a real (tiny) Qwen3-VL instead of MagicMock" feedback and substantially trims the skip list as a side effect. 

**What changed**
- **`tests/pipelines/dreamlite/test_pipeline_dreamlite.py`** & **`test_pipeline_dreamlite_mobile.py`**
    - `text_encoder` is now a real `Qwen3VLForConditionalGeneration(Qwen3VLConfig(...))` built tiny in memory, mirroring the recipe in `tests/pipelines/nucleusmoe_image/test_nucleusmoe_image.py`.
    - `tokenizer` / `processor` come from `hf-internal-testing/tiny-random-Qwen2VLForConditionalGeneration` (same fixture nucleusmoe_image relies on). 
    - Vision config tweaks needed to keep the tiny model self-consistent:
    - `vision_config.patch_size = 14` to match the cached image processor.
    - `vision_config.out_hidden_size = _CROSS_ATTN_DIM` so vision tokens are projected into the text hidden dim before being merged into the text stream. 
    - `_CROSS_ATTN_DIM = 16` matches the tiny `text_config.hidden_size` and the UNet's `cross_attention_dim`, so `encode_prompt` runs end-to-end and the UNet consumes the result without a custom shape adapter.

- **`src/diffusers/models/unets/unet_dreamlite.py`**
    - `DreamLiteUNetModel` now overrides `set_default_attn_processor()` to reinstall `DreamLiteAttnProcessor2_0` instead of falling through to the parent (which raises because our processor isn't in `ADDED_KV_ATTENTION_PROCESSORS` / `CROSS_ATTENTION_PROCESSORS`). This unblocks every `PipelineTesterMixin` test that round-trips through `set_default_attn_processor` — `test_save_load_local`, `test_save_load_optional_components`, `test_save_load_float16`, `test_dict_tuple_outputs_equivalent`, `test_float16_inference`, `test_pipeline_with_accelerator_device_map`, `test_*cpu_offload*`, etc.

 **Skip list, before vs after**
Removed (these now run for real):
- `test_to_dtype`, `test_torch_dtype_dict`
- `test_save_load_local`, `test_save_load_optional_components`, `test_save_load_float16`, `test_loading_with_variants`, `test_pipeline_with_accelerator_device_map`, `test_from_pipe_consistent_config`, `test_serialization`
- `test_dict_tuple_outputs_equivalent` 

Kept — the rationale is now intrinsic to DreamLite, not to the mock
setup:
- `test_num_images_per_prompt` / `test_inference_batch_consistent` / `test_inference_batch_single_identical`: DreamLite intentionally caps `batch_size = 1` (the 3-way CFG concat already triples the UNet batch); only `num_images_per_prompt > 1` is supported.
- `test_encode_prompt_works_in_isolation`: DreamLite's `encode_prompt` returns `(prompt_embeds, prompt_embeds_mask)`, while the mixin assumes a single tensor return. Same skip pattern as SD3 / Flux.
- `test_save_load_dduf`: `Qwen3VLProcessor.save_pretrained` doesn't currently round-trip through DDUF (the nested `image_processor` sub-config is dropped on reload). Orthogonal to DreamLite.

dg845:
@bot /style

src/diffusers/models/unets/unet_dreamlite.py
```diff
@@ -0,0 +1,2009 @@
...
+            ff_mult=ff_mult,
+            num_kv_heads=num_kv_heads,
+        )
+    raise ValueError(f"DreamLite does not support up_block_type={up_block_type!r}")
+
+
+# ---------------------------------------------------------------------------
+# Model
+# ---------------------------------------------------------------------------
+class DreamLiteUNetModel(UNet2DConditionModel):
```
dg845, line 1559: suggestion: rather than inheriting directly from `UNet2DConditionModel`, `DreamLiteUNetModel` could inherit from the appropriate mixins (`ModelMixin`, `ConfigMixin`, `AttentionMixin`, etc.) and implement the `forward` logic directly. I think this would make the exact logic used by the DreamLite U-Net more clear and fits the single-file design better. CC @yiyixuxu

tests/pipelines/dreamlite/test_pipeline_dreamlite.py
```diff
@@ -0,0 +1,360 @@
...
+
+        def spy_forward(*args, **kwargs):
+            x = args[0] if args else kwargs["sample"]
+            seen_batches.append(x.shape[0])
+            return original_forward(*args, **kwargs)
+
+        pipe.unet.forward = spy_forward
+        inputs = self.get_dummy_i2i_inputs(device)
+        inputs["num_inference_steps"] = 1
+        pipe(**inputs)
```
dg845, line 286:
```suggestion
        inputs = self.get_dummy_i2i_inputs(device)
        inputs["num_inference_steps"] = 1
        with patch.object(pipe.unet, "forward", wraps=pipe.unet.forward) as spy:
            pipe(**inputs)
        seen_batches = [call.args[0].shape[0] for call in spy.call_args_list]
```
nit: I think using `unittest.mock.patch` here would work better than doing it manually (for example, it's more robust if an error occurs).

dg845:
Hi @claude, can you do a review for this PR?

dg845 (commented):
Thanks for the changes! I think this PR is close to merge.

For the changes Claude suggested, I think they all look reasonable and in particular the blocking issues (1, 2, 3) and issue 9 (making sure that the mixin initializations aren't broken) look like the most important issues to address.

Also, can you look into the `DreamLite` pipeline test failures in [this CI run](https://github.com/huggingface/diffusers/actions/runs/27324474979/job/80722946819?pr=13815)? The issues look like they are coming from the `Qwen3VLForConditionalGeneration` text encoder, could this be due to a mismatch between the `transformers` version used in the CI and the version expected by the DreamLite code?

Carlofkl:
Thanks @dg845! Pushed two commits:

- `6653dd491` — addresses issues 1, 2, 3, and 9.
- `a02bb23be` — fixes the CI failures.

The CI failures were caused by recent versions of `Qwen3VLProcessor` emitting a new `mm_token_type_ids` field, which `Qwen3VLModel.compute_3d_position_ids` now requires whenever `image_grid_thw` is provided. Switched both pipelines to `self.text_encoder(**tk_out, output_hidden_states=True)` (matching `NucleusMoEImagePipeline`) so all processor outputs are forwarded automatically.

Could you re-approve the workflow run when you get a chance?

dg845:
@bot /style

docs/source/en/api/pipelines/dreamlite.md
```diff
@@ -0,0 +1,160 @@
...
+
+```python
+import torch
+from diffusers import DreamLitePipeline
+from diffusers.utils import load_image
+
+pipe = DreamLitePipeline.from_pretrained("carlofkl/DreamLite-base", revision="diffusers", torch_dtype=torch.bfloat16)
+pipe = pipe.to("cuda")
+
+source = load_image("https://huggingface.co/datasets/diffusers/diffusers-images-docs/resolve/main/cat.png")
```
dg845: nit: can you check if using this image works? I tried it and got an error, I think because the image doesn't exist.

docs/source/en/api/pipelines/dreamlite.md
```diff
@@ -0,0 +1,160 @@
...
+
+image = pipe(
+    prompt="turn the cat into a corgi",
+    image=source,
+    height=1024,
+    width=1024,
+    num_inference_steps=28,
+    guidance_scale=3.5,
+    image_guidance_scale=1.5,
```
dg845: Are `guidance_scale=3.5` and `image_guidance_scale=1.5` the recommended values? If so, I think we should set them as the pipeline defaults.

docs/source/en/api/pipelines/dreamlite.md
```diff
@@ -0,0 +1,160 @@
...
+
+```python
+import torch
+from diffusers import DreamLiteMobilePipeline
+from diffusers.utils import load_image
+
+pipe = DreamLiteMobilePipeline.from_pretrained("carlofkl/DreamLite-mobile", revision="diffusers", torch_dtype=torch.bfloat16)
+pipe = pipe.to("cuda")
+
+source = load_image("https://huggingface.co/datasets/diffusers/diffusers-images-docs/resolve/main/cat.png")
```
dg845: Same comment as https://github.com/huggingface/diffusers/pull/13815#discussion_r3400460741.

src/diffusers/pipelines/dreamlite/pipeline_dreamlite.py
```diff
@@ -0,0 +1,592 @@
...
+EXAMPLE_DOC_STRING = """
+    Examples:
+        ```py
+        >>> import torch
+        >>> from PIL import Image
+        >>> from diffusers import DreamLitePipeline
+
+        >>> pipe = DreamLitePipeline.from_pretrained(
+        ...     "carlofkl/DreamLite-base", revision="diffusers", torch_dtype=torch.float16
+        ... )
```
dg845, line 55:
```suggestion
        >>> pipe = DreamLitePipeline.from_pretrained(
        ...     "carlofkl/DreamLite-base", revision="diffusers", torch_dtype=torch.bfloat16
        ... )
```
nit: should this be `torch.bfloat16` here (rather than `torch.float16`), following the examples in the documentation?

src/diffusers/pipelines/dreamlite/pipeline_dreamlite_mobile.py
```diff
@@ -0,0 +1,569 @@
...
+EXAMPLE_DOC_STRING = """
+    Examples:
+        ```py
+        >>> import torch
+        >>> from PIL import Image
+        >>> from diffusers import DreamLiteMobilePipeline
+
+        >>> pipe = DreamLiteMobilePipeline.from_pretrained(
+        ...     "carlofkl/DreamLite-mobile", revision="diffusers", torch_dtype=torch.float16
+        ... )
```
dg845, line 143:
```suggestion
        >>> pipe = DreamLiteMobilePipeline.from_pretrained(
        ...     "carlofkl/DreamLite-mobile", revision="diffusers", torch_dtype=torch.bfloat16
        ... )
```
nit: same comment as https://github.com/huggingface/diffusers/pull/13815#discussion_r3400475086.

dg845 (approved):
Thanks for your work on this PR! Left some small comments on the docs/examples.

Carlofkl:
Thanks @dg845! Pushed `f38937d97` addressing all five nits:

1 & 3. Replaced the broken `cat.png` URL with the standard `huggingface/documentation-images/.../diffusers/cat.png` (verified `200 OK`).

2. Promoted `guidance_scale=3.5` / `image_guidance_scale=1.5` to defaults on `DreamLitePipeline.__call__` and removed the now-redundant explicit args from the docs examples.

4 & 5. Switched the `EXAMPLE_DOC_STRING` examples to `torch.bfloat16` in both pipelines.

dg845:

PR #13823, opened 2026-05-28
Add Cosmos3 action generation support

yzhautouskay:
# What does this PR do?

Fixes # (issue)

## Before submitting

## Who can review?


src/diffusers/models/transformers/transformer_cosmos3.py
```diff
@@ -146,6 +146,39 @@ def forward(self, x):
...
 
 
+class DomainAwareLinear(nn.Module):
+    """Linear projection with one weight/bias pair per embodiment domain."""
+
+    def __init__(self, input_size: int, output_size: int, num_domains: int) -> None:
+        super().__init__()
+        self.input_size = int(input_size)
+        self.output_size = int(output_size)
+        self.num_domains = int(num_domains)
```
dg845:
```suggestion
        self.input_size = input_size
        self.output_size = output_size
        self.num_domains = num_domains
```
nit: since we're already annotating `input_size`, etc. as `int`s, I don't think we need to defensively cast using `int(...)` here.

src/diffusers/models/transformers/transformer_cosmos3.py
```diff
@@ -333,6 +369,13 @@ def __init__(
         self.proj_out = nn.Linear(hidden_size, patch_latent_dim, bias=True)
         self.time_proj = Timesteps(num_channels=256, flip_sin_to_cos=True, downscale_freq_shift=0)
         self.time_embedder = TimestepEmbedding(in_channels=256, time_embed_dim=hidden_size)
+        self.action_gen = action_gen
+        self.action_dim = int(32 if action_dim is None else action_dim)
```
dg845:
```suggestion
        self.action_dim = 32 if action_dim is None else action_dim
```
Similar comment to https://github.com/huggingface/diffusers/pull/13823#discussion_r3322571941.

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -396,37 +460,158 @@ def _prepare_sound_segment(
             "sound_len": sound_len,
         }
 
+    def _pack_action_tokens(
```
dg845:
```suggestion
    def _prepare_action_segment(
```
nit: rename `_pack_action_tokens` to `_prepare_action_segment` so that it follows the naming for other modalities (e.g. `_prepare_sound_segment`, `_prepare_vision_segment`, etc.) and so that it is more distinct from the transformer's `_pack_audio_latents` method.

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -435,8 +620,41 @@ def prepare_latents(
...
+        action_image_size: torch.Tensor | None = None
+        vision_condition_frames: list[int] | None = None
+
         # Build the vision conditioning tensor (always [1, 3, T, H, W], in [-1, 1], on device).
-        if is_image:
+        if action_mode is not None:
+            if action_chunk_size is None:
+                raise ValueError("action_mode requires action_chunk_size.")
+            if video is None:
+                raise ValueError(f"action_mode={action_mode!r} requires loaded video conditioning.")
```
dg845: nit: I think we should move the validation checks here (and elsewhere in `prepare_latents`) into the `check_inputs` method so that we can fail earlier.

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -670,22 +1017,34 @@ def current_timestep(self):
...
         negative_prompt: str | list[str] | None = None,
         image: torch.Tensor | None = None,
+        video: Any | None = None,
         num_frames: int = 189,
         height: int = 720,
         width: int = 1280,
         fps: float = 24.0,
         num_inference_steps: int = 35,
         guidance_scale: float = 6.0,
+        flow_shift: float | None = None,
```
dg845:
```suggestion
```
Rather than having a `flow_shift` argument and using it to modify the scheduler in `__call__`, we should respect the pipeline's scheduler config. Users can always specify a new scheduler with the desired `flow_shift` using something like

```python
new_scheduler = UniPCMultistepScheduler.from_config(pipe.scheduler.config, flow_shift=flow_shift)
pipe.scheduler = new_scheduler
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -670,22 +1017,34 @@ def current_timestep(self):
...
         enable_sound: bool = False,
         generator: torch.Generator | None = None,
         latents: torch.Tensor | None = None,
         sound_latents: torch.Tensor | None = None,
+        action_latents: torch.Tensor | None = None,
```
dg845, line 1211: nit: I think we should document the new action arguments in the `__call__` docstring so that it's easier to understand how each argument is supposed to be used.

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -670,22 +1017,34 @@ def current_timestep(self):
...
+        action_mode: str | None = None,
+        action: torch.Tensor | None = None,
```
dg845: I find the `action` name here confusing, especially since there's also an `action_latents` argument. Maybe we could rename it to something like `action_tokens` (not sure if that's the best name)?

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -770,13 +1129,33 @@ def __call__(
         if isinstance(callback_on_step_end, (PipelineCallback, MultiPipelineCallbacks)):
             callback_on_step_end_tensor_inputs = callback_on_step_end.tensor_inputs
 
+        if action_mode is not None and action_mode not in _ACTION_MODES:
+            raise ValueError(f"Unsupported action_mode={action_mode!r}; expected one of {sorted(_ACTION_MODES)}.")
```
dg845: nit: similarly to https://github.com/huggingface/diffusers/pull/13823#discussion_r3322591408, I think we should move this check into `check_inputs`.

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -770,13 +1129,33 @@ def __call__(
...
+        if action_mode is not None and action_chunk_size is not None:
+            num_frames = action_chunk_size + 1
```
dg845: I think we should raise a warning to inform the user whenever `num_frames` get overwritten here.

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -886,29 +1304,58 @@ def __call__(
...
+        if flow_shift is not None:
+            inference_scheduler.register_to_config(
+                use_flow_sigmas=True,
+                use_karras_sigmas=False,
+                use_exponential_sigmas=False,
+                use_beta_sigmas=False,
+                flow_shift=flow_shift,
+                shift_terminal=None,
+                final_sigmas_type="zero",
+            )
```
dg845:
```suggestion
```
Following https://github.com/huggingface/diffusers/pull/13823#discussion_r3322617541, I think we should remove this logic and respect the `scheduler` that's been set on the pipeline.

dg845 (commented):
Thanks for the PR! Left some initial comments :).

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
yzhautouskay: Done

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
yzhautouskay: done

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
yzhautouskay: Renamed to `raw_actions` as `action_tokens`  is taken by tensor of per-step noisy action latents derived from `action_latents`

src/diffusers/models/transformers/transformer_cosmos3.py
```diff
@@ -146,6 +146,39 @@ def forward(self, x):
...
+
+    def __init__(self, input_size: int, output_size: int, num_domains: int) -> None:
+        super().__init__()
+        self.input_size = input_size
+        self.output_size = output_size
+        self.num_domains = num_domains
+        self.fc = nn.Embedding(self.num_domains, self.output_size * self.input_size)
+        self.bias = nn.Embedding(self.num_domains, self.output_size)
+        nn.init.xavier_uniform_(self.fc.weight)
+        nn.init.zeros_(self.bias.weight)
```
yiyixuxu:
```suggestion
```

src/diffusers/models/transformers/transformer_cosmos3.py
```diff
@@ -333,6 +369,13 @@ def __init__(
         self.proj_out = nn.Linear(hidden_size, patch_latent_dim, bias=True)
         self.time_proj = Timesteps(num_channels=256, flip_sin_to_cos=True, downscale_freq_shift=0)
         self.time_embedder = TimestepEmbedding(in_channels=256, time_embed_dim=hidden_size)
+        self.action_gen = action_gen
+        self.action_dim = 32 if action_dim is None else action_dim
```
yiyixuxu:
```suggestion
        self.action_dim = action_dim
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -130,6 +134,62 @@ def get_3d_mrope_ids_vae_tokens(
 _SYSTEM_PROMPT_IMAGE = "You are a helpful assistant who will generate images from a give prompt."
 _SYSTEM_PROMPT_VIDEO = "You are a helpful assistant who will generate videos from a give prompt."
 
+_ACTION_MODE_FORWARD_DYNAMICS = "forward_dynamics"
+_ACTION_MODE_INVERSE_DYNAMICS = "inverse_dynamics"
+_ACTION_MODE_POLICY = "policy"
+_ACTION_MODES = {_ACTION_MODE_FORWARD_DYNAMICS, _ACTION_MODE_INVERSE_DYNAMICS, _ACTION_MODE_POLICY}
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+            Cosmos 3 embodiment domains.
+        resolution_tier (`int`, defaults to `480`):
+            Action conditioning resolution *tier* (one of `256`, `480`, `704`, `720`). The tier picks a predefined
+            canvas whose aspect ratio is closest to the input; the input is downscaled (never upscaled) and padded
+            into it for conditioning. This is not the output frame size, which tracks the input content. Match the
+            tier to the input's native resolution: a lower tier discards detail, while a higher tier adds no
+            resolution (no upscaling), wastes compute on padding, and is a train/inference mismatch that can hurt
+            quality.
+        raw_action_dim (`int`, *optional*):
+            Number of meaningful (unpadded) action channels to keep when slicing predicted actions. Required for
```
yiyixuxu: is the value fixed for each `domain_name`?  
if not, how should a user configure it?

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+        if resolution_key not in _ACTION_RESOLUTION_BINS:
+            raise ValueError(
+                f"Unsupported action resolution_tier={self.resolution_tier!r}; "
+                f"expected one of {sorted(int(k) for k in _ACTION_RESOLUTION_BINS)}."
+            )
+        return VideoProcessor.classify_height_width_bin(
+            source_height, source_width, ratios=_ACTION_RESOLUTION_BINS[resolution_key]
+        )
+
+    def validate(self, *, action_dim: int | None = None) -> None:
```
yiyixuxu:
```suggestion
    def __post_init__(self):
```
can we make this class self-contained? I think we can keep the validation logic on it, moving that into `__post_init__`, instead of the `validate` method called from pipeline.

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+    resolution_tier: int = 480
+    raw_action_dim: int | None = None
+    raw_actions: torch.Tensor | None = None
+    image: Image.Image | np.ndarray | torch.Tensor | None = None
+    video: list | np.ndarray | torch.Tensor | None = None
+
+    @property
+    def num_frames(self) -> int:
+        """Number of conditioning frames the paired video must provide (`chunk_size + 1`)."""
+        return self.chunk_size + 1
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+
+    @property
+    def conditioning_clip(self) -> Any:
+        """Conditioning clip as a single source: `image` wrapped as a one-frame clip, else `video`."""
+        return [self.image] if self.image is not None else self.video
+
+    def target_size(self, source_height: int, source_width: int) -> tuple[int, int]:
```
yiyixuxu: can we move this to pipeline

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+        )
+
+    def validate(self, *, action_dim: int | None = None) -> None:
+        """Validate every action-specific field. Called by [`Cosmos3OmniPipeline.check_inputs`].
+
+        Args:
+            action_dim (`int`, *optional*):
+                The model's action head width. When provided, `raw_actions` channels are checked against it.
+        """
+        if self.mode not in _ACTION_MODES:
```
yiyixuxu:
```suggestion
        if self.mode not in ["policy", "forward_dynamics", "inverse_dynamics"]:
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -130,6 +134,62 @@ def get_3d_mrope_ids_vae_tokens(
...
+    "umi": 6,
+    "bridge_orig_lerobot": 7,
+    "droid_lerobot": 8,
+    "robomind-franka": 8,
+    "galbot": 9,
+    "robomind-franka-dual": 12,
+    "robomind-ur": 13,
+    "agibotworld": 15,
+    "agibot_gear_gripper": 15,
+    "agibot_gear_gripper_ext": 15,
```
yiyixuxu, line 194: why do we have 3 different domain names share same id?

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+                f"expected one of {sorted(_EMBODIMENT_TO_DOMAIN_ID)}."
+            )
+        if str(self.resolution_tier) not in _ACTION_RESOLUTION_BINS:
+            raise ValueError(
+                f"Unsupported action resolution_tier={self.resolution_tier!r}; "
+                f"expected one of {sorted(int(k) for k in _ACTION_RESOLUTION_BINS)}."
+            )
+        if self.image is not None and self.video is not None:
+            raise ValueError("Provide either `image` or `video` for the action condition, not both.")
+        if self.mode == _ACTION_MODE_INVERSE_DYNAMICS:
```
yiyixuxu:
```suggestion
        if self.mode == "inverse_dynamics":
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+                f"Unknown Cosmos3 action domain_name={self.domain_name!r}; "
+                f"expected one of {sorted(_EMBODIMENT_TO_DOMAIN_ID)}."
+            )
+        if str(self.resolution_tier) not in _ACTION_RESOLUTION_BINS:
+            raise ValueError(
+                f"Unsupported action resolution_tier={self.resolution_tier!r}; "
+                f"expected one of {sorted(int(k) for k in _ACTION_RESOLUTION_BINS)}."
+            )
+        if self.image is not None and self.video is not None:
+            raise ValueError("Provide either `image` or `video` for the action condition, not both.")
```
yiyixuxu, line 321:
```suggestion
            raise ValueError("Provide either `image` or `video` for the action condition, not both.")
        elif if self.image is None and self.video is None:
            raise ValueError("`image` and `video` cannot both be None")
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+        if self.mode == _ACTION_MODE_INVERSE_DYNAMICS:
+            if self.video is None:
+                raise ValueError("action mode='inverse_dynamics' requires `video` conditioning.")
+        else:
+            if self.image is None and self.video is None:
+                raise ValueError(f"action mode={self.mode!r} requires `image` or `video` conditioning.")
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+        if self.mode in {_ACTION_MODE_POLICY, _ACTION_MODE_INVERSE_DYNAMICS} and self.raw_action_dim is None:
+            raise ValueError(f"action mode={self.mode!r} requires `raw_action_dim` for output slicing.")
+        if self.mode == _ACTION_MODE_FORWARD_DYNAMICS:
```
yiyixuxu:
```suggestion
        if self.mode =="forward_dynamics":
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+    image: Image.Image | np.ndarray | torch.Tensor | None = None
+    video: list | np.ndarray | torch.Tensor | None = None
+
+    @property
+    def num_frames(self) -> int:
+        """Number of conditioning frames the paired video must provide (`chunk_size + 1`)."""
+        return self.chunk_size + 1
+
+    @property
+    def conditioning_clip(self) -> Any:
```
yiyixuxu: can we move this to pipeline?

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -142,10 +202,130 @@ class Cosmos3OmniPipelineOutput(BaseOutput):
...
+        if self.mode in {_ACTION_MODE_POLICY, _ACTION_MODE_INVERSE_DYNAMICS} and self.raw_action_dim is None:
+            raise ValueError(f"action mode={self.mode!r} requires `raw_action_dim` for output slicing.")
+        if self.mode == _ACTION_MODE_FORWARD_DYNAMICS:
+            if self.raw_actions is None:
+                raise ValueError("action mode='forward_dynamics' requires `raw_actions`.")
+            if self.raw_actions.ndim != 2:
+                raise ValueError(f"`raw_actions` must have shape [T, D], got {tuple(self.raw_actions.shape)}.")
+            if self.raw_actions.shape[0] < 1:
+                raise ValueError("action mode='forward_dynamics' requires at least one action token.")
+            if action_dim is not None and self.raw_actions.shape[-1] > action_dim:
```
yiyixuxu: we can check this in pipeline

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -670,12 +1085,17 @@ def current_timestep(self):
...
+    def do_classifier_free_guidance(self):
+        return self._guidance_scale != 1.0
+
     @torch.no_grad()
     def __call__(
         self,
         prompt: str | list[str],
         negative_prompt: str | list[str] | None = None,
         image: torch.Tensor | None = None,
+        video: Any | None = None,
```
yiyixuxu:
```suggestion
```
let's add this when we add v2v

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -770,13 +1208,54 @@ def __call__(
         if isinstance(callback_on_step_end, (PipelineCallback, MultiPipelineCallbacks)):
             callback_on_step_end_tensor_inputs = callback_on_step_end.tensor_inputs
 
+        # Action runs size via `action.resolution_tier` and derive their frame count from `action.chunk_size`, so
```
yiyixuxu: let's not ignore anything, instead we throw an error (move this to `check_input`)
```py
if action is not None and num_frames is not None:
    raise ValueError(" `num_frames` has to be None if action is not None")
if action is not None and (height is not None or width is not None):
    raise ValueEerror(" `height` and `width` has to be None if action is not None")
```
bit akward becase these are default to be not None, so uwer has to explicitly set to None when using action
but we don't guess user intent &  ignore/override values

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -670,12 +1085,17 @@ def current_timestep(self):
...
+
     @torch.no_grad()
     def __call__(
         self,
         prompt: str | list[str],
         negative_prompt: str | list[str] | None = None,
         image: torch.Tensor | None = None,
+        video: Any | None = None,
         num_frames: int = 189,
         height: int = 720,
```
yiyixuxu: Ideally, we keep these to be `None` since these are not used for action

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -527,6 +914,21 @@ def check_inputs(
...
+        if action is not None:
+            # API-conflict checks live here; all action-field validation is delegated to action.validate().
+            if image is not None or video is not None:
+                raise ValueError(
+                    "Pass action conditioning via `action.image` / `action.video`, not the top-level "
+                    "`image` / `video` arguments."
+                )
+            if not getattr(self.transformer.config, "action_gen", False):
+                raise ValueError("`action` requires a transformer trained with action_gen=True.")
+            action.validate(action_dim=self.transformer.action_dim)
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -527,6 +914,21 @@ def check_inputs(
...
 
+        if action is not None:
+            # API-conflict checks live here; all action-field validation is delegated to action.validate().
+            if image is not None or video is not None:
+                raise ValueError(
+                    "Pass action conditioning via `action.image` / `action.video`, not the top-level "
+                    "`image` / `video` arguments."
+                )
+            if not getattr(self.transformer.config, "action_gen", False):
+                raise ValueError("`action` requires a transformer trained with action_gen=True.")
```
yiyixuxu, line 962:
```suggestion
             action.raw_actions.shape[-1] >self.transformer.config.action_dim:
                 raise ValueError(...)
```
let's check the action.raw_actions shape against transformer config here

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
```diff
@@ -770,13 +1208,54 @@ def __call__(
...
+            callback_on_step_end_tensor_inputs,
+            action,
         )
 
+        # `action_mode` is the only action field consumed directly in __call__ (prompt template + output slicing);
+        # all other action fields are read from `action` at their point of use (e.g. in prepare_latents).
+        action_mode = action.mode if action is not None else None
+
+        if action is not None:
+            num_frames = action.num_frames
```
yiyixuxu: can you rewrite this here, without the methods defined in the action data class? inline the code here instead

yiyixuxu (commented):
thanks, I left some feedbacks

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
yzhautouskay: `domain_name` selects which domain-aware action projection weights are used, rather than uniquely identifying a dataset
So compatible embodiments can share an ID (e.g., `droid_lerobot` and `robomind-franka` both map to 8 for the same setup and compatible action semantics). This shared mapping is also how the model was trained

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
yzhautouskay: We can indeed map `raw_action_dim` from `domain_name` for nearly all domains, so I've added that table and dropped the required arg in [commit](https://github.com/huggingface/diffusers/pull/13823/changes/b5fa4f779a8b19b355ed9d5308e332abcae46ebb). Two domains depend on dataset config instead, resulting in different `raw_action_dim` values: e.g. for `libero` width changes with the rotation representation (axis-angle 3D / 6D / 9D -> 7/10/13 total, will follow up on support for `libero` and `hand_pose`.

dg845:
@bot /style

src/diffusers/pipelines/cosmos/pipeline_cosmos3_omni.py
yzhautouskay: Confirmed `raw_action_dim` value for `hand_pose` and added, `libero` is P1, so we keep it unsupported

yiyixuxu (approved):
thanks!

src/diffusers/models/transformers/transformer_cosmos3.py
```diff
@@ -488,7 +565,14 @@ def forward(
         sound_mse_loss_indexes: torch.Tensor | None = None,
         sound_timesteps: torch.Tensor | None = None,
         sound_noisy_frame_indexes: list[torch.Tensor] | None = None,
-    ) -> tuple[list[torch.Tensor], list[torch.Tensor] | None]:
+        action_tokens: list[torch.Tensor] | None = None,
```
dg845, line 568: nit: can you document the new action-related arguments in the `forward` docstring? The CI is complaining about them not being documented (for example, https://github.com/huggingface/diffusers/actions/runs/26925442980/job/79435285950?pr=13823).

yiyixuxu:
thanks a lot for the PR!

PR #13831, opened 2026-05-29
override torch stuff to prevent them from getting updated

sayakpaul:
# What does this PR do?

Fix https://github.com/huggingface/diffusers/actions/runs/26616425981/job/78433040396

Tested by replicating the CI workflow. The previous PR didn't fully realize the "install dependencies" step, which is what necessitates this PR.


sayakpaul:
Merging to unblock the slow runs.

PR #13836, opened 2026-05-29
moved to a webhook

tarekziade:
# What does this PR do?

Moves the serge bot into using the github app deployed on our infra


.github/workflows/serge_review.yml
```diff
@@ -8,59 +24,75 @@ on:
...
 
-      - name: Check out PR head (shallow)
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
-        with:
-          ref: refs/pull/${{ steps.pr.outputs.number }}/head
-          fetch-depth: 1
+          SIG="sha256=$(openssl dgst -sha256 -hmac "${WEBHOOK_SECRET}" payload.json | awk '{print $NF}')"
 
-      - name: Strip fork-supplied reviewer/agent config
-        # ai-reviewer fetches its config (.ai/review-rules.md, .ai/review-tools.json,
```
sayakpaul, line 50: How does it get fetched then? It's important that these files always get fetched from the upstream `main`.

.github/workflows/serge_review.yml
```diff
@@ -8,59 +24,75 @@ on:
...
+              exit 1
+            }
 
-      - uses: huggingface/ai-reviewer@main
-        with:
-          llm_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
-          llm_api_base: https://api.anthropic.com
-          llm_model: claude-opus-4-6
-          llm_stream: 'true'
-          mention_trigger: '@askserge'
```
sayakpaul, line 66: How do we configure these things?

sayakpaul (commented):
I have configured the secrets.

.github/workflows/serge_review.yml
tarekziade: exactly like the github action : the repo is cloned in the githubapp server and it reads it

.github/workflows/serge_review.yml
tarekziade: in the admin UI. the webook then picks the right now depending on what org/repo calls the service

.github/workflows/serge_review.yml
sayakpaul: For review rules, we always use upstream main right?

sayakpaul (approved):
Good to merge after the conflicts are resolved. Excited to try it out.

.github/workflows/serge_review.yml
tarekziade: no we may use the forked / PR one. I can change that

tarekziade: correction: it is the case, but we do get the context-script from the PR/fork, which is a big issue. Fixing

PR #13851, opened 2026-06-01
Fix duplicate safetensors.load_file call in _onload_from_disk when st…

gagandhakrey:
# Fix: Duplicate `safetensors.load_file` Call in `_onload_from_disk`

## Problem

In `src/diffusers/hooks/group_offloading.py`, `ModuleGroup._onload_from_disk` called `safetensors.torch.load_file()` twice when `stream is None` (the default, non-streaming path):

```python
# Called unconditionally — result discarded when stream is None
device = str(self.onload_device) if self.stream is None else "cpu"
loaded_tensors = safetensors.torch.load_file(
    self.safetensors_file_path,
    device=device,
)

if self.stream is not None:
    ...
else:
    onload_device = (
        self.onload_device.type
        if isinstance(self.onload_device, torch.device)
        else self.onload_device
    )

    # Second load — overwrites the first, which is silently discarded
    loaded_tensors = safetensors.torch.load_file(
        self.safetensors_file_path,
        device=onload_device,
    )
```

The first load result was immediately overwritten and garbage collected. As a result, every `_onload_from_disk` invocation in the default (`use_stream=False`) path performed two full tensor loads from disk while only using the second result.

Since `offload_to_disk_path` is typically used without streaming, this affected the documented and most common usage pattern.

## Impact

* **2× disk I/O** on every group load, slowing inference when disk offloading is enabled.
* **Transient 2× memory usage during loading** (CPU memory and potentially VRAM depending on the target device), increasing the risk of OOM on memory-constrained systems.
* Triggered on **every `_onload_from_disk` call**, which can occur hundreds or thousands of times during an inference run depending on model size and denoising steps.

## Solution

Move the `load_file()` call into the corresponding execution branch so that tensors are loaded exactly once using the correct device argument for that path.

This removes the redundant disk read and avoids allocating an unused copy of the loaded tensors while preserving existing behavior.

### DIFF
--- a/src/diffusers/hooks/group_offloading.py
+++ b/src/diffusers/hooks/group_offloading.py
@@ -259,17 +259,16 @@ class ModuleGroup:
         current_stream = self._torch_accelerator_module.current_stream() if self.record_stream else None
 
         with context:
-            # Load to CPU (if using streams) or directly to target device, pin, and async copy to device
-            device = str(self.onload_device) if self.stream is None else "cpu"
-            loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device=device)
-
             if self.stream is not None:
+                # Load to CPU first, pin memory, then async copy to the target device
+                loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device="cpu")
                 for key, tensor_obj in self.key_to_tensor.items():
                     pinned_tensor = loaded_tensors[key].pin_memory()
                     tensor_obj.data = pinned_tensor.to(self.onload_device, non_blocking=self.non_blocking)
                     if self.record_stream:
                         tensor_obj.data.record_stream(current_stream)
             else:
+                # Load directly to the target device
                 onload_device = (
                     self.onload_device.type if isinstance(self.onload_device, torch.device) else self.onload_device
                 )


gagandhakrey:
Hey @AryanV @sayakpaul, I found a bug in `_onload_from_disk` where `safetensors.load_file()` was called twice in the default (`use_stream=False`) path, causing unnecessary disk I/O and memory overhead. Would appreciate a review. Thanks!

DN6 (approved):
Thanks @gagandhakrey 👍🏽

gagandhakrey:
Thanks @DN6  for the review, looks like CI is green as well , just checking if there is anything else needed from my side so  this can be merged

src/diffusers/hooks/group_offloading.py
```diff
@@ -259,17 +259,16 @@ def _onload_from_disk(self):
         current_stream = self._torch_accelerator_module.current_stream() if self.record_stream else None
 
         with context:
-            # Load to CPU (if using streams) or directly to target device, pin, and async copy to device
-            device = str(self.onload_device) if self.stream is None else "cpu"
-            loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device=device)
-
             if self.stream is not None:
+                # Load to CPU first, pin memory, then async copy to the target device
+                loaded_tensors = safetensors.torch.load_file(self.safetensors_file_path, device="cpu")
```
sayakpaul, line 264: Why should it have to be conditioned on `if self.strea is not None`?

gagandhakrey: @sayakpaul  these two branches handle tensors differently after loading , the stream is not **None** branch needs device="cpu" specifically because it calls .pin_memory() on the tensor before doing a non_blocking async copy to the device — pinned memory is required for async host->device transfers to work correctly, the else branch loads directly to the target device and assigns immediately with no intermediate step, so there's no need to go through **cpu** at all , making the load unconditional with "cpu" would add an unnecessary CPU->device transfer in the no-stream path

sayakpaul (approved):

PR #13854, opened 2026-06-02
Fix resource leak in load_video GIF handling

gagandhakrey:
# Fix : resource leak in load_video GIF handling

## Problem

In `load_video`, `PIL.Image.open()` is called on a GIF file, but the underlying file handle is never explicitly closed after all frames are extracted.

```python
gif = PIL.Image.open(video)  # file handle opened
try:
    while True:
        pil_images.append(gif.copy())
        gif.seek(gif.tell() + 1)
except EOFError:
    pass
# gif.close() never called
```

## Impact

`PIL.Image.open()` keeps the underlying OS file descriptor open for the lifetime of the image object. Without an explicit `close()`, the descriptor is only released when Python's garbage collector destroys the object, which is non-deterministic.

Under repeated calls to `load_video` with GIF inputs, file descriptors can accumulate and eventually exhaust the process limit, resulting in:

```python
OSError: [Errno 24] Too many open files
```

## Fix

Wrap `PIL.Image.open()` in a context manager so the file handle is deterministically closed once all frames have been copied.

```python
with PIL.Image.open(video) as gif:
    try:
        while True:
            pil_images.append(gif.copy())
            gif.seek(gif.tell() + 1)
    except EOFError:
        pass
```

This change is safe because `gif.copy()` creates independent in-memory `PIL.Image` objects. Once the frames are copied, the underlying GIF file can be closed without affecting the returned images.


gagandhakrey:
Hi @sayakpaul, could you please take a quick look at this small GIF file handle fix ? thanks

gagandhakrey:
Hi @DN6 @sayakpaul, could you please take a quick look at this pr

PR #13855, opened 2026-06-03
[tests] fix anyflow tests

sayakpaul:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/actions/runs/26819669538/job/79072404192?pr=13712. Note that the approach I have taken to fix the dtype mismatch error includes an explicit cast. We can disable the layerwise casting tests if that's preferred.

Additionally, removes the `unittest` stuff and solely uses `pytest`.


src/diffusers/models/transformers/transformer_anyflow_far.py
```diff
@@ -121,6 +121,9 @@ def __call__(
         if attn.norm_k is not None:
             key = attn.norm_k(key)
 
+        # Make layerwise upcasting work.
+        value = value.to(query.dtype)
```
dg845:
```suggestion
        query = query.to(target_dtype)
        key = key.to(target_dtype)
```
After some debugging with Claude, I think this should fix all of the layerwise casting issues (where `target_dtype = hidden_states.dtype` is the effective compute dtype, cached before the `to_q`/`to_k`/`to_v` projections happen). I tested this out locally and all the tests pass after this change, including `TestAnyFlowFARTransformer3DMemory::test_layerwise_casting_training` and `TestAnyFlowFARTransformer3DTraining::test_mixed_precision_training`, which were still failing even with the PR.

My understanding of the issue is as follows:

1. Because `norm_q` and `norm_k` are skipped due to `_skip_layerwise_casting_patterns`, and because they are `diffusers.models.normalization.RMSNorm`s which upcast to FP32 during the variance calculation, they end up upcasting `query` and `key` to FP32.
2. Because `value` is not normed, it stays in the effective compute dtype `target_dtype` (for example, the autocast dtype inside a `torch.amp.autocast` region like in `test_layerwise_casting_training`).
3. If we cast `value` to `query`'s dtype, we will therefore end up casting `value` to FP32, regardless of the effective compute dtype. In this case, the Flex Attention operation will always be performed in FP32.
4. However, if we are training, during the backward pass the gradients from the first layer after Flex Attention, which is `to_out`, will be in `target_dtype` (for example because `torch.amp.autocast` will automatically downcast it to the autocast dtype).
5. Whenever `target_dtype != torch.float32`, we will then get a dtype mismatch error when running the Flex Attention backward pass, because the attention scores will be in FP32, but the incoming gradients will be in `target_dtype`.

So casting `query` and `key` to `target_dtype` after the QK norm solves the underlying issue by putting `query`/`key`/`value`, and therefore the Flex Attention computation, in the effective compute dtype, which works for both inference and training.

tests/models/transformers/test_models_transformer_anyflow_far.py
```diff
@@ -137,15 +135,18 @@ def test_gradient_checkpointing_is_applied(self):
     # GPU-only (`torch.nn.attention.flex_attention` raises NotImplementedError on CPU). The
     # bidi transformer test file covers training on the SDPA path; FAR training correctness
     # is exercised end-to-end on H200 via the pipeline replay (L2=0 against NVlabs/AnyFlow).
-    @unittest.skipIf(torch_device == "cpu", "FlexAttention has no CPU backward kernel.")
+    pytest.mark.skipif(torch_device == "cpu", "FlexAttention has no CPU backward kernel.")
+
```
dg845:
```suggestion
```
I think we can remove this because Flex Attention does have a CPU backward path, and the below test (`test_training`) actually passes after the fixes.

tests/models/transformers/test_models_transformer_anyflow_far.py
```diff
@@ -137,15 +135,18 @@ def test_gradient_checkpointing_is_applied(self):
...
     def test_training(self):
         super().test_training()
 
-    @unittest.skipIf(torch_device == "cpu", "FlexAttention has no CPU backward kernel.")
+    pytest.mark.skipif(torch_device == "cpu", "FlexAttention has no CPU backward kernel.")
+
```
dg845:
```suggestion
```
Same reason as https://github.com/huggingface/diffusers/pull/13855#discussion_r3354246526.

tests/models/transformers/test_models_transformer_anyflow_far.py
```diff
@@ -137,15 +135,18 @@ def test_gradient_checkpointing_is_applied(self):
...
     def test_training_with_ema(self):
         super().test_training_with_ema()
 
-    @unittest.skipIf(torch_device == "cpu", "FlexAttention has no CPU backward kernel.")
+    pytest.mark.skipif(torch_device == "cpu", "FlexAttention has no CPU backward kernel.")
+
```
dg845:
```suggestion
```
Same reason as https://github.com/huggingface/diffusers/pull/13855#discussion_r3354246526.

dg845 (commented):
Thanks for the PR! Left some comments.

sayakpaul:
Can you please open a PR to this branch? I think it's just easier that way.

dg845:
Opened a PR to this branch with the suggested fix: #13863.

dg845 (approved):
Thanks!

PR #13859, opened 2026-06-03
Add Ideogram 4

apolinario:
Adds the Ideogram 4 text-to-image model: transformer, standard pipeline, modular pipeline, docs, and tests.

Checkpoint: [`ideogram-ai/ideogram-4-nf4`](https://huggingface.co/ideogram-ai/ideogram-4-nf4) (gated, please accept the gate before using with diffusers)

#### diffusers — remote prompt upsampling

For the best possible results, use the Ideogram prompt upsampling.

```python
import json, os, torch, requests
from diffusers import Ideogram4Pipeline

pipe = Ideogram4Pipeline.from_pretrained(
    "ideogram-ai/ideogram-4-nf4",
    torch_dtype=torch.bfloat16,
    token=os.environ["HF_TOKEN"], # or: token="hf_xxxxxxxxx", token is needed as the repo is gated
).to("cuda")

# Expand the prompt into a structured JSON caption with Ideogram's free hosted magic-prompt API.
# Get a key at https://developer.ideogram.ai/  (set IDEOGRAM_API_KEY).
resp = requests.post(
    "https://api.ideogram.ai/v1/ideogram-v4/magic-prompt",
    headers={"Api-Key": "your_ideogram_api_key"},
    json={"text_prompt": "a ginger cat wearing a tiny wizard hat reading a spellbook", "aspect_ratio": "1x1"},
).json()
caption = json.dumps(resp["json_prompt"])  # already a structured JSON caption

# Pass the caption straight to the pipeline (no prompt_upsampling — it's already upsampled).
image = pipe(
    caption,
    height=1024,  # model supports up to 2048
    width=1024,   # model supports up to 2048
    generator=torch.Generator("cuda").manual_seed(0),
).images[0]
image.save("ideogram4.png")
```

#### diffusers — local prompt upsampling

For a full-local experience, diffusers ships a `prompt_upsampling` that utilizes the same `Qwen3-VL-8B` model as the text-encoder for the upsampling. Expect a quality decrease compared to remote prompt upsampling.

```shell
pip install outlines  # to force the json structure
```

```python
import torch
from diffusers import Ideogram4Pipeline

pipe = Ideogram4Pipeline.from_pretrained(
    "ideogram-ai/ideogram-4-nf4",
    torch_dtype=torch.bfloat16,
    token=os.environ["HF_TOKEN"], # or: token="hf_xxxxxxxxx", token is needed as the repo is gated
).to("cuda")

[...]

PR #13860, opened 2026-06-03
Add structured prompt upsampling to Ideogram4

apolinario:
Ideogram 4 is trained on a structured JSON caption format and quality degrades noticeably on short free-form prompts. The model strongly recommends keeping prompt upsampling on.

This adds an opt-in on-device upsampler that rewrites the user prompt into Ideogram's native JSON caption schema using the pipeline's own Qwen3-VL encoder grafted with a hosted LM head (`diffusers/qwen3-vl-8b-instruct-lm-head`). Constrained decoding via `outlines` (soft dep) enforces the schema. Also wired into the modular pipeline as `Ideogram4PromptUpsampleStep`.

```python
import torch
from diffusers import Ideogram4Pipeline

pipe = Ideogram4Pipeline.from_pretrained("ideogram-ai/ideogram-4-nf4", torch_dtype=torch.bfloat16)
pipe.to("cuda")

# prompt_upsampling=True rewrites the prompt into Ideogram's structured JSON caption on-device,
# using the pipeline's own Qwen3-VL encoder + a grafted LM head.
# Install `outlines` (pip install outlines) for schema-constrained decoding.
image = pipe(
    "a ginger cat wearing a tiny wizard hat reading a spellbook",
    height=2048,
    width=2048,
    prompt_upsampling=True,
    generator=torch.Generator("cuda").manual_seed(0),
).images[0]
image.save("ideogram4.png")
```

Stacked on #13859.


src/diffusers/pipelines/ideogram4/pipeline_ideogram4.py
```diff
@@ -186,6 +194,58 @@ def __init__(
         self.patch_size = 2
         self.image_processor = VaeImageProcessor(vae_scale_factor=self.vae_scale_factor * self.patch_size)
 
+        # Outlines logits processor derived from `prompt_enhancer`; rebuilt by `load_prompt_enhancer`.
+        self._caption_logits_processor = None
+
+    def load_prompt_enhancer(
```
yiyixuxu: I think maybe easiest way is to have this loaded outside and pass it to pipeline as a component 

```py
prompt_enhancer = ...
pipe = DiffusionPipeline.from_pretrained(..., prompt_enhancer=prompt_enhancer)
```

apolinario: That would be more elegant I think but would double the text-encoder memory. 

The reason why I thought of doing it like this is because we can essentially benefit from the already loaded text encoder and just graft the LM Head that it doesn't have (so the same `Qwen3VLModel` that is the text encoder is used here for prompt enhancing, we just use this `graft_lm_head` to slap the lm_head into it and use it for prediction). 

If we do it separately with a typical
`prompt_enhancer = Qwen3VLForConditionalGeneration.from_pretrained(...)`
We couldn't re-use the already loaded components of it

dg845: I think saving the LM head weights as the state dict of a `torch.nn.Module` subclass might work? It would need to be a `ModelMixin` subclass specifically I think to work with e.g. `from_pretrained` (but if it's only passed as a component it could probably be something like a plain `nn.Linear`):

```python
class Ideogram4Pipeline(DiffusionPipeline):
    ...
    _optional_components = ["prompt_upsampler_head"]

    def __init__(
        self,
        ...,
        prompt_upsampler_head: Ideogram4PromptUpsampleHead | None = None,
    ):
        ...
```

and then later on we can still use something similar to `graft_lm_head` (perhaps now as a pipeline method) to combine the existing `self.text_encoder` with the optional `self.prompt_upsampler_head`.

apolinario: applied your suggestion!

src/diffusers/pipelines/ideogram4/pipeline_ideogram4.py
```diff
@@ -156,7 +163,7 @@ class Ideogram4Pipeline(DiffusionPipeline):
     """
 
     model_cpu_offload_seq = "text_encoder->transformer->unconditional_transformer->vae"
```
dg845:
```suggestion
    model_cpu_offload_seq = "prompt_enhancer_head->text_encoder->transformer->unconditional_transformer->vae"
```
Add `prompt_enhancer_head` to `model_cpu_offload_seq` so that it participates in model offloading.

src/diffusers/pipelines/ideogram4/pipeline_ideogram4.py
```diff
@@ -492,6 +562,17 @@ def __call__(
...
+        # 0. Optionally rewrite the prompt(s) into Ideogram4's native structured JSON caption.
+        if prompt_upsampling:
+            prompt = self.upsample_prompt(
+                prompt,
+                height=height,
+                width=width,
+                temperature=prompt_upsampling_temperature,
+                generator=generator,
+                device=device,
+            )
```
dg845, line 575:
```suggestion
            prompt = self.upsample_prompt(
                prompt,
                height=height,
                width=width,
                temperature=prompt_upsampling_temperature,
                max_new_tokens=max_sequence_length,
                generator=generator,
                device=device,
            )
```
I think the call to `upsample_prompt` here should respect `max_sequence_length`, especially since we use it afterward to get stuff like the `neg_position_ids`. (Not sure if the above suggestion is exactly correct since `max_sequence_length` is a limit on all tokens, while `max_new_tokens` looks like a limit on newly-generated non-prompt tokens.)

src/diffusers/modular_pipelines/ideogram4/encoders.py
```diff
@@ -31,6 +38,137 @@
...
+            ),
+            InputParam(
+                name="prompt_upsampling_temperature",
+                type_hint=float,
+                default=PROMPT_UPSAMPLE_TEMPERATURE,
+                description="Sampling temperature for prompt upsampling.",
+            ),
+            InputParam.template("height"),
+            InputParam.template("width"),
+            InputParam.template("generator"),
```
dg845, line 121:
```suggestion
            InputParam.template("height"),
            InputParam.template("width"),
            InputParam.template("max_sequence_length"),
            InputParam.template("generator"),
```
Similar to https://github.com/huggingface/diffusers/pull/13860#discussion_r3353403087, I think the modular pipeline should also respect `max_sequence_length`.

dg845 (approved):
Thanks for the PR! Left some small comments.

src/diffusers/pipelines/ideogram4/pipeline_ideogram4.py
apolinario: Indeed `max_new_tokens` is a limit only on the newly genreated tokens; but that is exactly the prompt that will then be fed into the model - skipping the input prompt. e.g.: 
If i enter "A dog eating pizza" and it generates a big json that is fed into the model, the `a dog eating pizza` tokens are not considered

So except for maybe some special characters or tokens matching `max_new_tokens=max_sequence_length` is perfect

src/diffusers/modular_pipelines/ideogram4/encoders.py
```diff
@@ -31,6 +38,137 @@
...
+            width = block_state.width or components.default_width
+            block_state.prompt = generate_captions(
+                self._prompt_enhancer,
+                components.tokenizer,
+                self._caption_logits_processor,
+                block_state.prompt,
+                height,
+                width,
+                temperature=block_state.prompt_upsampling_temperature,
+                generator=block_state.generator,
```
apolinario, line 166:
```suggestion
                max_new_tokens=block_state.max_sequence_length, 
                generator=block_state.generator,
```

Continuation of add `max_sequence_length` modular

src/diffusers/modular_pipelines/ideogram4/encoders.py
```diff
@@ -31,6 +38,139 @@
...
+            ),
+            InputParam(
+                name="prompt_upsampling_temperature",
+                type_hint=float,
+                default=PROMPT_UPSAMPLE_TEMPERATURE,
+                description="Sampling temperature for prompt upsampling.",
+            ),
+            InputParam.template("height"),
+            InputParam.template("width"),
+            InputParam.template("max_sequence_length"),
```
apolinario:
```suggestion
            InputParam.template("max_sequence_length", default=2048),
```

match the pipe default against the generic `512`

PR #13862, opened 2026-06-04
[tests] Improve ideogram4 tests

sayakpaul:
# What does this PR do?

New model tests must use `pytest` and newly added mixins.


tests/models/transformers/test_models_transformer_ideogram4.py
```diff
@@ -110,24 +83,84 @@ def prepare_init_args_and_inputs_for_common(self):
...
+
+class TestIdeogram4Transformer(Ideogram4TransformerTesterConfig, ModelTesterMixin):
+    """Core model tests for Ideogram 4 Transformer."""
+
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16], ids=["fp16", "bf16"])
+    def test_from_save_pretrained_dtype_inference(self, tmp_path, dtype):
+        # Skip: the non-persistent fp32 RoPE inv_freq buffer is truncated to fp16 by the in-memory
+        # .to(dtype) path but kept fp32 by from_pretrained, so the two outputs diverge well beyond any
+        # meaningful tolerance. Dtype preservation is already covered by test_from_save_pretrained_dtype
+        # and test_keep_in_fp32_modules.
```
sayakpaul, line 149: This seems a bit concerning to me.

dg845: I think an underlying issue with `test_from_save_pretrained_dtype_inference` is that the `model.to(dtype)` cast at

https://github.com/huggingface/diffusers/blob/5d10b4de3b65b1debb8a94de3a0e3bfa51fc628a/tests/models/testing_utils/common.py#L484

is not dtype-aware (it will cast everything to `dtype`), but `from_pretrained(..., torch_dtype=dtype)` is dtype-aware (it respects `_keep_in_fp32_modules`, etc.). This causes the behavior of the two to diverge in several scenarios:

1. There are `_keep_in_fp32_modules` specified on the `model` (the more common case)
2. A non-persistent buffer like `inv_freq` is created with an explicit dtype (which is the case here)

which leads to a divergence between the outputs of `model` and `model_loaded`.

dg845 (approved):
Thanks for the PR!

PR #13868, opened 2026-06-05
fix the bug on the leading time schedule

TheLovesOfLadyPurple:
# What does this PR do?

When the current UniPC Multistep Scheduler call the set_timesteps, it would not generate the correct leading time schedule.  To be more explicitly, if you input different num_inference_steps when you call this function, it will generate different first timestep at the beginning of the inference.  However, user will expect the first step is fixed on 999.

Fixes # input different num_inference_steps to the set_timesteps of the UniPC scheduler would generated different start time step.

We fix this bug on the UniPC Multistep Scheduler.  Now the first inference step is fixed to 999.  We write a simple test code and ensure that the first step of the inference is correct.

## Who can review?

@yiyixuxu

### DIFF
--- a/src/diffusers/schedulers/scheduling_unipc_multistep.py
+++ b/src/diffusers/schedulers/scheduling_unipc_multistep.py
@@ -356,7 +356,7 @@ class UniPCMultistepScheduler(SchedulerMixin, ConfigMixin):
                 .astype(np.int64)
             )
         elif self.config.timestep_spacing == "leading":
-            step_ratio = self.config.num_train_timesteps // (num_inference_steps + 1)
+            step_ratio = self.config.num_train_timesteps // (num_inference_steps)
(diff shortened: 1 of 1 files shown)


TheLovesOfLadyPurple:
If necessary, you can also add a breakpoint on line 1031 in the file src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py to check it.  It's actually very possible that other multi-step solver also has such a problem.

TheLovesOfLadyPurple:
@yiyixuxu Should I first post an issue before I fix this bug? I do think entering an incorrect timestep label will harm the perceptual quality of the generative image, and this simple bug fix is very necessary.

TheLovesOfLadyPurple:
fix the issue in https://github.com/huggingface/diffusers/issues/13887#event-26473373943

PR #13872, opened 2026-06-05
[tests] fix test_from_save_pretrained_dtype_inference

sayakpaul:
# What does this PR do?

Context: https://github.com/huggingface/diffusers/pull/13862#discussion_r3359815645

I have run the tests we were skipping before and they're passing now.


src/diffusers/models/modeling_utils.py
```diff
@@ -1482,6 +1482,12 @@ def cuda(self, *args, **kwargs):
     def to(self, *args, **kwargs):
         from ..hooks.group_offloading import _is_group_offload_enabled
 
+        fp32_modules = self._keep_in_fp32_modules or []
+        if fp32_modules is not None:
```
yiyixuxu: i think we should only warn when it is dtype cast no?
i.e. move it a bit lower and do
`if dtype_present_in_args: ....`

sayakpaul: Indeed. Addressed it.

sayakpaul:
Some test failures in https://github.com/huggingface/diffusers/actions/runs/27183408863/job/80247268518?pr=13872 are unrelated. Fixing the `test_from_save_pretrained_dtype_inference` related ones.

sayakpaul:
Failing tests are now unrelated and I have opened a PR to fix those:
https://github.com/huggingface/diffusers/pull/13894

src/diffusers/models/modeling_utils.py
```diff
@@ -1501,6 +1503,11 @@ def to(self, *args, **kwargs):
                     dtype_present_in_args = True
                     break
 
+        if dtype_present_in_args and fp32_modules is not None:
+            logger.debug(
```
DN6: Would make this `logger.warning`

DN6 (commented):
Looks good. Minor comment.

DN6 (approved):
LGTM. Good to merge once conflicts are resolved

src/diffusers/models/modeling_utils.py
```diff
@@ -1528,6 +1530,11 @@ def to(self, *args, **kwargs):
                     dtype_present_in_args = True
                     break
 
+        if dtype_present_in_args and fp32_modules is not None:
+            logger.warning(
+                f"There are modules in {self.__class__.__name__} that should be kept in float32: {fp32_modules}. A bare `to()` might lead to inconsistent results."
```
DN6:
```suggestion
                f"There are modules in {self.__class__.__name__} that should be kept in float32: {fp32_modules}. Casting directly with `to()` can lead to inconsistent results; set `torch_dtype` in `from_pretrained()` instead to keep these modules in float32."
```

sayakpaul:
Failing tests are unrelated.

PR #13875, opened 2026-06-05
Add TorchAO disk group offload support

hlky:
Split out from huggingface/diffusers#13719 after review feedback to keep the TorchAO safetensors PR focused on save/load.

This follow-up contains only the TorchAO disk group-offload work:
- allow disk group offload for TorchAO tensor subclasses when torchao safetensors support is available
- flatten/unflatten TorchAO subclass tensors through safetensors metadata for disk offload
- add focused TorchAO disk group-offload coverage

Validation so far:
- `python -m ruff check src/diffusers/hooks/group_offloading.py tests/quantization/torchao/test_torchao.py`
- `python -m compileall -q src/diffusers/hooks/group_offloading.py tests/quantization/torchao/test_torchao.py`

Opening as draft because the parent save/load PR should land first, and this PR may need review after that scope settles.


hlky:
@sayakpaul @DN6 This part is ready for review

PR #13884, opened 2026-06-08
switch to a webhook

tarekziade:
# What does this PR do?

Refactored Serge action so it uses a webhook

### DIFF
--- a/.github/workflows/serge_review.yml
+++ b/.github/workflows/serge_review.yml
@@ -4,7 +4,7 @@ name: Claude AI Review with inline comments
 # a thin, VPN-side relay to the Serge GitHub App hosted at
 # https://serge.huggingface.tech/. The App's /webhook endpoint sits behind a VPN
 # that GitHub's own webhook delivery cannot reach, so a runner inside the VPN
-# (group: aws-general-8-plus) re-delivers the triggering comment event to the App.
+# re-delivers the triggering comment event to the App.
 #
 # The relay reproduces a genuine GitHub App webhook delivery:
 #   - body: the original event payload with `installation.id` injected (the App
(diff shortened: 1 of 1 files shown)


.github/workflows/serge_review.yml
```diff
@@ -4,7 +4,7 @@ name: Claude AI Review with inline comments
 # a thin, VPN-side relay to the Serge GitHub App hosted at
 # https://serge.huggingface.tech/. The App's /webhook endpoint sits behind a VPN
 # that GitHub's own webhook delivery cannot reach, so a runner inside the VPN
-# (group: aws-general-8-plus) re-delivers the triggering comment event to the App.
```
sayakpaul, line 7: What re-delivers?

.github/workflows/serge_review.yml
```diff
@@ -46,10 +46,20 @@ jobs:
...
+    # pre-existing tailscaled collided with the action's own daemon). The
+    # Tailscale step below joins this runner to the tailnet so
+    # https://serge.huggingface.tech/ is reachable.
+    runs-on: ubuntu-latest
     steps:
+      - name: Connect to Tailscale
+        uses: tailscale/github-action@v4
+        with:
+          oauth-client-id: ${{ secrets.TS_OAUTH_CLIENT_ID_AI_REVIEW }}
+          oauth-secret: ${{ secrets.TS_AUDIENCE_AI_REVIEW }}
```
sayakpaul, line 59: Is this configured org-wide?

sayakpaul (commented):
Thanks! Left two questions.

.github/workflows/serge_review.yml
tarekziade: yes

.github/workflows/serge_review.yml
tarekziade: I should have removed the 3 lines.
the new workflow uses a standard Github runner, that calls the wehbook right after it has connected to tailscale using the tailscale action

sayakpaul (approved):

Issue #13650, opened 2026-04-29
modular_pipeline_infrastructure model/pipeline review


hlky:
# `modular_pipeline_infrastructure` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

Files reviewed:
- `src/diffusers/modular_pipelines/__init__.py`
- `src/diffusers/modular_pipelines/components_manager.py`
- `src/diffusers/modular_pipelines/mellon_node_utils.py`
- `src/diffusers/modular_pipelines/modular_pipeline.py`
- `src/diffusers/modular_pipelines/modular_pipeline_utils.py`

Duplicate search: checked GitHub Issues and PRs for `modular_pipeline_infrastructure`, `ModularPipeline`, `MellonPipelineConfig`, `from_custom_block`, `output_param_to_mellon_param`, `MODULAR_MODEL_CARD_TEMPLATE`, and the specific failure modes. No exact duplicate found. Related merged PRs exist for nearby modular/Mellon work, especially https://github.com/huggingface/diffusers/pull/13193 and https://github.com/huggingface/diffusers/pull/13051.

## Issue 1: `ModularPipeline()` crashes with an internal `UnboundLocalError`

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/modular_pipelines/modular_pipeline.py#L1688-L1705

Problem:
When `blocks=None` and no `default_blocks_name` is available, `blocks_class` is never initialized, but line 1702 still reads it. The public base class therefore raises an internal `UnboundLocalError` instead of a clear configuration error.

Impact:
Users experimenting with custom modular blocks get a misleading crash before they can recover. This also weakens the fallback behavior added by the related merged PR https://github.com/huggingface/diffusers/pull/13193.

Reproduction:
```python
from diffusers import ModularPipeline

try:
    ModularPipeline()
except Exception as e:
    print(type(e).__name__)
    print(e)
```

Relevant precedent:
`DiffusionPipeline` and other loaders generally raise explicit `ValueError`/`EnvironmentError` messages when required pipeline metadata is missing, rather than leaking local variable errors.

Suggested fix:
```python
blocks_class = None
if blocks is None:
    if modular_config_dict is not None:
        blocks_class_name = modular_config_dict.get("_blocks_class_name")
    else:
        blocks_class_name = self.default_blocks_name

[...]

PR #13888, opened 2026-06-08
fix Mellon custom block required input handling

wunianze666-netizen:
## Summary

This PR preserves required custom block inputs when generating a Mellon config.

## Motivation

While reviewing the Mellon custom block flow, I noticed that MellonPipelineConfig.from_custom_block() always emitted an empty 
equired_inputs list, even when a block input was declared with 
equired=True. The same code path also printed a debug line for every input.

This makes the generated Mellon schema less accurate for custom blocks and can leak noisy stdout from a library API. This came up while reading through the modular Mellon findings in #13650.

## Changes

- track required custom block inputs in rom_custom_block()
- stop printing a debug line for each processed input
- add a regression test covering the generated 
equired_inputs metadata

## Testing

- python -m py_compile src/diffusers/modular_pipelines/mellon_node_utils.py tests/modular_pipelines/test_modular_pipelines_custom_blocks.py

## Notes

I could not run the pytest target in this local environment because pytest and the repo's development dependencies are not installed here, so I limited verification to a syntax check and a focused regression test addition.


yiyixuxu (approved):
thanks

wunianze666-netizen:
Fixed the modular CI regression in `6ab3e4d`: `required_inputs` is retained in the internal node spec and consumed when building `node_params`, but it is not part of the public `node_params` output. The regression test now checks the spec-level metadata and the resulting required label separately. After updating the branch to current `main`, the focused test passes locally on Windows (`1 passed in 3.91s`).

PR #13900, opened 2026-06-09
fix: log ignored longcat positional id arguments

wunianze666-netizen:
## Summary

This PR replaces direct `print()` warnings in the LongCat positional-id helpers with standard logger warnings and fixes the `Unknown type` error message.

## Problem

`prepare_pos_ids()` in both LongCat image pipelines currently prints warnings directly to stdout when callers pass ignored arguments. That is noisy for library consumers and harder to capture in tests. The helper also raises a `KeyError` with a small typo in the message.

## Changes

- replace direct `print()` calls with `logger.warning()` in both LongCat image helpers
- fix `Unknow type` to `Unknown type`
- add a small regression test covering both helpers

## Verification

- `python -m py_compile src/diffusers/pipelines/longcat_image/pipeline_longcat_image.py src/diffusers/pipelines/longcat_image/pipeline_longcat_image_edit.py tests/pipelines/longcat_image/test_prepare_pos_ids.py`
- Not run: `pytest` is not available in this environment, so I could not execute the new test locally.

### DIFF
--- a/src/diffusers/pipelines/longcat_image/pipeline_longcat_image.py
+++ b/src/diffusers/pipelines/longcat_image/pipeline_longcat_image.py
@@ -110,7 +110,7 @@ def prepare_pos_ids(modality_id=0, type="text", start=(0, 0), num_token=None, he
     if type == "text":
         assert num_token
         if height or width:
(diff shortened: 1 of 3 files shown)


wunianze666-netizen:

PR #13905, opened 2026-06-10
[tests] fix consistency decoder tests

sayakpaul:
# What does this PR do?

Fix https://github.com/huggingface/diffusers/actions/runs/27254779798/job/80486772812?pr=13153


tests/models/testing_utils/common.py
```diff
@@ -400,8 +412,12 @@ def recursive_check(tuple_object, dict_object):
         model.to(torch_device)
         model.eval()
 
-        outputs_dict = model(**self.get_dummy_inputs())
-        outputs_tuple = model(**self.get_dummy_inputs(), return_dict=False)
+        inputs_dict = self.get_dummy_inputs()
+        outputs_dict = model(**inputs_dict)
+        # Re-create inputs only if they contain a generator (which needs to be reset)
+        if "generator" in inputs_dict:
+            inputs_dict = self.get_dummy_inputs()
```
DN6: Only the Autoencoders use a generator in their forward, and these Mixin classes should be model agnostic.   

A better pattern is:
```python
  image     = model(**self.get_dummy_inputs(), return_dict=False)[0]
  new_image = new_model(**self.get_dummy_inputs(), return_dict=False)[0]
```

And we enforce deterministic outputs in every `self.get_dummy_inputs()` call.

DN6:
BTW the old format tests are also present in the consistency decoder file
https://github.com/huggingface/diffusers/blob/76e7d164a5eb92a8aeaa83f46b84c67e377591fb/tests/models/autoencoders/test_models_consistency_decoder_vae.py#L105

sayakpaul:
@DN6 fair points. They have been addressed now.

DN6 (approved):

PR #13911, opened 2026-06-10
Add Sound Encoder to Cosmos3

MaciejBalaNV:
# What does this PR do?

Fixes # (issue)

## Before submitting

## Who can review?


src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -221,7 +612,61 @@ def __init__(
...
+    def _disable_encoder(self):
+        self.encoder = None
+        self._encoder_available = False
+        self.register_to_config(encoder_enabled=False)
+
+    def _fix_state_dict_keys_on_load(self, state_dict: OrderedDict) -> None:
+        super()._fix_state_dict_keys_on_load(state_dict)
+        if self.encoder is not None and not any(key.startswith("encoder.") for key in state_dict):
+            self._disable_encoder()
+
```
yiyixuxu, line 557: why do we need these two methods?

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -58,6 +70,211 @@ def forward(self, hidden_states):
         return hidden_states
 
 
+class Cosmos3AudioSnakeBeta(nn.Module):
```
dg845: It looks like the existing `Snake1d` module implements essentially the same logic as `Cosmos3AudioSnakeBeta`, could we use it as well for the encoder?

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -58,6 +70,211 @@ def forward(self, hidden_states):
...
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        alpha = self.alpha[None, :, None]
+        beta = self.beta[None, :, None]
+        if self.logscale:
+            alpha = torch.exp(alpha)
+            beta = torch.exp(beta)
+        return hidden_states + (beta + 1e-9).reciprocal() * torch.sin(alpha * hidden_states).pow(2)
+
+
+class Cosmos3AudioLayerNorm(nn.Module):
```
dg845: Could we potentially reuse the existing `diffusers.models.normalization.FP32LayerNorm` module here? Like `Cosmos3AudioLayerNorm`, it also upcasts the `weight` and `bias` (if available) to FP32:

https://github.com/huggingface/diffusers/blob/0cc1cdb2ee2f2275ed7cf65ca8e6d6360e3b9139/src/diffusers/models/normalization.py#L87-L93

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -58,6 +70,211 @@ def forward(self, hidden_states):
...
+            )
+
+        self.norm = Cosmos3AudioLayerNorm(hidden_dim)
+        self.pwconv1 = nn.Conv1d(hidden_dim, intermediate_dim, kernel_size=1)
+        self.act = Cosmos3AudioSnakeBeta(intermediate_dim) if use_snake else nn.GELU()
+        self.pwconv2 = (
+            _zero_module(nn.Conv1d(intermediate_dim, hidden_dim, kernel_size=1))
+            if identity_init
+            else nn.Conv1d(intermediate_dim, hidden_dim, kernel_size=1)
+        )
```
dg845:
```suggestion
        self.pwconv2 = nn.Conv1d(intermediate_dim, hidden_dim, kernel_size=1)
        if identity_init:
            nn.init.zeros_(self.pwconv2.weight)
            nn.init.zeros_(self.pwconv2.bias)
```
I think the above suggestion would be more clear and would allow us to remove the `_zero_module` helper method, as we prefer not to have too many small methods.

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -13,24 +13,36 @@
...
+from ...utils.torch_utils import randn_tensor
 from ..modeling_utils import ModelMixin
 
 
+def _zero_module(module: nn.Module) -> nn.Module:
+    for parameter in module.parameters():
+        parameter.detach().zero_()
+    return module
+
+
```
dg845:
```suggestion
```
Follow up suggestion to https://github.com/huggingface/diffusers/pull/13911#discussion_r3392854955.

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -58,6 +70,211 @@ def forward(self, hidden_states):
...
+        batch_size, num_channels, num_samples = audio.shape
+        if num_channels != self.input_channels:
+            raise ValueError(
+                f"Cosmos3 AVAE encoder expected {self.input_channels} audio channels, got {num_channels}."
+            )
+
+        if num_channels > 1:
+            audio = audio.reshape(batch_size * num_channels, 1, num_samples)
+
+        with torch.autocast(device_type=audio.device.type, enabled=False):
```
dg845: Would it be possible to remove the `autocast` region here? We generally prefer not to use `autocast` regions and I think mixed-precision training would still work correctly without it (since `_spectrogram` doesn't use any ops that could be dispatched to a lower precision dtype).

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -180,22 +397,114 @@ def forward(self, hidden_state):
         return hidden_state
 
 
+class Cosmos3AudioDiagonalGaussianDistribution:
```
dg845: Would it be possible to reuse the existing `OobleckDiagonalGaussianDistribution` module here? I believe the logic is essentially the same as in `Cosmos3AudioDiagonalGaussianDistribution`.

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -221,7 +612,61 @@ def __init__(
...
+            hidden_states = hidden_states / (hidden_states.abs().max() + 1e-5) * 0.95
+
+        if force_pad or not self.training:
+            sample_length = hidden_states.shape[-1]
+            padding = (self._hop_size - (sample_length % self._hop_size)) % self._hop_size
+            if padding > 0:
+                hidden_states = F.pad(hidden_states, (0, padding), mode="constant", value=0)
+
+        encoder_dtype = next(self.encoder.parameters()).dtype if self.encoder is not None else hidden_states.dtype
+        moments = self._encode(hidden_states.to(dtype=encoder_dtype))
```
dg845:
```suggestion
        encoder_dtype = get_parameter_dtype(self.encoder) if self.encoder is not None else hidden_states.dtype
        moments = self._encode(hidden_states.to(dtype=encoder_dtype))
```
Using `diffusers.models.modeling_utils.get_parameter_dtype` here is more robust to things like layerwise casting, where the storage dtype (for `self.encoder`'s weights) may differ from the compute dtype (which we want `hidden_states` to be in).

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -221,7 +612,61 @@ def __init__(
...
+        super()._fix_state_dict_keys_on_load(state_dict)
+        if self.encoder is not None and not any(key.startswith("encoder.") for key in state_dict):
+            self._disable_encoder()
+
+    def _encode(self, sample: torch.Tensor) -> torch.Tensor:
+        if self.encoder is None or not self._encoder_available:
+            raise ValueError(
+                "This Cosmos3 AVAE sound tokenizer was loaded from decoder-only weights and cannot encode audio. "
+                "Re-convert the AVAE checkpoint with encoder weights to use `encode()`."
+            )
```
dg845: I think it might make more sense to move this check into `encode`, so that we fail earlier.

dg845 (commented):

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: It's an extra safety net for checkpoints that do not have the encoder weights. We will update the main checkpoint to have encoder weights, but I think it's still fine to keep this method in case of e.g. cached local checkpoints. We don't want them to break if people don't need the encoder weights.

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: The math should be the same, but we'd need a reshape on load, since `Cosmos3AudioSnakeBeta` has 1D parameters instead of 3D. Let me think about it for a bit.

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Yes, it works fine, done

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: I kept the separate classes for native checkpoint loading, but shared a forward implementation

dg845: Could we potentially reshape the encoder Snake `alpha`/`beta` weights to 3D in the `scripts/convert_cosmos3_to_diffusers.py` conversion script? I think this would allow us to reuse `Snake1d` for the encoder.

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -47,17 +54,203 @@ def __init__(self, hidden_dim, logscale=True):
         self.logscale = logscale
 
     def forward(self, hidden_states):
-        shape = hidden_states.shape
+        return self._forward(hidden_states, self.alpha, self.beta, self.logscale)
 
-        alpha = self.alpha if not self.logscale else torch.exp(self.alpha)
-        beta = self.beta if not self.logscale else torch.exp(self.beta)
+    @staticmethod
+    def _forward(hidden_states, alpha, beta, logscale):
```
dg845: Since `Snake1d` is `# Copied from` the Oobleck audio VAE for the Cosmos 3 decoder, I don't think we should modify it here (and this breaks the CI check that the implementations are synced). I think it would be better (for example) to convert the encoder Snake weights from 1D to 3D when converting the checkpoint, as suggested in https://github.com/huggingface/diffusers/pull/13911#discussion_r3407432010.

tests/models/autoencoders/test_models_autoencoder_cosmos3_audio.py
```diff
@@ -0,0 +1,166 @@
...
+
+from diffusers.models.autoencoders.autoencoder_cosmos3_audio import (
+    Cosmos3AudioSnakeBeta,
+    Cosmos3AVAEAudioTokenizer,
+    Snake1d,
+)
+from diffusers.models.autoencoders.autoencoder_oobleck import OobleckDiagonalGaussianDistribution
+
+
+def _get_tiny_cosmos3_audio_tokenizer() -> Cosmos3AVAEAudioTokenizer:
```
dg845: Would it be possible to refactor the tests to use the standard `diffusers` model test config + test mixins? For reference, here is what the LTX-2 audio VAE tests do:

https://github.com/huggingface/diffusers/blob/f7f1f6304d90215122e43544e36f4f894ec66d4c/tests/models/autoencoders/test_models_autoencoder_kl_ltx2_audio.py#L77-L85

So for example we would move the `_get_tiny_cosmos3_audio_tokenizer` logic into the `get_init_dict` method of a new `Cosmos3AVAEAudioTokenizerTesterConfig` class. I think we would still keep the Cosmos 3-specific tests below.

dg845 (commented):
Thanks for the changes! Left some follow up comments.

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

tests/models/autoencoders/test_models_autoencoder_cosmos3_audio.py
MaciejBalaNV: Done

dg845:
@bot /style

src/diffusers/models/autoencoders/autoencoder_cosmos3_audio.py
```diff
@@ -238,3 +615,23 @@ def decode(self, latents: torch.Tensor) -> torch.Tensor:
...
+        self,
+        sample: torch.Tensor,
+        sample_posterior: bool = False,
+        return_dict: bool = True,
+        generator: torch.Generator | None = None,
+        force_pad: bool = False,
+    ) -> Cosmos3AudioDecoderOutput | tuple[torch.Tensor]:
+        """Encode then decode a waveform; ``sample_posterior=False`` (default) decodes the distribution mode (mean),
+        whereas the upstream Cosmos3 AVAE always samples — pass ``sample_posterior=True`` for reference-equivalent
+        behavior."""
```
dg845: Can you update the `Cosmos3AVAEAudioTokenizer.forward` docstring to document the arguments and return values? This should fix the failure in [this CI run](https://github.com/huggingface/diffusers/actions/runs/27588511382/job/81564365354?pr=13911).

dg845 (approved):
Thanks for your work on this PR! Left one comment about fixing a CI failure.

MaciejBalaNV:
Fixed CI. I think the PR is ready to be merged now. The checkpoints on HF should be updated today or tomorrow with encoder weights, but the current encoder-less checkpoints also work, as long as we don't call the `encode` method.

PR #13922, opened 2026-06-11
Fix `Ideogram4MRoPE` collapsing under `torch.autocast` (compute rotary in float32)

### LINKED ISSUE #13920: Ideogram4: `Ideogram4MRoPE` breaks under `torch.autocast`: all image positions collapse, producing flat single-color images
### Describe the bug

`Ideogram4MRoPE.forward` computes the rotary frequencies with a matmul:

```python
freqs = inv_freq @ pos.unsqueeze(2)
```

`torch.matmul` is on autocast's lower-precision op list, so when the transformer's forward runs inside
`torch.autocast("cuda", torch.bfloat16)` — common in training code, and users also frequently wrap pipeline calls in
autocast — this matmul executes in bfloat16 even though the module explicitly casts both operands to float32.

Ideogram4's image-token positions are `IMAGE_POSITION_OFFSET + (t, h, w)` = 65536…65536+grid. bfloat16 has an 8-bit
mantissa, so its representable step at 65536 is 512: **every image position in a ≤512-wide grid rounds to exactly
65536**. All image tokens receive identical rotary embeddings, the model loses all spatial information, and sampling
degenerates to a near-uniform velocity field — the decoded image is a single flat color (with faint patch-sized
texture). Text positions (0…2047) are also quantized (step 8–16 above 1024), but the image-position collapse is total.

This was painful to find because it is invisible at the weight/IO level: text encoding, packing, scheduler, and the
checkpoint all check out, and the same code produces correct images as soon as the autocast context is removed.
Tracing per-module outputs (autocast vs. no autocast) shows `rotary_emb` as the first divergent module, with
max |Δcos| ≈ 2.0 (i.e. fully flipped rotations).

Other diffusers RoPE implementations that matmul/outer raw position values are exposed to the same class of bug, but Ideogram4 is uniquely catastrophic because of the 65536 position offset.

---

*Drafted by Claude*

The rationale for Ideogram4 using position offsets >= 65536 is here:
  - Definition: venv/src/diffusers/src/diffusers/models/transformers/transformer_ideogram4.py:42 — IMAGE_POSITION_OFFSET = 65536, with the upstream comment "Image grid coordinates start at this offset so they never collide with text token indices."

### Reproduction

Weight-free, runs in under a second:

```python
import torch
from diffusers.models.transformers.transformer_ideogram4 import IMAGE_POSITION_OFFSET, Ideogram4MRoPE

rope = Ideogram4MRoPE(head_dim=256, base=5_000_000, mrope_section=(24, 20, 20)).to("cuda")

# three image-grid positions as built by Ideogram4Pipeline._prepare_ids: (t, h, w) + IMAGE_POSITION_OFFSET
pos = torch.tensor([[[0, 0, 0], [0, 0, 1], [0, 63, 63]]], device="cuda") + IMAGE_POSITION_OFFSET

cos_ref, sin_ref = rope(pos)
with torch.autocast(device_type="cuda", dtype=torch.bfloat16):
    cos_ac, sin_ac = rope(pos)

print("fp32:     token(0,0) == token(0,1):", torch.equal(cos_ref[0, 0], cos_ref[0, 1]))
print("autocast: token(0,0) == token(0,1):", torch.equal(cos_ac[0, 0], cos_ac[0, 1]))
print("autocast: token(0,0) == token(63,63):", torch.equal(cos_ac[0, 0], cos_ac[0, 2]))
print("max |cos_autocast - cos_fp32|:", (cos_ac - cos_ref).abs().max().item())
```

Output:

```
fp32:     token(0,0) == token(0,1): False
autocast: token(0,0) == token(0,1): True
autocast: token(0,0) == token(63,63): True
max |cos_autocast - cos_fp32|: 1.933509349822998
```

End-to-end: running the `Ideogram4Pipeline` denoising loop inside `torch.autocast("cuda", torch.bfloat16)` produces a
flat single-color image; the identical call without autocast (same seed, same weights) renders correctly.

### Expected behavior

Rotary embeddings should be computed in float32 regardless of an ambient autocast context. transformers guards all of
its rotary embedding modules for exactly this reason, e.g. Qwen3VL
(`transformers/models/qwen3_vl/modeling_qwen3_vl.py`):

```python
with maybe_autocast(device_type=device_type, enabled=False):  # Force float32
```

(That guard is also why the Qwen3-VL *text encoder* used by this very pipeline is unaffected — only the diffusers-side
DiT MRoPE breaks.)

Suggested fix in `Ideogram4MRoPE.forward`:

```python
def forward(self, position_ids: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
    ...
    device_type = position_ids.device.type if position_ids.device.type != "mps" else "cpu"
    with torch.autocast(device_type=device_type, enabled=False):
        pos = position_ids.permute(2, 0, 1).to(dtype=torch.float32)
        inv_freq = self.inv_freq.to(dtype=torch.float32)[None, None, :, None].expand(3, batch_size, -1, 1)
        freqs = inv_freq @ pos.unsqueeze(2)
        ...
```

### Logs

```shell

```

### System Info

- diffusers 0.39.0.dev0 (main @ `9a0aaba36`)
- torch 2.12.0+cu130, CUDA, NVIDIA GeForce RTX 4070 Ti SUPER
- transformers 5.9.0
- Python 3.12, Linux

### Who can help?

Transformers/Attention @DN6 @yiyixuxu @sayakpaul

cc @bghira

HaozheZhang6:
# What does this PR do?

Fixes #13920

`Ideogram4MRoPE` produces collapsed rotary embeddings under `torch.autocast`, so denoising inside an autocast context (common in training, and when users wrap pipeline calls) renders a flat single-color image.

## Root cause

Image-token positions are `IMAGE_POSITION_OFFSET (65536) + (t, h, w)`. `Ideogram4MRoPE.forward` casts its operands to float32, but the frequency matmul is on autocast's downcast list, so under `torch.autocast("cuda", torch.bfloat16)` it executes in bfloat16 anyway. bfloat16's representable step at 65536 is 512, so **every image position in a ≤512-wide grid rounds to the same value** — all image tokens get identical rotary embeddings, spatial information is lost, and sampling degenerates to a flat field.

Reproduced with the weight-free snippet from the issue (`max |cos_autocast − cos_fp32| ≈ 1.93`, distinct positions become equal).

## Fix

Wrap the frequency computation in `torch.autocast(device_type=..., enabled=False)` so the rotary embeddings are always computed in float32 regardless of an ambient autocast — the same guard `transformers` applies to its RoPE modules. After the fix the autocast and float32 paths are bit-identical (`max |Δ| = 0.0`).

Scope is `Ideogram4MRoPE`, the catastrophic case (others noted in the issue are far milder without the 65536 offset). Happy to extend the same guard to the sibling RoPE modules in a follow-up if you'd like.

## Tests

Added `test_ideogram4_mrope_is_autocast_invariant` — it fails on `main` (collapsed positions) and passes with the fix. Full file green:

```
pytest tests/models/transformers/test_models_transformer_ideogram4.py
38 passed, 4 skipped
```

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? #13920
- [x] Did you make sure to update the documentation with your changes? (n/a — bug fix)

## Who can review?

@DN6 @sayakpaul


dxqb:
before committing that (and thereby closing my report), please consider that other modules might be affected, just not as bad. bfloat16 becomes inaccurate for integers starting 257.0 (which is rounded to 256.0).

that's within the range of text token ids

HaozheZhang6:
You're right — confirmed bf16 rounds 257→256, 259→260, so text positions past 256 lose precision in any RoPE that matmuls raw position ids under autocast. Ideogram4 is just the pathological case: the 65536 offset collapses a whole ≤512-wide grid onto a single value, where the others degrade gradually instead of all-at-once.

I'd checked the other diffusers transformers — Ideogram4 is the only RoPE with a large position offset, so the only catastrophic one — but the gradual loss you describe is real for the rest. I can extend the same `autocast(enabled=False)` guard to the other RoPE forwards in this PR, or keep this one targeted at the Ideogram4 regression and do a follow-up sweep, whichever you and the maintainers prefer. Either way it shouldn't close your report until the broader case is covered.

src/diffusers/models/transformers/transformer_ideogram4.py
```diff
@@ -70,20 +70,25 @@ def forward(self, position_ids: torch.Tensor) -> tuple[torch.Tensor, torch.Tenso
...
-            idx = torch.arange(offset, length, 3, device=freqs_t.device)
-            freqs_t[..., idx] = freqs[axis][..., idx]
-
-        emb = torch.cat((freqs_t, freqs_t), dim=-1)
-        return emb.cos(), emb.sin()
+        # Rotary frequencies must be computed in float32: Ideogram4's image positions start at
+        # IMAGE_POSITION_OFFSET (65536), so an ambient autocast would otherwise run the matmul and
+        # cos/sin in bfloat16, rounding every image position to the same value and collapsing the
+        # rotary embeddings (all spatial information is lost).
+        with torch.autocast(device_type=position_ids.device.type, enabled=False):
```
sayakpaul, line 79: We don't use `autocast` within our modeling implementation like this.

HaozheZhang6: Good catch — dropped the `autocast` guard and compute the freqs in `float64` instead, which autocast doesn't downcast (matching the float64 rope path Flux uses). The autocast and float32 paths come out bit-identical (`max|Δ| = 0`), and the regression test still passes.

dxqb: > We don't use `autocast` within our modeling implementation like this.

@sayakpaul Why?
Maybe this can be reconsidered.
1. It's the right solution. Casting to float64 as @HaozheZhang6's AI suggested below is a bad workaround
2. This is what huggingface transformers does: https://github.com/huggingface/transformers/blob/08a7ef05bcf9723cb2e58855afb8dc2c799323ff/src/transformers/models/qwen3_vl/modular_qwen3_vl.py#L304

sayakpaul: We are considering this as a library-wide thing i.e., to handle these kinds of situations. So, expect a PR soon that will also include this case.

Cc: @dg845 @yiyixuxu @DN6 as we were discussing this.

dg845: After some internal discussion, we decided that using `torch.autocast(..., enabled=False)` makes sense here, so the original implementation which uses it is fine.

src/diffusers/models/transformers/transformer_ideogram4.py
```diff
@@ -70,8 +70,12 @@ def forward(self, position_ids: torch.Tensor) -> tuple[torch.Tensor, torch.Tenso
             raise ValueError(f"`position_ids` must have shape (B, L, 3), got {tuple(position_ids.shape)}.")
         batch_size, seq_len, _ = position_ids.shape
 
-        pos = position_ids.permute(2, 0, 1).to(dtype=torch.float32)
-        inv_freq = self.inv_freq.to(dtype=torch.float32)[None, None, :, None].expand(3, batch_size, -1, 1)
+        # Rotary frequencies are computed in float64: Ideogram4's image positions start at
+        # IMAGE_POSITION_OFFSET (65536), which float32 cannot represent distinctly once an ambient
+        # autocast runs the matmul/cos/sin in bfloat16, collapsing every image position to the same
+        # embedding. float64 is not downcast by autocast, matching the float64 rope path Flux uses.
```
dg845:
```suggestion
        # Ideogram4's image position ids start at IMAGE_POSITION_OFFSET (65536). If an ambient autocast downcasts the
        # matmul to bfloat16, the image positions will collapse to only a few distinct values because bfloat16 cannot
        # represent consecutive integers at this value (after pos 65536 each 512-integer block will collapse to the
        # same value), which causes the image to become essentially flat. Therefore, we need to disable autocast here.
```
nit: I think the suggested comment more accurately describes the situation (in particular it's `bfloat16`, not `float32`, which can't represent the consecutive integers at the offset).

dg845 (commented):
Thanks for the PR! Would you be willing to extend the fix to other RoPE modules? Using `torch.autocast(enabled=False)` would be fine, as described in https://github.com/huggingface/diffusers/pull/13922#discussion_r3433126738.

HaozheZhang6:
Done — switched Ideogram4 back to `torch.autocast(..., enabled=False)` around the rope matmul and took your comment, thanks @dg845.

On extending it: a fair number of the other transformer RoPE modules build their freqs the same way (matmul of raw position ids), so they'd downcast under an ambient autocast too — most won't collapse as hard as Ideogram4's 65536 offset, but they lose precision once positions pass ~257 in bf16. Happy to wrap each in `autocast(enabled=False)`. Do you want them all in this PR, or a focused first set? A few are `# Copied from`-linked so I'll edit the source and run `make fix-copies`.

src/diffusers/models/transformers/transformer_ideogram4.py
```diff
@@ -70,20 +70,25 @@ def forward(self, position_ids: torch.Tensor) -> tuple[torch.Tensor, torch.Tenso
...
-            idx = torch.arange(offset, length, 3, device=freqs_t.device)
-            freqs_t[..., idx] = freqs[axis][..., idx]
-
-        emb = torch.cat((freqs_t, freqs_t), dim=-1)
-        return emb.cos(), emb.sin()
+        # Ideogram4's image position ids start at IMAGE_POSITION_OFFSET (65536). If an ambient autocast downcasts the
+        # matmul to bfloat16, the image positions will collapse to only a few distinct values because bfloat16 cannot
+        # represent consecutive integers at this value (after pos 65536 each 512-integer block will collapse to the
+        # same value), which causes the image to become essentially flat. Therefore, we need to disable autocast here.
+        with torch.autocast(device_type=position_ids.device.type, enabled=False):
```
dg845, line 79: I think it would be better to tighten the `torch.autocast` region to just the `freqs` matmul, since it's the operation that is actually precision-sensitive and needs the guard. So maybe something like

```python
        # <explanatory comment from above>
        pos = position_ids.permute(2, 0, 1).to(dtype=torch.float32)
        inv_freq = self.inv_freq.to(dtype=torch.float32)[None, None, :, None].expand(3, batch_size, -1, 1)
        with torch.autocast(device_type=position_ids.device.type, enabled=False):
            freqs = inv_freq @ pos.unsqueeze(2)
        freqs = freqs.transpose(2, 3)  # (3, B, L, inv_freq_size)

        # Rest of the implementation (setting up interleaved mrope, cos/sin call)
        ...
```

dg845:
Hi @HaozheZhang6, I think fixing all of the RoPE modules that build their `freqs` like Ideogram 4 and thus have the bug  (with the changes propagated via `make fix-copies`) would be best.

HaozheZhang6:
Went through every RoPE module and extended the guard to the unguarded ones.

**Fixed** (wrapped in `torch.autocast(enabled=False)`): `ideogram4` + `cosmos3` (position-id matmul), `ernie_image` + `helios` (position-id `einsum`).

**Already handled, left alone**: `omnigen` (matmul already under `autocast(enabled=False)`), `mochi` (einsum already forced to fp32), `hidream_image` (computes in float64, which autocast leaves alone).

The `torch.outer`-based ones (cogview4, glm_image, qwenimage, cosmos, kandinsky, z_image, longcat, nucleusmoe, joyimage) aren't affected: autocast's lower-precision list covers the GEMM family (matmul/mm/bmm/einsum) but not `outer`/`ger`, so their positions stay full-precision (confirmed locally — `outer` keeps float32 under autocast).

`make fix-copies` is clean (none of the edited functions are `# Copied from` sources/copies). On tests: the matmul collapse reproduces on CPU (the existing Ideogram4 test covers it, and cosmos3 is the same path), but the einsum collapse is CUDA-only — CPU autocast doesn't downcast einsum — so I relied on the matmul test plus the mochi/omnigen precedent rather than a CPU test that can't fail there. Happy to add per-module autocast-invariance tests (CUDA-gated for ernie/helios) if you'd prefer.

dg845 (approved):
Thanks for the follow up changes! Can you add tests similar to the Ideogram4 one for the other models that got modified (Cosmos 3, Ernie, Helios)?

HaozheZhang6:
Done — tightened the guard to just the matmul in `ideogram4` and `cosmos3` (`ernie_image`/`helios` already wrap only the einsum). Test still passes.

PR #13928, opened 2026-06-12
Add PRXPixelPipeline: pixel-space PRX text-to-image pipeline

DavidBert:
# What does this PR do?

Adds a pixel-space variant of PRX that denoises raw RGB directly (no VAE), conditioned on a Qwen3-VL text encoder:

- PRXTransformer2DModel: new optional config args `bottleneck_size` (two-layer img_in projection for large patch dims) and `resolution_embeds` (PRXResolutionEmbedder conditions the timestep modulation on the latent resolution)
- PRXPipeline: support for subclass-tuned tokenizer max length, light text cleaning, x-prediction flow matching (x0 -> velocity conversion), and non-unit initial noise scale
- PRXPixelPipeline: thin subclass wiring the above together (vae optional/None, vae_scale_factor=1, 1024px default)
- conversion script support for the pixel checkpoint format
- registration in __init__ files + dummy objects, docs autodoc entry, fast pipeline tests

## Before submitting

## Who can review?
@dg845


src/diffusers/pipelines/prx/pipeline_prx.py
```diff
@@ -297,6 +303,20 @@ def __init__(
         self.text_preprocessor = TextPreprocessor()
         self.default_sample_size = default_sample_size
         self._guidance_scale = 1.0
+        # Max number of text tokens. When None, falls back to the tokenizer's own ``model_max_length``.
+        # Subclasses (e.g. the PRXPixel pipeline, whose Qwen tokenizer has a very large model_max_length)
+        # can pin this to the value used at training time.
+        self.tokenizer_max_length = None
+        # When True, prompts get only light cleaning (``basic_clean``) instead of the DeepFloyd ``clean_text``.
+        # Set by subclasses whose text encoder was trained without the heavy cleaning (e.g. the Qwen tower).
+        self.skip_text_cleaning = False
```
dg845: I think the `tokenizer_max_length` and `skip_text_cleaning` attributes make more sense as `__call__` arguments so that users can set them for each inference run.

src/diffusers/pipelines/prx/pipeline_prx.py
```diff
@@ -297,6 +303,20 @@ def __init__(
...
+        # What the transformer predicts. "flow_matching" -> velocity (consumed directly by the scheduler).
+        # "x_prediction_flow_matching" -> the clean sample x0, converted to velocity before each scheduler step
+        # (see "Back to Basics: Let Denoising Generative Models Denoise", https://arxiv.org/abs/2511.13720).
+        self.prediction_type = "flow_matching"
+        # Standard deviation of the initial noise. Some PRX variants train with a non-unit noise scale and must
+        # start sampling from `randn * noise_scale` to match the learned flow-matching trajectory.
+        self.noise_scale = 1.0
```
dg845: Since `prediction_type` and `noise_scale` sound like they are pipeline-level attributes, I think it makes sense to have them as `__init__` arguments (like what `PRXPixelPipeline` currently does for `noise_scale`) so that it's easier to instantiate pipelines which use different values for these attributes.

src/diffusers/pipelines/prx/pipeline_prx_pixel.py
```diff
@@ -0,0 +1,121 @@
...
+
+logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
+
+# PRXPixel is a 1024px model.
+PRX_PIXEL_DEFAULT_RESOLUTION = 1024
+# Number of text tokens used at training time (the Qwen tokenizer's own ``model_max_length`` is far larger).
+PRX_PIXEL_DEFAULT_MAX_TOKENS = 256
+
+
+class PRXPixelPipeline(PRXPipeline):
```
dg845: `PRXPixelPipeline` should not inherit directly from `PRXPipeline`. We should either inherit from `DiffusionPipeline` as usual and copy over any shared methods using `# Copied from`, or, since the VAE is already optional for `PRXPipeline` and the code changes to support pixel-space models are small, we could let `PRXPipeline` handle both latent-space and pixel-space models.

tests/pipelines/prx/test_pipeline_prx_pixel.py
```diff
@@ -0,0 +1,307 @@
...
+
+from ..pipeline_params import TEXT_TO_IMAGE_PARAMS
+from ..test_pipelines_common import PipelineTesterMixin
+
+
+@pytest.mark.xfail(
+    condition=is_transformers_version(">", "4.57.1"),
+    reason="See https://github.com/huggingface/diffusers/pull/12456#issuecomment-3424228544",
+    strict=False,
+)
```
dg845: I think we should update the tests so that they work on `transformers>4.57.1`. (If they already work for newer `transformers` versions, then we should remove this `xfail` mark.)

dg845 (commented):

src/diffusers/pipelines/prx/pipeline_prx.py
DavidBert: Done — both are now explicit `__call__` (and `encode_prompt`) arguments in `PRXPipeline`, defaulting to `None`/`False` and threaded through to `_tokenize_prompts`. The instance-attribute pattern has been removed.

For `PRXPixelPipeline`, since it is now standalone (see the reply to the inheritance comment below), the Qwen cleaning and 256-token budget are baked into its own `_tokenize_prompts` implementation, so these args do not appear in that pipeline at all.

src/diffusers/pipelines/prx/pipeline_prx.py
DavidBert: `prediction_type` has been removed entirely — `PRXPipeline` always predicts velocity and `PRXPixelPipeline` always predicts x0, so there is no need for a runtime flag; the x0→velocity conversion is baked inline in the pixel pipeline's `__call__`.

`noise_scale` is now a proper `__init__` arg of `PRXPixelPipeline` (default `2.0`), registered to config so it round-trips through `save_pretrained`/`from_pretrained`. `PRXPipeline` no longer carries it.

src/diffusers/pipelines/prx/pipeline_prx_pixel.py
DavidBert: Done — `PRXPixelPipeline` now inherits directly from `DiffusionPipeline`. Shared methods (`encode_prompt`, `_encode_prompt_standard`, `do_classifier_free_guidance`, `guidance_scale`) are copied with `# Copied from`. Methods that differ (`_tokenize_prompts`, `check_inputs`, `prepare_latents`, `__call__`) are reimplemented directly in the pixel pipeline.

A separate doc page `docs/source/en/api/pipelines/prx_pixel.md` has been added, and `prx.md` has been restored to its pre-PR state.

tests/pipelines/prx/test_pipeline_prx_pixel.py
DavidBert: Removed. The tests pass without it on the current transformers version. The condition (`> 4.57.1`) was a precaution for a pre-release window that has since closed.

src/diffusers/models/transformers/transformer_prx.py
```diff
@@ -666,6 +715,8 @@ def __init__(
         theta: int = 10000,
         time_factor: float = 1000.0,
         time_max_period: int = 10000,
+        bottleneck_size: Optional[int] = None,
```
dg845:
```suggestion
        bottleneck_size: int | None = None,
```
nit: use implicit type hints here, in line with the rest of the file.

src/diffusers/models/transformers/transformer_prx.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any
+from typing import Any, Optional
```
dg845:
```suggestion
from typing import Any
```
nit: follow up change to https://github.com/huggingface/diffusers/pull/13928#discussion_r3418021652.

src/diffusers/pipelines/prx/pipeline_prx_pixel.py
```diff
@@ -0,0 +1,681 @@
...
+    _callback_tensor_inputs = ["latents", "prompt_embeds"]
+    _optional_components = ["vae"]
+
+    def __init__(
+        self,
+        transformer: PRXTransformer2DModel,
+        scheduler: FlowMatchEulerDiscreteScheduler,
+        text_encoder: PreTrainedModel,
+        tokenizer: AutoTokenizer | PreTrainedTokenizerBase,
+        vae: AutoencoderKL | AutoencoderDC | None = None,
```
dg845: I think the `vae` should be removed rather than retained as an optional component for `PRXPixelPipeline`. My understanding is that if a non-trivial VAE is supplied with either `vae_scale_factor > 1` or `self.vae.config.latent_channels != self.transformer.config.in_channels`, the code will break.

tests/pipelines/prx/test_pipeline_prx_pixel.py
```diff
@@ -0,0 +1,312 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        # Ensure PRXPixelPipeline has an _execution_device property expected by __call__
+        if not isinstance(getattr(PRXPixelPipeline, "_execution_device", None), property):
+            try:
+                setattr(PRXPixelPipeline, "_execution_device", property(lambda self: torch.device("cpu")))
+            except Exception:
+                pass
+
```
dg845:
```suggestion
```
I think we can remove the `setUpClass` method as `DiffusionPipeline` already defines a `_execution_device` property, and all of the tests pass without the method.

docs/source/en/_toctree.yml
```diff
@@ -591,6 +587,8 @@
         title: PixArt-Σ
       - local: api/pipelines/prx
         title: PRX
+      - local: api/pipelines/prx_pixel
+        title: PRX Pixel
       - local: api/pipelines/qwenimage
```
dg845, line 592: Can you add documentation for `PRXPixelPipeline`? The `api/pipelines/prx_pixel` file is referenced here but hasn't been added yet, which causes an error in [this CI run](https://github.com/huggingface/diffusers/actions/runs/27550420134/job/81562112940?pr=13928).

dg845 (commented):
Thanks for iterating! Left some follow up comments.

src/diffusers/models/transformers/transformer_prx.py
DavidBert: Done, changed to `int | None = None`.

src/diffusers/models/transformers/transformer_prx.py
DavidBert: Done, removed `Optional` from the import.

src/diffusers/pipelines/prx/pipeline_prx_pixel.py
DavidBert: Good catch — agreed. `vae` has been removed entirely from `PRXPixelPipeline` (the `__init__` signature, `register_modules`, `_optional_components`, and `model_cpu_offload_seq`). `vae_scale_factor` is now a simple property returning 1. The Hub checkpoint's `model_index.json` also needs the `"vae": [null, null]` entry removed so `from_pretrained` continues to work — updating that separately.

tests/pipelines/prx/test_pipeline_prx_pixel.py
DavidBert: Done, removed.

docs/source/en/_toctree.yml
DavidBert: The file was created locally but accidentally left untracked and never committed — fixed, it is now included in the branch.

dg845 (approved):
Thanks for your work on this PR!

PR #12915, opened 2026-01-06
Add LTX 2.0 Video Pipelines

dg845:
# What does this PR do?

This PR adds pipelines for the LTX 2.0 video generation model ([code](https://github.com/Lightricks/LTX-2), [weights](https://huggingface.co/Lightricks/LTX-2)). LTX 2.0 is an audio-video foundation model that generates videos with synced audio; it supports generation tasks such as text-to-video (T2V), text-image-to-video (TI2V), and more.

An example usage script for I2V is as follows:

```python
import torch
from diffusers.pipelines.ltx2 import LTX2ImageToVideoPipeline
from diffusers.pipelines.ltx2.export_utils import encode_video
from diffusers.utils import load_image

pipe = LTX2ImageToVideoPipeline.from_pretrained("Lightricks/LTX-2", torch_dtype=torch.bfloat16)
pipe.enable_model_cpu_offload()

image = load_image(
    "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/diffusers/astronaut.jpg"
)
prompt = "An astronaut hatches from a fragile egg on the surface of the Moon, the shell cracking and peeling apart in gentle low-gravity motion. Fine lunar dust lifts and drifts outward with each movement, floating in slow arcs before settling back onto the ground. The astronaut pushes free in a deliberate, weightless motion, small fragments of the egg tumbling and spinning through the air. In the background, the deep darkness of space subtly shifts as stars glide with the camera's movement, emphasizing vast depth and scale. The camera performs a smooth, cinematic slow push-in, with natural parallax between the foreground dust, the astronaut, and the distant starfield. Ultra-realistic detail, physically accurate low-gravity motion, cinematic lighting, and a breath-taking, movie-like shot."
negative_prompt = "shaky, glitchy, low quality, worst quality, deformed, distorted, disfigured, motion smear, motion artifacts, fused fingers, bad anatomy, weird hand, ugly, transition, static."

frame_rate = 24.0
video, audio = pipe(
    image=image,
    prompt=prompt,
    negative_prompt=negative_prompt,
    width=768,
    height=512,
    num_frames=121,
    frame_rate=frame_rate,
    num_inference_steps=40,
    guidance_scale=4.0,
    output_type="np",
    return_dict=False,
)
video = (video * 255).round().astype("uint8")
video = torch.from_numpy(video)

encode_video(
    video[0],
    fps=frame_rate,
    audio=audio[0].float().cpu(),
[...]

PR #13931, opened 2026-06-12
Fix LTX2 connector token/register layout (regression from #13564)

### LINKED ISSUE #13930: LTX2 text connectors pass reversed prompt tokens and misplaced registers to the transformer (regression from #13564)
### Describe the bug

`LTX2ConnectorTransformer1d.forward` (used by both the LTX 2.0 and LTX 2.3 pipelines) lays out its input sequence incorrectly before running the connector blocks: the prompt tokens reach the transformer **in reversed order**, and the learnable registers that fill the rest of the sequence are placed at the wrong positions.

The reference implementation ([`ltx_core` `_replace_padded_with_learnable_registers`](https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-core/src/ltx_core/text_encoders/gemma/embeddings_connector.py), matched by ComfyUI) front-aligns the valid tokens *preserving their order* and fills the tail with the tiled registers *indexed by absolute position*. The connector blocks apply RoPE, so the layout is part of what the model was trained on.

Toy example — 8 slots, 3 valid tokens `t1 t2 t3` (left-padded), register tile `R0 R1 R2 R3`:

```
reference (ltx_core / ComfyUI):  [t1 t2 t3 | R3 R0 R1 R2 R3]
diffusers main:                  [t3 t2 t1 | R0 R3 R2 R1 R0]
```

Even a full-length prompt (no padding at all) is reversed.

### Where it was introduced

#12915 originally ported this correctly (per-row boolean-mask gather). #13564 (`ebaa1871`, merged May 8) replaced that gather — which forces a GPU→CPU sync due to data-dependent indexing — with a vectorized masked-write followed by `torch.flip(hidden_states, dims=[1])`. The flip does move the embeddings to the front (as its comment intends), but it also reverses the token order and the register tile. The regression is on `main` only; v0.38.0 (May 1) predates it.

### Impact

Measured with the LTX-2.3 checkpoint (`diffusers/LTX-2.3-Diffusers` connectors) on real prompts: the post-connector text embeddings produced by `main` correlate with the reference layout's output at only **0.11–0.34 in the token region** (0.38–0.39 for the audio context). Short prompts — typically the *negative* prompt, whose 1024-slot context is mostly registers — are distorted the worst, so classifier-free guidance is hit hardest. After restoring the reference layout, the connector output matches ComfyUI's independent implementation of the same checkpoint at correlation 1.000.

### Reproduction

```python
import torch

S, L = 8, 3
tokens = torch.arange(1, L + 1).float()
regs = torch.arange(4).float()                # register tile R0..R3
tiled = regs.repeat(S // 4)

hs = torch.cat([torch.zeros(S - L), tokens])  # left-padded
mask = torch.cat([torch.zeros(S - L), torch.ones(L)])

# reference: gather valid tokens in order, registers by absolute position
reference = torch.cat([tokens, tiled[L:]])

# what main's connector layout produces (masked write + flip)
current = torch.flip(mask * hs + (1 - mask) * tiled, dims=[0])

print(reference.tolist())  # [1, 2, 3, 3, 0, 1, 2, 3]
print(current.tolist())    # [3, 2, 1, 0, 3, 2, 1, 0]
```

A fix that keeps #13564's sync-free goal (stable argsort + gather, all fixed-shape device ops) is in #PENDING — opening it alongside this issue.

### System Info

diffusers `main` (0.39.0.dev0); any platform.

### Who can help?

@dg845 @sayakpaul

Boffee:
# What does this PR do?

Fixes #13930.

## The problem

`LTX2ConnectorTransformer1d.forward` replaces the padding slots of the (left-padded) text sequence with learnable registers and then front-aligns the embeddings with `torch.flip(hidden_states, dims=[1])`. The flip does move the embeddings to the front, but it also **reverses the token order and the register tile**. The connector blocks apply RoPE, so this layout is part of what the LTX models were trained on — the original implementation ([`ltx_core` `_replace_padded_with_learnable_registers`](https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-core/src/ltx_core/text_encoders/gemma/embeddings_connector.py), matched by ComfyUI) front-aligns the valid tokens *in their original order* and fills the tail with the tiled registers *by absolute position*.

Toy example — 8 slots, 3 valid tokens `t1 t2 t3` (left-padded), register tile `R0 R1 R2 R3`:

```
reference (ltx_core / ComfyUI):  [t1 t2 t3 | R3 R0 R1 R2 R3]
main (flip):                     [t3 t2 t1 | R0 R3 R2 R1 R0]
```

Even a full-length prompt with no padding comes out reversed. Short prompts (typically the *negative* prompt, whose 1024-slot context is mostly registers) are distorted the worst, so CFG quality is hit hardest. Measured with the real `diffusers/LTX-2.3-Diffusers` connector weights, `main`'s output correlates with the reference layout's output at only **0.11–0.34 in the token region**; with this fix the output matches ComfyUI's independent implementation of the same checkpoint at correlation **1.000**.

## Where it was introduced

#12915 ported this correctly (per-row boolean-mask gather). #13564 (`ebaa1871`) replaced the gather — which forces a GPU→CPU sync due to data-dependent indexing — with the vectorized masked-write + flip, unintentionally changing the semantics. The regression is on `main` only (v0.38.0 predates it), so fixing it now keeps it out of the next release.

## The fix

Replace the masked-write + flip with a **stable argsort of the inverted mask + `torch.gather` + `torch.where`**: valid tokens move to the front in original order (stable sort preserves relative order), registers fill the tail by absolute position, computed per batch row (the pipelines batch negative+positive prompts of different lengths). All ops are fixed-shape device ops, so #13564's sync-elimination goal is preserved — no data-dependent indexing, no host sync.

## Tests

`tests/pipelines/ltx2/test_ltx2_connectors.py` checks the exact layout semantics through the module forward (`num_layers=0` reduces forward to layout + final RMSNorm) for left-padded, mixed-length batch, fully-valid, and single-token inputs. The tests fail on `main` and pass with the fix:

```
$ python -m pytest tests/pipelines/ltx2/test_ltx2_connectors.py -q   # on main
4 failed in 0.40s

$ python -m pytest tests/pipelines/ltx2/test_ltx2_connectors.py -q   # with this PR
4 passed in 0.57s

$ python -m pytest tests/pipelines/ltx2/ -q                          # with this PR
169 passed, 5 skipped, 34 warnings in 519.62s (0:08:39)
```

## Before submitting
- [x] Was this discussed via a GitHub issue? #13930

This investigation and fix were developed with AI assistance (Claude Code) and verified end-to-end as described above and in #13930.

## Who can review?

@dg845 @sayakpaul


dg845 (approved):
Thanks for the PR!

PR #13936, opened 2026-06-13
Log WanAnimate interpolation resizing as a warning

wunianze666-netizen:
## Summary
- replace the direct interpolation `print(...)` in `WanAnimatePipeline.prepare_prev_segment_cond_latents()` with `logger.warning(...)`
- add a regression test that captures the warning when previous-segment conditioning frames are resized

## Why
This path runs inside library code, so writing directly to stdout makes the message harder to control in downstream applications. The rest of the pipeline already uses structured warnings for similar resizing notices.

## Validation
- `python -m pytest tests/pipelines/wan/test_wan_animate.py -k test_prepare_prev_segment_cond_latents_logs_interpolation_warning -q`
- `ruff check src/diffusers/pipelines/wan/pipeline_wan_animate.py tests/pipelines/wan/test_wan_animate.py`
- `ruff format --check src/diffusers/pipelines/wan/pipeline_wan_animate.py tests/pipelines/wan/test_wan_animate.py`
- `python -m py_compile src/diffusers/pipelines/wan/pipeline_wan_animate.py tests/pipelines/wan/test_wan_animate.py`

## Issue Linkage
This is a small logging cleanup and is not tied to an existing tracked issue.

### DIFF
--- a/src/diffusers/pipelines/longcat_image/pipeline_longcat_image.py
+++ b/src/diffusers/pipelines/longcat_image/pipeline_longcat_image.py
@@ -110,7 +110,7 @@ def prepare_pos_ids(modality_id=0, type="text", start=(0, 0), num_token=None, he
     if type == "text":
         assert num_token
         if height or width:
-            print('Warning: The parameters of height and width will be ignored in "text" type.')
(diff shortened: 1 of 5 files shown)


wunianze666-netizen:

PR #13939, opened 2026-06-13
Log VQ remapping configuration through the model logger

wunianze666-netizen:
## Summary
- route `VectorQuantizer` remap configuration output through the model logger instead of writing directly to stdout
- add a regression test that captures the remap message when remapping is enabled

## Why
This message is emitted from library/model initialization code. Sending it to stdout makes it harder for downstream users to manage output, while the logger keeps the signal available without forcing terminal noise.

## Validation
- `python -m pytest tests/models/autoencoders/test_models_vq.py -k test_vector_quantizer_logs_remap_configuration -q`
- `ruff check src/diffusers/models/autoencoders/vae.py tests/models/autoencoders/test_models_vq.py`
- `ruff format --check src/diffusers/models/autoencoders/vae.py tests/models/autoencoders/test_models_vq.py`
- `python -m py_compile src/diffusers/models/autoencoders/vae.py tests/models/autoencoders/test_models_vq.py`

## Issue Linkage
This is a small logging cleanup and is not tied to an existing tracked issue.

### DIFF
--- a/src/diffusers/models/autoencoders/vae.py
+++ b/src/diffusers/models/autoencoders/vae.py
@@ -17,7 +17,7 @@ import numpy as np
 import torch
 import torch.nn as nn
 
-from ...utils import BaseOutput
+from ...utils import BaseOutput, logging
 from ...utils.torch_utils import randn_tensor
 from ..activations import get_activation
 from ..attention_processor import SpatialNorm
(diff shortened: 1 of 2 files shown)


wunianze666-netizen:

PR #13940, opened 2026-06-13
Fix ZImageControlNetModel.from_transformer sharing weights with the base transformer

### LINKED ISSUE #13077: ZImageControlNet.from_transformer creates a shallow copy of the transformer weights

HaozheZhang6:
# What does this PR do?

`ZImageControlNetModel.from_transformer` assigns the transformer's `t_embedder`, `all_x_embedder`, `cap_embedder`, `rope_embedder`, `noise_refiner`, `context_refiner` and the `x_pad_token` / `cap_pad_token` parameters straight onto the control net. They get registered as the control net's own submodules, so anything done to the control net later also touches the transformer: `controlnet.to(dtype/device)` or an optimizer over `controlnet.parameters()` mutates the base transformer's weights in place.

This deep-copies the reused components so the control net owns independent weights, the same way `FluxControlNetModel` holds its own copies. `t_scale` is a float and stays a plain assignment.

There was no test under `tests/models/controlnets/` for this model, so I added one for `from_transformer`: it checks the carried-over modules are independent objects, start out as exact copies, and that mutating the control net leaves the transformer untouched. Fails on `main`, passes here.

A larger `__init__` refactor is also possible (build these submodules and load them via `state_dict`, like Flux); happy to do that instead if you'd prefer. I kept this one minimal.

Fixes #13077

## Before submitting
- [x] Was this discussed via a GitHub issue? Fixing reported issue #13077.

## Who can review?

@hlky — you added the Z-Image ControlNet in #12792, so this is your area.

### DIFF
--- a/src/diffusers/models/controlnets/controlnet_z_image.py
+++ b/src/diffusers/models/controlnets/controlnet_z_image.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
 import math
 from typing import Literal
 
@@ -517,15 +518,18 @@ class ZImageControlNetModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOrigi
 
     @classmethod
     def from_transformer(cls, controlnet, transformer):
+        # Deep-copy the reused transformer modules/params so the control net owns independent weights;
+        # a shallow assignment registers them on the control net and leaks later ``.to()``/optimizer
+        # updates back into the transformer.
         controlnet.t_scale = transformer.t_scale
-        controlnet.t_embedder = transformer.t_embedder
-        controlnet.all_x_embedder = transformer.all_x_embedder
-        controlnet.cap_embedder = transformer.cap_embedder
-        controlnet.rope_embedder = transformer.rope_embedder
-        controlnet.noise_refiner = transformer.noise_refiner
-        controlnet.context_refiner = transformer.context_refiner
-        controlnet.x_pad_token = transformer.x_pad_token
-        controlnet.cap_pad_token = transformer.cap_pad_token
+        controlnet.t_embedder = copy.deepcopy(transformer.t_embedder)
+        controlnet.all_x_embedder = copy.deepcopy(transformer.all_x_embedder)
+        controlnet.cap_embedder = copy.deepcopy(transformer.cap_embedder)
+        controlnet.rope_embedder = copy.deepcopy(transformer.rope_embedder)
+        controlnet.noise_refiner = copy.deepcopy(transformer.noise_refiner)
+        controlnet.context_refiner = copy.deepcopy(transformer.context_refiner)
+        controlnet.x_pad_token = copy.deepcopy(transformer.x_pad_token)
+        controlnet.cap_pad_token = copy.deepcopy(transformer.cap_pad_token)
         return controlnet
 
     @staticmethod
--- a/tests/models/controlnets/test_models_controlnet_z_image.py
+++ b/tests/models/controlnets/test_models_controlnet_z_image.py
@@ -1,0 +1,96 @@
+# coding=utf-8
+# Copyright 2025 HuggingFace Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import torch
+
+from diffusers import ZImageControlNetModel, ZImageTransformer2DModel
+
+
+def _get_tiny_transformer():
+    return ZImageTransformer2DModel(
+        all_patch_size=(2,),
+        all_f_patch_size=(1,),
+        in_channels=16,
+        dim=16,
+        n_layers=1,
+        n_refiner_layers=1,
+        n_heads=1,
+        n_kv_heads=2,
+        qk_norm=True,
+        cap_feat_dim=16,
+        rope_theta=256.0,
+        t_scale=1000.0,
+        axes_dims=[8, 4, 4],
+        axes_lens=[256, 32, 32],
+    )
+
+
+def _get_tiny_controlnet():
+    return ZImageControlNetModel(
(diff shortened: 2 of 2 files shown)


hlky:
@HaozheZhang6 

This was deliberate design decision, the original code had a variant of the transformer with controlnet inline, this did not fit with Diffusers design where controlnet are implemented as a standalone model.

For context, the controlnet part of that transformer variant in the original code uses the exact modules from the transformer, as such they cannot be trained independently, see this [comment](https://github.com/huggingface/diffusers/pull/13128#issuecomment-3890906666). So naively running `optimizer over controlnet.parameters()` as you mentioned is not correct, this is mirrored in the training script from the original repo that only trains modules starting with `control` prefix iirc.

As for the design decision itself, sharing the modules directly from the transformer saves vram compared to having the controlnet own copies.

Regarding `controlnet.to(dtype/device)` I don't believe this would happen in practice with diffusers pipelines, even with offloading etc. iirc the controlnet should always end up on the same device as transformer, and there is no practical use for running the controlnet or transformer in different dtypes. If you have discovered a concrete failure related to this then it should be looked at, but it appears that your concerns are more hypothetical.

I would also be cautious about accepting the existence of an issue to mean the issue itself is validated. This particular issue has already had several previous PRs attempting a similar change that have missed the surrounding architectural context.

If you are interested in contributing I would make a biased suggestion of tackling something from https://github.com/huggingface/diffusers/issues/13656. Those issues are already well scoped, can generally be considered validated and the maintainers would probably appreciate the help working through all that low hanging fruit.

On that note, I will delegate to a maintainer for final word on this PR: @yiyixuxu, as we discussed the design decision at the time it was implemented.

HaozheZhang6:
That makes sense, thanks for the context. I missed that the module sharing is intentional (saves VRAM, and only the `control*` modules are trained), and I don't have a concrete device/dtype failure to back up that part — so my concern was hypothetical. Closing this. I'll take a look at #13656.

Issue #13630, opened 2026-04-29
ovis_image model/pipeline review


hlky:
# `ovis_image` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules. `AGENTS.md` is referenced by `.ai/review-rules.md` but is not present in this checkout; the other referenced rule files were read and applied.

Files/categories reviewed: target pipeline/model files, public imports and lazy loading, config/serialization, dtype/device/offload paths, attention processor behavior, docs, and fast/slow test coverage.

Duplicate search status: `gh search` hit the GitHub API rate limit, so I checked GitHub web issue/PR searches for `OvisImage`, `ovis_image`, `OvisImageTransformer2DModel AttentionMixin`, and `OvisImagePipeline num_images_per_prompt`. I did not find an exact Ovis duplicate. Related but not duplicate: https://github.com/huggingface/diffusers/issues/12186 covers the same missing-`AttentionMixin` pattern for `WanVACETransformer3DModel`.

## Issue 1: Transformer does not expose attention processor APIs

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/models/transformers/transformer_ovis_image.py#L22-L28
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/models/transformers/transformer_ovis_image.py#L386-L392

Problem:
`OvisImageTransformer2DModel` defines `OvisImageAttention` modules but does not inherit `AttentionMixin`. That leaves the model without the standard `attn_processors`, `set_attn_processor`, `fuse_qkv_projections`, and `unfuse_qkv_projections` APIs expected by related transformer families.

Impact:
Users and tests cannot swap attention processors, inspect processors, or use QKV fusion through the model-level API.

Reproduction:
```python
from diffusers import OvisImageTransformer2DModel

print(hasattr(OvisImageTransformer2DModel, "set_attn_processor"))
print(hasattr(OvisImageTransformer2DModel, "fuse_qkv_projections"))
assert hasattr(OvisImageTransformer2DModel, "set_attn_processor")
```

Relevant precedent:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/models/transformers/transformer_flux.py#L28-L33
[...]

Issue #12186, opened 2025-08-19
`WanVACETransformer3DModel` does not inherit `AttentionMixin` while `WanTransformer3DModel` does


J4BEZ:
Hello,

While installing the latest version of diffusers from source and implementing a custom Wan VACE pipeline, I noticed a potential inconsistency that I hope may be helpful to point out.

`WanTransformer3DModel` inherits the following:
`ModelMixin`, `ConfigMixin`, `PeftAdapterMixin`, `FromOriginalModelMixin`, `CacheMixin`, and `AttentionMixin`. [link](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_wan.py#L497C1-L499C3)

However, `WanVACETransformer3DModel inherits` all of the above except for `AttentionMixin`. [link](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_wan_vace.py#L137)

This difference does not cause major issues for video generation or LoRA application, but when attempting to call the `transformer.set_attn_processor()` method in a custom pipeline, I encountered an error with `WanVACETransformer3DModel`, unlike `WanTransformer3DModel`.

Thank you very much for all the incredible work and dedication you put into maintaining the diffusers library, which makes it much easier for researchers and developers to enjoy working with generative models. Wishing you good health and a peaceful day ahead!

Regards, J4BEZ

DN6:
Hi @J4BEZ yes, we do need to refactor VACE to use the AttentionMixin. Would it be possible for you to open a PR for it?

J4BEZ:
@DN6 Yes, I would be glad to contribute to diffusers again.
Thank you very much for your kind reply😄

DN6:
Thanks @J4BEZ. Closing since #12268  was merged.

J4BEZ:
@DN6 Thank you! I hope you a peaceful day 😄

PR #13941, opened 2026-06-13
Add `AttentionMixin` to transformers that are missing it

HaozheZhang6:
# What does this PR do?

Several transformers build their attention with `AttentionModuleMixin` but don't inherit the model-level `AttentionMixin`, so `attn_processors` / `set_attn_processor` / `fuse_qkv_projections` / `unfuse_qkv_projections` raise `AttributeError`. This adds `AttentionMixin` to all of them, the same way `WanVACETransformer3DModel` was handled (#12186):

- `OvisImageTransformer2DModel` (the original #13630 finding)
- `BriaTransformer2DModel`, `BriaFiboTransformer2DModel`
- `AnyFlowTransformer3DModel`, `AnyFlowFARTransformer3DModel`
- `ErnieImageTransformer2DModel`
- `LongCatAudioDiTTransformer`

The attention modules already use `AttentionModuleMixin`, so the four APIs work without further changes and resolve to each model's own processor. Each model gets a regression test that `attn_processors` is populated and `set_attn_processor` round-trips.

Note: `AnyFlowTransformer3DModel`, `AnyFlowFARTransformer3DModel`, and `ErnieImageTransformer2DModel` have some unrelated pre-existing failures in their model test suites on `main` (not touched here); the new `test_attention_processor_api` passes for all seven.

Addresses the `AttentionMixin` pattern from #13656.

## Before submitting
- [x] Was this discussed via a GitHub issue? Yes — #13656 / #13630.
- [x] Did you write any new tests?

## Who can review?

@hlky — batched per your suggestion, from your model reviews in #13656.

### DIFF
--- a/src/diffusers/models/transformers/transformer_anyflow.py
+++ b/src/diffusers/models/transformers/transformer_anyflow.py
@@ -29,7 +29,7 @@ from ...configuration_utils import ConfigMixin, register_to_config
 from ...loaders import FromOriginalModelMixin, PeftAdapterMixin
 from ...utils import apply_lora_scale, logging
 from ...utils.torch_utils import maybe_adjust_dtype_for_device
-from ..attention import AttentionModuleMixin, FeedForward
+from ..attention import AttentionMixin, AttentionModuleMixin, FeedForward
 from ..attention_dispatch import dispatch_attention_fn
 from ..embeddings import PixArtAlphaTextProjection, TimestepEmbedding, Timesteps, get_1d_rotary_pos_embed
 from ..modeling_outputs import Transformer2DModelOutput
@@ -504,7 +504,7 @@ class AnyFlowTransformerBlock(nn.Module):
         return hidden_states
 
 
-class AnyFlowTransformer3DModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOriginalModelMixin):
+class AnyFlowTransformer3DModel(ModelMixin, ConfigMixin, PeftAdapterMixin, FromOriginalModelMixin, AttentionMixin):
     r"""
     Bidirectional 3D Transformer for AnyFlow flow-map sampling.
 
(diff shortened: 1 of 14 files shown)


HaozheZhang6:
While here I checked the other transformers that build their attention with `AttentionModuleMixin` but don't inherit the model-level `AttentionMixin` — the same gap is in `AnyFlowTransformer3DModel`, `AnyFlowFARTransformer3DModel`, `BriaTransformer2DModel`, `BriaFiboTransformer2DModel`, and `ErnieImageTransformer2DModel` (all raise `AttributeError` on `attn_processors`). Happy to fix those in a follow-up batch if this approach looks right.

hlky:
@HaozheZhang6 If you can identify a pattern from all the issues, then it is better to fix those all at once. `AttentionMixin` is a good case for that as the change is minimal. Other patterns where the change is more involved may be riskier to fix as a whole. There may be others beyond than those you found, I would either point your agent at the meta issue ask it to check all the sub-issues, or do another pass on repo as a whole. For this particular pattern the part I would exercise caution with is the tests, depending on maintainers view (I would ask @DN6 or @sayakpaul about this) it may be better to add the test to `ModelTesterMixin` rather than individually.

HaozheZhang6:
Makes sense — I'll consolidate them into this PR. I'll re-scan for the full set (including any beyond the five I listed) and add `AttentionMixin` + a regression test to each here, then close the separate Bria PR I'd just opened (#13943).

HaozheZhang6:
Done — all seven are in now: `OvisImage`, `Bria`, `BriaFibo`, `AnyFlow`, `AnyFlowFAR`, `ErnieImage`, and `LongCatAudioDiTTransformer` (that last one I'd missed in the first scan). Each gets `AttentionMixin` + a small regression test.

PR #13945, opened 2026-06-14
Warn and skip transformer LoRA loading when the transformer is absent (#13487)

### LINKED ISSUE #13487: Loading loras on "text_pipe" of a modular pipeline fails

archievi:
# What does this PR do?

Loading a transformer-only LoRA onto a Modular sub-pipeline that does not contain the transformer (for example a text-encoder-only `text_pipe`) currently crashes:

```
AttributeError: 'Flux2ModularPipeline' object has no attribute 'transformer'
```

This happens because `load_lora_weights` resolves the transformer with `getattr(self, self.transformer_name)`, which raises when the (sub-)pipeline has no transformer component.

Following @yiyixuxu's suggestion on the issue ("maybe warn and skip is better"), `load_lora_weights` now resolves the transformer with a `None` default and, when no transformer is present, logs a warning and skips loading the transformer LoRA layers instead of raising.

The change is made at the copied-from source (`CogVideoXLoraLoaderMixin.load_lora_weights`) and propagated to its 14 copies with `make fix-copies`, so the behavior is consistent across the transformer LoRA loaders (Flux2, Flux, SD3, CogView4, etc.). Normal pipelines that do have a transformer are unaffected.

Fixes #13487

## Before submitting
  - [x] Did you point it at the project conventions in [`.ai/`](https://github.com/huggingface/diffusers/tree/main/.ai)? I read and applied `.ai/AGENTS.md` (copied-code convention) and `.ai/modular.md`; the fix is made at the copied-from source and propagated with `make fix-copies` rather than editing individual copies.
  - [x] Did you self-review the diff against [`.ai/review-rules.md`](https://github.com/huggingface/diffusers/blob/main/.ai/review-rules.md)? No ephemeral context, correctness-focused, no dead code.
- [ ] Did you read our [philosophy doc](https://github.com/huggingface/diffusers/blob/main/PHILOSOPHY.md)? (small, localized fix)
- [x] Was this discussed/approved via a GitHub issue? Yes — #13487, where @yiyixuxu suggested the warn-and-skip approach and invited a PR.
- [ ] Did you make sure to update the documentation with your changes? (no user-facing API/doc change)
- [x] Did you write any new necessary tests? Added `Flux2LoRAModularPipelineTests` in `tests/lora/test_lora_layers_flux2.py`.

### Verification (run locally)
- New regression test passes, and it reproduces the original `AttributeError` when the fix is reverted (regression-proven).
- `python utils/check_copies.py` passes (copies consistent after `make fix-copies`).
- `ruff check` and `ruff format --check` are clean on the changed files.

## Who can review?

@yiyixuxu @sayakpaul @DN6


src/diffusers/loaders/lora_pipeline.py
```diff
@@ -1381,9 +1381,18 @@ def load_lora_weights(
         if not is_correct_format:
             raise ValueError("Invalid LoRA checkpoint. Make sure all LoRA param names contain `'lora'` substring.")
 
+        transformer = self.transformer if hasattr(self, "transformer") else getattr(self, self.transformer_name, None)
+        if transformer is None:
+            logger.warning(
+                f"No `{self.transformer_name}` module was found in {self.__class__.__name__}, so no LoRA weights will be "
+                "loaded into the transformer. This can happen when loading LoRA weights into a modular pipeline that "
+                "does not contain the transformer (for example, a text-encoder-only sub-pipeline)."
+            )
```
sayakpaul, line 1400: This is not specific to a modular pipeline, though.

tests/lora/test_lora_layers_flux2.py
```diff
@@ -148,3 +148,33 @@ def test_simple_inference_with_text_denoiser_block_scale_for_all_dict_options(se
     @unittest.skip("Not supported in Flux2.")
     def test_modify_padding_mode(self):
         pass
+
+
+@require_peft_backend
+class Flux2LoRAModularPipelineTests(unittest.TestCase):
```
sayakpaul: It should be added to https://github.com/huggingface/diffusers/blob/9b249df46dd1eb2bdabf6124c3ebeccfb41201e6/tests/lora/utils.py#L123 as I don't think the underlying issue _only_ applies to modular pipelines.

sayakpaul (commented):
Thanks! I left some comments.

src/diffusers/loaders/lora_pipeline.py
archievi: Good call — reworded to drop the modular framing. Digging in, I realized the `getattr(self, self.transformer_name)` crash was not limited to the copied-from group: SD3, Flux, Wan (+SkyReelsV2 via copies), LTX2, Helios and Anima each define their own `load_lora_weights` and still raised. Extended warn-and-skip to all of them. For the loaders that also load text-encoder LoRAs (SD3/Flux/Anima) it skips only the transformer part and still loads the text-encoder layers rather than returning early.

tests/lora/test_lora_layers_flux2.py
archievi: Moved it — the regression test now lives in `PeftLoraLoaderMixinTests` (`tests/lora/utils.py`) as `test_load_lora_weights_without_transformer_warns_and_skips`, so it runs for every transformer-based loader (skipped for UNet pipelines). It builds the dummy pipeline, nulls out the transformer to emulate a sub-pipeline without one, and asserts we warn and skip instead of raising. Dropped the Flux2-specific class.

archievi:
Rebased on `main` and fixed CI. The `check_repository_consistency` failure was because `main` added `Krea2LoraLoaderMixin` and `Ideogram4LoraLoaderMixin`, both `# Copied from` `CogVideoXLoraLoaderMixin.load_lora_weights` — so this PR's warn-and-skip guard needed propagating to them. Merged `main` (no conflicts) and ran `make fix-copies`; the full repo-consistency suite passes locally against the pinned `ruff==0.9.10`. Ready for another look.

Issue #13584, opened 2026-04-29
flux model/pipeline review


hlky:
# `flux` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

Duplicate search status: searched current GitHub Issues and PRs for `flux`, affected class/function names, and failure modes. No likely duplicates found. Broader `gh` searches hit rate limits after targeted searches, so remaining modular-pipeline searches were checked through the GitHub connector; related PRs such as `#12272` and `#13482` are not duplicates.

## Issue 1: Flux IP-Adapter masks are accepted but ignored

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/models/transformers/transformer_flux.py#L183-L268
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/models/transformers/transformer_flux.py#L345-L346

Problem:
`FluxIPAdapterAttnProcessor.__call__` accepts `ip_adapter_masks`, and `FluxAttention.forward` explicitly quiets that kwarg, but the processor never reads or applies the masks.

Impact:
Region-specific IP-Adapter conditioning silently has no effect for Flux. Users can pass masks without warnings, making masked IP-Adapter results misleading.

Reproduction:
```python
import torch
from diffusers.models.transformers.transformer_flux import FluxAttention, FluxIPAdapterAttnProcessor

torch.manual_seed(0)
attn = FluxAttention(
    query_dim=4,
    heads=1,
    dim_head=4,
    added_kv_proj_dim=4,
    processor=FluxIPAdapterAttnProcessor(hidden_size=4, cross_attention_dim=4, num_tokens=(2,), scale=10.0),
)

hidden = torch.randn(1, 3, 4)
encoder = torch.randn(1, 2, 4)
ip = [torch.randn(1, 2, 4)]
mask0 = [torch.zeros(1, 3, 1, 1)]
mask1 = [torch.ones(1, 3, 1, 1)]

out0 = attn(hidden, encoder_hidden_states=encoder, ip_hidden_states=ip, ip_adapter_masks=mask0)[0]
out1 = attn(hidden, encoder_hidden_states=encoder, ip_hidden_states=ip, ip_adapter_masks=mask1)[0]
print(torch.max(torch.abs(out0 - out1)).item())  # 0.0
```

[...]

PR #13955, opened 2026-06-14
fix(flux): tighten check_inputs validation

akshan-main:
# What does this PR do?

Addresses part of the flux review (#13584) - three input-validation gaps in the flux pipelines:

- **Issue 3:** `FluxPipeline` / `FluxKontextPipeline` accepted `negative_prompt_embeds` with a different shape than `prompt_embeds`. Added the shape check the img2img/inpaint variants already have (propagated to kontext via `# Copied from`).
- **Issue 4:** `FluxControlNetImg2ImgPipeline` checked `height % self.vae_scale_factor * 2`, which parses as `(height % vae_scale_factor) * 2`, so invalid sizes like 72 passed. Added the parentheses to match every other flux pipeline.
- **Issue 5:** `FluxPriorReduxPipeline` only validated scale length for list images and ignored `pooled_prompt_embeds_scale`. Now computes the image batch size for tensors too and validates both scales.

Verified each finding's repro now raises/flags, and that matching/valid inputs still pass.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? (Part of #13656, finding from #13584) and discussed on slack with @sayakpaul

## Who can review?

@hlky @sayakpaul @yiyixuxu


src/diffusers/pipelines/flux/pipeline_flux_controlnet_image_to_image.py
```diff
@@ -454,7 +454,7 @@ def check_inputs(
         if strength < 0 or strength > 1:
             raise ValueError(f"The value of strength should in [0.0, 1.0] but is {strength}")
 
-        if height % self.vae_scale_factor * 2 != 0 or width % self.vae_scale_factor * 2 != 0:
+        if height % (self.vae_scale_factor * 2) != 0 or width % (self.vae_scale_factor * 2) != 0:
```
sayakpaul, line 457: What value is this adding though?

src/diffusers/pipelines/flux/pipeline_flux.py
```diff
@@ -491,6 +491,14 @@ def check_inputs(
...
             )
 
+        if prompt_embeds is not None and negative_prompt_embeds is not None:
+            if prompt_embeds.shape != negative_prompt_embeds.shape:
+                raise ValueError(
+                    "`prompt_embeds` and `negative_prompt_embeds` must have the same shape when passed directly, but"
+                    f" got: `prompt_embeds` {prompt_embeds.shape} != `negative_prompt_embeds`"
+                    f" {negative_prompt_embeds.shape}."
+                )
+
```
sayakpaul: I think this should also be checked when `do_true_cfg` is True? Otherwise, it's not used right?

sayakpaul (commented):
Left some comments.

src/diffusers/pipelines/flux/pipeline_flux_controlnet_image_to_image.py
akshan-main: For consistency, all flux pipelines have the same 2 * (height // (vae_scale_factor * 2)) round-down in prepare_latents, but the others raise in check_inputs first, so an invalid size like 72 errors there and never reaches it. This is the only flux pipeline that silently resizes to 64 instead of erroring. Precedencies!

src/diffusers/pipelines/flux/pipeline_flux.py
akshan-main: moved it into the do_true_cfg branch in both base flux and kontext, so the shape check only runs when the negative embeds are used.

akshan-main: fixed!

tests/pipelines/flux/test_pipeline_flux.py
```diff
@@ -234,6 +234,25 @@ def test_flux_true_cfg(self):
...
+            "negative_prompt_embeds": torch.randn(1, 5, 32, device=torch_device),
+            "negative_pooled_prompt_embeds": torch.randn(1, 32, device=torch_device),
+            "height": 16,
+            "width": 16,
+            "num_inference_steps": 1,
+            "output_type": "latent",
+        }
+
+        with self.assertRaises(ValueError):
+            pipe(**base_inputs, true_cfg_scale=2.0, generator=torch.manual_seed(0))
```
sayakpaul: Let's also inspect the message and assert around it.

sayakpaul (approved):
Thanks! Left one comment.

sayakpaul (approved):

PR #13961, opened 2026-06-15
[tests] component manager tests

sayakpaul:
# What does this PR do?

Adds tests for the component manager in modular pipelines.

* Correctness-related checks (offloading effectivity, component eviction, etc.). These tests are in `ComponentsManagerTesterMixin`.
* Checks when the component manager is used in actual modular pipelines. These are in `ModularPipelineOffloadTesterMixin`. I have only used it in the Flux pipeline tests to keep things minimal.

**Notes**:

We're mocking the available memory and free memory in the tests and checking against those simulations. So, in many cases we don't actually need a real GPU or a GPU with a certain amount of memory.

In the next PR, I want to introduce the `maximum_total_models_size` flag (and related tests).


yiyixuxu (approved):
thanks!

PR #13974, opened 2026-06-17
[tests] port final set of model tests and others

sayakpaul:
# What does this PR do?

* Port the remaining model tests to use our new mixins and remove the monolithic `ModelTesterMixin` class.
* Port the remaining standalone testing classes to use `pytest`.

I have run all the GPU tests, too (and they pass).


sayakpaul:
Failing tests are also exist on `main`.

tests/models/transformers/test_models_transformer_sana_video.py
```diff
@@ -82,16 +73,44 @@ def prepare_init_args_and_inputs_for_common(self):
...
+                (batch_size, sequence_length, text_encoder_embedding_dim),
+                generator=self.generator,
+                device=torch_device,
+            ),
+            "timestep": torch.randint(0, 1000, size=(batch_size,), generator=self.generator).to(torch_device),
+        }
 
 
-class SanaVideoTransformerCompileTests(TorchCompileTesterMixin, unittest.TestCase):
-    model_class = SanaVideoTransformer3DModel
```
dg845, line 94: It looks like `SanaVideoTransformer3DModel` previously had a `TorchCompileTesterMixin` test subclass that got removed, would it be possible to add it back?

tests/models/transformers/test_models_transformer_consisid.py
```diff
@@ -97,9 +82,36 @@ def prepare_init_args_and_inputs_for_common(self):
...
+
+class TestConsisIDTransformer(ConsisIDTransformerTesterConfig, ModelTesterMixin):
+    pass
+
+
+class TestConsisIDTransformerMemory(ConsisIDTransformerTesterConfig, MemoryTesterMixin):
+    pass
+
+
+class TestConsisIDTransformerTraining(ConsisIDTransformerTesterConfig, TrainingTesterMixin):
```
dg845, line 114: I think `ConsisIDTransformer3DModel` is the only model that didn't get an `AttentionTesterMixin` test subclass, is this intentional?

tests/models/transformers/test_models_transformer_cogvideox.py
```diff
@@ -81,49 +74,36 @@ def prepare_init_args_and_inputs_for_common(self):
...
+class CogVideoX15TransformerTesterConfig(CogVideoXTransformerTesterConfig):
     @property
-    def input_shape(self):
-        return (1, 4, 8, 8)
+    def output_shape(self) -> tuple:
+        return (2, 4, 8, 8)
 
     @property
-    def output_shape(self):
-        return (1, 4, 8, 8)
```
dg845, line 123: It looks like the CogVideoX-1.5 test `output_shape` was changed from `(1, 4, 8, 8)` to `(2, 4, 8, 8)`, is this intentional?

tests/models/transformers/test_models_transformer_sana_video.py
sayakpaul: I intentionally didn't add it because it's not used much.

tests/models/test_modeling_common.py
```diff
@@ -13,225 +13,46 @@
...
-class ModelUtilsTest(unittest.TestCase):
-    def tearDown(self):
-        super().tearDown()
-
+class TestModelUtils:
     def test_missing_key_loading_warning_message(self):
-        with self.assertLogs("diffusers.models.modeling_utils", level="WARNING") as logs:
+        logger = logging.getLogger("diffusers.models.modeling_utils")
+        with CaptureLogger(logger) as cap_logger:
             UNet2DConditionModel.from_pretrained("hf-internal-testing/stable-diffusion-broken", subfolder="unet")
```
dg845, line 43: Would it be better to use `pytest`'s `caplog` fixture here?

tests/models/transformers/test_models_transformer_consisid.py
sayakpaul: `ConsisIDTransformer3DModel` uses a custom attention processor and in the old suite it had `uses_custom_attn_processor` = True, which exempted it from the attention-processor tests.

dg845 (approved):
Thanks for the PR! Left some comments :).

tests/models/transformers/test_models_transformer_cogvideox.py
sayakpaul: 1.5 config's dummy input uses `num_frames=2` (vs `num_frames=1` for base CogVideoX), and the transformer outputs the same frame count, so the true per-sample output is `(2, 4, 8, 8)`. 

The old `(1, 4, 8, 8)` was actually inconsistent with its own `num_frames=2` input and it never tripped a test because the old `test_output` checked `output.shape` against the input tensor, not the declared `output_shape`; the declared value was only used to build the training-loss noise, where `(2,1,4,8,8)` happened to broadcast against the `(2,2,4,8,8)` output. I set it to the model's actual output shape so the loss target is built correctly instead of relying on broadcasting.

PR #13965, opened 2026-06-15
ci: don't remind on prs from admins, etc.

sayakpaul:
# What does this PR do?

We shouldn't remind maintainers, collaborators, and folks with write access. Otherwise, [this](https://github.com/huggingface/diffusers/pull/13921#issuecomment-4698011053) becomes annoying.

PR #13976, opened 2026-06-17
[ci] don't remind model authors about linking issues/.

sayakpaul:


sayakpaul:
https://github.com/huggingface/diffusers/pull/13965

PR #13978, opened 2026-06-17
Support batched multi-controlnet conditioning in SDXL ControlNet inpaint pipeline

### LINKED ISSUE #9511: Multi-controlnet batching for `StableDiffusionXLControlNetInpaintPipeline`
Currently, batching is not supported when we are conditioning the SDXL pipeline on multiple controlnets: https://github.com/huggingface/diffusers/blob/28f9d84549c0b1d24ef00d69a4c723f3a11cffb6/src/diffusers/pipelines/controlnet/pipeline_controlnet_inpaint_sd_xl.py#L771

We can pass a list of lists of images for batch inference.

One-by-one processing, which results in very low GPU utilization.

abhay23-AI:
## What does this PR do?

Adds batched multi ControlNet conditioning to `StableDiffusionXLControlNetInpaintPipeline`, so you can pass `control_image` as a list of lists and run multiple ControlNets over a batch of prompts. This is the same feature that #6334 added to `StableDiffusionControlNetPipeline`, ported to the SDXL inpaint pipeline. Resolves #9511.

```python
# two prompts, each with one conditioning image per ControlNet
pipe(
    prompt=[prompt_a, prompt_b],
    image=init_image,
    mask_image=mask,
    control_image=[[canny_a, pose_a], [canny_b, pose_b]],
)
```

## How

Following #6334, the change is intentionally minimal and `check_image` is left untouched:

1. In `check_inputs`, the nested list guard is replaced with a transpose `[list(t) for t in zip(*image)]` that regroups the conditioning images per ControlNet, then validates each sublist with the existing `check_image`.
2. The same transpose is applied in `__call__` right before the `prepare_control_image` loop, using the `control_image` variable.
3. The conditioning scale guard message is clarified (nested scales stay unsupported), and the now obsolete prompt list warning is removed, matching #6334.
4. The `control_image` docstring documents the list of lists behavior.

The inpaint mask path is unaffected: `init_image`, `mask`, and `masked_image_latents` are still batched to `batch_size * num_images_per_prompt` independently, and the transpose only regroups control images per ControlNet, so the batch dimension stays aligned with the UNet input.

## Tests

Added `StableDiffusionXLMultiControlNetInpaintPipelineFastTests` mirroring the existing SDXL multi ControlNet test class, with `test_inference_multiple_prompt_input` covering the batched conditioning case (asserts shape and that batch items differ) and the single set broadcast case (asserts it matches the batched output). Runs on CPU.

I also ran `make style`, `make quality`, and `utils/check_copies.py` locally, all clean. `prepare_control_image` is copie

## Who can review?

@yiyixuxu @sayakpaul since you reviewed and merged the original #6334.


abhay23-AI:
hi @yiyixuxu, gentle nudge on this when you have a moment. it ports the batched multi controlnet conditioning from your merged #6334 over to the SDXL inpaint pipeline, same transpose approach with check_image left untouched, plus a cpu test covering the batched and broadcast cases. all checks are green. thanks!

Issue #13618, opened 2026-04-29
bria_fibo model/pipeline review


hlky:
# `bria_fibo` model/pipeline review

Commit tested: `0f1abc4ae8b0eb2a3b40e82a310507281144c423`

Review performed against the repository review rules.

Duplicate search: checked GitHub issues/PRs for `bria_fibo`, `BriaFibo`, `FIBO`, affected class names, and failure modes. Found integration/refactor PRs `#12545`, `#12688`, `#12731`, `#12930`, `#13341`; no duplicate for the issues below. Public top-level imports succeeded.

Test note: standalone reproductions ran with `.venv`. Full target pytest collection failed before running tests because this Windows torch build lacks `torch._C._distributed_c10d`.

## Issue 1: `prompt_embeds` is public but unusable

Affected code:
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo.py#L254-L332
https://github.com/huggingface/diffusers/blob/0f1abc4ae8b0eb2a3b40e82a310507281144c423/src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo_edit.py#L412-L490

Problem:
Both pipelines expose `prompt_embeds`, but `encode_prompt()` only defines `prompt_layers` when it encodes `prompt` itself. Passing precomputed embeddings raises `UnboundLocalError`. `negative_prompt_embeds` is also not honored because negative embeddings are always recomputed when `guidance_scale > 1`.

Impact:
Users cannot use documented precomputed embedding workflows, prompt weighting, cached text encoder outputs, or callback-modified embeddings reliably.

Reproduction:
```python
import torch
from types import SimpleNamespace
from diffusers import BriaFiboPipeline

pipe = BriaFiboPipeline.__new__(BriaFiboPipeline)
pipe.transformer = SimpleNamespace(dtype=torch.float32)
pipe.text_encoder = SimpleNamespace(dtype=torch.float32)

pipe.encode_prompt(
    prompt=None,
    prompt_embeds=torch.zeros(1, 1, 64),
    guidance_scale=1.0,
    device=torch.device("cpu"),
)
```

Relevant precedent:
Flux validates all required precomputed conditioning inputs together:
[...]

PR #13981, opened 2026-06-17
fix(bria_fibo): fix guidance_embeds, prompt_embeds, tensor-image and multi-image crashes

akshan-main:
# What does this PR do?

Part of #13618 (`bria_fibo` review). Fixes four runtime crashes:

- `guidance_embeds=True` couldn't construct the transformer: the guidance embedder was built without the required `time_theta`, and the forward tested a tensor with `if guidance:`.
- `prompt_embeds` / `negative_prompt_embeds` were public inputs that can't work standalone: the transformer also needs the per-layer embeddings, which only come from encoding `prompt`. Removed both args from both pipelines.
- `BriaFiboEditPipeline` crashed on tensor `image` inputs, which referenced an undefined `self.latent_channels`.
- `num_images_per_prompt > 1` returned a malformed output shape (extra batch axis for numpy, nested lists for PIL). The decode now runs once over the full latent batch and post-processes once, which gives the right shape.

Verified each against the repros in the issue.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? Discussed on Slack with @yiyixuxu

## Who can review?

@yiyixuxu @sayakpaul

### DIFF
--- a/src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo.py
+++ b/src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo.py
@@ -260,6 +260,11 @@ class BriaFiboPipeline(DiffusionPipeline, FluxLoraLoaderMixin):
             )
             prompt_embeds = prompt_embeds.to(dtype=self.transformer.dtype)
             prompt_layers = [tensor.to(dtype=self.transformer.dtype) for tensor in prompt_layers]
+        else:
+            raise ValueError(
+                "`prompt_embeds` cannot be passed on its own; this pipeline also needs the per-layer embeddings "
+                "computed from `prompt`. Please pass `prompt` instead."
+            )
 
         if guidance_scale > 1:
             if isinstance(negative_prompt, list) and negative_prompt[0] is None:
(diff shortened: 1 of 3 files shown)


src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo.py
```diff
@@ -260,6 +260,11 @@ def encode_prompt(
             )
             prompt_embeds = prompt_embeds.to(dtype=self.transformer.dtype)
             prompt_layers = [tensor.to(dtype=self.transformer.dtype) for tensor in prompt_layers]
+        else:
```
yiyixuxu: let's just remove this argument?

src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo.py
```diff
@@ -773,10 +778,11 @@ def __call__(
             for scaled_latent in latents_scaled:
                 curr_image = self.vae.decode(scaled_latent.unsqueeze(0), return_dict=False)[0]
                 curr_image = self.image_processor.postprocess(curr_image.squeeze(dim=2), output_type=output_type)
-                image.append(curr_image)
```
yiyixuxu, line 776: Is there any reason that they have to decode and post-process latent one each time?

yiyixuxu (commented):
thanks, i left some comments

src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo.py
akshan-main: Done! removed both `prompt_embeds` and `negative_prompt_embeds` (the latter was recomputed from `negative_prompt` and ignored anyway).

src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo.py
akshan-main: No, they're already a single batched tensor, so switched to one decode + postprocess

yiyixuxu (approved):
thanks

PR #13983, opened 2026-06-18
fix(bria_fibo): honor timesteps, fix edit num_images batching, add _no_split_modules, fix docstring

akshan-main:
# What does this PR do?

Part of #13618 (`bria_fibo` review), follow-up to #13981. Rebased on main, which has since fixed three of the original findings on its own (the reference-latent rewrite expands to the effective batch, `timesteps` is now honored instead of ignored, and the edit example docstring was rewritten), so this PR is down to the two that remain:

- The pipelines hardcoded `vae_scale_factor = 16` instead of reading the VAE's `scale_factor_spatial`; now read from config.
- `BriaFiboTransformer2DModel` was missing `_no_split_modules`, so `device_map`/offload could split residual transformer blocks. Declared the two block classes the model uses (this also enables 4 previously-skipped sharding tests, which pass).

## Before submitting
- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? Discussed on Slack with @yiyixuxu

## Who can review?

@yiyixuxu @sayakpaul


src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo_edit.py
```diff
@@ -1044,6 +1038,9 @@ def prepare_image_latents(
         image_latents_cthw = torch.concat(latents_scaled, dim=0)
         image_latents_bchw = image_latents_cthw[:, :, 0, :, :]
 
+        repeat_by = batch_size // image_latents_bchw.shape[0]
```
yiyixuxu: we talked about standardlize this across pipelines using a function like this https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/ideogram4/pipeline_ideogram4.py#L115

maybe we can start addingthat while you're doing these pipeline refactors?

src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo_edit.py
```diff
@@ -898,7 +892,7 @@ def __call__(
             self.scheduler,
             num_inference_steps=num_inference_steps,
             device=device,
-            timesteps=None,
+            timesteps=timesteps,
```
yiyixuxu: we can just deprecate `timesteps` instead (it has not been used so far)

yiyixuxu: see example https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py#L896

yiyixuxu (commented):
thanks!

src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo_edit.py
akshan-main: sure!

src/diffusers/pipelines/bria_fibo/pipeline_bria_fibo_edit.py
akshan-main: done!

akshan-main:
friendly ping @yiyixuxu

PR #13992, opened 2026-06-19
Fix duplicate timestep shift in FlowMatchEulerDiscreteScheduler

### LINKED ISSUE #13243: [BUG] FlowMatchEulerDiscreteScheduler.__init__ computes sigma_min/sigma_max after shift, causing duplicate shift in set_timesteps

Whning0513:
Fixes #13243.

### DIFF
--- a/b/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
+++ b/b/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
@@ -1,0 +1,20 @@
+        self._shift = shift
+
+        self.sigmas = sigmas.to("cpu")  # to avoid too much CPU/GPU communication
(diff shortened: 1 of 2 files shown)


Whning0513:
Thanks for the context on #13243. Several earlier PRs already address this scheduler behavior, including #13246, #13706, and #13962, and the maintainers are still deciding the appropriate compatibility scope. This branch also contains an incorrectly added path, so I am closing it rather than adding another duplicate to the review queue. Sorry for the noise.

PR #13326, opened 2026-03-25
Add Flux2 LoKR adapter support prototype with dual conversion paths

CalamitousFelicitousness:
Adds support for Flux2 LoKR, with dual path to benchmark implementations.
- Custom lossless path: BFL LoKR keys → peft LoKrConfig (fuse-first QKV)
- Generic lossy path: optional SVD conversion via peft.convert_to_lora
- Fix alpha handling for lora_down/lora_up format checkpoints
- Re-fuse LoRA keys when model QKV is fused from prior LoKR load

Given that Civitai does not have a LoKR category I didn't feel like digging for a SFW one, so I just used what user brought up to me when they reported the issue.

<details>
<summary>Benchmark test</summary>

```
"""Benchmark: Lossless LoKR vs Lossy LoRA-via-SVD on Flux2 Klein 9B.

Generates images using both conversion paths for visual comparison.
Uses 4-bit quantization + CPU offload to fit on a single 24GB GPU.

Usage:
    python benchmark_lokr.py
    python benchmark_lokr.py --prompt "a cat in a garden" --ranks 32 64 128
"""

import argparse
import gc
import os
import time

import torch

# Use local model cache
os.environ["HF_HUB_CACHE"] = "/home/ohiom/database/models/huggingface"

from diffusers import Flux2KleinPipeline  # noqa: E402
from peft import convert_to_lora  # noqa: E402

MODEL_ID = "black-forest-labs/FLUX.2-klein-9B"
LOKR_PATH = "/home/ohiom/database/models/Lora/Flux.2 Klein 9B/klein_snofs_v1_1.safetensors"
OUTPUT_DIR = "/home/ohiom/diffusers/benchmark_output"


def load_pipeline():
    """Load Flux2 Klein 9B in bf16 with model CPU offload."""
    pipe = Flux2KleinPipeline.from_pretrained(MODEL_ID, torch_dtype=torch.bfloat16)
    pipe.enable_model_cpu_offload()
    return pipe


def generate(pipe, prompt, seed, num_steps=4, guidance_scale=1.0):
    """Generate a single image with fixed seed for reproducibility."""
    generator = torch.Generator(device="cpu").manual_seed(seed)
    image = pipe(
        prompt=prompt,
        num_inference_steps=num_steps,
        guidance_scale=guidance_scale,
        generator=generator,
        height=1024,
        width=1024,
    ).images[0]
    return image


def benchmark_lossless(pipe, prompt, seed):
    """Path A: Load LoKR natively (lossless)."""
[...]

Issue #13261, opened 2026-03-12
black-forest-labs/FLUX.2-klein-9B not working with lora with lokr


chaowenguo:
### Describe the bug

Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/aiohttp/web_protocol.py", line 510, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiohttp/web_app.py", line 569, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/studio_service/PROJECT/app.py", line 30, in imageGet
    request.app.pop('future').result().save(buffer, 'png')
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/studio_service/PROJECT/app.py", line 16, in imageGenerate
    flux.load_lora_weights(modelscope.model_file_download('chaowenguo/lora', 'klein_snofs_v1_1.safetensors'), adapter_name='snofs')
  File "/usr/local/lib/python3.11/site-packages/diffusers/loaders/lora_pipeline.py", line 5713, in load_lora_weights
    state_dict, metadata = self.lora_state_dict(pretrained_model_name_or_path_or_dict, **kwargs)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/diffusers/loaders/lora_pipeline.py", line 5682, in lora_state_dict
    state_dict = _convert_non_diffusers_flux2_lora_to_diffusers(state_dict)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #13997, opened 2026-06-19
Fix LoKr format LoRA weight loading for Flux2 pipelines

Whning0513:
LoKr weights were failing to load with a `ValueError` about non-empty `original_state_dict`. Adjusted the weight processing to handle the LoKr format correctly.

### DIFF
--- a/src/diffusers/loaders/lora_conversion_utils.py
+++ b/src/diffusers/loaders/lora_conversion_utils.py
@@ -2377,6 +2377,18 @@ def _convert_non_diffusers_flux2_lora_to_diffusers(state_dict):
             temp_state_dict[new_key] = v
         original_state_dict = temp_state_dict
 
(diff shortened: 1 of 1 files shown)


asomoza:
Hi @Whning0513 this is not the correct way to handle lokr, with this fix the lora will load but practically won't have any effect. 

Please look at this [issue](https://github.com/huggingface/diffusers/issues/13261) and this [PR](https://github.com/huggingface/diffusers/pull/13326) if you want to contribute

Whning0513:
You're right, I looked at the code again and this approach doesn't properly handle the LoKr factorization — it just suppresses the error without actually applying the weights correctly.

I see PR #13326 already has a proper implementation for this. I'll close this one.

Thanks for taking the time to review and point me to the right direction.

Whning0513:
Thanks for the guidance. I closed #13997 and will focus on the other fixes. Let me know if any of the remaining PRs need changes.

PR #14001, opened 2026-06-20
[Fix] Fix three final_layer LoRA conversion bugs in _convert_sd_scripts_to_ai_toolkit

### LINKED ISSUE #13998: FLUX kohya LoRA conversion crashes on `final_layer` (KeyError on missing adaLN_modulation_1; "Incompatible keys" on final_layer alphas)
### Describe the bug

The kohya / sd-scripts FLUX LoRA converter (`_convert_sd_scripts_to_ai_toolkit` in `lora_conversion_utils.py`) crashes on LoRAs that include `final_layer` weights, in two distinct ways. Both reproduce on `diffusers==0.38.0` and on current `main`.

**Case A — `KeyError` when `final_layer.linear` has no `adaLN_modulation_1` companion.**
`assign_remaining_weights` pops the `adaLN_modulation_1` source key *unconditionally*, so a LoRA that trained `final_layer.linear` but not `final_layer.adaLN_modulation.1` (a real, if uncommon, kohya export) raises `KeyError`.

**Case B — `Incompatible keys detected` when `final_layer` carries `.alpha` entries.**
`assign_remaining_weights` consumes only `lora_down`/`lora_up`, never the `.alpha` keys. The leftover `lora_unet_final_layer_*.alpha` keys then reach the `remaining_keys` guard, which only tolerates `lora_te*` prefixes, so the (otherwise valid) LoRA is rejected.

### Reproduction

```python
import torch
from diffusers.loaders.lora_pipeline import FluxLoraLoaderMixin

RANK, HID = 4, 3072

def kohya(name, out, inp, alpha=True):
    d = {f"{name}.lora_down.weight": torch.zeros(RANK, inp),
         f"{name}.lora_up.weight":   torch.zeros(out, RANK)}
    if alpha:
        d[f"{name}.alpha"] = torch.tensor(float(RANK))
    return d

def base():  # one recognized block so detection routes to the kohya converter
    return kohya("lora_unet_double_blocks_0_img_attn_proj", HID, HID)

# Case A: final_layer.linear without adaLN_modulation_1  ->  KeyError
sd = base() | kohya("lora_unet_final_layer_linear", 64, HID)
FluxLoraLoaderMixin.lora_state_dict(sd)
# KeyError: 'lora_unet_final_layer_adaLN_modulation_1.lora_down.weight'

# Case B: final_layer linear + adaLN present, alphas included  ->  Incompatible keys
sd = base() | kohya("lora_unet_final_layer_linear", 64, HID) \
            | kohya("lora_unet_final_layer_adaLN_modulation_1", 2 * HID, HID)
FluxLoraLoaderMixin.lora_state_dict(sd)
# ValueError: Incompatible keys detected:
#   lora_unet_final_layer_linear.alpha, lora_unet_final_layer_adaLN_modulation_1.alpha
```

### Root cause

In [`_convert_sd_scripts_to_ai_toolkit`](https://github.com/huggingface/diffusers/blob/7bf00006aa005eae37bcc639fd0f010c183365b4/src/diffusers/loaders/lora_conversion_utils.py#L618-L630):

- [`assign_remaining_weights`](https://github.com/huggingface/diffusers/blob/7bf00006aa005eae37bcc639fd0f010c183365b4/src/diffusers/loaders/lora_conversion_utils.py#L548-L557) does `value = source.pop(source_key)` ([L554](https://github.com/huggingface/diffusers/blob/7bf00006aa005eae37bcc639fd0f010c183365b4/src/diffusers/loaders/lora_conversion_utils.py#L554)) with no fallback — **Case A**. It also only handles `lora_down`/`lora_up`, so `final_layer` `.alpha` keys are never consumed.
- The unconsumed alphas then hit the [`remaining_keys` guard](https://github.com/huggingface/diffusers/blob/7bf00006aa005eae37bcc639fd0f010c183365b4/src/diffusers/loaders/lora_conversion_utils.py#L632-L636), which raises unless every leftover key starts with `lora_te`/`lora_te1` — **Case B**.

### Suggested fix

- In `assign_remaining_weights`, skip an assignment whose `source_key` is absent (`source.pop(source_key, None)` + continue on `None`) so a lone `final_layer.linear` still converts.
- Consume `final_layer` `.alpha` keys alongside `lora_down`/`lora_up` (or strip them before the `remaining_keys` guard), as the per-block `_convert_to_ai_toolkit` path already does.

### System Info

- diffusers 0.38.0 (and `main` @ 7bf00006)
- transformers 5.9.0, peft 0.19.1, torch 2.11+cu128, Python 3.11, Linux
- Model: black-forest-labs/FLUX.1 family (kohya/sd-scripts LoRAs with `final_layer` weights)

### Who can help?

@sayakpaul @BenjaminBossan

lcheng321:
Fix three final_layer LoRA conversion bugs in _convert_sd_scripts_to_ai_toolkit

Fixes https://github.com/huggingface/diffusers/issues/13998

Three independent bugs in assign_remaining_weights caused failures when loading FLUX LoRAs that include final_layer weights (common in kohya and ComfyUI/SwarmUI exports).

Case A (KeyError): source.pop(source_key) raised KeyError when a LoRA trained final_layer.linear but not final_layer.adaLN_modulation_1. Fixed by using source.pop(source_key, None) and skipping on None.

Case B (Incompatible keys): Unconsumed final_layer .alpha keys were left in the state dict after assign_remaining_weights, which only consumed lora_down/lora_up. These keys then hit the remaining_keys guard that only tolerates lora_te* prefixes, raising ValueError. Fixed by consuming final_layer .alpha keys after the assignment loop.

Case C (Silent weight corruption): swap_scale_shift was applied to both lora_A and lora_B. The correct transform is swap(B) @ A = swap(B) @ A, meaning only lora_B should be swapped. Applying it to lora_A as well produces swap(B) @ swap(A) which silently corrupts the norm_out LoRA for any FLUX LoRA that trains the final-layer adaLN modulation. Fixed by gating transform on lora_key == "lora_B".

### DIFF
--- a/src/diffusers/loaders/lora_conversion_utils.py
+++ b/src/diffusers/loaders/lora_conversion_utils.py
@@ -551,10 +551,18 @@ def _convert_kohya_flux_lora_to_diffusers(state_dict):
                 for target_fmt, source_fmt, transform in assignments:
                     target_key = target_fmt.format(lora_key=lora_key)
                     source_key = source_fmt.format(orig_lora_key=orig_lora_key)
-                    value = source.pop(source_key)
-                    if transform:
+                    value = source.pop(source_key, None)
+                    if value is None:
+                        continue
+                    if transform and lora_key == "lora_B":
                         value = transform(value)
                     ait_sd[target_key] = value
+
+            # Consume any leftover final_layer alpha keys so they don't
+            # reach the remaining_keys guard and cause a false "Incompatible keys" error.
+            for key in list(source.keys()):
+                if "final_layer" in key and key.endswith(".alpha"):
+                    source.pop(key)
 
         if any("guidance_in" in k for k in sds_sd):
             _convert_to_ai_toolkit(


sayakpaul (approved):
Thanks a lot!

PR #14017, opened 2026-06-21
Fix pipeline callback state initialized in __init__ (#11002)

### LINKED ISSUE #11002: Any chance class members like self._interrupt could be defined in __init__ across pipelines?

srlynch1:
Closes #11002. Pack real-issues e2e 2026-06-21.

### DIFF
--- a/src/diffusers/pipelines/bria/pipeline_bria.py
+++ b/src/diffusers/pipelines/bria/pipeline_bria.py
@@ -123,6 +123,7 @@ class BriaPipeline(DiffusionPipeline):
         image_encoder: CLIPVisionModelWithProjection = None,
         feature_extractor: CLIPImageProcessor = None,
     ):
+        super().__init__()
         self.register_modules(
             vae=vae,
             text_encoder=text_encoder,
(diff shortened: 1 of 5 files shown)


srlynch1:

PR #14019, opened 2026-06-21
Fix StableDiffusionPipeline legacy init incompatibility (#6969)

### LINKED ISSUE #6969: Expanded init fields in StableDiffusionPipeline cause incompatibilities with many/most inherited pipelines
### Describe the bug

class `StableDiffusionPipeline` has its init section as:
```py
    def __init__(
        self,
        vae: AutoencoderKL,
        text_encoder: CLIPTextModel,
        tokenizer: CLIPTokenizer,
        unet: UNet2DConditionModel,
        scheduler: KarrasDiffusionSchedulers,
        safety_checker: StableDiffusionSafetyChecker,
        feature_extractor: CLIPImageProcessor,
        image_encoder: CLIPVisionModelWithProjection = None,
        requires_safety_checker: bool = True,
    ):
```

and here `image_encoder` was recently introduced thus changing the class signature  
but most community pipelines do not include init field for recently included `image_encoder` and thus order or params is wrong  

for example, `examples/community/regional_prompting_stable_diffusion.py` has this in its init:  
```py
        super().__init__(
            vae,
            text_encoder,
            tokenizer,
            unet,
            scheduler,
            safety_checker,
            feature_extractor,
            requires_safety_checker,
        )
```
which means `bool` value from `requires_safety_checker` is going to be passed as `image_encoder` and pipeline will fail during initialization like this:  

```log
> diffusers/pipelines/pipeline_utils.py:546 in _fetch_class_library_tuple
object has no attribute '__module__
```

this is a conceptual problem with changing master class signature while all inherited classes pass args list as simple list.  
i don't see a simple solution as going back is bad and going forward requires updates to a lot of pipelines.  

### Reproduction

load nearly any community pipeline or any pipeline inherited from StableDiffusionPipeline, but not updated to use new signature in its `super().__init__` call

### Logs

_No response_

### System Info

diffusers==0.26.3

### Who can help?

@yiyixuxu @DN6 @sayakpaul @patrickvonplaten

srlynch1:
Closes #6969. Pack real-issues e2e 2026-06-21.

### DIFF
--- a/src/diffusers/pipelines/pipeline_loading_utils.py
+++ b/src/diffusers/pipelines/pipeline_loading_utils.py
@@ -958,6 +958,11 @@ def _fetch_class_library_tuple(module):
 
     # register the config from the original module, not the dynamo compiled one
     not_compiled_module = _unwrap_model(module)
+    if isinstance(not_compiled_module, (bool, int, float, str)) or not hasattr(not_compiled_module, "__module__"):
+        raise TypeError(
+            f"Cannot fetch library tuple for object of type {type(not_compiled_module).__name__}. "
+            "Expected a model, scheduler, tokenizer, or pipeline module."
+        )
     library = not_compiled_module.__module__.split(".")[0]
 
     # check if the module is a pipeline module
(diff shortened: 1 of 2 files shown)


srlynch1:

PR #9475, opened 2024-09-19
Profile and speedup pipelines

a-r-r-o-w:
# What does this PR do?

Currently tested only for CogVideoX. Noticing a ~42% speedup comparing compiled versions of this branch VS compiled version of `main`. Will share more concrete numbers for our major pipelines/schedulers soon.

<details>
<summary> testing code </summary>

```python
import argparse
import logging
import os
from platform import python_version

os.environ['TORCH_LOGS'] = '+graph_breaks,dynamo,recompiles'
os.environ['TORCHDYNAMO_VERBOSE'] = '1'

logging.basicConfig(level=logging.INFO)

import diffusers
import torch
from diffusers import CogVideoXPipeline
from diffusers.utils import export_to_video

from torch.profiler import profile, record_function, ProfilerActivity, tensorboard_trace_handler


print(diffusers.__version__)
print(torch.__version__)
print(python_version())


torch.set_float32_matmul_precision("high")
torch._inductor.conv_1x1_as_mm = True
torch._inductor.coordinate_descent_tuning = True
torch._inductor.epilogue_fusion = False
torch._inductor.coordinate_descent_check_all_directions = True

PROMPT_EMBEDS_FILE = "prompt_embeds.pt"
NEGATIVE_PROMPT_EMBEDS_FILE = "negative_prompt_embeds.pt"


def profiler_runner(fn, *args, **kwargs):
    with torch.profiler.profile(
            activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA],
            record_shapes=True,
            on_trace_ready=tensorboard_trace_handler("./cogvideox_trace")
    ) as prof:
        result = fn(*args, **kwargs)
    
    # prof.export_chrome_trace("cogvideox_trace.json")
    return result


@torch.no_grad()
def generate_embeds(pipe):
    prompt = (
        "A panda, dressed in a small, red jacket and a tiny hat, sits on a wooden stool in a serene bamboo forest. "
        "The panda's fluffy paws strum a miniature acoustic guitar, producing soft, melodic tunes. Nearby, a few other "
        "pandas gather, watching curiously and some clapping in rhythm. Sunlight filters through the tall bamboo, "
        "casting a gentle glow on the scene. The panda's face is expressive, showing concentration and joy as it plays. "
        "The background includes a small, flowing stream and vibrant green foliage, enhancing the peaceful and magical "
        "atmosphere of this unique musical performance."
    )
    prompt_embeds, negative_prompt_embeds = pipe.encode_prompt(
[...]

PR #14021, opened 2026-06-21
Speed up index_for_timestep in FlowMatchEulerDiscreteScheduler (#9417)

### LINKED ISSUE #9417: Suggestion for speeding up `index_for_timestep` by removing sequential `nonzero()` calls in samplers
First off, thanks for the great codebase and providing so many resources! I just wanted to provide some insight into an improvement I made for myself, in case you'd like to include it for all samplers. I'm using the `FlowMatchEulerDiscreteScheduler` and after profiling, I've noticed that it's unexpectedly slowing down my training speeds. I'll describe the issue and proposed solution here rather than making a PR, since this would touch a lot of code and perhaps someone on the diffusers team would like to implement it.

This line in particular is very slow because it is a for loop `step_indices = [self.index_for_timestep(t, schedule_timesteps) for t in timestep]` and the `self.index_for_timestep()` is calling a nonzero() function which is slow.

https://github.com/huggingface/diffusers/blob/b9e2f886cd6e9182f1bf1bf7421c6363956f94c5/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L149

I've changed the code as follows:

```python
# huggingface code
def index_for_timestep(self, timestep, schedule_timesteps=None):
    if schedule_timesteps is None:
        schedule_timesteps = self.timesteps

    indices = (schedule_timesteps == timestep).nonzero()

    # The sigma index that is taken for the **very** first `step`
    # is always the second index (or the last index if there is only 1)
    # This way we can ensure we don't accidentally skip a sigma in
    # case we start in the middle of the denoising schedule (e.g. for image-to-image)
    pos = 1 if len(indices) > 1 else 0

    return indices[pos].item()
```

changed to =>

```python
# my code
def index_for_timestep(self, timestep, schedule_timesteps=None):
    if schedule_timesteps is None:
        schedule_timesteps = self.timesteps

    num_steps = len(schedule_timesteps)
    start = schedule_timesteps[0].item()
    end = schedule_timesteps[-1].item()
    indices = torch.round(((timestep - start) / (end - start)) * (num_steps - 1)).long()

    return indices
```

and

```python
# huggingface code
# self.begin_index is None when scheduler is used for training, or pipeline does not implement set_begin_index
if self.begin_index is None:
    step_indices = [self.index_for_timestep(t, schedule_timesteps) for t in timestep]
```

changed to =>

```python
# my code
# self.begin_index is None when scheduler is used for training, or pipeline does not implement set_begin_index
if self.begin_index is None:
    step_indices = self.index_for_timestep(timestep, schedule_timesteps)
```

**Additional context.**
Just wanted to bring this modification to your attention since it could be a training speedup for folks. 🙂 Especially when someone has a large batch size > 1 and this for loop it occurring with nonzero search operations. Some other small changes might be necessary to ensure compatibility of the function changes, but I suspect it could help everyone. Thanks for the consideration!

srlynch1:
Closes #9417. Related to #9475. Pack real-issues e2e 2026-06-21.

### DIFF
--- a/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
+++ b/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py
@@ -219,7 +219,9 @@ class FlowMatchEulerDiscreteScheduler(SchedulerMixin, ConfigMixin):
 
         # self.begin_index is None when scheduler is used for training, or pipeline does not implement set_begin_index
         if self.begin_index is None:
-            step_indices = [self.index_for_timestep(t, schedule_timesteps) for t in timestep]
+            step_indices = self.index_for_timestep(timestep, schedule_timesteps)
+            if not torch.is_tensor(step_indices):
+                step_indices = torch.tensor([step_indices], device=sample.device, dtype=torch.long)
         elif self.step_index is not None:
             # add_noise is called after first denoising step (for inpainting)
             step_indices = [self.step_index] * timestep.shape[0]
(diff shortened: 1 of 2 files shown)


srlynch1:

PR #14026, opened 2026-06-21
[Fix] Resolve deprecated pipeline modules in from_pretrained

### LINKED ISSUE #14008: from_pretrained fails loading deprecated pipelines
Note: I understand these pipelines are deprecated and bugfixes aren't accepted anymore. But this seems to be a bug in _the process of moving pipelines to deprecated itself_

### Describe the bug

`_get_custom_components_and_folders` (`src/diffusers/pipelines/pipeline_loading_utils.py`) resolves custom pipeline components by checking `hasattr(diffusers.pipelines, module_candidate)`. Since #13157 moved several pipelines (e.g. Wuerstchen) to `diffusers.pipelines.deprecated.<name>`, that check no longer finds them, even though the relocated module still exists.

### Reproduction

```python
DiffusionPipeline.from_pretrained("warp-ai/wuerstchen-prior")
```
```
ValueError: prior/wuerstchen.py as defined in `model_index.json` does not exist in warp-ai/wuerstchen-prior and is not a module in 'diffusers/pipelines'.
```

`model_index.json` declares `"prior": ["wuerstchen", "WuerstchenPrior"]`. `hasattr(diffusers.pipelines, "wuerstchen")` is `False`, but `hasattr(diffusers.pipelines.deprecated, "wuerstchen")` is `True` — confirming the module just moved out from under the check.

**Suggested fix:** have `_get_custom_components_and_folders` also check `diffusers.pipelines.deprecated` before raising.

### System Info

- diffusers commit: 0f1abc4ae8b0eb2a3b40e82a310507281144c423 (main)
- Python: 3.12
- Platform: Linux

### Who can help?
@yiyixuxu @stevhliu 

---
Drafted by Claude

Nas01010101:
# What does this PR do?

Fixes #14008

## The bug

`DiffusionPipeline.from_pretrained` raises a spurious `ValueError` when loading any pipeline that was relocated under `diffusers.pipelines.deprecated` (for example Wuerstchen, moved in #13157):

```python
DiffusionPipeline.from_pretrained("warp-ai/wuerstchen-prior")
```

```
ValueError: prior/wuerstchen.py as defined in `model_index.json` does not exist in
warp-ai/wuerstchen-prior and is not a module in 'diffusers/pipelines'.
```

## Root cause

This is a bug in the deprecation mechanism itself, not in the deprecated pipeline.

`_get_custom_components_and_folders` decides whether a component refers to a built-in module via `hasattr(diffusers.pipelines, module_candidate)`. When a pipeline is moved to `diffusers.pipelines.deprecated`, its module name is no longer a top-level attribute of `diffusers.pipelines` (deprecated modules are nested under the `deprecated` key in `_import_structure`). So `hasattr` returns `False`, the component is mistaken for a missing custom module, and loading fails.

The fix suggested in the issue (`hasattr(diffusers.pipelines.deprecated, module_candidate)`) does not work: `diffusers.pipelines.deprecated` is a `_LazyModule`, so the module name (`wuerstchen`) is not an attribute until it is imported; only the lazily exposed class names are. `hasattr` therefore still returns `False`.

## Fix

Resolve the module spec under the deprecated namespace without importing it, via `importlib.util.find_spec("diffusers.pipelines.deprecated.<module>")`. This is import side-effect free (no heavy import is triggered) and correctly recognizes relocated modules. Genuinely missing modules still raise, since `find_spec` returns `None` for them.

## Tests

Added `GetCustomComponentsAndFoldersTests` in `tests/pipelines/test_pipeline_utils.py` (the existing home for `pipeline_loading_utils` unit tests). All are function level and require no network or GPU.

```
$ python -m pytest tests/pipelines/test_pipeline_utils.py::GetCustomComponentsAndFoldersTests -q
4 passed, 2 warnings in 2.28s
```

Reverting only the production change reproduces the issue (`from_pretrained` raises the `ValueError` above) and makes the new tests fail, confirming they guard the bug. `ruff check` and `ruff format --check` pass on the changed files.

## Before submitting

- [x] Did you use an AI agent to help with this PR? Yes, Claude Opus 4.8. The diff was self-reviewed against `.ai/review-rules.md`, and I understand and take responsibility for the change.

### DIFF
--- a/src/diffusers/pipelines/pipeline_loading_utils.py
+++ b/src/diffusers/pipelines/pipeline_loading_utils.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import importlib
+import importlib.util
 import os
 import re
 import warnings
@@ -1073,6 +1074,22 @@ def _update_init_kwargs_with_connected_pipeline(
     return init_kwargs
 
 
+def _is_deprecated_pipeline_module(module_candidate: str) -> bool:
+    """Return whether ``module_candidate`` is a pipeline module relocated under
+    ``diffusers.pipelines.deprecated``.
+
+    Deprecated pipelines (e.g. Wuerstchen) are no longer attributes of ``diffusers.pipelines``, so a
+    plain ``hasattr(diffusers.pipelines, module_candidate)`` check fails for them even though the
+    module still ships with diffusers. We resolve the spec without importing the module to avoid
+    triggering its (potentially heavy) import side effects.
+    """
+    try:
+        return importlib.util.find_spec(f"diffusers.pipelines.deprecated.{module_candidate}") is not None
+    except (ImportError, ModuleNotFoundError, ValueError):
+        # ValueError covers malformed candidate names (e.g. containing path separators).
+        return False
+
+
 def _get_custom_components_and_folders(
     pretrained_model_name: str,
     config_dict: dict[str, Any],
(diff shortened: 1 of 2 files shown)


dxqb:
Claude: Thanks for this — the `_is_deprecated_pipeline_module` check does get past the early `ValueError` in `_get_custom_components_and_folders` when loading e.g. `warp-ai/wuerstchen-prior`.

One thing to flag: with this PR applied, loading a deprecated pipeline still doesn't complete. `get_class_obj_and_candidates` later does `importlib.import_module(library_name)` with the bare module name, so it then raises:

```
ModuleNotFoundError: No module named 'wuerstchen'
```

(`pipeline_loading_utils.py`, in the `else: # else we just import it from the library.` branch.) The module resolves only under `diffusers.pipelines.deprecated.<library_name>`, so that import path may need the same deprecated-namespace fallback for the fix to make these pipelines actually loadable end-to-end.

Drafted by Claude.

myselfhimself:
For me the code changes by the 2 contributors of this PR fix the problem as tested on my side yesterday. I have not tried to run the related patched automated test collections.
Thank you for the fix!

PR #14027, opened 2026-06-21
fix(mag_cache): correct per-transformer step count for dual-transformer pipelines

### LINKED ISSUE #14025: MagCache on Wan 2.2 Dual-Transformer Pipelines: Incorrect Step Accounting and Limited Effectiveness on a 4-Step Distilled Model (e.g. Wan2.2)
</h2><p>While evaluating MagCache on Wan 2.2, I found what appears to be a design issue in the current implementation for dual-transformer pipelines.</p><p>The current implementation assumes that a transformer executes exactly <code inline="">num_inference_steps</code> times during inference. This assumption is valid for single-transformer architectures such as Flux, SDXL, and Wan 2.1, but it does not hold for Wan 2.2, where denoising is split between two transformers that each execute only during part of the diffusion process.</p><p>After implementing a prototype fix that assigns independent MagCache configurations to each transformer based on the number of timesteps it actually executes, I still observe poor practical results on a distilled 4-step Wan 2.2 model:</p><ul><li><p>At thresholds below 0.3, MagCache provides little or no measurable speedup.</p></li><li><p>At thresholds greater than or equal to 0.3, runtime improves (~25%), but output quality degrades severely.</p></li><li><p>The same behavior occurs whether MagCache is applied only to the first transformer or to both transformers.</p></li></ul><p>I am therefore opening this issue both as a bug report regarding the current dual-transformer implementation and as a request for feedback regarding the expected behavior of MagCache on heavily step-distilled Wan 2.2 models.</p><hr><h2>Background</h2><p>According to the current MagCache documentation, calibration produces a set of <code inline="">mag_ratios</code> that describe the model's intrinsic magnitude decay curve.</p><p>Example workflow:</p><pre><code class="language-python">calib_config = MagCacheConfig(
    calibrate=True,
    num_inference_steps=4,
)
</code></pre><p>After calibration, the reported ratios are passed back into the inference configuration:</p><pre><code class="language-python">mag_config = MagCacheConfig(
    mag_ratios=[...],
    num_inference_steps=4,
)
</code></pre><p>The documentation also states that ratios calibrated at higher step counts can be reused at lower step counts through interpolation.</p><p>For example, ratios obtained from a 50-step calibration can be interpolated and reused for a 20-step or lower-step inference schedule.</p><hr><h2>Experimental Setup</h2><p>The tested model is not the original Wan 2.2 inference configuration.</p><p>The workflow is:</p><ol><li><p>Start from the original Wan 2.2 14B model (designed for approximately 50 inference steps).</p></li><li><p>Apply a step-distillation LoRA to convert the model into a 4-step model.</p></li><li><p>Evaluate MagCache on the resulting distilled model.</p></li></ol><p>Therefore all observations below correspond to:</p><pre><code class="language-text">Wan 2.2 14B (50-step)
    ↓
Step Distillation LoRA
    ↓
4-step distilled model
    ↓
MagCache evaluation
</code></pre><p>I also tested using interpolated MagCache ratios derived from the original 50-step calibration, following the workflow described in the documentation.</p><p>The resulting outputs were significantly degraded on the distilled 4-step model.</p><hr><h2>Potential Bug: Dual-Transformer Step Accounting</h2><p>Wan 2.2 divides denoising between two transformers:</p><ul><li><p><code inline="">transformer</code> (high-noise stage)</p></li><li><p><code inline="">transformer_2</code> (low-noise stage)</p></li></ul><p>The split is determined by <code inline="">boundary_ratio</code>.</p><p>For example, with a 4-step schedule:</p>
Stage | Executions
-- | --
High-noise transformer | 2
Low-noise transformer | 2

<p>Neither transformer executes 4 times.</p><p>However, MagCache currently receives:</p><pre><code class="language-python">num_inference_steps = 4
</code></pre><p>for both transformers.</p><p>Internally, calibration completion depends on:</p><pre><code class="language-python">if state.step_index &gt;= self.config.num_inference_steps:
</code></pre><p>This implicitly assumes:</p><pre><code class="language-text">transformer executions == num_inference_steps
</code></pre><p>which is not true for Wan 2.2.</p><p>As a result, calibration and state management appear to be tied to the total diffusion step count rather than the number of executions of the individual transformer.</p><hr><h2>Prototype Fix</h2><p>To address this, I implemented a prototype that treats each transformer as an independent MagCache scope.</p><p>The scheduler timesteps are first split according to the pipeline's <code inline="">boundary_ratio</code>.</p><p>Example:</p><pre><code class="language-text">4 diffusion steps

High-noise transformer: 2 steps
Low-noise transformer: 2 steps
</code></pre><p>Each transformer then receives:</p><ul><li><p>its own <code inline="">MagCacheConfig</code></p></li><li><p>its own <code inline="">num_inference_steps</code></p></li><li><p>its own <code inline="">mag_ratios</code></p></li><li><p>its own MagCache state</p></li></ul><p>instead of sharing a single global diffusion-step count.</p><p>Conceptually:</p><pre><code class="language-text">transformer
 ├── own state
 ├── own ratios
 └── own step count

transformer_2
 ├── own state
 ├── own ratios
 └── own step count
</code></pre><p>This guarantees that:</p><pre><code class="language-python">if state.step_index &gt;= self.config.num_inference_steps:
</code></pre><p>is evaluated against the number of executions of that specific transformer rather than the total diffusion schedule.</p><hr><h2>Results</h2><p>After implementing the per-transformer step accounting described above, I evaluated MagCache on the distilled 4-step Wan 2.2 model.</p><h3>Threshold &lt; 0.3</h3><p>No meaningful speedup was observed.</p><p>Runtime remained very close to the baseline.</p><h3>Threshold &gt;= 0.3</h3><p>Runtime improved from approximately:</p><pre><code class="language-text">40 s → 30 s
</code></pre><p>which corresponds to roughly a 25% speedup.</p><p>However, output quality degraded significantly and became visibly worse than the baseline.</p><h3>Transformer Coverage</h3>
<p>I tested both:</p>
<ul>
<li><p>applying MagCache only to <code inline="">transformer</code></p></li>
<li><p>applying MagCache to both <code inline="">transformer</code> and <code inline="">transformer_2</code></p></li>
</ul>

<p>The behavior was similar in terms of speed/quality tradeoff, but there was one noticeable difference:</p>

<ul>
<li><p>Applying MagCache only to the first transformer (<code inline="">transformer</code>) produced better output quality compared to applying MagCache to both transformers, but provided no meaningful speedup and had the same inference time as the base model with LoRA.</p></li>
<li><p>However, the quality was still significantly worse than the baseline without MagCache.</p></li>
</ul>

<p>This suggests that the second transformer may be more sensitive to MagCache reuse, or that the error introduced by caching accumulates differently across the two denoising stages.</p>

baseline wan22 14b 4step distilled:
https://github.com/user-attachments/assets/01424605-874a-4635-b07e-4817111227c7

baseline wan22 14b 4step distilled + apply magcache on both transformers:
https://github.com/user-attachments/assets/22e52d25-a524-496d-9756-18df0942adcd

baseline wan22 14b 4step distilled + apply magcache on first transformer:
https://github.com/user-attachments/assets/4e99b509-b17f-4c1e-8463-90b273fa3420

lcheng321:
Fixes #14025

## Root cause

`MagCacheConfig` was applied to both `transformer` and `transformer_2` with `num_inference_steps` set to the full pipeline step count. In dual transformer pipelines such as Wan 2.2, `transformer` only runs the high noise stage and `transformer_2` only runs the low noise stage. Each of them executes a subset of the total steps, never the full count.

This causes three downstream problems:

- `_advance_step`: the check `state.step_index >= config.num_inference_steps` never fires for either transformer, since neither ever reaches the full step count. State never resets and calibration never completes.
- Retention window: `int(retention_ratio * num_inference_steps)` is computed against the full step count instead of the actual per transformer step count, so it protects more steps from skipping than intended.
- `mag_ratios` indexing: since `mag_ratios` is sized to the full step count, `step_index` no longer maps to the correct entry, so each transformer reads ratios that do not correspond to its own denoising timestep.

## First attempt and why it changed

The initial fix added `update_mag_cache_num_steps(module, actual_steps)`, which mutated `config.num_inference_steps` and `config.mag_ratios` in place on the attached `MagCacheConfig`.

Review feedback identified two problems with that approach:

1. If the same `MagCacheConfig` object is passed to both `transformer` and `transformer_2`, calling this function for one transformer corrupts the config for the other, since they share the same object.
2. The pipeline had to `import` a hook internal function directly, coupling pipeline code to hook implementation details.

## Fix

`MagCacheConfig` is now never mutated after construction. Instead:

- Pipelines set a plain attribute on each transformer, `transformer._mag_cache_expected_steps = n_steps`, no hook internals imported.
- The hooks hold a reference to the root transformer module (passed in at `apply_mag_cache` time) and read `_mag_cache_expected_steps` off it live, at forward and reset time, falling back to `config.num_inference_steps` if the attribute was never set.
- `mag_ratios` are computed on demand from `_original_mag_ratios` at the effective step count, never written back to `config.mag_ratios`.

This means a shared config object across both transformers can no longer be corrupted, since nothing ever writes to it after construction.

Files changed:

| File | Change |
| --- | --- |
| `src/diffusers/hooks/mag_cache.py` | Add `_get_effective_num_steps` and `_get_effective_mag_ratios`, both pure functions that read from a root module attribute and a config, no mutation. Hooks now take a `root_module` reference. Remove `update_mag_cache_num_steps`. |
| `src/diffusers/hooks/__init__.py` | Remove the export of the now deleted `update_mag_cache_num_steps`. |
| `src/diffusers/pipelines/wan/pipeline_wan.py` | Set `transformer._mag_cache_expected_steps` / `transformer_2._mag_cache_expected_steps` instead of calling into hook internals. |
| `src/diffusers/pipelines/wan/pipeline_wan_i2v.py` | Same as above. |
| `src/diffusers/pipelines/wan/pipeline_wan_vace.py` | Same as above. |

All pipeline changes are no-ops when MagCache is not applied.

## Evidence

Run locally, CPU only, no GPU or model weights required. Uses tiny `WanTransformerBlock` instances (`dim=32`) just to satisfy the hook registry.

### 1. The bug this PR fixes: shared config corruption under asymmetric split

Before this PR, the mutating approach corrupted a config object shared across two transformers. Reproduced directly against a plain `MagCacheConfig`, no pipeline needed:

```python
from diffusers.hooks.mag_cache import MagCacheConfig, nearest_interp

shared = MagCacheConfig(mag_ratios=[1.0, 0.9, 0.8, 0.7], num_inference_steps=10)

# old mutating approach, simulated inline since update_mag_cache_num_steps is removed
shared.num_inference_steps = 7
shared.mag_ratios = nearest_interp(shared._original_mag_ratios, 7)
print("after setting for transformer:", shared.num_inference_steps)

shared.num_inference_steps = 3
shared.mag_ratios = nearest_interp(shared._original_mag_ratios, 3)
print("after setting for transformer_2:", shared.num_inference_steps)

print("what transformer now reads:", shared.num_inference_steps)
```

Output:
after setting for transformer: 7
after setting for transformer_2: 3
what transformer now reads: 3

`transformer` was supposed to run 7 steps but the shared config now reports 3, because `transformer_2`'s update overwrote it. This is the exact corruption the reviewer flagged.

### 2. After the fix, the same shared config is never touched

```python
from diffusers.hooks.mag_cache import MagCacheConfig, apply_mag_cache, _get_effective_num_steps
import torch.nn as nn
from diffusers.models.transformers.transformer_wan import WanTransformerBlock

def make_tiny():
    class T(nn.Module):
        def __init__(self):
            super().__init__()
            self.transformer_blocks = nn.ModuleList([
                WanTransformerBlock(dim=32, ffn_dim=64, num_heads=2,
                    qk_norm="rms_norm_across_heads", cross_attn_norm=True,
                    eps=1e-6, added_kv_proj_dim=None)
                for _ in range(3)
            ])
    return T()

shared = MagCacheConfig(mag_ratios=[1.0, 0.9, 0.8, 0.7], num_inference_steps=10)
t1, t2 = make_tiny(), make_tiny()
apply_mag_cache(t1, shared)
apply_mag_cache(t2, shared)

t1._mag_cache_expected_steps = 7
t2._mag_cache_expected_steps = 3

print("t1 effective steps:", _get_effective_num_steps(shared, t1))
print("t2 effective steps:", _get_effective_num_steps(shared, t2))
print("shared.num_inference_steps, unchanged:", shared.num_inference_steps)
```

Output:
t1 effective steps: 7
t2 effective steps: 3
shared.num_inference_steps, unchanged: 10

Each transformer sees its own correct step count and the shared config object is never written to.

### 3. Full unit test suite
$ python -m pytest tests/hooks/test_mag_cache_wan22_steps.py -v
19 passed in 0.30s

Covers, beyond the case above:

- Retention window computed against the effective per transformer step count, not the global one.
- `mag_ratios` re-interpolated per transformer without mutating `config.mag_ratios`.
- Asymmetric splits `(3,2)`, `(7,3)`, `(1,9)` on a single shared config object.
- Single transformer pipelines (Flux, SDXL, Wan 2.1) fall back to `config.num_inference_steps` unchanged when `_mag_cache_expected_steps` was never set, confirming no behavior change there.
- Hooks correctly hold a reference to the root module they were applied to.

### 4. Double interpolation is still avoided

Unchanged from the original fix, still verified by the test suite: `mag_ratios` are always re-computed from `_original_mag_ratios`, never from an already interpolated `mag_ratios`, so calling this multiple times at different step counts does not compound rounding error.

## Scope note

This PR only fixes the step accounting and config mutation issues described above. The separately reported quality degradation on heavily step distilled models (50 step calibration collapsed into a 4 step schedule) is a distinct issue tracked in #14025 and not addressed here.

### DIFF
--- a/src/diffusers/hooks/mag_cache.py
+++ b/src/diffusers/hooks/mag_cache.py
@@ -130,6 +130,7 @@ class MagCacheConfig:
             if not torch.is_tensor(self.mag_ratios):
                 self.mag_ratios = torch.tensor(self.mag_ratios)
 
+            self._original_mag_ratios = self.mag_ratios.clone()
             if len(self.mag_ratios) != self.num_inference_steps:
                 logger.debug(
                     f"Interpolating mag_ratios from length {len(self.mag_ratios)} to {self.num_inference_steps}"
@@ -407,6 +408,7 @@ def apply_mag_cache(module: torch.nn.Module, config: MagCacheConfig) -> None:
     # Initialize registry on the root module so the Pipeline can set context.
     HookRegistry.check_if_exists_or_initialize(module)
 
+    module._mag_cache_config = config
     state_manager = StateManager(MagCacheState, (), {})
     remaining_blocks = []
 
@@ -466,3 +468,15 @@ def _apply_mag_cache_block_hook(
 
     hook = MagCacheBlockHook(state_manager, is_tail, config)
     registry.register_hook(hook, _MAG_CACHE_BLOCK_HOOK)
+
+
+def update_mag_cache_num_steps(module: torch.nn.Module, num_steps: int) -> None:
+    config: MagCacheConfig = getattr(module, "_mag_cache_config", None)
+    if config is None:
+        return
+    original_ratios = getattr(config, "_original_mag_ratios", config.mag_ratios)
+    config.num_inference_steps = num_steps
+    if original_ratios is not None:
+        config.mag_ratios = nearest_interp(original_ratios, num_steps)
+
+
--- a/src/diffusers/hooks/mag_cache.py
+++ b/src/diffusers/hooks/mag_cache.py
@@ -169,12 +169,25 @@ class MagCacheState(BaseState):
         self.calibration_ratios = []
 
 
+def _get_effective_num_steps(config: MagCacheConfig, root_module: torch.nn.Module) -> int:
+    expected = getattr(root_module, "_mag_cache_expected_steps", None)
+    return expected if expected is not None else config.num_inference_steps
+
+
+def _get_effective_mag_ratios(config: MagCacheConfig, effective_num_steps: int) -> torch.Tensor:
+    if effective_num_steps == config.num_inference_steps:
+        return config.mag_ratios
+    original = getattr(config, "_original_mag_ratios", config.mag_ratios)
+    return nearest_interp(original, effective_num_steps)
+
+
 class MagCacheHeadHook(ModelHook):
     _is_stateful = True
 
-    def __init__(self, state_manager: StateManager, config: MagCacheConfig):
+    def __init__(self, state_manager: StateManager, config: MagCacheConfig, root_module: torch.nn.Module):
         self.state_manager = state_manager
         self.config = config
+        self.root_module = root_module
         self._metadata = None
 
     def initialize_hook(self, module):
@@ -200,13 +213,16 @@ class MagCacheHeadHook(ModelHook):
             should_compute = True
         else:
             # MagCache Logic
+            effective_num_steps = _get_effective_num_steps(self.config, self.root_module)
+            effective_mag_ratios = _get_effective_mag_ratios(self.config, effective_num_steps)
+
             current_step = state.step_index
-            if current_step >= len(self.config.mag_ratios):
+            if current_step >= len(effective_mag_ratios):
                 current_scale = 1.0
             else:
-                current_scale = self.config.mag_ratios[current_step]
+                current_scale = effective_mag_ratios[current_step]
 
-            retention_step = int(self.config.retention_ratio * self.config.num_inference_steps + 0.5)
+            retention_step = int(self.config.retention_ratio * effective_num_steps + 0.5)
 
             if current_step >= retention_step:
                 state.accumulated_ratio *= current_scale
@@ -286,11 +302,18 @@ class MagCacheHeadHook(ModelHook):
 
 
 class MagCacheBlockHook(ModelHook):
-    def __init__(self, state_manager: StateManager, is_tail: bool = False, config: MagCacheConfig = None):
+    def __init__(
+        self,
+        state_manager: StateManager,
+        is_tail: bool = False,
+        config: MagCacheConfig = None,
+        root_module: torch.nn.Module = None,
+    ):
         super().__init__()
         self.state_manager = state_manager
         self.is_tail = is_tail
         self.config = config
+        self.root_module = root_module
         self._metadata = None
 
     def initialize_hook(self, module):
@@ -379,7 +402,8 @@ class MagCacheBlockHook(ModelHook):
 
     def _advance_step(self, state: MagCacheState):
         state.step_index += 1
-        if state.step_index >= self.config.num_inference_steps:
+        effective_num_steps = _get_effective_num_steps(self.config, self.root_module)
+        if state.step_index >= effective_num_steps:
             # End of inference loop
             if self.config.calibrate:
                 print("\n[MagCache] Calibration Complete. Copy these values to MagCacheConfig(mag_ratios=...):")
@@ -426,31 +450,36 @@ def apply_mag_cache(module: torch.nn.Module, config: MagCacheConfig) -> None:
     if len(remaining_blocks) == 1:
         name, block = remaining_blocks[0]
         logger.info(f"MagCache: Applying Head+Tail Hooks to single block '{name}'")
-        _apply_mag_cache_block_hook(block, state_manager, config, is_tail=True)
-        _apply_mag_cache_head_hook(block, state_manager, config)
+        _apply_mag_cache_block_hook(block, state_manager, config, module, is_tail=True)
+        _apply_mag_cache_head_hook(block, state_manager, config, module)
         return
 
     head_block_name, head_block = remaining_blocks.pop(0)
     tail_block_name, tail_block = remaining_blocks.pop(-1)
 
     logger.info(f"MagCache: Applying Head Hook to {head_block_name}")
-    _apply_mag_cache_head_hook(head_block, state_manager, config)
+    _apply_mag_cache_head_hook(head_block, state_manager, config, module)
 
     for name, block in remaining_blocks:
-        _apply_mag_cache_block_hook(block, state_manager, config)
+        _apply_mag_cache_block_hook(block, state_manager, config, module)
 
     logger.info(f"MagCache: Applying Tail Hook to {tail_block_name}")
-    _apply_mag_cache_block_hook(tail_block, state_manager, config, is_tail=True)
+    _apply_mag_cache_block_hook(tail_block, state_manager, config, module, is_tail=True)
 
 
-def _apply_mag_cache_head_hook(block: torch.nn.Module, state_manager: StateManager, config: MagCacheConfig) -> None:
+def _apply_mag_cache_head_hook(
+    block: torch.nn.Module,
+    state_manager: StateManager,
+    config: MagCacheConfig,
+    root_module: torch.nn.Module,
+) -> None:
     registry = HookRegistry.check_if_exists_or_initialize(block)
 
     # Automatically remove existing hook to allow re-application (e.g. switching modes)
     if registry.get_hook(_MAG_CACHE_LEADER_BLOCK_HOOK) is not None:
         registry.remove_hook(_MAG_CACHE_LEADER_BLOCK_HOOK)
 
-    hook = MagCacheHeadHook(state_manager, config)
+    hook = MagCacheHeadHook(state_manager, config, root_module)
     registry.register_hook(hook, _MAG_CACHE_LEADER_BLOCK_HOOK)
 
 
@@ -458,6 +487,7 @@ def _apply_mag_cache_block_hook(
     block: torch.nn.Module,
     state_manager: StateManager,
     config: MagCacheConfig,
+    root_module: torch.nn.Module,
     is_tail: bool = False,
 ) -> None:
     registry = HookRegistry.check_if_exists_or_initialize(block)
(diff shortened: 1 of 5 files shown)


lcheng321:
@sayakpaul I saw your name in the git blame history for mag_cache.py and noticed you are also a maintainer on this file, would you have time to review this PR. It fixes the per-transformer step accounting bug reported in #14025, where num_inference_steps was set to the full pipeline count instead of the actual per-transformer execution count. I verified the fix locally with a CPU only repro and 20 passing unit tests covering asymmetric splits and edge cases. Happy to answer any questions about the approach.

sayakpaul:
Cc: @AlanPonnachan

lcheng321:
@AlanPonnachan any chance you could take a look when free?

AlanPonnachan:
Sorry for the late reply. I'll review it today

src/diffusers/hooks/mag_cache.py
```diff
@@ -466,3 +468,15 @@ def _apply_mag_cache_block_hook(
...
+
+
+def update_mag_cache_num_steps(module: torch.nn.Module, num_steps: int) -> None:
+    config: MagCacheConfig = getattr(module, "_mag_cache_config", None)
+    if config is None:
+        return
+    original_ratios = getattr(config, "_original_mag_ratios", config.mag_ratios)
+    config.num_inference_steps = num_steps
+    if original_ratios is not None:
+        config.mag_ratios = nearest_interp(original_ratios, num_steps)
```
AlanPonnachan: It's recommended to keep `MagCacheConfig` fixed during pipeline execution rather than modifying it dynamically. If a user passes the same `MagCacheConfig` object to both transformers, or reuses the pipeline for a new request with a different step count, mutating `config.num_inference_steps` and `config.mag_ratios` in place will corrupt the configuration for subsequent runs?

src/diffusers/pipelines/wan/pipeline_wan.py
```diff
@@ -586,6 +586,15 @@ def __call__(
...
             boundary_timestep = None
 
+        if boundary_timestep is not None:
+            from ...hooks.mag_cache import update_mag_cache_num_steps
+            n_steps_t1 = sum(1 for t in timesteps if t >= boundary_timestep)
+            n_steps_t2 = len(timesteps) - n_steps_t1
+            if self.transformer is not None:
+                update_mag_cache_num_steps(self.transformer, n_steps_t1)
+            if self.transformer_2 is not None:
+                update_mag_cache_num_steps(self.transformer_2, n_steps_t2)
```
AlanPonnachan: we should avoid tightly coupling the core pipelines to specific hook implementations (like importing `update_mag_cache_num_steps` directly). I think a better solution would be
 can we just have the pipeline generically broadcast the expected step counts to the modules? For example:
> ```python
> n_steps_t1 = sum(1 for t in timesteps if t >= boundary_timestep)
> if self.transformer is not None:
>     self.transformer._expected_steps = n_steps_t1
> if self.transformer_2 is not None:
>     self.transformer_2._expected_steps = len(timesteps) - n_steps_t1
> ```

Then, `MagCacheBlockHook` can simply read `getattr(module, "_expected_steps", default)` during initialization. This keeps the pipeline cleanly decoupled from hook internals."

src/diffusers/hooks/mag_cache.py
lcheng321: Fixed in 6d19597.

Config was stored by reference and read live every step. Same config on both transformers under an asymmetric split meant the second update_mag_cache_num_steps call clobbered the first.

Now config is never mutated. Each transformer reads its own step count from a plain attribute on the root module (_mag_cache_expected_steps), falling back to config.num_inference_steps when unset. update_mag_cache_num_steps is gone, nothing left to mutate.

```python
def test_shared_config_object_untouched_by_apply(self):
    shared = _make_config(self.RATIOS, self.TOTAL)
    t1, t2 = _make_tiny_transformer(), _make_tiny_transformer()
    apply_mag_cache(t1, shared)
    apply_mag_cache(t2, shared)
    assert shared.num_inference_steps == self.TOTAL
    assert len(shared.mag_ratios) == self.TOTAL
```

Also covered asymmetric splits (3,2), (7,3), (1,9) on a shared config. 19/19 tests pass.

src/diffusers/pipelines/wan/pipeline_wan.py
lcheng321: Fixed in 6d19597, close to your suggestion.

Pipeline just sets a plain attribute now:

```python
if self.transformer is not None:
    self.transformer._mag_cache_expected_steps = n_steps_t1
if self.transformer_2 is not None:
    self.transformer_2._mag_cache_expected_steps = n_steps_t2
```

One tweak: reading it during initialize_hook doesn't work, that runs before the pipeline computes boundary_timestep, so the attribute isn't set yet. Hooks also live on individual blocks, not the root transformer. So hooks hold a root_module reference and read the attribute lazily at forward/reset time instead.

Issue #3212, opened 2023-04-24
Add clip skip for diffusion pipeline


NormXU:
### Introduction
clip skip is a trick to feed the early-stopped features encoded by `CLIPTextModel` into the cross-attention.   If `clip_skip = 2`, it means that we want to use the features from the layer before the last of the clip text encoder to guide our image generation. And our current diffusion pipeline can be regarded as `clip_skip = 1`, which means that we just use the feature from the last layer of clip text encoder.

Here is a brief introduction to clip skip [webui-wiki](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Features#clip-skip) and related discussion [link](https://github.com/AUTOMATIC1111/stable-diffusion-webui/discussions/5674) 

A dominant majority of models need `clip_skip=2` to reach a more aesthetic generation. I think adding this feature can give people more choices to optimize their generation.

### Implementation
Adding clip_skip into diffusers is both simple and difficult. The main idea of clip_skip is simple, however, since our text encoder is imported from transformers, it is not easy to hack the `CLIPTexModel` in diffusers.

To do so, we need to overwrite `CLIPTextModel `and `CLIPTextTransformer`. Here is my implementation:

```python
class MyCLIPTextTransformer(CLIPTextTransformer):

    def forward(
        self,
        input_ids: Optional[torch.Tensor] = None,
        attention_mask: Optional[torch.Tensor] = None,
        position_ids: Optional[torch.Tensor] = None,
        output_attentions: Optional[bool] = None,
        output_hidden_states: Optional[bool] = None,
        return_dict: Optional[bool] = None,
        clip_skip: Optional[int] = 1, # <-- newly added: take the last N layer of encoder as output
    ) -> Union[Tuple, BaseModelOutputWithPooling]:

        output_attentions = output_attentions if output_attentions is not None else self.config.output_attentions
        output_hidden_states = (
            output_hidden_states if output_hidden_states is not None else self.config.output_hidden_states
        )
        return_dict = return_dict if return_dict is not None else self.config.use_return_dict

        if input_ids is None:
            raise ValueError("You have to specify input_ids")

        input_shape = input_ids.size()
        input_ids = input_ids.view(-1, input_shape[-1])

        hidden_states = self.embeddings(input_ids=input_ids, position_ids=position_ids)

        bsz, seq_len = input_shape
        # CLIP's text model uses causal mask, prepare it here.
[...]

PR #14029, opened 2026-06-21
Fix kohya FLUX CLIP text-encoder LoRA loading under transformers>=5 (#13984)

### LINKED ISSUE #13984: FLUX LoRA with CLIP text-encoder weights fails (empty rank -> IndexError) under transformers>=5
### Describe the bug

Loading a kohya-style FLUX LoRA that contains **CLIP text-encoder** weights (`lora_te1_*`) fails with `IndexError: list index out of range` when `transformers>=5` is installed.

The text-encoder half of the LoRA is never loaded; `pipe.load_lora_weights(...)` raises before any adapter is injected.

### Root cause

`transformers>=5` flattened `CLIPTextModel`: the `text_model.` wrapper module was removed, so `text_encoder.named_modules()` now yields names like `encoder.layers.0.self_attn.k_proj` instead of `text_model.encoder.layers.0.self_attn.k_proj`.

The kohya→diffusers conversion (`_convert_kohya_flux_lora_to_diffusers`) still produces text-encoder keys prefixed with `text_model.`, e.g. `text_encoder.text_model.encoder.layers.0.self_attn.k_proj.lora_B.weight`.

In `diffusers/loaders/lora_base.py::_load_lora_into_text_encoder`, the `rank` dict is built by matching `text_encoder.named_modules()` against the (converted, PEFT-format) state-dict keys:

```python
for name, _ in text_encoder.named_modules():
    if name.endswith((".q_proj", ".k_proj", ".v_proj", ".out_proj", ".fc1", ".fc2")):
        rank_key = f"{name}.lora_B.weight"   # e.g. "encoder.layers.0.self_attn.k_proj.lora_B.weight"
        if rank_key in state_dict:           # but keys are "text_model.encoder.layers.0...."
            rank[rank_key] = state_dict[rank_key].shape[1]
```

Under transformers 5 the module names no longer carry the `text_model.` prefix while the state-dict keys still do, so **nothing matches**, `rank` stays empty, and `_create_lora_config` → `get_peft_kwargs` does `r = lora_alpha = list(rank_dict.values())[0]` → `IndexError`.

### Reproduction

```python
import torch

# any kohya FLUX LoRA that includes CLIP text-encoder weights (lora_te1_*)
pipe.load_lora_weights("kohya_flux_lora_with_text_encoder.safetensors")
```

Minimal isolation of the empty-rank step:

```python
from transformers import CLIPTextModel
te = CLIPTextModel.from_pretrained("black-forest-labs/FLUX.1-dev", subfolder="text_encoder")
print(hasattr(te, "text_model"))                       # False on transformers>=5 (was True on 4.x)
print([n for n, _ in te.named_modules()][:6])          # ['', 'embeddings', ..., 'encoder', 'encoder.layers']
# -> module names are now un-prefixed, but converted LoRA keys are still 'text_model.encoder....'
```

### Traceback

```
File ".../diffusers/loaders/lora_base.py", line 390, in _load_lora_into_text_encoder
    lora_config = _create_lora_config(state_dict, network_alphas, metadata, rank, is_unet=False)
File ".../diffusers/utils/peft_utils.py", line 355, in _create_lora_config
    lora_config_kwargs = get_peft_kwargs(...)
File ".../diffusers/utils/peft_utils.py", line 158, in get_peft_kwargs
    r = lora_alpha = list(rank_dict.values())[0]
IndexError: list index out of range
```

### Suggested fix

Reconcile the converted text-encoder key namespace with the model's `named_modules()` for `transformers>=5` (drop/normalize the stale `text_model.` segment), or build `rank` in a prefix-tolerant way. A clearer error than a bare `IndexError` when `rank` ends up empty would also help.

### System Info

- diffusers 0.38.0
- transformers 5.9.0
- peft 0.19.1
- torch 2.10–2.11 (+cu128), Python 3.11/3.12, Linux
- Model: black-forest-labs/FLUX.1-dev

### Who can help?

@sayakpaul @BenjaminBossan

Jefsky:
Fixes https://github.com/huggingface/diffusers/issues/13984

## Root cause

`transformers>=5` flattened `CLIPTextModel`: the `text_model.` wrapper module was removed, so `text_encoder.named_modules()` now yields names like `encoder.layers.0.self_attn.k_proj` instead of `text_model.encoder.layers.0.self_attn.k_proj`.

The kohya→diffusers conversion still emits text-encoder keys prefixed with `text_model.` (e.g. `text_encoder.text_model.encoder.layers.0.self_attn.k_proj.lora_B.weight`). In `_load_lora_into_text_encoder`, the `rank` dict is built by matching `named_modules()` against the converted state-dict keys — under transformers>=5 nothing matches, `rank` stays empty, and `get_peft_kwargs` does `list(rank_dict.values())[0]` → `IndexError`.

The PEFT-side fix (#3212) doesn't help here because the crash happens before any PEFT state-dict injection (confirmed by the issue reporter).

## Fix

In `_load_lora_into_text_encoder`, after the `convert_state_dict_to_peft` call, strip the stale `text_model.` prefix from the converted state-dict keys when the text encoder doesn't have the `text_model` submodule (i.e. the transformers>=5 layout). The check uses `hasattr(text_encoder, "text_model")`, not a version number, so it's forward-compatible.

## Test

Added `FluxLoRATests::test_kohya_clip_text_encoder_flattened_compat` — a fast CPU regression test that:

1. Builds the dummy FLUX pipeline
2. Removes the `text_model` attribute from `text_encoder` (simulating transformers>=5)
3. Passes a synthetic kohya-style state dict with the stale `text_model.` prefix
4. Asserts the adapter is correctly injected (no IndexError)

## Verification

- Before fix: the synthetic state dict with `text_model.` prefix causes `IndexError: list index out of range` in `get_peft_kwargs`
- After fix: the adapter loads correctly on both flattened (transformers>=5) and traditional (transformers<5, hasattr(text_model)=True) layouts

### DIFF
--- a/src/diffusers/loaders/lora_base.py
+++ b/src/diffusers/loaders/lora_base.py
@@ -376,6 +376,14 @@ def _load_lora_into_text_encoder(
         # convert state dict
         state_dict = convert_state_dict_to_peft(state_dict)
 
+        # Under transformers>=5, CLIPTextModel was flattened and no longer has the
+        # `text_model.` wrapper module, so named_modules() returns unprefixed names
+        # like "encoder.layers.0.self_attn.q_proj". Kohya-sourced LoRA state dict
+        # keys still carry the "text_model." prefix from the old module layout.
+        # Strip it so that rank-key matching works regardless of the transformers version.
+        if not hasattr(text_encoder, "text_model"):
+            state_dict = {k.removeprefix("text_model."): v for k, v in state_dict.items()}
+
         for name, _ in text_encoder.named_modules():
             if name.endswith((".q_proj", ".k_proj", ".v_proj", ".out_proj", ".fc1", ".fc2")):
                 rank_key = f"{name}.lora_B.weight"
--- a/tests/lora/test_lora_layers_flux.py
+++ b/tests/lora/test_lora_layers_flux.py
@@ -256,6 +256,48 @@ class FluxLoRATests(unittest.TestCase, PeftLoraLoaderMixinTests):
             "LoRA should lead to different results.",
         )
 
+    def test_kohya_clip_text_encoder_flattened_compat(self):
+        """Regression test for #13984: kohya FLUX CLIP text-encoder LoRA under transformers>=5.
+
+        Under transformers>=5, CLIPTextModel was flattened and no longer has the `text_model.`
+        wrapper module. Kohya-sourced state dict keys still carry the "text_model." prefix,
+        which caused an empty rank dict and IndexError. This test verifies the fix by passing
+        a synthetic kohya-style state dict with the stale prefix to a text encoder that
+        doesn't have the `text_model` submodule.
+        """
+        components, _, _ = self.get_dummy_components(FlowMatchEulerDiscreteScheduler)
+        pipe = self.pipeline_class(**components)
+        pipe = pipe.to(torch_device)
+        pipe.set_progress_bar_config(disable=None)
+
+        # Simulate transformers>=5 flattened CLIPTextModel by removing the text_model
+        # wrapper module attribute. Under transformers>=5 named_modules() no longer
+        # includes this prefix.
+        if hasattr(pipe.text_encoder, "text_model"):
+            del pipe.text_encoder.text_model
+
+        # Build a synthetic kohya-style state dict whose text-encoder keys carry the stale
+        # "text_model." prefix that real kohya conversion produces.
+        state_dict = {}
+        with torch.no_grad():
+            for name, module in pipe.text_encoder.named_modules():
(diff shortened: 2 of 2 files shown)


BenjaminBossan (commented):
Thanks for the PR. I agree that it should fix the specific issue that was mentioned and that should be safe to apply. My concern is that it is very specialized to this case and not a general solution. The PR covers this exact case:

https://github.com/huggingface/transformers/blob/bfd3604d83e84d7ff8bbc18bc09c21e8282d31f9/src/transformers/conversion_mapping.py#L608

But there could be other entries in the conversion mapping, now or added in the future, which are not covered by this patch. So I wonder if we should instead call Transformers `get_model_conversion_mapping` on the model (if it's a Transformers model) and apply the conversions from there.

I'll leave it up to the Diffusers maintainers to decide how to deal with this.

PR #14040, opened 2026-06-22
Add Boogu-Image generation, editing, and turbo pipelines

Boogu-Team:
# What does this PR do?

Adds the **Boogu-Image** family of pipelines to `diffusers`:

- **`BooguImagePipeline`** — text-to-image generation and instruction-based image editing.
- **`BooguImageTurboPipeline`** — few-step DMD distilled generation.
- **fp8** quantized inference examples for both, targeting the published fp8 checkpoints.

The integration is **purely additive**: it introduces new files only and does not modify any existing upstream module.

Published checkpoints (Hugging Face Hub, `Boogu/…`): `Boogu-Image-0.1-Base`, `Boogu-Image-0.1-Edit`, `Boogu-Image-0.1-Turbo`, and their `-fp8` variants.

## Pipelines & model

| Component | File |
|---|---|
| Generation / edit pipeline | `src/diffusers/pipelines/boogu/pipeline_boogu.py` |
| Turbo (DMD few-step) pipeline | `src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py` |
| Transformer backbone (model, attention module + processors, RoPE) | `src/diffusers/models/transformers/transformer_boogu.py` |
| Image processor | `src/diffusers/pipelines/boogu/image_processor.py` |

Docs: `docs/source/en/api/pipelines/boogu.md`. Runnable examples: `examples/boogu/` (base / edit / turbo + fp8, with `README.md`).

## Convention compliance

Implemented against the repo's `.ai` rules:

- **Single-file model.** The transformer backbone, its attention module (`BooguImageAttention`) + stateless processors, and the RoPE helper all live in `transformer_boogu.py` (one model = one file).
- **Attention.** Routed through **`dispatch_attention_fn`** (no direct `F.scaled_dot_product_attention` in the forward path); attention masks always materialized as `[B, 1, 1, L]` bool to stay bit-exact with the trained checkpoints under the native bf16 backend. `BooguImageAttention` (modelled on `Flux2Attention`) holds the `to_q`/`to_k`/`to_v`/`norm_q`/`norm_k`/`to_out` layers and dispatches to a stateless processor.
- **Inference-only.** No weight-init (`initialize_weights`) — the model loads pretrained weights. `__init__` follows the lazy convention.
- **Turbo pipeline.** `BooguImageTurboPipeline` is a **standalone `DiffusionPipeline`** (not a subclass of `BooguImagePipeline`), with shared methods carried via `# Copied from` and kept in sync by `make fix-copies`. Device placement / offloading / component registration reuse the `DiffusionPipeline` base class.
- **Standard CFG.** The pipeline exposes standard classifier-free guidance only. No dead code paths, no silent `except Exception` fallbacks, no unused "API-consistency" parameters — training / ablation / prompt-tuning code from the research repo was removed; only the inference path is integrated.

## Verification

- All 6 `examples/boogu/` scripts run end-to-end and produce correct images (base / edit / turbo + fp8).
- Every refactor verified against a pre-refactor reference. The **base path (T2I + CFG, which exercises both attention processors) is bit-exact (`maxdiff = 0`)** on GPU; the **edit path (TI2I double-guidance) matches within that path's inherent GPU nondeterminism** (~8–9e-2, confirmed equal to a same-code self-vs-self run). Checkpoints load strict (no missing/unexpected keys).
- CI gates pass locally: `ruff check`, `ruff format --check`, `make fix-copies` (no diff), `check_dummies`.
- Test suite under `tests/pipelines/boogu/` and `tests/models/transformers/test_models_transformer_boogu.py` passes.

## Notes for reviewers

- **Double-stream attention & checkpoint keys.** The single-stream attention processor is fully stateless. For the double-stream block, the per-stream projections (`img_to_q` / `instruct_to_q` / `img_out` / …) are stored in the *already-published* checkpoints under `…img_instruct_attn.processor.*`. Moving them onto `BooguImageAttention` to make that processor fully stateless would rename those state-dict keys and require re-saving the released Hub checkpoints. To keep existing downloads loading, those projections stay on the processor module for now; happy to do the key migration + re-publish in a follow-up if you prefer the fully-stateless form.
- **Boosted Orthogonal Guidance (BOG)** has been removed — it defaulted to off, so standard-CFG output is unchanged. We'd like to re-introduce it as its own guider class under `src/diffusers/guiders/` (and route advanced features through modular diffusers) in a follow-up PR, keeping this one focused on the standard pipeline.
- **TeaCache** was removed from both the model and the pipeline; caching is deferred to `CacheMixin` / FirstBlockCache (and the WIP in #12652).
- **`_keep_in_fp32_modules`** is intentionally not set: forcing `time_caption_embed` to fp32 changes bf16 inference numerics, so we left it to the model author / reviewer's call.
- The fp8 examples include a small DeepGEMM-disable shim with version branches that are load-bearing for `transformers` 5.10.x (the env var alone does not disable DeepGEMM there).

### Required Hub-side checkpoint changes (no code change in this PR)

> **Important for reviewers:** the published `Boogu/Boogu-Image-0.1-*` checkpoints on the Hub are still packaged for the **old custom-remote-code loading path** and will *not* load correctly against this PR's native diffusers classes until the Hub repos are updated. We will push these Hub-repo edits to land together with the PR. Diff below is Hub `main` → required state.

**1. `model_index.json` — point component classes at `diffusers`, not the bundled remote-code modules.**
On Hub `main` it currently is:
```json
"scheduler":   ["scheduling_flow_match_euler_discrete_time_shifting", "FlowMatchEulerDiscreteScheduler"],
"transformer": ["transformer_boogu", "BooguImageTransformer2DModel"],
```
Both must become library refs so `from_pretrained` resolves the in-tree classes added by this PR:
```json
"scheduler":   ["diffusers", "FlowMatchEulerDiscreteScheduler"],
"transformer": ["diffusers", "BooguImageTransformer2DModel"],
```

**2. Delete the two remote-code shim modules** (they are thin re-export stubs that `import boogu …` from the private research package and raise `ModuleNotFoundError: boogu` for any external user):
- `transformer/transformer_boogu.py`
- `scheduler/scheduling_flow_match_euler_discrete_time_shifting.py`

**3. `scheduler/scheduler_config.json` — drop the legacy custom-scheduler keys.** Hub `main` has:
```json
{ "_class_name": "FlowMatchEulerDiscreteScheduler", "_diffusers_version": "0.33.1",
  "do_shift": true, "dynamic_time_shift": false, "time_shift_version": "v1",
  "seq_len": 4096, "num_train_timesteps": 1000 }
```
Remove `do_shift`, `dynamic_time_shift`, `time_shift_version` (these drove the old custom scheduler's time-shift; the official scheduler ignores them). **Keep `seq_len`** — the pipeline's time-shift adapter reads `scheduler.config.seq_len` to compute the shift `mu`, so it is load-bearing → final config keeps `_class_name` / `_diffusers_version` / `seq_len` / `num_train_timesteps`.

**4. `transformer/config.json` — remove `prompt_tuning_configs`** (the prompt-tuning subsystem was dropped from this integration; the key is otherwise ignored with a warning).

The same four edits apply to all six repos (`Base`, `Edit`, `Turbo` and their `-fp8` variants). The `mllm/`, `processor/`, `vae/` and weight files are unchanged.

---

Fixes # (issue)

## Before submitting
- [x] Are you the author (or part of the team) of the model/pipeline (only applicable for model/pipeline related PRs)?

## Who can review?


src/diffusers/ops/triton/layer_norm.py
```diff
@@ -0,0 +1,1261 @@
+# This repository is a fork by Boogu Team; modifications have been made.
```
yiyixuxu: I wonder how much performance difference you're seeing (for inference)? can we remove this?

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1117 @@
+"""
```
yiyixuxu: we follows a single-file convention for models -> can you combine everything into the same file?

src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py
```diff
@@ -0,0 +1,1417 @@
...
+        num_enabled_offload_flags = sum(int(x) for x in enabled_offload_flags)
+        assert num_enabled_offload_flags <= 1, (
+            "At most one pipeline offload strategy can be enabled at a time. "
+            f"Got enable_model_cpu_offload_flag={enable_model_cpu_offload_flag}, "
+            f"enable_sequential_cpu_offload_flag={enable_sequential_cpu_offload_flag}, "
+            f"enable_group_offload_flag={enable_group_offload_flag}."
+        )
+
+    # Copied from diffusers.pipelines.boogu.pipeline_boogu.BooguImagePipeline.devices_manager
+    def devices_manager(
```
yiyixuxu: can we remove this?

src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py
```diff
@@ -0,0 +1,1417 @@
...
+            device=latents.device,
+            dtype=latents.dtype,
+        )
+        return (1 - sigma_expanded) * noise + sigma_expanded * latents
+
+    # ------------------------------------------------------------------ #
+    # Shared device / component utilities (copied from BooguImagePipeline) #
+    # ------------------------------------------------------------------ #
+    # Copied from diffusers.pipelines.boogu.pipeline_boogu.BooguImagePipeline._validate_device_format
+    def _validate_device_format(
```
yiyixuxu: can you remove this?

src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py
```diff
@@ -0,0 +1,1417 @@
...
+        self,
+        device: Literal[None, "cpu", "cuda", "cuda:x"] = "cpu",
+    ):
+        # get_device_validator() raises on an unsupported device string (e.g. "gpu", "cuda:x").
+        get_device_validator()(device.lower() if isinstance(device, str) else device)
+
+    # Copied from diffusers.pipelines.boogu.pipeline_boogu.BooguImagePipeline._check_device_strategy_validity
+    def _check_device_strategy_validity(
```
yiyixuxu: can you remove this?

src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py
```diff
@@ -0,0 +1,1417 @@
...
+                self.to(instant_device_2_use.lower())
+            else:
+                logger.info(
+                    "An offload strategy is enabled, so the user-requested device move to "
+                    "`instant_device_2_use=%r` will be ignored.",
+                    instant_device_2_use,
+                )
+
+    # Copied from diffusers.pipelines.boogu.pipeline_boogu.BooguImagePipeline.set_mllm
+    def set_mllm(self, mllm, device=None):
```
yiyixuxu: so a pipeline subclass like `BooguImageTurboPipeline`, should only have pipeline methods that are pipeline-specific steps, e.g. `encode_prompt`, `prepare_latents`, `check_inputs`. stuff like device management, offloading, setting components -> these are Base pipeline logic, you should re-use the ones implemented in our base class `DiffusionPipeline`, instead of implement custom ones

yiyixuxu (commented):
thanks a lot for the PR and the very thoughtful self-review :) 
i left some comments/questions

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in d202a2376. Merged block_lumina2.py, rope_boogu.py, and attention_processor_boogu.py into transformer_boogu.py so the model lives in a single file. It's a pure relocation — the class bodies are unchanged, so checkpoints load identically and base/edit/turbo stay bit-exact (verified end-to-end on GPU). The two pipelines and the transformer test now import BooguImageRotaryPosEmbed from transformer_boogu.

src/diffusers/ops/triton/layer_norm.py
Boogu-Team: Done ^-^ — removed in 9e672c233. The triton fused RMSNorm (and the flash-attn SwiGLU) were gated behind an `os.getenv("device")` guard that defaulted to "cpu", so the published inference path never actually used them — it always fell back to torch.nn.RMSNorm / a torch SwiGLU. So there's no inference perf to lose: I deleted ops/triton/layer_norm.py, ops/simple_layer_norm.py, the dead env-guard in block_lumina2, and the now-unused is_triton_available helper (which makes import_utils.py byte-identical to upstream again). Output is numerically identical to the previous default path.

src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py
Boogu-Team: Removed in 5cef903ce. The validator_utils helper it depended on was unused elsewhere, so I deleted that file too.

src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py
Boogu-Team: Removed in 5cef903ce. Offload-strategy validation is now handled by the base class's enable_*_offload methods.

src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py
Boogu-Team: Agreed — addressed in 5cef903ce. Removed all the custom infra (devices_manager, set_mllm / set_transformer / set_processor / set_scheduler, _validate_device_format, _check_device_strategy_validity, and the enable_*_offload_flag / user_set_pipe_device state) from both pipelines. They now rely on DiffusionPipeline for device placement (.to() / _execution_device), offloading (enable_model_cpu_offload etc.), and component registration;
__call__ dropped its redundant `device=` kwarg. The only model-specific bit (stripping the MLLM lm_head to keep just the encoder backbone) stays in __init__. As a bonus this made the inherited offload / to-device / float16 pipeline tests pass — those were the ~16 previously-failing device tests, now green.

src/diffusers/pipelines/boogu/pipeline_boogu_turbo.py
Boogu-Team: Removed in 5cef903ce, along with the rest of the custom device/offload layer (see the umbrella reply below). __call__ now resolves the device via the base class's self._execution_device.

Boogu-Team:
> thanks a lot for the PR and the very thoughtful self-review :)
> i left some comments/questions

  Thank you for the quick and thoughtful review, @yiyixuxu — much appreciated! 🙏

  I've addressed all the comments and pushed the changes (commits 9e672c2, d202a23, 5cef903):

  - **single-file convention** — merged the model's helper modules (rope /blocks / attention processors) into `transformer_boogu.py`.
  - **triton fused RMSNorm** — removed; the published path always fell back to `torch.nn.RMSNorm`, so it's numerically identical.
  - **pipeline-subclass responsibilities** — dropped all the custom device / offload / component-setter infrastructure from both pipelines; they now rely on `DiffusionPipeline` (`.to()` / `_execution_device` / `enable_*_offload`). This also made the previously-failing device/offload tests pass.

  I've replied inline on each thread with the specifics and resolved them.
  Happy to iterate further — thanks again!

src/diffusers/pipelines/boogu/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .image_processor import BooguImageProcessor
```
yiyixuxu: can you follow this pattern 
https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/flux2/__init__.py

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+# ----------------------------- RoPE -----------------------------
+class BooguImageRotaryPosEmbed:
+    """Namespace for Boogu's rotary-position-embedding frequency table.
+
+    Only the static `get_freqs_cis` is used (by the pipeline and the transformer's
+    internal double-stream RoPE); it does not hold any state.
+    """
+
+    @staticmethod
+    def get_freqs_cis(
```
yiyixuxu: I think we don't need a module here
just a function `get_freqs_cis` is fine

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+        axes_dim: Tuple[int, int, int], axes_lens: Tuple[int, int, int], theta: int
+    ) -> List[torch.Tensor]:
+        freqs_cis = []
+        freqs_dtype = torch.float32
+        for i, (d, e) in enumerate(zip(axes_dim, axes_lens)):
+            emb = get_1d_rotary_pos_embed(d, e, theta=theta, freqs_dtype=freqs_dtype)
+            freqs_cis.append(emb)
+        return freqs_cis
```
yiyixuxu:
```suggestion
```
it looks exactly same as `BooguImageRotaryPosEmbed`. no?
we only need one of them then

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+        x: torch.Tensor,
+        emb: Optional[torch.Tensor] = None,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+        emb = self.linear(self.silu(emb))
+        scale_msa, gate_msa, scale_mlp, gate_mlp = emb.chunk(4, dim=1)
+        x = self.norm(x) * (1 + scale_msa[:, None])
+        return x, gate_msa, scale_mlp, gate_mlp
+
+
+class LuminaLayerNormContinuous(nn.Module):
```
yiyixuxu: probably rename to BooguImageXXX? if it is same as the Lumina one, you can use a `#Copied from with xx-> `

yiyixuxu: same with other LuminaXX layers here

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+        self.caption_embedder = nn.Sequential(
+            RMSNorm(instruction_feat_dim, eps=norm_eps),
+            nn.Linear(instruction_feat_dim, hidden_size, bias=True),
+        )
+
+        self._initialize_weights()
+
+    def _initialize_weights(self):
+        nn.init.trunc_normal_(self.caption_embedder[1].weight, std=0.02)
+        nn.init.zeros_(self.caption_embedder[1].bias)
```
yiyixuxu:
```suggestion
```

diffusers is inference-only - we do not need worry about initialization usually

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+        self.instruct_to_k = nn.Linear(query_dim, kv_dim, bias=qkv_bias)
+        self.instruct_to_v = nn.Linear(query_dim, kv_dim, bias=qkv_bias)
+
+        # Separate output projections for instruction and image streams.
+        self.instruct_out = nn.Linear(query_dim, query_dim, bias=qkv_bias)
+        self.img_out = nn.Linear(query_dim, query_dim, bias=qkv_bias)
+
+        self.initialize_weights()
+
+    def initialize_weights(self) -> None:
```
yiyixuxu: I think we don't need this here!

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+        head_dim: Dimension of each attention head
+        num_attention_heads: Number of attention heads for queries
+        num_kv_heads: Number of key-value heads
+        qkv_bias: Whether to use bias in QKV linear layers
+    """
+
+    _attention_backend = None
+    _parallel_config = None
+
+    def __init__(
```
yiyixuxu, line 529: Our attention processor should be stateless, only contain computation logic 

Can you implement a `BooguImageAttention` that holds all the layers and can be used together with `BooguImageDoubleStreamSelfAttnProcessor`

See how it is done in flux2 https://github.com/huggingface/diffusers/blob/eb0a900caac8625dd02d4c403b8e7c3ae12b8733/src/diffusers/models/transformers/transformer_flux2.py#L494

`BooguImageAttnProcessor` is already stateless, it just needs to to work with `BooguImageAttention` instead of `Attention` for the layers (`to_q`/`to_k`/...)

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+        else:
+            self.norm1 = RMSNorm(dim, eps=norm_eps)
+
+        self.ffn_norm1 = RMSNorm(dim, eps=norm_eps)
+        self.norm2 = RMSNorm(dim, eps=norm_eps)
+        self.ffn_norm2 = RMSNorm(dim, eps=norm_eps)
+
+        self.initialize_weights()
+
+    def initialize_weights(self) -> None:
```
yiyixuxu: i think we don't need this :)

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+
+        self.initialize_weights()
+
+        # double_stream_processor owns its own q/k/v projections.
+        for param in self.img_instruct_attn.to_q.parameters():
+            param.requires_grad = False
+        for param in self.img_instruct_attn.to_k.parameters():
+            param.requires_grad = False
+        for param in self.img_instruct_attn.to_v.parameters():
+            param.requires_grad = False
```
yiyixuxu:
```suggestion
```

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+
+        del self.img_instruct_attn.to_k
+        del self.img_instruct_attn.to_v
+        del self.img_instruct_attn.to_q
+
+    def initialize_weights(self) -> None:
```
yiyixuxu: can we remove this?

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+
+        # double_stream_processor owns its own q/k/v projections.
+        for param in self.img_instruct_attn.to_q.parameters():
+            param.requires_grad = False
+        for param in self.img_instruct_attn.to_k.parameters():
+            param.requires_grad = False
+        for param in self.img_instruct_attn.to_v.parameters():
+            param.requires_grad = False
+
+        del self.img_instruct_attn.to_k
```
yiyixuxu: I think we won't need these anymore once you implement the img_instruct_attn properly

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+                instruct_norm1_out,
+                instruct_gate_msa,
+                instruct_scale_mlp,
+                instruct_gate_mlp,
+            ) = self.instruct_norm1(instruct_hidden_states, temb)
+            instruct_norm2_out, instruct_shift_mlp, _, _ = self.instruct_norm2(instruct_hidden_states, temb)
+
+            # Step 2: joint attention on [instruct + img].
+            # Call processor directly because Attention.forward does not expose these dual-stream args.
+            joint_attn_out = self.img_instruct_attn.processor(
```
yiyixuxu: you should be able to call self.img_instruct_attn() directly once it's refactored -> it will dispatch to the processor

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+
+        # TeaCache settings
+        self.enable_teacache = False
+        self.teacache_rel_l1_thresh = 0.05
+        self.teacache_params = TeaCacheParams()
+
+        # Polynomial (highest-degree first) rescaling the relative L1 distance used by TeaCache.
+        self.teacache_rescale_coefficients = [-5.48259225, 11.48772289, -4.47407401, 2.47730926, -0.03316487]
+
+    def initialize_weights(self) -> None:
```
yiyixuxu: can you remove this?

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+
+        # Distinguish multiple reference images.
+        self.image_index_embedding = nn.Parameter(torch.randn(5, hidden_size))  # support max 5 ref images
+
+        self.gradient_checkpointing = False
+
+        self.initialize_weights()
+
+        # TeaCache settings
+        self.enable_teacache = False
```
yiyixuxu: for TeaCache, I think we should have it backed in the model 
can you remove the related code for now? 

we've already supported `FirstBlockCache` via `CacheMixin` and there is already a tea cache support WIP https://github.com/huggingface/diffusers/pull/12652

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1885 @@
...
+        nn.init.constant_(self.ref_image_patch_embedder.bias, 0.0)
+
+        nn.init.zeros_(self.norm_out.linear_1.weight)
+        nn.init.zeros_(self.norm_out.linear_1.bias)
+        nn.init.zeros_(self.norm_out.linear_2.weight)
+        nn.init.zeros_(self.norm_out.linear_2.bias)
+
+        nn.init.normal_(self.image_index_embedding, std=0.02)
+
+    def img_patch_embed_and_refine(
```
yiyixuxu: can we move it into `forward`? we want all the layer calls visiable directly in `forward` https://github.com/huggingface/diffusers/blob/main/.ai/models.md#coding-style

src/diffusers/utils/teacache_util.py
```diff
@@ -0,0 +1,41 @@
...
+"""
+
+from dataclasses import dataclass
+from typing import Optional
+
+import torch
+
+
+@dataclass
+class TeaCacheParams:
```
yiyixuxu: can you remove this?

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1766 @@
...
+    """
+    if seq_len is None:
+        seq_len = scheduler.config.seq_len
+
+    # Static v1 shift: mu is a linear function of seq_len between (base_image_seq_len,
+    # base_shift) and (max_image_seq_len, max_shift).
+    slope = (scheduler.config.max_shift - scheduler.config.base_shift) / (
+        scheduler.config.max_image_seq_len - scheduler.config.base_image_seq_len
+    )
+    mu = scheduler.config.base_shift + slope * (seq_len - scheduler.config.base_image_seq_len)
```
yiyixuxu: can we use `calculate_shift` to calculate `mu`?
https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/flux/pipeline_flux.py#L74

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1766 @@
...
+
+    t = np.linspace(0.0, 1.0, num_inference_steps + 1, dtype=np.float32)[:-1]
+    # Boogu v1 == 1 - exponential_shift(mu, 1, 1 - t); reuse the parent's formula.
+    t = (1.0 - scheduler._time_shift_exponential(mu, 1.0, 1.0 - torch.from_numpy(t))).numpy()
+
+    timesteps = torch.from_numpy(t).to(dtype=torch.float32, device=device)
+    scheduler.timesteps = timesteps  # 0-1 sigma, fed to the transformer as the timestep
```
yiyixuxu: instead of overwriting scheduler.timesteps/sigmas/_step_index, can we just create the custom sigmas (with the same logic here to create o-1 sigma) and set it with `retrieve_timesteps(..., sigmas=sigmas)`?

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1766 @@
...
+        max_input_image_side_length: int = 2048 * 2,
+        align_res: bool = True,
+        num_inference_steps: int = 50,
+        text_guidance_scale: float = 4.0,
+        image_guidance_scale: float = 1.0,
+        empty_instruction_guidance_scale: float = 0.0,
+        cfg_range: Tuple[float, float] = (0.0, 1.0),
+        system_prompt_follows_task_type: bool = False,
+        ### Momentum Config
+        use_boosted_orthogonal_guidance: bool = False,
```
yiyixuxu: can we keep the standard pipeline with just CFG for now?  and add add support for modular diffusers and other features through modular system? 

the boosted-orthogonal-guidance can be implemented as its own guider class https://github.com/huggingface/diffusers/tree/main/src/diffusers/guiders

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1766 @@
...
+        use_ref_empty_instruct_pred = (
+            use_empty_neg_instruct_4_ref_img_pred_at_image_guide_in_double_guide
+            or use_empty_neg_instruct_4_ref_img_pred_at_text_guide_in_double_guide
+        )
+
+        enable_teacache = self.transformer.enable_teacache or getattr(
+            self.transformer, "enable_teacache_for_all_layers", False
+        )
+        self.transformer.enable_teacache = enable_teacache
+        if enable_teacache:
```
yiyixuxu: can we remove the tea cache support from pipeline as well?

yiyixuxu (commented):
thanks! 
i left more comments

Boogu-Team:
> thanks! i left more comments

Thanks for the detailed second-round review! I've pushed updates addressing it — one commit per topic so each is easy to review in isolation:

| Commit | What |
|---|---|
| `05cc899ae` | Drop all weight-init (`initialize_weights`), follow the lazy / inference-only `__init__` convention |
| `a597eed0b` | Rename the Lumina-derived layers to `BooguImage*` |
| `36298d3a6` | Collapse the RoPE namespace class into a `get_freqs_cis` function (and drop the dead duplicate) |
| `c9a6b5d06` | Remove TeaCache (model + pipeline) and delete `utils/teacache_util.py`; defer caching to `CacheMixin` / FirstBlockCache |
| `662e6ca11` | Compute the time-shift `mu` via `calculate_shift` (`# Copied from` flux) |
| `2977b2dce` | Introduce `BooguImageAttention` (flux2-style) that holds the layers; processors dispatch through it |
| `3b56e275c` | Remove boosted-orthogonal-guidance; the pipeline now exposes standard CFG only |

I've replied inline on each thread. Of the 20 comments, 18 are done; 2 I've left open with a question for you rather than guess:

1. **`retrieve_timesteps(..., sigmas=)`** (pipeline timesteps) — Boogu feeds the 0→1 sigma value *directly as the timestep*, whereas `set_timesteps(sigmas=)` forces `timesteps = sigmas * num_train_timesteps` + its own shift, which changes the numerics and breaks bit-exactness. I kept the explicit assignment for now and proposed a small scheduler flag as a clean alternative — would value your call on the thread.
2. **Inlining `img_patch_embed_and_refine` into `forward`** — happy to inline per the coding-style doc, just wanted to confirm the preference since `forward` is already large.

Two things I'd also like your steer on:

- **Double-stream attention & checkpoint keys.** The single-stream processor is now fully stateless. For the double-stream block, the per-stream projections (`img_to_q` / `instruct_to_q` / `img_out` / ...) are stored in the *already-published* checkpoints under `...img_instruct_attn.processor.*`. Moving them onto `BooguImageAttention` to make that processor fully stateless would rename those state-dict keys and require re-saving the released Hub checkpoints (Base / Edit / Turbo / Turbo-fp8). To preserve compatibility for existing downloads, I kept those projections on the processor for this round. If you'd prefer the fully-stateless form, we'll do the key migration + re-publish in a follow-up once we can coordinate the Hub re-upload.
- **Boosted Orthogonal Guidance.** I removed it outright here (it defaulted to off, so standard CFG output is unchanged / bit-exact). I'd like to bring it back as its own guider class under `src/diffusers/guiders/` — and route advanced features through modular diffusers — in a follow-up PR, keeping this one focused on the standard pipeline. Does that sequencing work for you?

Verification: every change is checked on GPU. The base path (T2I + CFG, which exercises both attention processors) is **bit-exact (maxdiff=0)** vs the pre-refactor reference; the edit path (TI2I double-guidance) matches within that path's inherent GPU nondeterminism (~8–9e-2, same as a same-code self-vs-self run). `ruff check` / `ruff format` / `check_copies` / `check_dummies` are clean, and the model + pipeline test suites pass. The six `examples/boogu` scripts (base / edit / turbo, each + fp8) all generate images correctly.

Thanks again — happy to iterate on the two open questions.

src/diffusers/pipelines/boogu/__init__.py
Boogu-Team: Done in 05cc899ae. Rewrote `pipelines/boogu/__init__.py` to the `_LazyModule` pattern used by flux2: `image_processor` is always importable, and the pipelines are guarded behind `is_torch_available()` + `is_transformers_available()` with a dummy-objects fallback. Imports are lazy via `_import_structure`.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 36298d3a6. Replaced the `BooguImageRotaryPosEmbed` module with a plain `get_freqs_cis(...)` function. It carried no parameters/buffers, so this is a pure call-site change; the rotary tables are byte-identical and base/edit/turbo stay bit-exact.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Right — it was a dead duplicate. Removed in 36298d3a6; the single `get_freqs_cis` function is now the only RoPE-frequency path.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in a597eed0b. Renamed the Lumina-derived layers to `BooguImage*`. Class names don't appear in the state dict (keys are attribute paths), so renaming is checkpoint-safe — verified strict load with 0 missing/unexpected keys and bit-exact output.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 05cc899ae. Removed the caption-embedder weight init (and the other `initialize_weights` methods across the model) — diffusers loads pretrained weights, so these were never exercised on the inference path.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Removed in 05cc899ae (one of the 5 `initialize_weights` methods dropped this round).

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Removed in 05cc899ae. :)

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 2977b2dce, following the flux2 pattern. Introduced `BooguImageAttention(nn.Module, AttentionModuleMixin)` that holds the per-head layers (to_q/to_k/to_v, norm_q/norm_k, to_out) and dispatches to the processor via its `forward`. The single-stream `BooguImageAttnProcessor` is now fully stateless and reads the layers off `BooguImageAttention` instead of `Attention`. 

qk-norm uses diffusers' RMSNorm so numerics are unchanged; base (T2I+CFG) is bit-exact (maxdiff=0) vs before, strict checkpoint load with no key changes.

One caveat on the double-stream block. The joint instruction<->image attention has *per-stream* projections (img_to_q / instruct_to_q / img_out /instruct_out, ...), and the already-published checkpoints store them under `...img_instruct_attn.processor.img_to_q` etc. Moving those onto `BooguImageAttention` to make `BooguImageDoubleStreamSelfAttnProcessor` fully stateless would rename those state-dict keys, which means re-saving the released Hub checkpoints (Base / Edit / Turbo / Turbo-fp8). We're not in a position to re-key the published weights right now, so for this round I kept the double-stream projections on the processor module to preserve checkpoint compatibility — no weight re-save required, and existing downloads keep loading.
The single-stream processors are stateless as requested.

If you'd prefer we go all the way to a fully stateless double-stream processor, we're happy to do the key migration and re-publish the checkpoints in a follow-up once we can coordinate the Hub re-upload — just let us know.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 2977b2dce. With `BooguImageAttention(has_qkv=False)` the double-stream module never creates its own to_q/to_k/to_v, so the `requires_grad=False` loop is gone entirely.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Removed in 05cc899ae.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Correct — removed in 2977b2dce. The `del self.img_instruct_attn.to_k/to_v/to_q` block is gone; `has_qkv=False` means those layers are never instantiated in the first place.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 2977b2dce. The block now calls `self.img_instruct_attn(...)` directly (both the modulated and non-modulated branches); `BooguImageAttention.forward` dispatches to `BooguImageDoubleStreamSelfAttnProcessor`. No more reaching into `.processor`.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Removed in 05cc899ae — together with the per-submodule ones above, all weight-init is gone and `__init__` follows the lazy/inference-only convention.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Thanks for your reminder. Done in c9a6b5d06. Removed the in-model TeaCache (the `enable_teacache` / `teacache_*` settings, the `TeaCacheParams` import, and the caching branch in `forward`). The single-stream forward is back to the plain block loop; we'll rely on `CacheMixin` / FirstBlockCache (and the WIP in #12652) for caching.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Not done yet — flagging for guidance. `img_patch_embed_and_refine` runs the noise/ref-image patch-embed + the refiner block loops (noise_refiner / ref_image_refiner), and it's called once from `forward`. Inlining it would make
all layer calls visible in `forward` per the coding-style doc, but `forward` is already fairly large, so I wanted to confirm the preference before moving it.

Do you want the whole helper inlined into `forward`, or is keeping it as a clearly-named single-call helper acceptable here? Happy to inline if that's the convention you want enforced.

src/diffusers/utils/teacache_util.py
Boogu-Team: Done in c9a6b5d06 — deleted `utils/teacache_util.py` entirely (it had no other users once the model + pipeline TeaCache paths were removed).

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: Done in 662e6ca11. Replaced the hand-written slope formula with the canonical `calculate_shift` (`# Copied from diffusers.pipelines.flux.pipeline_flux.calculate_shift`). For our config it produces the same mu (bit-identical), and base stays bit-exact on GPU.

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: I looked into this but held off, because Boogu's timestep convention differs from the standard FlowMatch path and `retrieve_timesteps(..., sigmas=sigmas)` would change the numerics:

- Boogu feeds the 0->1 sigma value *directly as the timestep* to the transformer (not sigma * num_train_timesteps).
- `FlowMatchEulerDiscreteScheduler.set_timesteps(sigmas=...)` instead forces `timesteps = sigmas * num_train_timesteps` and applies its own shift in step 2. That only reduces to our convention when shift=1 and use_dynamic_shifting=False, and even then the reported `timesteps` differ by  the 1000x scale that the transformer is sensitive to.

So switching to `retrieve_timesteps(sigmas=)` as-is would break bit-exactness on the base path. The current code builds the custom 0->1 sigmas and sets `scheduler.timesteps` / `scheduler.sigmas` / `_step_index` so the transformer receives exactly the sigma-as-timestep values it was trained on.

If you're okay with it, I'd prefer to keep the explicit assignment for now (it's the minimal, verified-bit-exact form). Alternatively, if you want this routed through `retrieve_timesteps`, I think the clean fix is a small scheduler option (e.g. a flag to skip the *1000 scaling / shift) so the public API is used without changing outputs — happy to do that in a follow-up if you prefer.

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: Done in 3b56e275c — the pipeline now exposes only standard classifier-free guidance. Rather than carry an unused code path, we removed the boosted-orthogonal-guidance (BOG) implementation outright for this round:
`MomentumRollingSum`, the `_project_matrix` / `_newtonschulz5_batched` / `bog_norm` / `calculate_boosted_orthogonal_guidance` helpers, and all the `use_boosted_orthogonal_guidance` / `*_momentum_*` / `bog_*` arguments from `__call__` and `processing`. BOG defaulted to off, so every guidance branch already fell through to the plain `model_pred - model_pred_uncond` delta — removing it leaves the default CFG numerics unchanged (base bit-exact, maxdiff=0).

We'd like to re-introduce BOG as its own guider class under `src/diffusers/guiders/` (and wire advanced features through modular diffusers) in a follow-up PR, so this one stays focused on the standard pipeline. Does that
sequencing work for you?

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: Done in c9a6b5d06. Removed the TeaCache plumbing from the denoising loop (the 8 teacache blocks) and the import. The pipeline no longer references TeaCache.

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1777 @@
...
+        )
+
+
+# --------------- Norm / FeedForward / Embedding ----------------
+def _torch_swiglu(x, y):
+    return F.silu(x.float(), inplace=False).to(x.dtype) * y
+
+
+swiglu = _torch_swiglu
+torch_swiglu = _torch_swiglu
```
yiyixuxu:
```suggestion
```

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1777 @@
...
+            freqs_cis,
+            l_effective_cap_len,
+            seq_lengths,
+            combined_img_freqs_cis,
+            combined_img_seq_lengths,
+        )
+
+
+# --------------- Norm / FeedForward / Embedding ----------------
+def _torch_swiglu(x, y):
```
yiyixuxu:
```suggestion
def swiglu(x, y):
```

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1777 @@
...
+        )
+        self.linear_3 = nn.Linear(
+            dim,
+            inner_dim,
+            bias=False,
+        )
+
+    def forward(self, x):
+        h1, h2 = self.linear_1(x), self.linear_3(x)
+        swiglu_fn = torch_swiglu if torch.compiler.is_compiling() else self.swiglu
```
yiyixuxu:
```suggestion
```
they're identical,no?

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1777 @@
...
+        return self.linear_2(swiglu_fn(h1, h2))
```
yiyixuxu:
```suggestion
        return self.linear_2(swiglu(h1, h2))
```

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1777 @@
...
+
+    def __init__(
+        self,
+        dim: int,
+        inner_dim: int,
+        multiple_of: Optional[int] = 256,
+        ffn_dim_multiplier: Optional[float] = None,
+    ):
+        super().__init__()
+        self.swiglu = swiglu
```
yiyixuxu:
```suggestion
```

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1777 @@
...
+            shift = 0
+            for ref_img_len in l_effective_ref_img_len[i]:
+                batch_ref_img_mask[idx, :ref_img_len] = True
+                batch_ref_image_hidden_states[idx, :ref_img_len] = ref_image_hidden_states[
+                    i, shift : shift + ref_img_len
+                ]
+                batch_ref_img_rotary_emb[idx, :ref_img_len] = ref_img_rotary_emb[i, shift : shift + ref_img_len]
+                batch_temb[idx] = temb[i]
+                shift += ref_img_len
+                idx += 1
```
yiyixuxu, line 1393: you can put these into a helper method that;'s fine

src/diffusers/models/transformers/transformer_boogu.py
```diff
@@ -0,0 +1,1777 @@
...
+                ref_image_hidden_states[i, shift : shift + ref_img_len] = batch_ref_image_hidden_states[
+                    idx, :ref_img_len
+                ]
+                shift += ref_img_len
+                idx += 1
+
+        combined_img_hidden_states = hidden_states.new_zeros(batch_size, max_combined_img_len, self.config.hidden_size)
+        for i, (ref_img_len, img_len) in enumerate(zip(l_effective_ref_img_len, l_effective_img_len)):
+            combined_img_hidden_states[i, : sum(ref_img_len)] = ref_image_hidden_states[i, : sum(ref_img_len)]
+            combined_img_hidden_states[i, sum(ref_img_len) : sum(ref_img_len) + img_len] = hidden_states[i, :img_len]
```
yiyixuxu, line 1424: same here too

src/diffusers/models/transformers/transformer_boogu.py
yiyixuxu: yes, let's inline it. the convention is that we want to have every `nn.Module`/ weight-layer call directly visible in `foward`, so basically these

```py
  ....
ref_image_hidden_states = self.ref_image_patch_embedder(ref_image_hidden_states)
 ...
for layer in self.noise_refiner:
    hidden_states = layer(hidden_states, padded_img_mask, noise_rotary_emb, temb)
 ....
```

you can still use helper method to abstract away the rest of the code inside `img_patch_embed_and_refine` if it makes sense

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1467 @@
...
+    XLA_AVAILABLE = True
+else:
+    XLA_AVAILABLE = False
+
+
+logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
+
+
+@dataclass
+class FMPipelineOutput(BaseOutput):
```
yiyixuxu: can you move it to boogu/pipeline_output.py?
https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/flux2/pipeline_output.py#L10

src/diffusers/pipelines/boogu/pipeline_boogu.py
yiyixuxu: would it work if you just feed `t/ num_train_timesteps` to the transformer model?

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1467 @@
...
+                    ref_latents.append(self.encode_vae(img_j.to(device=device)).squeeze(0))
+            else:
+                ref_latents = None
+
+            for _ in range(num_images_per_instruction):
+                latents.append(ref_latents)
+
+        return latents
+
+    def _check_and_wrap_input_images(
```
yiyixuxu, line 386: we do not accept path for `image`

also, this function is really complex. At diffusers, we try not to guess user intent and "normalize" it. Basically, we just need to make it clear what we expect as the `image` input and raise an error message otherwise; 

I think we can just accept None, a list of images (one sample), or list of list of images (bartched), we wrap the single image into a batch and raise on everything else, you might be able to use this for check input https://github.com/huggingface/diffusers/blob/main/src/diffusers/image_processor.py#L54

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1467 @@
...
+                vlm_inputs[k] = vlm_inputs[k].to(device)
+
+        instruction_mask = vlm_inputs["attention_mask"]
+
+        num_instruction_feature_layers = self.transformer.instruction_feature_configs.get(
+            "num_instruction_feature_layers", 1
+        )
+        final_instruction_mask = instruction_mask
+
+        with torch.no_grad():
```
yiyixuxu:
```suggestion
```

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1467 @@
...
+        latents = latents.to(dtype=dtype)
+        if self.vae.config.scaling_factor is not None:
+            latents = latents / self.vae.config.scaling_factor
+        if self.vae.config.shift_factor is not None:
+            latents = latents + self.vae.config.shift_factor
+        image = self.vae.decode(latents, return_dict=False)[0]
+
+        return image
+
+    def predict(
```
yiyixuxu, line 1308: can we inline this?

src/diffusers/models/transformers/transformer_boogu.py
yiyixuxu: ohh we usually do not change the checkpoint once it's merged in because it will break for diffusers users who're using it 
but I understand the concern on re-key the published weights, I think the best solution is to refactor here & release boogu-image directly in Modular Diffusers so that you don't have the rekey and the same repo can work for both diffusers & original repo

with modular, you can just add a second transformer repo that works for diffusers alongside the current one (e.g. `transformer_diffusers`) and in `modular_model_index.json` map the transformer components to that folder, see an example of modular repo https://huggingface.co/circlestone-labs/Anima-Base-v1.0-Diffusers/blob/main/modular_model_index.json 

other than the custom `guider`, in `predict,` the denoise loop handles guidance for 4 different cases, that's exact the type of use case that should be supported in modular system

src/diffusers/pipelines/boogu/pipeline_boogu.py
```diff
@@ -0,0 +1,1467 @@
...
+        if not input_images:
+            return "t2i"
+
+        if isinstance(input_images, (list, tuple)):
+            for x in input_images:
+                if x:
+                    return "ti2i"
+        return "t2i"
+
+    def processing(
```
yiyixuxu, line 1120: This method should be in-line; 
Also, it actually handles 4 different conditions; I think we should support it from modular instead, where you can write 4 different denoising loop blocks and dispatch at runtime based on user input; see quick doc https://huggingface.co/docs/diffusers/main/en/modular_diffusers/quickstart

It's pretty straightforward to migrate to a modular system, and AI agents have been pretty good at this task in our experience

yiyixuxu (commented):
thanks @Boogu-Team !
i left a bit more comments

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 9918a630d. Collapsed the SwiGLU helper to a single `def swiglu(x, y)` — the `_torch_swiglu` name and the `swiglu` / `torch_swiglu` aliases were redundant.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 9918a630d. Dropped both module-level aliases; there's now a single `def swiglu(x, y)`.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 9918a630d. Removed the `torch_swiglu if is_compiling() else self.swiglu` branch — both sides were the same function.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 9918a630d. `forward` is now `return self.linear_2(swiglu(h1, h2))`.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 9918a630d. The FFN no longer stores `self.swiglu`; it calls the module-level `swiglu` directly.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Kept as a helper (`_flatten_ref_images`) in 96fd58bbd — the weight-layer calls are now in `forward`, and this index-shuffling is abstracted away as you suggested.

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: Done in 9918a630d. Moved the output dataclass to `boogu/pipeline_output.py` (flux2 layout) and renamed `FMPipelineOutput` -> `BooguImagePipelineOutput`; both pipelines import it from there, it's exported via the package `__init__`, and the docs autodoc reference is updated.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Same — kept as `_combine_ref_and_noise_img` in 96fd58bbd; the layer calls live in `forward`, the index-shuffling is in the helper.

src/diffusers/models/transformers/transformer_boogu.py
Boogu-Team: Done in 96fd58bbd. Inlined the layer calls into `forward`: the x_embedder / ref_image_patch_embedder patch embeddings and the noise_refiner / ref_image_refiner block loops are now visible at the call site. The pure index-shuffling around them stays in three helpers (`_add_image_index_embedding`, `_flatten_ref_images`, `_combine_ref_and_noise_img`), per your note that helpers are fine for that. Also dropped the now-fully-dead `padded_ref_img_mask` (computed and threaded through but never read).

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: Done in b5fe1de08. Replaced the ~170-line intent-guessing normalizer with a strict validator: `None` / `[]` -> no reference images; `List[PIL.Image.Image]` -> one sample; `List[List[PIL.Image.Image]]` -> already batched (per-sample `None` / `[]` = no image); anything else (paths, a bare image, mixed / nested types) raises `ValueError`. The pipeline no longer loads images from disk or guesses intent.

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: Agreed this is the right shape for the modular system — the 4 guidance cases are the same "drop-condition -> delta -> weighted-combine" logic parameterized differently, which fits a guider-abstracted denoise block rather than inline branches. Deferring the inline + modularization to the modular migration; see the scope questions on the thread above (r3502711018) so we build it the way you want.

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: This is in the shared `predict` / denoise-loop path that the modular migration will restructure (per r3502737207). Rather than inline it now and rework it for modular, we'd like to fold it into that migration once the scope above is settled — happy to do it either way depending on your call.

Boogu-Team:
> thanks @Boogu-Team ! i left a bit more comments

https://github.com/huggingface/diffusers/pull/14040#discussion_r3502711018

Thanks — that makes sense, and agreed we shouldn't change an already-merged checkpoint. We're happy to go the Modular Diffusers route (a `transformer_diffusers` folder + `modular_model_index.json` so one repo serves both diffusers and the original code, avoiding a re-key), and to move the denoise-loop guidance into the modular system.

Before we start, could you confirm the intended scope so we build the right thing:

1. Scope — should the standard `BooguImagePipeline` / `BooguImageTurboPipeline` stay alongside a new `modular_pipelines/boogu/`, or is modular meant to be the primary integration going forward?
2. Sequencing — do you want the modular migration in THIS PR, or land the standard pipeline first and add modular in a follow-up PR?
3. Guidance — Boogu's denoise loop has 4 guidance cases (t2i CFG; ti2i text-only; ti2i image-only; ti2i double text+image guidance, plus an optional empty-instruction term). Should these compose from existing guiders, or would you prefer a dedicated Boogu guider class under `src/diffusers/guiders/` (which is also where we'd re-introduce BOG)?

Once we align on the above we'll map it onto the qwenimage/flux2 modular layout (encoders / before_denoise / denoise / decoders + block assembly). The model, image processor, and scheduler usage are all reusable as-is.

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: Done in b8be8067e. `__call__` is already decorated with `@torch.no_grad()`, so the inner `with torch.no_grad():` around the MLLM forward here was  redundant — removed the wrapper and dedented its body (propagated to the turbo pipeline via `make fix-copies`, whose `__call__` is also `@torch.no_grad()`).

src/diffusers/pipelines/boogu/pipeline_boogu.py
Boogu-Team: Done in 61bb0100e — thanks, this is the clean fix. `set_flow_match_timesteps` now installs the 0->1 sigma schedule through the public  `scheduler.set_timesteps(sigmas=...)` API instead of hand-assigning timesteps/sigmas/_step_index; the only fix-up left is the terminal sigma (set_timesteps appends 0.0, Boogu's inverted 0->1 schedule ends at 1.0). The timesteps now run 0..num_train_timesteps, and the denoise loop feeds `t / num_train_timesteps` to the transformer. Verified equivalent: at bf16/fp16 the transformer feed is bit-identical (8/28/50 steps), the step loop's final latent matches to maxdiff=0, and GPU base T2I+CFG is bit-exact vs before.

vladmandic:
any updates on this pr? looks like it stalled few weeks ago?

Boogu-Team:
> any updates on this pr? looks like it stalled few weeks ago?

Thanks for your reminder. We are awaiting confirmation of the expected scope.

DefTruth:
also pay attention to this pr

Boogu-Team:
We are awaiting your comments before proceeding to the next step.

Issue #14003, opened 2026-06-20
Split Q,K,V vs Fused QKV Projections in diffusers Discussion


dg845:
`diffusers` generally splits fused QKV projections to separate Q, K, V projections in its checkpoint format (for example, for diffusion transformer models). I am interested in collecting feedback on this design choice and any problems which may arise from it. To be honest upfront, I think it is unlikely that we will change this design in the near future, but we would appreciate your feedback nevertheless.

EDIT: the reasoning for using split Q, K, V projections is described in more detail in https://github.com/huggingface/diffusers/issues/14003#issuecomment-4774944332.

dg845:
@bghira, with respect to https://github.com/huggingface/diffusers/pull/13861#issuecomment-4757018693, would something like having an `inplace` flag on `AttentionMixin.fuse_qkv_projections`, where `inplace=True` fuses the weights and then deletes the old ones, mitigate your training concerns? I think doing the fusion in-place should make the fused QKV projection more friendly for training.

dxqb:
a good starting point for this discussion would be to understand, why diffusers is doing this. I don't understand it.

dg845:
My understanding is that the main reason is code readability and consistency across the codebase. Having a single convention makes it easier to understand the attention code across different models. I believe the reason the split convention is used in particular is partly historical (older models like Stable Diffusion (pre-SD3) used the split convention natively) and partly because it is more expressive modeling-wise (for example, fused projections have a harder time expressing per-projection techniques like IP-Adapter).

PR #14047, opened 2026-06-22
Refactor QKV Fusion Utilities

dg845:
# What does this PR do?

This PR refactors the attention QKV fusion utilities in `AttentionMixin` and `AttentionModuleMixin` to be more LoRA-aware. In particular, this PR adds guards when attempting to fuse/unfuse with a LoRA attached (because LoRAs cannot be easily transferred over when fusing/unfusing) and an `inplace` option to fuse without keeping copies of the split Q,K,V projections.

**Changelist**
1. Adds guards when attempting to fuse/unfuse Q,K,V with a LoRA attached (will raise an error).
2. Adds an `inplace` argument; if `inplace=True`, the module is modified to have only the fused QKV projection (e.g. `to_qkv`) with the split Q,K,V projections (e.g. `to_q`/`to_k`/`to_v`) removed. (`inplace=False`, the default, retains the current behavior).
3. Supports fusion in the case where `add_k_proj` and `add_v_proj` are present without `add_q_proj` also present (e.g. Wan models).
4. Adds `get_qkv` and `get_added_qkv` helper methods in `AttentionModuleMixin` which handles getting the Q, K, V (and added Q,K,V, for second stream projections in MM-DiT-style models like Flux) in both the fused and split case. This is intended to make it easier for attention processors to support both fused and split QKV.
5. Adds an experimental `restore_checkpoint_fusion_state` method to `AttentionMixin` to put models back in the fusion state of the original model checkpoint. A new `_native_fused_projections` attribute on `AttentionModuleMixin` is added to allow this state to be described. (The motivation is to make it easier to support PEFT adapters which target the original checkpoint structure.)

Partially addresses #14003.

## Before submitting

## Who can review?

@sayakpaul
@DN6


src/diffusers/models/attention.py
```diff
@@ -275,6 +315,16 @@ def fuse_projections(self):
...
+
+            if inplace:
+                # Keep the necessary K,V dims so that the individual projections can be reconstructed.
+                self._qkv_split_dims = (
+                    self.to_k.weight.shape[0],
+                    self.to_v.weight.shape[0],
+                    self.to_k.weight.shape[1],
+                )
+                delattr(self, "to_k")
+                delattr(self, "to_v")
```
sayakpaul, line 327: Does it free the memory? If not, what is the purpose of deleting these attributes? Also, from what I understand this and some of the other refactors introduced in this PR aren't particularly for LoRA-awareness?

src/diffusers/models/attention.py
```diff
@@ -287,27 +337,68 @@ def fuse_projections(self):
...
+                # Keep the necessary Q,K,V dims so that the individual projections can be reconstructed.
+                self._qkv_split_dims = (
+                    self.to_q.weight.shape[0],
+                    self.to_k.weight.shape[0],
+                    self.to_v.weight.shape[0],
+                    self.to_q.weight.shape[1],
+                )
+                delattr(self, "to_q")
+                delattr(self, "to_k")
+                delattr(self, "to_v")
```
sayakpaul, line 350: Same as above.

src/diffusers/models/attention.py
```diff
@@ -244,11 +261,34 @@ def set_use_memory_efficient_attention_xformers(
 
                 self.set_attention_backend("xformers")
 
+    @staticmethod
+    def _has_active_lora(module: nn.Module) -> bool:
+        """Checks for the presence of PEFT-style LoRA modules without needing to import `peft`."""
+        return any("lora_A" in name or "lora_B" in name for name, _ in module.named_modules())
```
sayakpaul, line 267: I think a better way to detect it is:

```py
for name, mod in module.named_modules():
    if isinstance(mod, BaseTunerLayer): ...
```

This way, we can cater towards any non-LoRA adapters in the future.

src/diffusers/models/attention.py
```diff
@@ -117,11 +117,28 @@ def unfuse_qkv_projections(self):
             if isinstance(module, AttentionModuleMixin) and module._supports_qkv_fusion:
                 module.unfuse_projections()
 
+    def restore_checkpoint_fusion_state(self, inplace: bool = False):
```
sayakpaul, line 120: Could you provide an example of where / how this is used?

My mental model says:

* Load the pipeline
* Call the `fuse_qkv_projections()` method on `pipe.transformer`.
* Then, before loading we call this method.
* And then load the LoRA weights?

tests/models/test_attention_mixins.py
```diff
@@ -0,0 +1,624 @@
+import pytest
```
sayakpaul, line 1: Would it not make sense to make these a common mixin that can be added to models?

tests/models/test_attention_mixins.py
```diff
@@ -0,0 +1,624 @@
...
+        super().__init__()
+        self.base = linear
+        self.lora_A = nn.Linear(linear.in_features, rank, bias=False)
+        self.lora_B = nn.Linear(linear.out_features, rank, bias=False)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.base(x) + self.lora_B(self.lora_A(x))
+
+
+class TestAttentionModuleMixin:
```
sayakpaul, line 102: The tests are missing:

* If outputs with and without fusion stay the same (they should).
* If expected errors are raised when LoRA is attached.

Additionally, are we checking if the projection attributes like `to_k`, `to_q`, etc. are actually getting deleted for in-place fusions and are successfully restored when the user wants them?

tests/models/test_attention_mixins.py
```diff
@@ -0,0 +1,624 @@
...
+    def test_unfuse_qkv_projections_unfuses_all(self, model):
+        model.fuse_qkv_projections()
+        model.unfuse_qkv_projections()
+        assert model.block1.fused_projections is False
+        assert model.block2.fused_projections is False
+        assert not hasattr(model.block1, "to_qkv")
+        assert not hasattr(model.block2, "to_kv")
+
+    # -------------------------------------------------------------------------
+    # restore_checkpoint_fusion_state
```
sayakpaul, line 595: Should the corresponding tests also check if we have the fused QKV projections rather than just checking the `fused_projections` attribute?

tests/models/test_attention_mixins.py
```diff
@@ -0,0 +1,624 @@
...
+    # -------------------------------------------------------------------------
+
+    def test_restore_checkpoint_noop_for_none(self, model):
+        # Default _native_fused_projections is None — state should be unchanged.
+        model.restore_checkpoint_fusion_state()
+        assert model.block1.fused_projections is False
```
sayakpaul, line 601: Prefer 


```suggestion
        assert not model.block1.fused_projections
```

src/diffusers/models/attention.py
dg845: Yeah, the `delattr` lines free memory relative to the non-`inplace` implementation by allowing the split projection weights to be freed. This can reduce the memory usage a fair bit for a full-size DiT (for example, ~1.7 GB for a 5.5B Flux transformer).

src/diffusers/models/attention.py
dg845: I have removed `restore_checkpoint_fusion_state` from this PR to keep the scope focused. (The idea was to allow checkpoints to be set to the original checkpoint's QKV state, but it needs some work.)

PR #14058, opened 2026-06-24
fix claude code review fix in PRs.

sayakpaul:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/actions/runs/28076024086/job/83120352829

Since we're still beta-testing Serge and there can be disruptions inside it, I think it's better to also have the Claude reviewer in a fixed state.


yiyixuxu (approved):
thanks

PR #14059, opened 2026-06-24
[tests] implement base model output caching in model-level tests

sayakpaul:
Implements fixture-level caching of base model output to avoid recomputation:

* `common.py`
* `memory.py`

Other mixins don't have the right contract to implement it yet.


sayakpaul:
/diffusers-bot pytest tests/models/transformers/test_models_transformer_flux.py -k "offload"

tests/models/testing_utils/common.py
```diff
@@ -258,7 +262,39 @@ def get_dummy_inputs(self) -> Dict[str, Any]:
...
+        speeds up the suite.
+
+        The hardware-gated tests that consume this fixture use `pytest.mark.skipif` (via the `require_*`
+        decorators), which pytest evaluates before fixture setup, so skipping on a machine without the
+        required accelerators never triggers this forward.
+
+        Tests that still need a live model (e.g. to save or offload it) build their own with the same
+        seed, so the reloaded model's weights match this cached output.
+        """
+        torch.manual_seed(0)
```
dg845, line 291: Should `BaseModelOutputMixin` also expose the seed it uses to create the `base_model_output` fixture? I think this would allow tests to more easily match its behavior.

dg845 (approved):
Thanks for the PR! Left one comment.

tests/models/testing_utils/common.py
sayakpaul: I thought about it but then this change should also be propagated to the `generator` property, e.g. 

https://github.com/huggingface/diffusers/blob/d8d3f906e0f782bb8a7eb90f8b78517c19bf37e5/tests/models/transformers/test_models_transformer_qwenimage.py#L64-L66

So, I think this should be done in a separate PR.

PR #14071, opened 2026-06-25
[mps] Fix float64 upcast in HeliosDMDScheduler.convert_flow_pred_to_x0

Shreyas-jk:
## What does this PR do?

`HeliosDMDScheduler.convert_flow_pred_to_x0` upcasts its tensors to float64 via
`.double()` for higher-precision math. The MPS backend has no float64, so this
raises `Cannot convert a MPS Tensor to float64 dtype as the MPS framework
doesn't support float64` on Apple Silicon the moment `step()` calls it.

The fix mirrors the existing `# mps does not support float64` pattern already
used across the schedulers (e.g. `scheduling_flow_match_euler_discrete.py`):
compute in float32 on `mps`, keep float64 on CUDA/CPU. Off-MPS numerics are
unchanged. Same fix shape as #13701.

## Testing

Verified locally on Apple Silicon (MPS available):

- New regression test in `tests/schedulers/test_scheduler_helios_dmd.py` fails
  on the unpatched method with the float64 `TypeError`, and passes with the fix
  (red → green).
- `make style && make quality` clean.
- `pytest tests/schedulers/ -k "flow or helios"` → 16 passed.

The test is gated with `@unittest.skipUnless(torch_device == "mps", ...)`, so it
skips on the Linux/CUDA CI runners and runs for contributors on Apple Silicon.

cc @yiyixuxu @sayakpaul

### DIFF
--- a/src/diffusers/schedulers/scheduling_helios_dmd.py
+++ b/src/diffusers/schedulers/scheduling_helios_dmd.py
@@ -275,7 +275,11 @@ class HeliosDMDScheduler(SchedulerMixin, ConfigMixin):
         # use higher precision for calculations
         original_dtype = flow_pred.dtype
         device = flow_pred.device
-        flow_pred, xt, sigmas, timesteps = (x.double().to(device) for x in (flow_pred, xt, sigmas, timesteps))
+        # mps does not support float64
+        compute_dtype = torch.float32 if device.type == "mps" else torch.float64
+        flow_pred, xt, sigmas, timesteps = (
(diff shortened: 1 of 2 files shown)


Shreyas-jk:
Superseded by #14375, which fixes the same `convert_flow_pred_to_x0` upcast plus both `set_timesteps` sites, and routes through `maybe_adjust_dtype_for_device` so it covers `npu` and `neuron` rather than just `mps`. Closing in favour of that one — details in [my comment there](https://github.com/huggingface/diffusers/pull/14375#issuecomment-5284939806).

PR #14073, opened 2026-06-26
[Tests] Skip layerwise casting tests on devices without float8_e4m3fn support

### LINKED ISSUE #14072: [Bug] Layerwise casting tests fail on MPS backend due to float8 unsupported dtype
### Describe the bug

When running the unit test suite on Apple Silicon (macOS) using the MPS backend (`torch_device = "mps"`), tests checking `enable_layerwise_casting` fail. This occurs because the MPS backend does not support the `torch.float8_e4m3fn` dtype.

Several test cases unconditionally configure layerwise casting with `storage_dtype=torch.float8_e4m3fn`, leading to a `TypeError: Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.` on MPS.

I intend to submit a pull request to fix this by dynamically skipping the tests when the active device doesn't support the `float8_e4m3fn` dtype.

### Reproduction

On an Apple Silicon Mac:
```bash
python3 -m pytest tests/pipelines/krea2/test_krea2.py -k test_layerwise_casting_inference
```

### Logs

```tb
___________ Krea2PipelineFastTests.test_layerwise_casting_inference ____________

self = <tests.pipelines.krea2.test_krea2.Krea2PipelineFastTests testMethod=test_layerwise_casting_inference>

    def test_layerwise_casting_inference(self):
        if not self.test_layerwise_casting:
            return
    
        pipe.to(torch_device, dtype=torch.bfloat16)
    
        denoiser = pipe.transformer if hasattr(pipe, "transformer") else pipe.unet
>       denoiser.enable_layerwise_casting(storage_dtype=torch.float8_e4m3fn, compute_dtype=torch.bfloat16)

tests/pipelines/test_pipelines_common.py:2305: 
...
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/torch/nn/modules/module.py:1370: TypeError
E           TypeError: Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.
```

### System Info

- Diffusers version: 0.39.0.dev0
- Platform: macOS (Apple Silicon)
- Python version: 3.11.9
- PyTorch version: 2.11.0
- MPS available: True

### Who can help?

@DN6 @sayakpaul @yiyixuxu

GiGiKoneti:
# What does this PR do?

`torch.float8_e4m3fn` is a storage-only dtype that is not supported on all hardware backends. On Apple Silicon (MPS), attempting to cast a tensor to this dtype raises a `TypeError`:

```
TypeError: Cannot convert a MPS Tensor to float8_e4m3fn dtype as the MPS framework doesn't support float8 types.
```

This causes five layerwise-casting tests to **crash** instead of being skipped:

| File | Test |
|------|------|
| `tests/pipelines/test_pipelines_common.py` | `test_layerwise_casting_inference` |
| `tests/models/testing_utils/memory.py` | `test_layerwise_casting_memory` |
| `tests/models/testing_utils/memory.py` | `test_layerwise_casting_training` |
| `tests/lora/utils.py` | `test_layerwise_casting_inference_denoiser` |
| `tests/lora/utils.py` | `test_layerwise_casting_peft_input_autocast_denoiser` |

### Fix

Each affected test now includes a runtime check at the very top:

```python
try:
    torch.zeros(1, device=torch_device).to(dtype=torch.float8_e4m3fn)
except TypeError:
    self.skipTest(f"Device {torch_device} does not support float8 storage dtype.")
```

If the device cannot handle `float8_e4m3fn`, the test is **skipped** with a clear message rather than crashing the entire test suite.

Fixes #14072

## Before submitting
- [ ] Did you read our [philosophy doc](https://huggingface.co/docs/diffusers/main/en/conceptual/philosophy)? (important for complex PRs)
  - Issue: https://github.com/huggingface/diffusers/issues/14072
- [ ] Did you make sure to update the documentation with your changes? *(N/A — no docs changes needed)*
- [ ] Did you write any new necessary tests? *(N/A — this fixes existing tests)*
- [ ] Are you the author (or part of the team) of the model/pipeline? *(N/A)*

## Who can review?

@pcuenca (MPS), @sayakpaul @DN6 (General functionalities)

### DIFF
--- a/tests/lora/utils.py
+++ b/tests/lora/utils.py
@@ -2105,12 +2105,12 @@ class PeftLoraLoaderMixinTests:
         self.assertTrue(not np.allclose(original_output, lora_output_diff_alpha, atol=1e-3, rtol=1e-3))
         self.assertTrue(not np.allclose(lora_output_diff_alpha, lora_output_same_rank, atol=1e-3, rtol=1e-3))
 
+    @pytest.mark.xfail(
+        condition=torch_device == "mps",
+        reason="MPS does not support float8 casting.",
+        strict=True,
+    )
     def test_layerwise_casting_inference_denoiser(self):
-        try:
-            torch.zeros(1, device=torch_device).to(dtype=torch.float8_e4m3fn)
-        except TypeError:
-            self.skipTest(f"Device {torch_device} does not support float8 storage dtype.")
-
         from diffusers.hooks._common import _GO_LC_SUPPORTED_PYTORCH_LAYERS
         from diffusers.hooks.layerwise_casting import DEFAULT_SKIP_MODULES_PATTERN
 
@@ -2154,6 +2154,11 @@ class PeftLoraLoaderMixinTests:
         pipe_float8_e4m3_bf16 = initialize_pipeline(storage_dtype=torch.float8_e4m3fn, compute_dtype=torch.bfloat16)
         pipe_float8_e4m3_bf16(**inputs, generator=torch.manual_seed(0))[0]
 
+    @pytest.mark.xfail(
+        condition=torch_device == "mps",
+        reason="MPS does not support float8 casting.",
+        strict=True,
+    )
     @require_peft_version_greater("0.14.0")
     def test_layerwise_casting_peft_input_autocast_denoiser(self):
         r"""
@@ -2169,10 +2174,6 @@ class PeftLoraLoaderMixinTests:
 
         See the docstring of [`hooks.layerwise_casting.PeftInputAutocastDisableHook`] for more details.
         """
-        try:
-            torch.zeros(1, device=torch_device).to(dtype=torch.float8_e4m3fn)
-        except TypeError:
-            self.skipTest(f"Device {torch_device} does not support float8 storage dtype.")
 
         from diffusers.hooks._common import _GO_LC_SUPPORTED_PYTORCH_LAYERS
         from diffusers.hooks.layerwise_casting import (
--- a/tests/lora/utils.py
+++ b/tests/lora/utils.py
@@ -2106,6 +2106,11 @@ class PeftLoraLoaderMixinTests:
         self.assertTrue(not np.allclose(lora_output_diff_alpha, lora_output_same_rank, atol=1e-3, rtol=1e-3))
 
     def test_layerwise_casting_inference_denoiser(self):
+        try:
+            torch.zeros(1, device=torch_device).to(dtype=torch.float8_e4m3fn)
+        except TypeError:
+            self.skipTest(f"Device {torch_device} does not support float8 storage dtype.")
+
         from diffusers.hooks._common import _GO_LC_SUPPORTED_PYTORCH_LAYERS
         from diffusers.hooks.layerwise_casting import DEFAULT_SKIP_MODULES_PATTERN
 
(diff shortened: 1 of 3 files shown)


GiGiKoneti:
## Alternative approaches considered

This PR uses an inline `try/except` guard at the top of each affected test. It works, but I wanted to flag two cleaner alternatives for your consideration:

### 1. `@require_float8_support` decorator (preferred)

Instead of repeating the same 4-line block across 5 tests, a reusable decorator in the test utilities would be more consistent with the existing `@require_*` pattern in diffusers:

```python
def require_float8_support(test_func):
    @functools.wraps(test_func)
    def wrapper(*args, **kwargs):
        try:
            torch.zeros(1, device=torch_device).to(dtype=torch.float8_e4m3fn)
        except TypeError:
            raise unittest.SkipTest(f"Device {torch_device} does not support float8_e4m3fn")
        return test_func(*args, **kwargs)
    return wrapper
```

Each test then becomes:
```python
@require_float8_support
def test_layerwise_casting_inference(self):
    ...
```

### 2. Guard in source code (`enable_layerwise_casting`)

A more complete fix would be to also add a check inside `enable_layerwise_casting()` itself, raising a clear `ValueError` when the device does not support the requested `storage_dtype`, rather than letting it crash deep inside the pipeline forward pass. The test-side skip would still be needed, but users calling the API directly would get a much better error message.

---

let me know which you prefer and I will update the PR. cc @pcuenca @DN6

sayakpaul (commented):
Would it not be better to use a `skip_mps` decorator directly to skip those tests? 

https://github.com/huggingface/diffusers/blob/bd2c91958881b777260eedb1c3d61d01c03e800f/src/diffusers/utils/testing_utils.py#L468

GiGiKoneti:
Using the `@skip_mps` decorator is much cleaner and fits the existing test codebase conventions perfectly.
The only reason I initially went with the dynamic `try/except` check was to make it future-proof and device-agnostic (in case other accelerators/configurations also lack float8 casting support and trigger a similar TypeError). But since MPS is the primary backend experiencing this crash and `@skip_mps` is idiomatic for the suite, using it is definitely better.

u want me to update PR ?

sayakpaul:
Future-proof would `xfail`ing those tests when `torch_device == "mps"`. That way, if support is added in the future, `xfail` won't work anymore.

GiGiKoneti:
@sayakpaul 
Since float8 support is a device capability (similar to bfloat16), would you prefer we `xfail` on MPS specifically to track when they add support, or use a generic `@require_float8` decorator to skip on any unsupported device?

GiGiKoneti:
@sayakpaul
I have updated the PR to use `@pytest.mark.xfail(condition=torch_device == "mps", reason="MPS does not support float8 casting.", strict=True)` for all five affected layerwise casting tests instead of skipping them. 

This way, if MPS adds float8 casting support in the future, the tests will unexpectedly pass (XPASS) and fail the suite, alerting us to remove the decorator.

sayakpaul (approved):

GiGiKoneti:
Hi @sayakpaul,

The PR checks show 2 failing workflows on the CI:
1. **Torch CUDA Tests (models)**:
   - `TestAutoencoderVidTokMemory::test_group_offloading` fails with a shape mismatch. This is because `AutoencoderVidTok`'s `forward` returns a raw tensor instead of a single-element tuple `(dec,)` when `return_dict=False` (violating the standard Diffusers VAE return contract).
   - `TestAutoencoderDCTraining::test_mixed_precision_training` fails with cuDNN engine search errors (`"GET was unable to find an engine..."`).
2. **PyTorch Example CPU tests**:
   - The unconditional image generation example test times out due to a blocking subprocess read under single-device CPU launch configurations.

I have tested and verified clean fixes for these locally (correcting `AutoencoderVidTok`'s return type, adding a `try-except` skip guard for the DC mixed-precision test, and adding a timeout/single-process constraint to the example run helper). 

Would you like me to push these fixes here to get the CI green for merge, or should we handle them in a separate PR?

PR #1234, opened 2022-11-09
Added sampling script for unconditional training example

lukovnikov:
Added sampling script that supports both DDPM and DDIM sampling for unconditional  training example

PR #14080, opened 2026-06-27
Complete Kohya LoRA conversion for Qwen and Z-Image

dxqb:
# What does this PR do?

This PR adds to the already implement kohya-to-diffusers conversion code some missing layers.
These are layers that mostly live outside the transformer blocks, but also one inside the transformer block.

I *guess* these weren't initially included because kohya-ss/musubi-tuner doesn't train them by default, but you *can* train them, with kohya-ss/musubi-tuner and other trainers that output the kohya format.

Details:

  **Qwen — top-level (non-block) modules**
  `convert_key` assumes every key lives under `transformer_blocks` and strips/re-prepends
  that prefix. The six top-level modules (`img_in`, `txt_in`, `proj_out`,
  `norm_out.linear`, `time_text_embed.timestep_embedder.linear_1/2`) collapse onto each
  other and trip the `state_dict should be empty` check. They're now resolved via an
  explicit flattened→dotted map before the block logic, preserving the
  `.lora_down`/`.lora_up`/`.alpha` suffix.

  **Z-Image — module names that contain underscores**
  The blanket `_`→`.` split over-splits modules whose own names contain underscores
  (`all_final_layer`, `all_x_embedder`, `adaLN_modulation`, `cap_embedder`, `t_embedder`),
  so they arrive as `all.final.layer`, `adaLN.modulation`, … and fail with "unexpected
  keys". The existing dotted→underscore post-normalization is extended to re-merge these
  names (it runs on the full key, so `.lora_A/B` and `.alpha` are handled alike).

## Who can review?

- PEFT: @sayakpaul @BenjaminBossan


dxqb:
> Hi @dxqb, thanks for the PR! It does not appear to link an issue it fixes. If this PR addresses an existing issue, please add a closing keyword (e.g. `Fixes #1234`) to the PR description so the issue is linked. See the [contribution guide](https://huggingface.co/docs/diffusers/main/en/conceptual/contribution#coding-with-ai-agents) for more details. If this PR intentionally does not fix a tracked issue, a maintainer can add the `no-issue-needed` label to silence this reminder.

I could open an issue, but it would just repeat what the PR summary already says. Please add `no-issue-needed`

sayakpaul (approved):
Sweet!

PR #14089, opened 2026-06-29
[lora training] add caption dropout and aspect ratio buckets to krea2 and qwen-image

linoytsaban:
Adds aspect-ratio bucketing and caption dropout to the krea2 and qwen-image dreambooth LoRA scripts (neither supported these before):
- **Aspect-ratio bucketing** for krea2 and qwen-image.
- **Caption dropout** (`--caption_dropout`): randomly replace an instance caption with `""` so that fraction trains unconditionally (improves CFG).

Split from the original PR; bucketing fixes to the other scripts that already support buckets are in #14158.


linoytsaban:
krea comparison with and without buckets and dropout
image

sayakpaul (approved):
Should we do it just one and then propagate to the others? This way, it's easier to review.

linoytsaban:
> Should we do it just one and then propagate to the others? This way, it's easier to review.

I can split the krea 2 + qwen image changes (that introduce aspect ratio buckets + caption dropout since they dont currently support these at all) from the changes to the other scripts (that already support bucketing) where it's fixing bucketing issues. wdyt?

sayakpaul:
Yes sure

sayakpaul (commented):
Same comments as the other PR. WDYT?

examples/dreambooth/train_dreambooth_lora_krea2.py
```diff
@@ -524,6 +527,26 @@ def parse_args(input_args=None):
...
+    )
+    parser.add_argument(
+        "--use_aspect_ratio_buckets",
+        action="store_true",
+        help=(
+            "Enable aspect-ratio bucketing. Without --aspect_ratio_buckets, the buckets are computed on the "
+            "fly from --resolution and capped to each image's own resolution, so smaller images are assigned "
+            "to a smaller bucket instead of being upscaled. Provide --aspect_ratio_buckets to use an explicit list."
+        ),
+    )
```
sayakpaul, line 549: If `--aspect_ratio_buckets` is provided and `--use_aspect_ratio_buckets` is not set, should we throw an error? We could also just keep `--aspect_ratio_buckets`, default it to `None`. If specified, then only use it for the conditionals. WDYT?

linoytsaban: Kept the error here to stay aligned with the behavior we landed on in #14158 ([discussion](https://github.com/huggingface/diffusers/pull/14158#discussion_r3619890919)) - keeping `--use_aspect_ratio_buckets` as the single explicit switch across all the dreambooth scripts, so a bucket list without it fails fast instead of silently enabling bucketing.

PR #14091, opened 2026-06-29
Fix model offloading and training tests + prevent examples timeout

### LINKED ISSUE #14090: Fix model offloading and training tests + prevent examples timeout
### Describe the bug

When running the full test suite on models and training examples, three distinct test failures/hangs consistently block test runners:

1. **AutoencoderVidTok return format mismatch:**
   `AutoencoderVidTok.forward` returns a raw tensor `dec` when `return_dict=False`. However, the standard VAE/autoencoder API contract in Diffusers requires returning a single-element tuple `(dec,)`. Because `return_dict=False` returned a raw tensor, the test fixture `base_model_output` indexed it with `[0]`, slicing the batch dimension out (converting shape `[4, 3, 16, 32, 32]` to `[3, 16, 32, 32]`). The training check run (`test_group_offloading`) returns a `DecoderOutput` where `[0]` gets `sample` (shape `[4, 3, 16, 32, 32]`), resulting in a shape mismatch failure.

2. **AutoencoderDC mixed-precision training crash:**
   `TestAutoencoderDCTraining::test_mixed_precision_training` crashes with `RuntimeError: "GET was unable to find an engine to execute this computation"` in specific CUDA/cuDNN runner environments due to missing matching execution engines for the DC architecture layers.

3. **Examples launcher hangs / timeouts:**
   In single-device/CPU CI workflows, `run_command` in `examples/test_examples_utils.py` uses `subprocess.check_output` without a timeout, causing execution to block indefinitely if a deadlock occurs. Additionally, the test runner does not enforce single-process launch, which can deadlock when training commands try to coordinate multi-device setups.

I have verified clean fixes for all of these issues locally and will submit a PR shortly.

### Reproduction

1. Run VidTok autoencoder memory tests:
   ```bash
   pytest tests/models/autoencoders/test_models_autoencoder_vidtok.py -k "test_group_offloading"
   ```
2. Run DC autoencoder training tests on a CUDA device with restricted cuDNN configurations.
3. Run unconditional image generation example tests on a single-GPU/CPU host:
   ```bash
   pytest examples/unconditional_image_generation/test_unconditional.py
   ```

### Logs

```shell
1. VidTok Offloading shape mismatch:
AssertionError: Shape mismatch — actual torch.Size([3, 16, 32, 32]) vs expected torch.Size([4, 3, ...])

2. AutoencoderDC mixed precision:
RuntimeError: "GET was unable to find an engine to execute this computation"

3. Examples hang:
pytest-timeout (>60s) while reading subprocess stdout
```

### System Info

- Platform: macOS / Linux (CI runner)
- PyTorch version: 2.x
- Diffusers version: 0.30.0.dev0

### Who can help?

@sayakpaul @DN6 @pcuenca

GiGiKoneti:
# What does this PR do?

This PR fixes three pre-existing bugs in model offloading, training tests, and example test runs that were causing CI failures:

1. **AutoencoderVidTok return format mismatch:**
   - Modified `AutoencoderVidTok.forward` to return `(dec,)` when `return_dict=False`, aligning it with the standard VAE return contract in Diffusers.
   - Unskipped `test_outputs_equivalence` since the outputs match correctly now.

2. **AutoencoderDC mixed precision skip:**
   - Added a `try-except` block to catch `RuntimeError: "GET was unable to find an engine to execute this computation"` and call `pytest.skip` when cuDNN cannot find matching computation engine configs.

3. **Examples timeout and distributed launch mitigation:**
   - Refactored `run_command` in `examples/test_examples_utils.py` to use `subprocess.run` with a configurable `timeout` parameter (default 300s) to prevent tests from hanging indefinitely.
   - Appended `--num_processes 1 --num_machines 1` to `ExamplesTestsAccelerate` launch arguments to prevent distributed launch deadlocks on single-device/CPU CI runners.

Fixes #14090

## Before submitting
  - Approved via issue: https://github.com/huggingface/diffusers/issues/14090
- [ ] Are you the author (or part of the team) of the model/pipeline?

## Who can review?

@sayakpaul @DN6 @pcuenca


examples/test_examples_utils.py
```diff
@@ -28,20 +28,31 @@ class SubprocessCallException(Exception):
     pass
 
 
-def run_command(command: List[str], return_stdout=False):
+def run_command(command: List[str], return_stdout=False, timeout=300):
```
sayakpaul: Timeout issues are common. We can leave them be.

tests/models/autoencoders/test_models_autoencoder_dc.py
```diff
@@ -91,6 +91,14 @@ def test_from_save_pretrained_dtype_inference(self, tmp_path, dtype):
 class TestAutoencoderDCTraining(AutoencoderDCTesterConfig, TrainingTesterMixin):
     """Training tests for AutoencoderDC."""
 
+    def test_mixed_precision_training(self):
+        try:
+            super().test_mixed_precision_training()
+        except RuntimeError as e:
+            if "GET was unable to find an engine" in str(e) or "unable to find an engine" in str(e).lower():
+                pytest.skip(f"Skipping mixed precision: backend engine missing: {e}")
+            raise
```
sayakpaul: It's better to gate / `xfail` it based on the availability of the hardware it's being tested. This is probably happening because our CI uses T4 and that doesn't support `bfloat16`. So, we could also implement this at this level: 
https://github.com/huggingface/diffusers/blob/1f9c2010835ca46dc995674143cfc40a3f331118/tests/models/testing_utils/training.py#L211

tests/models/autoencoders/test_models_autoencoder_vidtok.py
```diff
@@ -80,7 +79,6 @@ def get_dummy_inputs(self) -> dict:
 
 
 class TestAutoencoderVidTok(AutoencoderVidTokTesterConfig, ModelTesterMixin):
-    @pytest.mark.skip("VidTok output structure not compatible with recursive output check.")
```
sayakpaul, line 83: Why is this going away?

tests/models/autoencoders/test_models_autoencoder_dc.py
GiGiKoneti: Thanks! I've reverted the `ad-hoc` try/except in `test_models_autoencoder_dc.py` and implemented the hardware support check globally under the # Test with `bfloat16` block in `tests/models/testing_utils/training.py` using `torch.cuda.is_bf16_supported()`.

tests/models/autoencoders/test_models_autoencoder_vidtok.py
GiGiKoneti: Since we fixed `AutoencoderVidTok`'s return format to output a tuple (dec,) when `return_dict=False` (which aligns it with all other autoencoders), the `test_outputs_equivalence` comparison now passes successfully and no longer needs to be skipped.

tests/models/testing_utils/training.py
```diff
@@ -210,11 +210,14 @@ def test_mixed_precision_training(self):
 
         # Test with bfloat16
         if torch.device(torch_device).type != "cpu":
-            model.zero_grad()
-            with torch.amp.autocast(device_type=torch.device(torch_device).type, dtype=torch.bfloat16):
-                output = model(**inputs_dict, return_dict=False)[0]
+            if torch.device(torch_device).type == "cuda" and not torch.cuda.is_bf16_supported():
+                pass
```
sayakpaul: We should do it with `pytest.mark.skip` with this condition.

GiGiKoneti: Using `@pytest.mark.skipif` at the function level would skip the `float16` check on T4 GPUs as well (which passes fine). Using `pytest.skip()` inside the block allows us to only skip the `bfloat16` part while maintaining `float16` test coverage.

sayakpaul: Okay then let's use `pytest.skip()` for a proper skip.

sayakpaul (approved):

sayakpaul:
Failing tests are unrelated.

PR #14093, opened 2026-06-29
Invalidate HookRegistry child-registries cache on enable/disable cache

### LINKED ISSUE #14037: `HookRegistry._child_registries_cache` goes stale after `enable_cache`/`disable_cache` → `ValueError: No context is set`
### Describe the bug

`HookRegistry._get_child_registries()` caches the list of child-module registries (built by walking `named_modules()`) in `self._child_registries_cache` on first use and **never invalidates it**. But `register_hook` / `remove_hook` — invoked by `enable_cache()` / `disable_cache()` — change *which* child modules have a `_diffusers_hook`. When `cache_context()` is first entered while a model has **no** stateful block hooks (e.g. the cache is disabled), the cached child-registry list is built without those blocks. A later `enable_cache(FirstBlockCacheConfig(...))` adds `FirstBlockCache` hooks to the blocks, but `cache_context()._set_context()` still iterates the **stale** cached list, so the new block-level `StateManager`s are never given a context. The first cached forward then raises:

```
ValueError: No context is set. Please set a context before retrieving the state.
```

This is distinct from #12012 (Wan I2V pipeline not initializing the context, fixed in #12013): here the context *is* set on the transformer registry, but the staleness in `_child_registries_cache` prevents it reaching the freshly-registered block hooks. The bug is present on `main` (`register_hook`/`remove_hook` contain no cache invalidation) and was introduced by the `_child_registries_cache` optimization (docstring references Flux2).

### Reproduction

Fully self-contained — no downloads, no GPU. Builds a tiny randomly-initialized `FluxTransformer2DModel` on CPU (verified with `diffusers==0.38.0`, `torch==2.8.0`):

```python
import torch
from diffusers import FirstBlockCacheConfig, FluxTransformer2DModel

torch.manual_seed(0)

heads, head_dim = 2, 16
hidden = heads * head_dim  # 32
model = FluxTransformer2DModel(
    patch_size=1, in_channels=hidden, num_layers=2, num_single_layers=2,
    attention_head_dim=head_dim, num_attention_heads=heads,
    joint_attention_dim=32, pooled_projection_dim=16,
    guidance_embeds=False, axes_dims_rope=(2, 6, 8),  # sums to head_dim (16)
).eval()

B, T_img, T_txt = 1, 8, 4
def make_inputs():
    return dict(
        hidden_states=torch.randn(B, T_img, hidden),
        encoder_hidden_states=torch.randn(B, T_txt, 32),
        pooled_projections=torch.randn(B, 16),
        timestep=torch.tensor([1.0]),
        img_ids=torch.zeros(T_img, 3), txt_ids=torch.zeros(T_txt, 3),
        return_dict=False,
    )

# 1) Enter cache_context BEFORE any cache hooks exist (mirrors a warmup/inference pass that
#    wraps the call in cache_context() while caching is disabled). This populates
#    model._diffusers_hook._child_registries_cache WITHOUT the block registries.
with torch.no_grad(), model.cache_context("cond"):
    model(**make_inputs())

# 2) enable_cache() registers FirstBlockCache hooks on the blocks (new child registries) but
#    does NOT invalidate the parent's _child_registries_cache.
model.enable_cache(FirstBlockCacheConfig(threshold=0.1))

# 3) The stale cache means cache_context()._set_context() never reaches the new block hooks.
with torch.no_grad(), model.cache_context("cond"):
    model(**make_inputs())   # ValueError: No context is set
```

Output:

```
[1] warmup pass with no cache hooks: OK
[2] enable_cache: OK
[3] cached pass raised: ValueError: No context is set. Please set a context before retrieving the state.
[workaround] cleared _child_registries_cache
[4] cached pass after workaround: OK  (bug confirmed + workaround fixes it)
```

The same staleness also bites a plain `enable_cache → disable_cache → enable_cache` cycle on a long-lived module: the second `enable_cache` adds hooks whose registries aren't in the cache built during the first cycle.

### Root cause

`HookRegistry._get_child_registries()` memoizes `_child_registries_cache` and reuses it indefinitely:

```python
def _get_child_registries(self) -> list["HookRegistry"]:
    if not hasattr(self, "_child_registries_cache"):
        self._child_registries_cache = None
    if self._child_registries_cache is not None:
        return self._child_registries_cache          # <-- stale after register/remove_hook
    registries = []
    for module_name, module in unwrap_module(self._module_ref).named_modules():
        ...
    self._child_registries_cache = registries
    return registries
```

Neither `register_hook` nor `remove_hook` resets it, so the cache's invariant ("the set of descendant modules with `_diffusers_hook` is stable after first build") is violated by the very APIs that mutate that set.

### Suggested fix

Invalidate the cache whenever a descendant's `_diffusers_hook` is created or removed. The simplest correct option is to reset it in `register_hook` / `remove_hook` (and propagate to ancestors, or just reset on the registry being mutated and let parents rebuild lazily). E.g.:

```python
def register_hook(self, hook, name):
    ...
    self._child_registries_cache = None      # structural change -> invalidate

def remove_hook(self, name, recurse=True):
    ...
    self._child_registries_cache = None
```

Since a child registry is created via `HookRegistry.check_if_exists_or_initialize(child)` and a *parent's* cache is what goes stale, invalidation likely needs to clear the cache on the whole ancestor chain (or fall back to always walking `named_modules()` when correctness must hold). Happy to send a PR if you agree on the approach.

### Workaround

After `enable_cache(...)`, manually clear the cache before the next `cache_context()`:

```python
reg = getattr(transformer, "_diffusers_hook", None)
if reg is not None and getattr(reg, "_child_registries_cache", None) is not None:
    reg._child_registries_cache = None
```

### System Info

- diffusers: 0.38.0 (also present on `main`)
- transformers: 5.10.2
- torch: 2.8.0
- Python: 3.11
- Platform: Linux, NVIDIA H200

### Who can help?

Hook/caching: @a-r-r-o-w @DN6

SuryanshSS1011:
# What does this PR do?

Fixes #14037.

`HookRegistry._get_child_registries()` caches the child-module registries it finds by walking `named_modules()`, and never invalidates that cache. But `enable_cache()` / `disable_cache()` add and remove block-level hooks, changing which modules carry a `_diffusers_hook`. If `cache_context()` is first entered while no block hooks exist (e.g. a warmup pass with caching disabled), the parent registry caches an incomplete child list. A later `enable_cache(FirstBlockCacheConfig(...))` registers block hooks, but `_set_context()` still iterates the stale cache, so the new block `StateManager`s never receive a context and the next cached forward raises:

`ValueError: No context is set. Please set a context before retrieving the state.`

This adds `HookRegistry.invalidate_child_registries_cache()`, which clears the cached list across the module tree, and calls it from `enable_cache()` and `disable_cache()` after hooks are added/removed.

The staleness originates in `register_hook` / `remove_hook`, but those run on the *child* block registries, which can't reach the *parent* registry whose cache is stale. `enable_cache` / `disable_cache` operate on the root module, so invalidating there covers the reported scenario for every cache technique. Happy to move it into `register_hook` / `remove_hook` instead if you'd prefer it lower down.

The self-contained CPU reproduction from the issue passes after the fix, and a regression test is added in `tests/hooks/test_hooks.py`.

## Before submitting

## Who can review?
@DN6 @sayakpaul


tests/hooks/test_hooks.py
```diff
@@ -200,6 +200,26 @@ def test_stateful_hook(self):
         assert registry.get_hook("stateful_add_hook").increment == 1
         assert torch.allclose(output1, output2)
 
+    def test_child_registries_cache_invalidation(self):
```
sayakpaul, line 209: Should there also be testing when the context path is exercised, as reported in https://github.com/huggingface/diffusers/issues/14037?

SuryanshSS1011: Yes, that is a good call!

I've added `test_cache_context_after_enable_cache_with_prior_context`, which reproduces the issue end-to-end: it enters `cache_context()` before `enable_cache()` on a small FluxTransformer2DModel and runs a cached forward. It raises the original `ValueError: No context is set without the fix and passes with it`. Also fixed the `check_code_quality` failure (a docstring line-length restyle).

tests/hooks/test_hooks.py
```diff
@@ -372,3 +393,46 @@ def test_invocation_order_stateful_last(self):
...
+        "timestep": torch.tensor([1.0]),
+        "img_ids": torch.zeros(img_seq_len, 3),
+        "txt_ids": torch.zeros(txt_seq_len, 3),
+        "return_dict": False,
+    }
+
+    # Warmup pass inside a cache_context() while caching is disabled, then enable caching.
+    with torch.no_grad(), model.cache_context("cond"):
+        model(**inputs)
+    model.enable_cache(FirstBlockCacheConfig(threshold=0.1))
```
sayakpaul: Help me understand why would this be a practical flow though? 

Like `with torch.no_grad(), model.cache_context("cond"):` isn't doing anything from the perspectives of caching. So, does this reflect how caching is to be used actually in practice?

SuryanshSS1011: So our flow here is enabling caching on a transformer that has already been called once inside `cache_context()`.

The pipelines call `cache_context()` unconditionally around every transformer forward. For example `pipeline_qwenimage.py` does `with self.transformer.cache_context("cond"):` on every denoise step regardless of whether caching is enabled. So the first `cache_context()` in my test mirrors a normal pipeline run. It's a no-op for caching but it still builds `HookRegistry._child_registries_cache`. If the user then calls `enable_cache(...)` and runs again, which is a natural "run once, then turn on caching to speed up the next run" workflow, the second `cache_context()` hits that stale cache and the block hooks never get a context, which raises `ValueError: No context is set`.

To confirm it's not contrived, calling `enable_cache()` before any run works fine. The bug only appears when a `cache_context()` pass precedes `enable_cache()`. The test reduces that to the minimal repro, using direct `model(**inputs)` calls in place of a full pipeline loop so it stays CPU-only and fast, but the ordering is the same one a real pipeline produces.

We can reshape the test to go through an actual pipeline call if you'd prefer it to read more explicitly as the real flow.

sayakpaul: Yeah let's do that

SuryanshSS1011: I have committed the changes. 
 
 I moved the end-to-end test into `FirstBlockCacheTesterMixin` as `test_first_block_cache_enabled_after_inference`, so it now runs through a real pipeline: `pipe(**inputs)` once with caching off, then `enable_cache()`, then `pipe(**inputs)` again. Since it's in the mixin it covers every pipeline that runs the FirstBlockCache suite (Flux, LTX, Mochi, CogVideoX, Hunyuan). I confirmed it fails on `main` with the original `No context is set` and passes with the fix. Removed the old standalone test from `test_hooks.py`.

SuryanshSS1011:
Gentle bump on this one @sayakpaul. I moved the end-to-end test into `FirstBlockCacheTesterMixin` as you suggested. Anything else you'd like changed before this can go in?

tests/pipelines/test_pipelines_common.py
```diff
@@ -2994,6 +2994,31 @@ def run_forward(pipe):
...
+        pipe.set_progress_bar_config(disable=None)
+
+        def run_forward():
+            torch.manual_seed(0)
+            inputs = self.get_dummy_inputs(device)
+            inputs["num_inference_steps"] = 4
+            return pipe(**inputs)[0]
+
+        # First run populates the child-registry cache while caching is disabled.
+        run_forward()
```
sayakpaul: Are we assuming that the underlying pipeline implementation is following the following pattern (originally shown in https://github.com/huggingface/diffusers/issues/14037):

```
with torch.no_grad(), model.cache_context("cond"):
    model(**make_inputs())
```

There is also no presence of another

```
with torch.no_grad(), model.cache_context("cond"):
    model(**make_inputs())   # ValueError: No context is set
```

(as reported in https://github.com/huggingface/diffusers/issues/14037)

So, I think this test test is still incomplete.

SuryanshSS1011: Yes, you are right. I do realize that the test leans on the pipeline calling `cache_context()` internally and doesn't include the second context entry from the issue, so it's incomplete as-is.

I'll rework it to explicitly drive the two `cache_context("cond")` forwards from the issue's reproduction, with the second one being where `No context is set `was raised, rather than relying on the pipeline to do it implicitly. I'll make sure it fails on main and passes with the fix.

SuryanshSS1011:
@sayakpaul, I have pushed the rework. The test now drives the two `cache_context("cond")` forwards explicitly and asserts the pipeline actually called the transformer. Confirmed failing on main and passing with the fix across all the FirstBlockCache pipelines (Flux, LTX, Mochi, CogVideoX, HunyuanVideo, HunyuanImage). Ready for another look when you have a chance.

SuryanshSS1011:
@sayakpaul gentle bump. The test rework you asked for is pushed (drives the two cache_context("cond") forwards explicitly, fails on main and passes with the fix). Anything else before this can go in?

SuryanshSS1011:
Following up again. The test now drives both `cache_context('cond') `forwards explicitly, confirmed failing on main and passing with the fix. @DN6, tagging you too since you were listed as a reviewer, in case you have bandwidth to take a look.

tests/pipelines/test_pipelines_common.py
```diff
@@ -2831,6 +2831,52 @@ def run_forward(pipe):
             "Outputs from normal inference and after disabling cache should not differ."
         )
 
+    def test_first_block_cache_enabled_after_inference(self):
```
sayakpaul: Let's delete this test

sayakpaul (approved):
Thanks!

tests/pipelines/test_pipelines_common.py
SuryanshSS1011: Done, removed in c1a723b.

sayakpaul (approved):

PR #14096, opened 2026-06-30
support loading pipeline from transformer style (flat) repo 

yiyixuxu:
With this PR, we want to support loading a pipeline from [`google/diffusiongemma-26B-A4B-it`](https://huggingface.co/google/diffusiongemma-26B-A4B-it), and also future model releases in the same style :
* model released in transformers with the flat structure, i.e., model and processor are hosted at the root folder directly
* pipeline + scheduler in diffusers, and would be able to add a scheduler folder and `model_index.json` file into the same repo

Currently, we have some legacy support for loading components from the root, but very limited (e.g., it does not work if the repo also has a subfolder, or if your component comes with files other than the regular weights, etc.) . Since all our releases use subfolder structure, this was sufficient as it is (I'm actually not sure why the legacy support was there at all; I assume a very early model released that way, and we kept it for backward compatibility).

This PR extends that a bit further to support joint transformer + diffusers release (we will likely see more)
*  make it work with a mixed structure (some components at the root, some in subfolders). Here, `model` and `processor` live in the root, `scheduler` has its own folder)
* always allow a fixed list of known transformers auxiliary filenames at the root (`TRANSFORMERS_COMPONENT_AUX_FILES`: tokenizer/processor/chat-template/generation configs etc), currently, we only allow the list of files DiffusionGemma requires, but we can extend as we support more pipelines

you can test with this
```py
import torch
from diffusers import DiffusionGemmaPipeline

# model_id = "google/diffusiongemma-26B-A4B-it" 
model_id = "dg845/diffusiongemma-test"

pipe = DiffusionGemmaPipeline.from_pretrained(
    model_id,
    torch_dtype=torch.bfloat16,
    # revision="refs/pr/20",   # Omit if merged to main
)
pipe.to("cuda")

output = pipe(
    prompt="Why is the sky blue?",
    gen_length=256,
    num_inference_steps=48,
    cache_implementation="static",
    generator=torch.Generator("cuda").manual_seed(42),
)
print(output.texts[0])
```

### DIFF
--- a/src/diffusers/pipelines/pipeline_loading_utils.py
+++ b/src/diffusers/pipelines/pipeline_loading_utils.py
@@ -136,6 +136,8 @@ def is_safetensors_compatible(filenames, passed_components=None, folder_names=No
     )
 
     passed_components = passed_components or []
+    # only weight files matter for safetensors compatibility
+    filenames = filter_model_files(filenames)
     if folder_names:
         filenames = {f for f in filenames if os.path.split(f)[0] in folder_names}
 
--- a/src/diffusers/pipelines/pipeline_loading_utils.py
+++ b/src/diffusers/pipelines/pipeline_loading_utils.py
@@ -69,6 +69,27 @@ DUMMY_MODULES_FOLDER = "diffusers.utils"
 TRANSFORMERS_DUMMY_MODULES_FOLDER = "transformers.utils"
 CONNECTED_PIPES_KEYS = ["prior"]
 
+# Auxiliary (non-weight) files a transformers component saves next to its weights. Repos with a flat,
+# transformers-style layout host a component's files at the repo root instead of in a subfolder, where the
+# folder-based allow patterns of `DiffusionPipeline.download` would miss them. Root-hosted weights and
+# `config.json` are matched by their own patterns, so only these auxiliary filenames need listing.
+TRANSFORMERS_COMPONENT_AUX_FILES = [
+    "added_tokens.json",
+    "chat_template.jinja",
+    "chat_template.json",
+    "generation_config.json",
+    "merges.txt",
+    "preprocessor_config.json",
+    "processor_config.json",
+    "special_tokens_map.json",
+    "spiece.model",
+    "tokenizer.json",
+    "tokenizer.model",
+    "tokenizer_config.json",
+    "vocab.json",
+    "vocab.txt",
+]
+
 logger = logging.get_logger(__name__)
 
 LOADABLE_CLASSES = {
--- a/src/diffusers/pipelines/pipeline_loading_utils.py
+++ b/src/diffusers/pipelines/pipeline_loading_utils.py
@@ -73,21 +73,13 @@ CONNECTED_PIPES_KEYS = ["prior"]
 # transformers-style layout host a component's files at the repo root instead of in a subfolder, where the
 # folder-based allow patterns of `DiffusionPipeline.download` would miss them. Root-hosted weights and
 # `config.json` are matched by their own patterns, so only these auxiliary filenames need listing.
+# Currently the set needed by DiffusionGemma — extend as new flat-layout pipelines require it.
 TRANSFORMERS_COMPONENT_AUX_FILES = [
-    "added_tokens.json",
     "chat_template.jinja",
-    "chat_template.json",
     "generation_config.json",
-    "merges.txt",
-    "preprocessor_config.json",
     "processor_config.json",
-    "special_tokens_map.json",
-    "spiece.model",
     "tokenizer.json",
-    "tokenizer.model",
     "tokenizer_config.json",
-    "vocab.json",
-    "vocab.txt",
 ]
 
 logger = logging.get_logger(__name__)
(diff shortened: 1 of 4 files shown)


src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1751,6 +1752,11 @@ def download(cls, pretrained_model_name, **kwargs) -> str | os.PathLike:
                 p for p in allow_patterns if not (len(p.split("/")) == 2 and p.split("/")[0] in passed_components)
             ]
 
+            # Files explicitly allow-listed by the repo author via `_allow_files` in `model_index.json` are
+            # added to the download set. This supports repos that keep a component's config/tokenizer files at
+            # the root (instead of in its own subfolder), where the folder-based allow patterns would miss them.
+            allow_patterns += allow_filenames
```
yiyixuxu: need to work with field like this https://huggingface.co/YiYiXu/diffusiongemma-test/blob/main/model_index.json#L4

yiyixuxu: alternative is we could just add these processor-related files into the default `allow_patterns` here (without the model author having to specify in `model_index.json`), I leaned toward the design in this PR mainly because:

(1) it's more practical and more remote friendly - for a repo hosted this way with slightly different files, the model author can just update the `model_index.json` instead of PR into diffusers 
(2) more fundamentally, I don't want to broadly support loading components from the roots - it is just too ambiguous for diffusers because we have multiple components and we often co-host our checkpoints in the same repo as the research repo's original checkpoint and their checkpoint is hosted at the root.  We already have some implicit root-loading support today, but I don't think we need to expand it as default behavior. `_allow_files` keeps it explicit and opt-in

But I'm open to different opinions!

tests/pipelines/test_pipeline_utils.py
```diff
@@ -218,6 +218,34 @@ def test_diffusers_is_compatible_no_components_only_variants(self):
...
+        ]
+        self.assertTrue(is_safetensors_compatible(filenames))
+
+    def test_diffusers_is_compatible_no_components_safetensors_only_variants(self):
+        filenames = [
+            "diffusion_pytorch_model.fp16.safetensors",
+        ]
+        self.assertTrue(is_safetensors_compatible(filenames, variant="fp16"))
+
+    def test_diffusers_is_compatible_weightless_subfolder(self):
```
yiyixuxu: only this test will have a different result from `main`
Currently, in `main` this will marked as not `is_safetensors_compatible`

src/diffusers/pipelines/pipeline_loading_utils.py
```diff
@@ -136,6 +136,8 @@ def is_safetensors_compatible(filenames, passed_components=None, folder_names=No
     )
 
     passed_components = passed_components or []
+    # only weight files matter for safetensors compatibility
+    filenames = filter_model_files(filenames)
```
yiyixuxu, line 153: i think it's a bug here but let me know cc @DN6

dg845 (approved):
Thanks for the PR!

src/diffusers/pipelines/pipeline_utils.py
sayakpaul: The current one is better as it gives the idea that model authors need to be explicit about this kind of structure. Loading it implicitly will lead to confusion.

tests/pipelines/test_pipeline_utils.py
```diff
@@ -218,6 +218,34 @@ def test_diffusers_is_compatible_no_components_only_variants(self):
...
+        # transformers-style flat layout: safetensors weights at the root + a weight-less subfolder (scheduler/)
+        filenames = [
+            "diffusion_pytorch_model.safetensors",
```
sayakpaul: Should we follow the transformers-style weight naming convention here? (`model-00001-of-00011.safetensors`, and so on)?

sayakpaul (approved):

tests/pipelines/test_pipeline_utils.py
DN6: Oh so main is wrong here? This test should evaluate to `compatible == True`.

src/diffusers/pipelines/pipeline_loading_utils.py
DN6: LGTM 👍🏽

src/diffusers/pipelines/pipeline_utils.py
DN6: How does one define the allow list when saving the model? My one concern here is that this introduces a parameter that allows a repo owner to change which files are downloaded from the hub repo onto a users machine without the end user explicitly opting in for it. e.g. I can add `_allow_files=['weird.pt']`  and this would end up on a users machine. 

Allowing root level components is probably not ideal, but in this case we'd just be expanding to `transformers` components and we would be unlikely to add more file types. IMO this tradeoff is okay because with `_allow_file` we would be enabling downloading potentially any file type without explicit opt in.

yiyixuxu: ohh that's really good point!

DN6 (approved):
Thanks 👍🏽

PR #14100, opened 2026-07-01
Add Nunchaku Lite single-file quantization

rootonchair:
# What does this PR do?

Adds Nunchaku Lite single-file checkpoint loading for Diffusers models.

This introduces `NunchakuLiteQuantizationConfig` and a new Nunchaku Lite quantizer that can patch supported nn.Linear modules into runtime SVDQ/AWQ linear layers before strict checkpoint loading. The loader reads safetensors metadata during `from_single_file` so Nunchaku Lite checkpoints can use their embedded runtime manifest to decide which modules to replace.

## Deprecated API
```python
import torch
from diffusers import (
    ErnieImagePipeline,
    ErnieImageTransformer2DModel,
    NunchakuLiteQuantizationConfig,
)

checkpoint = hf_hub_download(
    repo_id="rootonchair/ERNIE-Image-Turbo-nunchaku-lite",
    filename="svdq-int4_r32-ernie-image-turbo-zero-svdq-fix-bias.safetensors",
)

transformer = ErnieImageTransformer2DModel.from_single_file(
    checkpoint,
    config="baidu/ERNIE-Image-Turbo",
    subfolder="transformer",
    quantization_config=NunchakuLiteQuantizationConfig(compute_dtype=torch.bfloat16),
    torch_dtype=torch.bfloat16,
)

pipe = ErnieImagePipeline.from_pretrained(
    "baidu/ERNIE-Image-Turbo",
    transformer=transformer,
    torch_dtype=torch.bfloat16,
)

pipe.to("cuda")

image = pipe(
    prompt="A modern red armchair in a quiet studio, soft window light, realistic product photography",
    height=512,
    width=512,
    num_inference_steps=8,
    guidance_scale=1.0,
    generator=torch.Generator(device="cuda").manual_seed(1234),
    use_pe=False,
).images[0]

image.save("ernie-image-turbo-nunchaku-lite.png")
```

## New API for `from_single_file` use
```python
import torch

from diffusers import (
    ErnieImagePipeline,
    ErnieImageTransformer2DModel,
    NunchakuLiteQuantizationConfig,
)


dtype = torch.bfloat16
device = "cuda"

checkpoint = hf_hub_download(
    repo_id="rootonchair/ERNIE-Image-Turbo-nunchaku-lite",
    filename="svdq-int4_r32-ernie-image-turbo-zero-svdq-fix-bias.safetensors",
)

svdq_targets = []
for name in [
    "self_attention.to_q",
    "self_attention.to_k",
    "self_attention.to_v",
    "self_attention.to_out.0",
    "mlp.gate_proj",
    "mlp.up_proj",
    "mlp.linear_fc2",
]:
    svdq_targets.extend([f"layers.{i}.{name}" for i in range(36)])

quantization_config = NunchakuLiteQuantizationConfig(
    compute_dtype=dtype,
    svdq_w4a4={
        "precision": "int4",
        "group_size": 64,
        "rank": 32,
        "targets": svdq_targets,
    },
    awq_w4a16={
        "precision": "int4",
        "group_size": 64,
        "targets": [
            "text_proj",
            "time_embedding.linear_1",
            "time_embedding.linear_2",
            "adaLN_modulation.1",
            "final_norm.linear",
            "final_linear",
        ],
    },
)

transformer = ErnieImageTransformer2DModel.from_single_file(
    checkpoint,
    config="baidu/ERNIE-Image-Turbo",
    subfolder="transformer",
    quantization_config=quantization_config,
    torch_dtype=dtype,
)

pipe = ErnieImagePipeline.from_pretrained(
    "baidu/ERNIE-Image-Turbo",
    transformer=transformer,
    torch_dtype=dtype,
)

pipe.to(device)

image = pipe(
    prompt="A modern red armchair in a quiet studio, soft window light, realistic product photography",
    height=512,
    width=512,
    num_inference_steps=8,
    guidance_scale=1.0,
    generator=torch.Generator(device=device).manual_seed(1234),
    use_pe=False,
).images[0]

image.save("ernie-image-turbo-nunchaku-lite.png")
```

Fixes # (issue)

## Before submitting
- [ ] Did you read the [contributor guideline](https://huggingface.co/docs/diffusers/main/en/conceptual/contribution)?
- [ ] Did you read our [philosophy doc](https://huggingface.co/docs/diffusers/main/en/conceptual/philosophy)? (important for complex PRs)

## Who can review?


src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
```diff
@@ -0,0 +1,97 @@
...
+
+logger = logging.get_logger(__name__)
+
+
+class NunchakuLiteQuantizer(DiffusersQuantizer):
+    def __init__(self, quantization_config, **kwargs):
+        super().__init__(quantization_config, **kwargs)
+        self.compute_dtype = quantization_config.compute_dtype
+        self.pre_quantized = quantization_config.pre_quantized
+        self.runtime_manifest = None
```
sayakpaul: What do we need this for?

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
```diff
@@ -0,0 +1,97 @@
...
+            raise ValueError("Nunchaku Lite checkpoints must include a JSON `quantization_config` metadata field.")
+        try:
+            quantization_config = json.loads(metadata["quantization_config"])
+        except json.JSONDecodeError as exc:
+            raise ValueError("Nunchaku Lite checkpoint metadata field `quantization_config` is not valid JSON.") from exc
+        if not isinstance(quantization_config, dict):
+            raise ValueError("Nunchaku Lite checkpoint metadata field `quantization_config` must decode to a JSON object.")
+        return quantization_config
+
+    def _check_strict_state_dict_match(self, model: "ModelMixin", state_dict: dict[str, Any]):
```
sayakpaul: Feels like it's better off in `utils.py`?

src/diffusers/quantizers/quantization_config.py
```diff
@@ -429,6 +430,24 @@ def __init__(self, compute_dtype: "torch.dtype" | None = None):
...
+
+    Args:
+        compute_dtype (`torch.dtype`, defaults to `torch.bfloat16`):
+            Runtime dtype used by the floating-point buffers in the quantized modules.
+    """
+
+    def __init__(self, compute_dtype: "torch.dtype" | None = None):
+        self.quant_method = QuantizationMethod.NUNCHAKU_LITE
+        self.compute_dtype = compute_dtype
+        self.pre_quantized = True
```
sayakpaul, line 494: Can we also guide the readers on how to obtain the checkpoints?

Also, can we ensure torch.compile compatibility?

tests/quantization/nunchaku/test_nunchaku.py
```diff
@@ -0,0 +1,161 @@
+import json
```
sayakpaul: For tests, WDYT of adding a mixin to https://github.com/huggingface/diffusers/blob/main/tests/models/testing_utils/quantization.py and then extending a popular model like Flux to use that mixin?

sayakpaul (commented):
Thanks for getting started! Just did a first pass and left high-level reviews.

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
rootonchair: In nunchaku checkpoint, it's a mix of SVDQuant and AWQ, so this for maintaining a list of config telling which and how Linear need to be replaced (borrows from https://github.com/rootonchair/diffuse-compressor/blob/main/docs/nunchaku_lite_manifest_v1.md#target-entries)

This is the schema of each items:
```
{
  "name": "single_transformer_blocks.0.attn.to_qkv_mlp_proj.linears.0",
  "checkpoint_prefix": "single_transformer_blocks.0.attn.qkv_proj",
  "source_modules": [
    "single_transformer_blocks.0.attn.to_qkv_mlp_proj.linears.0"
  ],
  "roles": [],
  "kind": "linear",
  "nunchaku_op": "svdq_w4a4",
  "precision": "fp4",
  "group_size": 16,
  "rank": 32,
  "has_bias": true,
  "op_options": {
    "outer_scale_splits": [4096, 4096, 4096]
  },
  "activation": {}
}
```

what we need is only 
 ```
{
  "checkpoint_prefix": "single_transformer_blocks.0.attn.qkv_proj",
  "kind": "linear",
  "nunchaku_op": "svdq_w4a4",
  "precision": "fp4",
  "group_size": 16,
  "rank": 32,
  "has_bias": true,
}
```

File: src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py (hunk 2 above)
Comment (rootonchair) [reply src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py]: Sure, I will move it to `utils.py`

File: tests/quantization/nunchaku/test_nunchaku.py (hunk 4 above)
Comment (rootonchair) [reply tests/quantization/nunchaku/test_nunchaku.py]: Yes, let's do it that way

File: src/diffusers/quantizers/quantization_config.py (hunk 3 above)
Comment (rootonchair) [reply src/diffusers/quantizers/quantization_config.py:494]: The kernels are compatible with torch.compile, as well as `SVDQLinear` and `AWQLinear`, I will make a test to assure that the compatibility still remains when we integrate to diffusers

> Can we also guide the readers on how to obtain the checkpoints?

I'm a little confused here. Could you help provide more context

**Comment by rootonchair** [comment]
I just did some benchmark on RTX PRO 6000, here is the visual result between bf16 and nvfp4 checkpoint for ERNIE-Image-Turbo

| BF16 | Nunchaku NVFP4 |
|:----:|:----:|
| image| image |


| Case | Full mean | Denoise mean | Denoise peak alloc | Full peak alloc | Speedup |
|:---:|:---:|:---:|:----:|:----:|:----:|
| original | 3.003s | 2.862s | 29.429GB | 31.081GB | 1.0x |
| nunchaku_lite NVFP4 | 2.271s | 2.127s | 18.926GB | 20.578GB | 1.35x |
| nunchaku_lite NVFP4 + compile | 1.675s | 1.525s | 18.672GB | 20.578GB | 1.8x |
| nunchaku_lite NVFP4 + bnb text encoder | 2.285s | 2.132s | 14.317GB | 15.969GB | 1.35x |

By replacing Nunchaku Linear, we have reduced the latency of these linear operations by 2x with large shape

| Target | Op | Rows | Shape | Normal ms | Nunchaku ms | Speedup |
|---|---|---:|---:|---:|---:|---:|
| `layers.0.self_attention.to_q` | `svdq_w4a4` | 4096 | 4096 -> 4096 | 0.3660 | 0.1563 | 2.34x |
| `layers.0.mlp.gate_proj` | `svdq_w4a4` | 4096 | 4096 -> 12288 | 1.0646 | 0.4272 | 2.49x |
| `layers.0.mlp.linear_fc2` | `svdq_w4a4` | 4096 | 12288 -> 4096 | 1.0269 | 0.4596 | 2.23x |

One note here, AWQ only benefit with adaLN layer, so other modules like time embedding or final linear can stay as bf16

| Target | Op | Rows | Shape | Normal ms | Nunchaku ms | Speedup |
|---|---|---:|---:|---:|---:|---:|
| `text_proj` | `awq_w4a16` | 1 | 3072 -> 4096 | 0.0111 | 0.0201 | 0.55x |
| `time_embedding.linear_1` | `awq_w4a16` | 1 | 4096 -> 4096 | 0.0129 | 0.0251 | 0.52x |
| `time_embedding.linear_2` | `awq_w4a16` | 1 | 4096 -> 4096 | 0.0129 | 0.0247 | 0.52x |
| `adaLN_modulation.1` | `awq_w4a16` | 1 | 4096 -> 24576 | 0.1389 | 0.0243 | 5.71x |
| `final_norm.linear` | `awq_w4a16` | 1 | 4096 -> 8192 | 0.0220 | 0.0245 | 0.90x |
| `final_linear` | `awq_w4a16` | 4114 | 4096 -> 128 | 0.0248 | 0.0635 | 0.39x |

**Comment by rootonchair** [comment]
I have just implemented the native loading feature, which now can load by `from_pretrained` with converted repo:
- int4: https://huggingface.co/rootonchair/ERNIE-Image-Turbo-nunchaku-lite-int4
- nvfp4: https://huggingface.co/rootonchair/ERNIE-Image-Turbo-nunchaku-lite-nvfp4

```python
import torch
from diffusers import ErnieImagePipeline

pipe = ErnieImagePipeline.from_pretrained(
    "rootonchair/ERNIE-Image-Turbo-nunchaku-lite-int4",
    torch_dtype=torch.bfloat16,
).to("cuda")

image = pipe(
    prompt="A modern red armchair in a quiet studio, soft window light, realistic product photography",
    height=1024,
    width=1024,
    num_inference_steps=8,
    guidance_scale=1.0,
    use_pe=False,
).images[0]

image.save("ernie-image-turbo-nunchaku-lite-int4.png")

```

Quantization config now change to:
```
"quantization_config": {
    "awq_w4a16": {
      "group_size": 64,
      "precision": "int4",
      "targets": [
        "text_proj",
        "time_embedding.linear_1",
        "time_embedding.linear_2",
        "adaLN_modulation.1",
        "final_norm.linear",
        "final_linear"
      ]
    },
    "compute_dtype": "bfloat16",
    "quant_method": "nunchaku_lite",
    "svdq_w4a4": {
      "group_size": 16,
      "precision": "fp4",
      "rank": 32,
      "targets": [
        "layers.0.self_attention.to_q",
        "layers.1.self_attention.to_q",
        "layers.2.self_attention.to_q",
        "layers.3.self_attention.to_q",
         ...
      ]
    }
```

If we agree to use this schema, I will remove the old metadata/from_single_file approach

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
```diff
@@ -0,0 +1,85 @@
...
+        logger.info(f"Applied Nunchaku quantization config with {len(runtime_manifest.targets)} targets.")
+
+    def _process_model_after_weight_loading(self, model: "ModelMixin", **kwargs):
+        return model
+
+    @property
+    def is_serializable(self):
+        return False
+
+    @property
```
sayakpaul, line 83: We should set `is_compileable()` property too:
https://github.com/huggingface/diffusers/blob/9159a587b1f148a2cc911deb72b6d755508ef38b/src/diffusers/quantizers/base.py#L263

src/diffusers/quantizers/quantization_config.py
sayakpaul: > I'm a little confused here. Could you help provide more context

How are the example checkpoints obtained? I think we're only dealing with pre-quantized checkpoints in this PR?

src/diffusers/quantizers/quantization_config.py
```diff
@@ -429,6 +430,46 @@ def __init__(self, compute_dtype: "torch.dtype" | None = None):
...
+            raise ValueError("Nunchaku compute_dtype must be a string or a torch.dtype.")
+        self.compute_dtype = compute_dtype
+        self.pre_quantized = True
+        self.svdq_w4a4 = svdq_w4a4
+        self.awq_w4a16 = awq_w4a16
+
+    def to_dict(self) -> dict[str, Any]:
+        output = super().to_dict()
+        output["compute_dtype"] = str(output["compute_dtype"]).split(".")[1]
+        return output
```
sayakpaul, line 551: Why do we need to override this?

sayakpaul (commented):
Looking good. I think we can remove all `metadata` related code?

src/diffusers/quantizers/quantization_config.py
```diff
@@ -429,6 +430,46 @@ def __init__(self, compute_dtype: "torch.dtype" | None = None):
...
+        if isinstance(compute_dtype, str):
+            if not hasattr(torch, compute_dtype):
+                raise ValueError(f"Unsupported Nunchaku compute dtype: {compute_dtype!r}.")
+            compute_dtype = getattr(torch, compute_dtype)
+        if not isinstance(compute_dtype, torch.dtype):
+            raise ValueError("Nunchaku compute_dtype must be a string or a torch.dtype.")
+        self.compute_dtype = compute_dtype
+        self.pre_quantized = True
+        self.svdq_w4a4 = svdq_w4a4
+        self.awq_w4a16 = awq_w4a16
```
sayakpaul, line 496: Do we need any validation around these two?

src/diffusers/quantizers/quantization_config.py
rootonchair: I reference BitsAndBytesConfig to serialize this torch.dtype property type
https://github.com/huggingface/diffusers/blob/main/src/diffusers/quantizers/quantization_config.py#L375

src/diffusers/quantizers/quantization_config.py
rootonchair: Sure, we do need it

src/diffusers/quantizers/quantization_config.py
rootonchair: Yes we are only dealing with pre-quantized checkpoint here. Perhaps we can leave a comment that said the checkpoints is quantized with diffuse-compressor + run diffuser format converter?

sayakpaul: I think it'd be better off in the docs?

sayakpaul (commented):
No rush but let us know once you would like another round of review.

rootonchair:
@sayakpaul I think this is ready for the next review

src/diffusers/quantizers/nunchaku/utils.py
```diff
@@ -0,0 +1,356 @@
...
+
+_ops = None
+
+
+def _get_ops():
+    global _ops
+    if _ops is None:
+        from kernels import get_kernel
+
+        _ops = get_kernel(_HF_KERNEL_REPO, version=_HF_KERNEL_VERSION, trust_remote_code=True).ops
```
SunMarc: Should we make it a trusted organization @sayakpaul ? not sure if this is good to leave `trust_remote_code=True`. Also, I think it is fine to not pin the version unless we really need it no ?

src/diffusers/quantizers/nunchaku/utils.py
```diff
@@ -0,0 +1,356 @@
...
+_HF_KERNEL_VERSION = 1
+
+
+_ops = None
+
+
+def _get_ops():
+    global _ops
+    if _ops is None:
+        from kernels import get_kernel
```
SunMarc: let's make this more specific since this function is just there to load the nunchaku_kernels. maybe just all it `nunchaku_kernels` ?

src/diffusers/quantizers/nunchaku/utils.py
```diff
@@ -0,0 +1,356 @@
...
+            message += f" Missing keys: {missing_keys[:10]}"
+            if len(missing_keys) > 10:
+                message += f" and {len(missing_keys) - 10} more"
+            message += "."
+        if unexpected_keys:
+            message += f" Unexpected keys: {unexpected_keys[:10]}"
+            if len(unexpected_keys) > 10:
+                message += f" and {len(unexpected_keys) - 10} more"
+            message += "."
+        raise ValueError(message)
```
SunMarc, line 349: we don't need this no ? we already print some logs when there are missing or unexpected keys.

src/diffusers/quantizers/nunchaku/utils.py
```diff
@@ -0,0 +1,356 @@
...
+        if op == "svdq_w4a4":
+            replacement = SVDQW4A4Linear(
+                in_features,
+                out_features,
+                rank=rank,
+                bias=bias is not None,
+                precision="nvfp4" if precision == "fp4" else precision,
+                group_size=group_size,
+                torch_dtype=compute_dtype,
+                device=_module_device(module),
```
SunMarc: the module is practically always on meta device when loading with low_cpu_mem_usage. I think it should be fine to leave device to `None` no ? Also, if you want to be sure that the modules are loaded on meta device, we usually use 
```
ctx = init_empty_weights if is_accelerate_available() else nullcontext
            with ctx():
```

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
```diff
@@ -0,0 +1,66 @@
...
+        if not is_kernels_available():
+            raise ImportError(
+                "Loading Nunchaku checkpoints requires the Hugging Face `kernels` package. "
+                "Install it with `pip install kernels`."
+            )
+
+    def update_torch_dtype(self, torch_dtype):
+        if torch_dtype is None:
+            torch_dtype = self.compute_dtype
+        return torch_dtype
```
SunMarc, line 59: what is the `torch_dtype` passed in from_pretrained don't match the `compute_dtype` specified in the `quantization_config` ? there will be an issue no ? Should be overwrite `compute_dtype` in that case ?

src/diffusers/quantizers/nunchaku/utils.py
```diff
@@ -0,0 +1,356 @@
...
+            raise ValueError(f"Nunchaku target {target!r} does not exist in the model.") from exc
+
+        in_features = getattr(module, "in_features", None)
+        out_features = getattr(module, "out_features", None)
+        bias = getattr(module, "bias", None)
+        if not isinstance(in_features, int) or not isinstance(out_features, int):
+            raise TypeError(f"Nunchaku target {target!r} must expose integer in_features/out_features.")
+
+        if op == "svdq_w4a4":
+            replacement = SVDQW4A4Linear(
```
SunMarc: this only works on specific machines, should we catch this beforehand or just let the user get an error when trying to get/run the kernels ?

src/diffusers/quantizers/nunchaku/utils.py
```diff
@@ -0,0 +1,356 @@
...
+                in_features,
+                out_features,
+                rank=rank,
+                bias=bias is not None,
+                precision="nvfp4" if precision == "fp4" else precision,
```
SunMarc: why not just use `nvfp4` instead of `fp4` when naming the precision ?

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,123 @@
...
+
+> [!TIP]
+> Use `fp4` on Blackwell GPUs. Running `int4` checkpoints on Blackwell can be slower than `fp4`.
+
+| Method | Precision | Group size | Notes |
+|---|---:|---:|---|
+| `svdq_w4a4` | `fp4` | 16 | Uses NVFP4 runtime kernels with SVDQ low-rank correction. |
+| `svdq_w4a4` | `int4` | 64 | Uses INT4 W4A4 kernels with SVDQ low-rank correction. |
+| `awq_w4a16` | `int4` | 64 | Uses INT4 weight-only AWQ-style kernels. |
+
```
SunMarc: maybe we should add some recommendation of hardware to be used with ?

SunMarc (commented):
Thanks ! added some comments

src/diffusers/quantizers/nunchaku/utils.py
sayakpaul: It's ALWAYS recommended to pin a version when using kernels. For now, `trust_remote_code=True` is fine. We can flip it once there's usage and more community trust.

src/diffusers/quantizers/nunchaku/utils.py
sayakpaul: Additionally, can we avoid using `global`?

rootonchair: This for loading the kernels lazily, I don't want the user to load the  nunchaku kernels if they don't use it. Or am I just overthink it? 😅

src/diffusers/quantizers/nunchaku/utils.py
rootonchair: I think it will be a good idea to make the precision name shorter. But I will change it back

docs/source/en/quantization/nunchaku.md
rootonchair: yes, I will be a good addition

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
rootonchair: yes we should

src/diffusers/quantizers/nunchaku/utils.py
rootonchair: I want a hard fail on this as partial loaded model will make inference incorrect and produce noisy results

src/diffusers/quantizers/nunchaku/utils.py
rootonchair: yes, I think using `ctx` would be more correct

src/diffusers/quantizers/nunchaku/utils.py
rootonchair: I think we should let the kernels library handling it when loading. @sayakpaul WDYT

asomoza:
nice, I tested it and made a krea model, it works pretty good!

comparison_bf16_vs_nvfp4

```python
import torch

pipe = DiffusionPipeline.from_pretrained("OzzyGT/Krea_2_Turbo_nunchaku_lite_nvfp4", torch_dtype=torch.bfloat16)
pipe.to("cuda")

prompt = (
    "A cozy corner bookstore-cafe on a rainy evening, cinematic wide shot. "
    'A large hand-lettered chalkboard sign in the window reads "FRESH COFFEE & OLD BOOKS" '
    "and below it in smaller chalk letters \"open 'til late\". "
    "Warm golden light spills onto wet cobblestones that mirror pink and blue neon reflections. "
    "Inside, tall mahogany shelves are packed with hundreds of colorful book spines with tiny legible titles, "
    "a barista in a striped apron pours delicate latte art, steam curling upward, "
    "a tabby cat sleeps on a windowsill beside a stack of paperbacks. "
    "Intricate detail, sharp focus, shallow depth of field, photorealistic, rich color grading."
)

image = pipe(
    prompt,
    num_inference_steps=8,
    guidance_scale=0.0,
    height=1024,
    width=1024,
    generator=torch.Generator("cuda").manual_seed(7),
).images[0]
image.save("sample.png")
```

src/diffusers/quantizers/nunchaku/utils.py
rootonchair: I have added the validation to `validate_environment` for this

src/diffusers/quantizers/nunchaku/utils.py
rootonchair: I think this has been resolved

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+## torch.compile
+
+Nunchaku Lite kernels and quantized linear layers are compatible with [`torch.compile`](../optimization/fp16#torchcompile).
+Compile the quantized transformer after loading the pipeline for faster inference.
+
+```python
+pipe.transformer = torch.compile(pipe.transformer, mode="default", fullgraph=False)
```
sayakpaul: Why can't we do it `fullgraph=True`? Can you open an issue on the nunchaku-lite repo with the trace?

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,123 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+-->
+
+# Nunchaku Lite
+
+Nunchaku Lite is a quantization backend for loading prequantized checkpoints in Diffusers. Use
```
sayakpaul: Should provide some citations to the original Nunchaku repo since Nunchaku Lite is based off of it?

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,123 @@
...
+import torch
+from diffusers import DiffusionPipeline
+
+model_id = "rootonchair/ERNIE-Image-Turbo-nunchaku-lite-nvfp4"
+
+pipe = DiffusionPipeline.from_pretrained(
+    model_id,
+    torch_dtype=torch.bfloat16,
+)
+pipe.to("cuda")
```
sayakpaul:
```suggestion
pipe = DiffusionPipeline.from_pretrained(
    model_id, torch_dtype=torch.bfloat16,
).to("cuda")
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,123 @@
...
+image = pipe(
+    prompt=prompt,
+    height=1024,
+    width=1024,
+    num_inference_steps=8,
+    guidance_scale=1.0,
+    use_pe=False,
+).images[0]
+image.save("ernie-image-turbo-nunchaku-lite.png")
+```
```
sayakpaul, line 54: Is it compatible with offloading, too?

docs/source/en/quantization/nunchaku.md
sayakpaul: I think it was not addressed? I think should mention a matrix of the quant schemes and their supported hardware / CUDA capabilities. Also, the note on hopper since we cannot support it.

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,123 @@
...
+
+| Method | Precision | Group size | Notes |
+|---|---:|---:|---|
+| `svdq_w4a4` | `fp4` | 16 | Uses NVFP4 runtime kernels with SVDQ low-rank correction. |
+| `svdq_w4a4` | `int4` | 64 | Uses INT4 W4A4 kernels with SVDQ low-rank correction. |
+| `awq_w4a16` | `int4` | 64 | Uses INT4 weight-only AWQ-style kernels. |
+
+## NunchakuLiteQuantizationConfig
+
+The `model.json` file must include a compact [`NunchakuLiteQuantizationConfig`]. It defines the runtime
```
sayakpaul: Did you mean `config.json`?

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,123 @@
...
+      "targets": ["layers.0.self_attention.to_q"]
+    },
+    "awq_w4a16": {
+      "precision": "int4",
+      "group_size": 64,
+      "targets": ["final_linear"]
+    }
+  }
+}
+```
```
sayakpaul, line 121: Can we briefly summarize what would this config do? Would it apply a combination of `svdq_w4a4` and `awq_w4a16` to the `targets`?

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
```diff
@@ -0,0 +1,89 @@
...
+logger = logging.get_logger(__name__)
+
+
+class NunchakuLiteQuantizer(DiffusersQuantizer):
+    def __init__(self, quantization_config, **kwargs):
+        super().__init__(quantization_config, **kwargs)
+        self.compute_dtype = quantization_config.compute_dtype
+        self.pre_quantized = quantization_config.pre_quantized
+
+    def validate_environment(self, *args, **kwargs):
```
sayakpaul, line 24: TODO: Hopper validation

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
sayakpaul: I guess this is addressed?

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
```diff
@@ -0,0 +1,89 @@
...
+    def _process_model_before_weight_loading(
+        self,
+        model: "ModelMixin",
+        state_dict: dict[str, Any] | None = None,
+        **kwargs,
+    ):
+        from .utils import (
+            check_strict_state_dict_match,
+            replace_with_nunchaku_linear,
+        )
```
sayakpaul:
```suggestion
        from .utils import check_strict_state_dict_match, replace_with_nunchaku_linear
```

src/diffusers/quantizers/quantization_config.py
```diff
@@ -429,6 +430,127 @@ def __init__(self, compute_dtype: "torch.dtype" | None = None):
...
+
+        self.post_init()
+
+    def post_init(self):
```
sayakpaul, line 500: Very nice validation!

tests/quantization/nunchaku/test_nunchaku.py
sayakpaul: We can remove this test then?

sayakpaul (commented):
Nice, thanks a lot for the further updates. I will post about the benchmarking I conducted through HF Jobs and Claude Code below.

sayakpaul:
Even though the below is generated with Claude Code (I have manually verified them).

## Nunchaku Lite (PR #14100) — testing notes

I ran the snippet from `docs/source/en/quantization/nunchaku.md` end-to-end. TL;DR: the compiled NVFP4 benchmark in the doc **reproduces cleanly** on the same class of hardware, but I hit two rough edges (hardware/torch-version compatibility) that are worth documenting or guarding.

## Reproduced the doc benchmark ✅

Ran the exact doc snippet (`rootonchair/ERNIE-Image-Turbo-nunchaku-lite-nvfp4`, 1024×1024, 8 steps, `use_pe=False`) on an **RTX PRO 6000 Blackwell** (via HF Jobs, `--flavor rtx-pro-6000`) — the same GPU the doc benchmarks on.

| Variant | Latency (mean of 3) |
|---|---|
| No compile | **2.279 s** |
| `torch.compile(mode="default", fullgraph=False)` | **1.643 s** |
| Speedup (compile vs eager) | **1.39×** |

- First compiled call (compilation): ~51 s
- Peak memory: 22.1 GB

These match the doc's reported **2.271 s → 1.675 s** almost exactly, so the NVFP4 + `torch.compile` claim holds up. 👍

## Rough edge 1: capability check is more permissive than the shipped kernels

`NunchakuLiteQuantizer.validate_environment` allows INT4 on any GPU with capability `>= (7, 5)` (Turing+). But the kernel build actually fetched for INT4 (`rootonchair/nunchaku-lite-kernels`, `torch212-cxx11-cu126`) only contains SM images for **sm_75 / 80 / 86 / 89**. Verified with `cuobjdump -lelf` on the downloaded `.so`.

Consequence: on an **H100 (sm_90)** the INT4 path **passes** `validate_environment` and then dies at inference with:

```
CUDA error: no kernel image is available for execution on the device
```

which is a confusing failure mode (looks like a runtime bug, not an unsupported-hardware message). So Hopper is effectively unsupported for INT4 today even though the check green-lights it.

Suggestions (pick whatever fits):
- Add sm_90 to the INT4 kernel build, **or**
- Tighten/clarify the check so unsupported archs fail early with a clear message (e.g. "no INT4 kernel image for sm_90"), **or**
- Note the actually-supported archs in the docs.

## Rough edge 2: kernels only ship for torch 2.11 / 2.12

`kernels.get_kernel` matches builds by torch `major.minor`. The repo currently only has builds for **torch 2.11 and 2.12** (cu126/cu128/cu130/cu132). Installing any other torch (I first tried 2.8) fails hard:

```
FileNotFoundError: Cannot find a build variant for this system ...
  torch211-cxx11-cu128-x86_64-linux: Torch version (2.11) does not match environment Torch version (2.8)
  ...
```

For **Blackwell** specifically, the working combo is **torch 2.11 + cu128** (`torch211-cxx11-cu128`).

Suggestion: document the supported torch versions in `nunchaku.md` (and ideally surface a friendlier error / requirement up front), since this isn't obvious and the failure only appears at load time.

## Environment

- HF Jobs `rtx-pro-6000` → NVIDIA RTX PRO 6000 Blackwell, capability (12, 0)
- torch 2.11.0+cu128, diffusers @ PR #14100 head (`452de403`), `kernels` 0.16.x

sayakpaul:
Now running the tests.

sayakpaul:
Testing report (again manually verified).

## Nunchaku Lite (PR #14100) — test run on HF Jobs

Ran the PR's test suite `tests/quantization/nunchaku/test_nunchaku.py` on Hugging Face Jobs.

### Setup

- **Hardware:** HF Jobs `rtx-pro-6000` — RTX PRO 6000 Blackwell, capability (12, 0)
- **Stack:** torch 2.11.0+cu128, kernels 0.16.0, diffusers @ PR #14100 head
- **Env:** `RUN_NIGHTLY=1`, `DIFFUSERS_TEST_DEVICE=cuda` (so nothing is skipped, including the `@nightly` CUDA kernel test)

### Result

**9 passed in 4.66s** ✅ (pytest exit code 0), including `NunchakuLiteCudaKernelsTests::test_awq_cuda_kernels`, which executes the real AWQ W4A16 kernel on GPU.

```
NunchakuLiteBasicTests::test_compact_config_round_trips_dtype_and_targets           PASSED
NunchakuLiteBasicTests::test_environment_requires_cuda                              PASSED
NunchakuLiteBasicTests::test_int4_environment_allows_turing_cuda                    PASSED
NunchakuLiteBasicTests::test_int4_environment_requires_turing_cuda                  PASSED
NunchakuLiteBasicTests::test_nvfp4_environment_allows_blackwell_cuda                PASSED
NunchakuLiteBasicTests::test_nvfp4_environment_requires_blackwell_cuda              PASSED
NunchakuLiteBasicTests::test_compact_config_replaces_svdq_and_awq_without_state_dict PASSED
NunchakuLiteBasicTests::test_nunchaku_lite_loads_with_from_pretrained               PASSED
NunchakuLiteCudaKernelsTests::test_awq_cuda_kernels                                 PASSED
```

### One caveat (not a PR bug)

The 3 tests that download the kernel initially failed due to an external dependency bug: `kernels` passes an empty `user_agent` to `huggingface_hub`, producing a User-Agent header with a trailing space:

```
httpx.LocalProtocolError: Illegal header value b'kernels/0.16.0; hf_hub/1.22.0; python/3.12.12; '
```

`huggingface_hub` 1.x uses httpx, which rejects trailing whitespace. It only triggers on a cold first `get_kernel` call. Worked around by stripping the UA (a small `sitecustomize.py` patching `huggingface_hub.utils._headers`). Pinning `huggingface_hub<1.0` does **not** help — it drags `kernels` down to 0.12.3, which lacks the `trust_remote_code=` argument the PR uses.

Suggestion: report to `kernels` upstream, and consider bumping the PR's `require_kernels_version_greater_or_equal("0.9.0")` to a version that actually provides `trust_remote_code`.

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
rootonchair: yep

docs/source/en/quantization/nunchaku.md
rootonchair: Not only the repo, I will reference their paper too

rootonchair: this is resolved

docs/source/en/quantization/nunchaku.md
rootonchair: I have updated the details

docs/source/en/quantization/overview.md
```diff
@@ -28,7 +28,7 @@ There are two ways to use [`~quantizers.PipelineQuantizationConfig`] depending o
 
 Initialize [`~quantizers.PipelineQuantizationConfig`] with the following parameters.
 
-- `quant_backend` specifies which quantization backend to use. Currently supported backends include: `bitsandbytes_4bit`, `bitsandbytes_8bit`, `gguf`, `quanto`, and `torchao`.
+- `quant_backend` specifies which quantization backend to use. Currently supported backends include: `bitsandbytes_4bit`, `bitsandbytes_8bit`, `gguf`, `nunchaku_lite`, `quanto`, and `torchao`. Nunchaku Lite is used for loading prequantized checkpoints; see the [Nunchaku Lite](./nunchaku) guide for checkpoint requirements.
```
rootonchair:
```suggestion
- `quant_backend` specifies which quantization backend to use. Currently supported backends include: `bitsandbytes_4bit`, `bitsandbytes_8bit`, `gguf`, `nunchaku_lite`, `quanto`, and `torchao`.
```

src/diffusers/quantizers/nunchaku/nunchaku_quantizer.py
rootonchair: Hopper validation added

docs/source/en/quantization/nunchaku.md
rootonchair: with the second kernel version, this has been resolved

docs/source/en/quantization/nunchaku.md
rootonchair: yes it is compatible with `enable_sequential_cpu_offload` and `enable_model_cpu_offload`

sayakpaul (approved):
Thanks a lot for this!

sayakpaul:
@bot /style

docs/source/en/quantization/overview.md
```diff
@@ -28,7 +28,7 @@ There are two ways to use [`~quantizers.PipelineQuantizationConfig`] depending o
 
 Initialize [`~quantizers.PipelineQuantizationConfig`] with the following parameters.
 
-- `quant_backend` specifies which quantization backend to use. Currently supported backends include: `bitsandbytes_4bit`, `bitsandbytes_8bit`, `gguf`, `quanto`, and `torchao`.
+- `quant_backend` specifies which quantization backend to use. Currently supported backends include: `bitsandbytes_4bit`, `bitsandbytes_8bit`, `gguf`, `nunchaku_lite`, `quanto`, and `torchao`.
```
stevhliu: seems like `nunchaku_lite` is more of a prequantized loader versus a backend that quantizes pretrained/non-quantized checkpoints on the fly which is what this section mainly demonstrates. it would be better to add a new "Loading prequantized checkpoints" section with this there

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+
+Nunchaku Lite kernels and quantized linear layers are compatible with [`torch.compile`](../optimization/fp16#torchcompile).
+Compile the quantized transformer after loading the pipeline for faster inference.
+
+```python
+pipe.transformer = torch.compile(pipe.transformer, mode="default", fullgraph=True)
+```
+
+An ERNIE-Image-Turbo benchmark on an RTX PRO 6000 reported that Nunchaku Lite NVFP4 with `torch.compile` reduced the full pipeline latency from 2.271s to 1.675s. Compared to the original BF16 pipeline, the compiled
+Nunchaku Lite NVFP4 pipeline reached a 1.8x speedup.
```
stevhliu: front-load the benefits

```suggestion
The compiled Nunchaku Lite NVFP4 pipeline runs 1.8x faster than the original BF16 pipeline (2.271s → 1.675s on an RTX PRO 6000).
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+).to("cuda")
+
+prompt = "A modern red armchair in a quiet studio, soft window light, realistic product photography"
+image = pipe(
+    prompt=prompt,
+    height=1024,
+    width=1024,
+    num_inference_steps=8,
+    guidance_scale=1.0,
+    use_pe=False,
```
stevhliu: may be helpful to either remove or very briefly explain `use_pe` is model-specific so users copying the snippet for another model won't get confused

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+## Using Optimized CUDA Kernels with Nunchaku Lite
+
+Nunchaku Lite uses optimized CUDA kernels through the `kernels` package. The kernels are loaded automatically when you
+load a Nunchaku Lite checkpoint. Install the runtime dependency before loading a Nunchaku Lite checkpoint.
+
+```bash
+pip install -U kernels
+```
```
stevhliu: move the installation before the first code snippet which needs it

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+
+> [!NOTE]
```
stevhliu, line 56: would move this to the loading section where architecture matching is more relevant versus under CUDA kernels

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+specific language governing permissions and limitations under the License.
+
+-->
+
+# Nunchaku Lite
+
+Nunchaku Lite is a quantization backend for loading prequantized checkpoints in Diffusers. Use
+[`diffuse-compressor`](https://github.com/rootonchair/diffuse-compressor) to create compatible checkpoints by choosing
+or adapting a target config for the model architecture, quantizing and exporting the transformer, and packaging the
+result as a Diffusers pipeline repository.
```
stevhliu:
```suggestion
Nunchaku Lite is a quantization backend for loading prequantized checkpoints in Diffusers. Create compatible checkpoints with [diffuse-compressor](https://github.com/rootonchair/diffuse-compressor). It quantizes and exports a transformer, then packages it as a Diffusers pipeline.
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+
+| Method | Precision | Group size | Notes |
+|---|---:|---:|---|
+| `svdq_w4a4` | `nvfp4` | 16 | Uses NVFP4 runtime kernels with SVDQ low-rank correction. |
+| `svdq_w4a4` | `int4` | 64 | Uses INT4 W4A4 kernels with SVDQ low-rank correction. |
+| `awq_w4a16` | `int4` | 64 | Uses INT4 weight-only AWQ-style kernels. |
+
+## NunchakuLiteQuantizationConfig
+
+The `config.json` file must include a compact [`NunchakuLiteQuantizationConfig`]. It defines the runtime
```
stevhliu:
```suggestion
The `config.json` file must include a [`NunchakuLiteQuantizationConfig`]. It defines the runtime
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+    guidance_scale=1.0,
+    use_pe=False,
+).images[0]
+image.save("ernie-image-turbo-nunchaku-lite.png")
+```
+
+## Using Optimized CUDA Kernels with Nunchaku Lite
+
+Nunchaku Lite uses optimized CUDA kernels through the `kernels` package. The kernels are loaded automatically when you
+load a Nunchaku Lite checkpoint. Install the runtime dependency before loading a Nunchaku Lite checkpoint.
```
stevhliu:
```suggestion
The kernels package supplies the optimized CUDA kernels, which load automatically. Install it first.
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+
+The `config.json` file must include a compact [`NunchakuLiteQuantizationConfig`]. It defines the runtime
+`compute_dtype` and the target modules for each Nunchaku Lite quantization method.
+
+- `compute_dtype`: runtime dtype for floating-point buffers in quantized modules, typically `torch.bfloat16`.
+- `svdq_w4a4`: SVDQ W4A4 target config with `precision`, `group_size`, `rank`, and `targets`.
+- `awq_w4a16`: AWQ W4A16 target config with `precision`, `group_size`, and `targets`.
+
+Each entry in `targets` must point to a linear layer. Diffusers replaces the linear layers listed under `svdq_w4a4`
+with SVDQ W4A4 layers and the linear layers listed under `awq_w4a16` with AWQ W4A16 layers. The example below shows the
```
stevhliu:
```suggestion
Each entry in `targets` must point to a linear layer. Diffusers swaps each `svdq_w4a4` target for an SVDQ W4A4 layer and each `awq_w4a16` target for an AWQ W4A16 layer. The example below shows the
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+expected shape with shortened target lists.
+
+> [!NOTE]
+> A target module can only use one Nunchaku Lite quantization method. Do not list the same target under both
+> `svdq_w4a4` and `awq_w4a16`.
```
stevhliu:
```suggestion
List each module you want to quantize under `svdq_w4a4` or `awq_w4a16`. A module can only use one method, so don't list the same target under both.
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+    height=1024,
+    width=1024,
+    num_inference_steps=8,
+    guidance_scale=1.0,
+    use_pe=False,
+).images[0]
+image.save("ernie-image-turbo-nunchaku-lite.png")
+```
+
+## Using Optimized CUDA Kernels with Nunchaku Lite
```
stevhliu:
```suggestion
## Install the CUDA kernels
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+```bash
+pip install -U kernels
+```
+
+> [!NOTE]
+> The exported state dict must match the target Diffusers model architecture exactly. For example, a checkpoint
+> quantized with fused QKV projections won't load into a model config that expects separate Q, K, and V projection
+> modules.
+
+## Supported Quantization Types
```
stevhliu:
```suggestion
## Supported quantization types
```

docs/source/en/quantization/nunchaku.md
```diff
@@ -0,0 +1,146 @@
...
+result as a Diffusers pipeline repository.
+
+Nunchaku Lite builds on the original [Nunchaku](https://github.com/nunchaku-ai/nunchaku) inference engine,
+[DeepCompressor](https://github.com/nunchaku-ai/deepcompressor) quantization library, and
+[SVDQuant paper](https://arxiv.org/abs/2411.05007).
+
+## Load a quantized pipeline
+
+Load the packaged prequantized pipeline with [`~DiffusionPipeline.from_pretrained`]. Diffusers reads the quantization
+config from `config.json`.
```
stevhliu:
```suggestion
Load the prequantized pipeline with [`~DiffusionPipeline.from_pretrained`], which reads the quantization
config from `config.json`.
```

stevhliu (approved):
thanks for the docs!

sayakpaul:
@rootonchair there are some CI failures. Would it be possible for you to fix them? The PR is basically ready to be merged.

docs/source/en/quantization/nunchaku.md
rootonchair:
```suggestion
```
let's remove it

rootonchair:
@sayakpaul @stevhliu I just updated the docs and the unit test, could you check again

sayakpaul (approved):

PR #14105, opened 2026-07-01
feat: auto-detect and fuse QKV for original PEFT adapters

### LINKED ISSUE #14002: Support Original Checkpoint-Compatible PEFT Adapters
My understanding is that `diffusers` does not currently support original checkpoint-compatible PEFT adapters aside from LoRAs.

I think `diffusers` should support loading general `peft`-compatible checkpoints which are compatible with the original checkpoint structure.

In particular, my proposal is that it could be supported by the following strategy:
1. Make the `diffusers` checkpoint and PEFT checkpoint weights coincide.
2. Inject the checkpoint normally using `peft`.

See https://github.com/huggingface/diffusers/pull/13861#issuecomment-4756872880 for a slightly more fleshed out version of the above. 

`diffusers` supports LoRA checkpoints by converting the checkpoints to the `diffusers` format. However, differences between `diffusers` checkpoints and potential LoRA checkpoints, most notably split Q,K,V projections in `diffusers` vs fused QKV projections in LoRA checkpoints, means that the adapter checkpoint weights need to be modified to make this possible. In general, if weight shape differences are present, each new PEFT technique (e.g. DoRA, IA3, etc.) would need their own technique-specific conversion logic. So, in particular, for (1) in the proposal above we should make the `diffusers` checkpoint match the original checkpoint in terms of weight shapes (for example, by fusing the QKV projections) to avoid the need for conversion logic for each PEFT technique.

**Additional context.**

Saurav-Gupta-9741:
Fixes #14002 
Hi @dg845 Sir! I saw your comment regarding the lack of support for loading generalized, original-checkpoint PEFT adapters without relying on conversion logic. 
To solve this, I added dynamic auto-detection to `PeftAdapterMixin.load_lora_adapter`. When an adapter is passed in (like DoRA, IA3, or standard LoRA), the code scans the state dictionary. If it expects a fused projection (e.g., `to_qkv`), it intercepts the load, checks if the model supports fusion, and automatically runs `self.fuse_qkv_projections()`. 
This guarantees that the Diffusers checkpoint weights natively coincide with the PEFT weights, allowing `peft` to inject it seamlessly without shape mismatch errors or giant conversion scripts! 
I ran the `tests/lora` suite and locally tested a mock fused adapter to verify that the PyTorch shapes align perfectly after the interception. Let me know what you think!
Looking forward for your Response

### DIFF
--- a/src/diffusers/loaders/peft.py
+++ b/src/diffusers/loaders/peft.py
@@ -225,6 +225,12 @@ class PeftAdapterMixin:
             if is_sai_sd_control_lora:
                 state_dict = convert_sai_sd_control_lora_state_dict_to_peft(state_dict)
 
+            # Auto-detect if the incoming state_dict expects fused QKV projections
+            # and fuse them in the model if the model supports it.
+            has_fused_qkv = any("to_qkv" in k for k in state_dict.keys())
+            if has_fused_qkv and hasattr(self, "fuse_qkv_projections"):
+                self.fuse_qkv_projections()
+
             rank = {}
             for key, val in state_dict.items():
                 # Cannot figure out rank from lora layers that don't have at least 2 dimensions.


Saurav-Gupta-9741:
Hi @sayakpaul Sir and @yiyixuxu Ma'am, I hope you are both doing well!

I submitted this PR to add auto-detection and QKV fusion for original PEFT adapters. It has been a little while since I opened it and it hasn't received a review yet. I understand you are both extremely busy, but I would be incredibly grateful if one of you could at least share a brief response or take a quick look whenever you have a moment.

Please let me know if there are any changes or improvements you would like me to make. I am more than happy to iterate on this to ensure it perfectly matches the diffusers architecture!

Looking forward to your response. Thank you both for your time.

Issue #14097, opened 2026-07-01
Regression (#13485) Broken TorchAO Compat


Beinsezii:
### Describe the bug

Some changes introduced in https://github.com/huggingface/diffusers/pull/13485 / https://github.com/huggingface/diffusers/commit/236e5dd9f38e21ae40c002539368b9be9a5e0fc8 have regressed compat with compiled 8-bit models through TorchAO.

### Reproduction

Simply `./regression.py`. Tested on 1xH200. Switch to previous commit `rev = "6d9331ea607183f84a21bdc37da6389a611fe7bd"` to observe it functional.
```py
#! /usr/bin/env -S uv run --script
# /// script
# requires-python = "==3.12.13"
# dependencies = [
#     "accelerate==1.14.0",
#     "diffusers",
#     "torch==2.12.1",
#     "torchao==0.17.0",
#     "transformers==5.5.4",
# ]
#
# [[tool.uv.index]]
# url = "https://download.pytorch.org/whl/cu132"
#
# [tool.uv.sources]
# diffusers = { git = "https://github.com/huggingface/diffusers.git", rev = "236e5dd9f38e21ae40c002539368b9be9a5e0fc8" }
# ///

import torch
from diffusers import AutoModel, Flux2KleinPipeline, TorchAoConfig
from torchao.quantization.quant_api import Int8DynamicActivationInt8WeightConfig

with torch.no_grad():
    device = "cuda"
    dtype = torch.bfloat16

    pipe = Flux2KleinPipeline.from_pretrained(
        "black-forest-labs/FLUX.2-klein-9B",
        transformer=torch.compile(
            AutoModel.from_pretrained(
                "black-forest-labs/FLUX.2-klein-9B",
                subfolder="transformer",
                torch_dtype=dtype,
                device_map=device,
                quantization_config=TorchAoConfig(
                    Int8DynamicActivationInt8WeightConfig(version=2),
                    modules_to_not_convert=[
                        "pos_embed",
                        "time_guidance_embed",
                        "double_stream_modulation_img",
                        "double_stream_modulation_txt",
                        "single_stream_modulation",
                        "x_embedder",
                        "context_embedder",
                        "norm_out",
                        "proj_out",
                    ],
                ),
            ),
            mode="max-autotune",
            fullgraph=True,
            dynamic=True,
        ),
        torch_dtype=dtype,
        device_map=device,
    )

    prompt = "A cat holding a sign that says hello world"
    image = pipe(
[...]

PR #14107, opened 2026-07-02
[Restore] torch.chunk in Flux2SwiGLU from #13485

JingyaHuang:
# What does this PR do?

Reverts the manual half-slicing workaround in `Flux2SwiGLU.forward` back to `torch.chunk` to fix the regression reported in #14097. The workaround was only added to support the Neuron `torch.compile` backend, which no longer requires it in their latest beta, so this change carries no downside for Neuron integration.


Beinsezii (approved):
confirmed on my end in real env

PR #14113, opened 2026-07-03
[tests] Pipeline test refactor

sayakpaul:
## Notes

* The PR doesn't have LoRA-related tests refactored yet. That will be a big change, and I want to do it in a different PR to keep the reviewing scope manageable.
* I haven't yet added a `generate_pipeline_level_tests.py` script. Will be done in subsequent PRs once other mixins, such as LoRA, are added.
* There were a lot of overlaps between the pipeline-level and model-level quantization tests. So, I have removed redundancy as much as possible by keeping only the pipeline-level quantization tests that genuinely require pipeline-level specs.
* I have used `pytest` features as much as possible. This is particularly reflected in the use of custom fixtures. For example, `base_pipe_output`. This way, we don't have to compute it for every test that requires it.

I have only refactored the Flux pipeline tests to use the new mixin, as we do in other PRs. Will open subsequent PRs for other pipelines and involve the community.


sayakpaul:
/diffusers-bot pytest tests/pipelines/flux/test_pipeline_flux.py::TestFluxPipeline::test_float16_inference

tests/pipelines/test_pipeline_utils.py
```diff
@@ -1,9 +1,15 @@
 import contextlib
```
sayakpaul, line 1: Changes here are migrating to `pytest`.

tests/pipelines/test_pipelines_common.py
```diff
@@ -2507,146 +2501,6 @@ def test_pipeline_level_group_offloading_inference(self, expected_max_difference
         self.assertLess(max_diff, expected_max_difference)
 
 
-@is_staging_test
-class PipelinePushToHubTester(unittest.TestCase):
```
sayakpaul, line 2489: Moved to tests/pipelines/test_pipeline_utils.py

sayakpaul:
All the failing tests are quite unrelated

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+    `get_dummy_components()`, `get_dummy_inputs()` and the shared fixtures).
+    """
+
+    def test_save_load_local(self, tmp_path, base_pipe_output, expected_max_difference=5e-4):
+        components = self.get_dummy_components()
+        for key in components:
+            if "text_encoder" in key and hasattr(components[key], "eval"):
+                components[key].eval()
+        pipe = self.pipeline_class(**components)
+        for component in pipe.components.values():
```
DN6: This section is legacy code when we supported attn processors that didn't use SDPA. We can remove this part since our minimum torch version is now `>=2.6`

tests/pipelines/testing_utils/utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+    return tensor
+
+
+def assert_outputs_close(actual, expected, atol=1e-4, rtol=0.0, msg=""):
+    """
+    `assert_tensors_close` for pipeline outputs, which are usually numpy arrays (`output_type="np"`). Mirrors the
+    model-level assertion style (concise diff messages) while accepting numpy/torch outputs.
+    """
+    assert_tensors_close(
```
DN6: Let's just use this directly in the tests?

tests/pipelines/testing_utils/utils.py
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+
+import torch
+
+from diffusers.models.attention import AttentionModuleMixin
+
+from ...testing_utils import assert_tensors_close
+
+
+def to_np(tensor):
```
DN6: Let's use torch tensors exclusively in tests and avoid the overheard to converting to numpy. We don't gain much by maintaining two output types (unless we're explicitly testing for output type).

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+        init_components = {k: v for k, v in init_components.items() if not isinstance(v, (str, int, float))}
+
+        pipe = self.pipeline_class(**init_components)
+
+        assert hasattr(pipe, "components")
+        assert set(pipe.components.keys()) == set(init_components.keys())
+
+    @pytest.mark.skipif(torch_device not in ["cuda", "xpu"], reason="float16 requires CUDA or XPU")
+    @require_accelerator
+    def test_float16_inference(self, expected_max_diff=5e-2):
```
DN6: I'm actually wondering about the relevance of this test. Default `dtype` of models is usually `bf16` or `fp16` and it's not very common to run `fp32` inference. Rather than comparing the outputs to the fp32 output, which doesn't provide us with much signal, I think it is better to run the inference in fp16/bf16 and verify that there are no nans.

tests/pipelines/flux/test_pipeline_flux.py
```diff
@@ -160,7 +145,7 @@ def test_flux_different_prompts(self):
 
         # Outputs should be different here
         # For some reasons, they don't show large differences
-        self.assertGreater(max_diff, 1e-6, "Outputs should be different for different prompts.")
+        assert max_diff > 1e-6, "Outputs should be different for different prompts."
 
     def test_fused_qkv_projections(self):
```
DN6: I think we can remove this test since it is essentially just checking if we can run a forward pass through the denoiser with fused qkv. We already test this at the model level. We don't get any additional signal from this.

tests/pipelines/flux/test_pipeline_flux.py
```diff
@@ -230,8 +209,8 @@ def test_flux_true_cfg(self):
         inputs["negative_prompt"] = "bad quality"
         inputs["true_cfg_scale"] = 2.0
         true_cfg_out = pipe(**inputs, generator=torch.manual_seed(0)).images[0]
-        self.assertFalse(
-            np.allclose(no_true_cfg_out, true_cfg_out), "Outputs should be different when true_cfg_scale is set."
+        assert not np.allclose(no_true_cfg_out, true_cfg_out), (
```
DN6: Let's use torch tensors throughout the test suite.

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+            "latents",
+            "output_type",
+            "return_dict",
+        ]
+    )
+
+    # set these to False/True in the child class if the pipeline does not support the corresponding functionality
+    test_attention_slicing = True
+    test_layerwise_casting = False
+    test_group_offloading = False
```
DN6: Let's remove these and skip via checks either within the relevant feature mixins or individual tests. Also attention slicing is pretty outdated at this point. We can completely remove it I think.

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+    The class also exposes the shared pytest fixtures used across the tester mixins, most notably
+    `base_pipe_output` which holds the output of a freshly constructed pipeline run on the standard dummy inputs
+    (computed once per test class and reused by comparison tests). Comparison tests construct their own pipeline
+    inline (mirroring the model-level testers) so the only difference from `base_pipe_output` is the behavior
+    under test.
+    """
+
+    # Canonical parameters that are passed to `__call__` regardless of the type of pipeline. They are always
+    # optional and have common sense default values.
+    required_optional_params = frozenset(
```
DN6: Nit. This legacy naming feels off to me (how can something be required and optional). Additionally, let's explicitly refer to these as input params. IMO naming it `optional_input_params` is better.

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+    @property
+    def params(self) -> frozenset:
+        raise NotImplementedError(
+            "You need to set the attribute `params` in the child test class. "
+            "`params` are checked for if all values are present in `__call__`'s signature. "
+            "You can set `params` using one of the common set of parameters defined in `pipeline_params.py`."
+        )
+
+    @property
+    def batch_params(self) -> frozenset:
```
DN6: Nit. Would rename this to `batch_input_params`

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+        torch.compiler.reset()
+        gc.collect()
+        backend_empty_cache(torch_device)
+        yield
+        torch.compiler.reset()
+        gc.collect()
+        backend_empty_cache(torch_device)
+
+    @pytest.fixture(scope="class")
+    def base_pipe_output(self, request):
```
DN6: Why not use the same output caching system as the model tests? Why couple to a config class?

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+        )
+
+    def get_dummy_inputs(self, device, seed=0):
+        raise NotImplementedError(
+            "You need to implement `get_dummy_inputs(self, device, seed)` in the child test class. "
+            "See existing pipeline tests for reference."
+        )
+
+    @property
+    def params(self) -> frozenset:
```
DN6: Nit. Would rename this to `required_input_params_in_call_signature` to be more explicit about it's meaning.

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+            "See existing pipeline tests for reference."
+        )
+
+    def get_dummy_components(self):
+        raise NotImplementedError(
+            "You need to implement `get_dummy_components(self)` in the child test class. "
+            "See existing pipeline tests for reference."
+        )
+
+    def get_dummy_inputs(self, device, seed=0):
```
DN6: Do we need to pass device and seed here? Can't we use torch_device and just set the seed to 0 always? Or use the `get_generator` method?

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+        pipe.set_progress_bar_config(disable=None)
+
+        model_dtypes = [component.dtype for component in components.values() if getattr(component, "dtype", None)]
+        assert all(dtype == torch.float32 for dtype in model_dtypes)
+
+        pipe.to(dtype=torch.float16)
+        model_dtypes = [component.dtype for component in components.values() if getattr(component, "dtype", None)]
+        assert all(dtype == torch.float16 for dtype in model_dtypes)
+
+    def test_attention_slicing_forward_pass(self, expected_max_diff=1e-3):
```
DN6: Think we can remove this test. It's an outdated way to save memory and we have better features for that funcition.

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,954 @@
...
+
+    # ==================== Fixtures ====================
+
+    @pytest.fixture(autouse=True)
+    def cleanup(self):
+        """Skip deprecated pipelines and free VRAM before/after each test (replaces unittest setUp/tearDown)."""
+        from diffusers.pipelines.pipeline_utils import DeprecatedPipelineMixin
+
+        if issubclass(self.pipeline_class, DeprecatedPipelineMixin):
+            pytest.skip(reason=f"Deprecated Pipeline: {self.pipeline_class.__name__}")
```
DN6: Let's move this section into it's own fixture 

```python
@pytest.fixture(autouse=True)
def skip_if_deprecated(self):
    from diffusers.pipelines.pipeline_utils import DeprecatedPipelineMixin
    if issubclass(self.pipeline_class, DeprecatedPipelineMixin):
        pytest.skip(f"Deprecated pipeline: {self.pipeline_class.__name__}")
```

tests/quantization/bnb/test_4bit.py
```diff
@@ -82,7 +81,7 @@ def get_some_linear_layer(model):
 @require_torch
 @require_torch_accelerator
 @slow
-class Base4bitTests(unittest.TestCase):
+class Base4bitTests:
```
DN6, line 84: For pipeline quantization tests, why not use the Mixin approach that we have with models? Or is that planned for a follow up?

tests/pipelines/testing_utils/common.py
sayakpaul: I am just a porter here :D Didn't reason about it during the porting.

tests/quantization/bnb/test_4bit.py
sayakpaul: Followup.

tests/pipelines/testing_utils/common.py
sayakpaul: Good point. Addressed.

tests/quantization/quanto/test_quanto.py
```diff
@@ -1,31 +1,26 @@
 import gc
-import tempfile
-import unittest
+
+import pytest
 
 from diffusers import FluxPipeline, FluxTransformer2DModel, QuantoConfig
-from diffusers.models.attention_processor import Attention
-from diffusers.utils import is_optimum_quanto_available, is_torch_available
+from diffusers.utils import is_torch_available
```
DN6, line 6: Think we can just remove the quanto test entirely and fully deprecate support. Can be done in a separate PR, but just flagging.

tests/quantization/modelopt/test_modelopt.py
```diff
@@ -245,17 +247,17 @@ def test_training(self):
 
         for module in quantized_model.modules():
             if isinstance(module, LoRALayer):
-                self.assertTrue(module.adapter[1].weight.grad is not None)
+                assert module.adapter[1].weight.grad is not None
 
 
-class SanaTransformerFP8WeightsTest(ModelOptBaseTesterMixin, unittest.TestCase):
+class TestSanaTransformerFP8Weights(ModelOptBaseTesterMixin):
```
DN6, line 256: These would all be removed yes? Because they are model level tests, and should actually be run through the sana model test file?

tests/pipelines/testing_utils/ip_adapter.py
```diff
@@ -0,0 +1,149 @@
...
+        # forward pass without ip adapter
+        inputs = self._modify_inputs_for_ip_adapter_test(self.get_dummy_inputs())
+        if expected_pipe_slice is None:
+            output_without_adapter = pipe(**inputs)[0]
+        else:
+            output_without_adapter = expected_pipe_slice
+
+        # 1. Single IP-Adapter test cases
+        adapter_state_dict = create_flux_ip_adapter_state_dict(pipe.transformer)
+        pipe.transformer._load_ip_adapter_weights(adapter_state_dict)
```
DN6: Should load via the pipeline ip adapter method.

tests/pipelines/testing_utils/ip_adapter.py
```diff
@@ -0,0 +1,149 @@
...
+import torch
+
+from diffusers.loaders import FluxIPAdapterMixin
+
+from ...models.transformers.test_models_transformer_flux import create_flux_ip_adapter_state_dict
+from ...testing_utils import assert_tensors_close, is_ip_adapter, torch_device
+
+
+@is_ip_adapter
+class FluxIPAdapterTesterMixin:
```
DN6: This is pipeline specific no? We should either turn this into a generic IPAdapterMixin or move it into the flux pipeline test file. 

Model class example:
https://github.com/huggingface/diffusers/blob/284419b35267dea7d4435fd0ef1eaa132bf73efe/tests/models/testing_utils/ip_adapter.py#L41

And usage:

https://github.com/huggingface/diffusers/blob/284419b35267dea7d4435fd0ef1eaa132bf73efe/tests/models/transformers/test_models_transformer_flux.py#L263

https://github.com/huggingface/diffusers/blob/284419b35267dea7d4435fd0ef1eaa132bf73efe/tests/models/unets/test_models_unet_2d_condition.py#L957

tests/pipelines/testing_utils/ip_adapter.py
```diff
@@ -0,0 +1,149 @@
...
+    def _modify_inputs_for_ip_adapter_test(self, inputs: Dict[str, Any]):
+        inputs["negative_prompt"] = ""
+        if "true_cfg_scale" in inspect.signature(self.pipeline_class.__call__).parameters:
+            inputs["true_cfg_scale"] = 4.0
+        # Request torch outputs so comparisons run on torch tensors directly (see `BasePipelineTesterConfig`).
+        inputs["output_type"] = "pt"
+        inputs["return_dict"] = False
+        return inputs
+
+    def test_ip_adapter(self, expected_max_diff: float = 1e-4, expected_pipe_slice=None):
```
DN6: I think this was copied over from the existing test, but it would better to run each test case as an individual test.

tests/pipelines/testing_utils/cache.py
```diff
@@ -0,0 +1,379 @@
...
+            dummy_component_kwargs["num_layers"] = num_layers
+        if "num_single_layers" in dummy_component_parameters:
+            num_single_layers = 2
+            dummy_component_kwargs["num_single_layers"] = num_single_layers
+
+        components = self.get_dummy_components(**dummy_component_kwargs)
+        pipe = self.pipeline_class(**components)
+        pipe.set_progress_bar_config(disable=None)
+
+        self.pab_config.current_timestep_callback = lambda: pipe.current_timestep
```
DN6: Since `pab_config` is a class attribute, this callback would persist across tests.  At the model level, we recreate cache config instances per test:
https://github.com/huggingface/diffusers/blob/284419b35267dea7d4435fd0ef1eaa132bf73efe/tests/models/testing_utils/cache.py#L81

Can reuse the model cacheing test design:
https://github.com/huggingface/diffusers/blob/284419b35267dea7d4435fd0ef1eaa132bf73efe/tests/models/testing_utils/cache.py#L251-L277

tests/pipelines/testing_utils/cache.py
```diff
@@ -0,0 +1,379 @@
...
+
+    def test_pyramid_attention_broadcast_inference(self, expected_atol: float = 0.2):
+        # We need to use higher tolerance because we are using a random model. With a converged/trained model, the
+        # tolerance can be lower.
+        device = "cpu"  # ensure determinism for the device-dependent torch.Generator
+        num_layers = 2
+        components = self.get_dummy_components(num_layers=num_layers)
+        for key in components:
+            if "text_encoder" in key and hasattr(components[key], "eval"):
+                components[key].eval()
```
DN6: Why do we need to set the TE to eval mode?

tests/pipelines/testing_utils/cache.py
```diff
@@ -0,0 +1,379 @@
...
+        pipe = self.pipeline_class(**components)
+        pipe = pipe.to(device)
+        pipe.set_progress_bar_config(disable=None)
+
+        # Run inference without PAB
+        inputs = self.get_dummy_inputs()
+        inputs["num_inference_steps"] = 4
+        output = pipe(**inputs)[0]
+        original_image_slice = output.flatten()
+        original_image_slice = torch.cat((original_image_slice[:8], original_image_slice[-8:]))
```
DN6: Base pipeline output can be used here no?

tests/pipelines/testing_utils/cache.py
```diff
@@ -0,0 +1,379 @@
...
+@is_cache
+class FasterCacheTesterMixin(CacheTesterMixin):
+    faster_cache_config = FasterCacheConfig(
+        spatial_attention_block_skip_range=2,
+        spatial_attention_timestep_skip_range=(-1, 901),
+        unconditional_batch_skip_range=2,
+        attention_weight_callback=lambda _: 0.5,
+    )
+
+    def test_faster_cache_basic_warning_or_errors_raised(self):
```
DN6: This is mostly testing the FasterCacheConfig object that the pipeline + cacheing. I think we can remove it from here.

tests/pipelines/testing_utils/cache.py
```diff
@@ -0,0 +1,379 @@
...
+        for name, module in denoiser.named_modules():
+            if hasattr(module, "_diffusers_hook"):
+                if name == "":
+                    # Skip the root denoiser module
+                    continue
+                count += 1
+                assert isinstance(module._diffusers_hook.get_hook(_FASTER_CACHE_BLOCK_HOOK), FasterCacheBlockHook), (
+                    "Hook should be of type FasterCacheBlockHook."
+                )
+        assert count == expected_hooks, "Number of hooks should match expected number."
```
DN6: This part if more of a model level test. I would move it to the model level mixin.

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,878 @@
...
+                for name, buf in module.named_buffers():
+                    if not buf.is_floating_point():
+                        buf.data = buf.data.to(torch_device)
+                    elif any(
+                        module_to_keep_in_fp32 in name.split(".")
+                        for module_to_keep_in_fp32 in module._keep_in_fp32_modules
+                    ):
+                        buf.data = buf.data.to(torch_device).to(torch.float32)
+                    else:
+                        buf.data = buf.data.to(torch_device).to(torch.float16)
```
DN6, line 404: We can probably move this into a utilty function.

tests/pipelines/testing_utils/ip_adapter.py
sayakpaul: Didn't get your point.

tests/pipelines/testing_utils/common.py
sayakpaul: Keeping it inline for better readability.

sayakpaul:
@DN6 I believe I have resolved all your comments. PTAL.

tests/pipelines/testing_utils/utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+from diffusers.models.attention import AttentionModuleMixin
+
+
+def check_same_shape(tensor_list):
+    shapes = [tensor.shape for tensor in tensor_list]
+    return all(shape == shapes[0] for shape in shapes[1:])
+
+
+def check_qkv_fusion_matches_attn_procs_length(model, original_attn_processors):
```
DN6, line 32: Okay to have these here. But my preference would be to remove these checks from pipeline level testing. This can be in a follow up after all pipeline tests are migrated to the new format.

tests/pipelines/testing_utils/memory.py
```diff
@@ -0,0 +1,444 @@
...
+        if denoiser is None or not hasattr(denoiser, "enable_layerwise_casting"):
+            pytest.skip(f"{self.pipeline_class.__name__} has no denoiser that supports layerwise casting.")
+
+        pipe.to(torch_device, dtype=torch.bfloat16)
+        pipe.set_progress_bar_config(disable=None)
+
+        denoiser.enable_layerwise_casting(storage_dtype=torch.float8_e4m3fn, compute_dtype=torch.bfloat16)
+
+        inputs = self.get_dummy_inputs()
+        _ = pipe(**inputs)[0]
```
DN6: Think we can just test that the output is not NaN here.

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,878 @@
...
+
+        pipe.to(dtype=torch.float16)
+        model_dtypes = [component.dtype for component in components.values() if getattr(component, "dtype", None)]
+        assert all(dtype == torch.float16 for dtype in model_dtypes)
+
+    def test_num_images_per_prompt(self):
+        sig = inspect.signature(self.pipeline_class.__call__)
+
+        if "num_images_per_prompt" not in sig.parameters:
+            return
```
DN6: Should be `pytest.skip` here I think.

tests/pipelines/testing_utils/common.py
```diff
@@ -0,0 +1,878 @@
...
+    Common tests for each PyTorch pipeline: saving and loading, equivalence of dict and tuple outputs, batching,
+    dtype/device handling, callbacks, and variants.
+
+    Designed to be composed with `BasePipelineTesterConfig` (which provides `pipeline_class`,
+    `get_dummy_components()`, `get_dummy_inputs()` and the shared fixtures).
+    """
+
+    def test_save_load_local(self, tmp_path, base_pipe_output, expected_max_difference=5e-4):
+        components = self.get_dummy_components()
+        for key in components:
```
DN6: Rather than applying this fix to individual tests, why not add a util to this mixin to fetch the built pipe and apply it once there. Also makes sure all components in the pipe are in eval mode. 

```python
def get_pipeline(self, **components):
    components = components or self.get_dummy_components()
    for m in pipe.components.values():
        if hasattr(m, "eval"):
            m.eval()

    return pipe
```

DN6 (approved):
LGTM. Some minor additions requested.

tests/pipelines/testing_utils/utils.py
sayakpaul: We should then ensure if these are included in the model level testing? I have included a TODO for that as a comment.

tests/quantization/quanto/test_quanto.py
sayakpaul: https://github.com/huggingface/diffusers/pull/14254

tests/quantization/modelopt/test_modelopt.py
sayakpaul: https://github.com/huggingface/diffusers/pull/14256/

PR #14118, opened 2026-07-03
align snapshot_download to respect hfh latest version

### LINKED ISSUE #14117: `DiffusionPipeline.download()` breaks with huggingface_hub>=1.22.0 in offline mode (IncompleteSnapshotError)
## Summary

`huggingface_hub` v1.22.0 introduced a breaking change: `snapshot_download(..., local_files_only=True)` now verifies that the cached snapshot is **complete** (it caches the `/tree` call, so even offline it can tell whether the cached folder is partial). Previously, in offline mode it returned whatever snapshot folder it found in the cache without checking for completeness.

This breaks `DiffusionPipeline.download()` when a pipeline is loaded in offline mode after being downloaded (partially) once.

## Mechanism

In `DiffusionPipeline.download()`:

1. The first (online) call downloads with `snapshot_download(..., allow_patterns=<computed patterns>)`, which deliberately skips files like `.gitattributes` and `README.md`.
2. When later run in offline mode, `model_info()` raises `OfflineModeIsEnabled`, so `local_files_only=True` is set and the `if not local_files_only:` block that computes `allow_patterns` is **skipped** → `allow_patterns` stays `None`.
3. It then calls `snapshot_download(..., allow_patterns=None, local_files_only=True)`. With `allow_patterns=None`, hfh assumes the full repo is wanted, sees the cached folder is incomplete (missing `.gitattributes` / `README.md`), and raises `IncompleteSnapshotError` → surfaced as `OSError: Cannot load model...`.

- **hub 1.22.0:** completeness check fails → error.
- **hub 1.21.0:** returned the cached snapshot without checking completeness → succeeded.

This is independent of the diffusers version. It was first spotted in PEFT CI (their `hub_online_once` test helper flips to `HF_HUB_OFFLINE=1` on the second load of a model), but it is fundamentally a diffusers ↔ hfh interaction.

## Suggested fix

Always call `snapshot_download` with the **same** filters. Concretely, compute `allow_patterns` **before** the `model_info()` call in `download()` (or pass the same minimal `allow_patterns` in the offline fallback path), so hfh doesn't try to verify files diffusers never intended to download.

A hacky alternative is passing `ignore_patterns='*'` in offline mode (not recommended). On the hfh side, adding an opt-in `ignore_incomplete_snapshot` flag was floated as a fallback, but there are still some design decisions to settle before going that route.

## Repro

Load a pipeline once online, then load it again with `HF_HUB_OFFLINE=1` and `huggingface_hub>=1.22.0`.

cc @Wauplin @BenjaminBossan

---
Requested by Sayak Paul - [Slack thread](https://huggingface.slack.com/archives/C02V5EA0A95/p1783071444380129?thread_ts=1783071444.380129&cid=C02V5EA0A95) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-07-03T14-58-08-450Z_cf195c16-ce56-4f45-bb9d-ca03bc3af66d.jsonl)

sayakpaul:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/issues/14117

@BenjaminBossan does this work for you?


BenjaminBossan:
The PEFT tests succeed on this branch, thanks a lot.

(reminder to myself: testing with `pytest "tests/test_stablediffusion.py::TestStableDiffusionModel::test_disable_adapter" "tests/test_tuners_utils.py::TestPeftCustomKwargs::test_maybe_include_all_linear_layers_diffusion" -v`)

examples/instruct_pix2pix/train_instruct_pix2pix_sdxl.py
```diff
@@ -85,9 +85,11 @@ def log_validation(pipeline, args, accelerator, generator, global_step, is_final
         os.makedirs(val_save_dir)
 
     original_image = (
-        lambda image_url_or_path: load_image(image_url_or_path)
-        if urlparse(image_url_or_path).scheme
-        else Image.open(image_url_or_path).convert("RGB")
+        lambda image_url_or_path: (
```
BenjaminBossan: Changes seem to be unrelated.

BenjaminBossan (approved):
I can't comment on the logic of the fix, but it resolves the PEFT errors. Thanks.

BenjaminBossan:
A question: Is there going to be a patch release with this fix? If not, I need to prepare a fix on PEFT to exempt the failing tests until Diffusers 0.40 is released.

sayakpaul:
It's a breaking change from hfh and not diffusers. So, maybe you will need to wait till the next release.

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1776,6 +1781,27 @@ def download(cls, pretrained_model_name, **kwargs) -> str | os.PathLike:
...
+            try:
+                config_file = hf_hub_download(
+                    pretrained_model_name,
+                    cls.config_name,
+                    cache_dir=cache_dir,
+                    revision=revision,
+                    token=token,
+                    local_files_only=True,
+                )
+                return Path(config_file).parent
```
Wauplin: hacky but "works" though you don't benefit from the new "raise if incomplete snapshot"-feature i.e. if a previous download failed mid-way it won't be caught here, meaning a separate error will be raised downstream (likely when trying to load the model) - and the base exception will have been lost in the process

Wauplin (approved):
This PR feels correct 👍  (sorry for the delay reviewing it)

Though for a broader context I still do not understand why the allow_pattern/ignore_pattern arguments are different depending on the Hub availability

> It's a breaking change from hfh and not diffusers. So, maybe you will need to wait till the next release.

It's not considered as a breaking change on hfh-side so it won't be fixed in next releases no

Wauplin (commented):
(sorry just wanted to "comment" instead of "approve" -not well versed with diffusers internals- but the summary is the same^^)

BenjaminBossan:
> > It's a breaking change from hfh and not diffusers. So, maybe you will need to wait till the next release.
> 
> It's not considered as a breaking change on hfh-side so it won't be fixed in next releases no

I took this to mean that it will have to wait until the next Diffusers (minor) release, which will contain this fix, as there won't be a Diffusers patch release for this.

sayakpaul:
> It's not considered as a breaking change on hfh-side so it won't be fixed in next releases no

Next release of Diffusers. I am not sure why it's not considered a breaking change on HFH side because it clearly broke things in Diffusers as we can see. From what I understand, the behaviour of the underlying HFH-side functions also changed. Maybe I am missing something.

Wauplin:
> . I am not sure why it's not considered a breaking change on HFH side because it clearly broke things in Diffusers as we can see. 

Mostly because it's considered as bug fix (I know it's mostly semantic^^)

> because it clearly broke things in Diffusers as we can see

did it broke things only in the tests or in real life? I feel like if we are now raising an error for an incomplete download it's better than having diffusers raise because a file is not found in a local snapshot that was supposed to contain it

src/diffusers/pipelines/pipeline_utils.py
```diff
@@ -1657,10 +1662,37 @@ def download(cls, pretrained_model_name, **kwargs) -> str | os.PathLike:
...
+                )
+            except LocalEntryNotFoundError:
+                config_file = None
+
+        if config_file is not None:
+            snapshot_folder = Path(config_file).parent
+            if local_files_only:
+                filenames = {
+                    f.relative_to(snapshot_folder).as_posix() for f in snapshot_folder.rglob("*") if f.is_file()
+                }
```
Wauplin: requesting to download a single file to then use `Path(...).parent` is IMO a hacky way of getting the snapshot_folder. (hence why it needs 7 lines of explanation in comments) 

to avoid that I've opened https://github.com/huggingface/huggingface_hub/pull/4500 to expose the `snapshot_folder` in the error in case of `IncompleteSnapshotError`. I think this is the most explicit way to do things. In diffusers the code will look like this:

```py
try:
    snapshot_path = snapshot_download(...)
except IncompleteSnapshotError as error:
    incomplete_snapshot_path = error.snapshot_path
```

making it explicit for anyone reading the code that the folder is incomplete and that we knowingly ignored it

(even though, once again, I do think the best way to tackle this is to provide the correct allow_pattern/ignore_pattern filters in the first place. I still don't understand why it's not the solution we are trying to have in this PR. If there is a rational behind not trying to fix the root cause I genuinely believe it should at the very least be mentioned/documented somewhere)

sayakpaul: Thanks for offering further thoughts. Post-mortem:

The reason we can't write the patterns down without any file listing: they're data-dependent. allow_patterns are largely concrete filenames chosen by `variant_compatible_siblings` (variant / sharded / safetensors-vs-bin resolution), and `_get_ignore_patterns` picks formats based on what actually exists in the repo. 

For the online path, that listing comes from `model_info().siblings`. For the offline path, the only faithful source is the cached snapshot itself, which holds exactly what the same computation downloaded when online — so both modes compute the same patterns. We also test it here:
https://github.com/huggingface/diffusers/blob/e92e2590cb7bbf5e13c084e96cb2cd2001a2f640/tests/pipelines/test_pipelines.py#L508

More precisely, the offline path falls through into the same pattern-computation code as online, fed with a filenames' set that came from the cache instead of `model_info().siblings`. Before this PR, the offline path skipped all of this (patterns stayed None).

Regarding `hf_hub_download(config_file)`, the online branch makes the identical call because `model_index.json`'s content dictates the computation (component folders, `_ignore_files`, custom pipeline):
https://github.com/huggingface/diffusers/blob/e92e2590cb7bbf5e13c084e96cb2cd2001a2f640/src/diffusers/pipelines/pipeline_utils.py#L1655-L1664

(All of that is used for computing the patterns)
 
For the offline path, we are also reusing its parent as the snapshot dir, matching the pre-existing `snapshot_folder = Path(config_file).parent` on `main`.

https://github.com/huggingface/huggingface_hub/pull/4500 is cool, thanks!  `snapshot_path` on the error does locate the folder correctly. But the error only fires from `snapshot_download`.We need the folder _before_ calling `snapshot_download` (its listing is the input to the pattern computation). Using it would mean a deliberate no-patterns call that fails on every offline diffusers load just to catch the error. 

Maybe there's a way to refactor this but that's for another PR.

Wauplin: Thanks for the detailed response! That makes sense then. So if I understand correctly what's missing before-hand is the list of files in the repo when in offline mode, right? Luckily we now have this information in cache -hence why we are able to tell if a folder is incomplete-. We could definitely add a method to retrieve that cached list to that you can:

1. make a repo info call to get  `repo.siblings`
2. if that fails, get `repo.siblings` from the cache (requires a method from `huggingface_hub`)
3. and then compute the allow/ignore patterns from the list of files, no matter if you are in online or offline mode

Would that be interesting for you?

> Using it would mean a deliberate no-patterns call that fails on every offline diffusers load just to catch the error.

Yes that's what I had in mind. I feel that it's less hacky than a deliberate `hf_hub_download` call on the index file, just to do a `Path(index file).parent` afterwards.

Wauplin: opened https://github.com/huggingface/huggingface_hub/pull/4513 to add a `get_cached_repo_tree` method useful when `repo.siblings` or `list_repo_tree` fails (in offline mode). _A priori_, this is the cleanest way to determine the patterns down the line (with variants, etc.) without raising an error and without skipping the "incomplete folder check". Should be shipped today or tomorrow with the rest.

sayakpaul: @Wauplin thanks! https://github.com/huggingface/huggingface_hub/pull/4513 is it released? If so, I can work reviving this PR.

Wauplin: Yes it is! https://github.com/huggingface/huggingface_hub/pull/4513#issuecomment-4926352199

sayakpaul: if you want to review again.

tests/pipelines/test_pipelines.py
```diff
@@ -504,6 +505,51 @@ def test_local_files_only_are_used_when_no_internet(self):
...
+                    cache_dir=tmpdirname,
+                    local_files_only=True,
+                )
+                offline_kwargs = mock_snapshot_download.call_args.kwargs
+
+            assert os.path.samefile(offline_folder, cached_folder)
+            assert set(offline_kwargs["allow_patterns"]) == set(online_kwargs["allow_patterns"])
+            assert set(offline_kwargs["ignore_patterns"]) == set(online_kwargs["ignore_patterns"])
+
+    @require_hf_hub_version_greater("1.21.0")
```
DN6: Nit, we don't need this check, since we're bumping the min required hub version in `setup.py`

tests/pipelines/test_pipelines.py
```diff
@@ -504,6 +505,51 @@ def test_local_files_only_are_used_when_no_internet(self):
...
+                    "hf-internal-testing/tiny-stable-diffusion-torch", cache_dir=tmpdirname
+                )
+                online_kwargs = mock_snapshot_download.call_args.kwargs
+
+                offline_folder = DiffusionPipeline.download(
+                    "hf-internal-testing/tiny-stable-diffusion-torch",
+                    cache_dir=tmpdirname,
+                    local_files_only=True,
+                )
+                offline_kwargs = mock_snapshot_download.call_args.kwargs
```
DN6, line 532: Just FYI, in the offline case, these kwargs are never passed to `snapshot_download`. It early returns the existing snapshot folder:
https://github.com/huggingface/diffusers/blob/375840dc6cb77af64398d5265cb660de94649cca/src/diffusers/pipelines/pipeline_utils.py#L1806-L1809

Add a `force_download=True` to the offline version to verify the computed patterns.

DN6 (approved):

PR #14119, opened 2026-07-03
Feat -- Stable Audio 3

### LINKED ISSUE #13793: Support for Stable Audio 3 Medium
### Model/Pipeline/Scheduler description

Stability AI recently released Stable Audio 3 Medium, a new 2-billion-parameter text-to-audio diffusion model. It is capable of generating high-quality music and sound effects, and performing targeted audio editing based on text prompts.

Given Diffusers' expanding support for audio modalities, adding a dedicated pipeline for Stable Audio 3 Medium would allow the community to easily download, run, and fine-tune this model using standard Diffusers syntax.

### Open source status

- [x] The model implementation is available.
- [x] The model weights are available (Only relevant if addition is not a scheduler).

### Provide useful links for the implementation

Model Weights: https://huggingface.co/stabilityai/stable-audio-3-medium

Original Code/Tools: https://github.com/Stability-AI/stable-audio-tools

Announcement/Paper: https://stability.ai/news/stable-audio-open

Motivation:
Stable Audio 3 Medium represents a massive leap in open-weight audio generation. Currently, running it requires pulling custom code from the original research repository, which can be difficult to manage in production environments.

Integrating this into the diffusers ecosystem would
i. Make it vastly more accessible to developers building custom audio tools, AI-assisted DAWs, or automated sound design workflows.
ii. Allow the model to benefit from standard Diffusers memory optimizations (like sequential CPU offloading), enabling users to run this heavy model efficiently on consumer hardware.
iii. Allow seamless chaining with other modalities (e.g., generating video with a video pipeline, then passing the context to this pipeline for matching audio).

buffett0323:
# What does this PR do?

Adds Stable Audio 3 Medium to Diffusers.

Fixes #13793 

## Before submitting
- [x] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/c/discussion-related-to-httpsgithubcomhuggingfacediffusers/63)? Please add a link to it if that's the case. https://github.com/huggingface/diffusers/issues/13793

## Who can review?
@yiyixuxu @DN6 @sayakpaul


src/diffusers/pipelines/stable_audio_3/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .modeling_stable_audio_3 import StableAudio3DurationEmbedder
```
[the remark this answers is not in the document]
buffett0323: solved, thanks!

docs/source/en/api/pipelines/autodoc.md
```diff
@@ -0,0 +1,2 @@
+# autodoc
```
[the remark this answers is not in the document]
buffett0323: removed, thanks!

buffett0323:
Hi, I just updated the code, fixing serge review's suggestion, thanks!

sdevil7th:
Hi @buffett0323 , thanks for working on Stable Audio 3 support. The PR already covers a lot of important ground- conversion, SAME autoencoder, transformer, schedulers, text-to-audio, inpainting, docs, and tests!

One workflow gap I noticed is native audio-to-audio variation. Stability’s reference library exposes this via `init_audio` + `init_noise_level`:

```python
model.generate(
    init_audio=init_audio,
    init_noise_level=0.5,
    prompt="...",
    duration=30,
)
```

In this PR, I see text-to-audio and inpainting via `audio` + mask controls, but not an equivalent init-audio/noise-strength path. Inpainting can approximate some variation-like behavior, but it is not the same workflow: true audio-to-audio variation noises/denoises the whole source according to `init_noise_level`.

Suggested improvement: either add native audio-to-audio support, e.g. `StableAudio3AudioToAudioPipeline` or explicit `init_audio` / `init_noise_level` args, or document clearly that audio-to-audio variation is not exposed yet so users do not confuse it with inpainting.

One related base-checkpoint question: the PR docs mention `stable-audio-3-medium-base`, while Stability’s reference docs treat `cfg_scale` / `negative_prompt` as base-model-only controls. The current pipeline does not expose `guidance_scale` / `negative_prompt`. If base checkpoints are in scope for this PR, should that be supported or explicitly documented as out of scope?

Also, while testing the converter, I noticed two implementation issues:

1. Scheduler selection appears to miss one valid SA3 Medium config shape. The converter currently reads `model.diffusion_objective` / top-level `diffusion_objective`, but my `model_config.json` has it under `model.diffusion.diffusion_objective` with value `rf_denoiser`. In that case the converter silently falls back to `StableAudio3EulerScheduler` with 100 steps instead of `PingPongScheduler` with 8 steps.

2. The converter uses Unicode characters in console output (`──`, `→`, `✓`, etc.). On Windows environments using a non-UTF-8 code page, this can raise `UnicodeEncodeError` during conversion. It may be safer for conversion scripts to either force UTF-8 output or keep logs ASCII-safe.

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,784 @@
...
+    t_rot = (t_rot * freqs.cos()) + (_rotate_half(t_rot) * freqs.sin())
+    return torch.cat((t_rot.to(out_dtype), t_pass), dim=-1)
+
+
+# ──────────────────────────────────────────────────────────────────────────────
+# Normalization layers
+# ──────────────────────────────────────────────────────────────────────────────
+
+
+class _DynamicTanh(nn.Module):
```
dg845:
```suggestion
class DynamicTanh(nn.Module):
```
nit: I don't think we need to prefix all of the helper module names with `_`, as the modeling code for other autoencoders doesn't do this. For example, see `AutoencoderOobleck`:

https://github.com/huggingface/diffusers/blob/208704a27a6f362b67cd1a04fa1db0b98036d26f/src/diffusers/models/autoencoders/autoencoder_oobleck.py#L30

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,784 @@
...
+
+class _Attention(nn.Module):
+    """
+    Self-attention used inside TRB transformer blocks.
+
+    When *use_differential* is True (default for SAME-L), the block runs two independent attention maps — Attn(Q1,K1,V)
+    − Attn(Q2,K2,V) — so that attention patterns common to both heads cancel out, improving focus.
+    """
+
+    def __init__(self, dim: int, dim_heads: int = 128, use_differential: bool = True, qk_norm_eps: float = 1e-3):
```
dg845: The `qk_norm_eps` argument here doesn't appear to be used, is this intentional? If it isn't needed we should remove it.

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,784 @@
...
+        gate = torch.sin(math.pi * gate) if self.sinusoidal else F.silu(gate)
+        return self.proj_out(x * gate)
+
+
+# ──────────────────────────────────────────────────────────────────────────────
+# Self-attention (standard or differential, with QK-DyT norm and partial RoPE)
+# ──────────────────────────────────────────────────────────────────────────────
+
+
+class _Attention(nn.Module):
```
dg845: Would it be possible to refactor `_Attention` to use the `diffusers` attention module (`AttentionModuleMixin`) + attention processor pattern? For example, this is how Flux 2 implements the attention module:

https://github.com/huggingface/diffusers/blob/208704a27a6f362b67cd1a04fa1db0b98036d26f/src/diffusers/models/transformers/transformer_flux2.py#L494-L496

and an attention processor:

https://github.com/huggingface/diffusers/blob/208704a27a6f362b67cd1a04fa1db0b98036d26f/src/diffusers/models/transformers/transformer_flux2.py#L326-L328

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,784 @@
...
+        freqs = self.rope(N, x.device)
+
+        if self.use_differential:
+            q1, q2, k1, k2, v = self.to_qkv(x).chunk(5, dim=-1)
+            q1, q2, k1, k2, v = map(self._heads, (q1, q2, k1, k2, v))
+            q1, q2 = self.q_norm(q1), self.q_norm(q2)
+            k1, k2 = self.k_norm(k1), self.k_norm(k2)
+            q1, q2 = _apply_rotary(q1, freqs), _apply_rotary(q2, freqs)
+            k1, k2 = _apply_rotary(k1, freqs), _apply_rotary(k2, freqs)
+            out = F.scaled_dot_product_attention(q1, k1, v, attn_mask=attn_mask) - F.scaled_dot_product_attention(
```
dg845: Could we use `dispatch_attention_fn` instead of `F.scaled_dot_product_attention` here? This would allow us to support different attention backends such as Flash Attention (with `torch` native SDPA still supported as the default `"native"` backend). For example, this is what Flux 2 does:

https://github.com/huggingface/diffusers/blob/208704a27a6f362b67cd1a04fa1db0b98036d26f/src/diffusers/models/transformers/transformer_flux2.py#L369-L376

For reference, you can also look at the attention backend docs [here](https://huggingface.co/docs/diffusers/v0.39.0/en/optimization/attention_backends).

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,784 @@
...
+        encoder_sinusoidal_blocks: List[int] = (0,),
+        decoder_sinusoidal_blocks: List[int] = (0,),
+        sampling_rate: int = 44100,
+    ):
+        super().__init__()
+
+        # Derived constants — set as attributes so they're accessible without
+        # touching the (possibly absent) sub-modules.
+        self.sampling_rate = sampling_rate
+        self.latent_dim = latent_dim
```
dg845:
```suggestion
```
I don't think we need to define these attributes as they are already registered to the config and available via e.g. `self.config.sampling_rate`.

src/diffusers/models/transformers/transformer_stable_audio3.py
```diff
@@ -0,0 +1,557 @@
...
+        half = self.dim // 2
+        ramp = torch.linspace(0.0, 1.0, half, device=timestep.device, dtype=torch.float32)
+        freqs = torch.exp(ramp * (math.log(self.max_freq) - math.log(self.min_freq)) + math.log(self.min_freq))
+        args = timestep.float() * freqs * 2.0 * math.pi
+        out = torch.cat([torch.cos(args), torch.sin(args)], dim=-1)
+        return out.to(timestep.dtype)
+
+
+class StableAudio3RotaryEmbedding(nn.Module):
+    """Shared rotary positional embedding (partial RoPE over the first `2 * (dim // 2)` head channels)."""
```
dg845:
```suggestion
    """Shared rotary positional embedding. Stable Audio 3 applies the RoPE embeddings partially over only the first
    `2 * (dim // 2)` head channels).
    """
```
nit: I think the suggested docstring is a little more clear since the partial RoPE behavior isn't implemented inside `StableAudio3RotaryEmbedding` itself.

src/diffusers/models/transformers/transformer_stable_audio3.py
```diff
@@ -0,0 +1,557 @@
...
+        return torch.cat((freqs, freqs), dim=-1)  # (seq_len, rot_dim)
+
+
+def _rotate_half(x: torch.Tensor) -> torch.Tensor:
+    x = x.unflatten(-1, (2, -1))
+    x1, x2 = x.unbind(-2)
+    return torch.cat((-x2, x1), dim=-1)
+
+
+def _apply_rotary(t: torch.Tensor, freqs: torch.Tensor) -> torch.Tensor:
```
dg845:
```suggestion
def _apply_rotary_partial(t: torch.Tensor, freqs: torch.Tensor) -> torch.Tensor:
```
nit: I think renaming this to `_apply_rotary_partial` could be more clear as it more precisely describes its behavior.

src/diffusers/models/transformers/transformer_stable_audio3.py
```diff
@@ -0,0 +1,557 @@
...
+        inner = int(dim * mult)
+        self.proj_in = nn.Linear(dim, inner * 2, bias=True)
+        self.proj_out = nn.Linear(inner, dim, bias=True)
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        hidden_states, gate = self.proj_in(hidden_states).chunk(2, dim=-1)
+        return self.proj_out(hidden_states * F.silu(gate))
+
+
+class StableAudio3SelfAttention(nn.Module):
```
dg845: Could we refactor `StableAudio3SelfAttention` and `StableAudio3CrossAttention` below to use the `diffusers` Attention module + attention processor pattern? For example, I think we could potentially used a shared `AttentionModuleMixin` class and have separate self- and cross-attention attention processors. This would help support using different attention backends.

src/diffusers/models/transformers/transformer_stable_audio3.py
```diff
@@ -0,0 +1,557 @@
...
+
+    def forward(self, hidden_states: torch.Tensor, rope: torch.Tensor) -> torch.Tensor:
+        if self.use_differential:
+            q1, q2, k1, k2, v = self.to_qkv(hidden_states).chunk(5, dim=-1)
+            q1, q2, k1, k2, v = map(self._heads, (q1, q2, k1, k2, v))
+            q1, q2 = self.q_norm(q1), self.q_norm(q2)
+            k1, k2 = self.k_norm(k1), self.k_norm(k2)
+            q1, q2 = _apply_rotary(q1, rope), _apply_rotary(q2, rope)
+            k1, k2 = _apply_rotary(k1, rope), _apply_rotary(k2, rope)
+            out = F.scaled_dot_product_attention(q1, k1, v) - F.scaled_dot_product_attention(q2, k2, v)
```
dg845: As with the SAME autoencoder, could we use `dispatch_attention_fn` for both self- and cross-attention in place of the `F.scaled_dot_product_attention` calls? This would allow us to support different attention backends like Flash Attention.

src/diffusers/models/transformers/transformer_stable_audio3.py
```diff
@@ -0,0 +1,557 @@
...
+        timestep_features_dim: int = 256,
+        ff_mult: float = 4.0,
+        num_memory_tokens: int = 64,
+        use_differential_attention: bool = True,
+    ):
+        super().__init__()
+
+        self.patch_size = patch_size
+        self.io_channels = io_channels
+        self.num_memory_tokens = num_memory_tokens
```
dg845:
```suggestion
```
I don't think we need to define these attributes explicitly as they will be available via the config. For example, we could use `self.config.patch_size` in `forward` below without needing to define `self.patch_size` here.

src/diffusers/pipelines/stable_audio_3/pipeline_stable_audio_3.py
```diff
@@ -0,0 +1,478 @@
...
+            device: Target device.
+            num_waveforms_per_prompt: Tile factor.
+            batch_size: Number of prompts.
+
+        Returns:
+            ``(batch * num_waveforms_per_prompt, output_dim)`` tensor.
+        """
+        if not isinstance(duration, (int, float)):
+            raise ValueError(f"`duration` must be a single `float`, got {type(duration)}.")
+
```
dg845:
```suggestion
```
nit: I don't think we need to defensively check whether `duration` is a `float` or `int` since we're already annotating its type as `float` here and in `__call__`.

src/diffusers/pipelines/stable_audio_3/pipeline_stable_audio_3.py
```diff
@@ -0,0 +1,478 @@
...
+                    prompt_embeds = callback_outputs.pop("prompt_embeds", prompt_embeds)
+
+                progress_bar.update()
+
+                if XLA_AVAILABLE:
+                    xm.mark_step()
+
+        # 7. Decode latents
+        if output_type == "latent":
+            return AudioPipelineOutput(audios=latents)
```
dg845: nit: I think we should also support `return_dict=False` in the `output_type="latent"` case here. For example, we could do something like

```python
        if output_type == "latent":
            audio = latents
        else:
            audio = self.vae.decode(latents).sample
            ...
        ...
        if not return_dict:
            return (audio,)
        return AudioPipelineOutput(audios=audio)
```

src/diffusers/pipelines/stable_audio_3/pipeline_stable_audio_3_inpaint.py
```diff
@@ -0,0 +1,312 @@
...
+from typing import Callable, List, Optional, Union
+
+import torch
+import torch.nn.functional as F
+
+from ..pipeline_utils import AudioPipelineOutput
+from .pipeline_stable_audio_3 import StableAudio3Pipeline
+
+
+class StableAudio3InpaintPipeline(StableAudio3Pipeline):
```
dg845:
```suggestion
class StableAudio3InpaintPipeline(DiffusionPipeline):
```
`StableAudio3InpaintPipeline` should not inherit from `StableAudio3Pipeline` but rather should inherit directly from `DiffusionPipeline`. We can copy over shared methods such as `encode_prompt`, `encode_duration`, etc. with the `# Copied from` mechanism to sync the implementations:

```python
class StableAudio3InpaintPipeline(StableAudio3Pipeline):
    ...
    # Copied from diffusers.pipelines.stable_audio_3.pipeline_stable_audio3.StableAudio3Pipeline.encode_prompt
    def encode_prompt(
        self,
        prompt: Optional[Union[str, List[str]]],
        ...,
    ):
        ...
    ...
```

src/diffusers/pipelines/stable_audio_3/pipeline_stable_audio_3_inpaint.py
```diff
@@ -0,0 +1,312 @@
...
+            audio_latents = audio_latents[:, :, :target_length]
+
+        # Masked latent: zero out the inpaint region
+        masked_latent = audio_latents * mask
+
+        # Concat along channel dim: (batch, 1 + latent_dim, L)
+        return torch.cat([mask, masked_latent], dim=1)
+
+    @torch.no_grad()
+    def __call__(
```
dg845, line 457: nit: could we also have an example for the inpainting pipeline?

src/diffusers/schedulers/scheduling_stable_audio3_euler.py
```diff
@@ -0,0 +1,196 @@
...
+            Computed sample at the previous (less noisy) timestep.
+        pred_original_sample (`torch.Tensor`, *optional*):
+            The predicted de-noised sample (x₀ prediction) at the current step.
+    """
+
+    prev_sample: torch.Tensor
+    pred_original_sample: Optional[torch.Tensor] = None
+
+
+class StableAudio3EulerScheduler(SchedulerMixin, ConfigMixin):
```
dg845: Can we reuse the existing `FlowMatchEulerDiscreteScheduler` in place of `StableAudio3EulerScheduler`? I think supplying the log-SNR sigma schedule as a custom `sigmas` schedule to `FlowMatchEulerDiscreteScheduler.set_timesteps` could potentially be equivalent.

src/diffusers/schedulers/scheduling_ping_pong.py
```diff
@@ -0,0 +1,200 @@
...
+            Computed sample at the previous (less noisy) timestep.
+        pred_original_sample (`torch.Tensor`, *optional*):
+            The predicted de-noised sample (x₀ prediction) at the current step.
+    """
+
+    prev_sample: torch.Tensor
+    pred_original_sample: Optional[torch.Tensor] = None
+
+
+class PingPongScheduler(SchedulerMixin, ConfigMixin):
```
dg845: Similarly, could we also reuse `FlowMatchEulerDiscreteScheduler` in place of `PingPongScheduler`? I think using `FlowMatchEulerDiscreteScheduler` with `stochastic_sampling=True` might be equivalent as it does a similar denoise + re-noise to the next noise level operation like `PingPongScheduler` does:

https://github.com/huggingface/diffusers/blob/208704a27a6f362b67cd1a04fa1db0b98036d26f/src/diffusers/schedulers/scheduling_flow_match_euler_discrete.py#L506-L509

dg845 (commented):

src/diffusers/models/transformers/transformer_stable_audio3.py
buffett0323: yeah, i agree with that and am going to change it, thanks!

buffett0323:
@dg845 Hi, thanks for your time and suggestion. I've revised the code. If there's anything else needed, I would love to help, thanks again!

buffett0323:
@dg845 Sorry to bother you again, I saw there was a CI/CD bug after you merged from main and I just solved it. Can you take a look and do it again? Thanks for your help and time!

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,810 @@
...
+            q, k = attn.q_norm(q), attn.k_norm(k)
+            q, k = _apply_rotary(q, freqs), _apply_rotary(k, freqs)
+            out = dispatch_attention_fn(
+                q, k, v, attn_mask=attn_mask, backend=self._attention_backend, parallel_config=self._parallel_config
+            )
+
+        return attn.to_out(out.flatten(2, 3))
+
+
+class Attention(nn.Module, AttentionModuleMixin):
```
dg845:
```suggestion
class SAMEAttention(nn.Module, AttentionModuleMixin):
```
nit: I think renaming to `SAMEAttention` here is more clear, especially as it uses logic (such as differential attention) that isn't in most other `Attention` modules.

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,810 @@
...
+    return torch.cat([x, x.new_zeros(shape)], dim=dim)
+
+
+def _rotate_half(x: torch.Tensor) -> torch.Tensor:
+    x = x.unflatten(-1, (2, -1))
+    x1, x2 = x.unbind(-2)
+    return torch.cat((-x2, x1), dim=-1)
+
+
+def _apply_rotary(t: torch.Tensor, freqs: torch.Tensor) -> torch.Tensor:
```
dg845:
```suggestion
# Copied from diffusers.models.transformers.transformer_stable_audio3._apply_rotary_partial
def _apply_rotary_partial(t: torch.Tensor, freqs: torch.Tensor) -> torch.Tensor:
```
Is the `_apply_rotary` function here intended to be the same as `_apply_rotary_partial` in the transformer code? If so we can sync the implementations using `# Copied from` as in the suggestion above.

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,810 @@
...
+    size = x.shape[dim]
+    pad = (multiple - size % multiple) % multiple
+    if pad == 0:
+        return x
+    shape = list(x.shape)
+    shape[dim] = pad
+    return torch.cat([x, x.new_zeros(shape)], dim=dim)
+
+
+def _rotate_half(x: torch.Tensor) -> torch.Tensor:
```
dg845, line 86:
```suggestion
# Copied from diffusers.models.transformers.transformer_stable_audio3._rotate_half
def _rotate_half(x: torch.Tensor) -> torch.Tensor:
```
Similarly, if `_rotate_half` should be the same as in the transformer code, we can sync the implementations with `# Copied from`.

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,810 @@
...
+        )
+        self.transformers = nn.ModuleList(
+            [
+                TransformerBlock(
+                    tdim,
+                    dim_heads=min(dim_heads, tdim),
+                    use_differential=use_differential,
+                    ff_mult=ff_mult,
+                    # The trailing `sinusoidal_blocks` layers use sin gating (matches the reference).
+                    sinusoidal=(transformer_depth - i) < sinusoidal_blocks,
```
dg845, line 360:
```suggestion
                    sinusoidal=(transformer_depth - i) <= sinusoidal_blocks,
```
I think there may be an off-by-one error here, where if `transformer_depth >= sinusoidal_blocks`, we only end up with `sinusoidal_blocks - 1` rather than `sinusoidal_blocks` sinusoidal layers, because we accidentally drop the sinuoisdal layer with the smallest index (`transformer_depth - sinusoidal_blocks`).

For example, if `transformer_depth == sinusoidal_blocks == 1`, the only layer `0` ends up with `sinusoidal = (transformer_depth - 0) < 1 = 1 < 1 = False`, so we have 0 sinusoidal layers, not 1 as expected.

buffett0323:
@dg845 Doc build failures are resolved, and I believe the only remaining failure is unrelated to this PR. Thanks again for your time and support!

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,810 @@
...
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        stride: int,
+        mode: str = "encoder",
+        transformer_depth: int = 3,
+        dim_heads: int = 128,
+        use_differential: bool = True,
+        chunk_size: int = 128,
```
dg845: The docstring above says that `chunk_size` is no longer used, and `self.chunk_size` doesn't seem to be currently used anywhere in `SAMETransformerResamplingBlock` below. Given this, could we remove the `chunk_size` argument?

src/diffusers/models/transformers/transformer_stable_audio3.py
```diff
@@ -0,0 +1,594 @@
...
+        self.is_cross_attention = context_dim is not None
+
+        if self.is_cross_attention:
+            n_q = 2 if use_differential else 1
+            n_kv = 3 if use_differential else 2
+            self.to_q = nn.Linear(dim, dim * n_q, bias=False)
+            self.to_kv = nn.Linear(context_dim, dim * n_kv, bias=False)
+        else:
+            n_proj = 5 if use_differential else 3
+            self.to_qkv = nn.Linear(dim, dim * n_proj, bias=False)
```
dg845: Since we branch on `self.is_cross_attention` here, and we also use different Q,K,V projections for self- and cross-attention (totally fused vs only KV fused), I think it would be more clear to have separate `AttentionModuleMixin` subclasses for self- and cross-attention:

```python
class StableAudio3SelfAttention(nn.Module, AttentionModuleMixin):
    ...
    _default_processor_cls = StableAudio3SelfAttnProcessor
    _available_processors = [StableAudio3SelfAttnProcessor]
    ...


class StableAudio3CrossAttention(nn.Module, AttentionModuleMixin):
    ...
    _default_processor_cls =StableAudio3CrossAttnProcessor
    _available_processors = [StableAudio3CrossAttnProcessor]
    ...
```

 This would also make it easier to specify the `forward` signature, since it is different between `StableAudio3SelfAttnProcessor` and `StableAudio3CrossAttnProcessor`.

src/diffusers/models/transformers/transformer_stable_audio3.py
```diff
@@ -0,0 +1,594 @@
...
+        cond_token_dim (`int`, defaults to 768): Dimension of the cross-attention context tokens.
+        global_cond_dim (`int`, defaults to 768): Dimension of the global duration embedding.
+        local_add_cond_dim (`int`, defaults to 257): Channels of the local-additive (inpaint) tensor.
+        timestep_features_dim (`int`, defaults to 256): Output dimension of the Fourier timestep features.
+        ff_mult (`float`, defaults to 4.0): SwiGLU feed-forward expansion factor.
+        num_memory_tokens (`int`, defaults to 64): Number of learnable memory tokens.
+        use_differential_attention (`bool`, defaults to `True`): Enable differential self/cross attention.
+    """
+
+    _supports_gradient_checkpointing = True
```
dg845, line 434:
```suggestion
    _supports_gradient_checkpointing = True
    _no_split_modules = ["StableAudio3DiTBlock"]
    _repeated_blocks = ["StableAudio3DiTBlock"]
```
We can set `_no_split_modules` here to support splitting the DiT across devices (e.g. via `from_pretrained(..., device_map="auto")` and `_repeated_blocks` to support regional compilation.

src/diffusers/pipelines/stable_audio_3/modeling_stable_audio_3.py
```diff
@@ -0,0 +1,111 @@
...
+                Duration values in seconds.
+
+        Returns:
+            `torch.Tensor` of shape `(batch, output_dim)` — duration embeddings.
+        """
+        cfg = self.config
+
+        # 1. Normalize to [0, 1]
+        seconds = seconds.float().clamp(cfg.min_val, cfg.max_val)
+        t_norm = (seconds - cfg.min_val) / (cfg.max_val - cfg.min_val)
```
dg845:
```suggestion
        # 1. Normalize to [0, 1]
        seconds = seconds.float().clamp(self.config.min_val, self.config.max_val)
        t_norm = (seconds - self.config.min_val) / (self.config.max_val - self.config.min_val)
```
nit: I think using `self.config.min_val` and `self.config.max_val` directly is more clear here.

src/diffusers/pipelines/stable_audio_3/modeling_stable_audio_3.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # 2. Exponential Fourier features — run in fp32 for stability
+        t_norm = t_norm.reshape(-1, 1)  # (B, 1)
+        args = t_norm * self.freqs.unsqueeze(0) * 2.0 * math.pi  # (B, half)
+        fourier = torch.cat([args.cos(), args.sin()], dim=-1)  # (B, fourier_dim)
+
+        # 3. Linear projection
+        return self.linear(fourier.to(self.linear.weight.dtype))  # (B, output_dim)
```
dg845:
```suggestion
        return self.linear(fourier.to(self.dtype))  # (B, output_dim)
```
Since `StableAudio3DurationEmbedder` inherits from `ModelMixin`, we can use `ModelMixin.dtype` here, which is also more robust if the model is quantized.

src/diffusers/pipelines/stable_audio_3/pipeline_stable_audio_3.py
```diff
@@ -0,0 +1,553 @@
...
+    XLA_AVAILABLE = False
+
+logger = logging.get_logger(__name__)
+
+EXAMPLE_DOC_STRING = """
+    Examples:
+        ```py
+        >>> import torch
+        >>> import soundfile as sf
+        >>> from diffusers import StableAudio3Pipeline, PingPongScheduler
```
dg845:
```suggestion
        >>> from diffusers import StableAudio3Pipeline
```
nit: remove the `PingPongScheduler` import as it isn't used in the example.

src/diffusers/pipelines/stable_audio_3/pipeline_stable_audio_3_audio2audio.py
```diff
@@ -0,0 +1,521 @@
...
+logger = logging.get_logger(__name__)
+
+
+EXAMPLE_DOC_STRING = """
+    Examples:
+        ```py
+        >>> import torch
+        >>> import soundfile as sf
+        >>> import torchaudio
+        >>> from diffusers import StableAudio3AudioToAudioPipeline, PingPongScheduler
```
dg845:
```suggestion
        >>> from diffusers import StableAudio3AudioToAudioPipeline
```
nit: remove `PingPongScheduler` as it is unused by the example.

src/diffusers/pipelines/stable_audio_3/pipeline_stable_audio_3_audio2audio.py
```diff
@@ -0,0 +1,521 @@
...
+        >>> generator = torch.Generator("cuda").manual_seed(0)
+        >>> audio = pipe(
+        ...     "A gentle piano melody with soft strings in a concert hall",
+        ...     duration=10.0,
+        ...     audio=audio,
+        ...     init_noise_level=0.6,
+        ...     generator=generator,
+        ... ).audios
+
+        >>> sf.write("output.wav", audio[0].T.cpu().float().numpy(), samplerate=44100)
```
dg845:
```suggestion
        >>> sf.write("output.wav", audio[0].T.cpu().float().numpy(), samplerate=pipe.vae.config.sampling_rate)
```
nit: I think it would more robust to use `pipe.vae.config.sampling_rate` here and elsewhere, but especially here since we already use it above when resampling the input `audio`.

src/diffusers/schedulers/scheduling_ping_pong.py
```diff
@@ -0,0 +1,155 @@
...
+
+from typing import Literal, Optional, Union
+
+import torch
+
+from ..configuration_utils import register_to_config
+from .scheduling_flow_match_euler_discrete import FlowMatchEulerDiscreteScheduler
+
+
+class PingPongScheduler(FlowMatchEulerDiscreteScheduler):
```
dg845: I think we should remove `PingPongScheduler` and `StableAudio3EulerScheduler` in favor of reusing the existing `FlowMatchEulerDiscreteScheduler` scheduler with a custom log-SNR `sigmas` schedule, as they are essentially equivalent except for the choice of sigma schedule.

For example, I think we could define something like

```python
def logsnr_sigma_schedule(num_inference_steps, logsnr_min, logsnr_max, ...):
    # log-SNR logic from current `PingPongScheduler.set_timesteps`.
    ...
```

and then support a `sigmas` argument to `__call__` and follow the standard `retrieve_timesteps` logic to support custom `sigmas` schedules, for example as used by the Flux 2 pipeline:

https://github.com/huggingface/diffusers/blob/13a7bee4878d62fccc8d25f97e480e68de96fa03/src/diffusers/pipelines/flux2/pipeline_flux2.py#L934-L945

If we want to support different log-SNR-style `sigmas` schedules, I think having `logsnr_min` and `logsnr_max` as additional `__call__` arguments makes sense for now.

src/diffusers/schedulers/scheduling_ping_pong.py
```diff
@@ -0,0 +1,155 @@
...
+                Compresses the schedule to ``[sigma_max, 0]`` instead of ``[1, 0]``, while keeping the full step count.
+                Used for audio-to-audio variation (``init_noise_level``): a lower `sigma_max` starts the denoising loop
+                closer to the clean signal, so less of the original audio is altered.
+        """
+        n = num_inference_steps if num_inference_steps is not None else self.config.num_inference_steps
+
+        logsnr = torch.linspace(self.config.logsnr_min, self.config.logsnr_max, n + 1)
+        sigmas = torch.sigmoid(-logsnr)  # (N+1,), decreasing
+        sigmas[0] = 1.0  # sigma_max: start from pure noise
+        sigmas[-1] = 0.0  # end fully denoised
```
dg845: Is overwriting `sigmas[0]` to be `1.0` and `sigmas[-1]` to be `0.0` intentional here? This seems a little weird as we are effectively overwriting `logsnr_min` and `logsnr_max` to $`-\infty`$ and $`\infty`$, respectively (although they still determine the spacing of the schedule).

src/diffusers/schedulers/scheduling_ping_pong.py
```diff
@@ -0,0 +1,155 @@
...
+        sigmas = sigmas * sigma_max
```
dg845: Similarly, is scaling the `sigmas` schedule by `sigma_max` intentional? This also seems a little weird since the resulting `sigmas` schedule is in general no longer uniform in log-SNR space.

tests/models/autoencoders/test_models_autoencoder_same.py
```diff
@@ -0,0 +1,201 @@
...
+    model.to(device).eval()
+    return model
+
+
+def _make_audio(batch: int, channels: int, n_samples: int, device: str = "cpu") -> torch.Tensor:
+    return torch.randn(batch, channels, n_samples, device=device)
+
+
+@require_torch
+class TestAutoencoderSAMETinyConfig(unittest.TestCase):
```
dg845: Can we migrate the SAME autoencoder tests to use the standard model test mixins? You can generate a standard model test suite skeleton using

```bash
python utils/generate_model_tests.py src/diffusers/models/autoencoders/autoencoder_same.py
```

tests/models/transformers/test_models_transformer_stable_audio3.py
```diff
@@ -0,0 +1,363 @@
...
+    torch.manual_seed(0)
+    hidden_states = torch.randn(batch, IO_CHANNELS, T_AUDIO, device=device)
+    timestep = torch.rand(batch, device=device)
+    encoder_hidden_states = torch.randn(batch, T_TEXT, COND_DIM, device=device)
+    global_hidden_states = torch.randn(batch, GLOBAL_DIM, device=device)
+    return hidden_states, timestep, encoder_hidden_states, global_hidden_states
+
+
+@require_torch
+class TestStableAudio3DiTModelTinyConfig(unittest.TestCase):
```
dg845: Similarly, can we also migrate the Stable Audio 3 transformer tests to use the standard model test mixins? You can generate a standard model test suite skeleton using

```bash
python utils/generate_model_tests.py src/diffusers/models/transformers/transformer_stable_audio3.py
```

scripts/convert_stable_audio_3_to_diffusers.py
```diff
@@ -0,0 +1,877 @@
...
+            dummy = torch.zeros(1, 2, 44100)
+            with torch.no_grad():
+                lat = pipeline.vae.encode(dummy).latents
+                rec = pipeline.vae.decode(lat).sample
+            print(f"  ✓ VAE encode→decode: input {dummy.shape} → latent {lat.shape} → output {rec.shape}")
+
+            # Quick duration embedder check
+            with torch.no_grad():
+                d = pipeline.duration_embedder(torch.tensor([10.0]))
+            print(f"  ✓ DurationEmbedder output shape: {d.shape}")
```
dg845:
```suggestion
            print(f"[DONE] DurationEmbedder output shape: {d.shape}")
```
Following @sdevil7th's comment, I think we should make sure that anything printed to `stdout` is ASCII, as this is more robust and also what the other conversion scripts do.

dg845 (commented):
Thanks for iterating! Left some follow up comments :).

buffett0323:
@dg845 Hi, I just updated the code, thanks again for your time and effort!

buffett0323:
@dg845 Sorry to bother you again, gentle reminder for this PR whenever you get a chance! Let me know if you have any questions or feedback. Thanks again for your help!!

docs/source/en/api/models/autoencoder_same.md
```diff
@@ -0,0 +1,42 @@
...
+
+The **SAME** (Semantically-Aligned Music Encoder) autoencoder is used by [Stable Audio 3](https://stability.ai/news/stable-audio-3)
+to compress stereo audio waveforms into a compact latent sequence and reconstruct them.
+
+The encoder stacks [`SAMETransformerResamplingBlock`] modules, each of which groups a fixed number of audio
+patch frames and produces one learnable output token via a differential transformer. The decoder inverts this
+process, expanding each latent token back to a patch of audio frames.
+
+A soft-norm bottleneck ([`SoftNormBottleneck`]) normalises latents before and after the diffusion model,
+providing stable training dynamics.
```
dg845:
```suggestion
A soft-norm bottleneck (`SoftNormBottleneck`) normalises latents before and after the diffusion model,
providing stable training dynamics.
```
nit: drop the brackets since `SoftNormBottleneck` has no corresponding `autodoc` entry.

scripts/convert_stable_audio_3_to_diffusers.py
```diff
@@ -0,0 +1,864 @@
...
+
+        text_encoder = text_encoder.to(dtype)
+        text_encoder_cls_name = type(text_encoder).__name__
+        text_encoder.save_pretrained(output_dir / "text_encoder")
+        tokenizer.save_pretrained(output_dir / "tokenizer")
+        print(f"  Saved -> {output_dir / 'text_encoder'}, {output_dir / 'tokenizer'}")
+    except ImportError:
+        print("  WARNING: transformers not installed; skipping text_encoder & tokenizer.")
+    except Exception as exc:
+        print(f"  WARNING: text encoder conversion failed: {exc}")
```
dg845:
```suggestion
```
I think we should let exceptions raise normally instead of catching them all here and allowing the conversion script to continue, which could result in the script outputting a silently broken `text_encoder`.

scripts/convert_stable_audio_3_to_diffusers.py
```diff
@@ -0,0 +1,864 @@
...
+    out = {}
+    skipped = []
+
+    ae_prefix = "pretransform.model."
+
+    enc_base = ae_prefix + "encoder.layers"
+    dec_base = ae_prefix + "decoder.layers"
+
+    enc_depth = _infer_trb_depth(ref_sd, enc_base)
+    dec_depth = _infer_trb_depth(ref_sd, dec_base + ".0") if f"{dec_base}.0.new_tokens" in ref_sd else 0
```
dg845: nit: since we ultimately loop over `enc_depth` when converting the SAME decoder blocks below, I think we can safely remove the `dec_depth` logic here.

src/diffusers/pipelines/stable_audio_3/pipeline_stable_audio_3.py
```diff
@@ -0,0 +1,632 @@
...
+        shape = (batch_size, latent_dim, latent_length)
+        if isinstance(generator, list) and len(generator) != batch_size:
+            raise ValueError(
+                f"You have passed a list of generators of length {len(generator)}, but requested an effective batch"
+                f" size of {batch_size}. Make sure the batch size matches the length of the generators."
+            )
+        if latents is None:
+            latents = randn_tensor(shape, generator=generator, device=device, dtype=dtype)
+        else:
+            latents = latents.to(device)
```
dg845:
```suggestion
        if latents is None:
            latents = randn_tensor(shape, generator=generator, device=device, dtype=dtype)
        else:
            latents = latents.to(device=device, dtype=dtype)
```
nit: I think it would be better to also cast pre-computed `latents` to the supplied `dtype`, since this matches the dtype behavior in the `if latents is None:` branch above.

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,800 @@
...
+        )
+        self.transformers = nn.ModuleList(
+            [
+                TransformerBlock(
+                    tdim,
+                    dim_heads=min(dim_heads, tdim),
+                    use_differential=use_differential,
+                    ff_mult=ff_mult,
+                    # The trailing `sinusoidal_blocks` layers use sin gating (matches the reference).
+                    sinusoidal=(transformer_depth - i) <= sinusoidal_blocks,
```
dg845:
```suggestion
                    # The trailing `sinusoidal_blocks` layers use sin gating (matches the reference).
                    sinusoidal=(transformer_depth - i) < sinusoidal_blocks,
```
Sorry, I think my previous suggestion was wrong: using `<=` results in samples which are pure noise, and reverting back to using `<` produces good samples again. Can you confirm this against the reference implementation and update the comment to clarify which is correct?

dg845 (commented):
Sorry for the delay and thanks for your patience! I think this PR is close to merge, left a few small comments :).

src/diffusers/models/autoencoders/autoencoder_same.py
buffett0323: Confirmed, [`Stability-AI/stable-audio-tools/stable_audio_tools/models/autoencoders.py:143-144`](https://github.com/Stability-AI/stable-audio-tools/blob/main/stable_audio_tools/models/autoencoders.py), thanks!

buffett0323:
Fixed, thanks again for your time @dg845 , please take a look!

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,800 @@
...
+            _wn_conv1d(in_channels, out_channels, 1, padding=0) if in_channels != out_channels else nn.Identity()
+        )
+        self.transformers = nn.ModuleList(
+            [
+                TransformerBlock(
+                    tdim,
+                    dim_heads=min(dim_heads, tdim),
+                    use_differential=use_differential,
+                    ff_mult=ff_mult,
+                    # The trailing `sinusoidal_blocks` layers use sin gating (matches the reference).
```
dg845, line 359:
```suggestion
                    # Sin-gate the trailing `sinusoidal_blocks - 1` layers ("<", not "<="), which matches the
                    # `stable-audio-tools` SAME reference (`stable_audio_tools/models/autoencoders.py`).
```
nit: update the comment to clarify the reference behavior.

src/diffusers/models/autoencoders/autoencoder_same.py
```diff
@@ -0,0 +1,800 @@
...
+        in_channels: Number of input channels.
+        out_channels: Number of output channels.
+        stride: Down-/up-sampling factor.
+        mode: ``"encoder"`` or ``"decoder"``.
+        transformer_depth: Number of :class:`TransformerBlock` layers.
+        dim_heads: Attention head dimension.
+        use_differential: Whether to use differential attention.
+        ff_mult: Feed-forward expansion factor.
+        sliding_window: Sliding-window half-width in latents (band half-width is ``sliding_window * (stride + 1)``).
+        sinusoidal_blocks: Number of trailing transformer layers that use ``sin`` FFN gating instead of SiLU.
```
dg845, line 321:
```suggestion
        sinusoidal_blocks (`int`, defaults to `0`):
            The `sinusoidal_blocks - 1` trailing layers will use `sin` FFN gating instead of SiLU gating. This matches
            the reference `stable-audio-tools` SAME transformer block `sinusoidal_blocks` semantics.
```
nit: update the docstring to clarify the `sinusoidal_blocks` behavior.

dg845 (approved):
Thanks again for your patience and work on this PR! Can you run `make fix-copies` to fix the [CI failure](https://github.com/huggingface/diffusers/actions/runs/32198061357/job/95906471855?pr=14119)? This should propagate the `prepare_latents` change in https://github.com/huggingface/diffusers/pull/14119#discussion_r3801701339 to the inpainting and audio-to-audio pipeline copies.

buffett0323:
Fixed and uploaded, can you please help me for the CI/CD? Thanks!
@dg845

dg845:

PR #14120, opened 2026-07-04
[tests] fix tests for transformers latest stable.

sayakpaul:
# What does this PR do?

Fixes https://github.com/huggingface/diffusers/actions/runs/28672597948/job/85039226641?pr=14118

Starting with transformers 5.13.0 (PR #46259), every model's `_init_weights` now calls the generic `super()._init_weights()` before applying its model-specific initialization, so each weight consumes extra RNG draws and a seeded `torch.manual_seed(0); CLIPTextModel(config)` produces different random weights than before — same distributions, different values. 

For fast pipeline tests, we build the dummy text encoders exactly that way and compare outputs against hardcoded slices generated under the old init, hence the test failures.


dg845 (approved):
Thanks for the PR! Two questions:

1. For `diffusers`, were only pipelines that used a CLIP model component affected by the change?
2. Does `tests/pipelines/stable_diffusion/test_stable_diffusion_inpaint.py::StableDiffusionInpaintPipelineFastTests::test_ip_adapter` also need to have an expected slice updated? I see that it fails in this PR run: https://github.com/huggingface/diffusers/actions/runs/28695148034/job/85103458225?pr=14120.

sayakpaul:
> For diffusers, were only pipelines that used a CLIP model component affected by the change?

Seems like it, yes.

> Does tests/pipelines/stable_diffusion/test_stable_diffusion_inpaint.py::StableDiffusionInpaintPipelineFastTests::test_ip_adapter also need to have an expected slice updated? I see that it fails in this PR run: https://github.com/huggingface/diffusers/actions/runs/28695148034/job/85103458225?pr=14120.

Sorry, I should have caught it earlier. Not sure why it didn't show up :o Updated now.

PR #14131, opened 2026-07-06
Make doc builds faster

mishig25:
Companion to the doc-builder tracker: huggingface/doc-builder#802

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #805, opened 2022-10-11
[ONNX] Loading textual inversion .bin/.pt embeddings


GreenLandisaLie:
I tried to follow this guide:
https://colab.research.google.com/github/huggingface/notebooks/blob/main/diffusers/stable_conceptualizer_inference.ipynb
But OnnxRuntimeModel does not have any of the necessary functions that CLIPTextModel has and since I'm no dev I couldn't figure out how to get this working.
If this is already possible then I guess this issue is just about the lack of documentation on this subject, else this is something that should be added eventually.
Also, it would be cool if we could have a simple single function within the pipelines for this purpose.

anton-l:
Hi @GreenLandisaLie! Unfortunately there's no way to seamlessly support ONNX models for textual inversion examples, since they modify the model itself, and the onnx graph is fairly static in its current implementation. 

I guess you could try loading the embeddings into a pytorch model first, save it, and then export it to ONNX with the default conversion script. But you'd have to do that for each learned concept :sweat_smile: 

Another alternative is to try updating the onnx model with the new embedding weights, similar to this discussion: https://github.com/onnx/onnx/issues/2978

GreenLandisaLie:
Thank you @anton-l for looking into this. 
Its was silly of me to not have thought about taking advantage of the conversion scripts.
With that information I can probably get it working on my end. Still its something that needs to be considered in the future once ONNX gets past its experimental stage so I'll leave this issue open.

PR #14134, opened 2026-07-07
[fix] missing doc-builder imports

stevhliu:
https://github.com/huggingface/diffusers/pull/14131 removed the old container which provided `sentencepiece` which causes the doc builder to break now. This fixes it by guarding the import to fail at use time.

The other thing is to add `bitsandbytes` to https://github.com/huggingface/doc-builder/blob/main/src/doc_builder/mock_deps/diffusers.txt cc @mishig25


mishig25:
Done in huggingface/doc-builder#805 — adds `real:sentencepiece` (installed for real by `doc-builder light-install`; kolors' tokenizer imports it at module level) and `bitsandbytes` as a mock to `mock_deps/diffusers.txt`. Validated the full 363-page diffusers docs build in a fresh venv with that registry. Since the shared workflows read the registry from the doc-builder@main checkout, the next main doc build here will pick it up as soon as #805 merges — no diffusers-side change needed for that part. Guarding the import as in this PR is still a good hardening on top.

mishig25:
Update: after #805 (sentencepiece + bitsandbytes) the build progressed and then hit `KeyError: 'optimum_quanto'` on `api/quantization.md`. Two findings:

1. huggingface/doc-builder#806 adds the quantization backends (`optimum`, `torchao`, `gguf`) as mocks — with `DIFFUSERS_SLOW_IMPORT=yes` the quantization page needs them importable, and mocking them means autodoc documents the real config classes rather than dummies. Verified the full 363-page build under exact CI conditions (`DIFFUSERS_SLOW_IMPORT=yes`, fresh venv, diffusers `--no-deps`).

2. The `KeyError` also exposed a latent diffusers bug the old container was masking: `dummy_optimum_quanto_objects.py` declares backend `"optimum_quanto"` but `BACKENDS_MAPPING` in `utils/import_utils.py` only has the key `"quanto"` — so when optimum-quanto is missing, `requires_backends` crashes with `KeyError` instead of raising the intended "requires optimum-quanto" ImportError. Worth a one-line fix here (add an `"optimum_quanto"` key or rename), independent of the doc build.

mishig25:
huggingface/doc-builder#806 is merged and the [main doc build](https://github.com/huggingface/diffusers/actions/runs/28853508918) now passes ✅ — diffusers is fully on the light path (no container, `--no-deps` install, sentencepiece real, torch/bitsandbytes/optimum/torchao/gguf mocked), and the build populated the page cache for the runs that follow.

Nothing in the docs pipeline depends on this PR anymore, but the import guard and the `optimum_quanto`/`quanto` `BACKENDS_MAPPING` mismatch noted above are still worth landing on their own merit.

DN6:
Thanks for flagging @mishig25 👍🏽 We're planning on removing quanto support, so that should resolve the issue.

sayakpaul:
Do we still need this change given https://github.com/huggingface/diffusers/pull/14134#issuecomment-4906813619?

sayakpaul (approved):

PR #14157, opened 2026-07-10
Add tests and agent docs for `kwargs_type` input/output

yiyixuxu:
add tests and docs for `kwargs_type`, which comes very handy for cosmos3

### DIFF
--- a/.ai/modular.md
+++ b/.ai/modular.md
@@ -102,6 +102,37 @@ class HeliosChunkDenoiseStep(HeliosChunkLoopWrapper):
 
 Note: sub-blocks inside `LoopSequentialPipelineBlocks` receive `(components, block_state, i, t)` for denoise loops or `(components, block_state, k)` for chunk loops.
 
+## Key pattern: `kwargs_type` bags (`denoiser_input_fields`)
+
+The conditioning inputs a denoiser needs often vary by workflow — especially for omni models like Cosmos3, where the action workflow requires additional action conditioning, and a workflow that generates sound along with video requires additional sound inputs. Tag these outputs with `kwargs_type="denoiser_input_fields"` when they are written, and have the denoiser declare the bag once and receive everything tagged — this avoids creating a new denoiser block for each workflow just to list its specific inputs:
+
+```python
+# producer side: standard conditioning outputs already carry the tag via their templates
+OutputParam.template("prompt_embeds")  # kwargs_type="denoiser_input_fields"
+# workflow-specific fields declare it explicitly
+OutputParam(
+    "action_embeds",
+    kwargs_type="denoiser_input_fields",
+    type_hint=torch.Tensor,
+    description="Action conditioning fed into the transformer.",
+)
+
+# consumer side (the loop denoiser): declare the bag once
+InputParam.template("denoiser_input_fields")
+
+# inside the denoiser __call__: every tagged value arrives in one dict —
+# and also individually (block_state.prompt_embeds, block_state.action_embeds, ...)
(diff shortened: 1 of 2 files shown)


.ai/modular.md
```diff
@@ -102,6 +102,37 @@ class HeliosChunkDenoiseStep(HeliosChunkLoopWrapper):
 
 Note: sub-blocks inside `LoopSequentialPipelineBlocks` receive `(components, block_state, i, t)` for denoise loops or `(components, block_state, k)` for chunk loops.
 
+## Key pattern: `kwargs_type` bags (`denoiser_input_fields`)
+
+The conditioning inputs a denoiser needs often vary by workflow — especially for omni models like Cosmos3, where the action workflow requires additional action conditioning, and a workflow that generates sound along with video requires additional sound inputs. Tag these outputs with `kwargs_type="denoiser_input_fields"` when they are written, and have the denoiser declare the bag once and receive everything tagged — this avoids creating a new denoiser block for each workflow just to list its specific inputs:
```
sayakpaul: What is a bag here?

tests/modular_pipelines/test_modular_pipelines_custom_blocks.py
```diff
@@ -578,6 +578,110 @@ def test_loop_block_requirements_save_load(self, tmp_path):
...
+            InputParam(kwargs_type="typea"),
+            InputParam(name="b", default=None),
+            InputParam(name="expected_a", default=None),
+            InputParam(name="expected_b", default=None),
+            InputParam(name="expected_typea", default=None),
+        ]
+
+    def __call__(self, components, state: PipelineState) -> PipelineState:
+        block_state = self.get_block_state(state)
+        assert block_state.typea == block_state.expected_typea
```
sayakpaul: I am a bit confused here. 

I'd imagine that the `block_state.typea` and `block_state.expected_typea` are different state-level components. But seems like they are the same. In what circumstance, a user would want to define input params like that?

tests/modular_pipelines/test_modular_pipelines_custom_blocks.py
```diff
@@ -578,6 +578,110 @@ def test_loop_block_requirements_save_load(self, tmp_path):
...
+    `OutputParam(..., kwargs_type=...)`, and a user-passed input is tagged if the pipeline's
+    `InputParam` for it declares a kwargs_type. A named input declared *without* a kwargs_type
+    therefore never reaches the bag, even though it is available in state by name.
+    """
+
+    def test_tagged_block_outputs_are_delivered_to_consumer(self):
+        blocks = SequentialPipelineBlocks.from_blocks_dict(
+            {"producer": DummyKwargsProducerStep(), "consumer": DummyKwargsConsumerStep()}
+        )
+        pipe = blocks.init_pipeline()
```
sayakpaul, line 648: When `pipe` is constructed, should we also assert against the params accepted by the `call` method of `pipe`?

sayakpaul (commented):
Thanks for the work and I left some comments.

I found it a bit confusing as a user as to when one should use, for example, `pipe(a="testa", ...)` and when one should use `pipe(typea={"a": "testa"})`. Maybe we could simplify that a bit (can be in a different PR and doesn't have to block this one).

tests/modular_pipelines/test_modular_pipelines_custom_blocks.py
yiyixuxu: it's for testing only, the `expcted_typea` is the value we expect to receive in block_state for `typea`, and we use only use it for `assert` statement inside the block - but i restructured the test to make it lesss confusing

PR #14158, opened 2026-07-10
[lora training] fix aspect ratio bucketing in dreambooth scripts (+ caption dropout, on-the-fly buckets)

linoytsaban:
Split from #14089. Bucketing fixes for the scripts that already support aspect-ratio buckets (flux2, flux2_img2img, flux2_klein(+img2img), flux_kontext, z_image):
- resize-to-cover then crop instead of resizing to exact bucket dims (which distorted images and made the crop a no-op)
- bucket class/prior-preservation images too (were forced to a fixed square)
- `--bucket_no_upscale`: pad images smaller than their bucket instead of upscaling
- caption dropout added to flux2_klein


examples/dreambooth/test_dreambooth_lora_flux2_klein.py
```diff
@@ -260,3 +260,55 @@ def test_dreambooth_lora_with_metadata(self):
             self.assertTrue(loaded_lora_alpha == lora_alpha)
             loaded_lora_rank = raw["transformer.r"]
             self.assertTrue(loaded_lora_rank == rank)
+
+    def test_dreambooth_lora_flux2_aspect_ratio_buckets(self):
```
sayakpaul, line 266: I think we can remove this test for now. Let's see if this feature becomes more and more used and then add it back. So, we could just `skip` it.

examples/dreambooth/test_dreambooth_lora_flux2_klein.py
```diff
@@ -260,3 +260,55 @@ def test_dreambooth_lora_with_metadata(self):
...
+                --output_dir {tmpdir}
+                """.split()
+
+            run_command(self._launch_args + test_args)
+            self.assertTrue(os.path.isfile(os.path.join(tmpdir, "pytorch_lora_weights.safetensors")))
+            lora_state_dict = safetensors.torch.load_file(os.path.join(tmpdir, "pytorch_lora_weights.safetensors"))
+            self.assertTrue(all("lora" in k for k in lora_state_dict.keys()))
+            self.assertTrue(all(key.startswith("transformer") for key in lora_state_dict.keys()))
+
+    def test_dreambooth_lora_flux2_caption_dropout(self):
```
sayakpaul, line 293: Same for this.

examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -450,6 +450,11 @@ def parse_args(input_args=None):
             "Images will be resized and cropped to fit the nearest bucket. If provided, --resolution is ignored."
         ),
     )
+    parser.add_argument(
+        "--bucket_no_upscale",
+        action="store_true",
+        help="If set, images smaller than their aspect-ratio bucket are padded instead of upscaled.",
+    )
```
sayakpaul, line 463: Does this help at all? If so, I would maybe keep it simple and just upscale. Also, should we enable bucketing through an arg?

examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -890,15 +895,6 @@ def __init__(
...
             self.class_data_root = None
 
-        self.image_transforms = transforms.Compose(
-            [
-                transforms.Resize(size, interpolation=transforms.InterpolationMode.BILINEAR),
-                transforms.CenterCrop(size) if center_crop else transforms.RandomCrop(size),
-                transforms.ToTensor(),
-                transforms.Normalize([0.5], [0.5]),
-            ]
-        )
```
sayakpaul, line 900: If we're going with the above suggestion i.e. make aspect-ratio bucketing conditional on an arg, then I guess this should stay?

sayakpaul (commented):
Left some comments. LMK.

examples/dreambooth/train_dreambooth_lora_flux2.py
linoytsaban: Yeah I agree, it made me also rethink how currently we ask users to pass a list to enable the buckets so I reworked this rather than keeping `--bucket_no_upscale`. 
**New behavior:**
- Bucketing is now enabled via this arg: **`--use_aspect_ratio_buckets`**. 
But so we its not a breaking change, passing an explicit `--aspect_ratio_buckets` list also turns it on.
- If **no list** is provided, buckets are computed **on the fly** from `--resolution`, ostris-style: the ladder is capped to each image's own resolution (`min(resolution, sqrt(h*w))`), so smaller images land in a *smaller* bucket instead of being upscaled or padded. This let me drop `--bucket_no_upscale` and the padding branch entirely.
- **List provided** → use it, resize-to-cover + crop, upscale small images with no padding.
- Flux Kontext seeds its on-the-fly ladder from the pipeline's own `PREFERRED_KONTEXT_RESOLUTIONS`; the others generate from a `resolution²` pixel budget across a standard spread of aspect ratios. Ladder dims are rounded to 16.

The generator lives in `training_utils.generate_aspect_ratio_buckets(...)` so all the scripts share it.

(For reference, ostris avoids upscaling the same way (cap the bucket resolution to the image's own resolution and pick a smaller bucket rather than padding), it can do that because it generates buckets on the fly, which is what we now do too. kohya exposes the same idea as its own `--bucket_no_upscale`.)

examples/dreambooth/test_dreambooth_lora_flux2_klein.py
linoytsaban: Skipped it (`@unittest.skip`) — will re-enable once bucketing sees more use. Also removed the now-gone `--bucket_no_upscale` flag from it so it's not stale if un-skipped.

examples/dreambooth/test_dreambooth_lora_flux2_klein.py
linoytsaban: Skipped this one too.

examples/dreambooth/train_dreambooth_lora_flux2.py
linoytsaban: It can stay removed, bucketing is now arg-gated (`--use_aspect_ratio_buckets` / `--aspect_ratio_buckets`), and the non-bucketing case is handled by a single `[(resolution, resolution)]` bucket that routes through the same transform — so the old fixed-size `image_transforms` path is fully covered by the fallback and isn't needed separately.

examples/dreambooth/train_dreambooth_lora_flux2.py
```diff
@@ -446,8 +447,18 @@ def parse_args(input_args=None):
...
             "Aspect ratio buckets to use for training. Define as a string of 'h1,w1;h2,w2;...'. "
-            "e.g. '1024,1024;768,1360;1360,768;880,1168;1168,880;1248,832;832,1248'"
-            "Images will be resized and cropped to fit the nearest bucket. If provided, --resolution is ignored."
+            "e.g. '1024,1024;768,1360;1360,768;880,1168;1168,880;1248,832;832,1248'. "
+            "Passing this enables aspect-ratio bucketing; images are resized to cover and cropped to the "
+            "nearest listed bucket (smaller images are upscaled). If provided, --resolution is ignored."
+        ),
+    )
+    parser.add_argument(
+        "--use_aspect_ratio_buckets",
```
sayakpaul, line 456: If aspect-ratio bucket lists are provided without `--use_aspect_ratio_buckets` then we should also raise an error.

sayakpaul (approved):
Just one comment. Phenonmenal work. I guess we could also move this modern dataloader to `utils.py` or `training_data_utils.py` to remove some boilerplate code. Doesn't have to be done in this PR, though.

examples/dreambooth/train_dreambooth_lora_flux2.py
linoytsaban: Done! A list passed without `--use_aspect_ratio_buckets` now raises a `ValueError`. Also reworded the auto-compute info log and updated the flux2/z_image READMEs.

PR #14160, opened 2026-07-10
Cast attention inputs to the autocast dtype in dispatch_attention_fn

### LINKED ISSUE #14104: Flux-family attention: flash-attn and other backends fail in an autocast context
### Describe the bug

- q and k are normed before attention in flux-style attention.
- in an autocast context, norms are calculated in float32 no matter their input - and are returned as float32
- the result gets passed on to attention
- no problem for torch scaled_dot_product_attention, it can handle float32 - doesn't fail, and still executes in low precision in an autocast context
- other attention backends don't though. even though there are checks (if activated), it still fails with a check then.

Not sure about the solution.

1. avoid propagation to float32 for norms, as some other functions do?
2. autocast-like behaviour in the attention backend to accept float32 there, even though flash-attn itself doesn't?

Details:

`FluxAttnProcessor.__call__` (and the equivalent processors for Flux2, Chroma, Krea2, etc., which share the same `norm_q`/`norm_k` + `apply_rotary_emb` + `dispatch_attention_fn` pattern) can hand `dispatch_attention_fn` a `query`/`key` that ended up in `float32` even though the model is meant to run in `bfloat16`, whenever the whole forward pass runs inside an active `torch.autocast` context (e.g. this happens routinely in training code that mixes weight dtypes across parts of the model).

Root cause, in order:

1. `attn.norm_q` / `attn.norm_k` are `torch.nn.RMSNorm`. Under an active `torch.autocast`, `RMSNorm` is on PyTorch's "compute in fp32" autocast policy list, so it returns an `fp32` tensor even when its input was `bf16`. This part is intentional, documented PyTorch autocast behavior, not itself a bug.
2. `apply_rotary_emb` [...] is a no-op in terms of dtype
3. `value` never goes through a norm or RoPE, so it stays `bf16`.
4. `dispatch_attention_fn` is called with `query`/`key` in `fp32` and `value` in `bf16`.

For the `native`/`_native_cudnn`/`_native_efficient`/`_native_flash` (torch-SDPA-based) backends this is harmless: `torch.nn.functional.scaled_dot_product_attention` is itself a PyTorch op with a registered autocast policy, so under an active autocast context it downcasts **all** of its inputs to the autocast dtype before running, regardless of what dtype they arrive in (verified below — this holds even under an explicit `sdpa_kernel()` backend restriction).

But several other backends call an external kernel function directly (`flash_attn_func` for `AttentionBackendName.FLASH`, and similarly for the hub/sage/xformers/aiter/npu/xla variants) — these are not PyTorch ops with an autocast policy, so no such implicit cast happens. They receive the raw fp32 query/key + bf16 value:
crash inside the external kernel's own dtype assertion (e.g. flash-attn's `"FlashAttention only support fp16 and bf16 data type"`), or

---

the latter part *Drafted by Claude* but shortened

### Reproduction

for example, run Chroma in an autocast context, install flash-attn, set the backend

### Logs

```
  File "\OneTrainer\modules\ui\TrainUI.py", line 716, in __training_thread_function
    trainer.train()
  File "\OneTrainer\modules\trainer\GenericTrainer.py", line 744, in train
    model_output_data = self.model_setup.predict(self.model, batch, self.config, train_progress)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\modules\modelSetup\BaseChromaSetup.py", line 241, in predict
    packed_predicted_flow = model.transformer(
                            ^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\nn\modules\module.py", line 1773, in _wrapped_call_impl
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\nn\modules\module.py", line 1784, in _call_impl
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\expermential guff\venv\src\diffusers\src\diffusers\models\transformers\transformer_chroma.py", line 577, in forward
    encoder_hidden_states, hidden_states = block(
                                           ^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\nn\modules\module.py", line 1771, in _wrapped_call_impl
    return self._compiled_call_impl(*args, **kwargs)  # type: ignore[misc]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_dynamo\eval_frame.py", line 736, in compile_wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\nn\modules\module.py", line 1784, in _call_impl
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\modules\util\checkpointing_util.py", line 90, in forward
    def forward(self, *args, **kwargs):
  File "\OneTrainer\venv\Lib\site-packages\torch\_dynamo\eval_frame.py", line 929, in _fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_functorch\aot_autograd.py", line 1241, in forward
    return compiled_fn(full_args)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_functorch\_aot_autograd\runtime_wrappers.py", line 370, in runtime_wrapper
    all_outs = call_func_at_runtime_with_args(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_functorch\_aot_autograd\utils.py", line 126, in call_func_at_runtime_with_args
    out = normalize_as_list(f(args))
                            ^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_functorch\_aot_autograd\utils.py", line 100, in g
    return f(*args)
           ^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\autograd\function.py", line 576, in apply
    return super().apply(*args, **kwargs)  # type: ignore[misc]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_functorch\_aot_autograd\runtime_wrappers.py", line 2074, in forward
    fw_outs = call_func_at_runtime_with_args(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_functorch\_aot_autograd\utils.py", line 126, in call_func_at_runtime_with_args
    out = normalize_as_list(f(args))
                            ^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_functorch\_aot_autograd\runtime_wrappers.py", line 556, in wrapper
    return compiled_fn(runtime_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_functorch\_aot_autograd\runtime_wrappers.py", line 750, in inner_fn
    outs = compiled_fn(args)
           ^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_inductor\output_code.py", line 584, in __call__
    return self.current_callable(inputs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_inductor\utils.py", line 2716, in run
    out = model(new_inputs)
          ^^^^^^^^^^^^^^^^^
  File "\AppData\Local\Temp\torchinductor\bq\cbqrfpkp4z55klkpln3fescrhmpvo7357m3j6qqgugdzy6ldyfze.py", line 3114, in call
    buf137 = torch.ops.flash_attn._flash_attn_forward.default(buf135, buf136, buf134, 0.0, 0.08838834764831845, False, -1, -1, 0.0, None, False)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_ops.py", line 829, in __call__
    return self._op(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_library\autograd.py", line 111, in autograd_impl
    result = forward_no_grad(*args, Metadata(keyset, keyword_only_args))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_library\autograd.py", line 40, in forward_no_grad
    result = op.redispatch(keyset & _C._after_autograd_keyset, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_ops.py", line 836, in redispatch
    return self._handle.redispatch_boxed(keyset, *args, **kwargs)  # type: ignore[return-value]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_library\custom_ops.py", line 344, in backend_impl
    result = self._backend_fns[device_type](*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_compile.py", line 53, in inner
    return disable_fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_dynamo\eval_frame.py", line 929, in _fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\torch\_library\custom_ops.py", line 377, in wrapped_fn
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "\OneTrainer\venv\Lib\site-packages\flash_attn\flash_attn_interface.py", line 91, in _flash_attn_forward
    out, softmax_lse, S_dmask, rng_state = flash_attn_gpu.fwd(
                                           ^^^^^^^^^^^^^^^^^^^
RuntimeError: FlashAttention only support fp16 and bf16 data type
Exception raised from mha_fwd at C:\Users\junya\actions-runner\_work\flash-attention-prebuild-wheels\flash-attention-prebuild-wheels\flash-attention\csrc\flash_attn\flash_api.cpp:374 (most recent call first):
00007FFC8050199400007FFC805018F0 c10.dll!c10::Error::Error [<unknown file> @ <unknown line number>]
00007FFC8050040A00007FFC805003B0 c10.dll!c10::detail::torchCheckFail [<unknown file> @ <unknown line number>]
00007FFAF25B42FD00007FFAF25A5EA0 flash_attn_2_cuda.cp312-win_amd64.pyd!c10::ivalue::Object::operator= [<unknown file> @ <unknown line number>]
00007FFAF25C3C9B00007FFAF25BD050 flash_attn_2_cuda.cp312-win_amd64.pyd!PyInit_flash_attn_2_cuda [<unknown file> @ <unknown line number>]
00007FFAF25C3DB400007FFAF25BD050 flash_attn_2_cuda.cp312-win_amd64.pyd!PyInit_flash_attn_2_cuda [<unknown file> @ <unknown line number>]
00007FFAF25AF2BC00007FFAF25A5EA0 flash_attn_2_cuda.cp312-win_amd64.pyd!c10::ivalue::Object::operator= [<unknown file> @ <unknown line number>]
00007FFC780B79E400007FFC780B73F4 python312.dll!PyThread_acquire_lock_timed [<unknown file> @ <unknown line number>]
00007FFC7810201800007FFC78101990 python312.dll!PyObject_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC781019C500007FFC78101990 python312.dll!PyObject_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC78102EA500007FFC78102610 python312.dll!PyEval_EvalFrameDefault [<unknown file> @ <unknown line number>]
00007FFC7810103C00007FFC78100EC0 python312.dll!PyFunction_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC7813281500007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC7813275F00007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC60AF7BBA00007FFC60ACFF50 torch_python.dll!torch::FunctionParameter::type_name [<unknown file> @ <unknown line number>]
00007FFC60AEDAC800007FFC60ACFF50 torch_python.dll!torch::FunctionParameter::type_name [<unknown file> @ <unknown line number>]
00007FFC60AF15EB00007FFC60ACFF50 torch_python.dll!torch::FunctionParameter::type_name [<unknown file> @ <unknown line number>]
00007FFC60AD17D600007FFC60ACFF50 torch_python.dll!torch::FunctionParameter::type_name [<unknown file> @ <unknown line number>]
00007FFC601E25F600007FFC601E1730 torch_python.dll!c10::ivalue::Future::devices [<unknown file> @ <unknown line number>]
00007FFC780B79E400007FFC780B73F4 python312.dll!PyThread_acquire_lock_timed [<unknown file> @ <unknown line number>]
00007FFC780E560E00007FFC780E556C python312.dll!PyObject_MakeTpCall [<unknown file> @ <unknown line number>]
00007FFC7824673800007FFC78223794 python312.dll!PyThread_tss_is_created [<unknown file> @ <unknown line number>]
00007FFC7813281500007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC7813275F00007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC781066EE00007FFC78102610 python312.dll!PyEval_EvalFrameDefault [<unknown file> @ <unknown line number>]
00007FFC7810103C00007FFC78100EC0 python312.dll!PyFunction_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC780FBD8E00007FFC780FB9BC python312.dll!PyArg_CheckPositional [<unknown file> @ <unknown line number>]
00007FFC7813281500007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC7813275F00007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC781066EE00007FFC78102610 python312.dll!PyEval_EvalFrameDefault [<unknown file> @ <unknown line number>]
00007FFC7810103C00007FFC78100EC0 python312.dll!PyFunction_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC7813281500007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC7813275F00007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC60AF7B2F00007FFC60ACFF50 torch_python.dll!torch::FunctionParameter::type_name [<unknown file> @ <unknown line number>]
00007FFB5B08924100007FFB5B089080 torch_cpu.dll!c10::Dispatcher::callBoxed [<unknown file> @ <unknown line number>]
00007FFC608BF02500007FFC608BEF60 torch_python.dll!torch::jit::invokeOperatorFromPython [<unknown file> @ <unknown line number>]
00007FFC608BC08D00007FFC608BBF50 torch_python.dll!torch::jit::_get_operation_for_overload_or_packet [<unknown file> @ <unknown line number>]
00007FFC6081AFCA00007FFC608153F0 torch_python.dll!c10d::PythonOnCompletionHook::PythonOnCompletionHook [<unknown file> @ <unknown line number>]
00007FFC6082F6C100007FFC608153F0 torch_python.dll!c10d::PythonOnCompletionHook::PythonOnCompletionHook [<unknown file> @ <unknown line number>]
00007FFC607C820600007FFC6074CED0 torch_python.dll!THPPointer<_frame>::dup [<unknown file> @ <unknown line number>]
00007FFC601E25F600007FFC601E1730 torch_python.dll!c10::ivalue::Future::devices [<unknown file> @ <unknown line number>]
00007FFC780B79E400007FFC780B73F4 python312.dll!PyThread_acquire_lock_timed [<unknown file> @ <unknown line number>]
00007FFC781327A600007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC781066EE00007FFC78102610 python312.dll!PyEval_EvalFrameDefault [<unknown file> @ <unknown line number>]
00007FFC7810103C00007FFC78100EC0 python312.dll!PyFunction_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC78133E8B00007FFC78133E18 python312.dll!PyObject_FastCallDictTstate [<unknown file> @ <unknown line number>]
00007FFC781D319B00007FFC781D311C python312.dll!PyObject_Call_Prepend [<unknown file> @ <unknown line number>]
00007FFC781D30C600007FFC781D2DF0 python312.dll!PyDictProxy_New [<unknown file> @ <unknown line number>]
00007FFC7810201800007FFC78101990 python312.dll!PyObject_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC781019C500007FFC78101990 python312.dll!PyObject_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC78102EA500007FFC78102610 python312.dll!PyEval_EvalFrameDefault [<unknown file> @ <unknown line number>]
00007FFC7810103C00007FFC78100EC0 python312.dll!PyFunction_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC78133E8B00007FFC78133E18 python312.dll!PyObject_FastCallDictTstate [<unknown file> @ <unknown line number>]
00007FFC781D319B00007FFC781D311C python312.dll!PyObject_Call_Prepend [<unknown file> @ <unknown line number>]
00007FFC781D30C600007FFC781D2DF0 python312.dll!PyDictProxy_New [<unknown file> @ <unknown line number>]
00007FFC7810201800007FFC78101990 python312.dll!PyObject_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC781019C500007FFC78101990 python312.dll!PyObject_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC78102EA500007FFC78102610 python312.dll!PyEval_EvalFrameDefault [<unknown file> @ <unknown line number>]
00007FFC7810103C00007FFC78100EC0 python312.dll!PyFunction_Vectorcall [<unknown file> @ <unknown line number>]
00007FFC7813281500007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC780893B700007FFC7808931C python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC6069275F00007FFC6068D170 torch_python.dll!torch::autograd::registerFunctionPreHook [<unknown file> @ <unknown line number>]
00007FFC780B7A1600007FFC780B73F4 python312.dll!PyThread_acquire_lock_timed [<unknown file> @ <unknown line number>]
00007FFC781327A600007FFC781326F0 python312.dll!PyObject_Call [<unknown file> @ <unknown line number>]
00007FFC781066EE00007FFC78102610 python312.dll!PyEval_EvalFrameDefault [<unknown file> @ <unknown line number>]
```

### System Info

6d71b76aceff935192e58fee38c5cc5d8d227cf0

### Who can help?

_No response_

ghost:
# What does this PR do?

Fixes #14104.

Starting with torch 2.12, `aten::rms_norm` registers an fp32 cast policy for `AutocastCUDA`/`AutocastXPU`. Bisected via `torch._C._dispatch_dump("aten::rms_norm")`: absent on 2.6 through 2.11, present on 2.12, and verified behaviorally on GPU for 2.8 vs 2.12. The policy is intentional on the torch side (pytorch/pytorch#174824, a numerics fix), currently debated as a memory/throughput regression in pytorch/pytorch#188955, but torch 2.12 stable ships with it either way. Under an active `torch.autocast` context the `torch.nn.RMSNorm` QK norms used by Flux-family attention processors therefore return float32 even for bf16/fp16 inputs, while `value`, which never passes through a norm, keeps the autocast dtype. `dispatch_attention_fn` forwarded that fp32/bf16 mix untouched:

- native backends worked: `aten::scaled_dot_product_attention` has its own autocast registration that downcasts its autocast-eligible floating inputs;
- external-kernel backends crashed: `flash`, `flash_hub`, `sage`, `aiter`, `xformers`, `npu`, `xla` call the kernel directly, outside autocast's dispatch handling. flash-attn dies with `RuntimeError: FlashAttention only support fp16 and bf16 data type`, and `flex` rejects the dtype mix as well;
- even the opt-in `DIFFUSERS_ATTN_CHECKS` dtype checks false-raised on the native backend under autocast, although the op itself would have run fine.

The affected pattern (`torch.nn.RMSNorm` QK-norm, then RoPE, then `dispatch_attention_fn`) appears in 18 model files: Flux, Flux2, Chroma, Bria, BriaFibo, OvisImage, LongCatImage, LTX, LTX2, Wan, WanAnimate, ChronoEdit, SkyReelsV2, Helios, JoyImage, JoyImage Edit Plus (merged July 8, a week after the issue was filed), MotifVideo, ErnieImage, plus the Anima condition embedder. So this fixes the dispatcher once rather than patching every processor: when autocast is active for the tensors' device type, `dispatch_attention_fn` now casts q/k/v and floating-point masks to `torch.get_autocast_dtype(device_type)`, mirroring SDPA's own autocast registration, including its eligibility rule (float64 tensors and bool/int masks pass through untouched). Since `dispatch_attention_fn` is the documented drop-in for `F.scaled_dot_product_attention`, this is parity with the op it replaces, not new policy:

- numerically a no-op for the native backend (SDPA performed this exact cast internally);
- provably dead branch outside autocast: mismatched dtypes still raise exactly as before, covered by a test;
- also repairs the context-parallel `_native_cudnn`/`_native_flash` paths, which call the decomposed aten ops directly and bypassed the composite's autocast handling;
- anyone who wants fp32 attention inside an autocast region uses `torch.autocast(..., enabled=False)`, same as with raw SDPA.

Two implementation details worth flagging for review:

- `torch.is_autocast_enabled(device_type)` raises for device types without an autocast dispatch key. The only such device type diffusers tensors realistically report is `meta` (meta-device forward works on main and keeps working), so the guard is a literal `device_type != "meta"` check. Everything left in the branch is Dynamo-traceable on torch 2.6, the repo floor: `torch.compile(dispatch_attention_fn, fullgraph=True)` traces with zero graph breaks, autocast on and off, verified on 2.6.0 and 2.12.1 and locked in by a `@is_torch_compile` regression test.
- Autocast is a first-party workflow (81 files under `examples/` use it, including Flux LoRA training validation), and this failure hits any trainer that selects a non-native attention backend. The issue was reported from OneTrainer, which has since made mixed precision always-on (Nerogar/OneTrainer#1481), so their whole user base lands on this path once they move to torch 2.12.

## Tests

New `TestAttentionDispatchAutocast` (plain pytest style, gated behind `require_torch_accelerator` so it stays off the fast PR CI) in `tests/models/test_attention_processor.py`, the first unit tests for `dispatch_attention_fn`:

- `test_autocast_casts_inputs_to_autocast_dtype`: mixed fp32 q/k + bf16 v under autocast, with the `DIFFUSERS_ATTN_CHECKS` validation enabled: the backend fn must receive uniform bf16 and the checks must not false-raise (both fail on main);
- `test_autocast_casts_floating_point_mask_but_not_bool_mask` (fails on main);
- `test_no_autocast_dtypes_pass_through_unchanged`: outside autocast the dispatcher still doesn't touch dtypes, the mismatch keeps raising SDPA's exact "same dtype" message;
- `test_rms_norm_qk_pattern_under_autocast` (`skipif` CUDA/XPU only): end-to-end shape of the bug through a real `torch.nn.RMSNorm`, asserting the fp32 upcast actually happens (guards against PyTorch policy changes hollowing the test) and that the backend receives bf16;
- `test_torch_compile_fullgraph_under_autocast` (`@is_torch_compile`): the autocast branch stays fullgraph-traceable.

Validation:

```
pytest tests/models/test_attention_processor.py::TestAttentionDispatchAutocast -v
  # CUDA (RTX A4000): 5 passed on torch 2.12.1+cu126, 5 passed on torch 2.13.0+cu126
  # CPU boxes: the class skips; with the accelerator gate lifted, the CPU-runnable
  # tests pass on torch 2.6.0 and 2.12.1, fail on main
pytest tests/models/transformers/test_models_transformer_flux.py \
       tests/models/transformers/test_models_transformer_flux2.py \
       tests/models/transformers/test_models_transformer_chroma.py  # 93 passed, 358 skipped
ruff check / ruff format --check / utils/check_copies.py  # clean
```

GPU validation (CUDA, real flash-attn2 kernels, RunPod RTX A4000, torch 2.12.1+cu126):

```
main:  FluxAttention forward under cuda autocast, backend=flash_hub
       -> RuntimeError: FlashAttention only support fp16 and bf16 data type
fixed: RMSNorm(bf16) under cuda autocast -> torch.float32 (policy confirmed on CUDA)
       FluxAttention forward with backend=flash_hub: OK, out=torch.bfloat16
       max |flash_hub - native| under autocast: < 1e-5
       TestAttentionDispatchAutocast on CUDA: 5 passed
       meta-device forward through the dispatcher: works, with and without autocast active
```

## Limitations

- Backends that only support fp16/bf16 still (correctly) raise if the autocast dtype itself can't be handled by the kernel. Unchanged semantics, same as native.
- The rms_norm fp32 policy is CUDA/XPU-only, so the CPU regression tests construct the dtype mix explicitly; the accelerator-gated test covers the real policy path.

## Before submitting

- [x] Did you use an AI agent to help with this PR? Yes, Claude Code.
  - [x] Read the "Coding with AI agents" guide.
- [x] Read the contributor guideline.
- [x] Coordination: maintainer go-ahead in https://github.com/huggingface/diffusers/issues/14104#issuecomment-4934552848 (full analysis + proposal in https://github.com/huggingface/diffusers/issues/14104#issuecomment-4886284118)
- [x] Wrote new necessary tests.
- Documentation: no doc changes needed. `docs/source/en/optimization/attention_backends.md` describes backend selection and opt-in checks, neither of which changes; autocast now simply behaves consistently across backends.

## Who can review?

@sayakpaul (attention dispatcher) @DN6 (recent `attention_dispatch.py` fixes)

### DIFF
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -408,16 +408,13 @@ def dispatch_attention_fn(
 ) -> torch.Tensor:
     attention_kwargs = attention_kwargs or {}
 
-    # Under autocast, RMSNorm QK norms return fp32 while value stays bf16/fp16. Torch SDPA casts
-    # its inputs to the autocast dtype internally, but backends calling external kernels
-    # (flash-attn, sage, ...) receive the mix as-is and crash, so do the same cast here.
-    # float64 is not autocast-eligible. See https://github.com/huggingface/diffusers/issues/14104.
-    # is_compiling() short-circuits the is_autocast_available guard (it protects eager runs on
-    # devices without autocast, e.g. "meta"; Dynamo can't trace it before torch 2.12).
+    # Under autocast, fp32-cast-policy ops (like the RMSNorm QK norms in Flux models) return fp32
+    # while `value` stays bf16/fp16. SDPA casts its inputs inside its C++ autocast kernel; backends
+    # calling external kernels (flash-attn, sage, ...) crash on the mix, so do the same cast here.
+    # float64 is not autocast-eligible; "meta" has no autocast key and `is_autocast_enabled` raises.
+    # See https://github.com/huggingface/diffusers/issues/14104.
     device_type = query.device.type
-    if (torch.compiler.is_compiling() or torch.amp.is_autocast_available(device_type)) and torch.is_autocast_enabled(
-        device_type
-    ):
+    if device_type != "meta" and torch.is_autocast_enabled(device_type):
         autocast_dtype = torch.get_autocast_dtype(device_type)
         if query.dtype != torch.float64:
             query, key, value = query.to(autocast_dtype), key.to(autocast_dtype), value.to(autocast_dtype)
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -408,6 +408,26 @@ def dispatch_attention_fn(
 ) -> torch.Tensor:
     attention_kwargs = attention_kwargs or {}
 
+    # Match the autocast behavior of `torch.nn.functional.scaled_dot_product_attention`: its autocast
+    # registration casts eligible floating-point inputs (autocast skips float64) to the autocast dtype
+    # before running. Under an active autocast context, ops with an fp32 cast policy (such as the
+    # `torch.nn.RMSNorm` QK norms in Flux-family models) return float32 even for bf16/fp16 inputs while
+    # `value` keeps the autocast dtype, and backends that call external kernels (flash-attn, sage,
+    # xformers, ...) bypass autocast's dispatch handling and fail on such inputs. Normalizing here keeps
+    # every backend consistent with the native backend. See https://github.com/huggingface/diffusers/issues/14104.
+    # `is_autocast_available` guards device types without an autocast dispatch key (e.g. "meta"), for
+    # which `is_autocast_enabled` raises. It is skipped when compiling: Dynamo cannot trace it before
+    # torch 2.12, and compiled graphs never run on such devices anyway.
+    device_type = query.device.type
+    if (torch.compiler.is_compiling() or torch.amp.is_autocast_available(device_type)) and torch.is_autocast_enabled(
+        device_type
+    ):
+        autocast_dtype = torch.get_autocast_dtype(device_type)
+        if query.dtype != torch.float64:
+            query, key, value = query.to(autocast_dtype), key.to(autocast_dtype), value.to(autocast_dtype)
+        if attn_mask is not None and torch.is_floating_point(attn_mask) and attn_mask.dtype != torch.float64:
+            attn_mask = attn_mask.to(autocast_dtype)
+
     if backend is None:
         # If no backend is specified, we either use the default backend (set via the DIFFUSERS_ATTN_BACKEND environment
         # variable), or we use a custom backend based on whether user is using the `attention_backend` context manager
--- a/src/diffusers/models/attention_dispatch.py
+++ b/src/diffusers/models/attention_dispatch.py
@@ -408,16 +408,12 @@ def dispatch_attention_fn(
 ) -> torch.Tensor:
     attention_kwargs = attention_kwargs or {}
 
-    # Match the autocast behavior of `torch.nn.functional.scaled_dot_product_attention`: its autocast
-    # registration casts eligible floating-point inputs (autocast skips float64) to the autocast dtype
-    # before running. Under an active autocast context, ops with an fp32 cast policy (such as the
-    # `torch.nn.RMSNorm` QK norms in Flux-family models) return float32 even for bf16/fp16 inputs while
-    # `value` keeps the autocast dtype, and backends that call external kernels (flash-attn, sage,
-    # xformers, ...) bypass autocast's dispatch handling and fail on such inputs. Normalizing here keeps
-    # every backend consistent with the native backend. See https://github.com/huggingface/diffusers/issues/14104.
-    # `is_autocast_available` guards device types without an autocast dispatch key (e.g. "meta"), for
-    # which `is_autocast_enabled` raises. It is skipped when compiling: Dynamo cannot trace it before
-    # torch 2.12, and compiled graphs never run on such devices anyway.
+    # Under autocast, RMSNorm QK norms return fp32 while value stays bf16/fp16. Torch SDPA casts
+    # its inputs to the autocast dtype internally, but backends calling external kernels
+    # (flash-attn, sage, ...) receive the mix as-is and crash, so do the same cast here.
+    # float64 is not autocast-eligible. See https://github.com/huggingface/diffusers/issues/14104.
+    # is_compiling() short-circuits the is_autocast_available guard (it protects eager runs on
+    # devices without autocast, e.g. "meta"; Dynamo can't trace it before torch 2.12).
     device_type = query.device.type
     if (torch.compiler.is_compiling() or torch.amp.is_autocast_available(device_type)) and torch.is_autocast_enabled(
         device_type
--- a/tests/models/test_attention_processor.py
+++ b/tests/models/test_attention_processor.py
@@ -1,3 +1,4 @@
+import contextlib
 import importlib.metadata
 import tempfile
 import unittest
@@ -8,9 +9,14 @@ import torch
 from packaging import version
 
 from diffusers import DiffusionPipeline
+from diffusers.models.attention_dispatch import (
+    AttentionBackendName,
+    _AttentionBackendRegistry,
+    dispatch_attention_fn,
+)
 from diffusers.models.attention_processor import Attention, AttnAddedKVProcessor
 
-from ..testing_utils import torch_device
+from ..testing_utils import is_torch_compile, require_torch_accelerator, torch_device
 
 
 class AttnAddedKVProcessorTests(unittest.TestCase):
@@ -133,3 +139,134 @@ class DeprecatedAttentionBlockTests(unittest.TestCase):
 
         self.assertTrue(np.allclose(pre_conversion, conversion, atol=1e-3))
         self.assertTrue(np.allclose(conversion, after_conversion, atol=1e-3))
+
+
+class AttentionDispatchAutocastTests(unittest.TestCase):
+    """Regression tests for https://github.com/huggingface/diffusers/issues/14104.
+
+    Under an active torch.autocast context, ops with an fp32 cast policy (e.g. the `torch.nn.RMSNorm`
+    QK norms in Flux-family models) return float32 while `value` keeps the autocast dtype. Native SDPA
+    downcasts all of its inputs itself, but backends calling external kernels (flash-attn, sage, ...)
+    receive the tensors as-is and reject the mismatch — so `dispatch_attention_fn` must normalize
+    dtypes the same way SDPA's autocast registration does.
+    """
+
+    def _qkv(self, qk_dtype=torch.float32, v_dtype=torch.bfloat16):
+        query = torch.randn(1, 8, 2, 16, device=torch_device, dtype=qk_dtype)
+        key = torch.randn(1, 8, 2, 16, device=torch_device, dtype=qk_dtype)
+        value = torch.randn(1, 8, 2, 16, device=torch_device, dtype=v_dtype)
+        return query, key, value
+
+    @contextlib.contextmanager
+    def _record_native_backend_input_dtypes(self):
+        original = _AttentionBackendRegistry._backends[AttentionBackendName.NATIVE]
+        received = {}
+
+        def recording_backend(query, key, value, **kwargs):
+            received["query"], received["key"], received["value"] = query.dtype, key.dtype, value.dtype
+            if kwargs.get("attn_mask") is not None:
+                received["attn_mask"] = kwargs["attn_mask"].dtype
+            return original(query, key, value, **kwargs)
+
+        _AttentionBackendRegistry._backends[AttentionBackendName.NATIVE] = recording_backend
+        try:
+            yield received
+        finally:
+            _AttentionBackendRegistry._backends[AttentionBackendName.NATIVE] = original
+
+    def test_autocast_casts_inputs_to_autocast_dtype(self):
+        query, key, value = self._qkv()
+        device_type = torch.device(torch_device).type
+
+        with self._record_native_backend_input_dtypes() as received:
+            with torch.autocast(device_type, dtype=torch.bfloat16):
+                out = dispatch_attention_fn(query, key, value, backend=AttentionBackendName.NATIVE)
+
+        self.assertEqual(received["query"], torch.bfloat16)
+        self.assertEqual(received["key"], torch.bfloat16)
+        self.assertEqual(received["value"], torch.bfloat16)
+        self.assertEqual(out.dtype, torch.bfloat16)
+
+    def test_autocast_casts_floating_point_mask_but_not_bool_mask(self):
+        query, key, value = self._qkv()
+        device_type = torch.device(torch_device).type
+        float_mask = torch.zeros(1, 2, 8, 8, device=torch_device, dtype=torch.float32)
+        bool_mask = torch.ones(1, 2, 8, 8, device=torch_device, dtype=torch.bool)
+
+        with self._record_native_backend_input_dtypes() as received:
+            with torch.autocast(device_type, dtype=torch.bfloat16):
+                dispatch_attention_fn(query, key, value, attn_mask=float_mask, backend=AttentionBackendName.NATIVE)
+        self.assertEqual(received["attn_mask"], torch.bfloat16)
+
+        with self._record_native_backend_input_dtypes() as received:
+            with torch.autocast(device_type, dtype=torch.bfloat16):
+                dispatch_attention_fn(query, key, value, attn_mask=bool_mask, backend=AttentionBackendName.NATIVE)
+        self.assertEqual(received["attn_mask"], torch.bool)
+
+    def test_autocast_passes_backend_dtype_checks(self):
+        # The opt-in debugging checks (DIFFUSERS_ATTN_CHECKS) validate the tensors the backend will
+        # actually receive, so mixed autocast inputs must not trip them.
+        query, key, value = self._qkv()
+        device_type = torch.device(torch_device).type
+
+        original = _AttentionBackendRegistry._checks_enabled
+        _AttentionBackendRegistry._checks_enabled = True
+        try:
+            with torch.autocast(device_type, dtype=torch.bfloat16):
+                out = dispatch_attention_fn(query, key, value, backend=AttentionBackendName.NATIVE)
+        finally:
+            _AttentionBackendRegistry._checks_enabled = original
+        self.assertEqual(out.dtype, torch.bfloat16)
+
+    def test_no_autocast_dtypes_pass_through_unchanged(self):
+        # Outside autocast the dispatcher must not touch dtypes: mismatched inputs keep raising
+        # (from SDPA itself for the native backend), exactly as before.
+        query, key, value = self._qkv()
+        with self.assertRaises(RuntimeError):
+            dispatch_attention_fn(query, key, value, backend=AttentionBackendName.NATIVE)
+
+    @require_torch_accelerator
+    def test_rms_norm_qk_pattern_under_autocast(self):
+        # End-to-end shape of the Flux-family bug: bf16 q/k go through `torch.nn.RMSNorm` under
+        # autocast and come out fp32, value stays bf16. The fp32 autocast policy for `aten::rms_norm`
+        # only registers for CUDA/XPU and only on recent torch, so probe for it instead of assuming.
+        device_type = torch.device(torch_device).type
+        if device_type not in ("cuda", "xpu"):
+            self.skipTest("aten::rms_norm only registers an fp32 autocast policy for CUDA/XPU.")
+
+        norm = torch.nn.RMSNorm(16, eps=1e-6, device=torch_device, dtype=torch.bfloat16)
+        query, key, value = self._qkv(qk_dtype=torch.bfloat16)
+
+        with torch.autocast(device_type, dtype=torch.bfloat16):
+            if norm(query).dtype != torch.float32:
+                self.skipTest("this torch version has no fp32 autocast policy for aten::rms_norm.")
+
+        with self._record_native_backend_input_dtypes() as received:
+            with torch.autocast(device_type, dtype=torch.bfloat16):
+                query, key = norm(query), norm(key)
+                out = dispatch_attention_fn(query, key, value, backend=AttentionBackendName.NATIVE)
+
+        self.assertEqual(received["query"], torch.bfloat16)
+        self.assertEqual(received["key"], torch.bfloat16)
+        self.assertEqual(received["value"], torch.bfloat16)
+        self.assertEqual(out.dtype, torch.bfloat16)
+
+    @is_torch_compile
+    def test_torch_compile_fullgraph_under_autocast(self):
+        # The autocast branch must stay fullgraph-traceable — probing autocast state with APIs that
+        # Dynamo cannot trace would graph-break every compiled model at each attention call.
+        # `backend` is left unset (default active backend, i.e. native): passing the enum explicitly
+        # takes the `AttentionBackendName(backend)` path, which Dynamo cannot trace before torch 2.12,
+        # and model processors compile with `backend=None` anyway.
+        query, key, value = self._qkv(qk_dtype=torch.bfloat16)
+        device_type = torch.device(torch_device).type
+
+        compiled = torch.compile(dispatch_attention_fn, fullgraph=True)
+        with torch.autocast(device_type, dtype=torch.bfloat16):
+            out = compiled(query, key, value)
+        self.assertEqual(out.dtype, torch.bfloat16)
+
+        torch.compiler.reset()
+        compiled = torch.compile(dispatch_attention_fn, fullgraph=True)
+        out = compiled(query, key, value)
+        self.assertEqual(out.dtype, torch.bfloat16)
--- a/tests/models/test_attention_processor.py
+++ b/tests/models/test_attention_processor.py
@@ -141,17 +141,20 @@ class DeprecatedAttentionBlockTests(unittest.TestCase):
         self.assertTrue(np.allclose(conversion, after_conversion, atol=1e-3))
 
 
+@require_torch_accelerator
 class TestAttentionDispatchAutocast:
     """Regression tests for https://github.com/huggingface/diffusers/issues/14104.
 
     Under an active torch.autocast context, ops with an fp32 cast policy (e.g. the `torch.nn.RMSNorm`
     QK norms in Flux-family models) return float32 while `value` keeps the autocast dtype. Native SDPA
     downcasts all of its inputs itself, but backends calling external kernels (flash-attn, sage, ...)
-    receive the tensors as-is and reject the mismatch — so `dispatch_attention_fn` must normalize
+    receive the tensors as-is and reject the mismatch, so `dispatch_attention_fn` must normalize
     dtypes the same way SDPA's autocast registration does.
     """
 
     def _qkv(self, qk_dtype=torch.float32, v_dtype=torch.bfloat16):
+        # The fp32 q/k + bf16 v defaults reproduce what the fp32 autocast policy for `aten::rms_norm`
+        # does to the QK-norm pattern: q/k leave the norm upcast, value never passes through it.
         query = torch.randn(1, 8, 2, 16, device=torch_device, dtype=qk_dtype)
         key = torch.randn(1, 8, 2, 16, device=torch_device, dtype=qk_dtype)
         value = torch.randn(1, 8, 2, 16, device=torch_device, dtype=v_dtype)
@@ -174,7 +177,10 @@ class TestAttentionDispatchAutocast:
         finally:
             _AttentionBackendRegistry._backends[AttentionBackendName.NATIVE] = original
 
-    def test_autocast_casts_inputs_to_autocast_dtype(self):
+    def test_autocast_casts_inputs_to_autocast_dtype(self, monkeypatch):
+        # Checks on: the opt-in DIFFUSERS_ATTN_CHECKS validation sees the tensors the backend will
+        # actually receive, so it must not raise on mixed autocast inputs (on main it false-raised).
+        monkeypatch.setattr(_AttentionBackendRegistry, "_checks_enabled", True)
         query, key, value = self._qkv()
         device_type = torch.device(torch_device).type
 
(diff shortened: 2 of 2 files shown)


src/diffusers/models/attention_dispatch.py
```diff
@@ -408,6 +408,26 @@ def dispatch_attention_fn(
...
+    # before running. Under an active autocast context, ops with an fp32 cast policy (such as the
+    # `torch.nn.RMSNorm` QK norms in Flux-family models) return float32 even for bf16/fp16 inputs while
+    # `value` keeps the autocast dtype, and backends that call external kernels (flash-attn, sage,
+    # xformers, ...) bypass autocast's dispatch handling and fail on such inputs. Normalizing here keeps
+    # every backend consistent with the native backend. See https://github.com/huggingface/diffusers/issues/14104.
+    # `is_autocast_available` guards device types without an autocast dispatch key (e.g. "meta"), for
+    # which `is_autocast_enabled` raises. It is skipped when compiling: Dynamo cannot trace it before
+    # torch 2.12, and compiled graphs never run on such devices anyway.
+    device_type = query.device.type
+    if (torch.compiler.is_compiling() or torch.amp.is_autocast_available(device_type)) and torch.is_autocast_enabled(
```
sayakpaul: Why does it have to depend on `torch.compiler.is_compiling()`?

tests/models/test_attention_processor.py
```diff
@@ -133,3 +139,134 @@ def test_conversion_when_using_device_map(self):
 
         self.assertTrue(np.allclose(pre_conversion, conversion, atol=1e-3))
         self.assertTrue(np.allclose(conversion, after_conversion, atol=1e-3))
+
+
+class AttentionDispatchAutocastTests(unittest.TestCase):
```
sayakpaul: This should be implemented with `pytest`.

sayakpaul (commented):
Left two high-level comments.

src/diffusers/models/attention_dispatch.py
danielxmed: Two stacked constraints. In eager, `torch.is_autocast_enabled(device_type)` raises for device types without an autocast dispatch key ("meta" is the one diffusers actually hits, e.g. during meta-device init), so it needs the `torch.amp.is_autocast_available(device_type)` guard. But Dynamo cannot trace that guard before torch 2.12 (it only entered the trace rules there): on torch 2.6, the floor in setup.py, compiling this function with `fullgraph=True` fails with

```
Unsupported: Graph break due to unsupported builtin torch._C._is_autocast_available
```

`torch.compiler.is_compiling()` resolves both sides: it folds to a constant inside the graph, so compiled code skips the untraceable guard, and nothing is lost because compiled graphs never execute on meta anyway. Eager keeps the guard. `test_torch_compile_fullgraph_under_autocast` locks the fullgraph behavior in; verified on 2.6.0 and 2.12.1.

Once the minimum torch reaches 2.12 the short-circuit can be dropped and the condition becomes just `is_autocast_available(...) and is_autocast_enabled(...)`. Happy to restructure if you would rather have it factored into a small helper.

tests/models/test_attention_processor.py
danielxmed: Done in d26452b84: dropped `unittest.TestCase` for plain asserts, `pytest.raises`/`pytest.skip`, and the `monkeypatch` fixture for the checks toggle. Renamed the class to `TestAttentionDispatchAutocast` so default pytest collection picks it up. Re-ran on torch 2.6.0 and 2.12.1 (CPU): 5 passed, 1 GPU-skip on both.

src/diffusers/models/attention_dispatch.py
sayakpaul: This all feels very very LLM-ish. Could you please cut the jargon and explain things out of your own understanding? It's just easier for everyone to follow. Otherwise, it's too much unnecessary cognitive load.

danielxmed: Sure, plain version: `torch.is_autocast_enabled` crashes on device types without autocast support ("meta" is the real case). The `is_autocast_available` check protects against that, but torch 2.6, our minimum, can't trace that check under `torch.compile(fullgraph=True)`. So `is_compiling()` skips it inside compiled code, which is safe: compiled graphs never execute on meta tensors. Eager keeps the protection, compile skips the untraceable call. Once the minimum torch is 2.12, this collapses to just the two autocast calls.

I also tightened the comment block in the code itself (049feefb3), same story in fewer lines.

One note: happy to simplify whenever it helps, and a plain "please make this simpler" does the job. The "LLM-ish" framing wasn't necessary.

sayakpaul: So, we are saying:

> Torch SDPA casts its inputs to the autocast dtype internally, but backends calling external kernels

How does SDPA handle it in a manner that doesn't offend Dynamo? 

> torch.is_autocast_enabled crashes on device types without autocast support ("meta" is the real case).

But is there any real merit to consider that? One would probably never initialize a model on a meta device and then switch between attention backends, I think. If we consider that, does the following still apply?

> but torch 2.6, our minimum, can't trace that check under torch.compile(fullgraph=True).

tests/models/test_attention_processor.py
```diff
@@ -133,3 +139,130 @@ def test_conversion_when_using_device_map(self):
...
+                received["attn_mask"] = kwargs["attn_mask"].dtype
+            return original(query, key, value, **kwargs)
+
+        _AttentionBackendRegistry._backends[AttentionBackendName.NATIVE] = recording_backend
+        try:
+            yield received
+        finally:
+            _AttentionBackendRegistry._backends[AttentionBackendName.NATIVE] = original
+
+    def test_autocast_casts_inputs_to_autocast_dtype(self):
```
sayakpaul: I would mark these tests with `require_torch_accelerator` because I don't think they need to run on our fast PR CI.

tests/models/test_attention_processor.py
```diff
@@ -133,3 +139,130 @@ def test_conversion_when_using_device_map(self):
...
+    """Regression tests for https://github.com/huggingface/diffusers/issues/14104.
+
+    Under an active torch.autocast context, ops with an fp32 cast policy (e.g. the `torch.nn.RMSNorm`
+    QK norms in Flux-family models) return float32 while `value` keeps the autocast dtype. Native SDPA
+    downcasts all of its inputs itself, but backends calling external kernels (flash-attn, sage, ...)
+    receive the tensors as-is and reject the mismatch — so `dispatch_attention_fn` must normalize
+    dtypes the same way SDPA's autocast registration does.
+    """
+
+    def _qkv(self, qk_dtype=torch.float32, v_dtype=torch.bfloat16):
```
sayakpaul, line 155: Is the `qk_type` set to `torch.float32` to emulate the QKnorm behaviour?

tests/models/test_attention_processor.py
```diff
@@ -133,3 +139,130 @@ def test_conversion_when_using_device_map(self):
...
+    def test_autocast_passes_backend_dtype_checks(self, monkeypatch):
+        # The opt-in debugging checks (DIFFUSERS_ATTN_CHECKS) validate the tensors the backend will
+        # actually receive, so mixed autocast inputs must not trip them.
+        query, key, value = self._qkv()
+        device_type = torch.device(torch_device).type
+
+        monkeypatch.setattr(_AttentionBackendRegistry, "_checks_enabled", True)
+        with torch.autocast(device_type, dtype=torch.bfloat16):
+            out = dispatch_attention_fn(query, key, value, backend=AttentionBackendName.NATIVE)
+        assert out.dtype == torch.bfloat16
```
sayakpaul: This is already tested in `test_autocast_casts_inputs_to_autocast_dtype`

tests/models/test_attention_processor.py
```diff
@@ -133,3 +139,130 @@ def test_conversion_when_using_device_map(self):
...
+
+    def test_no_autocast_dtypes_pass_through_unchanged(self):
+        # Outside autocast the dispatcher must not touch dtypes: mismatched inputs keep raising
+        # (from SDPA itself for the native backend), exactly as before.
+        query, key, value = self._qkv()
+        with pytest.raises(RuntimeError):
+            dispatch_attention_fn(query, key, value, backend=AttentionBackendName.NATIVE)
```
sayakpaul: I would even be specific and check that there's a dtype mismatch error in the `RuntimeError` message (assuming that's the case).

tests/models/test_attention_processor.py
```diff
@@ -133,3 +139,130 @@ def test_conversion_when_using_device_map(self):
...
+
+    @require_torch_accelerator
+    def test_rms_norm_qk_pattern_under_autocast(self):
+        # End-to-end shape of the Flux-family bug: bf16 q/k go through `torch.nn.RMSNorm` under
+        # autocast and come out fp32, value stays bf16. The fp32 autocast policy for `aten::rms_norm`
+        # only registers for CUDA/XPU and only on recent torch, so probe for it instead of assuming.
+        device_type = torch.device(torch_device).type
+        if device_type not in ("cuda", "xpu"):
+            pytest.skip("aten::rms_norm only registers an fp32 autocast policy for CUDA/XPU.")
```
sayakpaul: (nit): Would use `pytest.mark.skipif` and use `if torch_device not in ["cuda", "xpu"]` as the `reason`.

tests/models/test_attention_processor.py
```diff
@@ -133,3 +139,130 @@ def test_conversion_when_using_device_map(self):
...
+                query, key = norm(query), norm(key)
+                out = dispatch_attention_fn(query, key, value, backend=AttentionBackendName.NATIVE)
+
+        assert received["query"] == torch.bfloat16
+        assert received["key"] == torch.bfloat16
+        assert received["value"] == torch.bfloat16
+        assert out.dtype == torch.bfloat16
+
+    @is_torch_compile
+    def test_torch_compile_fullgraph_under_autocast(self):
```
sayakpaul, line 246: Cool test!

sayakpaul (commented):
Left further comments and apologies if my statement on LLM-ish framing came off as offensive. Thanks for speaking your mind!

src/diffusers/models/attention_dispatch.py
danielxmed: Fair challenge, and it made the code simpler.

On SDPA: it never checks anything at the Python level. `scaled_dot_product_attention` has autocast kernels registered in C++ (`torch._C._dispatch_dump("aten::scaled_dot_product_attention")` lists AutocastCUDA/CPU/XPU/MPS entries from `aten/src/ATen/autocast_mode.cpp`), so the dispatcher casts the inputs before Dynamo ever sees a branch. When AOTAutograd traces through the dispatcher, the casts just show up in the graph as ordinary ops. Our cast lives in Python, which is why traceability became a topic at all.

On meta: agreed that nobody switches attention backends on a meta model. I swept the codebase to be sure, and every meta usage in src/ and tests/ is weight init or offload bookkeeping, none reaches attention. The catch is that `torch.is_autocast_enabled("meta")` raises whether autocast is active or not, so an unguarded call would break plain meta-device forwards (shape or memory tracing) that work on main today. That protection doesn't require `is_autocast_available`, though. A string check covers it, and the guard is now

```python
if device_type != "meta" and torch.is_autocast_enabled(device_type):
```

Which also answers the last question: with `is_autocast_available` gone there is nothing left that Dynamo can't trace. `is_autocast_enabled` and `get_autocast_dtype` compile fine under `fullgraph=True` on torch 2.6 (zero graph breaks per `torch._dynamo.explain`, correct recompile when autocast toggles between calls), so `is_compiling()` is gone too. Done in 38a870dfc; re-validated on an A4000 (torch 2.12.1+cu126): flash_hub under cuda autocast passes, the full test class is green on CUDA, and the fullgraph test passes on 2.6.0 and 2.12.1.

tests/models/test_attention_processor.py
danielxmed: Done in e864233a9. The class is gated now, so all of it skips on the CPU-only fast PR CI and runs on the GPU workflows instead.

tests/models/test_attention_processor.py
danielxmed: Yes, exactly. fp32 q/k with bf16 v is what the `aten::rms_norm` policy produces on the QK-norm path, and value keeps the autocast dtype because it never passes through the norm. The helper now says this in a comment (e864233a9).

tests/models/test_attention_processor.py
danielxmed: They overlapped on the happy path, agreed. Merged in e864233a9: the checks toggle now lives inside `test_autocast_casts_inputs_to_autocast_dtype` and the standalone test is gone. What it guarded is the opt-in `DIFFUSERS_ATTN_CHECKS` validation, which runs before the backend is invoked and false-raised under autocast on main; keeping the toggle enabled in the surviving test covers that with one test fewer.

tests/models/test_attention_processor.py
danielxmed: Done in e864233a9: `match="Expected query, key, and value to have the same dtype"`. That's SDPA's shared input validation message, identical on torch 2.6.0 and 2.12.1, and the same on CUDA (checked on an A4000).

tests/models/test_attention_processor.py
danielxmed: Done in e864233a9. The runtime probe right below it stays: the fp32 policy also depends on the torch version (new in 2.12), so the version half can't be a static skipif.

danielxmed:
No offense taken, and thanks for saying it. All good.

Round summary, everything pushed:

- The guard is now `device_type != "meta" and torch.is_autocast_enabled(device_type)`. Both `torch.compiler.is_compiling()` and `torch.amp.is_autocast_available` are gone; reasoning in [the thread](https://github.com/huggingface/diffusers/pull/14160#discussion_r3559203249).
- Tests: class gated behind `require_torch_accelerator` (fast PR CI skips it), the checks test merged into the first test, `pytest.raises` asserts the exact dtype-mismatch message, `skipif` nit applied.
- Re-validated on an RTX A4000 (torch 2.12.1+cu126): flash_hub forward under cuda autocast passes with max |flash_hub - native| < 1e-5, the full class is green on CUDA, meta-device forward keeps working, and the fullgraph compile test passes on torch 2.6.0 and 2.12.1.

PR #14162, opened 2026-07-10
torch.dtype alias

stevhliu:
Alias `dtype` to `torch.dtype` so our Hub snippets work properly (see https://github.com/huggingface/huggingface.js/pull/2280 for more details)

### DIFF
--- a/tests/models/testing_utils/common.py
+++ b/tests/models/testing_utils/common.py
@@ -382,12 +382,12 @@ class ModelTesterMixin(BaseModelOutputMixin):
     @pytest.mark.parametrize("dtype", [torch.float32, torch.float16, torch.bfloat16], ids=["fp32", "fp16", "bf16"])
     def test_from_pretrained_dtype_alias(self, tmp_path, dtype):
         # `dtype` is an alias for `torch_dtype` in `from_pretrained`.
+        if torch_device == "mps" and dtype == torch.bfloat16:
+            pytest.skip(reason=f"{dtype} is not supported on {torch_device}")
+
         model = self.model_class(**self.get_init_dict())
         model.to(torch_device)
         model.eval()
-
-        if torch_device == "mps" and dtype == torch.bfloat16:
-            pytest.skip(reason=f"{dtype} is not supported on {torch_device}")
 
         model.to(dtype)
         model.save_pretrained(tmp_path)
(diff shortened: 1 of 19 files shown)


sayakpaul (commented):
Hmm, should we consider deprecating `torch_dtype` following what `transformers` did? @DN6 @yiyixuxu WDYT?

PR looks alright to me (and it also unblock and internal collaboration).

DN6:
PR looks good to me. I would just add a basic tests similar to these ones into the base [ModelTesterMixin](https://github.com/huggingface/diffusers/blob/01969142b55379991fee07608c9e7e8f80afced0/tests/models/testing_utils/common.py#L365) and [PipelineTesterMixin](https://github.com/huggingface/diffusers/blob/01969142b55379991fee07608c9e7e8f80afced0/tests/pipelines/test_pipelines_common.py#L1568) to verify the alias works as excepted. 

@sayakpaul Yeah, we can deprecate `torch_dtype`. I've opened a PR to remove [Jax/Flax](https://github.com/huggingface/diffusers/pull/14169), so it makes sense to consolidate to just `dtype`.

sayakpaul:
@DN6 did you want to open a PR?

tests/models/testing_utils/common.py
```diff
@@ -379,6 +379,21 @@ def test_from_save_pretrained_dtype(self, tmp_path, dtype):
...
 
+    @pytest.mark.parametrize("dtype", [torch.float32, torch.float16, torch.bfloat16], ids=["fp32", "fp16", "bf16"])
+    def test_from_pretrained_dtype_alias(self, tmp_path, dtype):
+        # `dtype` is an alias for `torch_dtype` in `from_pretrained`.
+        model = self.model_class(**self.get_init_dict())
+        model.to(torch_device)
+        model.eval()
+
+        if torch_device == "mps" and dtype == torch.bfloat16:
+            pytest.skip(reason=f"{dtype} is not supported on {torch_device}")
```
sayakpaul: We could move this block at the beginning of this method.

tests/pipelines/test_pipelines_common.py
```diff
@@ -2408,6 +2408,26 @@ def test_torch_dtype_dict(self):
                     f"Component '{name}' has dtype {component.dtype} but expected {expected_dtype}",
                 )
 
+    def test_dtype_alias(self):
+        # `dtype` is an alias for `torch_dtype` in `from_pretrained`.
+        components = self.get_dummy_components()
+        if not components:
+            self.skipTest("No dummy components defined.")
```
sayakpaul: We don't need this one because `components` is rarely `None`.

sayakpaul (approved):

sayakpaul:
@stevhliu some pipeline-level tests seem to be failing.

PR #14164, opened 2026-07-10
Add a sampling animation example for discrete diffusion

kashif:
Adds `examples/discrete_diffusion/animate_sampling.py`, a small tool to visualize how discrete diffusion language models write text.

It captures the intermediate canvas of every denoising step through each pipeline's `callback_on_step_end` (no changes to the pipelines) and renders it as a self-contained HTML animation or an animated GIF. The goal is to make the two families easy to compare: LLaDA2 reveals text from `[MASK]` slots, while DiffusionGemma sharpens a mask-free canvas of random tokens block by block.

For DiffusionGemma you can pick any of the three supported schedulers (`entropy_bound`, `discrete_ddim`, `block_refinement`) to compare them on the same prompt and seed. Trajectories with different step counts are resampled onto a shared timeline so the panels stay in sync, and the highlight colors are color-blind and grayscale safe.

The README has a new section with capture and render commands. The script is also runnable with `uv`.


kashif:
diffusion_gemma_sampling

PR #8391, opened 2024-06-03
ensure dtype match between diffused latents and vae weights

heyalexchoi:
# What does this PR do?

Simple fix to diffused latent dtype not matching vae weights dtype. See error below. I had this issue when loading pipeline in bfloat16 and using accelerate.

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/workspace/PixArt-sigma/diffusion/utils/image_evaluation.py", line 150, in generate_images
    batch_images = pipeline(
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/workspace/diffusers/src/diffusers/pipelines/pixart_alpha/pipeline_pixart_sigma.py", line 866, in __call__
    image = self.vae.decode(latents / self.vae.config.scaling_factor, return_dict=False)[0]
  File "/workspace/diffusers/src/diffusers/utils/accelerate_utils.py", line 46, in wrapper
    return method(self, *args, **kwargs)
  File "/workspace/diffusers/src/diffusers/models/autoencoders/autoencoder_kl.py", line 305, in decode
    decoded = self._decode(z, return_dict=False)[0]
  File "/workspace/diffusers/src/diffusers/models/autoencoders/autoencoder_kl.py", line 277, in _decode
    z = self.post_quant_conv(z)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/conv.py", line 460, in forward
    return self._conv_forward(input, self.weight, self.bias)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/conv.py", line 456, in _conv_forward
    return F.conv2d(input, weight, bias, self.stride,
RuntimeError: Input type (float) and bias type (c10::BFloat16) should be the same
```

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
[...]

PR #14184, opened 2026-07-14
Fix AuraFlow VAE dtype mismatch on pipeline reuse

### LINKED ISSUE #14183: AuraFlowPipeline: VAE dtype mismatch on pipeline reuse (latents cast skipped on 2nd call)
### Describe the bug

`AuraFlowPipeline.__call__` raises a dtype-mismatch `RuntimeError` in the VAE
decode when the same pipeline instance is called a **second** time. A single
call succeeds; the error only appears on reuse (e.g. a warmup call followed by a
real call).

Root cause is in the decode block:

```python
needs_upcasting = self.vae.dtype == torch.float16 and self.vae.config.force_upcast
if needs_upcasting:
    self.upcast_vae()                     # casts the ENTIRE vae to float32 in place
    latents = latents.to(next(iter(self.vae.post_quant_conv.parameters())).dtype)
image = self.vae.decode(latents / self.vae.config.scaling_factor, return_dict=False)[0]
```

`upcast_vae()` upcasts the whole VAE to float32 **in place**, so after the first
call `self.vae.dtype == torch.float32` permanently. On the second call
`needs_upcasting` is therefore `False`, the `latents.to(...)` line is skipped,
and float16 latents are fed to the now-float32 VAE:

```
RuntimeError: Input type (c10::Half) and bias type (float) should be the same
```

This is the same class of bug fixed for `pipeline_pixart_sigma.py` in #8391 —
that fix cast the latents to the VAE dtype unconditionally at the decode call.
The equivalent line in `pipeline_aura_flow.py` still guards the cast behind
`if needs_upcasting:`.

### Reproduction

```python
import torch
from diffusers import AuraFlowPipeline

pipe = AuraFlowPipeline.from_pretrained("fal/AuraFlow", torch_dtype=torch.float16).to("cuda")
gen = lambda: pipe(prompt=prompt, height=512, width=512, num_inference_steps=50,
                   guidance_scale=3.5, generator=torch.Generator("cuda").manual_seed(42))

gen()   # 1st call: OK (needs_upcasting True, latents cast runs, vae left in fp32)
gen()   # 2nd call: RuntimeError (needs_upcasting now False, latents stay fp16)
```

### Suggested fix

Apply the same pattern as #8391 — cast the latents to the VAE dtype inline at the
decode call so it always runs regardless of `needs_upcasting`:

```python
if needs_upcasting:
    self.upcast_vae()
image = self.vae.decode(latents.to(self.vae.dtype) / self.vae.config.scaling_factor, return_dict=False)[0]
```

### Verification

Applied the suggested fix to a checkout at commit `33becabe`, with **no** other
patches, and ran warmup + 1 run (the 2nd call is what triggers the bug). Both
produce a coherent image and no dtype error:

| Environment | Warmup | 2nd call | Result |
| --- | --- | --- | --- |
| NVIDIA RTX 5090, CUDA (torch 2.13 dev) | 18.51s | 7.31s | Pass |
| AMD Radeon (gfx1151), ROCm (torch 2.11) | 64.23s | 63.25s | Pass |

The offending line is unchanged on current `main`.

IvenHsu01:
## What does this PR do?

Fixes a VAE dtype-mismatch `RuntimeError` in `AuraFlowPipeline` that occurs when
the same pipeline instance is called more than once.

`upcast_vae()` upcasts the whole VAE to float32 in place, so on the second call
`needs_upcasting` is `False` and the `latents.to(...)` cast (guarded by that
`if`) is skipped, feeding fp16 latents to the fp32 VAE. This applies the same
pattern already merged for `pixart_sigma` in #8391 — cast the latents to the VAE
dtype inline at the decode call so it always runs.

```diff
             if needs_upcasting:
                 self.upcast_vae()
-                latents = latents.to(next(iter(self.vae.post_quant_conv.parameters())).dtype)
-            image = self.vae.decode(latents / self.vae.config.scaling_factor, return_dict=False)[0]
+            image = self.vae.decode(latents.to(self.vae.dtype) / self.vae.config.scaling_factor, return_dict=False)[0]
```

Companion to issue #14183, which has the full analysis and verification.
Happy to adjust the scope or approach based on maintainer feedback there.

## Coordination

Fixes #14183

## Tests run

Applied this exact change, no other patches, ran warmup + 1 run (the 2nd call
triggers the bug) at 512x512, 50 steps, guidance 3.5, seed 42:

- NVIDIA RTX 5090 (CUDA): warmup 18.51s, 2nd call 7.31s — Pass, coherent image
- AMD gfx1151 (ROCm): warmup 64.23s, 2nd call 63.25s — Pass, coherent image

## Before submitting
- [x] Read the [contributor guideline](https://github.com/huggingface/diffusers/blob/main/CONTRIBUTING.md)


klin2024:
@dg845 friendly ping, tiny one-line bugfix — could you take a look when you get a chance? Thanks!

dg845 (approved):
Thanks for the PR!

PR #14191, opened 2026-07-15
[skills] add notes about model implementation in our skills.

sayakpaul:
Add notes about the following in our model implementation skill:

* Prefer using fuller and more descriptive names for variables.
* A note on contiguity (example: https://github.com/huggingface/diffusers/pull/14186)


yiyixuxu (approved):
thanks

PR #14198, opened 2026-07-15
fix: allow local LoRA paths in offline mode

### LINKED ISSUE #14195: Offline mode rejects local LoRA files and directories without weight_name
### Describe the bug

When `HF_HUB_OFFLINE=1` is enabled, or `local_files_only=True` is passed,
loading LoRA weights from an existing local file or directory without an
explicit `weight_name` fails with:

`ValueError: When using the offline mode, you must specify a weight_name.`

Offline mode should prevent HTTP requests, but it should not prevent Diffusers
from inspecting an explicitly provided local filesystem path.

The issue is in `_best_guess_weight_name()`:

https://github.com/huggingface/diffusers/blob/1aadc65cacf94515e9c9a040def3e9a9e14d32c5/src/diffusers/loaders/lora_base.py#L277-L290

The function checks `local_files_only` / `HF_HUB_OFFLINE` before checking
`os.path.isfile()` and `os.path.isdir()`. As a result, existing local paths are
rejected before they can be inspected.

I understand that #6184 intentionally requires an explicit `weight_name` for a
remote repository ID in offline mode. Without access to `model_info()`,
Diffusers cannot automatically determine which file exists in a Hub repository.
This report does not propose changing that behavior.

This is related to #6089 and #6110 and follows up on the offline guard
introduced in #6184. Unlike the cached-repository case discussed there, this
issue only concerns inputs that already resolve to an explicit local file or
directory.

Expected behavior:

- An explicit local LoRA file should work in offline mode.
- An explicit local directory should support discovery of its LoRA weight file.
- A remote repository ID without `weight_name` should continue to raise in
  offline mode.
- `model_info()` should not be called in offline mode.

I am willing to submit a focused PR that moves the local file and directory
checks before the offline-mode guard and adds regression tests using temporary
local files without downloading any models.

Would this distinction and proposed scope be acceptable for a PR?

### `HF_HUB_OFFLINE=1`

Run this in a fresh Python process:

```python
import os

os.environ["HF_HUB_OFFLINE"] = "1"

import tempfile
from pathlib import Path

import torch

from diffusers.loaders import StableDiffusionLoraLoaderMixin


    weight_path = Path(tmpdir) / "adapter.safetensors"
    save_file(
        {"unet.test.lora_A.weight": torch.ones(1)},
        str(weight_path),
    )

    StableDiffusionLoraLoaderMixin.lora_state_dict(tmpdir)
```

### `local_files_only=True`

The same failure occurs with `HF_HUB_OFFLINE` disabled and only
`local_files_only=True`. Run this separately in a fresh Python process:

```python
import os

os.environ.pop("HF_HUB_OFFLINE", None)

import tempfile
from pathlib import Path

import torch

from diffusers.loaders import StableDiffusionLoraLoaderMixin


    weight_path = Path(tmpdir) / "adapter.safetensors"
    save_file(
        {"unet.test.lora_A.weight": torch.ones(1)},
        str(weight_path),
    )

    StableDiffusionLoraLoaderMixin.lora_state_dict(
        tmpdir,
        local_files_only=True,
    )
```

Both examples use a small temporary safetensors file and do not download any
models.

### Logs

```shell
  File "<stdin>", line 21, in <module>
  File ".../huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
  File ".../diffusers/loaders/lora_pipeline.py", line 318, in lora_state_dict
    state_dict, metadata = _fetch_state_dict(
  File ".../diffusers/loaders/lora_base.py", line 224, in _fetch_state_dict
    weight_name = _best_guess_weight_name(
  File ".../diffusers/loaders/lora_base.py", line 281, in _best_guess_weight_name
    raise ValueError("When using the offline mode, you must specify a `weight_name`.")
ValueError: When using the offline mode, you must specify a `weight_name`.
```

### System Info

- 🤗 Diffusers version: 0.40.0.dev0
- Reproduced on main commit: 1aadc65cacf94515e9c9a040def3e9a9e14d32c5
- Platform: Linux-5.4.210-39.1.pagevecsize-x86_64-with-glibc2.39
- Python version: 3.12.3
- PyTorch version (GPU?): 2.6.0+cpu (False)
- Jax version: not installed
- Huggingface_hub version: 0.36.2
- Transformers version: 4.57.6
- Accelerate version: 1.14.0
- PEFT version: 0.17.1
- Safetensors version: 0.8.0
- Using GPU in script?: No

### Who can help?

@sayakpaul

pctablet505:
Fixes #14195

Moves the local file and directory checks in `_best_guess_weight_name()` before the offline-mode guard, so explicit local LoRA files and directories can be loaded when `local_files_only=True` or `HF_HUB_OFFLINE=1` is set. Remote repository IDs without an explicit `weight_name` continue to raise in offline mode, and `model_info()` is still not called.

### DIFF
--- a/src/diffusers/loaders/lora_base.py
+++ b/src/diffusers/loaders/lora_base.py
@@ -277,15 +277,14 @@ def _fetch_state_dict(
 def _best_guess_weight_name(
     pretrained_model_name_or_path_or_dict, file_extension=".safetensors", local_files_only=False
 ):
-    if local_files_only or HF_HUB_OFFLINE:
-        raise ValueError("When using the offline mode, you must specify a `weight_name`.")
-
     targeted_files = []
 
     if os.path.isfile(pretrained_model_name_or_path_or_dict):
         return
     elif os.path.isdir(pretrained_model_name_or_path_or_dict):
         targeted_files = [f for f in os.listdir(pretrained_model_name_or_path_or_dict) if f.endswith(file_extension)]
+    elif local_files_only or HF_HUB_OFFLINE:
+        raise ValueError("When using the offline mode, you must specify a `weight_name`.")
     else:
         files_in_repo = model_info(pretrained_model_name_or_path_or_dict).siblings
         targeted_files = [f.rfilename for f in files_in_repo if f.rfilename.endswith(file_extension)]
--- a/tests/lora/test_lora_base.py
+++ b/tests/lora/test_lora_base.py
@@ -1,0 +1,88 @@
+# coding=utf-8
+# Copyright 2026 HuggingFace Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+import tempfile
+import unittest
+from pathlib import Path
+from unittest.mock import patch
+
+import torch
+from safetensors.torch import save_file
+
+from diffusers.loaders import StableDiffusionLoraLoaderMixin
+from diffusers.loaders.lora_base import _best_guess_weight_name
+
+
+class LoraBaseBestGuessWeightNameTests(unittest.TestCase):
+    def test_local_file_with_local_files_only_returns_none(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            weight_path = Path(tmpdir) / "adapter.safetensors"
+            save_file({"unet.test.lora_A.weight": torch.ones(1)}, str(weight_path))
+
+            weight_name = _best_guess_weight_name(str(weight_path), local_files_only=True)
+            self.assertIsNone(weight_name)
+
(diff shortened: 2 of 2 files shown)


fropych:
Thanks for looking into this. I opened #14195 to coordinate the fix and already have a local branch and commit ready, including focused regression tests. This would be my first open-source contribution, so I was hoping to take the implementation through the PR and review process myself.
Since the issue has not yet received maintainer acknowledgment, @sayakpaul, could you please advise whether I should open my PR or coordinate with #14198? I’m happy to collaborate and would like to avoid duplicate work.

sayakpaul:
@pctablet505 can you maintain the PR template and acknowledge how much AI assistance you had in this PR?

fropych:
@sayakpaul, I would like to clarify the coordination history here.
In #14195, I explicitly said that I wanted to implement this fix and asked for confirmation before opening a PR. Since my work was AI-assisted, I followed the contribution guide and waited for maintainer acknowledgment.
While I was waiting, #14198 was opened by another contributor with essentially the same fix, without coordinating with me.
I have now opened #14204. This would be my first open-source contribution, and I was hoping to take it through the review process myself.
My PR:
- follows the full PR template
- clearly discloses the AI assistance
- includes the self-review notes and actual check results
- covers every regression case described in the issue
- verifies that model_info() is not called in offline mode
- preserves the existing behavior for directories with no matching files or multiple matches.  

One public-API test in #14198 changes os.environ after HF_HUB_OFFLINE has already been imported. As a result, it does not actually exercise that offline path. My test patches the imported constant directly.
I’m happy to collaborate, but I would appreciate it if the coordination history and test coverage could be considered when deciding which PR should move forward.

